PR #74, opened 2023-08-09
Remove foreign key conditions

maxdeichmann:
Goals
- Unordered ingestion of events
- Increased ingestion performance by reducing number of sequential queries


src/pages/api/public/scores.ts
```diff
@@ -117,28 +117,21 @@ export default async function handler(
...
-            },
-          },
-          skip: (obj.page - 1) * obj.limit,
-          take: obj.limit,
-          orderBy: {
-            timestamp: "desc",
-          },
-        }),
+        prisma.$queryRaw<Score & { userIds: string[] }[]>(Prisma.sql`
+          SELECT s.id, s.timestamp, s.name, s.value, s.comment, s.trace_id as "traceId", s.observation_id as "observationId", json_build_object('userId', t.user_id) as "trace"
```
marcklingen, line 126: as `userIds`?

src/pages/api/public/events.ts
```diff
@@ -77,33 +77,32 @@ export default async function handler(
           ).id
         : obj.traceId;
 
-    // CHECK ACCESS SCOPE
-    const accessCheck = await checkApiAccessScope(authCheck.scope, [
-      ...(traceId ? [{ type: "trace" as const, id: traceId }] : []),
```
marcklingen, line 82: Before we hat db-level assertion that observation and trace are in the same project; need to add it (adds latency) or ignore as we have projectId as kind of a tenant id now on the observations

src/pages/api/public/spans.ts
```diff
@@ -51,10 +53,6 @@ export default async function handler(
     });
   // END CHECK AUTH
 
-  if (req.method !== "POST" && req.method !== "PATCH") {
-    return res.status(405).json({ message: "Method not allowed" });
-  }
-
   if (req.method === "POST") {
```
marcklingen, line 56: Suggestion: merge Patch and POST if both upsert

marcklingen: can still have e.g. default starttime if null if method === POST

src/pages/api/public/traces.ts
```diff
@@ -88,11 +90,25 @@ export default async function handler(
...
             projectId: authCheck.scope.projectId,
+            timestamp: new Date(),
+            name: name ?? undefined,
+            userId: userId ?? undefined,
+            metadata: metadata ?? undefined,
+            release: release ?? undefined,
+            version: version ?? undefined,
+          },
+          update: {
+            timestamp: new Date(),
```
marcklingen, line 111: Suggestion: remove timestamp from create and update; set by database on create and should not be updated when upserting the trace (unexpected behavior)

src/server/api/routers/traces.ts
```diff
@@ -26,67 +32,110 @@ export const traceRouter = createTRPCRouter({
...
+          SELECT s.*
+          FROM "scores" AS s
+          WHERE s."trace_id" IN (${traceIds})`
+        ),
+        ctx.prisma.$queryRaw<Observation[]>(
+          Prisma.sql`
+            SELECT o.*
+            FROM "observations" AS o
+            WHERE o."trace_id" IN (${traceIds})`
+        ),
```
marcklingen, line 108: Add token calculation to main sql query? Should be faster & simpler

marcklingen (changes requested):
Added some comments, let's discuss before I make changes

marcklingen (changes requested):
Update openapi spec and sdks before merging (if necessary, e.g. when adding traceid to PATCH observation)

src/server/api/routers/traces.ts
marcklingen: done

marcklingen (approved):

PR #100, opened 2023-08-24
remove overlapping in analytics page

DIWAKARKASHYAP:
from this
image

to this
image

### DIFF
--- a/src/pages/project/[projectId]/analytics.tsx
+++ b/src/pages/project/[projectId]/analytics.tsx
@@ -37,7 +37,7 @@ export default function AnalyticsPage() {
 const AnalyticsDisabled = () => (
   <Alert>
     <Construction className="h-4 w-4" />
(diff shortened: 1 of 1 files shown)


marcklingen:
Thank you for your contribution @DIWAKARKASHYAP, I cannot reproduce the error that you fix. The Alert component apparently includes css to place the icon (svg) in the right absolute spot.

Details
image

DIWAKARKASHYAP:
sorry , actually  i research and then i find that , i need to set "true" this option in my firefox browser "layout.css.has-selector.enabled" . to use is(), has(), where() pseudo-class in CSS

 in future, If anybody face this issue then simply follow these three step 

 1-    Open the `about:config` page by typing about:config in the address bar.
 2-   In the search bar, type `layout.css.has-selector.enabled.`
 3-   Double-click on the `layout.css.has-selector.enabled` setting to change it from `false` to `true`

PR #233, opened 2023-09-18
Datasets (experimentation & testing)

### LINKED ISSUE #196: [LFC-189] Dataset management for testing/experimentation
* Manage set of reference prompts/inputs to run tests on
* Add to the dataset from production traces
* Use within SDKs to run experiments

Workflows from SDK

1. Get all data set items by data set name (input & expected output), all that are not archived
2. User creates run (name) on data set
3. User runs app on each item and traces it with langfuse; needs to provide data set item id and run id (can both be from sdk state maybe) that the execution trace belongs to
4. Optionally: user runs evals and adds scores to the observation or trace

Todo

* [ ] Average score on run

<sub>From [SyncLinear.com](https://synclinear.com) | [LFC-189](https://linear.app/langfuse/issue/LFC-189/dataset-management-for-testingexperimentation)</sub>

marcklingen:
Contributes to #196


src/features/datasets/server/datasetRouter.ts
```diff
@@ -0,0 +1,345 @@
...
+            countRunItems: number;
+            scores: Record<string, number>;
+          }
+        >
+      >(Prisma.sql`
+        WITH avg_scores_by_run_id AS (
+          SELECT
+            ri.dataset_run_id run_id,
+            s.name score_name,
+            AVG(s.value) AS average_score_value
```
maxdeichmann, line 78: FYI: https://www.postgresql.org/docs/current/queries-select-lists.html#QUERIES-COLUMN-LABELS

maxdeichmann: Would keep it as it is for speed.

PR #303, opened 2023-10-05
api changes

maxdeichmann:


fern/server/definition/commons.yml
```diff
@@ -66,10 +66,10 @@ types:
       completionStartTime: optional<datetime>
       model: optional<string>
       modelParameters: optional<map<string, MapValue>>
-      prompt: optional<unknown>
```
maxdeichmann, line 69: This here does not work in our APIs, as we always return input, output

fern/server/definition/generations.yml
```diff
@@ -16,18 +16,6 @@ service:
       path: /generations
       request: UpdateGenerationRequest
       response: commons.Observation
-    get:
```
maxdeichmann, line 19: we could have one generic endpoint to return observations. As far as i know, this is only used in datasets. We should move it to 'observations' now as it is early.

PR #306, opened 2023-10-05
Latencies on tables

marcklingen:


marcklingen:
Next step: check performance impact

Issue #235, opened 2023-09-18
[LFC-202] Add playwright tests (to CI)


marcklingen:
We should start testing the frontend application as well to control for regressions affecting the front-end application.

To get started, I'd suggest to add playwright testing to the tests run on CI (npm run test). Ideas for test cases:

* [X] Init: #242
* [X] Create account #390
* [X] Sign in: #242
* [X] Create new project: #392

<sub>From [SyncLinear.com](https://synclinear.com) | [LFC-202](https://linear.app/langfuse/issue/LFC-202/add-playwright-tests-to-ci)</sub>

marcklingen:
@FrankenDeba Thanks for the contribution on this in #390 and #392. I'll close this for now. If you think there's another E2E test that would be high value, feel free to create a new issue

PR #392, opened 2023-10-21
Add E2E for project creation

FrankenDeba:
Issue# -> https://github.com/langfuse/langfuse/issues/235
I augmented additional E2Es in order to test the "create project" feature.

Now, the opening of the create project form & subsequent creation of a new project with a given name are monitored through this test!


marcklingen (approved):
Looks great, thanks for the contribution

PR #393, opened 2023-10-21
Auth: case-insensitive emails

marcklingen:
Users reported authentication issues when using non-all-lowercase emails.

- [x] Change auth to treat emails case-insensitively to resolve next-auth errors.
- [x] Add an E2E test to check for handling of uppercase emails.
- [x] Migrate emails of users who previously signed up using an uppercase email.


maxdeichmann (approved):

marcklingen:
Migrations applied (cloud)

PR #396, opened 2023-10-23
feat(auth): add AzureAD

marcklingen:


marcklingen:
Not working yet, getting random errors

PR #427, opened 2023-10-31
Docker: signal forwarding & migrations in entrypoint

marcklingen:
Changes

* signal forwarding using dumb-init

* prisma migrations applied in entrypoint

Fixes

* Add SALT to args of docker-compose.build.yml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,8 @@ ARG NEXTAUTH_SECRET
 ARG NEXTAUTH_URL
 ARG SALT
 
+RUN apk add --no-cache dumb-init
+
 WORKDIR /app
 
 ENV NODE_ENV production
@@ -68,6 +70,7 @@ RUN addgroup --system --gid 1001 nodejs
 RUN adduser --system --uid 1001 nextjs
 
 RUN npm install --no-package-lock --no-save cron
+RUN npm install -g --no-package-lock --no-save prisma
 
 COPY --from=builder /app/public ./public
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -79,11 +82,12 @@ COPY --from=builder --chown=nextjs:nodejs /app/prisma ./prisma
...
+RUN chmod +x ./entrypoint.sh
 
 USER nextjs
 
 # Default port to 3000
 ENV PORT 3000
 
 # CMD ["node", "server.js"]
-CMD ["/bin/sh", "entrypoint.sh"]
+CMD ["dumb-init", "--", "./entrypoint.sh"]
```
maxdeichmann, line 93: Was macht `dumb-init`?

maxdeichmann: I think best practice would be to have the migration on the docker-compose level as it is more obvious to an engineer that this migration step is happening with each docker container start. However, i understand it is easier here as we have access to prisma here. Hence i think it is ok.

README.md
```diff
@@ -107,9 +107,6 @@ cd langfuse
 
 # Run server and database
 docker compose up -d
-
-# Apply db migrations
-DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres DIRECT_URL=postgresql://postgres:postgres@localhost:5432/postgres npx prisma migrate deploy
```
maxdeichmann, line 112: We also need to adjust the docs then i assume

marcklingen: Yes. Will do once we merge.

Dockerfile
marcklingen: See description: signal forwarding using dumb-init
- eg ctrl-c or shutdown requests. Nodejs doesn't handle them well when run as PID 1

marcklingen: Goal: migrations work as part of the startup of the container wie package. Needs to also work in the non-docker-compose workflow when deploying the container

PR #442, opened 2023-11-06
add openai models

maxdeichmann:


derino:
@maxdeichmann New models should be added here as well (isOpenAiModel) ;)
https://github.com/langfuse/langfuse/blob/8bb349ce10433a2e2cf7bc4a4cba60e0ce589758/src/features/ingest/lib/usage.ts#L149

maxdeichmann:
> > @maxdeichmann New models should be added here as well (isOpenAiModel) ;)
> > https://github.com/langfuse/langfuse/blob/8bb349ce10433a2e2cf7bc4a4cba60e0ce589758/src/features/ingest/lib/usage.ts#L149
> 
> @derino you are right! Will add that asap.

https://github.com/langfuse/langfuse/pull/459

PR #471, opened 2023-11-10
Feat: latencies on dataset tables

marcklingen:
Implements LF-913


marcklingen:
https://github.com/langfuse/langfuse/assets/2834609/23eae0db-86d3-4094-8254-155294caaec0

PR #473, opened 2023-11-12
feat: add transfer project to new owner

### LINKED ISSUE #457: feat: OWNER of project can reassign ownership
### Describe the feature you'd like to request

User with owner role of project can make another user owner of a project

### Describe the solution you'd like to see

Transfer ownership dialog in project settings (include in danger zone)
- Only accessible by current owner
- Warn user that this will cause themselves to lose ownership

### Additional information

- Only one person can be owner
- It needs to be transactional, one user needs to be owner even if the tx fails
- Only show the dialog to the current owner, use `useHasAccess()` hook
- In the TRPC route, make sure to check if user is owner of the project with `throwIfNoAccess`

### Contribute

- [ ] Yes, I can implement this and raise a PR

<sub>[LFC-218](https://linear.app/langfuse/issue/LFC-218/feat-owner-of-project-can-reassign-ownership)</sub>

DIWAKARKASHYAP:
Resolve #457 
- [x] Only accessible by current owner
- [x] Warn user that this will cause themselves to lose ownership


src/features/rbac/constants/roleAccessRights.ts
```diff
@@ -33,6 +34,7 @@ export const roleAccessRights: Record<MembershipRole, Scope[]> = {
     "scores:CUD",
     "project:delete",
     "datasets:CUD",
+    "project:transfer",
```
maxdeichmann, line 37: would it not make more sense to have 'membership:delete' and 'membership:create' rights here?

src/features/projects/server/projectsRouter.ts
```diff
@@ -67,4 +67,45 @@ export const projectsRouter = createTRPCRouter({
...
+          projectId: input.projectId,
+          role: 'OWNER',
+        },
+      });
+
+      return await ctx.prisma.membership.deleteMany({
+        where: {
+          userId: ctx.session.user.id,
+          projectId: input.projectId,
+          role: 'OWNER',
```
maxdeichmann, line 107: are we sure the user is the "OWNER" here? Would we need to check that above before the transfer?

src/features/projects/components/TransferOwnershipButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <DialogTrigger asChild>
+          <Button variant="destructive">Transfer project</Button>
+        </DialogTrigger>
+        <DialogContent className="sm:max-w-[425px]">
+          <DialogHeader>
+            <DialogTitle className="text-lg font-semibold  ">
+              Transfer project
+            </DialogTitle>
+            <DialogDescription>
+              Remember, you can lose ownership, and you will no longer have access to this project
```
maxdeichmann, line 82: Can ewe rephrase this? We will lose ownership.

src/features/projects/components/TransferOwnershipButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+            </DialogDescription>
+          </DialogHeader>
+          <Form {...form}>
+            <form
+              // eslint-disable-next-line @typescript-eslint/no-misused-promises
+              onSubmit={form.handleSubmit(onSubmit)}
```
maxdeichmann, line 88: should this not be wrapped in a function like `{()=>form.handleSubmit(onSubmit)}`

src/features/projects/server/projectsRouter.ts
```diff
@@ -67,4 +67,45 @@ export const projectsRouter = createTRPCRouter({
...
+      })
+    )
+    .mutation(async ({ input, ctx }) => {
+      throwIfNoAccess({
+        session: ctx.session,
+        projectId: input.projectId,
+        scope: "project:transfer",
+      });
+
+      const user = await ctx.prisma.user.findUnique({
```
maxdeichmann, line 85: can we have a try .. catch block here where we return a 500 with internal error and log the exception to the console?

marcklingen:
On it, I want to get this merged, thanks for the contribution!

src/features/rbac/constants/roleAccessRights.ts
marcklingen: I removed the role, the owner can transfer their own project

marcklingen:
Made some changes to merge this, highlights
- Owner becomes Admin after transfer. This is a transaction to be sure the project is not lost without an owner.
- Danger zone always visible, buttons disabled if user has no access
- Errors are displayed in form to help user debug
- Upsert new owner role to also work when user already has access to project
- ui: add input label and meaningful placeholder

PR #508, opened 2023-11-16
fix: update example environment variable AUTH_DOMAINS_WITH_SSO_ENFORCEMENT

p5:
The example .env files contain an environment variable that does not match the code, and caused a bit of confusion when we tried self-hosting this.

This is a super small PR to update the example environment variable name.

You can see [here](https://github.com/langfuse/langfuse/blob/b42239095d9b36fa405af45071e85bb835b69306/src/env.mjs#L41), the correct name is `AUTH_DOMAINS_WITH_SSO_ENFORCEMENT`, not `AUTH_DOMAINS_WITH_OAUTH_ENFORCEMENT`


marcklingen:
@p5 thanks for spotting this! This has caused some confusion for others as well

PR #514, opened 2023-11-16
feat: delete-single-trace

### LINKED ISSUE #501: feat: delete single trace
### Describe the feature you'd like to request

Delete a single trace

### Describe the solution you'd like to see

- action on tables
- action on single trace view

### Additional information

While usually this is not necessary, there are various reasons why it can be useful to delete a single trace.

### Contribute

- [ ] Yes, I can implement this and raise a PR

TameemAsim:
This PR solves #501 .

After PR:

[Screencast from 11-17-2023 04:40:07 AM.webm](https://github.com/langfuse/langfuse/assets/88766161/199c3fb2-b6fe-4f00-a292-fdaf5bc2ff9f)


src/server/api/routers/traces.ts
```diff
@@ -275,4 +275,37 @@ export const traceRouter = createTRPCRouter({
         observations: enrichedObservations as ObservationReturnType[],
       };
     }),
+  delete: protectedGetTraceProcedure
```
marcklingen, line 278: This also allows public access if trace is published (meant for reading only). Use protectedProjectProcedure which requires the projectId as a param

src/server/api/routers/traces.ts
```diff
@@ -275,4 +275,37 @@ export const traceRouter = createTRPCRouter({
...
+    .input(z.object({ traceId: z.string() }))
+    .mutation(async ({ input, ctx }) => {
+      const trace = await ctx.prisma.trace.findFirstOrThrow({
```
marcklingen, line 281: add a check whether user has a sufficient role to perform this action

src/server/api/routers/traces.ts
```diff
@@ -275,4 +275,37 @@ export const traceRouter = createTRPCRouter({
...
+        where: {
+          id: input.traceId,
```
marcklingen, line 283: Add check whether the traceId is actually in the project the user has access to

src/server/api/routers/traces.ts
```diff
@@ -275,4 +275,37 @@ export const traceRouter = createTRPCRouter({
...
+        },
+        include: {
+          scores: true
+        }
+      })
+
+      await trace.scores.map(async (score) => {
```
marcklingen, line 290: Scores cascade, need to only delete all observations manually

src/components/delete-trace.tsx
```diff
@@ -0,0 +1,48 @@
...
+  icon?: boolean,
+}
+) {
+  const router = useRouter();
+
+  const mutDeleteTrace = api.traces.delete.useMutation();
+
+  const deleteTrace = async () => {
+    await mutDeleteTrace.mutateAsync({traceId: traceId});
+    if (icon) {
```
marcklingen, line 23: if(icon) is unclear, I get why you've implemented it this way but I would rename the icon to eg isTableAction

src/components/delete-trace.tsx
```diff
@@ -0,0 +1,48 @@
...
+      router.reload();
```
marcklingen, line 24: Do not reload but use invalidation of trpc cache, thereby the table updates without a full refresh.

src/components/delete-trace.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  if (icon) {
+    return (
+      <Button variant="ghost" size="xs" onClick={deleteTrace}>
+        <TrashIcon className='w-4 h-4'/>
+      </Button>
+    );
+  };
+
+  return (
```
marcklingen, line 38: Add popover to ask for confirmation, this is a dangerous action
UI: https://ui.shadcn.com/docs/components/popover

marcklingen (changes requested):
Thanks a lot for your contribution @18feb06 

I've made some suggestions as this is a dangerous action (need to check for access rights, ask for confirmation by user) and it manipulates the state (invalidate cache instead of refreshing the application)

src/server/api/routers/traces.ts
18feb06: ok... no problem

src/server/api/routers/traces.ts
18feb06: ok... no problem

src/components/delete-trace.tsx
18feb06: ok... no problem

src/components/delete-trace.tsx
18feb06: ok... no problem  working on all issues

src/components/delete-trace.tsx
18feb06: ok... no problem

src/server/api/routers/traces.ts
18feb06: I have checked scopes array, there is no scope for trace:CUD or trace:delete.
I am thinking to create new scopes. Kindly give any suggestion if you have.
If yes, then tell me that should only admin and owner be allowed to delete traces or member can also delete traces?

src/server/api/routers/traces.ts
marcklingen: Moved this to the sql query

18feb06: ok

marcklingen (approved):
Thanks a lot for the contribution, I made minor tweaks

- delete observations and trace in one transaction
- hide delete button for users who do not have access
- change text on popup
- change invalidation/redirect logic to not cause any erroring trpc calls

PR #515, opened 2023-11-17
feat(scores): types and new expert score form

marcklingen:
- Score types: user, expert, eval, default
  - API
  - UI

- Expert can add multiple scores via the UI


src/components/layouts/header.tsx
```diff
@@ -4,13 +4,15 @@ import { useRouter } from "next/router";
 import { useSession } from "next-auth/react";
 import DocPopup from "@/src/components/layouts/doc-popup";
 import { type Status, StatusBadge } from "./status-badge";
+import { cn } from "@/src/utils/tailwind";
```
maxdeichmann, line 7: Are the changes in this fiel supposed to be here?

src/components/session/index.tsx
```diff
@@ -113,15 +113,15 @@ export const SessionPage: React.FC<{
                 {trace.timestamp.toLocaleString()}
               </div>
               <div className="mb-1 mt-2 text-xs text-gray-500">Scores</div>
-              <div className="flex flex-wrap content-start items-start gap-1">
+              <div className="group/scores flex flex-wrap content-start items-start gap-1">
                 <GroupedScoreBadges scores={trace.scores} />
+                <ExpertScoreButton
```
maxdeichmann, line 118: image
It would be great to not have the delete button, when there is only one selected score

src/components/session/index.tsx
```diff
@@ -113,15 +113,15 @@ export const SessionPage: React.FC<{
                 {trace.timestamp.toLocaleString()}
               </div>
               <div className="mb-1 mt-2 text-xs text-gray-500">Scores</div>
-              <div className="flex flex-wrap content-start items-start gap-1">
+              <div className="group/scores flex flex-wrap content-start items-start gap-1">
```
maxdeichmann, line 116: What is `group/scores` about?

src/components/table/use-cases/traces.tsx
```diff
@@ -370,11 +371,19 @@ export default function TracesTable({
       cell: ({ row }) => {
         const traceId = row.getValue("id");
         return traceId && typeof traceId === "string" ? (
-          <DeleteTrace
-            traceId={traceId}
-            isTableAction={true}
-            projectId={projectId}
-          />
+          <>
+            <ExpertScoreButton
```
maxdeichmann, line 375: Not sure about the UI of the two buttons here. I would at least add a gap between them.

maxdeichmann: Maybe removing the action column and only using the multi select would be an option.

src/components/ui/command.tsx
```diff
@@ -42,14 +42,14 @@ const CommandInput = React.forwardRef<
   React.ComponentPropsWithoutRef<typeof CommandPrimitive.Input>
 >(({ className, ...props }, ref) => (
   <div
-    className="flex items-center rounded-lg border px-3"
```
maxdeichmann, line 45: are these changes intended? I assume yes, but we should make sure to keep these as close to the original as possible to also make upgrades with shadcn.

src/features/expert-scoring/components/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+                  />
+                  <FormField
+                    control={form.control}
+                    name={`scores.${index}.comment`}
+                    render={({ field }) => (
+                      <FormItem className="col-span-2">
+                        <FormControl>
+                          <Textarea
+                            {...field}
+                            placeholder="Comment"
```
maxdeichmann, line 247: Do you think it would make sense to make this box larger? I think that users would add more than just a couple of words. Maybe we could shift the comment blow name + score and increase the height.

src/components/session/index.tsx
maxdeichmann: I just learned that the button will not only have an impact on the form, but will remove the score from the trace. This was not entirely clear to me when playing around initially.

src/server/api/routers/scores.ts
```diff
@@ -128,73 +150,57 @@ export const scoresRouter = createTRPCRouter({
...
 
-      return ctx.prisma.score.create({
-        data: {
-          trace: {
-            connect: {
-              id: trace.id,
-            },
+      // Delete all scores that are not in the input
+      // User removed them in the UI
+      await ctx.prisma.score.deleteMany({
```
maxdeichmann, line 155: I think this should happen in a transaction together with all the upserts below.

PR #687, opened 2023-12-13
fix: upgrade sentry

maxdeichmann:


marcklingen:
@maxdeichmann added @sentry/types to fix this

maxdeichmann:
> @maxdeichmann added @sentry/types to fix this

Yes, makes sense that this fixes the issue. I was rather wondering why we need to include this dependency in the first place. I would expect the /next dependency to add this automatically.

PR #707, opened 2023-12-17
fix: timezone in chart query

marcklingen:


marcklingen:
better solution: https://github.com/langfuse/langfuse/pull/708

PR #708, opened 2023-12-17
fix(db): use utc time in prisma datetime filter

flxwu:
See https://discord.com/channels/1111061815649124414/1186038330761158787


flxwu:
`NODE_ENV=development npm run test -- src/__tests__/query-builder.servertest.ts -t 'should  order by a column'`

Local test passing:
image

flxwu:
Time filter still works:

image

nilsreichardt:
Thanks <3

PR #750, opened 2023-12-23
feat(ui, api): Add Ordering functionality to traces

flxwu:
## What does this PR do?

* Add orderBy params to the DataTable component 
* Parse and process orderBy on the Traces.all API contract and corresponding DB query
* Add Test Setup Infrastructure for testing Next Router + tRPC contracts -> I implemented this to be easily extensible & re-usable for new tests
* Add Unit tests 

This feature should now be easily addable to the other tables as well.

_NOTE TO REVIEWER : This PR looks big, but actually most of it are unit tests + infrastructure for setting up the tests :)
NOTE TO REVIEWER 2: I split up all subparts of this MR into separate standalone commits, which might be easier to follow for review._

**Video**

https://github.com/langfuse/langfuse/assets/16862997/f47afb00-dd95-44c7-91c3-5734d0c78ace

## Type of change

- [x] Refactor (does not change functionality, e.g. code style improvements, linting)

## Mandatory Tasks


marcklingen:
Just had a quick read, looks great, will try to get this merged soon

marcklingen:
Thx @flxwu

nilsreichardt:
Wow, nice that looks cool!

Would be great to also have this option in the generations. My use-case is to identify expensive traces/generations. Having a sorting option (desc or asc) for the "Cost" column in generation could be a low-hanging feature to coming close to the goal. 

Also, sorting the "Usage" column could be helpful because I want to check if everything works as expected with long contexts (is the output as expected? How much context do I need? -> relevant when switching models with different context sizes, e.g. currently using GPT-3.5 with 16k context but thinking about switching to a model with 8k context -> is this causing problems because I have many requests with a long context?)

flxwu:
> Would be great to also have this option in the generations [...] Having a sorting option (desc or asc) for the "Cost" column in generation

That'd be a 30min thing to add-on after this PR! :)

src/features/orderBy/server/orderByToPrisma.ts
```diff
@@ -0,0 +1,30 @@
...
+  orderBy: OrderByState,
+  tableColumns: ColumnDefinition[],
+): Prisma.Sql {
+  if (!orderBy) {
+    return Prisma.sql([`ORDER BY t.timestamp DESC`]);
+  }
+  // Get column definition to map column to internal name, e.g. "t.id"
+  const col = tableColumns.find(
+    // TODO: Only use id instead of name.
+    // It's less error-prone & decouples data fetching from the human-readable UI labels
```
flxwu, line 21: We're right now using `.name` to match the table column, which seems very error-prone to me since that's the UI copy - API behaviour should never depend on the copy on the UI imo :)

I didn't want to overblow this PR by doing this migration here, but if you're happy to do it & don't see any risks with it, i can also do it here.

src/features/filters/server/filterToPrisma.ts
```diff
@@ -18,7 +18,11 @@ export function filterToPrismaSql(
 ): Prisma.Sql {
   const statements = filters.map((filter) => {
     // Get column definition to map column to internal name, e.g. "t.id"
-    const col = tableColumns.find((c) => c.name === filter.column);
+    const col = tableColumns.find(
+      (c) =>
+        // TODO: Only use id instead of name
```
flxwu, line 23: See my other PR comment in orderByToPrisma

maxdeichmann:
@flxwu thanks a lot for the contribution! This will be part of our next release :)

PR #801, opened 2024-01-03
feat: add tags for traces

### LINKED ISSUE #616: feat: tagging of traces
detailed spec will follow

RichardKruemmel:
## What does this PR do?

Fixes #616

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Mandatory Tasks

## Checklist

- My code follows the style guidelines of this project (`npm run prettier`)


src/server/api/services/EventProcessor.ts
```diff
@@ -325,6 +325,7 @@ export class TraceProcessor implements EventProcessor {
           : undefined,
         public: body.public ?? undefined,
         project: { connect: { id: apiScope.projectId } },
+        tags: body.tags ?? undefined,
```
maxdeichmann, line 328: Should we use undefined here? I would expect that we always have an array, empty or filled.

maxdeichmann: Does this work? According to the prisma schema, tags cannot be null.

prisma/seed.ts
```diff
@@ -136,6 +138,11 @@ async function main() {
           Math.floor(Math.random() ** 1.5 * 90 * 24 * 60 * 60 * 1000),
       );
 
+      const envTag = envTags[Math.floor(Math.random() * envTags.length)];
+      const colorTag = colorTags[Math.floor(Math.random() * colorTags.length)];
+
+      const tags = [envTag, colorTag].filter((tag) => tag !== "");
```
maxdeichmann, line 144: nit: i think it would be nicer to have null in the envTags or colorTags and then filter by them. This way we dont have the risk of having empty strings in the array, which we might not see.

src/components/table/use-cases/traces.tsx
```diff
@@ -356,6 +358,11 @@ export default function TracesTable({
       enableHiding: true,
       enableSorting: true,
     },
+    {
+      accessorKey: "tags",
+      header: "Tags",
```
maxdeichmann, line 363: can you add `id: "tags",` here?

src/features/filters/components/filter-builder.tsx
```diff
@@ -334,6 +336,18 @@ function FilterBuilderForm({
                       }
                       values={Array.isArray(filter.value) ? filter.value : []}
                     />
+                  ) : filter.type === "arrayOptions" ? (
```
maxdeichmann, line 339: this is the same as above, right? Could we just write `filter.type === "stringOptions" || filter.type === "arrayOptions"` above?

src/features/filters/components/filter-builder.tsx
```diff
@@ -97,12 +97,14 @@ export function FilterBuilder({
                         ? new Date(filter.value).toLocaleDateString()
                         : filter.type === "stringOptions"
                           ? filter.value.join(", ")
-                          : filter.type === "number" ||
-                              filter.type === "numberObject"
-                            ? filter.value
-                            : filter.type === "boolean"
-                              ? `${filter.value}`
-                              : `"${filter.value}"`}
+                          : filter.type === "arrayOptions"
```
maxdeichmann, line 100: same as below

src/features/filters/server/filterToPrisma.ts
```diff
@@ -30,13 +34,19 @@ export function filterToPrismaSql(
 
     const colPrisma = Prisma.raw(col.internal);
     const operatorPrisma =
-      filter.operator in operatorReplacements
+      filter.type === "arrayOptions"
         ? Prisma.raw(
-            operatorReplacements[
-              filter.operator as keyof typeof operatorReplacements
+            arrayOperatorReplacements[
```
maxdeichmann, line 39: operatorPrisma wont be used if all of or none of.
- For all of: cant we use @> as operator?
- For none of: cant we use && as operator and then add another variable to the final return string like "func"? If func is not empty, we wrap the final prisma statement in `func(...)`

src/server/api/services/EventProcessor.ts
RichardKruemmel: From my understanding: If there are no tags present in body object the function should not upsert any tags at all. If there are no tags present in the body object it triggers the default of the prisma schema, adding an empty array to the trace which is the behaviour we want.
It gives the upsert and create function more flexibility for the case that that the body does not contain tags.

Can assume that every trace body when upserted or created will always have tags in it?

prisma/seed.ts
RichardKruemmel: Not sure if I misunderstood you but just to clarify in L144 we are filtering out any empty string. When envTag and colorTag are both [""], after filtering both the const tags will be [], which is the default of our prisma schema.

src/components/table/use-cases/traces.tsx
RichardKruemmel: Done 🙏🏻
Should we add id's to the other columns as well? Maybe we can have a quick follow up why we need or not need an id for certain columns. 😊

RichardKruemmel:
Screenshot 2024-01-04 at 15 20 28

Commit [abc2abe](https://github.com/langfuse/langfuse/pull/801/commits/abc2abee8efab49f7ef12e4ce7fb1e423bf0bebc) adds minimalistic UI implementation in trace detail

maxdeichmann (approved):

PR #834, opened 2024-01-07
feat(ui): improved/nested main navigation

marcklingen:


marcklingen:
> I am not sure i like this change. Alternatively, we could:
> 
> * have tabs on the Tracing view on top to select traces / spans / scores
> * I think that scores should be a main navigation point as we will add features here in the next weeks.
> * We could think about moving Feedback and Support further down towards the user email adress.
> * We could also move the feedback and support buttons to the very top right.

thx, good idea to improve main nav in future as complexity increases. as discussed, merging this for now as we need a secondary navigation for sure.

PR #879, opened 2024-01-15
feat(ui): add ui for tags on traces

RichardKruemmel:
- Add option to get back to default order by clicking table header
- Add better UI for tags in Trace Table and Trace Details
- Add Popover by clicking on tags
- Implement add/delete logic for tag popover
- Implement mutations using trpc's setData method


prisma/migrations/20240102010215_add_tags_in_traces/migration.sql
maxdeichmann, line 1: are we sure we want to rename this?

prisma/migrations/20240103170551_index_tags_in_traces/migration.sql
maxdeichmann, line 1: same as above?

src/features/tag/hooks/useTagManager.tsx
```diff
@@ -0,0 +1,32 @@
+import { useState, useMemo } from "react";
+
+type UseTagManagerProps = {
+  initialTags: string[];
+  availableTags: string[];
+};
+
+function useTagManager({ initialTags, availableTags }: UseTagManagerProps) {
```
maxdeichmann, line 8: The naming here is a bit confiuing. From reding the function name, i am not sure what is going on below. Do we wneed to extract this in the first place?

src/features/tag/components/TagCreateItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { cn } from "@/src/utils/tailwind";
+import { CommandItem } from "cmdk";
+
+type TagItemCreateProps = {
+  inputValue: string;
+  options: string[];
+  onSelect: () => void;
+};
+
+const TagItemCreate = ({
```
maxdeichmann, line 11: can we use `CommandItemCreate` here?

prisma/migrations/20240102010215_add_tags_in_traces/migration.sql
RichardKruemmel: No, we shouldn't do this. I corrected it.

src/features/tag/components/TagCreateItem.tsx
RichardKruemmel: CommandItemCreate has currently prompt name hard coded and uses different css.
We could look into generalising the CommandItemCreate component and then reuse.

src/features/tag/hooks/useTagManager.tsx
RichardKruemmel: Agreed. Naming is not ideal. In essence this customhook manages, which tags we show as selectable in the popover, it handles the state of which tags the user has selected and the creation of new tags.

Should we reuse parts of the logic of the tags for other components having UI logic in TagManager and business logic is useTagManager separated might be beneficial. Also should we decide to implement testing for this feature, separating parts of the logic in a custom hook will make testing easier.
Currently as we are not reusing this I agree it adds only overhead and makes it harder to understand what actually happens in the component.
What do you think?

src/features/tag/components/TagPopOver.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from "react";
+import { api } from "@/src/utils/api";
+import { useHasAccess } from "@/src/features/rbac/utils/checkAccess";
+import { type RouterOutput, type RouterInput } from "@/src/utils/types";
+import TagManager from "@/src/features/tag/components/TagMananger";
+
+type TagPopOverProps = {
```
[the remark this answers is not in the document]
RichardKruemmel: Agreed.
As point of reference we have the star toggle feature. It uses StarTraceDetailsToggle and StarTraceToggle. To use the same syntax we could rename it to TagTracePopver and TagTraceDetailsPopover. What do you think?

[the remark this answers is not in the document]
maxdeichmann: yes, sounds good!

maxdeichmann (approved):

Issue #888, opened 2024-01-16
bug: 1mb body size limit in uploading prevents large LLM context use to be saved/tracked


cw5d:
### Describe the bug

With gpt-4-1106-preview and Claude-2/2.1 the context length has been extended greatly. A simple math would work like this,

with 1 token = 4 characters, 200k tokens would come to 800k without taking into account other metadata and tags being sent to Langfuse although. We've seen in some rare occasions where the 1mb body size limit prevents data from being accepted/saved. 

Considering that it's easy to increase the limit, such as https://nextjs.org/docs/pages/building-your-application/routing/api-routes#custom-config I wonder if it's a quick fix that addresses some corner cases as people increasingly take use of large context window from certain models?

### To reproduce

Max out your context window when testing for real with a LLM such as Claude2, or just upload some data >1mb.

### Additional information

If you could put a more sensible ceiling such as 2mb, it would go a long way until the next LLM with extremely long context window comes along :)

maxdeichmann:
@cw5d i just merged the change. It will be live in a couple of minutes.

PR #898, opened 2024-01-17
fix: increase ingestion API body size

maxdeichmann:


maxdeichmann:
closes https://github.com/langfuse/langfuse/issues/888

PR #920, opened 2024-01-22
perf: native db upserts for TraceProcessor and ObservationProcessor

marcklingen:


src/server/api/services/EventProcessor.ts
```diff
@@ -124,19 +121,23 @@ export class ObservationProcessor implements EventProcessor {
...
-              version: this.event.body.promptVersion,
+              projectId_name_version: {
+                projectId: apiScope.projectId,
+                name: this.event.body.promptName,
+                version: this.event.body.promptVersion,
+              },
             },
           })
         : undefined;
+    if (prompt === null)
```
maxdeichmann, line 139: I woudl do a falsy check here to also catch undefined.

src/server/api/services/EventProcessor.ts
```diff
@@ -124,19 +121,23 @@ export class ObservationProcessor implements EventProcessor {
...
+      console.warn("Prompt not found for observation", this.event.body);
```
maxdeichmann, line 140: wont this be quite noisy?

marcklingen: added comment to clarify that this is only logged in error cases

PR #927, opened 2024-01-23
fix(security): add validation for html in inputs

RichardKruemmel:
- Validate Name in signup and projectName for html tags


src/features/projects/server/projectsRouter.ts
```diff
@@ -77,7 +78,7 @@ export const projectsRouter = createTRPCRouter({
     .input(
       z.object({
         projectId: z.string(),
-        newName: z.string(),
```
maxdeichmann, line 80: why are you making this change? Cant we apply the refine logic to the newName?

RichardKruemmel: The idea behind it is that the refine logic already exists in the projectNameSchema and reusing this felt more natural to me as we have less duplicate logic. The clear downside is that we now have to nest inside the router (newName.name etc). Do you prefer to have the refine logic in L80 instead?

maxdeichmann: Yes, i think the nesting is a bit confusing. We could do something like:

```
 z.object({
        projectId: z.string(),
        newName: .refine((value) => noHtmlCheck(value), {
          message: "Input should not contain HTML",
        }),
})
```

WDYT?

RichardKruemmel: That is fine with me. It reduces the readibility in this part of the code but at least we don't need to nest.

But I quickly looked if zod supports something like projectNameSchema.name to mitigate the nesting and found this option:
```
    .input(
      z.object({
        projectId: z.string(),
        newName: projectNameSchema.shape.name,
      }),
```

With this we can delete the nesting and reuse the logic that already exists in the projectNameSchema.

maxdeichmann: Ah i was not aware about this. Fine for me.

maxdeichmann: Thanks for the research!

RichardKruemmel: Great 😄 Will quickly implement it. And push it.

PR #959, opened 2024-01-25
feat(ui): improve prompt ui

RichardKruemmel:
- Improved prompt UI
- Refactor prompt routes
- Fix trash Icon UI

- [ ] Add improved seeding script for prompts -> moved to separate pr


src/features/rbac/components/ProjectMembersTable.tsx
```diff
@@ -119,8 +119,8 @@ export function ProjectMembersTable({ projectId }: { projectId: string }) {
                     {hasDeleteAccess ? (
                       <TableCell>
                         <Button
-                          variant="ghost"
-                          size="xs"
+                          variant="destructive"
```
maxdeichmann, line 122: I just checked that in the UI and think it is very dominant in the UI right now. What do you think? I think it might make sense to revert this again.

src/features/prompts/server/prompt-router.ts
```diff
@@ -27,18 +27,44 @@ export const promptRouter = createTRPCRouter({
         projectId: input.projectId,
         scope: "prompts:read",
       });
-      return ctx.prisma.prompt.findMany({
+      const prompts = await ctx.prisma.$queryRaw<Array<Prompt>>`
```
maxdeichmann, line 30: The typing here is dangerous. Typescript types this to Prompt but we did not select all the db columns (e.g. project_id, updated_at)

src/features/prompts/server/prompt-router.ts
```diff
@@ -27,18 +27,44 @@ export const promptRouter = createTRPCRouter({
...
+        SELECT id, name, version, created_at AS "createdAt", is_active AS "isActive"
+        FROM prompts
+        WHERE (name, version) IN (
+          SELECT name, MAX(version)
+          FROM prompts
+          WHERE "project_id" = ${input.projectId}
+          GROUP BY name
+        )
+        AND "project_id" = ${input.projectId}`;
```
maxdeichmann, line 39: WDYT about ordering this by name? This way the ordering of the prompts table remains the same even if we create new prompts on any of these. Currenlty, the prompt with the latest addition is added on top.

src/features/prompts/components/prompts-table.tsx
```diff
@@ -106,12 +76,12 @@ export function PromptTable(props: { projectId: string }) {
       accessorKey: "action",
```
maxdeichmann, line 76: We discussed this differently, but i think it does not make sense to have this button here anymore.

src/features/prompts/components/prompts-table.tsx
```diff
@@ -127,15 +97,14 @@ export function PromptTable(props: { projectId: string }) {
   ];
 
   const convertToTableRow = (
-    item: RouterOutput["prompts"]["all"][number],
+    item: RouterOutput["prompts"]["all"][0],
```
maxdeichmann, line 100: Why is this change necessary?

src/features/prompts/components/prompt-detail.tsx
```diff
@@ -9,77 +9,114 @@ import { api } from "@/src/utils/api";
...
+import { PromotePrompt } from "@/src/features/prompts/components/promote-prompt";
+import { ScrollArea } from "@radix-ui/react-scroll-area";
+import { useQueryParam, NumberParam } from "use-query-params";
 import router from "next/router";
 
 export type PromptDetailProps = {
   projectId: string;
-  promptId: string;
+  promptName: string;
+  //promptVersion: number;
```
maxdeichmann, line 21: can you remove this?

src/features/prompts/server/prompt-router.ts
```diff
@@ -136,6 +167,47 @@ export const promptRouter = createTRPCRouter({
         throw e;
       }
     }),
+  history: protectedProjectProcedure
```
maxdeichmann, line 170: i would call this .allVersions

src/features/prompts/server/prompt-router.ts
```diff
@@ -136,6 +167,47 @@ export const promptRouter = createTRPCRouter({
...
+            some: {
+              projectId: input.projectId,
+            },
+          },
+        },
+      });
+
+      const joinedPromptAndUsers = prompts.map((p) => {
+        const user = users.find((u) => u.id === p.createdBy);
+        if (!user) return { ...p, creator: "Unknown" };
```
maxdeichmann, line 203: We have the "API" case as well, right? In this case I would also return API. If it is unknown, i would also log this to the console, because this should never happen.

src/features/prompts/components/prompt-detail.tsx
```diff
@@ -9,77 +9,114 @@ import { api } from "@/src/utils/api";
...
 };
 
 export const PromptDetail = (props: PromptDetailProps) => {
-  const prompt = api.prompts.byId.useQuery({
-    id: props.promptId,
+  const [currentPromptVersion, setCurrentPromptVersion] = useQueryParam(
+    "version",
+    NumberParam,
+  );
+  const prompt = api.prompts.byName.useQuery({
```
maxdeichmann, line 29: Why do we need this API call? We could just take the currentPromptVersion and filter the promptHistory, correct? Then we could also delete the backend API.

src/features/prompts/components/prompt-detail.tsx
```diff
@@ -9,77 +9,114 @@ import { api } from "@/src/utils/api";
...
-          {extractedVariables.length > 0 ? (
-            extractedVariables.map((variable) => (
-              <Badge key={variable} variant="outline">
-                {variable}
-              </Badge>
-            ))
-          ) : (
-            <span className="text-xs">No variables</span>
-          )}
+    <div className="flex flex-col md:h-[calc(100vh-2rem)]">
```
maxdeichmann, line 48: Maybe we can put this view into a container similar to the dashboard view?

src/features/prompts/components/prompt-detail.tsx
```diff
@@ -9,77 +9,114 @@ import { api } from "@/src/utils/api";
...
+                  key="nav"
+                  currentId={prompt.data.name}
+                  path={(name) => `/project/${props.projectId}/prompts/${name}`}
+                  listKey="prompts"
+                />
+              </>
+            }
+          />
+        </div>
+        <div className="col-span-2 md:h-full">
```
maxdeichmann, line 91: Maybe we can put this view into a container similar to the dashboard view?

src/features/prompts/components/prompt-history.tsx
```diff
@@ -0,0 +1,75 @@
+import { StatusBadge } from "@/src/components/layouts/status-badge";
+import { type NextRouter, useRouter } from "next/router";
+
+type Prompt = {
```
maxdeichmann, line 4: Why do we need a new type here? We already have a prompt type or not?

src/features/prompts/components/prompt-history.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onClick={() => {
+      props.index === 0
+        ? props.setCurrentPromptVersion(undefined)
+        : props.setCurrentPromptVersion(props.prompt.version);
+    }}
+  >
+    <div // center all the content in the div below horizontally
+      className="flex items-center gap-2"
+    >
+      <span className="rounded-sm bg-gray-200 p-1 text-xs">PROMPT</span>
```
maxdeichmann, line 35: The Prompt tag is maybe a bit repeptitive as i see it again. What about putting the Version-X there?

src/features/prompts/components/promote-prompt.tsx
```diff
@@ -15,21 +15,21 @@ export function PromotePrompt({
...
   projectId: string;
   promptName: string;
   disabled: boolean;
+  variant?: "ghost" | "outline";
 }) {
-  const [isPromoted, setIsPromoted] = useState(false);
   const utils = api.useUtils();
   const posthog = usePostHog();
-
+  const [isOpen, setIsOpen] = useState(false);
```
maxdeichmann, line 28: Did the popover not close automatically?

RichardKruemmel: No, it didn't. In the Prompts table it worked as we update the underlying data in the table creating a refresh - reminds me of the tag popover 😄 
In the details view it is part of the toolbox so we needed an logic for it to close. See screencast in Linear for the behavior without the logic.

src/features/prompts/components/prompt-history.tsx
RichardKruemmel: Sure, I wasn't in love with it either. My thoughts were to keep it similar to the Trace detail observation tree.

src/features/prompts/components/prompt-history.tsx
RichardKruemmel: Because of the join with the users table.

the .allVersions route returns the Prompt Type, we already have, plus creator: string | null according to TS.
I am curios if there is a more native t3 way to handle this.

RichardKruemmel:
see #961 for final pr

RichardKruemmel:
All changes resolved in #961

PR #994, opened 2024-01-30
feat: tokenize in migration script

maxdeichmann:


prisma/migrations/20240130160110_claude_models/migration.sql
```diff
@@ -0,0 +1,40 @@
...
+  start_date,
+  input_price,
+  output_price,
+  total_price,
+  unit,
+  tokenizer_id,
+  tokenizer_config
+)
+VALUES
+  -- nothing earlier required for ada
```
marcklingen, line 30: ada?

prisma/migrations/20240130160110_claude_models/migration.sql
```diff
@@ -0,0 +1,40 @@
+-- This is an empty migration.
```
marcklingen, line 1: add comment

scripts/model-match.ts
```diff
@@ -44,6 +51,11 @@ export async function modelMatch() {
         model: true,
         unit: true,
         projectId: true,
+        promptTokens: true,
+        completionTokens: true,
+        totalTokens: true,
+        input: true,
```
marcklingen, line 57: need to reduce batch size if input/output is included to reduce memory usage?

PR #1025, opened 2024-02-01
perf: prevent ReDos for user-defined regex

marcklingen:


marcklingen:
close for now, revisit later

PR #1047, opened 2024-02-02
fix: remove `Expires` from S3 PutObject

### LINKED ISSUE #1046: bug: `Expires` doesn't actually delete files off S3
### Describe the bug

The `Expires` header in S3's PutObject is meant to indicate to clients (e.g. Chrome) when the file should no longer be cached.

https://github.com/langfuse/langfuse/blob/b86eab9e04e9039160ac0fe34d7a712060fdcd38/src/server/api/routers/generations.ts#L479

https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax

If you want to auto-delete the files, setting up a lifecycle rule on the bucket is the correct solution. Personally, I do _not_ want langfuse to automatically attempt to modify lifecycle rules; if the bucket gets too big, I'll enable that myself. =) 

### To reproduce

N/A

### Additional information

_No response_

marcklingen:
Fixes #1046

### DIFF
--- a/src/server/api/routers/generations.ts
+++ b/src/server/api/routers/generations.ts
@@ -476,7 +476,6 @@ export const generationsRouter = createTRPCRouter({
             Key: fileName,
             Body: output,
             ContentType: exportOptions[input.fileFormat].fileType,
-            Expires: new Date(Date.now() + 60 * 60 * 1000), // in 1 hour, file will be deleted
           }),
         );
         const signedUrl = await getSignedUrl(


Manouchehri:
I think adding a long cache might be a good idea to save bandwidth? =) 

https://github.com/langfuse/langfuse/issues/1046#issuecomment-1922680072

If privacy is a concern, `CacheControl: "private, immutable, max-age=31536000, s-maxage=0"` and `Expires: new Date(0)` is the way to go.

marcklingen:
Adding caching headers makes sense down the road when this feature is expanded and used to maybe unload data from Langfuse to S3 (which is then consumed more than once). Right now these files are created interactively in the UI and usually downloaded exactly once -> I think this does not make a difference here.

Manouchehri:
Fair, good point. =)

marcklingen:
Thanks again for raising this issue

PR #1088, opened 2024-02-06
feat(api): add notice regarding hostname to default 401

marcklingen:


src/features/public-api/server/apiAuth.ts
```diff
@@ -91,7 +91,9 @@ export async function verifyAuthHeaderAndReturnScope(
     console.error("Error verifying auth header: ", error);
     return {
       validKey: false,
-      error: error instanceof Error ? error.message : "Authorization error",
+      error:
+        (error instanceof Error ? error.message : "Authorization error") +
+        ". Confirm that you've configured the correct host.",
```
maxdeichmann, line 96: Should we add a link to our docs here?

marcklingen: we could, currently there is not the one page to refer to, will change with the docs relaunch, let's keep this in mind and add it then

PR #1095, opened 2024-02-06
Expand generation model units

hassiebp:
## What does this PR do?

Adds support for additional generative model usage units. Currently only `CHARACTERS` and `TOKENS` are supported. To add support for voice based models such as OpenAI's Whisper, we are adding `MILLISECONDS` and `SECONDS` as additional units to support cost calculation and analysis within Langfuse. For image-based models, we're adding the unit type `IMAGES` to account for e.g. OpenAI's DALL E per image pricing.

## Type of change

- [X] Chore (refactoring code, technical debt, workflow improvements)
- [X] New feature (non-breaking change which adds functionality)

## Mandatory Tasks


maxdeichmann (approved):

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "prebuild": "cp generated/openapi-client/openapi.yml public/openapi-client.yml && cp generated/openapi-server/openapi.yml public/openapi-server.yml",
     "build": "next build",
-    "dev": "next dev",
+    "dev": "npm run infra:dev:up && next dev && npm run infra:dev:down",
```
marcklingen, line 8: I like the current behavior as I mostly have the infra docker compose running in the background forever. This also breaks `npm run dx` which assumes that the db is already running.

marcklingen (changes requested):
Change looks great, unsure about changing the `dev` command though, see comment

PR #1099, opened 2024-02-06
chore: update start command to use node instead of next

marcklingen:


marcklingen:
breaks ci test, not important, closing for now

PR #1147, opened 2024-02-13
feat: add prompt config

RichardKruemmel:


src/__tests__/prompts.servertest.ts
```diff
@@ -173,14 +193,20 @@ describe("/api/public/prompts API Endpoint", () => {
     expect(fetchedObservations.body.version).toBe(2);
     expect(fetchedObservations.body.isActive).toBe(true);
     expect(fetchedObservations.body.createdBy).toBe("user-1");
+    expect(fetchedObservations.body.config).toEqual({ temperature: 0.2 });
   });
 
   it("should create and fetch a prompt", async () => {
+    const promptId = uuidv4();
     await makeAPICall("POST", "/api/public/prompts", {
+      id: promptId,
```
maxdeichmann, line 202: I think this here can be removed as the API does not take ids.

src/__tests__/prompts.servertest.ts
```diff
@@ -327,6 +354,36 @@ describe("/api/public/prompts API Endpoint", () => {
 
     expect(dbGeneration).toBeNull();
   });
+
+  it("should create empty object if no config is provided", async () => {
+    const promptId = uuidv4();
+    await makeAPICall("POST", "/api/public/prompts", {
+      id: promptId,
```
maxdeichmann, line 361: same as above

src/features/prompts/server/prompt-router.ts
```diff
@@ -315,13 +319,15 @@ export const createPrompt = async ({
...
   createdBy,
+  config,
   prisma,
 }: {
   projectId: string;
   name: string;
   prompt: string;
   isActive?: boolean;
   createdBy: string;
+  config: JsonValue;
```
maxdeichmann, line 330: i would define it as: config: z.infer<typeof jsonSchema>

src/pages/api/public/prompts.ts
```diff
@@ -14,6 +15,7 @@ const PromptCreateSchema = z.object({
   name: z.string(),
   prompt: z.string(),
   isActive: z.boolean(),
+  config: jsonSchema.optional().nullable().default({}),
```
maxdeichmann, line 18: i think nullable is enough here

src/features/prompts/server/prompt-router.ts
RichardKruemmel:
```suggestion
    config: z.infer<typeof jsonSchema>;
```

src/pages/api/public/prompts.ts
RichardKruemmel:
```suggestion
  config: jsonSchema.optional().nullable(),
```

src/features/prompts/components/new-prompt-button.tsx
```diff
@@ -229,6 +250,28 @@ export const NewPromptForm = (props: {
             </>
           )}
         />
+        <FormField
+          control={form.control}
+          name="config"
+          render={({ field }) => (
+            <FormItem>
+              <FormLabel>Config</FormLabel>
+              <JsonView
```
maxdeichmann, line 259: I just checked this and I was not able to create a nested json. Could you look into this?

prisma/migrations/20240209162415_add_prompt_config/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "prompts" ADD COLUMN     "config" JSONB NOT NULL DEFAULT '{}';
```
maxdeichmann, line 2: please update the timestamp of the migration folder so that this is the latest one

maxdeichmann (approved):

src/features/prompts/components/prompt-detail.tsx
```diff
@@ -112,6 +115,10 @@ export const PromptDetail = (props: PromptDetailProps) => {
               )}
             </div>
           </div>
+
+          {prompt.config && Object.keys(prompt.config).length > 0 && (
```
marcklingen, line 119: Object.keys does not work if JSON is not of type Object. JSON could also be an array or a scalar (number/string/…). I’d suggest to change to prompt.config !== {}

RichardKruemmel: Isn't the problem with `prompt.config !== {}` that in TS we don't compare the properties but the reference in memory and therefore the evaluation would always be true?

To better understand the background of the concern: The case that the config is array or scalar would be only possible for API ingested prompts right? Prompts edited via UI are always an object.

My first thought to solve this would actually be to write a custom function like:
`function isJsonEmpty(json: jsonSchema): boolean {
  if (json === null || json === undefined) {
    return true;
  }
  if (Array.isArray(json)) {
    return json.length === 0;
  }
  if (typeof json === 'object') {
    return Object.keys(json).length === 0;
  }
  // For scalar values
  return false;
}`

@marcklingen wdyt?

marcklingen: For simplicity we can also always show the config and remove this check.

RichardKruemmel: After some further research I realized that:
Object.keys().length also works for arrays and strings.
Therefore the only cases currently not covered would be int, floats, booleans as null and undefined are prohibited by the schema. Should we support those edge cases?

PR #1165, opened 2024-02-15
feat(ui): show full id in session table

RichardKruemmel:


src/components/table/table-link.tsx
```diff
@@ -17,7 +17,7 @@ export default function TableLink({
       href={path}
       title={value}
     >
-      {value.length > truncateAt
+      {value.length - truncateAt > 3
```
maxdeichmann, line 20: why do we need this change?

RichardKruemmel: Reasoning was that we had the following case: "session-1" got truncated to "...sion-1". Replacing "ses" with "..." seemed redundant to me.

RichardKruemmel: `{value.length > truncateAt
        ? '...${value.substring(value.length - truncateAt)}'`
When _value.length > truncateAt_ we slice the value to three dots + substring(value.length - truncateAt).
Therefore we should account for the three dots in the condition or not?

maxdeichmann (approved):

PR #1166, opened 2024-02-15
Api to delete a score issue No. 1148

Dharshan-K:
Added the Delete API to delete a Score at ```/api/public/score/[scoreId]```. Test is pending because I have problem with authentication headers.


marcklingen:
Added in #1167 which was the more complete implementation: used the /scores/[scoreId] route and included tests

PR #1199, opened 2024-02-19
feat: add support for AUTH_GOOGLE_ALLOWED_DOMAINS environment variabl…

aablsk:
test cloud build

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,29 @@
+steps:
+  - name: "gcr.io/cloud-builders/docker"
+    args:
+      [
+        "build",
+        "-t",
+        "europe-west1-docker.pkg.dev/langfuse-test/langfuse/langfuse:latest",
+        ".",
+      ]
+  - name: "gcr.io/cloud-builders/docker"
+    args:
+      [
+        "push",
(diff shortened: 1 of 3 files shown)


marcklingen:
Thanks for the contribution! I'd suggest to generalize this to all authentication methods, e.g. `ALLOWLISTED_DOMAINS`. What do you think? Is there anything specific to google auth here?

Can you remove the cloudbuild file?

aablsk:
Dear @marcklingen, apologies for this **unintended PR**! 🙇 
I'm testing new git tooling which seems to have opened this PR against the incorrect repository. I'm closing this PR.

I will look into adding a general allowlisting parameter for all providers, but I haven't evaluated what the secure way is for domain verification for other providers, yet, (e.g. it seems that using `email` to verify a domain for Google provider is not recommended) and create a new PR once I've had the time and I'm feeling confident about the solution. 

Unfortunately, I'm currently lacking a way to verify Azure AD as I do not have access to an Azure AD instance with a verified domain.

marcklingen:
No worries and thanks for your elaborate answer on this!

I think generalized allow-listing of domains would already be an improvement / useful feature and verification of emails on Google and Azure could be done in separate PRs.

PR #1215, opened 2024-02-20
feat(API): catch prisma errors individually for public apis

maxdeichmann:
- adds exception handling for prisma exceptions on public apis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -112,6 +112,7 @@
         "eslint-config-next": "^14.1.0",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
+        "node-mocks-http": "^1.14.1",
         "postcss": "^8.4.35",
         "prettier": "^3.2.5",
         "prettier-plugin-tailwindcss": "^0.5.11",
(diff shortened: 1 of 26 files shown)


package-lock.json
```diff
@@ -112,6 +112,7 @@
         "eslint-config-next": "^14.1.0",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
+        "node-mocks-http": "^1.14.1",
```
marcklingen, line 115: dev-dep

src/pages/api/public/metrics/daily.ts
```diff
@@ -149,6 +150,12 @@ export default async function handler(
     }
   } catch (error: unknown) {
     console.error(error);
+    if (isPrismaException(error)) {
+      return res.status(500).json({
+        message: "Error processing events",
```
marcklingen, line 155: the message does not make sense in this case. I'd just suggest to remove it alltogether in all routes as it does not add much detail/context

marcklingen (commented):
looks great, nice test as well, added 2 comments

PR #1216, opened 2024-02-20
feat(generation): extend generations table

RichardKruemmel:


src/components/table/use-cases/generations.tsx
```diff
@@ -339,6 +409,25 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
...
+      enableHiding: true,
+      enableSorting: true,
+      cell: ({ row }) => {
+        const value = row.getValue("promptId");
+        const promptName = row.original.promptName;
+        const promptVersion = row.original.promptVersion;
+
+        return typeof value === "string" ? (
+          <TableLink
+            path={`/project/${projectId}/prompts/${promptName}?version=${promptVersion}`}
```
maxdeichmann, line 425: Something is off here. I just got directed to this link when clicking this: `http://localhost:3000/project/7a88fb47-b4e2-43b8-a06c-a5ce950dc53a/prompts/Prompt%20with%20many%20versions?version=8`

src/components/table/use-cases/generations.tsx
```diff
@@ -354,8 +443,12 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
           traceName: generation.traceName,
           startTime: generation.startTime.toLocaleString(),
           endTime: generation.endTime?.toLocaleString() ?? undefined,
+          timeToFirstToken:
+            generation.completionStartTime?.toLocaleString() ?? undefined,
           latency: generation.latency === null ? undefined : generation.latency,
-          cost: generation.calculatedTotalCost,
+          totalCost: generation.calculatedTotalCost,
+          inputCost: generation.calculatedInputCost,
+          outputCost: generation.calculatedOutputCost,
```
maxdeichmann, line 451: Am i missing something or are these missing on `GenerationsTableRow`?

src/components/table/use-cases/generations.tsx
```diff
@@ -237,10 +263,54 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
       enableSorting: true,
     },
     {
-      accessorKey: "cost",
-      header: "Cost",
+      accessorKey: "latencyPerToken",
+      id: "latencyPerToken",
+      header: "Latency per Token",
+      cell: ({ row }) => {
+        const latency: number | undefined = row.getValue("latency");
```
maxdeichmann, line 270: I think this was a bit broken already: we cast the Decimals to numbers here and below. I think we should handle these gracefully.

src/server/api/routers/generations/getAllQuery.ts
```diff
@@ -23,6 +23,8 @@ export const getAllQuery = protectedProjectProcedure
         traceId: string;
         traceName: string;
         latency: number | null;
+        promptName: string | null;
+        promptVersion: string | null;
```
maxdeichmann, line 27: This should be always null, as we only have the `promptId` on the observations_view, right?

src/server/api/routers/generations/db/getAllGenerationsSqlQuery.ts
```diff
@@ -114,10 +114,14 @@ export function getAllGenerationsSqlQuery({
...
         o.calculated_total_cost as "calculatedTotalCost",
-        o."latency"
+        o."latency",
+        o.prompt_id as "promptId",
+        p.name as "promptName",
+        p.version as "promptVersion"
       FROM observations_view o
       JOIN traces t ON t.id = o.trace_id AND t.project_id = o.project_id
       LEFT JOIN scores_avg AS s_avg ON s_avg.trace_id = t.id and s_avg.observation_id = o.id
+      LEFT JOIN prompts p ON p.id = o.prompt_id
```
maxdeichmann, line 124: can you confirm that we have indices on these?

src/components/table/use-cases/generations.tsx
maxdeichmann: Also, i see where this comes from. However, i think it would be great to be able see the name + version in the UI and allow to browse these easilly as a user while keeping them as a link. Do you want to think about a nice illustration here?

src/components/table/use-cases/generations.tsx
```diff
@@ -213,6 +218,27 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
       enableHiding: true,
       enableSorting: true,
     },
+    {
+      accessorKey: "endTime",
+      id: "endTime",
+      header: "End Time",
+      enableHiding: true,
+      enableSorting: true,
```
maxdeichmann, line 226: is sorting already supported here?

src/components/table/use-cases/generations.tsx
marcklingen: Maybe relevant here as well: https://github.com/langfuse/langfuse/pull/1209/files

src/server/api/routers/generations/getAllQuery.ts
RichardKruemmel: Due to the left join in the rawSqlQuery coming from getAllGenerationsSqlQuery we have both properties available. Therefore we need to extend the type. The promptId is available via the ObservationView Type.

src/components/table/use-cases/generations.tsx
RichardKruemmel: You are right. I added them. 🙂

src/components/table/use-cases/generations.tsx
RichardKruemmel: Yes, we are already sorting the startTime. I assumed we only need to enableSorting. It also works locally for me. Is there some magic in the background happing or do I need to do more than just setting the flag here?

src/server/api/routers/generations/db/getAllGenerationsSqlQuery.ts
RichardKruemmel: I added a migration that introduces indexing for the promptId inside the Observation Table.

web/prisma/schema.prisma
```diff
@@ -268,6 +268,7 @@ model Observation {
     @@index(projectId)
     @@index(parentObservationId)
     @@index(model)
+    @@index(promptId)
```
maxdeichmann, line 271: Very good!

PR #1245, opened 2024-02-23
feat(ui): allow filtering for null trace names

### LINKED ISSUE #1198: bug: "Unknown" Trace name in Dashboard but not present as trace name filter
### Describe the bug

In my trace dashboard, I am seeing a trace name show up as "Unknown", however, if I try to filter by "Unknown" on the individual trace page "Unknown" is not an option. Wondering why this is. Also, would be nice to be able to click on the number in the dashboard and it automatically apply the "name" filter so that I can dig in quicker and easily understand how that number came to be.

Screenshot 2024-02-18 at 7 28 12 AM

Screenshot 2024-02-18 at 7 28 46 AM

### To reproduce

Not sure how to reproduce as I am unaware of why or how this is happening.

### Additional information

_No response_

khareyash05:
## What does this PR do?

Allow filtering for null trace names in traces sections

Since naming null traces to `null` could cause conflicts while searching for traces. Therefore IMHO we should replace null traces with `[no-trace-name]`

Fixes #1198 

## Type of change

## Mandatory Tasks


marcklingen:
The filter does not work for me, can you confirm that it works for you? I cannot filter for name=null in trace-table

PR #1263, opened 2024-02-27
refactor: remove separate identify() to reduce number of posthog events

marcklingen:


marcklingen (commented):
Need to locally whether this still works as expected

PR #1265, opened 2024-02-27
feat(generations): extend CSV and JSON export

RichardKruemmel:


web/src/components/table/use-cases/generations.tsx
```diff
@@ -238,18 +246,15 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
...
 
         if (!timeToFirstToken) {
           return undefined;
         }
 
-        const latencyInMs =
-          new Date(timeToFirstToken).getTime() - new Date(startTime).getTime();
-        const latencyInSeconds = latencyInMs / 1000;
-
+        const latencyInSeconds = intervalInSeconds(startTime, timeToFirstToken);
```
maxdeichmann, line 257: This returns 0 in case of

web/src/components/table/use-cases/generations.tsx
```diff
@@ -56,7 +60,7 @@ export type GenerationsTableRow = {
   totalCost?: Decimal;
   traceName?: string;
   metadata?: string;
-  scores: Score[];
+  scores?: Record<string, number>;
```
maxdeichmann, line 63: Why is this different now?

web/src/server/api/routers/generations/getAllQuery.ts
```diff
@@ -62,26 +64,13 @@ export const getAllQuery = protectedProjectProcedure
     `,
     );
 
-    const scores = await ctx.prisma.score.findMany({
```
maxdeichmann, line 65: I once pulled this out for performance reasons. Lets check toghether whether this works.

web/src/components/table/use-cases/generations.tsx
RichardKruemmel: Good idea!

web/src/components/table/use-cases/generations.tsx
RichardKruemmel: The changes to the scores are required due to the changes in the getAllQuery.ts
traceId, startTime and completionStartTime (timeToFirstToken) did not have the right type definitions according the prisma schema.

web/src/server/api/routers/generations/db/getAllGenerationsSqlQuery.ts
```diff
@@ -70,19 +70,21 @@ export function getAllGenerationsSqlQuery({
         SELECT
           trace_id,
           observation_id,
-          jsonb_object_agg(name::text, avg_value::double precision) AS scores_avg
+          jsonb_agg(jsonb_build_object('name', name, 'value', avg_value, 'comment', comment)) AS scores_avg
```
maxdeichmann, line 73: This now returns the avg_value with all scores - correct? Instead we want to filter by the avg_value and fetch all the scores which belong to the generations. I doubt we can fetch them all in this cte.

web/src/server/api/routers/generations/getAllQuery.ts
```diff
@@ -10,28 +10,36 @@ import { getAllGenerationsSqlQuery } from "@/src/server/api/routers/generations/
...
+
+export type ScoreSimplified = {
+  name: string;
+  value: number;
+  comment?: string | null;
+};
+
 export type GetAllGenerationsInput = z.infer<typeof getAllGenerationsInput>;
 
+export type ObservationViewWithScores = ObservationView & {
```
maxdeichmann, line 22: For completeness, transformStreamToJsonLines should also take this type but only use the data it is currently using.

web/src/server/api/routers/generations/exportQuery/transforms/transformStreamToCsv.ts
```diff
@@ -1,51 +1,83 @@
...
+          "promptVersion",
         ];
 
         this.push(csvHeader.join(",") + "\n");
 
         isFirstChunk = false;
       }
-
       // Convert the generation object to a CSV line and push it
       const csvRow = [
```
maxdeichmann, line 50: can we also include parent_observation_id, type, modelParameters, completion_start_time

maxdeichmann: ... and do the same for the other export?

PR #1271, opened 2024-02-27
feat(sessions): extend table

RichardKruemmel:


web/src/components/table/use-cases/sessions.tsx
```diff
@@ -189,6 +201,34 @@ export default function SessionsTable({
...
+      accessorKey: "outputCost",
+      id: "outputCost",
+      header: "Output Cost",
+      enableHiding: true,
+      defaultHidden: true,
+      cell: ({ row }) => {
+        const value: number | undefined = row.getValue("outputCost");
+
+        return value !== undefined ? (
+          <span>{usdFormatter(value, 2, 2)}</span>
```
maxdeichmann, line 228: lets round here the same way we do on the generations table: https://github.com/langfuse/langfuse/blob/richard/lfe-461/web/src/components/table/use-cases/generations.tsx#L309

web/src/components/table/use-cases/sessions.tsx
```diff
@@ -203,6 +243,52 @@ export default function SessionsTable({
         ) : undefined;
       },
     },
+    {
+      accessorKey: "inputTokens",
```
maxdeichmann, line 247: Lets also steal the token ilustration from the traces table here

web/src/server/api/routers/sessions.ts
```diff
@@ -49,16 +49,41 @@ export const sessionRouter = createTRPCRouter({
...
+            totalLatency: number;
           }>
         >(Prisma.sql`
-      WITH observation_metrics AS (
+      WITH trace_latencies AS (
+        SELECT
+          trace_id,
+          EXTRACT(EPOCH FROM COALESCE(MAX("end_time"), MAX("start_time"))) - EXTRACT(EPOCH FROM MIN("start_time"))::double precision AS "latency"
+        FROM
+          "observations_view"
```
maxdeichmann, line 66: Can we fetch this from the observations table? This would have a major performance impact. Could you also quickly restructure this query the same way we did last week on the traces router? This has hughe performance improvements for our heavy users.

PR #1282, opened 2024-02-28
docs: fix link to db schema

sebhs:
## What does this PR do?

I realized that the link to the db schema in the contributing docs was broken so I fixed it :)


marcklingen:
thank you!

PR #1312, opened 2024-03-04
feat: add traces, spans, generations latency tables

maxdeichmann:


web/src/server/api/services/sqlInterface.ts
```diff
@@ -54,6 +54,8 @@ export const sqlInterface = z.object({
     "observations",
     "traces_scores",
     "traces_parent_observation_scores",
+    "traces_observationsview",
```
maxdeichmann, line 57: adding a dedicated view here to query this only if reuiqred.

web/src/server/api/services/query-builder.ts
```diff
@@ -399,8 +399,10 @@ const getMandatoryFilter = (
   switch (table) {
     case "traces":
     case "traces_scores":
+    case "traces_duration":
       return [traceFilter];
     case "traces_observations":
+    case "traces_observationsview":
     case "traces_parent_observation_scores":
       return [traceFilter, observationFilter];
     case "observations":
```
maxdeichmann, line 408: very important change: adds mandatory filters based on projectid

PR #1346, opened 2024-03-06
feat(prompts): extend prompts table

RichardKruemmel:


web/src/server/api/routers/generations/getAllQuery.ts
```diff
@@ -41,6 +41,7 @@ export const getAllQuery = protectedProjectProcedure
         count(*)
       FROM observations_view o
       JOIN traces t ON t.id = o.trace_id AND t.project_id = o.project_id
+      LEFT JOIN prompts p ON p.id = o.prompt_id
```
maxdeichmann, line 44: is this intended?

maxdeichmann (approved):

PR #1365, opened 2024-03-07
feat: add score `source`

marcklingen:


web/prisma/migrations/20240307175726_score_add_source/migration.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- Set default values conditionally with explicit casting
+UPDATE "scores"
+SET "source" = CASE
+    WHEN "name" = 'manual-score' THEN 'REVIEW'::"ScoreSource"
+    ELSE 'API'::"ScoreSource"
+END;
+
+-- Alter column to not null after setting default values
+ALTER TABLE "scores" ALTER COLUMN "source" SET NOT NULL;
```
maxdeichmann, line 15: This migration:
- locks the entire table during the migration. Ideally, we iterate over the table in chunks and update.
- removes the set not null constraint after updating. Between the two steps mentioned above, we could have added a new score in the meantime. In this case, the migration not null constraint would fail.

proposal:
- add another migration, in which:
  -  we get a full lock of the table
  - update the entire column again where the type is null
  - add the non null constraint
  - release the lock.

web/prisma/migrations/20240307185725_backfill_score_source/migration.sql
```diff
@@ -0,0 +1,32 @@
...
+   r record;
+   rows_processed int;
+BEGIN 
+   LOOP
+      rows_processed := 0;
+      FOR r in 
+         SELECT "name", "id"
+         FROM "scores"
+         WHERE "source" IS NULL
+         LIMIT 1000
```
maxdeichmann, line 14: i think we can speed this up by going to 10k

web/prisma/migrations/20240307185725_backfill_score_source/migration.sql
```diff
@@ -0,0 +1,32 @@
...
+         FOR UPDATE
```
maxdeichmann, line 15: I think we are good with this one as our environment is not highly concurrent. We mostly deal with historic data. For update gets a lock on the row which might timeout when doing that for 1k or more entries.

web/prisma/migrations/20240307175726_score_add_source/migration.sql
marcklingen: replaced by two steps:

step 1: https://github.com/langfuse/langfuse/pull/1365/files
step 2: https://github.com/langfuse/langfuse/pull/1369

web/prisma/migrations/20240307185725_backfill_score_source/migration.sql
marcklingen: thanks

PR #1402, opened 2024-03-12
feat(api): restful prompts api (/v2)

lmyslinski:
## What does this PR do?

This PR introduces new API endpoints for fetching prompts:

`/api/public/v2/prompts` - Returns a list of all prompts in a project. Only the latest version is returned for each prompt. The endpoint supports paging similarly to existing endpoints.
`/api/public/v2/prompts/[promptName]` - Returns all prompt versions of a single prompt. Supports filtering by the `active` and `version` parameters. When querying for a single/active version, the endpoint will return a single-array response

## Type of change

## Mandatory Tasks


lmyslinski:
Before I finish this out, I need to confirm some API design details with you. In the current form, using top-level GET for fetching a single object is unrestful. An obvious fix is to make it return the collection & add filtering options on top (ex. by tag, by version etc), while also adding a new `/prompts/{promptName}` endpoint for fetching singular prompts (althought I believe it makes sense to also return multiple prompts on this endpoint too).

This change will however break all existing API integrations, so I'd propose leaving the current API as-is, and adding the proposed changed under a different endpoint (`/prompt`?)

Alternatively, you could just start versioning the API with the `/v1/` prefix but seems a bit bigger effort-wise for my current capacity

marcklingen:
hi @lmyslinski, appreciate your contribution on this and sorry for slow response! it's unfortunate that we need to fix this, debated with @maxdeichmann what's the least effort way to do this without much effort and without breaking things for users of the langfuse sdks that get prompts via the current endpoint

Current idea
- all current apis stay on /api/public/.. including current api/public/prompts
- add new RESTful prompts routes on api/public/v2/prompts & api/public/vs/prompts/{promptId}

what do you think?

we could update the api reference and sdks to point to the v2 versions and keep the old one for backwards compatibility. unfortunately we cannot easily introduce param/header based versioning as the curent sdk do not include it -> api would need to default to the non-restful version -> unexpected and bad dx

lmyslinski:
> hi @lmyslinski, appreciate your contribution on this and sorry for slow response! it's unfortunate that we need to fix this, debated with @maxdeichmann what's the least effort way to do this without much effort and without breaking things for users of the langfuse sdks that get prompts via the current endpoint
> 
> Current idea
> 
> * all current apis stay on /api/public/.. including current api/public/prompts
> * add new RESTful prompts routes on api/public/v2/prompts & api/public/vs/prompts/{promptId}
> 
> what do you think?
> 
> we could update the api reference and sdks to point to the v2 versions and keep the old one for backwards compatibility. unfortunately we cannot easily introduce param/header based versioning as the curent sdk do not include it -> api would need to default to the non-restful version -> unexpected and bad dx

No worries, I'm aware you have your priorities. I think slowly adding v2 endpoints is a solid way to go, you can add new ones whenever they're needed. No need to force users to use one or the other exclusively. If that's fine with you I'd continue with this approach

marcklingen:
Agree, all same thoughts on this, happy to review once you are done with this!

lmyslinski:
@marcklingen A simple POC based on what we've talked about:

- /public/v2/prompts returns a list of all prompts in the `active` version.
- /public/v2/{promptName} also returns a list of prompt versions for a given prompt. Added `active` and `version` parameters which are mutually exclusive.

Took me a while to realize that `name` is the de facto primary key of a prompt - this is a bit problematic given the mutability of names & potential spaces in the query string but I'm sure you're already aware of that.

I've added what I consider minimal to support my use case - if you're happy with this range of features I'll add tests etc. to make it posh.

web/src/pages/api/public/v2/prompts.ts
```diff
@@ -0,0 +1,76 @@
...
+import { prisma } from "@langfuse/shared/src/db";
+import { NextApiRequest, NextApiResponse } from "next";
+import { z } from "zod";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+) {
+  await runMiddleware(req, res, cors);
+  if (req.method === "GET") {
```
marcklingen, line 13: we'd need pagination on this api similar to the traces or datasets endpoint

web/src/pages/api/public/v2/prompts.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+    if (authCheck.scope.accessLevel !== "all") {
+      return res.status(401).json({
+        message:
+          "Access denied - need to use basic auth with secret key to GET prompts",
+      });
+    }
+    try {
+      const prompt = await prisma.prompt.findMany({
```
marcklingen, line 36: I'd suggest to return a unique list of prompt names here including metadata similar to the prompt table in the UI.

Context: Prompt mgmt will change (multiple environments instead of a single `isActive`) and by using isActive here to directly return the prompts we'll likely have to change this api soon again.

web/src/pages/api/public/v2/prompts/[promptName].ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const { promptName, version, active } = GetPromptSchema.parse(req.query);
+
+    if (version && active) {
+      return res.status(404).json({
+        message: "Cannot use the active and version query parameters together",
+      });
+    }
+
+    const prompt = await prisma.prompt.findMany({
```
Comment (marcklingen) [inline web/src/pages/api/public/v2/prompts/[promptName].ts:50]: I'd suggest to order the prompts by version desc

web/src/pages/api/public/v2/prompts/[promptName].ts
```diff
@@ -0,0 +1,88 @@
...
+    if (authCheck.scope.accessLevel !== "all") {
+      return res.status(401).json({
+        message:
+          "Access denied - need to use basic auth with secret key to GET traces",
+      });
+    }
+
+    console.log("Trying to get prompt:", req.body, req.query);
+
+    const { promptName, version, active } = GetPromptSchema.parse(req.query);
```
Comment (marcklingen) [inline web/src/pages/api/public/v2/prompts/[promptName].ts:42]: uridecoding promptName makes sense to be able to parse special characters in the name

marcklingen (commented):
Like where this is going and I'm sorry for the slow response as I was heads down working on a new release.

Added some thoughts and would try to copy/adapt existing prompt tests also to these v2 endpoints to make sure they are robust

marcklingen:
@lmyslinski do you continue with this PR? Otherwise I'd take it from here soon to make this available

lmyslinski:
> @lmyslinski do you continue with this PR? Otherwise I'd take it from here soon to make this available

Was about to resume the work, sorry but I've had a lot going on lately - will wrap this up in a day/two tops

marcklingen:
> > @lmyslinski do you continue with this PR? Otherwise I'd take it from here soon to make this available
> 
> @marcklingen Was about to resume the work, sorry but I've had a lot going on lately - will wrap this up in a day/two tops

awesome, thank you. ping me once you are ready and I'll try to get this to main asap

web/src/pages/api/public/v2/prompts/[promptName].ts
Comment (lmyslinski) [reply web/src/pages/api/public/v2/prompts/[promptName].ts:42]: added

lmyslinski:
There's one failing test due to missing OpenAPI keys but I suppose this is not related to my changes

marcklingen:
thanks @lmyslinski! i'll merge to a branch to run all of the CI and do some last fixes to get this merged ahead of further changes to the prompts data schema

marcklingen:
@lmyslinski, finally released today. thanks for the wait, launch week kept us busy

lmyslinski:
> @lmyslinski, finally released today. thanks for the wait, launch week kept us busy

All good! Just moved prompt management via LF to staging

marcklingen:
Awesome!

PR #1409, opened 2024-03-13
feat(generations): extend sorting and filtering

RichardKruemmel:


maxdeichmann (approved):

web/src/components/table/use-cases/generations.tsx
```diff
@@ -378,8 +386,10 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
     {
       accessorKey: "statusMessage",
       header: "Status Message",
+      id: "statusMessage",
       enableHiding: true,
       defaultHidden: true,
+      enableSorting: true,
```
maxdeichmann, line 392: Lets remove this one

web/src/components/table/use-cases/generations.tsx
```diff
@@ -407,10 +466,12 @@ export default function GenerationsTable({ projectId }: GenerationsTableProps) {
         );
       },
       enableHiding: true,
+      enableSorting: true,
```
maxdeichmann, line 469: Lets also remove this one here. Not sure whether this works

web/src/components/table/use-cases/traces.tsx
```diff
@@ -567,6 +566,7 @@ export default function TracesTable({
         }
         columnVisibility={columnVisibility}
         setColumnVisibility={setColumnVisibility}
+        omittedFilter={omittedFilter}
       />
```
maxdeichmann, line 570: Instead of omitting, we could also remove this from the filterColumnDefinition above, right?

RichardKruemmel: Not 100% sure what you mean. We can uplift this logic:
```
const filterableColumns = filterColumnDefinition.filter(
    (c) => !omittedFilter?.includes(c.name),
  );
```
from the DataTableToolBar component to the use-case components. Is this what you are asking about?
If yes, my thoughts were that the functionality to omit filteroptions should be closer coupled to the toolbar rather than the use-case datatables. The downside of my approach is that we prop drill the omitFilter one layer deeper, but it seemed more logical to couple the omitFilter logic to the Toolbar, but I don't have  a strong opinion on this.

PR #1410, opened 2024-03-13
feat(sessions): extend sorting and filtering

RichardKruemmel:


web/src/components/table/use-cases/sessions.tsx
```diff
@@ -176,8 +176,10 @@ export default function SessionsTable({
     {
       accessorKey: "userIds",
       enableColumnFilter: !omittedFilter.find((f) => f === "userIds"),
+      id: "userIds",
```
maxdeichmann, line 179: image
I just saw that in our seed:load command we end up in a state where we have a sessionid with traces null. Can you check what is going on to me? It should be at least 1 in my understanding. Can be another ticket as this is not really related to this change.

web/src/components/table/use-cases/sessions.tsx
```diff
@@ -58,7 +58,7 @@ export default function SessionsTable({
   const userIdFilter: FilterState = userId
```
maxdeichmann, line 58: image

Do you know why the selection in the dropdown is not selected?

web/src/server/api/definitions/sessionsView.ts
```diff
@@ -3,38 +3,75 @@ import { type ColumnDefinition } from "@/src/server/api/interfaces/tableDefiniti
...
     id: "id",
     type: "string",
     internal: 's."id"',
   },
   {
-    name: "userId",
+    name: "User Ids",
+    id: "userIds",
     type: "string",
     internal: "array_to_string(t.\"userIds\", ', ')",
```
maxdeichmann, line 15: What do you think about re-using the tags filter here? I think having a drop down with userids would be most helpful.

web/src/components/table/use-cases/sessions.tsx
RichardKruemmel: Found the issue and fixed it. 👍

RichardKruemmel: Fixed both

web/src/components/table/use-cases/sessions.tsx
RichardKruemmel: After short investigation I believe that this session didn't get created by the seeder as the syntax of the seeder script is: `session-${i % 10}`.

web/src/components/table/use-cases/sessions.tsx
```diff
@@ -176,8 +189,10 @@ export default function SessionsTable({
     {
       accessorKey: "userIds",
       enableColumnFilter: !omittedFilter.find((f) => f === "userIds"),
-      header: "User ID",
+      id: "userIds",
+      header: "User IDs",
       enableHiding: true,
+      enableSorting: true,
```
maxdeichmann, line 195: How does sorting for multiple user Ids work?

web/src/components/table/use-cases/sessions.tsx
```diff
@@ -282,6 +302,7 @@ export default function SessionsTable({
       id: "usage",
       header: "Usage",
       enableHiding: true,
+      enableSorting: true,
```
maxdeichmann, line 305: Same as traces table. Does this work?

RichardKruemmel: Yes it does. Usage column behaves internally like total_token column. As reference the sessionTable.ts defintion:
`  {
    name: "Usage",
    id: "usage",
    type: "number",
    internal: 'o."totalTokens"',
  },`

PR #1411, opened 2024-03-13
feat(traces): extend sorting and filtering

RichardKruemmel:


web/src/components/table/use-cases/traces.tsx
```diff
@@ -363,6 +371,7 @@ export default function TracesTable({
           />
         );
       },
+      enableSorting: true,
```
maxdeichmann, line 374: @RichardKruemmel how would sorting work here? As far as i understand, the API does not support that yet.

maxdeichmann: We can leave this out for now and solve this in another PR so we get this merged

web/src/components/table/use-cases/traces.tsx
```diff
@@ -436,22 +447,26 @@ export default function TracesTable({
...
       accessorKey: "input",
       header: "Input",
+      id: "input",
       cell: ({ row }) => {
         const value: unknown = row.getValue("input");
         return <JSONView json={value} className="w-[500px]" />;
       },
       enableHiding: true,
       defaultHidden: true,
+      enableSorting: true,
```
maxdeichmann, line 457: pls remove

web/src/components/table/use-cases/traces.tsx
```diff
@@ -436,22 +447,26 @@ export default function TracesTable({
...
       accessorKey: "output",
       header: "Output",
+      id: "output",
       cell: ({ row }) => {
         const value: unknown = row.getValue("output");
         return <JSONView json={value} className="w-[500px] bg-green-50" />;
       },
       enableHiding: true,
       defaultHidden: true,
+      enableSorting: true,
```
maxdeichmann, line 469: lets not sort by io for now

web/src/server/api/definitions/tracesTable.ts
```diff
@@ -2,32 +2,75 @@ import {
...
+    type: "number",
+    internal: 'tm."completionTokens"',
+  },
+  {
+    name: "Total Tokens",
+    id: "totalTokens",
+    type: "number",
+    internal: 'tm."totalTokens"',
+  },
+  {
```
maxdeichmann, line 48: This and the two below need to be removed as well i guess

PR #1422, opened 2024-03-15
feat: add pnpm and turbo monorepo (#1401)

maxdeichmann:


agamboa23:
The error when building the worker's docker image is fixed when adding the "prisma-kysely": "^1.8.0" dependency to the shared workspace. This issue does not pop up on the web build since the dependency is installed by the web workspace itself. Hope it helps, Pura Vida!

PR #1451, opened 2024-03-20
feat(models): add sorting and filtering

RichardKruemmel:


web/src/features/filters/components/filter-builder.tsx
```diff
@@ -368,11 +368,20 @@ function FilterBuilderForm({
                         <SelectValue placeholder="" />
                       </SelectTrigger>
                       <SelectContent>
-                        {["true", "false"].map((option) => (
-                          <SelectItem key={option} value={option}>
-                            {option}
-                          </SelectItem>
-                        ))}
+                        {filter.column === "Maintainer"
```
maxdeichmann, line 371: can you use true / false or string options here? The filterbuilder should be abstract and now we move column specific logic in here.

Happy to support when you are blocked here.

marcklingen:
@maxdeichmann wdyt? I'd tend to not move to the table api here but rather wait for an updated end-to-end model UI in order not to add complexity that we will cut again in the near-term

PR #1505, opened 2024-03-26
fix: reset filter when navigating user details tab

RichardKruemmel:


RichardKruemmel:
FYI: I just renamed the PR again so that it is clear what this tries to fix 🙂

maxdeichmann:
@RichardKruemmel we will look at this again on new infrastructure.

PR #1569, opened 2024-04-03
feat(prompt): add filtering, sorting and pagination

RichardKruemmel:


RichardKruemmel:
Test comment

web/src/features/prompts/server/prompt-router.ts
```diff
@@ -19,40 +33,57 @@ export const CreatePrompt = z.object({
...
-        )
-        AND "project_id" = ${input.projectId}
-        ORDER BY name ASC`;
+      const orderByCondition =
+        input.orderBy?.column === "numberOfObservations"
+          ? Prisma.empty
+          : orderByToPrismaSql(input.orderBy, promptsTableCols);
+
+      const prompts = await ctx.prisma.$queryRaw<Array<Prompt>>(
+        generatePromptQuery(
```
marcklingen, line 69: inline this if only used here?

web/src/features/prompts/server/prompt-router.ts
```diff
@@ -497,3 +543,69 @@ export const createPrompt = async ({
...
+          GROUP BY name
+        )
+    AND "project_id" = ${projectId}
+  ${filterCondition}
+  ${orderCondition}
+  LIMIT ${limit} OFFSET ${page * limit};
+`;
+};
+
+function filterPromptsByCount(
```
marcklingen, line 572: either do this via sql or not at all, this does not work anymore when we add pagination to this table

web/src/features/prompts/server/prompt-router.ts
```diff
@@ -497,3 +543,69 @@ export const createPrompt = async ({
...
+        return p.observationCount <= value;
+      case ">=":
+        return p.observationCount >= value;
+      default:
+        return true;
+    }
+  });
+}
+
+function sortPromptsByObservationCount(
```
marcklingen, line 595: api-level sorting breaks when we add pagination

marcklingen: i'd reuse `orderByToPrismaSql`

web/src/features/prompts/server/prompt-router.ts
marcklingen: let's remove the filter and order of count alltogether

PR #1573, opened 2024-04-03
feat(dashboard): tag filter

RichardKruemmel:


web/src/pages/project/[projectId]/index.tsx
```diff
@@ -94,14 +94,22 @@ export default function Start() {
       },
     },
   );
-  const values = traceFilterOptions.data?.name || [];
+  const nameOptions = traceFilterOptions.data?.name || [];
+  const tagsOptions = traceFilterOptions.data?.tags || [];
 
   const traceName: ColumnDefinition[] = [
```
Comment (maxdeichmann) [inline web/src/pages/project/[projectId]/index.tsx:100]: can you rename this to e.g. `filterColumns`

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -43,7 +43,11 @@ export const TracesBarListChart = ({
     {
       projectId,
       from: "traces",
-      select: [{ column: "traceId", agg: "COUNT" }, { column: "traceName" }],
+      select: [
+        { column: "traceId", agg: "COUNT" },
+        { column: "traceName" },
+        { column: "tags", agg: "COUNT" },
```
maxdeichmann, line 49: what do we need this here for?

web/src/server/api/services/query-builder.ts
```diff
@@ -16,11 +16,11 @@ import { tableDefinitions } from "./tableDefinitions";
 import { tableColumnsToSqlFilter } from "@/src/features/filters/server/filterToPrisma";
 
 export type InternalDatabaseRow = {
-  [key: string]: bigint | number | Decimal | string | Date;
+  [key: string]: bigint | number | Decimal | string | Date | Array<string>;
 };
 
 export type DatabaseRow = {
-  [key: string]: string | number | Date | null;
+  [key: string]: string | number | Date | Array<string> | null;
```
maxdeichmann, line 23: Why do we need the above? I think we never have to fetch the tags via the API.

web/src/server/api/services/query-builder.ts
```diff
@@ -361,6 +361,8 @@ const outputParser = (output: InternalDatabaseRow[]): DatabaseRow[] => {
         // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
       } else if (val === null) {
         newRow[key] = val;
+      } else if (val instanceof Array) {
+        newRow[key] = val;
```
maxdeichmann, line 365: can this be removed as well?

web/src/server/api/services/tableDefinitions.ts
```diff
@@ -114,6 +115,14 @@ export const calculatedTotalCost = {
   internal: 'o."calculated_total_cost"',
 } as const;
 
+export const traceTags = {
+  name: "Tags",
+  id: "tags",
+  type: "arrayOptions",
+  options: [] as OptionsDefinition[],
```
maxdeichmann, line 122: what do we need these options for?

web/src/server/api/services/sqlInterface.ts
```diff
@@ -34,6 +34,7 @@ export const groupByInterface = z.array(
     }),
     z.object({ type: z.literal("number"), column: z.string() }),
     z.object({ type: z.literal("string"), column: z.string() }),
+    z.object({ type: z.literal("arrayOptions"), column: z.string() }),
```
maxdeichmann, line 37: If we only filter by tags, why do we need to have it in the group by?

web/src/server/api/services/tableDefinitions.ts
RichardKruemmel: traceTags needs to be of type ColumnDefinition. So far we declared tags columnDefinitions as type arrayOptions, therefore I thought we should continue doing this. But in this case it only adds overhead and does not change the functionality. Good point, I changed it to type string and this enables us to remove the options.

maxdeichmann (approved):

maxdeichmann (approved):

PR #1584, opened 2024-04-04
feat(ui): Add header tooltips for Input and Output columns

RichardKruemmel:


marcklingen:
@maxdeichmann I'd prefer to remove the restriction of how many items to show on tables, what do you think?

maxdeichmann:
> @maxdeichmann I'd prefer to remove the restriction of how many items to show on tables, what do you think?

For that we would need to test that. 100 network requests for a page will cause lots of state updates. I tested this once and ran into flaky UI.

PR #1619, opened 2024-04-08
feat(api): expose promptId via GET /observations

fancyweb:
## What does this PR do?

Expose `observations.prompt_id` on the `/api/public/observations` endpoint.

It fixes the fact that the `prompt_id` attribute from the `Observation` resource in the Python SDK (https://github.com/langfuse/langfuse-python/blob/a95e0718327793246a091ea15dca4e6369892301/langfuse/api/resources/commons/types/observation.py#L33) is always `None` since it's not exposed.

## Type of change


marcklingen (approved):
Thanks, this is great!

PR #1657, opened 2024-04-10
perf: increase trpc route timeout to support larger exports

hassiebp:
Increase trpc route timeout to support larger exports


marcklingen:
sounds like good intermediate solution to me. We can move exports to worker in v3 and notify via email when done

@maxdeichmann pls merge if ok for you

PR #1739, opened 2024-04-16
feat: sentry exception tracking for authentication error page

marcklingen:


web/src/components/error-page.tsx
```diff
@@ -24,3 +26,21 @@ export const ErrorPage = ({
...
+  title = "Error",
+  message,
+}: {
+  title?: string;
+  message: string;
+}) => {
+  useEffect(() => {
+    // Capture the error with Sentry
+    if (window !== undefined)
+      captureException(
```
maxdeichmann, line 40: Should this only be triggered if we have the sentry env variable available?

marcklingen: this "noops" when no sentry env is set

related issue: https://arc.net/l/quote/lcagvjcp

marcklingen: tested, works without sentry env without further wrapping

PR #1861, opened 2024-04-25
perf: reduce sentry sampling rate

maxdeichmann:


passionworkeer:
Hi! Thanks for the clear repro and the workaround sketch — that made this straightforward to track down.

I'd like to take a crack at this. The plan:

- Route `BatchEvaluationRunner._fetch_batch_with_retry` through `GET /api/public/v2/observations` (cursor pagination). The v3 `trace.list` and the legacy `observations_v1.get_many` endpoints are unavailable on v4 events_only deployments, and the v2 endpoint is the only read path that works there.
- For `scope='traces'`, fetch v2 observations and collapse to one representative per `trace_id`, preferring `is_root_observation=True`. The v2 endpoint has no trace-level read; this matches your "fetch v2 observations grouped by traceId" workaround.
- For `scope='observations'`, pass the response through unchanged.
- `MapperFunction` adds `ObservationV2` to its `item` union so existing mappers keep working with the input/output/metadata fields they already read.

A few choices I want to flag before I send the PR:

1. **Behaviour change for `scope='traces'`**: items passed to the mapper go from `TraceWithFullDetails` to `ObservationV2`. Mappers that previously consumed the full `observations` / `scores` lists on a trace need to adapt. Happy to scope down to `scope='observations'` only if you'd rather land `scope='traces'` in a follow-up.
2. **Default**: I'm switching the default endpoint rather than gating it behind a flag, since the v3 endpoints are deprecated (EOL 2026-11-16) and the v2 endpoint is available on both v3 and v4.

I've drafted the change locally with 5 unit tests (cursor pagination, trace grouping, unknown-scope rejection, no-root fallback, trace_id-less filtering) and verified it against an unmodified `main` (tests fail before the fix, pass after). Currently running the full v4 events_only docker-compose stack locally for an end-to-end check before I push.

I work on this through a local automation workflow, so I'll send the PR rather than comment / push piecemeal — happy to adjust direction based on what you'd prefer.

PR #1928, opened 2024-04-30
fix: openapi spec publishing and add postman collection

marcklingen:


marcklingen:
seems to have solved the vercel deployment in issue in preview deployment ✅

PR #1950, opened 2024-05-02
fix: use provided job delay instead of default if provided

### LINKED ISSUE #1949: bug: Evaluation config delay is not saved
### Describe the bug

When creating an evaluation config, the timeout set in the UI is not applied even when the data is sent, and the default of 10 seconds is set.

### To reproduce

Create an evaluation config with a non-default delay value.

_No response_

### Additional information

_No response_

Yes

TimBeyer:
## What does this PR do?

Currently a bug prevents custom job delays to actually be applied and instead it uses the default.
At the same time, the default was also hardcoded into the zod default validator.
This fixes it so that the constant is being used for the default and the input is used to pass to the job configuration object.

Fixes https://github.com/langfuse/langfuse/issues/1949

## Type of change

## Mandatory Tasks

### DIFF
--- a/web/src/ee/features/evals/server/router.ts
+++ b/web/src/ee/features/evals/server/router.ts
@@ -249,7 +249,7 @@ export const evalRouter = createTRPCRouter({
         filter: z.array(singleFilter).nullable(), // re-using the filter type from the tables
         mapping: z.array(variableMapping),
         sampling: z.number().gte(0).lte(1),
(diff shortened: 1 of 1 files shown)


maxdeichmann (approved):

maxdeichmann:
@TimBeyer thanks a lot for fixing that. I will release that in a minute.

PR #2011, opened 2024-05-08
chore: backfill projectid on score

marcklingen:


packages/shared/prisma/migrations/20240508132736_scores_backfill_project_id/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Backfill project_id on existing scores based on linked trace_id
+UPDATE scores
+SET project_id = traces.project_id
+FROM traces 
+WHERE scores.trace_id = traces.id;
```
maxdeichmann, line 5: 1. did you test this out locally?
2. please do not run this on production directly. First manually execute, then migrate this with null check on project_id column

marcklingen: 1. tested
2. done
3. good catch, added

PR #2014, opened 2024-05-08
chore(scores): drop fk constraint on traces/observations

marcklingen:


web/src/server/api/routers/scores.ts
```diff
@@ -121,29 +118,32 @@ export const scoresRouter = createTRPCRouter({
...
+  //                     some: {
+  //                       userId: ctx.session.user.id,
+  //                     },
+  //                   },
+  //                 },
+  //               },
+  //             }),
+  //       },
+  //     }),
+  //   ),
```
maxdeichmann, line 142: should this be deleted?

PR #2019, opened 2024-05-08
fix: make latest label a reserved label

hassiebp:
- Do not allow setting / unsetting 'latest' label in UI
- If latest prompt version is getting deleted, set the 'latest' label on the next prompt version if any

### DIFF
--- a/web/src/features/prompts/server/routers/promptRouter.ts
+++ b/web/src/features/prompts/server/routers/promptRouter.ts
@@ -19,6 +19,7 @@ import {
   singleFilter,
   tableColumnsToSqlFilterAndPrefix,
 } from "@langfuse/shared";
+import { LATEST_PROMPT_LABEL } from "@/src/features/prompts/constants";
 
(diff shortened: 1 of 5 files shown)


web/src/features/prompts/server/routers/promptRouter.ts
```diff
@@ -312,12 +313,44 @@ export const promptRouter = createTRPCRouter({
           ctx.prisma,
         );
 
-        await ctx.prisma.prompt.delete({
-          where: {
-            id: input.promptVersionId,
-            projectId: input.projectId,
-          },
-        });
+        const transaction = [
```
marcklingen, line 316: this seems like an interactive transaction but done without using interactive transactions. I'd assume these prisma calls to be executed before hitting `ctx.prisma.$transaction`. did you check that these actually get processed inside of a transaction?

I'd suggest to make them non-interactive
- fetch next `latest` prompt version outside of transaction
-  delete current latest version and mark next latest as latest within a transaction

PR #2078, opened 2024-05-16
feat: move scores table on trace and observation preview into seperate tab 

marliessophie:


web/src/components/trace/ObservationPreview.tsx
```diff
@@ -33,6 +28,12 @@ export const ObservationPreview = (props: {
   currentObservationId: string;
   traceId: string;
 }) => {
+  const [selectedTab, setSelectedTab] = useState("preview");
```
marcklingen, line 31:
```suggestion
  const [selectedTab, setSelectedTab] = useQueryParam(
    "tab",
    withDefault(StringParam, "preview"),
  );
```

moving this state to a query param makes the view shareable via link

web/src/components/trace/ObservationPreview.tsx
```diff
@@ -138,64 +144,49 @@ export const ObservationPreview = (props: {
...
+          </>
+        )}
 
-        {observationWithInputAndOutput.data?.metadata ? (
-          <JSONView
-            key={observationWithInputAndOutput.data.id + "-metadata"}
-            title="Metadata"
-            json={observationWithInputAndOutput.data.metadata}
+        {selectedTab === "scores" && isScoreAttached ? (
+          <ScoresTablePreview
```
marcklingen, line 184: is it possible to reuse the existing scores datatable? it already has pagination, filters, sorting etc implemented.

necessary props to pass in order to make it work
- column ids to show as we need to hide irrelevant ids such as traceId
- current traceid/observation id as a default filter. could be implemented in a similar way as on the user tabs which also reuse the main scores/trace/observation tables

web/src/components/trace/ObservationPreview.tsx
```diff
@@ -138,64 +144,49 @@ export const ObservationPreview = (props: {
               key={preloadedObservation.id}
             />
           ) : null}
+          {isScoreAttached && (
```
marcklingen, line 147: i'd probably always show the table to make the interface stable/predictable

web/src/components/trace/ObservationPreview.tsx
marcklingen: the selection of preview vs scores could be stable while clicking through the observations of a trace

web/src/components/trace/ObservationPreview.tsx
```diff
@@ -33,6 +28,12 @@ export const ObservationPreview = (props: {
   currentObservationId: string;
   traceId: string;
 }) => {
+  const [selectedTab, setSelectedTab] = useState("preview");
+
+  useEffect(() => {
```
marcklingen, line 33: if this is meant to reset the view between different observations, I think adding a `key` on the ObservationPreview would do the trick as well

packages/shared/prisma/seed.ts
```diff
@@ -521,6 +542,17 @@ function createObjects(
 
     scores.push(...traceScores);
 
+    if (i !== DEMO_USER_ID_NUMBER) {
+      const user = {
+        id: `user-${i}`,
+        email: `user-${Math.random()}@langfuse.com`,
+        name: `User ${i}`,
+        image: `https://avatar.iran.liara.run/public/${i}`,
```
marliessophie, line 550: Chose some easy avatar generation page, please let me know how to ensure this url is secure & if you have strong feelings on not introducing this dependency. If in doubt, happy to remove! Intension for adding was only for marginal UI improvement.

web/src/components/table/use-cases/scores.tsx
```diff
@@ -218,14 +287,16 @@ export default function ScoresTable({
...
       },
     },
   ];
 
+  const columns = rawColumns.filter(
+    (c) => !!c.id && !hiddenColumns.includes(c.id),
+  );
+
   const [columnVisibility, setColumnVisibility] =
     useColumnVisibility<ScoresTableRow>("scoresColumnVisibility", columns);
```
marcklingen, line 301: if the columns are different between the different embeds, we might need to pass a table name for each of them in order to save the local storage state on different strorage keys

web/src/components/table/use-cases/scores.tsx
```diff
@@ -33,14 +40,36 @@ export type ScoreFilterInput = Omit<
...
+  hiddenColumns = [],
 }: {
   projectId: string;
   userId?: string;
+  traceId?: string;
+  observationId?: string;
   omittedFilter?: string[];
+  hiddenColumns?: string[];
 }) {
   const [paginationState, setPaginationState] = useQueryParams({
```
marcklingen, line 74: we probably need observationId = Null here in order to only show trace scores on the trace level. otherwise we'd need to somehow identify them within the scores table on the trace level

packages/shared/prisma/seed.ts
marliessophie: Adjusted, thank you 👍

packages/shared/prisma/seed.ts
```diff
@@ -14,6 +14,7 @@ import { ModelUsageUnit } from "../src";
 import { getDisplaySecretKey, hashSecretKey } from "../src/server/auth";
 
 const LOAD_TRACE_VOLUME = 10_000;
+const DEMO_USER_ID_NUMBER = 1;
```
marcklingen, line 17: i'd revert all changes to the seeder if we do not use them ehre, the users are not attached to any project thus not really useful in the seeder for the rest of the application

web/src/components/table/use-cases/scores.tsx
marliessophie: Decided to scope out of this PR as we'd require a `is null` filter on tables for this request, which we've scoped out into a separate feature request ticket on linear. For now: show all nested scores in the trace > scores tab.

marcklingen (approved):

PR #2121, opened 2024-05-22
feat(scores): add nullable authorUserId to scores

marliessophie:


packages/shared/prisma/schema.prisma
```diff
@@ -399,6 +399,7 @@ model Score {
     name          String
     value         Float
     source        ScoreSource
+    authorUserId  String?        @map("author_user_id")
```
marcklingen, line 402: agree to not add an fk on user.id here, we need to gracefull handle this later on read though

PR #2126, opened 2024-05-22
feat(score_configs): add table via migration 

marliessophie:


packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +426,36 @@ enum ScoreSource {
...
+    updatedAt    DateTime      @default(now()) @updatedAt @map("updated_at")
+    projectId    String        @map("project_id")
+    project      Project       @relation(fields: [projectId], references: [id], onDelete: Cascade)
+    name         String
+    dataType     ScoreDataType
+    isArchived   Boolean       @default(false) @map("is_archived")
+    minValue     Float?        @map("min_value")
+    maxValue     Float?        @map("max_value")
+    categories   Json?         @map("categories")
+    authorUserId String?       @map("author_user_id")
```
marcklingen, line 441: do we need author of an annotation config? we can cover edit history via the auditlog table

packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +426,36 @@ enum ScoreSource {
...
 }
 
+model AnnotationConfig {
+    id           String        @id @default(cuid())
+    createdAt    DateTime      @default(now()) @map("created_at")
+    updatedAt    DateTime      @default(now()) @updatedAt @map("updated_at")
+    projectId    String        @map("project_id")
+    project      Project       @relation(fields: [projectId], references: [id], onDelete: Cascade)
+    name         String
+    dataType     ScoreDataType
```
marcklingen, line 436: missing map "data_type"

packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +426,36 @@ enum ScoreSource {
...
+    description  String?
+    score        Score[]
+
+    @@unique([id, projectId]) // used for upserts via prisma
+    @@index([dataType])
+    @@index([isArchived])
+    @@index([projectId])
+    @@index([categories])
+    @@index([authorUserId])
+    @@map("annotation_config")
```
marcklingen, line 451: plural

packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +426,36 @@ enum ScoreSource {
...
+}
+
+enum ScoreDataType {
```
marcklingen, line 454: score or annotation?

packages/shared/prisma/migrations/20240522141943_scores_add_annotation_config_id/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "scores" ADD COLUMN     "annotation_config_id" TEXT;
```
marcklingen, line 2: can be in main migration

packages/shared/prisma/schema.prisma
marliessophie: Renamed to annotation 👍

packages/shared/prisma/schema.prisma
marliessophie: I've removed it 👍

packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +426,34 @@ enum ScoreSource {
...
+
+    @@unique([id, projectId]) // used for upserts via prisma
+    @@index([dataType])
+    @@index([isArchived])
+    @@index([projectId])
+    @@index([categories])
+    @@map("score_configs")
+}
+
+enum ScoreConfigDataType {
```
marcklingen, line 452: simplify to ScoreDataType? This is tough to edit once there are tables that rely on this

marliessophie: I'd prefer `ScoreConfigDataType` as we have our `scores` table which I would expect to hold this data type when called `ScoreDataType`, not the score configs table. Another suggestion could be `ConfigDataType`, but that would be too generic imo. What would be the reason for simplifying? I think we'd loose context in doing so.

marcklingen: No strong opinion here. I'd see this as the config constraints a score to a datatype

marliessophie: OK, would you be happy to leave as `ScoreConfigDataType` then? Then we can get this merged ☺️

packages/shared/prisma/schema.prisma
```diff
@@ -422,6 +428,34 @@ enum ScoreSource {
...
+    createdAt   DateTime      @default(now()) @map("created_at")
+    updatedAt   DateTime      @default(now()) @updatedAt @map("updated_at")
+    projectId   String        @map("project_id")
+    project     Project       @relation(fields: [projectId], references: [id], onDelete: Cascade)
+    name        String
+    dataType    ScoreDataType @map("data_type")
+    isArchived  Boolean       @default(false) @map("is_archived")
+    minValue    Float?        @map("min_value")
+    maxValue    Float?        @map("max_value")
+    categories  Json?         @map("categories")
```
marliessophie, line 442: Future proofing question: we are sure we want to use the `Json` data type for categories rather than a separate relational model, right? It suffices our current use case for sure but trying to future proof. Json is likely simpler/more flexible for maintenance, more performant as we're always fetching the entire object. Con: we're not enforcing data integrity on a schema level. Should not be a concern while only creating scoreConfigs via app, but if we allow creation of configs via API or need scalability for writes (don't see strong use case) might become problematic.  

My view: relational model is an overkill in the mid-term. We can always revisit this conversation when allowing config creation via API, don't see it as a one-way door. 

Thoughts? If no concerns + you agree feel free to resolve comment.

marcklingen: agree, I'd keep this as JSON. If we really want to we can still backfill a separate table via a migration if we enforce the structure here.

marcklingen (commented):
overall looks good to me! added quick note re backfill, i think defaulting here is probably fine and makes this faster to deploy. we can later switch to non-null without default and make sure to always set the type on the application layer

packages/shared/prisma/schema.prisma
maxdeichmann: agree

packages/shared/prisma/schema.prisma
```diff
@@ -406,7 +406,7 @@ model Score {
     observationId String?        @map("observation_id")
     configId      String?        @map("config_id")
     stringValue   String?        @map("string_value")
-    dataType      ScoreDataType  @default(NUMERIC) @map("data_type")
+    dataType      ScoreDataType? @default(NUMERIC) @map("data_type")
```
marcklingen, line 409: non-null enum with default works without backfill. you'd need to regenerate the migration here. will work out of the box

default value will be written to metadata of column in postgres this way, thus needs no backfill. on read, behaves in the same way as if we did a backfill

marliessophie: Super cool, did not know. Thanks!

PR #2139, opened 2024-05-24
feat: use custom keys for playground

hassiebp:


packages/shared/prisma/seed.ts
```diff
@@ -165,6 +166,24 @@ async function main() {
 
     await uploadObjects(traces, observations, scores, sessions, events);
 
+    // If openai key is in environment, add it to the projects LLM API keys
+    const OPENAI_API_KEY = process.env.OPENAI_API_KEY;
```
maxdeichmann, line 170: thanks!

web/src/ee/features/playground/server/authorizeRequest.ts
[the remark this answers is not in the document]
maxdeichmann: WDYT of sharing code with this [middleware](https://github.com/langfuse/langfuse/blob/56e66c195119f0cea0535cd61eafe5d7068b6ff8/web/src/server/api/trpc.ts#L175)?
In case we change something to the middleware, the change would also be reflected there. Less important for this case, but we also have the admin logic in the middlerware which we could apply here.

web/src/ee/features/playground/server/authorizeRequest.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const authorizeRequestOrThrow = async (
+  projectId: string,
+): Promise<AuthorizeRequestResult> => {
+  if (!getIsCloudEnvironment())
+    throw new ApiError("This endpoint is available in Langfuse cloud only.");
+
+  const authOptions = await getAuthOptions();
+  const session = await getServerSession(authOptions);
```
maxdeichmann, line 18: How is the JWT cookie passed from the request in here? Maybe we can use [this](https://github.com/langfuse/langfuse/blob/56e66c195119f0cea0535cd61eafe5d7068b6ff8/web/src/server/auth.ts#L363-L364) here? This is also used in tRPC [here](https://github.com/langfuse/langfuse/blob/56e66c195119f0cea0535cd61eafe5d7068b6ff8/web/src/server/api/trpc.ts#L53)

web/src/features/email/lib/project-invitation.ts
```diff
@@ -11,7 +11,7 @@ const langfuseUrls = {
   STAGING: "https://staging.langfuse.com",
 };
 
-const authUrl =
+const getAuthURL = () =>
```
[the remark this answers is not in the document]
maxdeichmann: does this break docker in any way?

web/src/server/api/trpc.ts
```diff
@@ -242,7 +243,11 @@ const enforceTraceAccess = t.middleware(async ({ ctx, rawInput, next }) => {
     ({ id }) => id === trace.projectId,
   );
 
-  if (!trace.public && !sessionProject && ctx.session?.user?.admin !== true)
+  if (
```
maxdeichmann, line 246: can we delete const sessionProject = ctx.session?.user?.projects.find( above?

PR #2157, opened 2024-05-27
feat(scores): add annotation slider and score config settings for categorical annotation

marliessophie:


packages/shared/prisma/seed.ts
marliessophie, line 1: Will add some more helpful data to seeder, still left on to-do list 👍

marliessophie, line 1: FYI only Seeder data will still be improved before launch

pnpm-lock.yaml
marliessophie, line 1: See dependency installation question above, thanks

package.json
```diff
@@ -69,5 +69,9 @@
         "pr": ":rocket: _This pull request is included in v${version}. See [${releaseName}](${releaseUrl}) for release notes._"
       }
     }
+  },
+  "dependencies": {
+    "@radix-ui/react-toggle-group": "^1.0.4",
+    "vaul": "^0.9.1"
```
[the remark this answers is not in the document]
marcklingen: you need to cd into the web proejct and install dependencies there. that's why you probably got an alert by pnpm to check whether you really want to do an install on the workspace level

pnpm-lock.yaml
marcklingen: yep, need to move these to the `web` project

package.json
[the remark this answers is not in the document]
marliessophie: thanks, will do 🙏

packages/shared/prisma/migrations/20240531155458_backfill_manual_scores/migration.sql
marliessophie, line 1: Would be good to review this migration please

web/src/features/manual-scoring/components/multi-select-key-values.tsx
marliessophie, line 1: I created this component in a generic fashion to be able to replace the mutli-select component used in filters to only use this one. It should work out of the box but did not want to pollute this PR with an unrelated change. Have added to backlog.

marcklingen (commented):
backfill notes

web/src/features/manual-scoring/components/multi-select-key-values.tsx
marcklingen: looks really nice

web/src/server/api/routers/scores.ts
```diff
@@ -149,38 +159,49 @@ export const scoresRouter = createTRPCRouter({
         throw new Error("No trace with this id in this project.");
       }
 
-      const score = await ctx.prisma.score.create({
-        data: {
+      try {
+        const score = await ctx.prisma.score.create({
```
marcklingen, line 163: use upsert to make sure that no score config can be applied twice via annotation workflow

web/src/components/trace/index.tsx
```diff
@@ -103,11 +104,13 @@ export function Trace(props: {
             trace={props.trace}
             observations={props.observations}
             scores={props.scores}
+            configs={props.configs}
```
marcklingen, line 107: we do not need to pass the prop down, we could just fetch the config down in the annotation drawer as react query will cache it locally if is needed mutliple times within the same session

web/src/features/manual-scoring/components/multi-select-key-values.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { Separator } from "@/src/components/ui/separator";
+
+type MultiSelectOptions = {
+  value: string;
+  key?: string;
+  count?: number;
+  disabled?: boolean;
+};
+
+export function MultiSelectKeyValues<
```
marcklingen, line 30: move to generic components (src/components)?

web/src/server/api/routers/scoreConfigs.ts
```diff
@@ -0,0 +1,85 @@
...
+  projectId: z.string(), // Required for protectedProjectProcedure
+});
+
+const category = z.object({
+  label: z.string().min(1),
+  value: z.number(),
+});
+
+export const scoreConfigsRouter = createTRPCRouter({
+  all: protectedProjectProcedure
```
marcklingen, line 20: add pagination

web/src/server/api/routers/scoreConfigs.ts
```diff
@@ -0,0 +1,85 @@
...
+        return {
+          configs,
+          totalCount: configsCount,
+        };
+      } catch (error) {
+        console.log(error);
+      }
+    }),
+
+  create: protectedProjectProcedure
```
marcklingen, line 54: check if no other config with the same name exists?

web/src/features/manual-scoring/components/multi-select-key-values.tsx
marliessophie: Agree! Was planning to do in follow up, does that work?

marcklingen: follow up for marlies

web/src/server/api/routers/scoreConfigs.ts
marcklingen: future

packages/shared/prisma/migrations/20240604133340_backfill_manual_scores/migration.sql
```diff
@@ -0,0 +1,50 @@
+-- Create manual-score config for any project with manual scores and link config_id to scores
+BEGIN;
+WITH project_configs AS (
```
maxdeichmann, line 3: I think we could have found a simpler way to move this data, which might be less elegant and produces some garbage data but cleans up. As this would happen in a transaction, this would be save anyways.

packages/shared/prisma/migrations/20240604133340_backfill_manual_scores/migration.sql
```diff
@@ -0,0 +1,50 @@
...
+			sc.name = 'manual-score'
+			AND sc.project_id = s.project_id)
+	RETURNING
+		id,
+		project_id
+)
+UPDATE
+	scores
+SET
+	config_id = pc.id
```
maxdeichmann, line 44: i assume all scores with 'manual-score' in the same project should have the same config_id?

marliessophie: Correct 👍

PR #2173, opened 2024-05-28
fix(ui): Allow for long model names in filter

RichardKruemmel:


web/src/features/filters/components/multi-select.tsx
```diff
@@ -118,9 +118,9 @@ export function MultiSelect({
                     >
                       <Check className={cn("h-4 w-4")} />
                     </div>
-                    <span>{option.value}</span>
+                    <span className="overflow-scroll">{option.value}</span>
```
marcklingen, line 121:
```suggestion
                    <span className="overflow-x-scroll">{option.value}</span>
```

lmit this to x in order to prevent weird side-effects?

.env.dev.example
```diff
@@ -47,5 +47,5 @@ REDIS_PORT=6379
 REDIS_AUTH="myredissecret"
 
 LANGFUSE_WORKER_PASSWORD=mybasicauthsecret
-# openssl rand -base64 32 used only here
+# openssl rand hex 32 used only here
```
marcklingen, line 50: conscious change within this PR?

RichardKruemmel: Yes conscious change as I wanted to bring up the issue. When I use the key generates with -base64 I get errors. Can you confirm that it's not a local issue?

marcklingen: needs -hex flag, I'll edit and merge

PR #2180, opened 2024-05-29
feat(api): add prompt tags via API

RichardKruemmel:


marcklingen (changes requested):
need to add tag to all existing prompts with the same name as they should be the same across versions. see trpc route for creating new prompt version for details, I think there is an abstraction that can be reused

Change also needs tests, suggested test cases

- prompt version 1 [tag]
- version 2 no tags

--> check all versions, should all have [tag]

- add new version 3, [tag2,tag3]

--> all three versions now tag2&tag3

marcklingen (approved):

PR #2225, opened 2024-06-04
perf: boost sessions table performance

hassiebp:
boost sessions table performance

### DIFF
--- a/web/src/server/api/routers/sessions.ts
+++ b/web/src/server/api/routers/sessions.ts
@@ -62,56 +62,54 @@ export const sessionRouter = createTRPCRouter({
             totalTokens: number;
           }>
         >(Prisma.sql`
-      WITH observation_metrics AS (
         SELECT
-          t.session_id,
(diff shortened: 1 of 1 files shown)


web/src/server/api/routers/sessions.ts
```diff
@@ -62,56 +62,54 @@ export const sessionRouter = createTRPCRouter({
...
+            AND t.session_id = s.id
+          GROUP BY
+            t.session_id) AS t ON TRUE
+        LEFT JOIN LATERAL (
+          SELECT
+            EXTRACT(EPOCH FROM COALESCE(MAX(o. "end_time"), MAX(o. "start_time"), t. "max_timestamp")) - EXTRACT(EPOCH FROM COALESCE(MIN(o. "start_time"), t. "min_timestamp"))::double precision AS "sessionDuration",
+            SUM(COALESCE(o. "calculated_input_cost", 0)) AS "inputCost",
+            SUM(COALESCE(o. "calculated_output_cost", 0)) AS "outputCost",
+            SUM(COALESCE(o. "calculated_total_cost", 0)) AS "totalCost",
+            SUM(o.prompt_tokens) AS "promptTokens",
```
maxdeichmann, line 103: how does this work here? Do i understand correctly, that we sum over all observations which have any of the traceIds? I think we need to group these by sessionId. But maybe i am missing logic here.

PR #2258, opened 2024-06-06
fix(ui): improve ux across delete buttons

mishushakov:
## What does this PR do?

This PR improves UX across delete buttons. They are now always visible in the interface, but disabled if user has no access. Moreover, I have added a tooltip confirmation for the delete button in the models screen to match the UX in the rest of the app and standardised the trash button style.

Screenshot 2024-06-06 at 13 47 06

## Type of change

## Mandatory Tasks


web/src/components/table/use-cases/models.tsx
```diff
@@ -222,13 +228,12 @@ export default function ModelTable({ projectId }: { projectId: string }) {
       accessorKey: "actions",
       header: "Actions",
       cell: ({ row }) => {
-        return row.original.maintainer === "User" ? (
+        return (
```
marcklingen, line 231: I agree that showing disabled buttons for users without sufficient roles is an improvement here.

Do you think the UX is better when showing a disabled delete button on langfuse-provided models even though they cannot be deleted?

marcklingen (commented):
thanks! great changes, especially the disabled behavior is better for users without sufficient scopes. unsure about delete buttons on builtin models though

web/src/components/table/use-cases/models.tsx
mishushakov: It should be signified in the UI in this case. Another way would be to present an info icon and once clicked (or hovered) it would explain that built-in models are non-removable.

marcklingen: maybe a simple `title="Built-in models cannot be deleted"`?

mishushakov: yep, good enough for me

mishushakov: in general the screen deserves some more love - I'll follow up with another PR

mishushakov: hey @marcklingen are you waiting for me here? let me know if we're ready to move with this pr ;)

marcklingen: sorry, just got to this as today was super busy, thank you!

marcklingen: > in general the screen deserves some more love - I'll follow up with another PR

agree, this is one of the most complex tables right now. Some change to the price calculation logic are necessary soon, but the underlying logic will only become more complex.

One option would be to render the groups of models more nicely as the full list including changes over time are hardly interesting and could be moved to separate pages.

web/src/components/ui/button.tsx
```diff
@@ -6,7 +6,7 @@ import { cn } from "@/src/utils/tailwind";
 import { Loader2 } from "lucide-react";
 
 const buttonVariants = cva(
-  "inline-flex items-center justify-center whitespace-nowrap rounded-md text-sm font-medium ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:opacity-50",
+  "inline-flex items-center justify-center whitespace-nowrap rounded-md text-sm font-medium ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50",
```
marcklingen, line 9: strong preference for cursor-none (this was the default of this shadcn/ui button) to indicate that it is disabled

mishushakov: I believe in this case you wouldn't see the title tooltip?

marcklingen: good catch, reenabled pointer-events when disabled for this button and added cursor-not-allowed as title takes some time to show up: https://github.com/langfuse/langfuse/commit/cc08408f0a9d67ff27276e73819125c1e6c53307

PR #2285, opened 2024-06-10
feat(auth): add custom OIDC IDP provider

bbartels:
## What does this PR do?

This PR adds the ability to add any custom OIDC IDP SSO integration. We have an air-gapped on-premise setup at work and have an internal IDP that doesn't match any of the ones currently provided.

Docs PR here: https://github.com/langfuse/langfuse-docs/pull/684

Fixes # (issue)

The below screenshot is what the login page would look like with a custom provider and `AUTH_CUSTOM_NAME: IdentityServer`

image

## Type of change

## Mandatory Tasks

## Checklist

There do not seem to be any existing tests around SSO and it would be rather difficult to setup. I've confirmed the changes work locally and am happy to demonstrate to a maintainer if required.


marcklingen:
Thanks for your contribution on this and the docs change!

Did you try this change end-to-end with your setup?

What kind of name needs to be provided in the callback? Currently the docs change uses `cognito` which is most likely not the correct callback

bbartels:
The `AUTH_CUSTOM_NAME` is simply the string you want to display on the custom SSO button. So if you check the provided screenshot it says `IdentityServer` as that is what `AUTH_CUSTOM_NAME` was set to in that instance.

And yes it's fully end-to-end tested with a replica scenario (I was unable to build langfuse on prem due to some prisma dependencies on the internet). I've just used the Duende IdentityServer quickstart samples to somewhat reproduce what we have on-prem. I am happy to demonstrate!

web/src/env.mjs
```diff
@@ -71,6 +71,11 @@ export const env = createEnv({
     AUTH_COGNITO_CLIENT_SECRET: z.string().optional(),
     AUTH_COGNITO_ISSUER: z.string().url().optional(),
     AUTH_COGNITO_ALLOW_ACCOUNT_LINKING: z.enum(["true", "false"]).optional(),
+    AUTH_CUSTOM_CLIENT_ID: z.string().optional(),
+    AUTH_CUSTOM_CLIENT_SECRET: z.string().optional(),
+    AUTH_CUSTOM_ISSUER: z.string().optional(),
```
marcklingen, line 76:
```suggestion
    AUTH_CUSTOM_ISSUER: z.string().url().optional(),
```

Can we enforce url here as it is used as a hostname later on?

marcklingen (commented):
This looks good overall! I've made a few minor changes and have one open question

web/src/env.mjs
bbartels: Yep that should work, frankly unsure why it wasn't already aligned!

marcklingen:
did you `pnpm i` before running this?

marcklingen:
This looks like a build error and I cannot reproduce it locally

bbartels:
Ah yes, I hadn't realized you made any package changes 🤦. Tested locally with your changes, works end-to-end 🙂

marcklingen:
awesome, merging now, thanks again!

marcklingen (approved):

bbartels:
Thanks for the quick review, I appreciate it!

PR #2287, opened 2024-06-10
fix(ui): order tooltip items on dashboard by value desc

RichardKruemmel:


RichardKruemmel:
This change mainly concerned functionality. The UI was supposed to stay the same. I therefore copied the UI from tremors Default Chart Tool Tip. Please see: https://github.com/tremorlabs/tremor/blob/main/src/components/chart-elements/common/ChartTooltip.tsx#L82

I also fixed the dark mode of tab select for phone sizes in the dashboard.

marcklingen (approved):
really nice, thx!

PR #2297, opened 2024-06-11
fix: update number formatting to "en-US" locale

RichardKruemmel:


RichardKruemmel:
There is no "us" option for Intl.NumberFormat, which probably introduced unexpected behaviour.
Found this repo to visualize different locale options: https://codepen.io/maqnus/pen/KGzywv

marcklingen (approved):
awesome, thanks

PR #2329, opened 2024-06-12
fix: clean up broken migration history

maxdeichmann:


packages/shared/scripts/fix-migration-history.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
marliessophie, line 1: Delete

web/Dockerfile
```diff
@@ -78,7 +78,9 @@ COPY --from=builder --chown=nextjs:nodejs /app/web/public ./web/public
 COPY --from=builder --chown=nextjs:nodejs /app/packages/shared/prisma ./packages/shared/prisma
```
marliessophie, line 78: revert changes to this file

marliessophie (approved):
LGTM

PR #2360, opened 2024-06-17
feat: adds exports to sessions table

hassiebp:
Adds bulk data exports to sessions table


packages/shared/src/queries/getSessionTableSQL.ts
```diff
@@ -0,0 +1,88 @@
...
+        t. "userIds",
+        t. "countTraces",
+        o. "sessionDuration",
+        o. "totalCost" AS "totalCost",
+        o. "inputCost" AS "inputCost",
+        o. "outputCost" AS "outputCost",
+        o. "promptTokens" AS "promptTokens",
+        o. "completionTokens" AS "completionTokens",
+        o. "totalTokens" AS "totalTokens",
+        (count(*) OVER ())::int AS "totalCount"
```
maxdeichmann, line 47: was this copied? From experience, this one can kill performance.

web/src/server/api/routers/batchExport.ts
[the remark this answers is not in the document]
maxdeichmann: - I think it is fine that all users are able to see the export. Scoping this to user level could be another iteration it think.
- I think users should be able to download the data they have access to via the UI as well. As far as I know, we did not make any restrictions at least for the default tracing data.

packages/shared/src/features/batchExport/types.ts
```diff
@@ -0,0 +1,73 @@
...
+    fileType: string;
+  }
+> = {
+  CSV: { label: "CSV", extension: "csv", fileType: "text/csv" },
+  JSON: { label: "JSON", extension: "json", fileType: "application/json" },
+} as const;
+
+export const BatchExportQuerySchema = z.object({
+  tableName: z.nativeEnum(BatchExportTableName),
+  filter: z.array(singleFilter).nullable(),
```
maxdeichmann, line 38: If we keep it thsi generic, how do we ensure, that we always have the right project_id filters in here?

worker/src/features/batchExport/handleBatchExportJob.ts
```diff
@@ -0,0 +1,169 @@
...
+  streamTransformations,
+} from "@langfuse/shared/src/server";
+
+import { env } from "../../env";
+import logger from "../../logger";
+
+export const handleBatchExportJob = async (
+  batchExportJob: BatchExportJobType
+) => {
+  const { projectId, batchExportId } = batchExportJob;
```
maxdeichmann, line 26: Just to confirm. The security concept is as follows:
- User provides projectId via trpc route
- Only this project id is inserted into the database and sent to our events endpoint
- projectId is reade from the db here and passed on to getSessionTableSQL which is filtering by projectId

worker/src/features/batchExport/handleBatchExportJob.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+  );
+
+  // Stream upload results to S3
+  const accessKeyId = env.S3_ACCESS_KEY_ID;
+  const secretAccessKey = env.S3_SECRET_ACCESS_KEY;
+  const bucketName = env.S3_BUCKET_NAME;
+  const endpoint = env.S3_ENDPOINT;
+  const region = env.S3_REGION;
```
maxdeichmann, line 107: Heads up: they need to be available in the worker when deploying

PR #2361, opened 2024-06-17
feat(api): add GET /score-configs/[configId] and GET /score-configs

marliessophie:


web/src/__tests__/score-configs.servertest.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    expect(getScoreConfig.status).toBe(404);
+    expect(getScoreConfig.body).toMatchObject({
+      message: "Score config not found within authorized project",
+    });
+  });
+
+  it("should GET all score configs", async () => {
+    const fetchedConfigs = await makeAPICall<{
+      data: CastedConfig[];
```
marcklingen, line 94: can we be certain if we do not verify? unsure if we should verify here but I would not cast if we do not check here

marliessophie: I chose to verify and then cast to ensure data consistency. I don't see why we should allow any other config.categories formats at the moment. We can always become less permissive in the future if it's needed.

PR #2398, opened 2024-06-21
feat(ui): add separate date range filter to tables

RichardKruemmel:


web/src/utils/date-range-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+      date_formatter: (date) =>
+        date.toLocaleTimeString("en-US", {
+          hour: "numeric",
+          minute: "numeric",
+        }),
+      minutes: 5,
+    },
+  };
+
+export const tableDateRangeAggregationSettings: DateRangeAggregationSettings<TableDateRangeAggregationOption> =
```
marcklingen, line 133: do we use the date formatter here?

web/src/components/useDateRange.tsx
```diff
@@ -0,0 +1,88 @@
...
+  type DateRangeOptions,
+  findClosestTableIntervalToDate,
+  tableDateRangeAggregationSettings,
+  isValidOption,
+  type DateRangeAggregationOption,
+  DASHBOARD_AGGREGATION_PLACEHOLDER,
+  findClosestDashboardIntervalToDate,
+} from "@/src/utils/date-range-utils";
+
+export function useDateRange(type: "dashboard" | "table", initialDate?: Date) {
```
marcklingen, line 16: add typing for output to ensure that it is consistent

web/src/components/DateRangeDropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+            value as TableDateRangeAggregationOption
+          ];
+        fromDate = addMinutes(new Date(), -setting.minutes);
+      } else {
+        setDateRangeAndOption(DEFAULT_AGGREGATION_SELECTION, undefined);
+        return;
+      }
+      setDateRangeAndOption(value, {
+        from: fromDate,
+        to: new Date(),
```
marcklingen, line 62: do we need to set a to? it feels like it would be nice to keep this optional as a filter that is only used on dashboard if user sets explicit range

web/src/utils/date-range-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+  "1 month",
+  "7 days",
+  "24 hours",
+  "3 hours",
+  "1 hour",
+  "30 minutes",
+  "5 minutes",
+] as const;
+
+export const TABLE_AGGREGATION_OPTIONS = [
```
marcklingen, line 19: can we use min instead of minutes? thereby the component can be smaller

web/src/components/table/data-table-toolbar.tsx
```diff
@@ -82,12 +92,21 @@ export function DataTableToolbar<TData, TValue>({
...
-          columnsWithCustomSelect={columnsWithCustomSelect}
-        />
+        <>
+          <PopoverFilterBuilder
+            columns={filterColumnDefinition}
+            filterState={filterState}
+            onChange={setFilterState}
+            columnsWithCustomSelect={columnsWithCustomSelect}
+          />
+          {selectedOption && setDateRangeAndOption && (
```
marcklingen, line 102: why is this within the other condition?

web/src/components/useDateRange.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [queryParams, setQueryParams] = useQueryParams({
+    select: withDefault(StringParam, "Select a date range"),
+    from: StringParam,
+    to: StringParam,
+  });
+
+  const closestInterval = initialDate
```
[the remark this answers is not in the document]
marcklingen: decision: drop custom lookback set in db, use last 1 day as default for all tables

web/src/components/useDateRange.tsx
marcklingen: + replace initialDate with option from available options to select from

web/src/server/api/routers/users.ts
```diff
@@ -25,12 +27,23 @@ export const userRouter = createTRPCRouter({
   all: protectedProjectProcedure
     .input(UserAllOptions)
     .query(async ({ input, ctx }) => {
+      if (input.filter && input.from && input.to) {
+        input.filter = input.filter.filter(
```
marcklingen, line 31: duplication is ok

web/src/utils/date-range-utils.ts
marcklingen: simplify type to only use minutes

web/src/server/api/routers/users.ts
```diff
@@ -25,12 +27,23 @@ export const userRouter = createTRPCRouter({
   all: protectedProjectProcedure
     .input(UserAllOptions)
     .query(async ({ input, ctx }) => {
+      if (input.filter && input.from && input.to) {
```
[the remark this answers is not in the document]
marcklingen: we do not need it, move eveyrthing to filter itself, no changes on trpc api

web/src/components/useDateRange.tsx
marcklingen: decision: use 2 different hooks as table is very simple

web/src/hooks/useDashboardDateRange.tsx
```diff
@@ -0,0 +1,92 @@
...
+  selectedOption: DashboardDateRangeOptions;
+  dateRange: DashboardDateRange | undefined;
+  setDateRangeAndOption: (
+    option: DashboardDateRangeOptions,
+    range?: DashboardDateRange,
+  ) => void;
+}
+
+export function useDashboardDateRange(
+  initialDate?: Date,
```
marcklingen, line 24: we do not use this anymore? can you remove the logic?

marcklingen (approved):
made some minor changes, merging now

- split dropdown into two components
- reverse the order of options
- remove `to` from tabledaterange
- move types to utils
- change query key as `select` is too generic
- add clock icon to dropdown

PR #2414, opened 2024-06-24
feat: write token costs at ingestion time

hassiebp:
Write token costs at ingestion time


web/src/server/api/services/EventProcessor.ts
```diff
@@ -372,6 +395,42 @@ export class ObservationProcessor implements EventProcessor {
...
+      totalCost?: number | null;
+    },
+    tokenCounts: { input?: number; output?: number; total?: number },
+  ): { inputCost?: number; outputCost?: number; totalCost?: number } {
+    if (!model) return {};
+
+    const inputCost =
+      userProvidedCosts.inputCost ??
+      (tokenCounts.input && model.inputPrice
+        ? tokenCounts.input * Number(model.inputPrice)
```
maxdeichmann, line 412: should we do this with decimals? WDYT?

web/src/server/api/services/EventProcessor.ts
```diff
@@ -191,6 +192,28 @@ export class ObservationProcessor implements EventProcessor {
...
+      input: newInputCount,
+      output: newOutputCount,
+      total: newTotalCount,
+    };
+
+    const calculatedCosts = this.calculateTokenCosts(
+      internalModel,
+      userProvidedTokenCosts,
+      tokenCounts,
+    );
```
maxdeichmann, line 215: I think we have the following cases here:
- user provided the cost in first observation, user provided no cost in second observation [+other way around]
- user provided no cost at all, model match in first observation, no model in the update event [+other way around]
- user provided no cost in first event (but we match the model and calculate everything) and provides cost in the second one [+ other way around]
- if no tokens and no cost and no model match, we dont do anything

maxdeichmann (approved):

PR #2465, opened 2024-06-27
Test arm build

marcklingen:


marcklingen:
tested it and it worked well locally

PR #2466, opened 2024-06-27
chore: adds backfill script for generations calculated costs

hassiebp:
Adds backfill script for generations calculated costs


scripts/backfill_observations_calculated_cost.sql
```diff
@@ -0,0 +1,161 @@
...
+                FROM models
+                WHERE (models.project_id = o.project_id OR models.project_id IS NULL)
+                  AND models.model_name = o.internal_model
+                  AND (models.start_date < o.start_time OR models.start_date IS NULL)
+                  AND o.unit = models.unit
+                ORDER BY models.project_id, models.start_date DESC NULLS LAST
+                LIMIT 1
+            ) m ON true
+            WHERE o.id > last_id
+            ORDER BY o.id
```
maxdeichmann, line 49: can we order explicitly asc?

web/scripts/observations-backfill-calculated-cost.ts
```diff
@@ -0,0 +1,201 @@
...
+const backfillCalculatedGenerationCost = async (
+  params: BackfillCalculatedGenerationCostParams,
+) => {
+  try {
+    const { batchSize, maxRowsToProcess, sleepBetweenMs } = params;
+    log("Starting backfillCalculatedGenerationCost with params", params);
+
+    // Set the statement timeout to six hours
+    log("Setting statement_timeout to six hours...");
+    await prisma.$executeRaw(Prisma.sql`SET statement_timeout = '6h';`);
```
maxdeichmann, line 20: do we need a statement length of 6h if the script below executes multiple db queries and not one only?

web/scripts/observations-backfill-calculated-cost.ts
```diff
@@ -0,0 +1,201 @@
...
+    log("Dropping temporary column...");
+    await prisma.$executeRaw`ALTER TABLE observations DROP COLUMN IF EXISTS tmp_has_calculated_cost;`;
+    log("✅ Dropped temporary column");
+
+    log("✅ Finished backfillCalculatedGenerationCost");
+  } catch (err) {
+    console.error("Error executing script", err);
+  } finally {
+    // Reset the statement timeout to two minutes
+    await prisma.$executeRaw`SET statement_timeout = '2min';`;
```
maxdeichmann, line 176: For self hosters:
- lets store the variable somewhere and isnsert it back with raw unsave
- if it is 60 min it most likely means that we failed before and the script was triggered a second time. In this case, assume 2 min

PR #2467, opened 2024-06-27
feat: add markdown rendering for Trace page

JannisConen:
## What does this PR do?

This PR adds 1. the ability to detect Markdown in the trace page component and if so 2. renders it.
If markdown is detected in either the input or output, the pretty view becomes available.

It uses the react-markdown library to render the markdown which prevents XSS attacks and is easily extendable via plugins.

## Why
Rendering markdown enhances the capability of users to quickly assess inputs and output of traces.

To consider:
- decide whether to apply styling to markdown rendered content
- there **was** a problem with detecting and rendering markdown for the assistant message, if there was ChatML provided. Rn it just renders the JSONView in this case. -> I removed the JSON.stringify function applied to outputClean. Am not 100% sure if it breaks something, tested locally and from how I understand the JSONViewer it should still work.
- extend markdown capabilities by using plugins

## Type of change

## Checklist

- I haven't checked if new and existing unit tests pass locally with my changes (could not execute, yet but tested the added ones)


marliessophie:
Merged into [md-view branch](https://github.com/langfuse/langfuse/pull/2584), change will be finalised there and then merged into langfuse main. Thanks for the contribution @JannisConen!

PR #2476, opened 2024-06-27
fix(ci): platform arg optional in dockerfiles

marcklingen:
goal: unbreak deployment via porter


.github/workflows/porter_stack_web_staging.yml
```diff
@@ -1,7 +1,7 @@
 "on":
   push:
     branches:
-      - main
+      - target-and-build-platform
```
marcklingen, line 4:
```suggestion
      - main
```

PR #2484, opened 2024-06-28
feat(api): datasets v2 api & new typed public api interface for improved tests

marcklingen:


web/src/features/public-api/server/createAuthedAPIRoute.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    if (routeConfig.responseSchema) {
+      try {
+        // If the response schema is an object, we need to call strict() to ensure that the response object doesn't have any extra keys
+        const responseSchema = routeConfig.responseSchema;
+        if (responseSchema instanceof ZodObject) {
+          responseSchema.strict().parse(response);
+        } else {
+          responseSchema.parse(response);
```
maxdeichmann, line 84: do we also need strict in case of arrays containing objects or are they handled above already?

web/src/features/public-api/server/createAuthedAPIRoute.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+      } catch (error: unknown) {
+        console.error("Response validation failed:", error);
+        Sentry.captureException(error);
+        if (process.env.NODE_ENV !== "production")
+          throw new ApiError("DEV: output validation failed"); // rethrow in when not in production
```
maxdeichmann, line 90: why do we throw here?

web/src/pages/api/public/dataset-items.ts
```diff
@@ -1,123 +1,145 @@
...
+        throw new LangfuseNotFoundError("Dataset not found");
       }
-      const id = itemBody.id ?? uuidv4();
+
+      const itemId = id ?? uuidv4();
 
       const item = await prisma.datasetItem.upsert({
         where: {
-          id,
+          id: itemId,
```
maxdeichmann, line 44: is there a project_id condition missing here?

web/src/features/public-api/server/createAuthedAPIRoute.ts
marcklingen: arrays containing objects are objects -> strict will be applied.

there is no .strict() on refined zod types. in these cases we need to include strict in the definition

Example
https://github.com/langfuse/langfuse/blob/becc64cab99480b00a09a79e8abde5736e9883b1/web/src/features/public-api/types/datasets.ts#L129-L144

web/src/features/public-api/server/createAuthedAPIRoute.ts
marcklingen: To produce a 500 error in development -> tests should fail

maxdeichmann: Ok, this is divergence of behavior prod <> test. We should communicate this with the team as this can cause confusion.
Alternative: we do this in test code and write a helper function there to assert bodies with strict.

PR #2500, opened 2024-07-01
perf: read calculated cost from observations table directly if available

hassiebp:
Read calculated cost from observations table directly if available


web/src/features/dashboard/components/UserChart.tsx
```diff
@@ -34,7 +36,9 @@ export const UserChart = ({
   const user = api.dashboard.chart.useQuery(
     {
       projectId,
-      from: "traces_observationsview",
+      from: env.NEXT_PUBLIC_HAS_BACKFILLED_OBSERVATION_COST
```
maxdeichmann, line 39: Can you add it to the env.production file?
Maybe we should also rename this to NEXT_PUBLIC_LANGFUSE_3_MIGRATION_DONE or similar

marcklingen: fyi, we should try to not use NEXT_PUBLIC envs as they get flattened at build-time, thus, they are always null or static for our prebuilt docker images. Can we do this via a regular env that gets used server-side or NEXT_PUBLIC_CLOUD_ENV which only gets used in our cloud deployment and is passed at build-time in or cloud build pipeline?

packages/shared/prisma/migrations/20240701103900_observations_view_read_from_calculated/migration.sql
```diff
@@ -0,0 +1,88 @@
...
+    m.start_date AS "model_start_date",
+    m.input_price,
+    m.output_price,
+    m.total_price,
+    m.tokenizer_config AS "tokenizer_config",
+    CASE 
+        WHEN o.calculated_input_cost IS NULL AND o.input_cost IS NULL AND o.output_cost IS NULL AND o.total_cost IS NULL THEN
+            o.prompt_tokens::decimal * m.input_price
+        ELSE
+            COALESCE(o.calculated_input_cost, o.input_cost)
```
maxdeichmann, line 41: Dont we want to return `o.input_cost` if it is not null before returning `o.calculated_input_cost`?

maxdeichmann: Same for the output_cost.

web/src/features/dashboard/components/UserChart.tsx
maxdeichmann: As discussed, lets take NEXT_PUBLIC_LANGFUSE_CLOUD_REGION here.

PR #2506, opened 2024-07-01
perf(ui): load metdata individually on trace table

marcklingen:
In cases of large metadata the traces.all routes did not work well on serverless deployments due to limits of api response sizes. This moves to fetching individually, as already done for input and output of a trace.

### DIFF
--- a/web/src/server/api/routers/traces.ts
+++ b/web/src/server/api/routers/traces.ts
@@ -107,7 +107,7 @@ export const traceRouter = createTRPCRouter({
         async () =>
           await ctx.prisma.$queryRaw<
             Array<
-              Omit<Trace, "input" | "output" | "metdata"> & {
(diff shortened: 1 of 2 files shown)


web/src/server/api/routers/traces.ts
```diff
@@ -31,7 +31,6 @@ const TraceFilterOptions = z.object({
   searchQuery: z.string().nullable(),
   filter: z.array(singleFilter).nullable(),
   orderBy: orderBy,
-  returnIO: z.boolean().default(true),
```
marcklingen, line 34: @maxdeichmann was not used at all, ok to remove>

marcklingen, line 34: ok to remove as not used for anything else?

hassiebp (approved):

PR #2541, opened 2024-07-04
fix(build): arm builds

### LINKED ISSUE #2556: bug: can not start docker container on arm64 ubuntu
### Describe the bug

```bash
docker run -it --rm langfuse/langfuse:2 
exec /usr/local/bin/docker-entrypoint.sh: exec format error

docker run -it --rm langfuse/langfuse:2 sh
exec /usr/local/bin/docker-entrypoint.sh: exec format error

docker run -it --rm --entrypoint /bin/sh langfuse/langfuse:main -c "ls"
exec /bin/sh: exec format error
```

> The "exec format error" typically indicates that the executable you are trying to run is not compatible with the architecture of the system you are running the container on. This can happen for a few reasons:
> Architecture Mismatch: The image you are trying to run may be built for a different CPU architecture than the one your host machine uses. For example, if you are running on an ARM-based system (like a Raspberry Pi) but the image is built for x86_64, you might see this error.
> Corrupted or Misconfigured Image: The image might be corrupted or incorrectly configured, leading to the entrypoint not being executable.

### To reproduce

OS: ubuntu arm64
```
uname -a
Linux ubuntu 5.15.0-41-generic #44-Ubuntu SMP Thu Jun 23 11:20:13 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

docker container langfuse/langfuse:2

### Additional information

_No response_

No

bc291:
## What does this PR do?

Currently, arm builds, despite being marked as `linux/arm64` ones:

```bash
$ docker image inspect ghcr.io/langfuse/langfuse:2.59.1 --format '{{ .Os }}/{{ .Architecture }}'
> linux/arm64
```

are not build in `linux/arm64` context:

```bash
$ docker run -it --platform linux/arm64 ghcr.io/langfuse/langfuse:2.59.1 arch
> x86_64
```

## Cause && solution

both [web](https://github.com/langfuse/langfuse/blob/main/web/Dockerfile) and [worker](https://github.com/langfuse/langfuse/blob/main/worker/Dockerfile) dockerfiles utilize `BUILDPLATFORM` build argument.

So with current:
```
OS/Arch: linux/amd64
```
runners, only `linux/amd64` is being built.

`TARGETPLATFORM` should be used instead, which is fed based on `--platform` during buildx build.

## Type of change

Bug fix (non-breaking change which fixes an issue)


pmaciejp (approved):

marcklingen:
Do the current arm builds not work for you? When merging the contributing I verified that I was able to run them and read about build/run-platform here: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

bc291:
> (...) read about build/run-platform here: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

As blog post states:
```
BUILDPLATFORM — matches the current machine. (e.g. linux/amd64)
(...)
TARGETPLATFORM — The value set with --platform flag on build
```

so as long you want to: cross compile or produce platform agnostic results, it is fine to use `BUILDPLATFORM` for the sake of the performance. However, last stage needs to match the `TARGETPLATFORM`.

>  When merging the contributing I verified that I was able to run them. (...) Do the current arm builds not work for you?

It will run, but under the qemu emulation. You may verify it by issuing:

```bash
$ pmap  1
1: /usr/bin/qemu-x86_64 /bin/sh sh
0000000000200000     944K r--p  /usr/bin/qemu-x86_64
00000000002a0000    1672K r-xp  /usr/bin/qemu-x86_64
00000000004a0000     232K rw-p  /usr/bin/qemu-x86_64
00000000004f0000      84K rw-p  /usr/bin/qemu-x86_64
```

rennokki:
I just got back after a while away figuring out that the PR actually got merged. Last days, I wanted to test on an ARM instance in Hetzner, just to see it failing again. I tried to fix it this way, and wanted to create a PR right now, but this PR LGTM.

Yes, the build is broken and I got false negatives when I tried to build and test on an M1 machine. That is because the aarch64 works like charm, but when running on an arm64, it fails.

rennokki:
@bc291 Don't sweat it, the builds are expected to fail until @marcklingen looks at it. 😁

rennokki:
@bc291 sync?

marcklingen (approved):

marcklingen:
thanks for the additional context @bc291 & @rennokki, and the initial contribution of this @rennokki!

marcklingen:
@rennokki, we celebrated too early, this is insanely slow

CleanShot 2024-07-07 at 17 28 28
pipeline: https://github.com/langfuse/langfuse/actions/runs/9828226297/job/27131985936

PR #2562, opened 2024-07-08
fix: correctly sample evals

maxdeichmann:


worker/src/__tests__/eval-service.test.ts
```diff
@@ -199,6 +199,104 @@ describe("create eval jobs", () => {
...
+      data: {
+        id: randomUUID(),
+        projectId: "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a",
+        filter: JSON.parse("[]"),
+        jobType: "EVAL",
+        delay: 0,
+        sampling: new Decimal("0"),
+        targetObject: "traces",
+        scoreName: "score",
+        variableMapping: JSON.parse("[]"),
```
marliessophie, line 280: Nit: pass status: `ACTIVE` for clarity

worker/src/__tests__/eval-service.test.ts
```diff
@@ -199,6 +199,104 @@ describe("create eval jobs", () => {
...
+    const jobs = await kyselyPrisma.$kysely
+      .selectFrom("job_executions")
+      .selectAll()
+      .where("project_id", "=", "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a")
+      .execute();
+
+    expect(jobs.length).toBe(0);
+  }, 10_000);
+
+  test("does not create eval job for 0 sample rate", async () => {
```
marliessophie, line 245: Nit: Shall we add another test for a sample rate of say 50%?

marliessophie (approved):
Change makes sense to me, thank you! Note, there seems to be one failing test though.

worker/src/__tests__/eval-service.test.ts
maxdeichmann: also thought about this but hard to assert. I would not do it now.

PR #2575, opened 2024-07-09
fix: remove betterstack

maxdeichmann:


4t8dd:
what is the replacement then?

marcklingen:
@4t8dd, if you run Langfuse self-hosted, you can rely on any logging provider that works well with your self-hosted instances, such as the native tooling provided by AWS, Azure, or GCP.

We previously used Betterstack for Langfuse Cloud, but now we rely on the regular logs from the instances as well.

PR #2602, opened 2024-07-10
fix(score_configs): migrate to set categories NULL for numeric scores

marliessophie:


packages/shared/prisma/migrations/20240710114043_score_configs_drop_empty_categories_array_for_numeric_scores/migration.sql
```diff
@@ -0,0 +1,6 @@
+-- Migration script to update score_configs entries
+-- Set categories to NULL where data_type is 'NUMERIC' and categories is an empty array
+
+UPDATE score_configs
+SET categories = NULL
+WHERE data_type = 'NUMERIC' AND categories = '[]'::jsonb;
```
marcklingen, line 6:
```suggestion
WHERE data_type = 'NUMERIC' AND categories IS NOT NULL;
```

Is there a case where categories is not null while data_type is numeric? If not, this could be a bit broader

marcklingen (commented):
looks good, see one comment

packages/shared/prisma/migrations/20240710114043_score_configs_drop_empty_categories_array_for_numeric_scores/migration.sql
marliessophie: I actually had that initially and changed it, tested both, both work - no strong feelings as there are only cases where categories is an empty array 

Changed it back for simplicity ☺️

marcklingen (approved):

PR #2626, opened 2024-07-15
feat(ui): do not collapse strings in tables as rows have fixed height

marcklingen:
Previously, strings in tables were collapesed beyond 500 characters. As table height is fixed, there is no need to default to collapsed strings and this improves workflows when browsing traces or dataset runs

Idea: https://github.com/orgs/langfuse/discussions/2454


marcklingen:
@tkreuder thanks for suggesting this

PR #2631, opened 2024-07-15
perf: writes to clickhouse

hassiebp:
writes to clickhouse


worker/src/utils/shutdown.ts
```diff
@@ -0,0 +1,24 @@
+import logger from "../logger";
+
+import { evalJobCreator, evalJobExecutor } from "../queues/evalQueue";
+import { batchExportJobExecutor } from "../queues/batchExportQueue";
+import { flushIngestionQueueExecutor } from "../queues/ingestionFlushQueue";
+import { repeatQueueExecutor } from "../queues/repeatQueue";
+
+export const gracefulShutdown: NodeJS.SignalsListener = async (signal) => {
```
maxdeichmann, line 8: Please check the logic on main branch
- we need to return status 500 on the health endpoint once we got sigterm
- we need to shut down the express server as well and stop accepting connections.

worker/src/api/index.ts
```diff
@@ -116,7 +118,50 @@ router
...
+        return res.status(400).json(events.error);
+      }
+
+      const { batch, projectId } = events.data;
+
+      if (!ingestionFlushQueue) {
+        throw Error("Ingestion flush queue not available");
+      }
+
+      await new IngestionService(
```
maxdeichmann, line 148: Before this becomes the ingestion pipeline, we will need to add auth. Created a ticket.

worker/src/api/index.ts
```diff
@@ -116,7 +118,50 @@ router
...
-router.use("/emojis", emojis);
+router
+  .use(
+    basicAuth({
+      users: { admin: env.LANGFUSE_WORKER_PASSWORD },
+    })
+  )
+  .post("/ingestion", async (req, res) => {
+    try {
+      if (!redis) {
```
maxdeichmann, line 129: Id wrap this in the retry function from the current ingestion pipeline to become more fault tolerant

worker/src/services/IngestionService/index.ts
```diff
@@ -0,0 +1,788 @@
...
+
+          return null;
+        }
+
+        return parsed.data;
+      })
+      .filter(Boolean) as IngestionEventWithProjectIdType[];
+
+    if (eventList.length === 0) {
+      throw new Error(
```
maxdeichmann, line 129: We should not throw here. If we have one creation and one update event, the second one could read an empty queue. We should log this and handle gracefully.

web/src/env.mjs
```diff
@@ -110,6 +110,12 @@ export const env = createEnv({
     TURNSTILE_SECRET_KEY: z.string().optional(),
     // DB event log
     ENABLE_EVENT_LOG: z.enum(["true", "false"]).optional().default("true"),
+    // clickhouse
+    CLICKHOUSE_URL: z.string().optional(),
+    CLICKHOUSE_USER: z.string().optional(),
+    CLICKHOUSE_PASSWORD: z.string().optional(),
+    SERVE_FROM_CLICKHOUSE: z.enum(["true", "false"]).optional().default("true"),
```
maxdeichmann, line 117: this should be default false

web/src/server/api/services/EventProcessor.ts
```diff
@@ -179,6 +104,7 @@ export class ObservationProcessor implements EventProcessor {
           })
         : undefined;
 
+    // TODO: this is not yet implemented in the V3 ingestion pipeline. Necessary or not?
```
maxdeichmann, line 107: No, we do not have to. This is logic pre-ingestion-pipeline era.

worker/src/services/IngestionService/index.ts
```diff
@@ -0,0 +1,788 @@
...
+    entityId: string;
+  }): string {
+    return `ingestionBuffer_${params.projectId}_${params.eventType}_${params.entityId}`;
+  }
+
+  private parseProjectEntityKey(projectEntityKey: string) {
+    const split = projectEntityKey.split("_");
+
+    if (split.length !== 4) {
+      throw new Error(
```
maxdeichmann, line 191: entitykey can be set by our users. ProjectId and eventType cannot. Hence i would only check for the first three "_". Everything after that is user generated.

worker/src/services/IngestionService/index.ts
```diff
@@ -0,0 +1,788 @@
...
+
+    return traceEventList.map((trace) => {
+      const traceRecord: TraceRecordInsertType = {
+        id: entityId,
+        // in the default implementation, we set timestamps server side if not provided.
+        // we need to insert timestamps here and change the SDKs to send timestamps client side.
+        timestamp: this.getMicrosecondTimestamp(trace.body.timestamp),
+        name: trace.body.name,
+        user_id: trace.body.userId,
+        metadata: trace.body.metadata
```
maxdeichmann, line 663: we need to apply the same logic as for observations below.

maxdeichmann: // metadata needs to be converted to a record<string, string>.
      // prefix all keys with "metadata." if they are an array or primitive

packages/shared/clickhouse/migrations/0002_observations.up.sql
```diff
@@ -0,0 +1,50 @@
...
+    `input_usage_units` Nullable(Decimal64(12)),
+    `output_usage_units` Nullable(Decimal64(12)),
+    `total_usage_units` Nullable(Decimal64(12)),
+    `unit` Nullable(String),
+    `provided_input_cost` Nullable(Decimal64(12)),
+    `provided_output_cost` Nullable(Decimal64(12)),
+    `provided_total_cost` Nullable(Decimal64(12)),
+    `input_cost` Nullable(Decimal64(12)),
+    `output_cost` Nullable(Decimal64(12)),
+    `total_cost` Nullable(Decimal64(12)),
```
maxdeichmann, line 31: @hassiebp i like the naming here better than currently in the postgres db. Would you keep it this way or revert back to the current pg setup?
Disadvantage: we would need to translate everywhere between the setups.

web/package.json
```diff
@@ -60,10 +60,9 @@
     "@remixicon/react": "^4.2.0",
     "@repo/eslint-config": "workspace:*",
     "@repo/typescript-config": "workspace:*",
-    "@sentry/nextjs": "^8.13.0",
```
maxdeichmann, line 63: why these changes?

maxdeichmann (commented):
we should let everyone from the team know about goose and add that to the contributing file so that pnpm run dx still works

packages/shared/src/server/definitions.ts
```diff
@@ -0,0 +1,167 @@
...
+  observation_id: z.string().nullish(),
+  name: z.string().nullish(),
+  value: z.union([z.number(), z.string()]).nullish(),
+  source: z.string(),
+  comment: z.string().nullish(),
+  authorUserId: z.string().nullish(),
+  configId: z.string().nullish(),
+  dataType: z.enum(["NUMERIC", "CATEGORICAL", "BOOLEAN"]).nullish(),
+  stringValue: z.string().nullish(),
+});
```
maxdeichmann, line 115: the keys above need to be snake case or not?

web/src/server/api/repositories/clickhouse.ts
```diff
@@ -0,0 +1,183 @@
...
+      model: record.provided_model_name ?? null,
+      input:
+        (record.input ? parseJsonPrioritised(record.input) : undefined) ?? null,
+      output:
+        (record.output ? parseJsonPrioritised(record.output) : undefined) ??
+        null,
+      unit: record.unit ?? null,
+
+      createdAt: new Date(record.created_at),
+      updatedAt: new Date(record.created_at),
```
maxdeichmann, line 111:
```suggestion
      updatedAt: new Date(record.updated_at),
```

web/src/server/api/repositories/clickhouse.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export const convertTraces = (traces: unknown[]): Trace[] => {
+  const parsedRecord = z.array(traceRecordReadSchema).parse(traces);
+
+  return parsedRecord.map((record) => {
+    const convertedTrace: Trace = {
+      id: record.id,
+      timestamp: new Date(record.timestamp),
+      createdAt: new Date(record.created_at),
+      updatedAt: new Date(record.created_at),
```
maxdeichmann, line 162:
```suggestion
      updatedAt: new Date(record.updated_at),
```

worker/src/app.ts
```diff
@@ -41,30 +45,18 @@ logger.info(
...
+flushIngestionQueueExecutor?.on("failed", logQueueWorkerError);
 
-repeatQueueExecutor?.on("failed", (job, err) => {
-  logger.error(
-    err,
-    `Repeat Queue Job with id ${job?.id} failed with error ${err}`
-  );
-});
+process.on("SIGINT", () => onShutdown("SIGINT"));
+process.on("SIGTERM", () => onShutdown("SIGTERM"));
```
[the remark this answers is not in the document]
maxdeichmann: Sorry, you moved this entirely. Just saw it. All good.

worker/src/services/IngestionService/index.ts
maxdeichmann: Yes, you are right.

PR #2741, opened 2024-07-25
feat: adds server-side caching to prompts

hassiebp:
Adds server-side caching to prompts


packages/shared/src/env.ts
```diff
@@ -23,6 +23,8 @@ const EnvSchema = z.object({
       "ENCRYPTION_KEY must be 256 bits, 64 string characters in hex format, generate via: openssl rand -hex 32"
     )
     .optional(),
+  LANGFUSE_PROMPT_CACHE_ENABLED: z.enum(["true", "false"]).default("true"),
```
maxdeichmann, line 26: id default to false here to make life easier for self hosters. We only enable this on AWS.

web/src/features/prompts/server/actions/createPrompt.ts
```diff
@@ -82,7 +83,16 @@ export const createPrompt = async ({
       })),
     );
 
+  // Lock and invalidate cache for _all_ versions and labels of the prompt name
+  const promptService = new PromptService(prisma, redis);
+  await promptService.lockCache({ projectId, promptName: name });
```
maxdeichmann, line 88: Why did you decide to go with the locking strategy?
- if not locking, we would be serving wrong prompts from cache until ttl expires
- if locking, we could fail here and keep the lock forever

packages/shared/src/server/services/PromptService.ts
```diff
@@ -0,0 +1,207 @@
...
+      this.logError("Error deleting keys for prefix", cacheKeyPrefix, e);
+
+      throw e;
+    }
+  }
+
+  private getCacheKey(params: PromptParams): string {
+    const prefix = this.getCacheKeyPrefix(params);
+
+    return `${prefix}:${params.version ?? params.label}`;
```
maxdeichmann, line 168: For my understanding: we will end up with the same prompt multiple times if it was read with version or one of its labels, right?

web/src/features/prompts/server/actions/getPromptByName.ts
```diff
@@ -17,90 +14,31 @@ export const getPromptByName = async (
   params: GetPromptByNameParams,
 ): Promise<Prompt | null> => {
   const { promptName, projectId, version, label } = params;
+  const promptService = new PromptService(prisma, redis);
```
maxdeichmann, line 17: This here is the only function where we get prompts, right?

PR #2772, opened 2024-07-29
feat(data-table): virtual detail column interface to add individual score columns

marliessophie:


web/src/server/api/routers/scores.ts
```diff
@@ -293,6 +293,32 @@ export const scoresRouter = createTRPCRouter({
         },
       });
     }),
+  getNamesList: protectedProjectProcedure
```
marcklingen, line 296: if this is not too slow for large projects, it is a good idea to reuse this function across tables and have no time filter on it

web/src/components/table/utils/scoreDetailColumnHelpers.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const { key, header } = parseColFct(col);
+  return {
+    header,
+    id: key,
+    accessorKey: key,
+    enableHiding: true,
+  };
+};
+
+export function getDetailColumns(
```
marcklingen, line 80: does this need to be memo'ed?

web/src/components/table/utils/scoreDetailColumnHelpers.tsx
```diff
@@ -0,0 +1,196 @@
...
+              showColorCoding
+            />
+          )
+        );
+      },
+    };
+  });
+};
+
+export const getDetailMetricsColumns = (
```
marcklingen, line 133: why is this logic based on dataset router types?

web/src/components/table/utils/scoreDetailColumnHelpers.tsx
```diff
@@ -0,0 +1,196 @@
...
+        return acc;
+      },
+      {} as
+        | DatasetRunRowData["scores"]["numeric"]
+        | DatasetRunRowData["scores"]["qualitative"],
+    ) ?? {}
+  );
+};
+
+export const constructDetailMetricsColumns = <T extends DatasetRunRowData>(
```
marcklingen, line 157: why is this based on dataset router types?

web/src/features/column-visibility/hooks/useColumnVisibility.ts
```diff
@@ -4,12 +4,51 @@ import useLocalStorage from "@/src/components/useLocalStorage";
...
+
 function useColumnVisibility<TData>(
   localStorageKey: string,
   columns: LangfuseColumnDef<TData>[],
 ) {
   const initialVisibilityState = () => {
     const visibilityState: VisibilityState = {};
+    const storedVisibilityState = readStoredVisibilityState(localStorageKey);
+
+    // With virtual detail columns we must ensure state in local storage is valid given current project data and possibly upsert
```
marcklingen, line 30: why is there a need to monitor this?

we could we not keep localstorage as is, ignore non-existing columns on read, and write valid set of columns on change

web/src/features/datasets/components/DatasetRunItemsTable.tsx
```diff
@@ -34,6 +38,9 @@ type RowData = {
   scores: APIScore[];
   latency?: number;
   totalCost?: string;
+
+  // any number of additional detail columns for individual scores
+  [key: string]: any; // any of type APIScore[] for detail columns
```
marcklingen, line 43: is there a way around this to still get type errors on other columns?

marcklingen (commented):
wip notes

web/src/components/grouped-score-badge.tsx
```diff
@@ -43,9 +59,13 @@ const ScoresOfGroup = (props: {
 export const GroupedScoreBadges = ({
```
marcklingen, line 59: do we still need this when moving to single-score aggregates that are not grouped together?

web/src/components/grouped-score-badge.tsx
```diff
@@ -150,9 +177,11 @@ export const QualitativeScoreBadge = ({
           {distribution.length > DETAIL_HOVER_AFTER ? (
```
marcklingen, line 177: i'd try to use css for ellipsis of text

web/src/components/table/data-table-column-visibility-filter.tsx
```diff
@@ -10,22 +10,28 @@ import {
...
+  DropdownMenuLabel,
 } from "@/src/components/ui/dropdown-menu";
 import { type VisibilityState } from "@tanstack/react-table";
 import { ChevronDown, Columns } from "lucide-react";
 import { type LangfuseColumnDef } from "@/src/components/table/types";
 import { usePostHogClientCapture } from "@/src/features/posthog-analytics/usePostHogClientCapture";
 
 interface DataTableColumnVisibilityFilterProps<TData, TValue> {
-  columns: LangfuseColumnDef<TData, TValue>[];
+  baseColumns: LangfuseColumnDef<TData, TValue>[];
```
marcklingen, line 22: is there a native way to do this with tanstack table, e.g. grouping columns?

https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers

marliessophie: Yes, we could use groupings (also shown in table with additional group header) in the future, and infer groups of columns based on this 👍 I quite like the approach. 

However, this requires a bit of a refactor of our table component and types that I'd prefer to decouple from this functionality PR. I'd prefer to have a separate PR in the near future to add grouping support to our table component, wdyt?

web/src/features/datasets/components/DatasetRunItemsTable.tsx
marliessophie: Unfortunately not really. Typescript does not support regex on strings, which we would need here to type the string more explicitly and inform typescript that there can not be any overlap between the dynamic keys and the predefined ones. I would have preferred to avoid this implicit technical complexity, but could not find a robust way to achieve this

PR #2824, opened 2024-08-05
feat(api): add `lastConfig` attribute and `fromUpdatedAt` and `toUpdatedAt` query params to `GET /v2/prompts`

marcklingen:
based on discussion: #2684

### DIFF
--- a/fern/apis/server/definition/prompts.yml
+++ b/fern/apis/server/definition/prompts.yml
@@ -41,6 +41,12 @@ service:
           limit:
             type: optional<integer>
             docs: limit of items per page
+          fromUpdatedAt:
+            type: optional<datetime>
+            docs: Optional filter to only include prompt versions created/updated on or after a certain datetime (ISO 8601)
+          toUpdatedAt:
+            type: optional<datetime>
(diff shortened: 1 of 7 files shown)


marcklingen:
ready to be merged, waiting for feedback by @ashutoshsaboo in discussion thread

ashutoshsaboo:
Hey Marc, I'll get back on the PR by tomorrow.

marcklingen:
Thanks

ashutoshsaboo:
@marcklingen since we are returning lastConfig for even non-filter cases, we should probably also return for which version of the prompt was the config returned, else it can get confusing for the client to figure out the same. It might be a good idea to also return the same for the filter cases, i.e, show matching version number for which query filters were passed and config was returned.

marcklingen:
Thanks for the feedback, I think the current output is good enough as "last" refers to max(versions[]). thanks for your additional feedback in the discussion thread. will add the time-based filters as well as you suggested

ashutoshsaboo:
this is great, thanks Marc! @marcklingen

Issue #2468, opened 2024-06-27
feat(auth): add GitLab auth provider


jsirex:
> **EDIT** (@marcklingen): GitLab auth can be added to Langfuse via the GitLab auth [provider](https://next-auth.js.org/providers/gitlab) in nextauth. Let's keep this issue open to track the implementation of this provider. Happy to help you contribute this.

### Describe the bug

I tried to integration langfuse with gitlab via application oauth.
And it seemed for me that oauth process itself was ok, but at the very end step it failed with:
```
 prisma:error 
 Invalid `prisma.account.create()` invocation:
 
 {
   data: {
     provider: "custom",
     type: "oauth",
     providerAccountId: "21949",
     access_token: "AAAAAAAAAAAAAAAAAAAAA",
     token_type: "Bearer",
     expires_at: 1719504313,
     refresh_token: "RRRRRRRRRRRRRRRRRRRRRR",
     scope: "openid email profile",
     created_at: 1719497112,
     ~~~~~~~~~~
     id_token: "xxxxxxxxxxxxxxx.yyyyyyyyyyyyyyyyyyyy.zzzzzzzzzzzzzzzzzz-qqqqqqqqqqqqqqq-wwwwwwwwwwwwwwww-eeeeeee-rrrrrrrrr-tttttttttttttt",
     userId: "HHHHHHHHHHHHH",
 ?   id?: String,
 ?   expires_in?: Int | Null,
 ?   ext_expires_in?: Int | Null,
 ?   session_state?: String | Null
   }
 }
 
 Unknown argument `created_at`. Available options are marked with ?.
 [next-auth][error][adapter_error_linkAccount] 
 https://next-auth.js.org/errors#adapter_error_linkaccount 
 Invalid `prisma.account.create()` invocation:
 

```

### To reproduce

1. Create an oauth application at gitlab instance
2. Run latest langfuse with AUTH_CUSTOM_* variables as described in docs
3. Try to sign in

ENVS:
```
...
AUTH_CUSTOM_CLIENT_ID=XXXXXXXXXXXXXXXXX
AUTH_CUSTOM_CLIENT_SECRET=YYYYYYYYYYYYYYYYYYYYY
AUTH_CUSTOM_ISSUER=https://gitlab.company.com
AUTH_CUSTOM_NAME=gitlab
AUTH_CUSTOM_ALLOW_ACCOUNT_LINKING=true
....
```

### SDK and container versions

container version: 2.57.0

[...]

PR #2224, opened 2024-06-04
Feat/add keycloak

abartrimcmgx:
This pull request introduces changes primarily to the authentication system of the web application, with a focus on adding support for Keycloak.

Keycloak support:

* [`web/src/env.mjs`](diffhunk://#diff-2f0c198838dc2284f6c49d813934f356195aca6d17f3be3948402a4422ba538cR70-R73): The environment variables for Keycloak have been added, including the client ID, client secret, issuer, and a boolean for account linking. [[1]](diffhunk://#diff-2f0c198838dc2284f6c49d813934f356195aca6d17f3be3948402a4422ba538cR70-R73) [[2]](diffhunk://#diff-2f0c198838dc2284f6c49d813934f356195aca6d17f3be3948402a4422ba538cR178-R182)
* [`web/src/pages/auth/sign-in.tsx`](diffhunk://#diff-4b54b8e565677da17095f4f4031037c2d692fec31c2a09d34e02dc226601ba9dR51): Keycloak has been added as an authentication provider. This includes changes to the `PageProps` type, the `getServerSideProps` function, and the `SSOButtons` function. [[1]](diffhunk://#diff-4b54b8e565677da17095f4f4031037c2d692fec31c2a09d34e02dc226601ba9dR51) [[2]](diffhunk://#diff-4b54b8e565677da17095f4f4031037c2d692fec31c2a09d34e02dc226601ba9dR81-R84) [[3]](diffhunk://#diff-4b54b8e565677da17095f4f4031037c2d692fec31c2a09d34e02dc226601ba9dR158-R171)
* [`web/src/server/auth.ts`](diffhunk://#diff-e58e5990c970014f04ffaa1ab621f49c03a311caff05ded1408ecfb18f5208ebL187-R212): The Keycloak provider has been added to the static providers array. An additional hack has been implemented to remove certain properties from the account object to prevent Prisma from throwing an error. [[1]](diffhunk://#diff-e58e5990c970014f04ffaa1ab621f49c03a311caff05ded1408ecfb18f5208ebL187-R212) [[2]](diffhunk://#diff-e58e5990c970014f04ffaa1ab621f49c03a311caff05ded1408ecfb18f5208ebR1-R31)

Code organization:

[...]

PR #2846, opened 2024-08-07
feat(add): add keycloak support #2017 #2224 also related to #2468

mparada-suva:
## What does this PR do?

Add Keycloak support.

- `web/srv/env.mjs`: Add new environment variables to configurate Keycloak as Identity Provider (IP).
- `web/src/pages/auth/sign-in.tsx`: Adds Keycloak sign-in button to the UI.
- `web/src/server/auth.ts`: Adds the next-auth KeycloakProvider.
  - Keycloak returns per default some fields which do not match the schema of the prisma account. My solution was to add a `cleanupToken` function that would take a token (json) produced by an IP and remove unnecessary fields (and possibly add default values). I believe such a construct would allow other IPs to be easily integrated in case they also have extra fields.

Related Issues and previous PRs: #2017 #2224 #2468

## Notes

- This is my first PR! And also the first time I'm working with Typescript. There is probably a ton of things that could be done better. I'm happy to include feedback to make this better. But I'll definitely need some support 😅

## Type of change

## Mandatory Tasks

## Checklist

- My code doesn't follow the style guidelines of this project (`npm run prettier`) -> I didn't manage to install and run prettier. I tried to keep the style guide.
- My PR needs changes to the documentation -> Changes are needed probably around here: https://langfuse.com/docs/deployment/self-host#sso
- I checked if my changes generate new warnings (`npm run lint`) -> 2 warnings related to unused values. The values are actually unused (since we ignore them). Don't know how to fix this.
- I haven't added tests that prove my fix is effective or that my feature works -> Since it is an itegration, I don't know how is the right approach to test it. I also looked for similar tests for other IPs and couldn't find them.
- I checked if new and existing unit tests pass locally with my changes -> Some tests fail, but none seem to be related to my changes. As far as I can tell from the logs, the errors are related to the IngestionService.


marcklingen:
thanks for raising this, does this work for you locally end-to-end?

mparada-suva:
Hi @marcklingen! Thanks for the super quick feedback!
Yes, it works locally for me end-to-end.

mparada-suva:
I think I found which type is the one that `linkAccount` expects and added it to the (now renamed) `cleanupAccount` function. But I don't know how to solve the other lint error: unused variables. I create these variables only to deconstruct the account Object from Keycloak and remove the unnecessary extra fields. Does anyone knows a better approach? I believe one can use the `delete` operator but this one works in place, and I'm not sure I will break something somewhere else if I delete these fields. Should I simply try?
Something like:
```typescript
delete account['refresh_expires_in']
delete account['not-before-policy']
```

mparada-suva:
okay... I went ahead and did use `delete`: now all tests (apart from the Vercel ones) are passing (Goodhart's law notwithstanding). Do I need to improve the PR in some other way to get it merged?

web/src/server/auth.ts
```diff
@@ -139,6 +140,13 @@ if (env.SMTP_CONNECTION_URL && env.EMAIL_FROM_ADDRESS) {
   );
 }
 
+// Function to be used in the prismaAdapter to cleanup extra
+// fields in the Account Object of specific Providers. Default does
+// nothing.
+let cleanupAccount = (account: AdapterAccount) => {
```
marcklingen, line 146: I think adding a mutable function that is updated when keycloak is used can can have some side effects on other providers on the same langfuse instance.

Is this strictly necessary? It seems like this issue was addressed here:
- https://github.com/nextauthjs/next-auth/pull/7369
- https://github.com/nextauthjs/next-auth/issues/7655

If not, is there a cleaner way of only using this function on the KeycloakProvider?

mparada-suva: Thanks a lot for the links! I'll check them out and the code from next-auth. There probably is a better way because we surely are not the first to try to use Keycloak with next-auth!

mparada-suva: Did I understood it correctly that this change will happen only in next-auth v5? This [message](https://github.com/nextauthjs/next-auth/issues/7655#issuecomment-1567440434) seems to imply those changes were merged to main but will only be released with v5 and that until then such a "workaround" is fine (scary quotes in original). Though I suspect your concern is valid in case of a Langfuse Server being integrated with multiple IPs at the same time.

The [docs](https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5) also mention it as a breaking change when upgrading to v5 (see last bullet point in Breaking Changes).

RTae: @mparada-suva Yeah, I think so. (https://github.com/langfuse/langfuse/pull/2866) My solution is a workaround by removing the data before linking accounts, but as a contributor to this issue, https://github.com/nextauthjs/next-auth/issues/7655 mention that this issue should be solved in version v5 by letting developers manually pass chosen fields in the provider's account callback. So, if Langfuse's team has a plan to migrate this library, they may need to make an adjustment to support this change.

RTae: @mparada-suva  BTW, if you have any suggestions for my PR, feel free to add them. I feel my solution is quite bulky since each IP have own additional parameters, so we may need to add a code to delete these parameters in each IP, like this [comment](https://github.com/orgs/langfuse/discussions/2165#discussioncomment-10274924)

mparada-suva: Closing this PR in favor of #2866

mparada-suva:
PR closed in favor of #2866

PR #2922, opened 2024-08-13
feat(ui): add deletion of organizations via org settings

marcklingen:


marcklingen:
flaky test is unrelated to this change, merging to main

PR #2971, opened 2024-08-19
feat(comments): add comments on objects

marliessophie:


web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,192 @@
...
+  });
+
+  if (!object) {
+    throw new Error(
+      `No ${prismaModel} with id ${objectId} in project ${projectId}`,
+    );
+  }
+};
+
+export const commentsRouter = createTRPCRouter({
```
maxdeichmann, line 51: can you add try .. catch blocks within the routers? If you then log the message, it helps us to find errors in our logs. Thx!

marliessophie: Of course, thx 👍

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,238 @@
...
+      projectId,
+    },
+  });
+
+  if (!object) {
+    throw new Error(
+      `No ${prismaModel} with id ${objectId} in project ${projectId}`,
+    );
+  }
+};
```
marliessophie, line 47: Would love if this fct was more typesafe, but unsure if that's achievable. Open to feedback :)

packages/shared/prisma/schema.prisma
```diff
@@ -635,6 +636,27 @@ model Events {
...
+    project      Project           @relation(fields: [projectId], references: [id], onDelete: Cascade)
+    objectType   CommentObjectType @map("object_type")
+    objectId     String            @map("object_id")
+    timestamp    DateTime          @default(now())
+    content      String
+    authorUserId String?           @map("author_user_id")
+
+    @@index([projectId, objectType, objectId])
+    @@map("comments")
+}
```
marliessophie, line 651: I considered the cost/benefit of adding a uniqueness constraint on DB level and deemed it v. slightly net negative here as it requires the database to check for existing values before inserting/updating a record and does not provide much value given these are simple comments -- but ultimately I don't have strong feelings here, wdyt?

marcklingen: on what do you want to check uniqueness here? on id will happen automatically, is there another uniqueness that you'd like to enforce?

packages/shared/prisma/schema.prisma
```diff
@@ -635,6 +636,27 @@ model Events {
     @@map("events")
 }
 
+model Comment {
```
marcklingen, line 639: I'd suggest to switch to prisma managed created_at, updated_at as on most other tables. thereby you do not need to manage this in the apis and in the future updtedAt will manage edits as well (edited if updated!==created)

web/src/features/comments/CommentList.tsx
```diff
@@ -0,0 +1,241 @@
...
+    return null;
+
+  function onSubmit(values: z.infer<typeof CreateCommentData>) {
+    createCommentMutation
+      .mutateAsync({
+        ...values,
+      })
+      .catch((error) => {
+        console.error(error);
+      });
```
marcklingen, line 130: can be removed

web/src/features/comments/CommentList.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  useEffect(() => {
+    form.reset({ content: "", projectId, objectId, objectType });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [objectId, objectType]);
+
+  const utils = api.useUtils();
+
+  const createCommentMutation = api.comments.create.useMutation({
+    onError: (error) => form.setError("content", { message: error.message }),
```
marcklingen, line 100: does not work as you reset the form onSettled. Suggestion:
- move the invalidation/reset to onSuccess to make sure the draft message is not deleted if the trpc call does not work
- remove the onError thereby use the global onError trpc handler which will render a notification in the top right

web/src/features/comments/CommentList.tsx
```diff
@@ -0,0 +1,262 @@
...
+      day: "numeric",
+    });
+  }
+};
+
+export function CommentList({
+  projectId,
+  objectId,
+  objectType,
+  isVisible,
```
marcklingen, line 51: Is `isVisible` necessary? Just tried removing the prop, the tRPC route was still only executed when the drawer was open. I believe it can be removed.

web/src/features/comments/CommentCountIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import { MessageCircleMore } from "lucide-react";
+
+export function CommentCountIcon({ count }: { count?: number }) {
+  if (!count) return null;
+
+  return (
+    <span className="relative mr-1 text-xs">
+      <MessageCircleMore className="h-5 w-5 text-muted-foreground" />
+      <span className="absolute -top-0.5 left-3 flex h-4 w-4 items-center justify-center rounded-full border border-muted-foreground bg-accent-light-blue text-xs">
+        {count}
```
marcklingen, line 10: does not look good if count is too large for fixed-pixel circle. I'd rather use smaller font, fixed padding, and rounded-full -> becomes more like a pill if there are e.g. 120 comments

web/src/features/comments/CommentCountIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import { MessageCircleMore } from "lucide-react";
+
+export function CommentCountIcon({ count }: { count?: number }) {
+  if (!count) return null;
+
+  return (
+    <span className="relative mr-1 text-xs">
+      <MessageCircleMore className="h-5 w-5 text-muted-foreground" />
+      <span className="absolute -top-0.5 left-3 flex h-4 w-4 items-center justify-center rounded-full border border-muted-foreground bg-accent-light-blue text-xs">
```
marcklingen, line 9: i think muted looks off compared to the icons next to it

web/src/components/trace/TracePreview.tsx
```diff
@@ -20,15 +20,20 @@ import ScoresTable from "@/src/components/table/use-cases/scores";
 import { ScoresPreview } from "@/src/components/trace/ScoresPreview";
 import { AnnotateDrawer } from "@/src/features/scores/components/AnnotateDrawer";
 import useLocalStorage from "@/src/components/useLocalStorage";
+import { CommentDrawerButton } from "@/src/features/comments/CommentDrawerButton";
 
 export const TracePreview = ({
   trace,
   observations,
   scores,
+  showComments,
```
marcklingen, line 29: why is showcomments necessary?

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
...
+        return comments;
+      } catch (error) {
+        console.error(error);
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Fetching comments by object id failed.",
+        });
+      }
+    }),
+  getCountsByObjectIds: protectedProjectProcedure
```
marcklingen, line 195: allowing for multiple types seems like bad practice if there are collisions on ids across object types, should not be mixed but rather then done in separate trpc calls (e.g. one for trace and one for observations)

web/src/server/api/routers/comments.ts
marcklingen: mh, this is creative. I'd probably just skip it and not validate the reference object, but now that it exists I'd keep it as long as Max does not have a strong opinion here

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
...
+      } catch (error) {
+        console.error(error);
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Creating comment failed.",
+        });
+      }
+    }),
+  delete: protectedProjectProcedure
+    .input(DeleteCommentData)
```
marcklingen, line 90: id -> commentId?

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
...
+        const comment = await ctx.prisma.comment.findFirst({
+          where: {
+            id: input.id,
+            projectId: input.projectId,
+            objectId: input.objectId,
+            objectType: input.objectType,
+          },
+        });
+        if (!comment) {
+          throw new Error("No comment with this id in this project.");
```
marcklingen, line 108: throw trpc errors

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
...
+        });
+
+        await auditLog({
+          session: ctx.session,
+          resourceType: "comment",
+          resourceId: comment.id,
+          action: "delete",
+          before: comment,
+        });
+      } catch (error) {
```
marcklingen, line 133: do we need try catch block here? hinders us from throwing more specific error messages as they will be catched here

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
...
+        console.error(error);
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Deleting comment failed.",
+        });
+      }
+    }),
+  getByObjectId: protectedProjectProcedure
```
marcklingen, line 141: looks good to me! nicely done with org check

web/src/server/api/routers/comments.ts
```diff
@@ -0,0 +1,242 @@
+import { z } from "zod";
+
+import { throwIfNoProjectAccess } from "@/src/features/rbac/utils/checkProjectAccess";
+import {
+  createTRPCRouter,
+  protectedProjectProcedure,
+} from "@/src/server/api/trpc";
+import { CommentObjectType } from "../../../../../packages/shared/dist/prisma/generated/types";
```
marcklingen, line 8: fix the import

marcklingen (changes requested):
this is awesome! some notes her + made some ui/ux tweaks to the PR directly

web/src/server/api/routers/comments.ts
marliessophie: With the try ... catch blocks we can ensure we log the errors. I have now slightly adjusted the code to then throw the caught error if it's a TRPC error and only default to internal service error if the error is of an unknown type -- we'd still get the specific error messages in the logs, but have the guarantee that we'd always be throwing TRPC errors. Wdyt?

web/src/components/trace/TracePreview.tsx
marliessophie: Because we don't show comments on the trace timeline view and this currently controls that

web/src/server/api/routers/comments.ts
marliessophie: Max's happy with the function 👍

web/src/server/api/routers/comments.ts
marcklingen: ok, will share try/catch to catch non-trpcerrors in trpc routes

web/src/components/trace/TracePreview.tsx
marcklingen: ok, lets remove

PR #2984, opened 2024-08-20
feat(ui): support OpenAI's image content part format via markdown on trace/observation preview 

marliessophie:


web/src/server/api/routers/utilities.ts
```diff
@@ -82,8 +82,8 @@ const isValidImageUrl = async (url: string): Promise<boolean> => {
     const contentType = response.headers.get("content-type");
     const contentLength = response.headers.get("content-length");
 
-    // ensure image is not larger than 1MB in size
-    if (!contentLength || parseInt(contentLength, 10) > 1 * 1024 * 1024) {
+    // ensure image is not larger than 1.25MB in size
```
marliessophie, line 85: slightly increased image size to provide some flexibility for edge cases where images are slightly over 1MB

Issue #3017, opened 2024-08-22
bug: Javascript error when using AUTH_GOOGLE_ALLOWED_DOMAINS


LucasMagnum:
### Describe the bug

When trying to authenticate with a user with an empty `hd` it returns a javascript error instead of returning a more user friendly error.

<img width="1020" alt="Screenshot 2024-08-22 at 14 53 03" src="https://github.com/user-attachments/assets/f9292fdb-0991-417b-b637-dd31fc99de03">

### To reproduce

Start the application with Google provider and set the `AUTH_GOOGLE_ALLOWED_DOMAINS` to an invalid domain (or a domain that is different that your actual domain).

AUTH_GOOGLE_CLIENT_ID=<...>
AUTH_GOOGLE_SECRET_ID=<...>
AUTH_GOOGLE_ALLOWED_DOMAINS=example.com

Open the application, try to authenticate with a `@gmail.com` user (in some cases the `hd` will be empty -> https://developers.google.com/identity/openid-connect/openid-connect#an-id-tokens-payload) and you see the error in the screenshot. 

### SDK and container versions

Latest version

### Are you interested to contribute a fix for this bug?

Yes

marcklingen:
thanks for reporting and fixing this! will be part of the next release later today

PR #3019, opened 2024-08-22
fix(auth): Cannot read properties of undefined, GoogleProfile.hd

LucasMagnum:
Fix: cannot read properties of undefined (reading 'toLowerCase') whe accessing GoogleProfile.hd.

As mentioned by the documentation (https://developers.google.com/identity/openid-connect/openid-connect#an-id-tokens-payload), GoogleProfile.hd can be null, hence, I have changed the code to check that before trying to run `toLowerCase()` 

## What does this PR do?
Fixes # ([issue](https://github.com/langfuse/langfuse/issues/3017))

Screenshot 2024-08-22 at 17 29 08

## Type of change

## Mandatory Tasks


marcklingen (approved):
thanks for fixing this!

PR #3022, opened 2024-08-22
feat: add tag filter to generations table

RichardKruemmel:


web/src/server/api/routers/generations/filterOptionsQuery.ts
```diff
@@ -109,6 +109,26 @@ export const filterOptionsQuery = protectedProjectProcedure
...
 
+    const tags = await ctx.prisma.$queryRaw<
+      Array<{
+        tag: string | null;
+        count: number;
+      }>
+    >(Prisma.sql`
+        SELECT
+          tag,
+          count(*)::int AS count
```
marliessophie, line 120: Can we please drop the count here to reduce latency? The counts for the filter options are quite expensive unfortunately and would be helpful to not include them for now. We did some investigation on datadog yesterday and this was one of the findings ☺️ 

Also, if you like you may include a free text option to make the filter option more flexible given we're limiting it ☺️

RichardKruemmel: Thanks for pointing out 👍  I will drop it.

web/src/server/api/routers/generations/filterOptionsQuery.ts
```diff
@@ -109,6 +109,24 @@ export const filterOptionsQuery = protectedProjectProcedure
       `),
     ]);
 
+    const tags = await ctx.prisma.$queryRaw<
```
marcklingen, line 112: please include in parallelized queries above in order not to add latency

marcklingen (changes requested):

marcklingen (approved):

PR #3055, opened 2024-08-23
feat: Supports filter string using regular expression matching

chai3:
## What does this PR do?

closes #XXX(not yet issues)
https://github.com/orgs/langfuse/discussions/3014
- I implemented only that regular expression matching is true and case-insensitively.

image

## Type of change


chai3:
invalid regex got

image

PR #3067, opened 2024-08-26
feat: add tags filter to sessions table

RichardKruemmel:


web/src/server/api/routers/sessions.ts
```diff
@@ -178,8 +178,27 @@ export const sessionRouter = createTRPCRouter({
           LIMIT 1000;
         `);
 
+        const tags = await ctx.prisma.$queryRaw<
```
marcklingen, line 181: please parallelize and add time filter similar to trace filterOptions query

marcklingen (changes requested):

packages/shared/src/queries/createSessionsAllQuery.ts
```diff
@@ -50,9 +50,12 @@ export const createSessionsAllQuery = (
             MIN(t. "timestamp") AS "min_timestamp",
             array_agg(t.id) AS "traceIds",
             array_agg(DISTINCT t.user_id) AS "userIds",
-            count(t.id)::int AS "countTraces"
+            count(t.id)::int AS "countTraces",
+            array_agg(DISTINCT u.tag) AS "tags"
```
marcklingen, line 54: can this work? please test
array_agg(DISTINCT unnest(t.tags)) AS "tags"

marcklingen (approved):

PR #3071, opened 2024-08-26
feat: add tags filter to scores table

RichardKruemmel:


web/src/server/api/routers/scores.ts
```diff
@@ -177,8 +177,18 @@ export const scoresRouter = createTRPCRouter({
         },
       });
 
+      const tags = await ctx.prisma.$queryRaw<{ value: string }[]>`
```
marcklingen, line 180: please parallelize with name query + add timestamp filter to imrpove performance. have a look at trace router for example

marcklingen (changes requested):

marcklingen (approved):

PR #3076, opened 2024-08-26
chore(tables): persist date range in session storage

marliessophie:


web/src/utils/date-range-utils.ts
```diff
@@ -118,77 +108,73 @@ export const dashboardDateRangeAggregationSettings: DashboardDateRangeAggregatio
...
-      interval,
-      diff: Math.abs(duration - minutes * 60 * 1000),
-    };
-  });
-
-  diffs.sort((a, b) => a.diff - b.diff);
-
-  return diffs[0]?.interval;
-};
-
```
marliessophie, line 191: dead code, not used

PR #3221, opened 2024-09-04
feat: add structured logging option

Steffen911:
## What does this PR do?

- Adds a new environment variable `LANGFUSE_LOG_FORMAT` that supports `text` and `json` logging. The default is `text` based logging. 
- Moves large parts of console.log statements to the new logger. To keep the size manageable, I'll create a second PR to change the remaining invocations.

**This changes the existing log format in many cases. If we consider logging an API, this would be a breaking change.**

Example `text`:
```
2024-09-04T06:56:14.691Z info       Received SIGINT, closing server...
```

Example `json`:
```
{"level":"info","message":"Received SIGINT, closing server...","timestamp":"2024-09-04T06:54:44.808Z"}
```

## Type of change

## Deploy notes

- [ ] Remove `LANGFUSE_WORKER_BETTERSTACK_TOKEN` from environment.
- [x] Revert GitHub Action change.


packages/shared/src/env.ts
```diff
@@ -39,6 +39,10 @@ const EnvSchema = z.object({
     .positive()
     .default(60 * 10),
   SALT: z.string().optional(), // used by components imported by web package
+  LANGFUSE_LOG_LEVEL: z
+    .enum(["trace", "debug", "info", "warn", "error", "fatal"])
+    .optional(),
+  LANGFUSE_LOG_FORMAT: z.enum(["text", "json"]).default("text"),
```
Steffen911, line 45: How do these environment variables work? `LANGFUSE_LOG_LEVEL` is also/still defined on the worker. Do shared environments bubble up, i.e. shall I remove the reference in the worker env.ts?

packages/shared/src/db.ts
```diff
@@ -11,17 +11,38 @@ import {
...
 // Instantiated according to the Prisma documentation
 // https://www.prisma.io/docs/orm/more/help-and-troubleshooting/help-articles/nextjs-prisma-client-dev-practices
 
 const prismaClientSingleton = () => {
-  return new PrismaClient({
-    log:
-      env.NODE_ENV === "development"
-        ? ["query", "error", "warn"]
-        : ["error", "warn"],
+  const client = new PrismaClient<
```
Steffen911, line 20: We could also use a child logger here. That would allow us to add an attribute "prisma" to the structured output instead of using a prefix, which would simplify filtering for prisma logs within our monitoring tools.

packages/shared/src/server/logger.ts
```diff
@@ -0,0 +1,33 @@
+import { env } from "../env";
+import winston from "winston";
+
+const getWinstonLogger = (
+  nodeEnv: "development" | "production" | "test",
```
Steffen911, line 5: @maxdeichmann We included the nodeEnv in the getLogger method in the past to decide on the logTransport to use. As it's not necessary anymore, I'd prefer to drop it. Any concerns?

web/package.json
```diff
@@ -46,6 +46,7 @@
     "@opentelemetry/instrumentation": "^0.52.1",
     "@opentelemetry/instrumentation-http": "^0.52.1",
     "@opentelemetry/instrumentation-ioredis": "^0.42.0",
+    "@opentelemetry/instrumentation-winston": "^0.40.0",
```
Steffen911, line 49: Could this move into the shared package?

worker/src/__tests__/eval-service.test.ts
```diff
@@ -391,9 +391,6 @@ describe("create eval jobs", () => {
     expect(jobs.length).toBe(1);
     expect(jobs[0].project_id).toBe("7a88fb47-b4e2-43b8-a06c-a5ce950dc53a");
     expect(jobs[0].job_input_trace_id).toBe(traceId);
-    console.log(jobs[0]);
-    const j = await getEvalQueue()?.getJob(jobs[0].id);
-    console.log(j);
```
Steffen911, line 396: Aside from logging the result `j` was never asserted. Hence, I removed it. Happy to revert if it serves a purpose or adjust the assertions.

web/package.json
```diff
@@ -156,8 +157,8 @@
     "@types/jest": "^29.5.12",
     "@types/lodash": "^4.17.7",
     "@types/node": "20.10.5",
-    "@types/react": "^18.2.79",
-    "@types/react-dom": "^18.2.25",
+    "@types/react": "~18.2.79",
```
Steffen911, line 160: Align minor version with locked react version of 18.2.0.

packages/shared/src/env.ts
maxdeichmann: Yes, this bubbles up. Yes, removing this from worker makes sense. can you also add this to `.env.prod.example`?

packages/shared/src/server/logger.ts
maxdeichmann: No, feel free to drop.

web/package.json
maxdeichmann: Probably not as we need to instantiate the instrumentation in the instrumentation files of worker / web. We could do this, if we then expose this to web / dev.

worker/src/logger.ts
```diff
@@ -1,33 +0,0 @@
-import pino from "pino";
-import { env } from "./env";
-
-const getBetterstackLogginTransport = (minLevel: string) => {
```
maxdeichmann, line 4: When deploying this, we should remove the LANGFUSE_WORKER_BETTERSTACK_TOKEN from our environment. Can you also remove it from everywhere else in the repo?

maxdeichmann:
There are still console.logs in the project left. Any reason for this?

Steffen911:
@maxdeichmann See
> Moves large parts of console.log statements to the new logger. To keep the size manageable, I'll create a second PR to change the remaining invocations.

I wanted to avoid big merge conflicts for everyone and, therefore, eventually stopped the transition. I'll follow this up with a separate PR.

Steffen911:
Corresponding doc PR: https://github.com/langfuse/langfuse-docs/pull/789

worker/src/logger.ts
Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/3221/commits/1b2409018704dbc34d10aac1b67db2ede6b5ed77.

packages/shared/src/env.ts
Steffen911: Clarified offline with Marc. Will remove it from the worker and leave it in the shared package only. Corresponding doc adjustment is prepared in https://github.com/langfuse/langfuse-docs/pull/789.

maxdeichmann (approved):

PR #3229, opened 2024-09-04
feat: migrate shared package to new logger

Steffen911:
## What does this PR do? 

Expand the new logger functionality from https://github.com/langfuse/langfuse/pull/3221 to all shared components.


packages/shared/src/index.ts
```diff
@@ -1,13 +1,10 @@
 export * from "./constants";
-export * from "./queries";
 export * from "./interfaces/filters";
 export * from "./interfaces/orderBy";
 export * from "./interfaces/cloudConfigSchema";
 export * from "./interfaces/parseDbOrg";
 export * from "./tableDefinitions";
 export * from "./types";
-export * from "./filterToPrisma";
-export * from "./orderByToPrisma";
```
Steffen911, line 10: Those are used on the backend only and use the logger module. If we keep them at this level, some *.tsx files try to import them which breaks the compilation. Hence, I moved them into the `server` sub-module.

PR #3270, opened 2024-09-09
perf: drop large events table

maxdeichmann:


packages/shared/prisma/schema.prisma
```diff
@@ -621,21 +620,6 @@ model DatasetRunItems {
     @@map("dataset_run_items")
 }
 
-model Events {
-    id        String   @id @default(cuid())
-    createdAt DateTime @default(now()) @map("created_at")
-    updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
-    projectId String   @map("project_id")
-    project   Project  @relation(fields: [projectId], references: [id], onDelete: Cascade)
```
maxdeichmann, line 629: From my understanding, the relation has no effect. If projects are deleted it will cause events to delete. If events are deleted, it has no impact on projects.

PR #3295, opened 2024-09-10
fix(ui): populate default prompt data for chat prompts with custom role

marliessophie:
- populate default prompt data for chat prompts with custom roles 
   - note: the langfuse API/SDKs allow to ingest prompts with any string as user defined role, see [API reference for details](https://api.reference.langfuse.com/#post-/api/public/v2/prompts)  
- allow to toggle between all available roles (ie `assistant`, `user`) and any custom role given a custom role was present in the original message body

### DIFF
--- a/web/src/components/ChatMessages/ChatMessageComponent.tsx
+++ b/web/src/components/ChatMessages/ChatMessageComponent.tsx
@@ -25,7 +25,8 @@ export const ChatMessageComponent: React.FC<ChatMessageProps> = ({
   const toggleRole = () => {
     if (message.role === ChatMessageRole.System) return;
 
-    if (!!availableRoles) {
+    // if user has set custom roles, available roles will be non-empty and we toggle through custom and default roles (assistant, user)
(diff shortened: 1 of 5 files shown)


hassiebp:
Ah got it! So the user issue was that since they used custom roles for the chat prompts, we could not render them correctly as we were more strict in the UI to support only predetermined roles?

web/src/components/ChatMessages/ChatMessageComponent.tsx
```diff
@@ -9,27 +9,38 @@ import type { MessagesContext } from "./types";
...
     if (message.role === ChatMessageRole.System) return;
 
-    updateMessage(
-      message.id,
-      "role",
-      message.role === ChatMessageRole.User
-        ? ChatMessageRole.Assistant
-        : ChatMessageRole.User,
-    );
+    if (!!availableRoles) {
```
hassiebp, line 28: *suggestion*: This existence check might return true for empty array of availableRoles which might not be intended

PR #3332, opened 2024-09-12
feat(ui): export traces

marliessophie:


worker/src/features/batchExport/handleBatchExportJob.ts
```diff
@@ -16,12 +19,193 @@ import {
...
+              t."input",
+              t."output",
+              t."metadata",
+              t."tags",
+              COALESCE(tm."promptTokens", 0)::bigint AS "usage.promptTokens",
+              COALESCE(tm."completionTokens", 0)::bigint AS "usage.completionTokens",
+              COALESCE(tm."totalTokens", 0)::bigint AS "usage.totalTokens",
+              COALESCE(tm."calculatedInputCost", 0)::numeric AS "inputCost",
+              COALESCE(tm."calculatedOutputCost", 0)::numeric AS "outputCost",
+              COALESCE(tm."calculatedTotalCost", 0)::numeric AS "totalCost"
```
marliessophie, line 168: _Question_: I have ensured all possible trace table rows are included in our query here. However for the sessions query we have included a `totalCount` column which is not included in the table, may I ask why it was included in the first place or was it legacy? And would we like to keep it? Also asking as it's quite expensive to compute.

marliessophie (commented):
_Global comment_: Ideally I'd like to make the current code more typesafe in the sense that currently a dev needs to remember to add a column to the export query when adding it to the traces/sessions table. Do you have a straightforward idea to introduce tighter coupling between the front-end and server side export table row types/implementation? Not high enough priority to justify doing straight away but the less devs need to remember the better, so at least I'd create a backlog ticket - wdyt?

worker/src/features/batchExport/handleBatchExportJob.ts
```diff
@@ -16,12 +19,193 @@ import {
...
+  const createdAtCutoffFilter: FilterCondition = {
+    column: tableNameToTimeFilterColumn[tableName],
+    operator: "<",
+    value: cutoffCreatedAt,
+    type: "datetime",
+  };
+
+  switch (tableName) {
+    case "sessions":
+      // TODO: improve performance for (count(*) OVER ())::int AS "totalCount"
```
marliessophie, line 82: _See comment below_

Adding comment to resolve after removing todo before merging

clemra:
Excited about this!

packages/shared/src/server/queries/createTracesQuery.ts
```diff
@@ -0,0 +1,87 @@
...
+         ${observationTimeseriesFilter}
+  ) AS tl ON true
+  LEFT JOIN LATERAL (
+    SELECT
+      ${selectScoreValues ? Prisma.sql`jsonb_object_agg(name::text, "values") AS "scores_values",` : Prisma.empty}
+      jsonb_object_agg(name::text, avg_value::double precision) AS "scores_avg"
+    FROM (
+        SELECT
+            name,
+            ${selectScoreValues ? Prisma.sql`array_agg(COALESCE(string_value, value::text)) AS "values",` : Prisma.empty}
```
marliessophie, line 69: Note: the numeric score values are cast as string and show up in a format like `["0.8042602923476814","0.7271858775794038","-0.2272392299725476","0.9284890924655911"]` in the csv or json output example -- while this is not ideal I think it's fine for the time being and can always be improved later, wdyt?

hassiebp: Yes, agree :)

edwardzjl:
Hello there, I'm also very excited about this feature. However, after deploying `2.79.0` I did not see any export button on my traces panel. Do I need to do some configuration to make it work?

PR #3348, opened 2024-09-13
temp: log db connection params

maxdeichmann:


packages/shared/src/db.ts
```diff
@@ -28,6 +28,12 @@ const prismaClientSingleton = () => {
     ],
   });
 
+  const connectionString = env.DATABASE_URL;
+  const paramsString = connectionString?.split("?")[1];
+  console.log(
+    `database connection params: ${paramsString}, ${connectionString}`
```
maxdeichmann, line 34: remove connection

Steffen911 (approved):
Please drop connectionString as this would print db password

PR #3371, opened 2024-09-16
chore: simplify daily metrics query

Steffen911:
## What does this PR do?

On our database, we see that lots of time is spend on the daily metrics query. It's not exposed to the frontend, but called actively by a few projects. With my change, we should perform the traces and observations join only once and reuse that result more effectively to generate essentially the same result set.


web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,73 +47,47 @@ export default withMiddlewares({
...
-                  "totalUsage",
-                  'totalCost',
-                  "totalCost",
-                  'countObservations',
-                  "countObservations",
-                  'countTraces',
-                  "countTraces")) daily_usage_json
+              sum("countTraces")::integer "countTraces",
+              sum("countObservations")::integer "countObservations",
+              sum("totalCost")::DOUBLE PRECISION "totalCost",
```
Steffen911, line 69: This performs essentially the same aggregation by date as the `daily_stats` query did before. The main difference is that we use `observation.start_time` as the baseline for the date instead of `traces.timestamp`. Since we've joined on the column in the end, I believe that this actually improves the situation as we're using a "single date of truth".

web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,73 +47,47 @@ export default withMiddlewares({
...
               ${userCondition}
               ${tagsCondition}
               ${fromTimestampCondition}
               ${toTimestampCondition}
-            GROUP BY
-              1,
-              2
-            ORDER BY
-              1,
-              2
```
Steffen911, line 65: Order by should be irrelevant to the grouping in daily_model_usage (aside from a potentially different order in the aggregated json array). IMO that should be negligible as users can still match by model key. We can revert this if necessary, though.

marcklingen: agree

web/src/pages/api/public/metrics/daily.ts
marcklingen: sounds good, if this is mentioned in the api reference (fern), we need to update this there as well

web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,53 +47,43 @@ export default withMiddlewares({
...
+            GROUP BY 1
           ),
-          daily_stats AS (
+          trace_usage AS (
             SELECT
               DATE_TRUNC('DAY', t.timestamp) "date",
-              count(distinct t.id)::integer count_traces,
-              count(distinct o.id)::integer count_observations,
-              SUM(o.calculated_total_cost)::DOUBLE PRECISION total_cost
+              count(distinct t.id)::integer as "countTraces"
```
Steffen911, line 85: Unfortunately, I couldn't drop the whole query as the aggregation in the first query would double count traces if we group by model. But we can still reduce this query to avoid the join with observations which should be a net-positive.

web/src/pages/api/public/metrics/daily.ts
Steffen911: @marcklingen I don't see that fern would need an update. All statements in there are still accurate. Good callout though to double check there 🙏

marcklingen: just thought that we might have inlcluded a sentence in the `docs` part of the fern schema to highlight how the filters apply and what values are returned. if we return the generation start_at, we could be more specific there to let people know

web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,53 +53,44 @@ export default withMiddlewares({
...
               ${fromTimestampCondition}
               ${toTimestampCondition}
-            GROUP BY
-              1,
-              2
-            ORDER BY
-              1,
-              2
+              ${fromObservationStartTimeCondition}
+              ${toObservationStartTimeCondition}
```
Steffen911, line 68: @marcklingen My understanding from our documentation is that traces should wrap all observations, i.e. my understanding is that t.timestamp <= o.start_time. Could you confirm this?

In addition, I updated the fern doc to make this change here explicit.

I would like to add the observation.start_time condition, because it allows us to use the observation indexes more effectively. Without it, we seem to join _all_ observations for the given project whereas this flag allows for some pre-filtering.

web/src/pages/api/public/metrics/daily.ts
Steffen911: @marcklingen Added a section in the fern schema to highlight how I intend to use the timestamp provided.

web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,53 +53,44 @@ export default withMiddlewares({
...
+            GROUP BY 1
           ),
-          daily_stats AS (
+          trace_usage AS (
             SELECT
               DATE_TRUNC('DAY', t.timestamp) "date",
-              count(distinct t.id)::integer count_traces,
-              count(distinct o.id)::integer count_observations,
-              SUM(o.calculated_total_cost)::DOUBLE PRECISION total_cost
+              count(distinct t.id)::integer as "countTraces"
```
marcklingen, line 92: Can we remove distinct here now that observations are not joined anymore?

web/src/pages/api/public/metrics/daily.ts
```diff
@@ -47,53 +53,44 @@ export default withMiddlewares({
               SUM(o.completion_tokens) "outputUsage",
               SUM(o.total_tokens) "totalUsage",
               COALESCE(SUM(o.calculated_total_cost), 0)::DOUBLE PRECISION as "totalCost"
-            FROM
-              traces t
-            LEFT JOIN observations_view o ON o.trace_id = t.id AND o.project_id = t.project_id
+            FROM traces t
+            LEFT JOIN observations_view o
```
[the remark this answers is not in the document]
marcklingen: I think we should go for this, seems like the biggest performance gain that's possible here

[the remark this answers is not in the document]
marcklingen: We could remove the distinct here in case this yields any performance gain:

count(o.id)::integer as "countObservations"

web/src/pages/api/public/metrics/daily.ts
Steffen911: @marcklingen Yeah, it seems to perform a nested loop join and therefore only uses the trace_id later in the flow. Adding the start_time makes it use the trace_id_project_id_start_time index which seems more performant.

PR #3377, opened 2024-09-17
chore: drop redundant indexes on postgres

Steffen911:
## What does this PR do?

We seem to declare multiple indexes that are covered by another index, e.g. (project_id) on observations even though there is (project_id, start_time, type). On writes this causes additional overhead, because multiple indexes have to be maintained and we also store the included columns redundantly. On reads, they should offer next to no performance gains (they're a slight bit easier to load into memory). 

Overall, from what I've read it's generally recommended to drop indexes that are covered by others.


packages/shared/prisma/migrations/20240917072611_remove_covered_indexes/migration.sql
```diff
@@ -0,0 +1,18 @@
+-- DropIndex
+-- Removes indexes that are covered by other indexes on the same table, e.g. (project_id) if (project_id, timestamp) exists.
+DROP INDEX CONCURRENTLY IF EXISTS "traces_project_id_idx"; -- covered by traces_project_id_timestamp_idx
+DROP INDEX CONCURRENTLY IF EXISTS "datasets_project_id_idx"; -- covered by datasets_project_id_name_key (unique)
```
Steffen911, line 4: This is an interesting case, because datasets_project_id_idx is a hash index while datasets_project_id_name is a standard btree. This may make for slightly faster reads using the first index so also happy to revert this line if there is a strong preference for it.

packages/shared/prisma/migrations/20240917072611_remove_covered_indexes/migration.sql
```diff
@@ -0,0 +1,18 @@
+-- DropIndex
```
Steffen911, line 1: @marcklingen @maxdeichmann Apparently, Prisma does dumb things and wraps this into a single transaction (https://github.com/prisma/prisma/issues/22922) 😞 
Could you review if you agree with the general premise of this change? If yes, I'll separate all those changes into separate folders to avoid the transactions in Prisma - I'd just prefer not to do this if we don't want to follow through with this change 😄

marcklingen: Agree. AFAIK they need to be in a separate migration files as otherwise concurrently also doesn't work correctly.

marcklingen:
Agree with the overall take that we overindex as they organically grew over the last year to solve for different read time performance requirements.

It is unclear to me if dropping an index on eg project_id is a good idea based on having another index on project_id+timestamp. 

project_id+timestamp is used for the main tables and read APIs which all fetch the first n of a given project in a given timeframe. Thus this index helps to cut down the scanning a lot, especially when more filters are applied and the whole timeframe is read.

If we now drop project_id, how does this impact queries that do not involve the timestamp filter. Based on my understanding, these would then do a full table scan (all projects) as the composite index doesn't apply.

Same applies the other way round. If we don't have the composite index on project_id+timestamp, my understanding is that the get api / table queries get slower as the initial index scan doesn't include the time filter leading to scanning all objects of a project + not being able to get the first 50 based on the index, thus needing to sort all rows again.

@maxdeichmann wdyt?

packages/shared/prisma/migrations/20240917072611_remove_covered_indexes/migration.sql
maxdeichmann: Agree as well.

packages/shared/prisma/migrations/20240917072611_remove_covered_indexes/migration.sql
maxdeichmann: Agree, i would delete this.

PR #3388, opened 2024-09-17
chore: fix dev server speed

maxdeichmann:


Steffen911 (commented):
I would propose to set this setting to `false` in https://github.com/langfuse/langfuse/blob/main/.env.dev.example.

web/src/instrumentation.ts
```diff
@@ -1,7 +1,14 @@
 // See: https://vercel.com/docs/observability/otel-overview
 
 export async function register() {
-  if (process.env.NEXT_RUNTIME === "nodejs") {
+  // This variable is set in the .env file or environment variables
+  // Value is true if NEXT_PUBLIC_LANGFUSE_RUN_NEXT_INIT_ENABLED is "true" or undefined
+  const isInitLoadingEnabled =
+    process.env.NEXT_PUBLIC_LANGFUSE_RUN_NEXT_INIT_ENABLED !== undefined
```
marcklingen, line 7: Suggestion

- I would not add another env (NEXT_PUBLIC_LANGFUSE_RUN_NEXT_INIT_ENABLED)
- We could rely on whether LANGFUSE_INIT_ORG_ID exists as it is the topmost check in the initalization file, probably worth adding a note to the initialization script that the condition needs to be updated if the init script changes
- LANGFUSE_INIT_ORG_ID does not exist on localhost usually but is sufficiently tested in ci as our ci pipeline depends on it

Steffen911: @marcklingen In that case I'd only make the `initialize` script conditional. Trying the otel/datadog init to `LANGFUSE_INIT_ORG_ID` feels wrong and unintuitive.

maxdeichmann: I fully agree here with @Steffen911 and hence added a new env variable.

maxdeichmann:
Regarding of the impact of both of these scripts: The dd instrumentation adds the majority of the duration. Hence, i would remove both.

maxdeichmann:
> I would propose to set this setting to `false` in https://github.com/langfuse/langfuse/blob/main/.env.dev.example.

Yes, we use this in the pipeline as well. As we are using LANGFUSE_INIT_ORG_ID there to work well, we should keep it out of there. -> I will add it to the examples file and remove it in the pipeline.

web/src/instrumentation.ts
marcklingen: Sorry for the confusion, I remembered that you wanted to keep the datadog instrumentation at all times. With the timings below it makes sense to remove datadog as well.

Change looks good to me

.env.dev.example
```diff
@@ -51,3 +51,6 @@ REDIS_AUTH="myredissecret"
 LANGFUSE_WORKER_PASSWORD=mybasicauthsecret
 # openssl rand -hex 32 used only here
 ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 
+
+# speeds up local development by not executing init scripts on server startup
+NEXT_PUBLIC_LANGFUSE_RUN_NEXT_INIT_ENABLED="false"
```
Steffen911, line 56: Small nitpick: Naming feels a bit redundant. "Run next init enabled" works better without the enabled. IMO we could also drop the _ENABLED here.

Steffen911 (approved):

PR #3400, opened 2024-09-18
chore: move web-ee logic from console.log to logger

Steffen911:
## What does this PR do?

Move additional logic from console.log to the langfuse logger to take advantage of structured logging.


web/src/ee/features/billing/stripeClientReference.ts
```diff
@@ -1,9 +1,10 @@
 import { env } from "@/src/env.mjs";
+import { logger } from "@langfuse/shared/src/server";
 
 // used client-side to create a stripe customer reference
 export const createStripeClientReference = (orgId: string) => {
   if (!env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION) {
-    console.error(
+    logger.error(
```
marcklingen, line 7: This is run client side, see comment above

Steffen911: @marcklingen Are you sure about this one? The file is imported in the `cloudBillingRouter.ts` which creates a new TRPCRouter and in the stripeWebhookApiHandler.ts where it implements a POST endpoint for stripe webhooks. Hence, my interpretation from the code is that we purely use it on the API/Nodejs side, not in React.

That would lead me to the interpretation that it's the "endpoint" which is used client side to create a stripe customer reference.

marcklingen: sorry for the confusion, you are right. I just updated the comment. Originally I planned to use this client side but then forgot to update the comment when moving this to server-side-only.

good to go

marcklingen (approved):

PR #3411, opened 2024-09-19
fix(dataset-table): cost and latency queries

hassiebp:


hassiebp:
- add comment with limitation that all items must be consistently linked to either a trace or an observation for a given run
- move queries to observartion_view as cost are not backfilled for self hosters

PR #3494, opened 2024-09-27
fix(ui): include non-generation warnings/error in rolled-up trace status

marcklingen:
Previously the level was only rolled up from generations and non-generation observations were ignored.


marcklingen:
tested e2e, ui & exports

PR #3509, opened 2024-09-27
chore: make credentials for export bucket optional

Steffen911:
## What does this PR do?

Makes accessKeyId, accessKeySecret, region, and endpoint optional for all export bucket interactions. This has the advantage that we can skip them in our own setup and instead fallback to the default credential provider chain which would use an IAM role that we assign to our service. That way, we can remove long-living credentials from our stack that we've previously used to authenticate with a S3 compliant interface.

### DIFF
--- a/worker/src/features/batchExport/handleBatchExportJob.ts
+++ b/worker/src/features/batchExport/handleBatchExportJob.ts
@@ -36,13 +36,13 @@ const tableNameToTimeFilterColumn = {
 };
 
 const isGenerationTimestampFilter = (
-  filter: FilterCondition
+  filter: FilterCondition,
 ): filter is TimeFilter => {
   return filter.column === "Start Time" && filter.type === "datetime";
 };
 
 const isTraceTimestampFilter = (
-  filter: FilterCondition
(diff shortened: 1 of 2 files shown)


Steffen911:
@hassiebp As you've been mostly working on the exports based on the Git blame file, could you take a look whether you see any issues with this change?

worker/src/features/batchExport/handleBatchExportJob.ts
```diff
@@ -333,8 +327,8 @@ export const handleBatchExportJob = async (
   const endpoint = env.S3_ENDPOINT;
   const region = env.S3_REGION;
 
-  if (!accessKeyId || !secretAccessKey || !bucketName || !endpoint || !region) {
-    throw new Error("S3 credentials not found");
+  if (!bucketName) {
+    throw new Error("No S3 bucket configured for exports.");
```
Steffen911, line 331: This is the only non-prettier/linting change in this file. Again, we only validate that there is a bucket and rely on S3 errors if further credentials are missing.

hassiebp (approved):

PR #3511, opened 2024-09-27
feat(generations-table): show tokens per second

hassiebp:


marcklingen:
@hassiebp can this be merged?

PR #3537, opened 2024-09-30
chore: use s3 event store to avoid event data in redis

Steffen911:
## What

Use the raw events in S3 to avoid sending the full event body via Redis. Reconstruct the event information on the worker side for downstream processing.


.env.dev.example
```diff
@@ -34,11 +34,25 @@ EMAIL_FROM_ADDRESS=""  # Defines the email address to use as the from address.
...
-S3_SECRET_ACCESS_KEY=
-S3_BUCKET_NAME=
-S3_REGION=
+S3_ENDPOINT=http://localhost:9090
+S3_ACCESS_KEY_ID=minio
+S3_SECRET_ACCESS_KEY=miniosecret
+S3_BUCKET_NAME=mybucket
+S3_REGION=us-east-1
+## Necessary for minio compatibility
+S3_FORCE_PATH_STYLE=true
```
Steffen911, line 43: TODO: Needs a doc update once we deploy it to prod.

packages/shared/src/server/ingestion/legacy/index.ts
```diff
@@ -208,7 +180,7 @@ export const isNotNullOrUndefined = <T>(
 export const isUndefinedOrNull = <T>(val?: T | null): val is undefined | null =>
   val === undefined || val === null;
 
-export const sendToWorkerIfEnvironmentConfigured = async (
+export const addTracesToTraceUpsertQueue = async (
```
Steffen911, line 183: Adding a clearer naming to explain the purpose of the function.

packages/shared/src/server/ingestion/legacy/index.ts
```diff
@@ -235,32 +207,6 @@ export const sendToWorkerIfEnvironmentConfigured = async (
       }
 
       await queue.addBulk(convertTraceUpsertEventsToRedisEvents(traceEvents));
-    } else if (
-      env.LANGFUSE_WORKER_HOST &&
-      env.LANGFUSE_WORKER_PASSWORD &&
```
Steffen911, line 240: Remove old http based upsert as we've removed this in our cloud deployment.

packages/shared/src/server/services/S3StorageService.ts
```diff
@@ -40,10 +39,11 @@ export class S3StorageService {
 
     this.client = new S3Client({
       credentials,
-      endpoint,
-      region,
+      endpoint: params.endpoint,
+      region: params.region,
+      forcePathStyle: params.forcePathStyle,
```
Steffen911, line 44: `forcePathStyle` is a required variable to make the S3 client minio compatible. Useful for local deploys and self-hosters.

packages/shared/src/server/queues.ts
```diff
@@ -20,6 +21,29 @@ export const LegacyIngestionEvent = z.object({
...
+  authCheck: z.object({
+    validKey: z.literal(true),
+    scope: z.object({
+      projectId: z.string(),
+      accessLevel: z.enum(["all", "scores"]),
+    }),
+  }),
+});
+
+export const LegacyIngestionEvent = z.discriminatedUnion("useS3EventStore", [
```
maxdeichmann, line 42: thx for the backwards compatibility

Steffen911: @maxdeichmann I've also verified that an event without the `useS3EventStore` property is processed successfully on new workers, i.e. we're also good during the deployment _and_ have this as a fallback.

worker/src/queues/legacyIngestionQueue.ts
```diff
@@ -8,21 +8,77 @@ import {
...
+      if (
+        env.LANGFUSE_S3_EVENT_UPLOAD_ENABLED !== "true" ||
+        !env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET
+      ) {
+        throw new Error(
+          "S3 event store is not enabled but useS3EventStore is true",
+        );
+      }
+      // If we used the S3 store we need to fetch the ingestionEvents from S3
+      const s3Client = new S3StorageService({
```
maxdeichmann, line 41: How efficient is it to always re-instantiate the service and also within the S3 client? Should this be treated as singleton to make use of longer lasting TCP connections?

Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/3537/commits/61b5623302e2fb47841110820913ae2b04774065. Good catch!

PR #3547, opened 2024-10-01
fix(prisma): HMR support

hassiebp:
Fixes hitting max_connection limit during hot module reload when developing next JS app.


packages/shared/src/db.ts
```diff
@@ -73,14 +73,40 @@ export class KyselySingleton {
...
+            // Don't forget to customize these to match your database!
+            createAdapter: () => new PostgresAdapter(),
+            createIntrospector: (db) => new PostgresIntrospector(db),
+            createQueryCompiler: () => new PostgresQueryCompiler(),
+          },
+        }),
+    })
+  );
+}
+s;
```
marliessophie, line 106: Leaving comment so we don't accidentally merge this

maxdeichmann (approved):

PR #3562, opened 2024-10-02
FIX: Typscript Error - ts(2686)

vanshavenger:
## What does this PR do?

In some of the Next js  (React) files, some lines were showing typescript error, Please find the error attached in the Screenshot.
The error resolution for react 17 or above is to just an jsx line in the tsconfig.json

I found it when i was working on auth providers and felt that it can be resolved easily.

Screenshot 2024-10-02 at 10 36 02 PM

## Type of change

## Mandatory Tasks


vanshavenger:
@marcklingen can you check this?
It also reduces the bundle size and some other small positive improvements

marcklingen:
thanks for the contribution on this. Langfuse uses NextJS which automatically reverts this to `"jsx": "preserve"`, see for example this thread: https://github.com/vercel/next.js/discussions/19155

From my point of view preserve seems to be the correct setting for NextJS outputs as it is the standard

vanshavenger:
Yeah completely agree with you @marcklingen, But then for the resolution of the typescript error which you can see we either have to add React imports to each file or ig App router handles it by itself. 
Kindly discuss and suggest, will make changes accordingly.

Otherwise will close this PR as obviously your point is valid
thank you

marcklingen:
is there a good way to ignore/overrule the warning for JSX?

marcklingen:
Interesting I cannot reproduce this warning locally right now, for which files do you get this?

vanshavenger:
Marc i found this in sign-in.tsx file 
in files where there is 
<></>
used 

I shared the same in the screenshot

marcklingen:
Interesting, I cannot reproduce this, no warning/error in sign-in for me

vanshavenger:
Screenshot 2024-10-04 at 12 20 54 PM

PR #3563, opened 2024-10-02
feat: send transactional email to owners and admins when a new member joins

### LINKED ISSUE #3504: feat: send transactional email to organization owners/admins when a new member joins the org
See issue title, Langfuse already has logic based on react.email and nodemailer for transactional emails (invites, password reset). From a security perspective, org admins/owners should be notified whenever a new person joins the organization.

mohanadkandil:
## What does this PR do?

This PR introduces email notifications for admins and owners when a new member joins an organization, ensuring timely updates and improved oversight.

Fixes #3504

## Type of change

## Mandatory Tasks

## Checklist


marcklingen (changes requested):
Overall this is great, thanks for the contribution!

I think we should adjustt the time at which this email is sent. Currently it is only sent for users who get their project-level roles changed.

I usggest to not send an email in those cases, but in all in which a user is added to a net-new org. This happens either when they are added to an org in which they were not a member before, or if they were invited to an org and then became members when creating a new account with the same email address. Can you make these changes?

PR #3564, opened 2024-10-02
FEAT: Gitlab Auth Provider

vanshavenger:
## What does this PR do?
The PR add the next Auth gitlab provider support to langfuse, Providing a better DX and Gitlab option for them to sigup 🚀

Fixes # (issue)
#2468 

## Type of change

## Mandatory Tasks

Kindly let me know if any more changes are required.


marcklingen (approved):
this is great, thank you!

ptempier:
I am not sure how it's supposed to work.
Isn't there a AUTH_GITLAB_ISSUER variable ?

There's not a single gitlab server, there's one gitlab per company.
Github however is the publix github instance available for everyone.
So you need to specify somewhere on which gitlab instance the ID and secret are valid.

Maybe i have a cache issue, but i tried AUTH_GITLAB_ISSUER and it didnt  do anything
and trying with  AUTH_CUSTOM_* didn't change anything compared to before this version

marcklingen:
@vanshavenger did you test this? I've assumed that you self-reviewed this as my review focused on that the provider was consistently added across the application. can you fix this?

vanshavenger:
For that 
issuer: process.env.GITLAB_ISSUER,
 thing need to be added as i have worked on something similar before, But I have followed the docs and understood the code structure and ran it myself to add on my local machine too. 
 
I am not sure if it is needed here. 

unless you're working with a custom OAuth2 provider or GitLab instance that requires it. The issuer is typically used for OpenID Connect (OIDC) flows, where it refers to the base URL of the identity provider.

or the gitlab is self hosted.

The above line change will be enough for that. GitlabProvier has that option.
 
 Reference 
 https://docs.gitlab.com/ee/integration/oauth_provider.html
 https://next-auth.js.org/providers/gitlab (Since we using Next-Auth) not Auth JS

ptempier:
Let's say i have a gitlab instance, what should i configure in langfuse to be able to authenticate through it ?

vanshavenger:
in the GitLap Auth Provider, Pass this

Make sure env is set in env.mjs in server and runtime

issuer: env.GITLAB_ISSUER,

vanshavenger:
also @marcklingen can we move env.mjs to env.ts, Both are valid syntaxes but Trying to keep it in Fully TS

marcklingen:
can you make the change? Feel free to tag me on the PR

Please don’t change it. The mjs is used in non-TS scripts afaik and this will break things.

vanshavenger:
@marcklingen Sure won't change that. 

And okay raising A PR to add the issuer to the NextAuth Gitlab Provider :))

PR #3617, opened 2024-10-07
chore: reuse DB client instantiation functions

Steffen911:
## What

Reuse the client instantiation functions to create global singletons.

----

> [!IMPORTANT]
> Refactor `db.ts` to reuse client instantiation functions for creating global singletons with optional forced instantiation.
> 
>   - **Behavior**:
>     - `PrismaClientSingleton.getInstance()` and `KyselySingleton.getInstance()` now accept a `forceNew` parameter to control singleton instantiation.
>     - In development mode, global singletons `prismaGlobal` and `kyselyPrismaGlobal` are instantiated with `forceNew` set to `true`.
>   - **Code Simplification**:
>     - Removed redundant instantiation logic in `db.ts` by reusing `getInstance()` methods for global singletons.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 759540af00f98dc8f4df3cd27a9b92cd6f21f3c4. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/db.ts
+++ b/packages/shared/src/db.ts
@@ -16,8 +16,8 @@ import { logger } from "./server";
 export class PrismaClientSingleton {
   private static instance: PrismaClient;
 
-  public static getInstance(): PrismaClient {
-    if (PrismaClientSingleton.instance) {
+  public static getInstance(forceNew = false): PrismaClient {
+    if (!forceNew && PrismaClientSingleton.instance) {
       return PrismaClientSingleton.instance;
     }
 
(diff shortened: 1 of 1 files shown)


packages/shared/src/db.ts
```diff
@@ -86,23 +88,9 @@ declare const globalThis: {
...
-            // Don't forget to customize these to match your database!
-            createAdapter: () => new PostgresAdapter(),
-            createIntrospector: (db) => new PostgresIntrospector(db),
-            createQueryCompiler: () => new PostgresQueryCompiler(),
-          },
-        }),
-    })
-  );
-};
+  globalThis.prismaGlobal ??= PrismaClientSingleton.getInstance(true);
```
hassiebp, line 91: I believe this will recreate the dev issues we saw previously. In this line, `prismaGlobal` will rely on the reference to the instance property on `PrismaClientSingleton` via `getInstance`. On hot reload, modules incl. the `shared` package will be reloaded such that the `PrismaClientSingleton.instance` will be undefined again, forcing the creation of a fresh client. Since on hot reload connections aren't closed gracefully, we will hit max_connections eventually. 

To fix the dev issues we have to create a separate client here (as previously done) or handle the `prismaGlobal` inside the PrismaClientSingleton to return the client bound to globalThis but not rely on the reference on `PrismaClientSingleton.instance`

PR #3695, opened 2024-10-10
Fix "none of" label filtering in prompts

### LINKED ISSUE #2279: bug: Label filters not working
### Describe the bug

Label filters not working in prompts

### To reproduce

Add a label development. Try to find prompts in the filter with none of the development. It won't work!

_No response_

### Additional information

_No response_

No

maverpolo022:
## What does this PR do?

The "none of" label requires that we exclude any prompts where that label is present in any of the versions. Currently, due to the sequence of Group By after Where clause, those prompts that should have been excluded were being added due to some of their versions passing the filter.

Fixes #2279 

## Type of change

## Mandatory Tasks

## Checklist

- My code doesn't follow the style guidelines of this project (`npm run prettier`)

----

> [!IMPORTANT]
> Fixes filtering logic for 'none of' labels in prompts by using a SQL subquery in `tableColumnsToSqlFilter`.
> 
>   - **Behavior**:
>     - Fixes filtering logic for 'none of' labels in `tableColumnsToSqlFilter` in `filterToPrisma.ts`.
>     - Uses SQL subquery to exclude prompts with 'none of' label in any version.
>   - **Misc**:
>     - Addresses issue #2279.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 8080c01ef93053c6554090a71ea25ba30e89f6d1. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/filterToPrisma.ts
```diff
@@ -136,9 +136,11 @@ export function tableColumnsToSqlFilter(
         : [Prisma.empty, Prisma.empty];
     const [funcPrisma1, funcPrisma2] =
       filter.type === "arrayOptions" && filter.operator === "none of"
-        ? [Prisma.raw("NOT ("), Prisma.raw(")")]
+        ? (table==="prompts" && filter.column === "Labels"
+          ? [Prisma.raw("name NOT IN (SELECT DISTINCT name FROM prompts p WHERE "), Prisma.raw(")")]
```
hassiebp, line 140: There is a subtle issue here in that the `project_id` filter is missing in this custom subquery. That means that if there is another project that has a prompt with the same name and label set, it will not show in the dashboard even though the prompt in the user project has that label not set.

hassiebp (changes requested):
Thanks a lot for your work @mr-loop-1 ! As you suggested I would prefer having the none-of case specifically handled in the promptRouter rather than in the filterToPrisma condition to be more explicit.

packages/shared/src/server/filterToPrisma.ts
mr-loop-1: I checked a lot of approaches today
1. it would be difficult to handle in `promptRouter` as there can be combination of filters which are injected into the generate function together.
2. Additionally, the filter sql is returned from shared package's `filterToPrisma` as combined and joined. Hence, we can't break it down after calling the function.

hassiebp:
Thanks again for your contribution - closing this PR as stale. Please open a Github issue if this persists

PR #3719, opened 2024-10-11
chore: migrate clickhouse ingestion to S3 based events

Steffen911:
## What does this PR do?

This adds a new ingestionQueue which receives single events from the /ingestion endpoint. It then reads all relevant information from S3 before merging them together and writing into Clickhouse. The logic replaces the current ingestionFlushQueue behaviour and therefore the latter is removed.

I keep most of the IngestionService logic in tact as part of this PR to decouple changes and make the total PR smaller. The eventual goal is to drop the Clickhouse reads in the ingestion service and only rely on the events coming from S3. 

We will also change the clickhouse writer behaviour in a follow-up PR to drop the queueing and instead rely on the async insert behaviour of Clickhouse.

----

> [!IMPORTANT]
> Migrates ingestion to use S3-based events with a new `IngestionQueue`, removing the old `ingestionFlushQueue`.
> 
>   - **Behavior**:
>     - Introduces `IngestionQueue` to process single events from S3 and write to Clickhouse, replacing `ingestionFlushQueue`.
>     - Removes `ingestionFlushQueue` and related logic from `ingestionFlushQueue.ts` and `enqueueIngestionEvents.ts`.
>     - Updates `handler` in `ingestion.ts` to use `IngestionQueue` for async processing.
>   - **Environment**:
>     - Removes `LANGFUSE_INGESTION_FLUSH_DELAY_MS` and `LANGFUSE_INGESTION_FLUSH_ATTEMPTS` from `env.ts`.
>     - Adds `LANGFUSE_INGESTION_QEUEUE_PROCESSING_CONCURRENCY` to `env.ts`.
>   - **Services**:
>     - Adds `listFiles` method to `S3StorageService` for listing S3 files.
>     - Updates `IngestionService` to handle event merging and writing to Clickhouse.
>   - **Tests**:
>     - Updates `redis-consumer.test.ts` to test `IngestionQueue` instead of `ingestionFlushQueue`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 9bbc3c8e36dd4851fd334c485d1ab547bfc11733. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/queues.ts
```diff
@@ -105,27 +120,29 @@ export type TQueueJobTypes = {
...
     payload: EvalExecutionEventType;
     name: QueueJobs.EvaluationExecution;
-    _tracecontext?: TCarrier;
   };
   [QueueName.BatchExport]: {
     timestamp: Date;
     id: string;
     payload: BatchExportJobType;
     name: QueueJobs.BatchExportJob;
-    _tracecontext?: TCarrier;
```
maxdeichmann, line 122: why are these gone?

web/src/pages/api/public/ingestion.ts
```diff
@@ -244,6 +245,41 @@ export default async function handler(
       });
     }
 
+    // Send each event individually to IngestionQueue for new processing
```
maxdeichmann, line 248: i think we should add here another env to be able to determine whether we want to send these events or not.

maxdeichmann: I think we never want to not upload anymore to have full history in S3. But we might want to turn off the processing.

worker/src/queues/ingestionQueue.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    // Download all events from folder into a local array
+    const eventFiles = await s3Client.listFiles(
+      `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${job.data.payload.authCheck.scope.projectId}/${eventName}/${job.data.payload.data.eventBodyId}/`,
+    );
+
+    const events: IngestionEventType[] = (
+      await Promise.all(
+        eventFiles.map(async (key) => {
+          const file = await s3Client.download(key);
```
maxdeichmann, line 77: do we have to download all the files? I thought that the list endpoint already returns the objects.

packages/shared/src/server/queues.ts
Steffen911: @maxdeichmann bullmq automatically propagates the context for us. Therefore, we do not need them to the event body.

web/src/pages/api/public/ingestion.ts
Steffen911: @maxdeichmann Shall we just check whether the CLICKHOUSE_URL is set? That was the "gate" in the old flow.

Steffen911: I added LANGFUSE_ASYNC_CLICKHOUSE_INGESTION_PROCESSING for gating it.

PR #3770, opened 2024-10-15
chore: batch events by eventBodyId for reduces S3 interactions

Steffen911:
## What does this PR do?

Batches events by eventBodyId and uploads them into S3 as one. We use the eventId of the first event as the key in the queue. The worker downloads the entire file and flattens the results, i.e. there is no need to change any processing on the worker side.

----

> [!IMPORTANT]
> Batch events by `eventBodyId` in `ingestion.ts` to reduce S3 interactions, using the first event's `eventId` as the key, with no changes needed on the worker side.
> 
>   - **Behavior**:
>     - Batch events by `eventBodyId` in `ingestion.ts` to reduce S3 interactions.
>     - Use the `eventId` of the first event as the key for S3 uploads.
>     - Worker processes entire file, no changes needed on worker side.
>   - **Async Processing**:
>     - Modify S3 upload logic to handle batches by `eventBodyId` in `ingestion.ts`.
>     - Update `IngestionQueue` and `LegacyIngestionQueue` to handle batched events.
>   - **Error Handling**:
>     - Log errors for failed S3 uploads and queue additions, fallback to sync processing if necessary.
>   - **Misc**:
>     - Increase `removeOnFail` limit in `LegacyIngestionQueue` from 100,000 to 500,000.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 17ee7ed2aacfca88880093a63c66e9e6793fdadd. It will automatically update as commits are pushed.</sup>


web/src/pages/api/public/ingestion.ts
```diff
@@ -212,6 +212,36 @@ export default async function handler(
...
+            data: IngestionEventType[];
+            key: string;
+            type: (typeof eventTypes)[keyof typeof eventTypes];
+          }
+        >,
+        event,
+      ) => {
+        if (!event.body?.id) {
+          return acc;
+        }
```
Steffen911, line 231: If there is no body or id, we skip the upload and worker processing. Based on the typing this should only occur for SdkLog events.

web/src/pages/api/public/ingestion.ts
```diff
@@ -212,6 +212,36 @@ export default async function handler(
 
     const sortedBatch = sortBatch(batch);
 
+    // We group events by eventBodyId which allows us to store and process them
+    // as one which reduces infra interactions per event. Only used in the S3 case.
+    const sortedBatchByEventBodyId = sortedBatch.reduce(
```
Steffen911, line 217: @maxdeichmann This will store all events by body id in order, but we won't have a guaranteed global order. Is that ok? My understanding is that only events with the same eventBodyId need to be processed in the correct order.

web/src/pages/api/public/ingestion.ts
```diff
@@ -297,35 +326,28 @@ export default async function handler(
...
-            },
-            {
-              removeOnFail: 1_000_000,
-              removeOnComplete: true,
-              attempts: 5,
-              backoff: {
-                type: "exponential",
-                delay: 1000,
-              },
-            },
```
Steffen911, line 327: This is actually covered by the default settings on the queue. No need to repeat that here for each event.

packages/shared/src/server/redis/legacyIngestion.ts
```diff
@@ -21,7 +21,7 @@ export class LegacyIngestionQueue {
             connection: newRedis,
             defaultJobOptions: {
               removeOnComplete: true,
-              removeOnFail: 100_000,
+              removeOnFail: 500_000,
```
Steffen911, line 24: Currently, we overwrite that to 1 million. I think picking the middle here is a reasonable call.

web/src/pages/api/public/ingestion.ts
maxdeichmann: Correct, only events with the same eventBodyId need to e processed in the same order.

web/src/pages/api/public/ingestion.ts
```diff
@@ -212,6 +212,36 @@ export default async function handler(
...
+          return acc;
+        }
+        if (!acc[event.body.id]) {
+          acc[event.body.id] = {
+            data: [],
+            key: event.id,
+            type: event.type,
+          };
+        }
+        acc[event.body.id].data.push(event);
```
maxdeichmann, line 239: We could also have the same id for different types (e.g. scores, traces, observations). This would be valid. We need to ensure that they are processed separately. Hence checking for type + id is probably better here.

Steffen911: @maxdeichmann Covered in https://github.com/langfuse/langfuse/pull/3770/commits/17ee7ed2aacfca88880093a63c66e9e6793fdadd.

PR #3810, opened 2024-10-17
feat: read traces and observations from clickhouse

maxdeichmann:
> [!IMPORTANT]
> Add functionality to read traces and observations from Clickhouse with eligibility checks and component updates.
> 
>   - **Behavior**:
>     - Add `getGenerationsTable` and `getTracesTable` functions to query Clickhouse for generations and traces data.
>     - Introduce `isClickhouseEligible` function to check if a user can query Clickhouse.
>     - Modify `traceRouter` and `getAllQueries` to support querying Clickhouse based on user eligibility.
>   - **Models**:
>     - Add new types `GenerationsReturnType` and `TracesTableReturnType` for Clickhouse query results.
>     - Define interfaces for `Trace`, `Observation`, `Score`, and `ObservationWide` in `schema.ts`.
>   - **Components**:
>     - Update `GenerationsTable` and `TracesTable` components to use Clickhouse data when eligible.
>     - Add `useClickhouse` hook to toggle Clickhouse querying in components.
>   - **Misc**:
>     - Rename `clickhouse.ts` to `clickhouse/clickhouse.ts`.
>     - Export new Clickhouse-related functions and types in `index.ts`.
>     - Add placeholder `filter` directory with commented-out `FilterFactory` class.
>     - Add migrations to add and remove `queue_id` column in `scores` table.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for db2addb2435638971385f41c3e1eb4aaddd3503e. It will automatically update as commits are pushed.</sup>


packages/shared/clickhouse/migrations/0004_scores_add_queue_id.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scores
+ADD COLUMN queue_id Nullable(String) AFTER string_value;
```
maxdeichmann, line 2: @Steffen911 FYI, i assume this is missing in the migration as well?

packages/shared/src/server/queries/clickhouse-filter/clickhouse-filter.ts
```diff
@@ -0,0 +1,253 @@
...
+export interface Filter {
+  apply(): ClickhouseFilter;
+  clickhouseTable: string;
+}
+type ClickhouseFilter = {
+  query: string;
+  params: { [x: string]: any };
+};
+
+export class StringFilter implements Filter {
```
Steffen911, line 12: Filteroperations look solid. Very nice!

packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -1,51 +1,56 @@
...
 ) => {
-  const query = `SELECT * FROM observations FINAL WHERE id = '${observationId}' AND project_id = '${projectId}' LIMIT 1`;
-  const records = await queryClickhouse(query);
+  const query = `SELECT * FROM observations FINAL WHERE id = {observationId: String} AND project_id = {projectId: String} LIMIT 1`;
+  const records = await queryClickhouse({
+    query,
+    params: { observationId, projectId },
+  });
 
   return records.length ? convertObservations(records)[0] : undefined;
```
Steffen911, line 32: Is records empty if there is no result? Then `return convertObservations(records).shift()` is way easier to read.

packages/shared/src/server/repositories/scores.ts
```diff
@@ -0,0 +1,104 @@
...
+  string_value: string | null;
+  queue_id: string | null;
+  created_at: string;
+  updated_at: string;
+  event_ts: string;
+  is_deleted: number;
+  projectId: string;
+};
+
+const convertToScore = (row: FetchScoresReturnType) => {
```
Steffen911, line 26: How about we use the results as is and skip the conversion?

packages/shared/src/server/repositories/scores.ts
```diff
@@ -0,0 +1,104 @@
+import { ScoreDataType, ScoreSource } from "@prisma/client";
+import { queryClickhouse } from "./clickhouse";
+
+export type FetchScoresReturnType = {
```
Steffen911, line 4: Wouldn't this be the same as the type we define in the schema?

packages/shared/src/server/repositories/traces.ts
```diff
@@ -0,0 +1,159 @@
...
+> & {
+  level: ObservationLevel;
+  observation_count: number | null;
+  latency: number;
+  usage_details: Record<string, number>;
+  cost_details: Record<string, number>;
+  scores_avg: Array<{ name: string; avg_value: number }>;
+};
+
+export const getTracesTable = async (
```
Steffen911, line 34: Overall, I think this makes for a good DX until we find a better solution or an ORM. As long as we keep all within a single file that should be good.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -0,0 +1,159 @@
...
+      sumMap(cost_details) as cost_details,
+      trace_id,
+      project_id
+    FROM
+        observations final
+    WHERE ${observationsStatsRes.query}
+    group by trace_id, project_id
+),
+
+         scores_avg AS (SELECT project_id,
```
Steffen911, line 74: Maybe we can have some linting here :-)

packages/shared/src/server/repositories/traces.ts
```diff
@@ -0,0 +1,159 @@
...
+        arrayExists(x -> x = 'ERROR', groupArray(level)), 'ERROR',
+        arrayExists(x -> x = 'WARNING', groupArray(level)), 'WARNING',
+        arrayExists(x -> x = 'DEFAULT', groupArray(level)), 'DEFAULT',
+        'DEBUG'
+      ) AS level,
+      sumMap(cost_details) as cost_details,
+      trace_id,
+      project_id
+    FROM
+        observations final
```
Steffen911, line 69: Those finals may become problematic - perhaps we can already have a backup plan ready in case we need to migrate off of them.

web/src/server/api/routers/traces.ts
```diff
@@ -49,6 +55,76 @@ export type ObservationReturnType = Omit<
...
+    tags: row.tags,
+    bookmarked: row.bookmarked,
+    release: row.release,
+    version: row.version,
+    projectId: row.project_id,
+    userId: row.user_id,
+    sessionId: row.session_id,
+    public: row.public,
+  };
+};
```
Steffen911, line 88: Shouldn't those go into a traces repository?

packages/shared/clickhouse/migrations/0004_scores_add_queue_id.up.sql
maxdeichmann: When a user creates a score from the Annotation queue, we add this to the score upon score creation. I would add it here for sure.

packages/shared/src/server/clickhouse/schema.ts
```diff
@@ -0,0 +1,292 @@
...
+  input?: string;
+  output?: string;
+  provided_model_name?: string;
+  internal_model_id?: string;
+  model_parameters?: string;
+  provided_usage_details?: Record<string, number>; // Map(LowCardinality(String), UInt64)
+  usage_details?: Record<string, number>; // Map(LowCardinality(String), UInt64)
+  provided_cost_details?: Record<string, number>; // Map(LowCardinality(String), Decimal64(12))
+  cost_details?: Record<string, number>; // Map(LowCardinality(String), Decimal64(12))
+  total_cost?: number; // Nullable(Decimal64(12))
```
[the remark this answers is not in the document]
maxdeichmann: Thanks! I would leave it as it is as it will be less specific over time.

packages/shared/src/server/repositories/scores.ts
maxdeichmann: Unfortunately not possible as we somehow need to map casing and names

packages/shared/src/server/repositories/traces.ts
maxdeichmann: We would need to group by and select the first one.

packages/shared/src/server/repositories/traces.ts
maxdeichmann: unfortunately not :(

web/src/server/api/routers/traces.ts
maxdeichmann: If we want to do this, we would need all the types to shared. I would prefer to keep them where they are and maybe move them to a dedicated file.

packages/shared/clickhouse/migrations/0004_scores_add_queue_id.up.sql
Steffen911: @maxdeichmann I included this change in https://github.com/langfuse/langfuse/pull/3960/files#diff-ac64ef43424291171fd4ebfbd3fbaca280f563f6ddce31e08b834586804dea34R15. Hence you can remove it here.

packages/shared/clickhouse/migrations/0004_scores_add_queue_id.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scores
+DROP COLUMN queue_id;
```
Steffen911, line 2: Covered via https://github.com/langfuse/langfuse/pull/3960/files#diff-ac64ef43424291171fd4ebfbd3fbaca280f563f6ddce31e08b834586804dea34R15. Can be removed.

maxdeichmann: Thanks!

PR #3811, opened 2024-10-17
feat(cost): make cost tracking flexible

hassiebp:
> [!IMPORTANT]
> Introduce flexible cost tracking by replacing fixed fields with dynamic maps and adding a `Price` model, updating ingestion logic and tests accordingly.
> 
>   - **Database Schema Changes**:
>     - Replace fixed cost and usage fields with `Map` types in `observations`, `observations_wide`, and `traces_wide`.
>     - Add `Price` model to `types.ts` and create `prices` table in `migration.sql`.
>     - Make `unit` column nullable in `models` table.
>   - **Ingestion Service**:
>     - Update `IngestionService` to calculate costs using `usage_details` and `cost_details` maps.
>     - Add `getModelPrices()` to fetch prices for models.
>     - Modify `getUsageUnits()` to handle dynamic usage details.
>   - **Scripts and Initialization**:
>     - Add `upsertDefaultModelPrices.ts` to upsert default model prices.
>     - Initialize default model prices in `initialize.ts`.
>     - Create `createDefaultModelPricesJson.ts` to export model prices to JSON.
>   - **Tests**:
>     - Update integration and unit tests in `IngestionService.integration.test.ts` and `calculateTokenCost.unit.test.ts` to verify new cost calculation logic.
>   - **Miscellaneous**:
>     - Add `default-model-prices.json` with default model prices.
>     - Minor logging and import changes in `app.ts` and `utils.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e0806ec754e8037d2d95776b4f0a491c39d5b687. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/schema.prisma
```diff
@@ -753,15 +753,30 @@ model Model {
...
     @@map("models")
 }
 
+model Price {
+    id        String   @id @default(cuid())
+    createdAt DateTime @default(now()) @map("created_at")
+    updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
+    modelId   String   @map("model_id") // Model is already linked to project (or default), so we don't need projectId here
+    Model     Model    @relation(fields: [modelId], references: [id], onDelete: Cascade)
+    itemName  String   @map("item_name")
```
maxdeichmann, line 772: what does itemName mean? Maybe we can find better naming here.

worker/src/services/IngestionService/index.ts
```diff
@@ -418,54 +419,56 @@ export class IngestionService {
...
+      ...final_cost_details,
       internal_model_id: internalModel?.id,
-      unit: observationRecord.unit ?? internalModel?.unit,
     };
   }
 
-  private getTokenCounts(
+  private async getModelPrices(modelId?: string): Promise<Price[]> {
+    return modelId
+      ? ((await this.prisma.price.findMany({ where: { modelId } })) ?? [])
```
maxdeichmann, line 454: do we also need a project_id check on the models table here? Either matching project_id or null

hassiebp: Each price is already linked to a `model_id` which in turn is linked to a project or null. Would be duplicate here and some of the prisma queries do not work well with project_id null in a composite key

packages/shared/prisma/schema.prisma
hassiebp: Agree, will come up with more options here

maxdeichmann: 1.	usageType - Generic and flexible, covers any type of usage measurement
	2.	metricType - Generic term for different types of measurements
	3.	countType - Simple and clear for counting different elements
	4.	measureType - Indicates what kind of measurement we're tracking
	5.	llmMetricType - Specific to LLM but generic about what's being measured
	6.	usageCategory - Categorizes different types of usage
	7.	consumptionType - Focuses on what's being consumed
	8.	processType - If you're tracking different stages of processing
	
	
	what do you like best?

worker/src/initialize.ts
```diff
@@ -0,0 +1,3 @@
+import { upsertDefaultModelPrices } from "./scripts/upsertDefaultModelPrices";
+
+upsertDefaultModelPrices();
```
maxdeichmann, line 3: is it intended that this here is not awaiting?

worker/src/scripts/upsertDefaultModelPrices.ts
```diff
@@ -0,0 +1,187 @@
...
+              );
+            })
+        );
+      }
+
+      await Promise.all(promises);
+      logger.debug(`Completed batch ${i + 1} of ${numBatches}.`);
+    }
+
+    logger.debug("Finished upserting default model prices.");
```
maxdeichmann, line 178: Can you make this an info logger and add latency here how much time this function took?
Also, please do the same in the catch block below.

worker/src/initialize.ts
hassiebp: yes, we are running this at server bootstrap and there is no top-level await in node currently. This is in any case fine to run async 👍🏾

packages/shared/prisma/schema.prisma
hassiebp: I like `usageType` - let's go with that

worker/src/scripts/upsertDefaultModelPrices.ts
hassiebp: Done!

packages/shared/prisma/migrations/20241024100928_add_prices_table/migration.sql
```diff
@@ -0,0 +1,62 @@
...
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "model_id" TEXT NOT NULL,
+    "usage_type" TEXT NOT NULL,
+    "price" DECIMAL(65,30) NOT NULL,
+
+    CONSTRAINT "prices_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "prices_model_id_idx" ON "prices"("model_id");
```
Steffen911, line 14: FYI: This index is unnecessary as it's completely covered by `prices_model_id_usage_type_key`.

web/src/server/api/routers/models.ts
```diff
@@ -132,19 +132,42 @@ export const modelRouter = createTRPCRouter({
...
+          if (price != null) {
+            await tx.price.create({
+              data: {
+                modelId: createdModel.id,
+                usageType,
+                price,
+              },
+            });
+          }
+        }
```
Steffen911, line 168: Is there a strong reason to do this one by one? A Promise.all would likely return significantly faster here for a longer list and reduce the time in which the transaction is pending.

worker/src/scripts/upsertDefaultModelPrices.ts
```diff
@@ -0,0 +1,192 @@
...
+          where: {
+            projectId: null,
+          },
+          select: {
+            id: true,
+            updatedAt: true,
+          },
+        })
+      ).map((model) => [model.id, model.updatedAt])
+    );
```
Steffen911, line 62: Do you see any issues in doing this part here outside of the update transaction? If multiple workers start at the same time, which would be expected during updates, all of them may read the same initial state and then make updates based on outdated knowledge.

worker/src/scripts/upsertDefaultModelPrices.ts
```diff
@@ -0,0 +1,192 @@
...
+                    defaultModelPrice.tokenizer_config ?? undefined,
+                  tokenizerId: defaultModelPrice.tokenizer_id,
+                  createdAt: defaultModelPrice.created_at,
+                  updatedAt: defaultModelPrice.updated_at,
+                },
+              });
+
+              for (const [usageType, price] of Object.entries(
+                defaultModelPrice.prices
+              )) {
```
Steffen911, line 138: Here, a Promise.all might also help with a shorter transaction duration compared to the for of loop.

worker/src/scripts/upsertDefaultModelPrices.ts
```diff
@@ -0,0 +1,192 @@
...
+                `Error upserting default model ${defaultModelPrice.model_name} (${defaultModelPrice.id}): ${error.message}`,
+                {
+                  error,
+                }
+              );
+            })
+        );
+      }
+
+      await Promise.all(promises);
```
Steffen911, line 175: Promise.allSettled may be more applicable here. That way you do not put the fate of the entire batch on each single operation, but may perform all updates that are valid.

worker/src/scripts/upsertDefaultModelPrices.ts
```diff
@@ -0,0 +1,192 @@
+import { z } from "zod";
```
Steffen911, line 1: Would we need any locking to ensure that only one worker performs those updates at a time? Should be quite easy by setting a timestamp and id in redis on start and unlocking after completion. That way we may avoid that we perform the same update for the same version N times.

web/src/server/api/routers/models.ts
hassiebp: Thanks, yes this is unoptimized since sequential. Performance diffs should be low however as there are only max a handful of prices per model

packages/shared/prisma/migrations/20241024100928_add_prices_table/migration.sql
hassiebp: Yes, thanks for spotting - prices table is small but will drop the unnecessary index :)

worker/src/scripts/upsertDefaultModelPrices.ts
hassiebp: Here I think the additional logic would not be worth the gain - most of the times we would not do any upserting at all due to the initial updated_at check and in case of upserting the upserts are cheap. Do you have a strong opinion here?

worker/src/scripts/upsertDefaultModelPrices.ts
hassiebp: Similar line of thinking as above - even if multiple workers attempt the same updates, they are quite cheap and were not yet worth heavily optimizing. Let me know if you see a clear need for that?

worker/src/scripts/upsertDefaultModelPrices.ts
hassiebp: Yes agree, again did not push for heavy optimization due to only few prices per model

worker/src/scripts/upsertDefaultModelPrices.ts
hassiebp: Each promise/transaction has a catch clause, so `all` and `allSettled` are equivalent as they will all resolve

PR #3848, opened 2024-10-21
feat(comments): add support for comment button on each trace preview in session view

marliessophie:
> [!IMPORTANT]
> Add comment button to each trace in session view, enabling comment interaction and fetching trace comment counts, with updates to components, API, and tests.
> 
>   - **Behavior**:
>     - Add `CommentDrawerButton` to each trace in `SessionPage` in `index.tsx` for comment interaction.
>     - Fetch trace comment counts using `getTraceCommentCountsBySessionId` in `comments.ts`.
>     - Requires user authentication for comment viewing and interaction.
>   - **Components**:
>     - Update `CommentDrawerButton` to accept `className` prop for styling.
>     - Modify `CommentList` to check for user authentication before displaying comments.
>   - **API**:
>     - Add `getTraceCommentCountsBySessionId` to `comments.ts` to retrieve comment counts for traces in a session.
>     - Update `validateCommentReferenceObject` to handle Clickhouse data for `TRACE` and `OBSERVATION` types.
>   - **Tests**:
>     - Rename `comments.servertest.ts` to `async/comments-api.servertest.ts` and update tests to use Clickhouse helpers.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 9d89e2f360f83ca28306718b50ef523ab71290a6. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/sessions.ts
```diff
@@ -285,6 +285,18 @@ export const sessionRouter = createTRPCRouter({
...
+            by: ["objectId"],
+            where: {
+              objectId: { in: session.traces.map((t) => t.id) },
+              projectId: input.projectId,
+              objectType: "TRACE",
+            },
+            _count: {
+              objectId: true,
+            },
+          }),
```
marliessophie, line 299: I'm concerned about added latency, as this query is expensive. I'll add a DD metric prior to merging this code

web/src/server/api/routers/comments.ts
```diff
@@ -288,4 +261,104 @@ export const commentsRouter = createTRPCRouter({
...
+            `;
+
+            const traceIds = new Set(clickhouseTraces.map((t) => t.id));
+            return new Map(
+              allCommentCounts
+                .filter((c) => traceIds.has(c.objectId))
+                .map(({ objectId, count }) => [objectId, Number(count)]),
+            );
+          },
+        });
```
marliessophie, line 355: If one of our clickhouse experts could quickly review this please, that would be very helpful. Just want to make sure my implementation is correct for clickhouse. No need to check postgres, will likely never be used. 

Context: comment table is really small (don't want to share public stats here, will dm if needed)

web/src/features/comments/validateCommentReferenceObject.ts
```diff
@@ -39,27 +48,42 @@ export const validateCommentReferenceObject = async ({
...
-      id: objectId,
-      projectId,
-    },
-  });
+    return !!clickhouseObject
+      ? {}
+      : {
+          errorMessage: `Reference object, ${objectType}: ${objectId} not found in Clickhouse. Skipping creating comment.`,
+        };
+  } else {
```
marliessophie, line 65: If one of our clickhouse experts could quickly review this please, that would be very helpful. Just want to make sure my implementation is correct for clickhouse. Implementation is used for POST comments API and create comments trpc procedure.

web/src/features/comments/validateCommentReferenceObject.ts
```diff
@@ -39,27 +48,42 @@ export const validateCommentReferenceObject = async ({
...
 }): Promise<{ errorMessage?: string }> => {
   const { objectId, objectType, projectId } = input;
-  const prismaModel = COMMENT_OBJECT_TYPE_TO_PRISMA_MODEL[objectType];
 
-  if (!prismaModel) {
-    return {
-      errorMessage: `No prisma model for object type ${objectType}`,
-    };
-  }
+  if (env.CLICKHOUSE_URL && isObservationOrTrace(objectType)) {
```
maxdeichmann, line 52: Is this v3 behavior only? If yes, you can kill the env check.

web/src/features/comments/validateCommentReferenceObject.ts
maxdeichmann: LGTM

web/src/__tests__/async/comments-api.servertest.ts
```diff
@@ -9,34 +9,39 @@ import {
...
+  createTracesCh,
+} from "@/src/__tests__/async/repositories/clickhouse-helpers";
+import {
+  createObservation,
+  createTrace,
+} from "@/src/__tests__/fixtures/tracing-factory";
 
 describe("Create and get comments", () => {
   beforeEach(async () => await pruneDatabase());
   afterEach(async () => await pruneDatabase());
```
maxdeichmann, line 25: This here is why the tests are failing. With this we remove models and prices from the db.

marliessophie: Ah interesting - thank you! 
Do we need models and prices for most tests? If so, we could consider adding a flag to the `pruneDatabase` function that keeps models and prices by default, unless explicitly required them be deleted? Because behaviour in the comments test resulting in a failure of a test in `/public/observations` feels a little off :)

web/src/features/comments/validateCommentReferenceObject.ts
marliessophie: Yes, will not cherry pick for v2. Great call :)

web/src/server/api/routers/comments.ts
```diff
@@ -288,4 +261,104 @@ export const commentsRouter = createTRPCRouter({
...
+                code: "NOT_FOUND",
+                message: "Session not found in project",
+              });
+            }
+
+            const clickhouseTraces = await getTracesIdentifierForSession(
+              input.projectId,
+              input.sessionId,
+            );
+
```
maxdeichmann, line 337: actually, we can improve the above. No need to check for postgres tracesession at all. You can just query clickhouse for all traces in the sessionId.

web/src/server/api/routers/comments.ts
maxdeichmann: Quick q: why did you think that the comment table size matters? We look up CH only by session_id, correct? I dont want to miss something here.

web/src/__tests__/async/comments-api.servertest.ts
maxdeichmann: I think we should not need pruneDatabase in the first place. We could build a helper to ensure prices and models are always there (this is the default in our database at all times in production).

marliessophie: Fair point on not requiring pruneDatabase. I like the idea with a helper to ensure prices and models are always there 🙌

maxdeichmann: There is also a helper to create orgs, projects and api keys one off https://github.com/langfuse/langfuse/blob/2f14d08569e60b2f36d25745d7d0a45811494f93/web/src/__tests__/test-utils.ts#L142

web/src/server/api/routers/comments.ts
marliessophie: Was just an info re the allCommentCounts query, nothing clickhouse specific -- you did not miss anything :)

web/src/server/api/routers/comments.ts
marliessophie: 😍 true, will remove redundant code!

PR #3895, opened 2024-10-24
chore: add background migrations for long-running tasks

Steffen911:
## What does this PR do?

This adds long-running background migrations that we will use for operations that usually won't complete on startup. This includes actions like backfilling certain values for optional columns or migrating data between systems (Postgres to Clickhouse).

We also add four background migrations:
1. Backfill generation cost to complete the switch from compute on read to compute on write (https://github.com/langfuse/langfuse/pull/2466)
2. Move traces from Postgres to Clickhouse
3. Move observations from Postgres to Clickhouse
4. Move scores from Postgres to Clickhouse

## Reviewer notes

I would recommend to start with the new README.md to understand the thought process behind background migrations. From there, the backgroundMigrationManager.ts would be a good next step for the flow. Only then, it may make sense to review the remaining migrations. All of them will not run if we merge this as the background migration environment is off by default. We can run them locally though, as described in the README.

----

> [!IMPORTANT]
> Add support for long-running background migrations with specific implementations for data backfilling and migration from Postgres to Clickhouse.
> 
>   - **Background Migrations**:
>     - Introduces `BackgroundMigration` model in `schema.prisma` and `types.ts`.
>     - Adds `background_migrations` table in `20241024111800_add_background_migrations_table/migration.sql`.
>     - Implements `IBackgroundMigration` interface in `IBackgroundMigration.ts`.
>     - Adds `BackgroundMigrationManager` in `backgroundMigrationManager.ts` to manage migrations.
>   - **Specific Migrations**:
>     - Adds `addGenerationsCostBackfill` migration in `addGenerationsCostBackfill.ts`.
>     - Adds `migrateTracesFromPostgresToClickhouse` in `migrateTracesFromPostgresToClickhouse.ts`.
>     - Adds `migrateObservationsFromPostgresToClickhouse` in `migrateObservationsFromPostgresToClickhouse.ts`.
>     - Adds `migrateScoresFromPostgresToClickhouse` in `migrateScoresFromPostgresToClickhouse.ts`.
>   - **Worker Integration**:
>     - Updates `app.ts` to run background migrations if enabled.
>     - Updates `shutdown.ts` to handle migration shutdown.
>   - **Miscellaneous**:
>     - Removes `observations-backfill-calculated-cost.ts` script from `web/scripts`.
>     - Updates `package.json` to remove `generationCost:backfill` script.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 3fbbec245df02a0aaa28a713bbde2d86ead6489c. It will automatically update as commits are pushed.</sup>


worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
```diff
@@ -0,0 +1,160 @@
...
+    const maxDate = new Date((args.maxDate as string) ?? new Date());
+
+    await addTemporaryColumnIfNotExists();
+
+    let processedRows = 0;
+    while (!this.isAborted && processedRows < maxRowsToProcess) {
+      const observations = await prisma.$queryRaw<
+        Array<Observation>
+      >(Prisma.sql`
+        SELECT id, trace_id, project_id, type, parent_observation_id, start_time, end_time, name, metadata, level, status_message, version, input, output, unit, prompt_id, input_cost, output_cost, total_cost, created_at, updated_at
```
Steffen911: @maxdeichmann I find it hard to map the following fields: `provided_model_name, model_parameters, provided_input_usage_units, provided_output_usage_units, provided_total_usage_units, input_usage_units, output_usage_units, total_usage_units, provided_input_cost, provided_output_cost, provided_total_cost, prompt_name, prompt_version`. Could you provide more context on what they mean and how we want to transition existing data to them?

packages/shared/prisma/schema.prisma
```diff
@@ -2,948 +2,962 @@
...
+  project             Project   @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+  @@index(projectId)
+  @@index(publicKey)
+  @@index(hashedSecretKey)
+  @@index(fastHashedSecretKey)
+  @@map("api_keys")
+}
+
+model BackgroundMigration {
```
maxdeichmann, line 164: maybe we should add a short comment here for the purpose and that this table is not related to any project / org

Steffen911: @maxdeichmann I wonder why we would need to make that explicit? Theoretically, no table at all is org/project specific. They may have a column that indicates org/project membership for a row, but tables itself are always instance-entities.

worker/src/backgroundMigrations/addGenerationsCostBackfill.ts
```diff
@@ -0,0 +1,236 @@
...
+        type: "string",
+        short: "d",
+        default: new Date().toISOString(),
+      },
+    },
+  });
+
+  const migration = new AddGenerationsCostBackfill();
+  await migration.validate(args.values);
+  await migration.run(args.values);
```
maxdeichmann, line 223: Are the implementations above copy-paste from Hassiebs initial solution?

worker/src/backgroundMigrations/backgroundMigrationManager.ts
```diff
@@ -0,0 +1,178 @@
...
+            BackgroundMigrationManager.activeMigration = {
+              id: migration.id,
+              name: migration.name,
+              args: migration.args as any,
+              migration: new (require(`./${migration.script}`).default)(),
+            };
+          },
+          {
+            maxWait: 5000,
+            isolationLevel: Prisma.TransactionIsolationLevel.Serializable,
```
maxdeichmann, line 91: Thanks for thinking of this! Elegant way to solve concurrency.

worker/src/backgroundMigrations/backgroundMigrationManager.ts
```diff
@@ -0,0 +1,178 @@
...
+            // Acquire lock
+            await tx.backgroundMigration.update({
+              where: {
+                id: migration.id,
+              },
+              data: {
+                workerId: BackgroundMigrationManager.workerId,
+                lockedAt: new Date(),
+              },
+            });
```
maxdeichmann, line 78: thought we could also add the migration id to the otel span

worker/src/backgroundMigrations/backgroundMigrationManager.ts
```diff
@@ -0,0 +1,178 @@
...
+      data: {
+        lockedAt: new Date(),
+      },
+    });
+
+    // Schedule next heartbeat in 15s
+    setTimeout(BackgroundMigrationManager.heartBeat, 15 * 1000);
+  }
+
+  public static async run(): Promise<void> {
```
maxdeichmann, line 38: Thought: maybe we could add prefixes to the logs in this class. This way, self hosters can easily filter logs and share with us.

worker/src/backgroundMigrations/backgroundMigrationManager.ts
```diff
@@ -0,0 +1,178 @@
...
+        where: {
+          id: BackgroundMigrationManager.activeMigration.id,
+          workerId: BackgroundMigrationManager.workerId,
+        },
+        data: {
+          lockedAt: null,
+        },
+      });
+      BackgroundMigrationManager.activeMigration = undefined;
+    }
```
maxdeichmann, line 176: should we add a log line here that the active migration was aborted successfully?

worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
```diff
@@ -0,0 +1,165 @@
...
+    await addTemporaryColumnIfNotExists();
+
+    let processedRows = 0;
+    while (!this.isAborted && processedRows < maxRowsToProcess) {
+      const observations = await prisma.$queryRaw<
+        Array<Observation>
+      >(Prisma.sql`
+        SELECT id, trace_id, project_id, type, parent_observation_id, start_time, end_time, name, metadata, level, status_message, version, input, output, unit, completion_start_time, internal_model_id, prompt_id, input_cost, output_cost, total_cost, created_at, updated_at
+        FROM observations
+        WHERE tmp_migrated_to_clickhouse = FALSE AND created_at <= ${maxDate}
```
maxdeichmann, line 64: can we order by time, project_id easily? Time would be most important. Otherwise, we will randomly select data, which might result into adding data to many partitions at once. CH is not good with that.

maxdeichmann: We should do that for the other migrations as well.

worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export default class MigrateObservationsFromPostgresToClickhouse
+  implements IBackgroundMigration
+{
+  private isAborted = false;
+
+  async validate(
+    args: Record<string, unknown>,
+  ): Promise<{ valid: boolean; invalidReason: string | undefined }> {
+    if (!env.CLICKHOUSE_URL) {
```
maxdeichmann, line 37: nice!

worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
maxdeichmann: Ah yes, see my other comment below with the commit. Happy to jump on a call to discuss.

packages/shared/prisma/schema.prisma
maxdeichmann: This was just a thought. Usually all the tables have project_id, org_id. But the readme is very extensive as well. Good enough to keep it the way it is.

worker/src/backgroundMigrations/backgroundMigrationManager.ts
Steffen911: @maxdeichmann IMO that wouldn't be very useful as all of them are a one-time thing. We have the wrapper span for all migrations that we perform and if we link it to logs that should be sufficient.

worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
Steffen911: @maxdeichmann I fully get the partition part and will sort by created_at desc here. Regarding the project_id I'm less sure whether it adds value. My concern is that project_id as a secondary sorting key will introduce more fragmentation and makes the query more expensive. Unless you have a strong reason on Clickhouse side why we should include it I'd prefer to leave it out.

Steffen911: Updated in https://github.com/langfuse/langfuse/pull/3895/commits/f5b3b54ad12139292da4f24f9b382ac90ad263fe.

worker/src/backgroundMigrations/migrateObservationsFromPostgresToClickhouse.ts
```diff
@@ -0,0 +1,165 @@
...
+      `);
+      if (observations.length === 0) {
+        logger.info("No more observations to migrate. Exiting...");
+        break;
+      }
+
+      await clickhouseClient.insert({
+        table: "observations",
+        values: observations.map((observation) => ({
+          ...observation,
```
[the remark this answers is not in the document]
Steffen911: Let's chat about this offline. On the observation side this may need some thought.
Me may also need to check whether our "data quality" script needs any updates after those updates have been applied.

worker/src/backgroundMigrations/backgroundMigrationManager.ts
maxdeichmann: ok makes sense

worker/src/backgroundMigrations/README.md
```diff
@@ -0,0 +1,38 @@
+# Background Migrations
+
+Background migrations are longer running jobs that must not be complete before a new application version can
+be served correctly.
```
marcklingen, line 4: Related, should we add to the requirements that these migrations generally should not be blocking and that the instance needs to be available while the migration is running? It is assumed that the instances continuously receive new traces and frontend events during the migration, which must not lead to an inconsistent state.

marcklingen (commented):
spec and readme looks good, one minor comment regarding requirements

packages/shared/prisma/schema.prisma
```diff
@@ -2,948 +2,962 @@
...
-    withDefaultModels Boolean  @default(true) @map("with_default_models")
-    config            Json?
+  provider          String
+  adapter           String // This controls the interface that is used to connect with the LLM, e.g. 'openai' or 'anthropic'
+  displaySecretKey  String   @map("display_secret_key")
+  secretKey         String   @map("secret_key")
+  baseURL           String?  @map("base_url")
+  customModels      String[] @default([]) @map("custom_models")
+  withDefaultModels Boolean  @default(true) @map("with_default_models")
+  config            Json?
```
hassiebp, line 190: @Steffen911 Do you remember why this was added? 🤔

PR #3944, opened 2024-10-28
refactor: `full-screen-page` to drop magic in height calculation  

marliessophie:
## Description 

Previously we would calculate the available space for the full-screen-page using some fixed expected padding. This introduced tech-debt and does not scale with different forms of page layout. Hence this PR removes the height calculation and relies on built in tailwind-classes instead.


marcklingen (approved):

web/src/components/layouts/layout.tsx
```diff
@@ -575,7 +575,9 @@ export default function Layout(props: PropsWithChildren) {
               </Button>
             </div>
           ) : null}
-          <main className="p-3">{props.children}</main>
+          <main className="h-[calc(100dvh-64px)] p-3 lg:h-dvh">
```
marliessophie, line 579: We need to keep the magic to account for the top langfuse nav bar on mobile only -- I'd be okay with this as we'll be adding the shadcn sidebar soon which will result in a refactor of this file and will put us in a position to remove the magic once and for all :)

PR #3978, opened 2024-10-31
chore: join postgres state into Clickhouse results and add event_ts

Steffen911:
## What does this PR do?

- Joins the postgres record in the ingestion pipeline to create a combination of S3 events, postgres, and Clickhouse.
- Adds the event_ts to the postgres to clickhouse migration fields.
- Calculates the event_ts based on the max of all provided timestamps, unless a Clickhouse value already exists. This allows us to use event_ts's as a sorting key.

----

> [!IMPORTANT]
> Integrates Postgres state with Clickhouse results in the ingestion pipeline, adding conversion functions and updating migrations and tests for data consistency.
> 
>   - **Behavior**:
>     - Joins Postgres records with Clickhouse results in the ingestion pipeline, allowing for a combined view of S3 events, Postgres, and Clickhouse data.
>     - Adds `event_ts` to migration fields from Postgres to Clickhouse, calculated as the max of all timestamps unless a Clickhouse value exists.
>   - **Functions**:
>     - Adds `convertPostgresTraceToInsert`, `convertPostgresObservationToInsert`, and `convertPostgresScoreToInsert` to convert Postgres records to Clickhouse format.
>     - Updates `mergeScoreRecords`, `mergeTraceRecords`, and `mergeObservationRecords` in `IngestionService` to include Postgres records.
>   - **Migrations**:
>     - Updates `migrateObservationsFromPostgresToClickhouse.ts`, `migrateScoresFromPostgresToClickhouse.ts`, and `migrateTracesFromPostgresToClickhouse.ts` to handle new conversion functions and ensure data consistency.
>   - **Tests**:
>     - Adds integration tests in `IngestionService.integration.test.ts` to verify merging of Postgres and event data in Clickhouse.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for ec63b6b77ae03c8c662001e51f392be523919b57. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/definitions.ts
```diff
@@ -172,3 +172,153 @@ export const convertScoreReadToInsert = (
     event_ts: new Date(record.event_ts).getTime(),
   };
 };
+
+/**
+ * Expects a single record from a `select * from traces` query. Must be a raw query to keep original
+ * column names, not the Prisma mapped names.
+ */
+export const convertPostgresTraceToInsert = (
```
Steffen911, line 180: Logic is fully copied from the migration scripts. That way we have the same mapping in the ingestion pipeline.

packages/shared/src/server/definitions.ts
```diff
@@ -172,3 +172,153 @@ export const convertScoreReadToInsert = (
...
+    tags: trace.tags,
+    input: trace.input,
+    output: trace.output,
+    session_id: trace.session_id,
+    created_at:
+      trace.created_at?.toISOString().replace("T", " ").slice(0, -1) ?? null,
+    updated_at:
+      trace.updated_at?.toISOString().replace("T", " ").slice(0, -1) ?? null,
+    event_ts:
+      trace.timestamp?.toISOString().replace("T", " ").slice(0, -1) ?? null,
```
Steffen911, line 209: New is that we also set the event_ts in the migrations here (timestamp for trace, score; start_time for observations).
@maxdeichmann We initially said that we'd just default to 1970-01-01 and overwrite - but per our discussion overwriting is not an option and would produce duplicates instead. Hence, I try to infer the domain timestamp right away to set this.

worker/src/services/IngestionService/index.ts
```diff
@@ -332,13 +373,26 @@ export class IngestionService {
...
+   * as it's not mutable on the clickhouse side.
+   * Otherwise, we use the maximum timestamp. Either it should be provided, or always be the same, and using the
+   * max has the nice property, that we overwrite the default date of 1970-01-01.
+   */
+  private getEventTimestamp(
+    clickhouseTimestamp: number | undefined,
+    otherTimestamps: number[],
+  ): number {
+    return clickhouseTimestamp ?? Math.max(...otherTimestamps);
+  }
```
Steffen911, line 394: @maxdeichmann Any concerns about this approach? This is a slight deviation from what we've discussed, as we're using the max now, but IMO that's safer and shouldn't make a difference for all well-behaved clients.

worker/src/services/IngestionService/index.ts
```diff
@@ -252,61 +280,67 @@ export class IngestionService {
...
 
-    const recordsToMerge = clickhouseObservationRecord
-      ? [clickhouseObservationRecord, ...observationRecords]
-      : observationRecords;
+    // Set clickhouse first as this is the baseline for immutable fields
+    const recordsToMerge = [
+      clickhouseObservationRecord,
+      postgresObservationRecord,
+      ...observationRecords,
+    ].filter(Boolean) as ObservationRecordInsertType[];
```
Steffen911, line 343: @maxdeichmann Regarding the domain timestamps: We always set clickhouse as the first record, postgres as second, and the any other observationRecord (sorted by event type and timestamp). As the domain timestamp is part of the immutableColumn list, we never overwrite what is written in the first entity. I.e. if we have a clickhouse record, we will deterministically use it, otherwise the domain timestamp from postgres and otherwise the one from the earliest event.

Based on my understanding this logic should cover our requrirements.

packages/shared/src/server/definitions.ts
maxdeichmann: very nice!

packages/shared/src/server/definitions.ts
```diff
@@ -172,3 +172,130 @@ export const convertScoreReadToInsert = (
...
+    source: score.source,
+    comment: score.comment,
+    author_user_id: score.author_user_id,
+    config_id: score.config_id,
+    data_type: score.data_type,
+    string_value: score.string_value,
+    created_at: score.created_at?.getTime(),
+    updated_at: score.updated_at?.getTime(),
+    event_ts: score.timestamp?.getTime(),
+    is_deleted: 0,
```
maxdeichmann, line 299: queue_id is missing here, i think

worker/src/services/IngestionService/index.ts
maxdeichmann: Thanks!

packages/shared/src/server/definitions.ts
Steffen911: Added in https://github.com/langfuse/langfuse/pull/3978/commits/42ebd90c136e753633f636956ab4a71d90eb0f21

worker/src/services/IngestionService/index.ts
```diff
@@ -590,24 +622,93 @@ export class IngestionService {
...
+  }) {
+    const recordParser = {
+      traces: convertPostgresTraceToInsert,
+      scores: convertPostgresScoreToInsert,
+      observations: convertPostgresObservationToInsert,
+    };
+    const { projectId, entityId, table } = params;
+
+    return await instrumentAsync(
+      { name: `get-postgres-${table}` },
```
maxdeichmann, line 685: nit: we can remove this as prisma has instrumentation built in.

worker/src/services/IngestionService/index.ts
```diff
@@ -759,7 +852,7 @@ export class IngestionService {
         prompt_version: prompt?.version,
         created_at: Date.now(),
         updated_at: Date.now(),
-        event_ts: new Date(obs.timestamp).getTime(),
+        event_ts: new Date(obs.body?.startTime ?? obs.timestamp).getTime(),
```
maxdeichmann, line 855: good one!

maxdeichmann: above, we are setting the start_time like: 

```
 start_time: this.getMillisecondTimestamp(
          obs.body.startTime ?? obs.timestamp,
        ),

```
I think we should remove this and default to undefined. Can you also check whether this is for scores + traces the same?

worker/src/services/IngestionService/index.ts
Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/3978/commits/4a99a1008368155c3f56844267eb84c9a41c4748

worker/src/services/IngestionService/index.ts
Steffen911: Good catch 😁 Fixed in https://github.com/langfuse/langfuse/pull/3978/commits/7ce2bb895e0b38d4276ab4bf7535869dac3b1369

PR #3989, opened 2024-10-31
feat(multimodal): add media upload / download endpoints

hassiebp:
> [!IMPORTANT]
> Add media upload/download endpoints with S3 integration and update database schema for media management.
> 
>   - **Database Schema**:
>     - Add `Media` and `TraceMedia` tables in `migration.sql` and `schema.prisma`.
>     - Update `types.ts` to include `Media` and `TraceMedia` types.
>   - **API Endpoints**:
>     - Add `GET` and `PATCH` endpoints in `[mediaId].ts` for retrieving and updating media metadata.
>     - Add `POST` endpoint in `upload-url.ts` for generating media upload URLs.
>   - **S3 Integration**:
>     - Update `S3StorageService.ts` to include `getSignedUploadUrl()` for generating S3 upload URLs.
>     - Add `getMediaStorageServiceClient.ts` to manage S3 client instances.
>   - **Validation and Utilities**:
>     - Add `getFileExtensionFromContentType.ts` for mapping content types to file extensions.
>     - Define media-related schemas in `validation.ts`.
>   - **Environment Configuration**:
>     - Update `env.ts` and `env.mjs` to include S3 media upload configurations.
>   - **Testing**:
>     - Add `media.servertest.ts` for end-to-end testing of media upload and download functionalities.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 33377980c68938073ddac9f7f16823319e5c8181. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/migrations/20241106122605_add_media_tables/migration.sql
```diff
@@ -0,0 +1,76 @@
+-- CreateTable
+CREATE TABLE "media" (
+    "id" TEXT NOT NULL,
+    "sha_256_hash" CHAR(44) NOT NULL,
+    "project_id" TEXT NOT NULL,
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "expires_at" TIMESTAMP(3),
```
hassiebp, line 8: Remove expiresAt and deletedAt while it is not used yet. Only add when cleanup job is added

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,111 @@
...
+    responseSchema: z.void(),
+    fn: async ({ query, body, auth }) => {
+      if (auth.scope.accessLevel !== "all") throw new ForbiddenError();
+
+      const { projectId } = auth.scope;
+      const { mediaId } = query;
+      const { uploadedAt, uploadHttpStatus, uploadHttpError } = body;
+
+      try {
+        await prisma.media.update({
```
Comment (hassiebp) [inline web/src/pages/api/public/media/[mediaId].ts:78]: TODO: add observability

web/src/pages/api/public/media/upload-url.ts
```diff
@@ -0,0 +1,225 @@
...
+        observationId,
+        field,
+      } = body;
+
+      if (contentLength > env.LANGFUSE_S3_MEDIA_MAX_CONTENT_LENGTH)
+        throw new InvalidRequestError(
+          `File size must be less than ${env.LANGFUSE_S3_MEDIA_MAX_CONTENT_LENGTH} bytes`,
+        );
+
+      const { mediaId, uploadUrl } = await prisma.$transaction(async (tx) => {
```
hassiebp, line 44: TODO: add observability

packages/shared/prisma/migrations/20241106122605_add_media_tables/migration.sql
```diff
@@ -0,0 +1,76 @@
...
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "media_id" TEXT NOT NULL,
+    "trace_id" TEXT NOT NULL,
+    "observation_id" TEXT NOT NULL,
+    "field" TEXT NOT NULL,
+
+    CONSTRAINT "observation_media_pkey" PRIMARY KEY ("id")
+);
+
```
maxdeichmann, line 47: would it not have been easier to have one table and nullable observation_id? observation_media and trace_media only differentiate by this column.

packages/shared/src/server/services/S3StorageService.ts
```diff
@@ -121,17 +121,20 @@ export class S3StorageService {
...
+    asAttachment: boolean = true,
   ): Promise<string> {
     try {
       return await getSignedUrl(
         this.client,
         new GetObjectCommand({
           Bucket: this.bucketName,
           Key: fileName,
-          ResponseContentDisposition: `attachment; filename="${fileName}"`,
+          ResponseContentDisposition: asAttachment
```
maxdeichmann, line 135: why did you need to change this?

web/src/features/media/validation.ts
```diff
@@ -0,0 +1,92 @@
...
+});
+
+export type GetMediaUploadUrlQuery = z.infer<
+  typeof GetMediaUploadUrlQuerySchema
+>;
+
+export const GetMediaUploadUrlResponseSchema = z.object({
+  uploadUrl: z.string().nullish(),
+  mediaId: z.string(),
+});
```
maxdeichmann, line 62: do we need these responses necessarily in the SDKs? I would try to return as little as possible to SDKs to not have an API dependency here.

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+  LangfuseNotFoundError,
+} from "@langfuse/shared";
+import { Prisma, prisma } from "@langfuse/shared/src/db";
+import { recordIncrement, recordHistogram } from "@langfuse/shared/src/server";
+
+export default withMiddlewares({
+  GET: createAuthedAPIRoute({
+    name: "Get Media data",
+    querySchema: GetMediaQuerySchema,
+    responseSchema: GetMediaResponseSchema,
```
Comment (maxdeichmann) [inline web/src/pages/api/public/media/[mediaId].ts:24]: do we need custom rate limits here?

Comment (maxdeichmann) [reply web/src/pages/api/public/media/[mediaId].ts:24]: I assume this is an add on.

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+      });
+
+      if (!media) throw new LangfuseNotFoundError("Media asset not found");
+      if (!media.uploadHttpStatus)
+        throw new LangfuseNotFoundError("Media not yet uploaded");
+      if (media.uploadHttpStatus !== 200)
+        throw new LangfuseNotFoundError(`Media upload failed`);
+
+      const mediaStorageClient = getMediaStorageServiceClient(media.bucketName);
+      const ttlSeconds = env.LANGFUSE_S3_MEDIA_DOWNLOAD_URL_EXPIRY_SECONDS;
```
Comment (maxdeichmann) [inline web/src/pages/api/public/media/[mediaId].ts:45]: Id prefer to have all expiry / timing env variables in MS to avoid confusion by having it the same everywhere.

packages/shared/src/server/services/S3StorageService.ts
hassiebp: Otherwise file is always downloaded as attachment (as we have with batch exports). For media we want to render inline.

web/src/features/media/validation.ts
hassiebp: Yes, we do. UploadURL necessary to upload the file, mediaID necessary to create the reference string with which e.g. the base64 string is replaced

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+            id: mediaId,
+          },
+          data: {
+            uploadedAt,
+            uploadHttpStatus,
+            uploadHttpError: uploadHttpStatus === 200 ? null : uploadHttpError,
+          },
+        });
+
+        recordIncrement("media.upload_http_status", 1, {
```
Comment (maxdeichmann) [inline web/src/pages/api/public/media/[mediaId].ts:92]: can you prefix all metrics with `langfuse.`? This helps us to differentiate our custom metrics with default ones generated by our infra.

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:24]: Can be added later if necessary, is there a good reason to add them already now?

web/src/pages/api/public/media/upload-url.ts
```diff
@@ -0,0 +1,227 @@
+import { randomUUID } from "crypto";
```
maxdeichmann, line 1: The API is not restful. what about POST api/public/media ?

web/src/pages/api/public/media/upload-url.ts
```diff
@@ -0,0 +1,227 @@
...
+import { instrumentAsync } from "@langfuse/shared/src/server";
+
+export default withMiddlewares({
+  POST: createAuthedAPIRoute({
+    name: "Get Media Upload URL",
+    bodySchema: GetMediaUploadUrlQuerySchema,
+    responseSchema: GetMediaUploadUrlResponseSchema,
+    successStatusCode: 201,
+    fn: async ({ body, auth }) =>
+      instrumentAsync({ name: "create-media-upload-url" }, async () => {
```
maxdeichmann, line 28: Spans are created by Next already. I would only create one if we are interested in timings of sub-parts of this API.

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:45]: 🤔 For expiry etc I would have opted for seconds and named the variable clearly as done above. Expiry TTL by milliseconds does not make much sense imo. If you have a strong opinion that *all* timing vars should be always in MS, let's change

web/src/pages/api/public/media/upload-url.ts
hassiebp: Thanks! Will remove that

web/src/pages/api/public/media/upload-url.ts
```diff
@@ -0,0 +1,227 @@
...
+          observationId,
+          field,
+        } = body;
+
+        if (contentLength > env.LANGFUSE_S3_MEDIA_MAX_CONTENT_LENGTH)
+          throw new InvalidRequestError(
+            `File size must be less than ${env.LANGFUSE_S3_MEDIA_MAX_CONTENT_LENGTH} bytes`,
+          );
+
+        const { mediaId, uploadUrl } = await prisma.$transaction(async (tx) => {
```
maxdeichmann, line 46: why do we need this upsert logic? Would it be enough to just insert? I am worried that we will maintain this table in CH in the future and will run into the same update issues we have right now

web/src/pages/api/public/media/upload-url.ts
```diff
@@ -0,0 +1,227 @@
...
+          const uploadUrl = await s3Client.getSignedUploadUrl({
+            path: bucketPath,
+            ttlSeconds: 60 * 60, // 1 hour
+            sha256Hash,
+            contentType,
+            contentLength,
+          });
+
+          await Promise.all([
+            tx.media.upsert({
```
maxdeichmann, line 138: cant we go to the database only once in this API?
- first we create the S3 files
- afterwards we insert into the database.

web/src/pages/api/public/media/upload-url.ts
maxdeichmann, line 1: We could also use POST api/public/media, right?

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+  }),
+
+  PATCH: createAuthedAPIRoute({
+    name: "Update Media Uploaded At",
+    querySchema: z.object({
+      mediaId: z.string(),
+    }),
+    bodySchema: PatchMediaBodySchema,
+    responseSchema: z.void(),
+    fn: async ({ query, body, auth }) => {
```
Comment (maxdeichmann) [inline web/src/pages/api/public/media/[mediaId].ts:71]: @Steffen911 curious on your opinion here:
I would consider adding all the body params that we use for /api/public/media/upload-url here as well. In this case, if we consider moving this table to CH at some point, we have all values here again and could create another entry in Clickhouse with non nullable columns, without checking the database, and without data modification. 
This way we could track a media-upload-activity table only with changing "status" and additional fields added in this API path.

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:24]: add limits now

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+      if (auth.scope.accessLevel !== "all") throw new ForbiddenError();
+
+      const { projectId } = auth.scope;
+      const { mediaId } = query;
+
+      const media = await prisma.media.findFirst({
+        where: {
+          projectId,
+          id: mediaId,
+        },
```
Comment (Steffen911) [inline web/src/pages/api/public/media/[mediaId].ts:35]: Would it make sense to add an index for this query as well? Right now, it doesn't seem to be covered.

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+      });
+
+      if (!media) throw new LangfuseNotFoundError("Media asset not found");
+      if (!media.uploadHttpStatus)
+        throw new LangfuseNotFoundError("Media not yet uploaded");
+      if (media.uploadHttpStatus !== 200)
+        throw new LangfuseNotFoundError(`Media upload failed`);
```
Comment (Steffen911) [inline web/src/pages/api/public/media/[mediaId].ts:42]: Could be nice to share the upload error here as it's present on the media object anyway.

web/src/server/api/routers/media.ts
```diff
@@ -0,0 +1,74 @@
...
+import { getMediaStorageServiceClient } from "@/src/features/media/server/getMediaStorageClient";
+import {
+  createTRPCRouter,
+  protectedProjectProcedure,
+} from "@/src/server/api/trpc";
+import { logger } from "@langfuse/shared/src/server";
+import { TRPCError } from "@trpc/server";
+
+export const mediaRouter = createTRPCRouter({
+  get: protectedProjectProcedure
```
Steffen911, line 13: Is the trpc router case sensitive? Otherwise, GET may be preferable to be consistent with other records.

web/src/server/api/routers/media.ts
```diff
@@ -0,0 +1,74 @@
...
+import { env } from "@/src/env.mjs";
+import { getMediaStorageServiceClient } from "@/src/features/media/server/getMediaStorageClient";
+import {
+  createTRPCRouter,
+  protectedProjectProcedure,
+} from "@/src/server/api/trpc";
+import { logger } from "@langfuse/shared/src/server";
+import { TRPCError } from "@trpc/server";
+
+export const mediaRouter = createTRPCRouter({
```
Steffen911, line 12: Isn't this exactly the same behaviour as the get in `web/src/pages/api/public/media/[mediaId].ts`? Can we do with only one of the two?

web/src/pages/api/public/media/[mediaId].ts
```diff
@@ -0,0 +1,123 @@
...
+        mediaId,
+        contentType: media.contentType,
+        contentLength: Number(media.contentLength),
+        url,
+        urlExpiry,
+      };
+    },
+  }),
+
+  PATCH: createAuthedAPIRoute({
```
Comment (Steffen911) [inline web/src/pages/api/public/media/[mediaId].ts:64]: Why is there no corresponding POST to this patch? Or is that covered by the `upload-url` endpoint?

web/src/pages/api/public/media/upload-url.ts
Steffen911, line 44: Does everything need to happen within one transaction here? Especially with the presigned link generation we may hold a long lock on the table which may lead to issues for active users.

web/src/pages/api/public/media/[mediaId].ts
Comment (Steffen911) [reply web/src/pages/api/public/media/[mediaId].ts:71]: @maxdeichmann In that case I'd challenge that we use two separate endpoints. We can just use POST and give it an upsert semantic for that.

web/src/server/api/routers/media.ts
hassiebp: This file sneaked in, was not ready yet to be part of this PR :)

web/src/server/api/routers/media.ts
hassiebp: This file sneaked in, was not ready yet to be part of this PR :)

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:24]: will be added in https://linear.app/langfuse/issue/LFE-2682/add-rate-limits-to-media-create-patch

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:35]: The `id` column is unique and already indexed, I added the `projectId` here for access safety only, query should be highly performant without a composite index already :)

web/src/pages/api/public/media/[mediaId].ts
Comment (hassiebp) [reply web/src/pages/api/public/media/[mediaId].ts:64]: We will move the POST to `api/public/media` to return mediaID and upload URL. There will be not `PUT` / `POST` on the mediaID endpoint as the upload goes directly to S3.

What we could do is rather than a POST to `api/public/media` we have a POST to `api/public/media/{mediaId}` but require a valid UUID in the mediaID path.

I prefer the first, let me know if you see good arguments for no 2 🤔

web/src/pages/api/public/media/upload-url.ts
hassiebp: Great point, I think it is indeed wise to split this up 🙏🏾

web/src/pages/api/public/media/[mediaId].ts
Comment (Steffen911) [reply web/src/pages/api/public/media/[mediaId].ts:64]: First one sounds good to me 👍

web/src/pages/api/public/media/index.ts
```diff
@@ -0,0 +1,232 @@
...
+        sha256Hash,
+        contentType,
+        contentLength,
+      });
+
+      return await prisma.$transaction<{
+        mediaId: string;
+        uploadUrl: string;
+      }>(async (tx) => {
+        if (!env.LANGFUSE_S3_MEDIA_UPLOAD_BUCKET)
```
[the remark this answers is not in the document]
hassiebp: ts does not infer that in the tx block

PR #4057, opened 2024-11-05
chore: update cost calculation to handle null values from DB

Steffen911:
## What

Modify the cost and usage schemata to not overwrite `null` with `0` values as those overwrite our calculations.

----

> [!IMPORTANT]
> Update cost calculation logic to handle null values in database schemas and tests, ensuring accurate cost computations.
> 
>   - **Behavior**:
>     - Update `UsageCostSchema` in `definitions.ts` to allow null values and transform them to numbers.
>     - Modify `convertObservationReadToInsert()` in `definitions.ts` to directly assign `provided_usage_details` and `usage_details` without conversion.
>     - Update `verifyClickhouseObservation()` in `verifyClickhouseRecords/index.ts` to handle cost details comparison with null values.
>   - **Tests**:
>     - Add test case in `IngestionService.integration.test.ts` to verify merging observations and calculating costs with null values.
>     - Update existing test cases in `IngestionService.integration.test.ts` to handle null values in usage and cost details.
>   - **Misc**:
>     - Remove `rus-diff` dependency from `package.json`.
>     - Update logging in `verifyClickhouseRecords/index.ts` to remove diff logging.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for fb741e03b0c7e0334ac913f23c56c3b097529421. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/definitions.ts
```diff
@@ -9,8 +9,14 @@ export const clickhouseStringDateSchema = z
...
 // clickhouse returns int64 as string
-export const UsageCostStringSchema = z.record(z.string(), z.string());
-export const UsageCostNumberSchema = z.record(z.string(), z.coerce.number());
+export const UsageCostStringSchema = z.record(
+  z.string(),
+  z.coerce.number().nullish(),
+);
+export const UsageCostNumberSchema = z.record(
+  z.string(),
+  z.coerce.number().nullish(),
```
hassiebp, line 18: question: Why would we need to loosen the constraint here? I think there are great benefits to having all existing keys a number attached rather than needing to check for nullish values

packages/shared/src/server/repositories/definitions.ts
```diff
@@ -9,10 +9,26 @@ export const clickhouseStringDateSchema = z
...
+        if (isNaN(parsed)) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: `Key ${key} is not a number`,
+          });
+        } else {
+          result[key] = parsed;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
maxdeichmann: Expected behavior of this: any string or number inserted into this will pass this function and will be transformed to a number.

worker/src/scripts/verifyClickhouseRecords/index.ts
```diff
@@ -527,7 +523,8 @@ async function verifyClickhouseObservation(postgresObservation: any) {
       case "prompt_tokens": {
         if (
           pgValue !== 0 &&
-          pgValue !== (clickhouseRecord as any)["input_usage_units"]
+          pgValue.toString() !==
```
maxdeichmann, line 526: Why is tostring required here?

packages/shared/src/server/repositories/definitions.ts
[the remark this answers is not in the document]
Steffen911: Correct, and in addition we only retain keys if they have a numeric non-nullish value.

worker/src/scripts/verifyClickhouseRecords/index.ts
Steffen911: Oh that may be obsolete. In the old incarnation zod coerced those values into strings for UInt64. I'll review this script when we run it again.

packages/shared/src/server/repositories/definitions.ts
```diff
@@ -9,10 +9,26 @@ export const clickhouseStringDateSchema = z
 
 //https://clickhouse.com/docs/en/integrations/javascript#integral-types-int64-int128-int256-uint64-uint128-uint256
 // clickhouse returns int64 as string
-export const UsageCostStringSchema = z.record(z.string(), z.string());
-export const UsageCostNumberSchema = z.record(z.string(), z.coerce.number());
-export type UsageCostStringType = z.infer<typeof UsageCostStringSchema>;
-export type UsageCostNumberType = z.infer<typeof UsageCostNumberSchema>;
+export const UsageCostSchema = z
```
hassiebp, line 12:
```typescript
z
  .record(z.string(), z.union([z.string(), z.number()]).nullish())
  .transform((val) =>
    Object.fromEntries(
      Object.entries(val).filter(([_, value]) => value != null),
    ),
  )
  .pipe(z.record(z.string(), z.coerce.number()));
```

PR #4081, opened 2024-11-06
chore: inject eval and annotation scores into clickhouse

Steffen911:
> [!IMPORTANT]
> Injects evaluation scores into Clickhouse by writing scores to S3 and adding them to the ingestion queue, with tests to verify functionality.
> 
>   - **Behavior**:
>     - Injects evaluation scores into Clickhouse by writing scores to S3 and adding them to the ingestion queue in `evalService.ts`.
>     - Adds `source` field to `BaseScoreBody` in `types.ts`, defaulting to `ScoreSource.API`.
>   - **Tests**:
>     - Adds test `evals a valid event and inserts score to ingestion pipeline` in `evalService.test.ts` to verify score ingestion.
>     - Updates `OpenAIServer` in `network.ts` to handle Minio requests for score ingestion.
>   - **Misc**:
>     - Adds `LANGFUSE_S3_EVENT_UPLOAD_ENABLED` to `.env.dev.example` for enabling S3 event uploads.
>     - Renames `schema-utils.ts` to `schemaUtils.ts` in `clickhouse` directory.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 753e9bb940b1a0a3ff643a91f41c711fd2429d8e. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/scores.ts
```diff
@@ -322,6 +328,7 @@ export const scoresRouter = createTRPCRouter({
         },
       });
 
+      // TODO: Can we drop this upsert logic and just error if a score already exists?
```
Steffen911, line 331: @maxdeichmann I think we should drop the `upsert` logic within the create endpoint here. It doesn't seem to be used (only reference in https://github.com/langfuse/langfuse/blob/main/web/src/features/scores/components/AnnotateDrawerContent.tsx#L398) and may produce unintuitive results in my view. 

I can check whether a record already exists for the given Clickhouse key or just create an insert that overwrites whatever is there. The former is probably preferable, but may increase load.

Let me know if you see any reason to keep this bit around and supported, otherwise, I'll go ahead with the removal tomorrow. I also checked with @marliessophie via Slack, but didn't get her assessment yet.

packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -1,34 +1,100 @@
...
+      if (env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET === undefined) {
+        throw new Error("S3 event store is enabled but no bucket is set");
+      }
+      const s3Client = getS3StorageServiceClient(
+        env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET,
+      );
+      await Promise.all(
+        records.map((record) => {
+          s3Client.uploadJson(
+            `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${record.project_id}/${record.id}/${randomUUID()}.json`,
```
maxdeichmann, line 48: `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${authCheck.scope.projectId}/${getClickhouseEntityType(type)}/${eventBodyId}/${key}.json`,

sollten wir den link genau gleich machen wie in der ingestion pipeline?

web/src/server/api/routers/scores.ts
maxdeichmann: I would overwrite any existing score in clickhouse. With whatever should be updated. Hence get from CH and replace. @marliessophie what is your take?

worker/src/features/evaluation/evalService.ts
```diff
@@ -319,18 +342,77 @@ export const evaluate = async ({
...
+  try {
+    if (env.LANGFUSE_S3_EVENT_UPLOAD_ENABLED === "true" && env.CLICKHOUSE_URL) {
+      if (env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET === undefined) {
+        throw new Error("S3 event store is enabled but no bucket is set");
+      }
+      const s3Client = getS3StorageServiceClient(
+        env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET,
+      );
+      await s3Client.uploadJson(
+        `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${event.projectId}/${scoreId}/${scoreId}.json`,
```
maxdeichmann, line 371: also here, not sure the path is correct?

packages/shared/src/server/ingestion/validateAndInflateScore.ts
```diff
@@ -76,7 +76,7 @@ function inflateScoreBody(
   const { body, projectId, scoreId, config } = params;
 
   const relevantDataType = config?.dataType ?? body.dataType;
-  const scoreProps = { ...body, id: scoreId, projectId, source: "API" };
```
Steffen911, line 79: Source is taken either from the body or we default to API via zod schema.

packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -1,34 +1,106 @@
...
+      await Promise.all(
+        opts.records.map((record) => {
+          s3Client.uploadJson(
+            `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${record.project_id}/${getClickhouseEntityType(eventType)}/${record.id}/${randomUUID()}.json`,
+            [
+              {
+                id: randomUUID(),
+                timestamp: new Date().toISOString(),
+                type: eventType,
+                body: record,
```
Steffen911, line 57: @maxdeichmann The body we write to S3 here has the Clickhouse format (i.e. using snake_case). Other events follow the zod schema. I could try to use our generic mapping functions to move back and forth between them here, but this will add more table specific logic to this generic upsert. Do you have a gut feel whether we should handle it in here or do some more explicit mapping on the respective repositories?

web/src/server/api/routers/scores.ts
```diff
@@ -364,6 +355,50 @@ export const scoresRouter = createTRPCRouter({
...
+            `No trace with id ${input.traceId} in project ${input.projectId} in Clickhouse`,
+          );
+        }
+
+        const clickhouseScore = await getScoreByNameAndTraceId(
+          input.projectId,
+          input.name,
+          input.traceId,
+          ScoreSource.ANNOTATION,
+        );
```
marliessophie, line 376: Some additional context: a combined key of `projectId, name, traceId, observationId` OR `projectId, configId, traceId, observationId` can be used to satisfy uniqueness only on scores with source annotation. Right now we are not including observationId in this filter which might lead to unexpected behaviour. Let's take an example. Within a given project, we have traceA which a child generationA. traceA has score1 (name) set with source ANNOTATION. Now we want to set score1 (name) on generationA. This check would find the score1 attached to traceA and falsely flag this as a duplicate. So we'll need to include the observationId in the filter, and make sure the explicitly filter if it should be null (ie we're trying to set a trace score) or if the observation is present (ie we're trying to set an observation score). Does this make sense? :)

worker/src/features/evaluation/evalService.ts
```diff
@@ -319,18 +342,77 @@ export const evaluate = async ({
...
+            timestamp: new Date().toISOString(),
+            type: eventTypes.SCORE_CREATE,
+            body: {
+              ...baseScore,
+              dataType: "NUMERIC",
+            },
+          },
+        ],
+      );
+
```
marliessophie, line 384: Curiosity question: why do we chose to not queue the score create event if we fail to upload the score json to s3? Is this a general requirement we are enforcing across the platform now?

web/src/server/api/routers/scores.ts
```diff
@@ -381,6 +416,41 @@ export const scoresRouter = createTRPCRouter({
...
+          );
+        } else {
+          await upsertScore({
+            id: input.id,
+            project_id: input.projectId,
+            timestamp: convertDateToClickhouseDateTime(
+              clickhouseScore.timestamp,
+            ),
+            value: input.value !== null ? input.value : undefined,
+            string_value: input.stringValue,
```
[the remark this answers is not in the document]
Steffen911: @marliessophie It's a typescript thing :) `value` expects `string | undefined` and the input is `string | null | undefined`. Hence, I added this workaround to cast nulls to undefineds. For string_value this doesn't seem necessary.
Expected behaviour is exactly the same.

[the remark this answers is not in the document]
marliessophie: Cool, thanks for letting me know, interesting that string value is happy with null and value is not :)

worker/src/features/evaluation/evalService.ts
marliessophie: Awesome, thanks for the detailed context! 🙌 Makes sense

web/src/server/api/routers/scores.ts
Steffen911: @marliessophie Could you validate whether https://github.com/langfuse/langfuse/pull/4081/commits/f79bc50ef26cf543cc71ae04eb975d4271cabc3a covers the requirements that you've outlined?

My logic going forward:

- Search for projectId, name and (if provided) observationId
- _AND_ in addition for a match of either name _or_ config.

Since name is required according to the input object we shouldn't throw the error in the util function in any case, but receive an object if there is any match. In that case we reject the creation.

PR #4086, opened 2024-11-06
feat: support evals triggered via dataset run  

marliessophie:
> [!IMPORTANT]
> Add support for evaluations triggered by dataset runs, updating models, queues, evaluation logic, UI, and API.
> 
>   - **Behavior**:
>     - Support evaluations triggered by dataset runs.
>     - Add `job_input_observation_id` and `job_input_dataset_item_id` to `JobExecution`.
>     - Create migration `20241107165056_job_executions_add_observation_and_dataset_item`.
>   - **Models and Schema**:
>     - Update `Observation` and `DatasetItem` models in `schema.prisma`.
>     - Update `JobExecution` model with new fields and relations.
>   - **Queue and Event Handling**:
>     - Add `DatasetRunItemUpsertEventSchema` and `UpsertEventSchema` in `queues.ts`.
>     - Implement `DatasetRunItemUpsertQueue` in `datasetRunItemUpsert.ts`.
>     - Modify `evalJobCreatorQueueProcessor` in `evalQueue.ts` to handle dataset events.
>   - **Evaluation Logic**:
>     - Update `createEvalJobs` and `evaluate` functions in `evalService.ts` to support dataset items.
>     - Add logic to extract variables from dataset items in `extractVariables` function.
>   - **UI and API**:
>     - Update `EvaluatorForm` and `EvaluatorTable` components to support dataset evaluations.
>     - Modify `evalRouter` in `router.ts` to handle new dataset evaluation configurations.
>   - **Testing**:
>     - Add tests for dataset evaluation in `evalService.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 34d0d7902e966ddfe7897455c80ca9ccf3c90ada. It will automatically update as commits are pushed.</sup>

----

TODO: 

- [x] Observability for `DatasetRunItemUpsert` Queue --> QQ: does the `metricWrapper` already handle this for us? 
- [x] Add env variable `QUEUE_CONSUMER_DATASET_RUN_ITEM_UPSERT_QUEUE_IS_ENABLED`


packages/shared/prisma/schema.prisma
```diff
@@ -898,6 +900,12 @@ model JobExecution {
   jobInputTraceId String? @map("job_input_trace_id")
   trace           Trace?  @relation(fields: [jobInputTraceId], references: [id], onDelete: SetNull) // job remains when traces are deleted
 
+  jobInputObservationId String?      @map("job_input_observation_id")
+  observation           Observation? @relation(fields: [jobInputObservationId], references: [id], onDelete: SetNull) // job remains when traces are deleted
+
+  jobInputDatasetItemId String?      @map("job_input_dataset_item_id")
+  datasetItem           DatasetItem? @relation(fields: [jobInputDatasetItemId, projectId], references: [id, projectId], onDelete: Cascade) // job is removed when dataset run items are deleted
```
marliessophie, line 907: Given that the `DatasetItem` Model is requiring the combined key of `id` and `projectId` to be unique, we are forced to cascade `onDelete` here as we cannot and do not want to set `projectId` to `NULL` (SetNull [definiton](https://www.prisma.io/docs/orm/prisma-schema/data-model/relations/referential-actions#setnull): "when the referenced record is deleted, set these fields to NULL").  

Is there any technical issue/limitation with cascading on delete? If not, I'd prefer to keep the implementation as is. Thanks!

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -93,6 +105,41 @@ export default withMiddlewares({
...
+              timestamp: new Date(),
+              name: QueueJobs.DatasetRunItemUpsert as const,
+            },
+            {
+              attempts: 3, // retry 3 times
+              backoff: {
+                type: "exponential",
+                delay: 1000,
+              },
+              delay: 180_000, // 180 seconds
```
marliessophie, line 135: The schema of the `jobExecution` has a fk relationship enforced on `traceId`. I added a generous delay to ensure we do not attempt to create a dataset eval job execution before the trace has been created via the ingestion queue. Do you have any insight here how to make this approach more robust? Once added to the queue, we validate the existence of the trace, and retry if it's not found.

web/src/ee/features/evals/components/evaluator-form.tsx
```diff
@@ -464,21 +533,44 @@ export const InnerEvalConfigForm = (props: {
...
+                        This will run on all future traces that match these
+                        filters
+                      </FormDescription>
+                      <FormMessage />
+                    </>
+                  ) : (
+                    <>
+                      <FormControl>
+                        <InlineFilterBuilder
+                          columns={datasetTableColsWithOptions(
```
maxdeichmann, line 559: image

You are probably aware, but I do not know what exactly i am selecting here.

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -23,6 +27,10 @@ export default withMiddlewares({
         metadata,
       } = body;
 
+      /**************
```
maxdeichmann, line 30: Nit: I think instead of having these headers, we could also move each part into functions that we call in here.

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -93,6 +105,41 @@ export default withMiddlewares({
...
+              timestamp: new Date(),
+              name: QueueJobs.DatasetRunItemUpsert as const,
+            },
+            {
+              attempts: 3, // retry 3 times
+              backoff: {
+                type: "exponential",
+                delay: 1000,
+              },
+              delay: 90000, // 90 seconds
```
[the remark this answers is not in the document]
maxdeichmann: This FK needs to go away for V3 anyways. I think removing this would be an option. Can you check whether this breaks anything in the eval queue downstream (also for the current evals)

web/src/ee/features/evals/components/evaluator-form.tsx
marliessophie: Thanks for flagging -- open ticket for me to replace the component such that it can support key value pairs as the data model requires the IDs of the datasets whereas we need to show their respective names to the user in the UI :)

packages/shared/prisma/schema.prisma
```diff
@@ -386,7 +386,8 @@ model Observation {
   completionStartTime DateTime? @map("completion_start_time")
   project             Project   @relation(fields: [projectId], references: [id], onDelete: Cascade)
 
-  promptId String? @map("prompt_id") // no fk constraint, prompt can be deleted
+  promptId     String?        @map("prompt_id") // no fk constraint, prompt can be deleted
+  jobExecution JobExecution[]
```
maxdeichmann, line 390: We cannot have this FK anymore as this will not work in CH.

packages/shared/prisma/schema.prisma
maxdeichmann: Why would it set the project_id to null? We have the FK to project_id. If possible, I would keep the execution if we delete a dataset item.

packages/shared/src/server/queues.ts
```diff
@@ -65,7 +66,19 @@ export const BatchExportJobSchema = z.object({
 export const TraceUpsertEventSchema = z.object({
   projectId: z.string(),
   traceId: z.string(),
+  type: z.literal("trace"),
```
maxdeichmann, line 69: This here will break production. Currently, we have events in the queue without the type. If we deploy a worker which requires a type, this will fail.
Proposal: i think we should move the trace and dataset upsert event through their own queue.

packages/shared/src/server/queues.ts
```diff
@@ -120,6 +138,12 @@ export type TQueueJobTypes = {
     payload: TraceUpsertEventType;
     name: QueueJobs.TraceUpsert;
   };
+  [QueueName.DatasetRunItemUpsert]: {
+    timestamp: Date;
+    id: string;
+    payload: DatasetRunItemUpsertEventType;
+    name: QueueJobs.DatasetRunItemUpsert;
+  };
```
maxdeichmann, line 146: Here we define a new queue for upserts already

packages/shared/src/tableDefinitions/tracesTable.ts
```diff
@@ -138,17 +138,44 @@ export const tracesTableCols: ColumnDefinition[] = [
   },
 ];
 
-export const evalTableCols: ColumnDefinition[] = tracesOnlyCols;
+// Used only for dataset evaluator, not on dataset table
+export const datasetOnlyCols: ColumnDefinition[] = [
```
maxdeichmann, line 142: in which UI elements is this added? This is potentially breaking with V3 in case we require a join observations x dataset

web/src/server/api/services/tableDefinitions.ts
```diff
@@ -273,4 +273,16 @@ export const tableDefinitions: TableDefinitions = {
       traceTags,
     ],
   },
+
+  dataset_items: {
+    table: ` dataset_items di`,
```
maxdeichmann, line 278: same here, which UI tables require this?

worker/src/features/evaluation/evalService.ts
```diff
@@ -53,24 +58,17 @@ const getS3StorageServiceClient = (bucketName: string): S3StorageService => {
   return s3StorageServiceClient;
 };
 
-// this function is used to determine which eval jobs to create for a given trace
-// there might be multiple eval jobs to create for a single trace
-export const createEvalJobs = async ({
+const query = kyselyPrisma.$kysely.selectFrom("job_configurations").selectAll();
+
+type JobConfigurationsRaw = InferResult<typeof query>;
```
maxdeichmann, line 63: Nit: can we take this type (https://github.com/langfuse/langfuse/blob/7d753a0930451e8c79e372e4d436e7b0a68c962b/packages/shared/prisma/generated/types.ts#L270)

worker/src/features/evaluation/evalService.ts
```diff
@@ -195,6 +201,217 @@ export const createEvalJobs = async ({
...
+
+    // safeguard only, this should never happen
+    if (config.target_object !== "dataset") {
+      logger.debug(
+        `Skipping eval config ${config.id} as target object is not dataset run item`,
+      );
+      continue;
+    }
+
+    logger.info("Creating eval job for config", config.id);
```
maxdeichmann, line 229: If we bring this live, we need to exclude these logs from the DD log index? https://github.com/langfuse/infrastructure/blob/388b4a65de116742eb08a979dfb9ea58301b36d5/terraform/org/infrastructure/datadog_log_archive.tf#L3

worker/src/features/evaluation/evalService.ts
```diff
@@ -195,6 +201,217 @@ export const createEvalJobs = async ({
...
+          },
+          delay: config.delay, // milliseconds
+          removeOnComplete: true,
+          removeOnFail: 1_000,
+        },
+      );
+    } else {
+      // if we do not have a match, and execution exists, we mark the job as cancelled
+      // we do this, because a second trace event might 'deselect' a trace
+      logger.info(`Eval job for config ${config.id} did not match trace`);
```
maxdeichmann, line 366: this should be an error

packages/shared/prisma/schema.prisma
marliessophie: My understanding: 
- ideal behaviour: When deleting a `DatasetItem`, the `jobInputDatasetItemId` will be set to NULL for all created execution jobs. 
- approach 1, given `@relation(fields: [jobInputDatasetItemId, projectId]`: set `onDelete` to `SetNull` --> this will never work as `SetNull` will attempt to null the `jobInputDatasetItemId` (fine) AND the `projectId` on the `jobExecution` object 
- approach 2, assume we change `@relation(fields: [jobInputDatasetItemId, projectId]` to `@relation(fields: [jobInputDatasetItemId]`--> this throws an error ceteris paribus as the argument `references` must refer to a unique criterion in the related model. The reason we are seeing this error is because we require the combined key `id, projectId` to be unique on the `DatasetItem`. 

So the trade-off we are facing: 
- Option A: adjust the schema of the `DatasetItem` such that the field `id` is `@unique` in the model `DatasetItem` and unlock approach 2 above 
- Option B: accept cascading the job on delete of the `DatasetItem` and leave the `DatasetItem` in tact

I would prefer not extending the scope of this change further, opting for Option B, and my key question to you is do you see a strong case for Option A over Option B?

web/src/pages/api/public/dataset-run-items.ts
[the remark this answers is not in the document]
marliessophie: Will do, sounds good 🙌

packages/shared/src/server/queues.ts
marliessophie: uuups, good catch, consider it fixed 🙌

packages/shared/src/tableDefinitions/tracesTable.ts
marliessophie: Only used for the dataset evaluator form, no tables as per comment above. No join required :)

web/src/server/api/services/tableDefinitions.ts
marliessophie: Same as above, will move to different file to make clearer this is not related to tables

worker/src/features/evaluation/evalService.ts
marliessophie: This has been live since we released model based evals... And can you give me context on why we'd need to exclude these logs from the DD index?

PR #4127, opened 2024-11-08
chore: trace merge reproduction

Steffen911:


worker/src/services/IngestionService/tests/IngestionService.integration.test.ts
```diff
@@ -661,6 +661,78 @@ describe("Ingestion end-to-end tests", () => {
...
+      }),
+    ]);
+
+    await Promise.all([
+      ingestionService.processTraceEventList({
+        projectId,
+        entityId: traceId,
+        traceEventList: traceEventList1,
+      }),
+    ]);
```
Steffen911, line 727: Note that there is _no_ flush in between.

My hypothesis is that traceEventList2 got processed and sets the larger event_ts. Afterwards, traceEventList1 gets processed, merges everything together, and writes another row. This row gets removed because the toDate(timestamp), project_id, and event_id is the same and the first, incomplete event has a higher event_ts.

PR #4196, opened 2024-11-12
chore: migrate score analytics dashboard to clickhouse

Steffen911:
> [!IMPORTANT]
> Migrate score analytics dashboard to Clickhouse by adding new query functions, updating the dashboard router, and modifying UI components to use Clickhouse data.
> 
>   - **Behavior**:
>     - Add `getNumericScoreTimeSeries` and `getCategoricalScoreTimeSeries` in `dashboards.ts` for querying score time series data from Clickhouse.
>     - Add `getNumericScoreHistogram` in `scores.ts` for querying numeric score histograms from Clickhouse.
>     - Update `dashboard-router.ts` to handle new query types `numeric-score-time-series` and `categorical-score-chart`.
>   - **UI Components**:
>     - Update `CategoricalScoreChart.tsx`, `NumericScoreHistogram.tsx`, and `NumericScoreTimeSeriesChart.tsx` to use Clickhouse queries.
>   - **Models**:
>     - Extend `dashboardColumnDefinitions` in `mapDashboards.ts` to include score-related columns.
>   - **Misc**:
>     - Add `resultPG.json` for testing or reference data.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 78803e12398cc61a44d98e3915d215559e967b76. It will automatically update as commits are pushed.</sup>


web/src/features/dashboard/lib/score-analytics-utils.ts
```diff
@@ -108,8 +108,10 @@ function groupCategoricalScoreDataByTimestamp(
 ): Record<string, DatabaseRow[]> {
   return data.reduce(
     (acc, row) => {
-      const timestamp = row[scoreTimestampAccessor];
-      if (!(timestamp instanceof Date)) return acc;
+      if (row[scoreTimestampAccessor] === null) {
+        return acc;
+      }
+      const timestamp = new Date(row[scoreTimestampAccessor] ?? new Date());
```
Steffen911, line 114: @marliessophie It seems that clickhouse doesn't return the metadata that trpc needs to auto-cast date columns to a Date type. Instead, I try to cast any string here as long as there is a value set.
Do you have any concerns about this approach? It looks like the function is only used on the dashboard and there it seems to work fine.

Steffen911: Clarified offline - change should be good.

PR #4209, opened 2024-11-13
chore: apply trace timestamp filter correctly on user model query

Steffen911:
> [!IMPORTANT]
> Fix timestamp filter in `getModelUsageByUser` by checking `field.includes("start_time")` and using `DateTime64(3)`.
> 
>   - **Behavior**:
>     - In `getModelUsageByUser`, change `f.field === "start_time"` to `f.field.includes("start_time")` to correctly apply timestamp filters.
>     - Update timestamp format to `DateTime64(3)` in `getModelUsageByUser` for accurate filtering.
>   - **Misc**:
>     - Minor formatting changes in SQL query for readability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 370742d084c0a0a47df355263910796fa9ca4b91. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/repositories/dashboards.ts
+++ b/packages/shared/src/server/repositories/dashboards.ts
@@ -8,6 +8,7 @@ import {
 import { dashboardColumnDefinitions } from "../../tableDefinitions/mapDashboards";
 import { convertDateToClickhouseDateTime } from "../clickhouse/client";
 import {
+  OBSERVATIONS_TO_TRACE_INTERVAL,
   SCORE_TO_TRACE_OBSERVATIONS_INTERVAL,
   TRACE_TO_OBSERVATIONS_INTERVAL,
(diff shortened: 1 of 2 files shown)


packages/shared/src/server/repositories/dashboards.ts
```diff
@@ -318,7 +318,7 @@ export const getModelUsageByUser = async (
   const timeFilter = chFilter.find(
     (f) =>
       f.clickhouseTable === "observations" &&
-      f.field === "start_time" &&
+      f.field.includes("start_time") &&
```
Steffen911, line 321: Field is set to o.start_time. With this approach we only check if something like it is present.

PR #4313, opened 2024-11-19
fix: dataset item api fix for large dataset runs

maxdeichmann:
> [!IMPORTANT]
> Enhances dataset run handling for large datasets using Clickhouse with new schema, functions, and tests.
> 
>   - **Behavior**:
>     - Introduces `datasetRunsTableSchema` in `service.ts` for input validation.
>     - Updates `runsByDatasetId` in `dataset-router.ts` to use `datasetRunsTableSchema`.
>     - Adds `createDatasetRunsTable` and `createTempTableInClickhouse` in `service.ts` for handling large dataset runs.
>     - Implements `getDatasetRunsFromPostgres` in `service.ts` to fetch dataset runs with pagination.
>   - **Functionality**:
>     - Renames `upsertClickhouse` to `upsertClickhouseWithS3` in `clickhouse.ts`, `scores.ts`, and `traces.ts`.
>     - Modifies `clickhouseClient` to be a function returning a client instance in `client.ts`.
>     - Replaces `randomCharacters` with `clickhouseCompliantRandomCharacters` in `clickhouse-filter.ts`.
>   - **Testing**:
>     - Adds `dataset-service.servertest.ts` to test dataset run data fetching.
>     - Updates `clickhouse-helpers.ts` to use the new `clickhouseClient` function.
>   - **Misc**:
>     - Fixes `clickhouseClient.close()` calls to `clickhouseClient().close()` in `seed.ts`, `load-seed.ts`, and `prepareClickhouse.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 02adfef0d50da2575e49aff9028fd09f23525b1c. It will automatically update as commits are pushed.</sup>


web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,332 @@
...
+
+export type DatasetRunsTableInput = z.infer<typeof datasetRunsTableSchema>;
+
+// we might have many traces / observations in Postgres which belong to data in clickhouse.
+// We need to create a temp table in CH, dump the data in there, and then join in CH.
+export const createDatasetRunsTable = async (input: DatasetRunsTableInput) => {
+  const tableName = `dataset_runs_${clickhouseCompliantRandomCharacters()}`;
+  const clickhouseSession = v4();
+  try {
+    const runs = await getDatasetRunsFromPostgres(input);
```
maxdeichmann, line 48: get all runs + run items from CH

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,332 @@
...
+    const obsAgg = await getObservationLatencyAndCostForDataset(
+      input,
+      tableName,
+      clickhouseSession,
+    );
+    const traceAgg = await getTraceLatencyAndCostForDataset(
+      input,
+      tableName,
+      clickhouseSession,
+    );
```
maxdeichmann, line 75: For each run, we need to get aggregate data:
- average cost
- average latency

dataset items are linked to traces or observations. Depending on that, we need to get the average from all trace or observation latencies / cost

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,332 @@
...
+    format: "JSONEachRow",
+  });
+};
+
+export const createTempTableInClickhouse = async (
+  tableName: string,
+  clickhouseSession: string,
+) => {
+  const query = `
+      CREATE TEMPORARY TABLE IF NOT EXISTS ${tableName}
```
maxdeichmann, line 144: we cannot use a param here. Hence, table name must not be editable by our users.

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+      tableName,
+      clickhouseSession,
+    );
+    const traceAgg = await getTraceLatencyAndCostForDataset(
+      input,
+      tableName,
+      clickhouseSession,
+    );
+
+    await deleteTempTableInClickhouse(tableName);
```
Steffen911, line 77: You could move this into a `finally` block to handle it once and not in the main and the catch path.

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+    clickhouseConfigs: { session_id: clickhouseSession },
+  });
+};
+
+export const deleteTempTableInClickhouse = async (tableName: string) => {
+  const query = `
+      DROP TABLE IF EXISTS ${tableName}
+  `;
+  await commandClickhouse({ query, params: { tableName } });
+};
```
Steffen911, line 164: No session_id here? How do we ensure it gets routed to the same node?

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+          'observation_id', ri.observation_id,
+          'ri_id', ri.id
+        )) AS run_items
+      FROM
+        datasets d
+        JOIN dataset_runs runs ON d.id = runs.dataset_id AND d.project_id = runs.project_id
+        JOIN dataset_run_items ri ON ri.dataset_run_id = runs.id
+          AND ri.project_id = runs.project_id
+      WHERE
+        d.id = ${input.datasetId}
```
Steffen911, line 189: This prisma syntax still freaks me out a bit. There is no need to wrap those inputs again in any way, right?

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+  tableName: string,
+  clickhouseSession: string,
+) => {
+  const query = `
+      WITH agg AS (
+      SELECT
+        run_id,
+        dateDiff('milliseconds', start_time, end_time) AS latency_ms,
+        total_cost AS cost
+      FROM ${tableName} tmp JOIN observations o
```
Steffen911, line 243: Does it make a difference whether we join the tmp table from right? That way we may avoid an in-memory hashmap of the observations table.

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+        dateDiff('milliseconds', start_time, end_time) AS latency_ms,
+        total_cost AS cost
+      FROM ${tableName} tmp JOIN observations o 
+        ON tmp.project_id = o.project_id 
+        AND tmp.observation_id = o.id 
+        AND tmp.trace_id = o.trace_id
+      WHERE o.project_id = {projectId: String}
+      AND tmp.project_id = {projectId: String}
+      AND tmp.dataset_id = {datasetId: String}
+      AND tmp.observation_id IS NOT NULL
```
Steffen911, line 250: Any chance we can filter by start_time/type here?

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+  clickhouseSession: string,
+) => {
+  const query = `
+      WITH agg AS (
+      SELECT
+        o.trace_id,
+        run_id,
+        dateDiff('milliseconds', min(start_time), max(end_time)) AS latency_ms,
+        sum(total_cost) AS cost
+      FROM ${tableName} tmp JOIN observations o
```
Steffen911, line 294: Same here as above.

web/src/features/datasets/server/service.ts
```diff
@@ -0,0 +1,329 @@
...
+      input,
+      tableName,
+      clickhouseSession,
+    );
+
+    await deleteTempTableInClickhouse(tableName);
+
+    const enrichedRuns = runs.map(({ run_items, ...run }) => {
+      const observation = obsAgg.find((o) => o.runId === run.run_id);
+      const trace = traceAgg.find((t) => t.runId === run.run_id);
```
Steffen911, line 81: Depending on the cardinality from obsAgg and traceAgg it might be more efficient to create an in-memory map first instead of searching the list each time.

web/src/features/datasets/server/service.ts
maxdeichmann: No, not in this case. This will be sent as a SQL object as prepared statement. If I would have used $queryRawUnsafe, this would be an issue.

web/src/features/datasets/server/service.ts
maxdeichmann: thx!

web/src/features/datasets/server/service.ts
maxdeichmann: thx

PR #4319, opened 2024-11-19
chore: use clickhouse in evaluations queues

Steffen911:
> [!IMPORTANT]
> Integrates ClickHouse into evaluations queue, replacing Redis for event processing and storage, and updates related functions and tests.
> 
>   - **Behavior**:
>     - Replaces Redis with ClickHouse for storing and processing trace and observation events in `evalService.ts` and `IngestionService/index.ts`.
>     - Updates `createEvalJobs()` and `evaluate()` in `evalService.ts` to use ClickHouse for event existence checks and data retrieval.
>     - Removes `addTracesToTraceUpsertQueue()` from `processEventBatch.ts` and `legacy/index.ts`.
>   - **Queues**:
>     - Increases `attempts` and adds `delay` for `DatasetRunItemUpsertQueue`, `EvalExecutionQueue`, and `TraceUpsertQueue`.
>     - Adds trace to `TraceUpsertQueue` in `IngestionService/index.ts` for eval processing.
>   - **Tests**:
>     - Updates tests in `evalService.test.ts` to use ClickHouse for data setup and verification.
>     - Removes `eventsApi.test.ts` as it is no longer relevant.
>   - **Environment**:
>     - Adds `LANGFUSE_RETURN_FROM_CLICKHOUSE` to `env.ts` to toggle between Redis and ClickHouse.
>   - **Misc**:
>     - Updates `getObservationById()` and `getTraceById()` to fetch data from ClickHouse in `observations.ts` and `traces.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 99f868f3ec4bb51931ce1bece00048e3274366c7. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/observations.ts
```diff
@@ -38,6 +38,30 @@ import {
   TRACE_TO_OBSERVATIONS_INTERVAL,
 } from "./constants";
 
+export const checkObservationExists = async (
```
Steffen911, line 41: Can we add a start_time filters by using the assumption that evals happen within a couple hours of receiving a trace/observation?

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -106,6 +109,7 @@ export default withMiddlewares({
        * ASYNC RUN ITEM EVAL *
        ********************/
 
+      // TODO: Move this check to TraceUpsert queue?
```
Steffen911, line 112: Pending feedback from Marlies.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -73,6 +72,48 @@ export const getTracesTableCount = async (props: {
...
+      operator: "=",
+      value: traceId,
+    }),
+  );
+
+  const tracesFilterRes = tracesFilter.apply();
+
+  const query = `
+    SELECT id, project_id
+    FROM traces t FINAL
```
Steffen911, line 99: Using FINAL here should be acceptable, because we do not rely on any skip indexes and always provide a timestamp with the current usage pattern.

web/src/ee/features/evals/server/addDatasetRunItemsToEvalQueue.ts
```diff
@@ -18,30 +21,17 @@ export const addDatasetRunItemsToEvalQueue = async ({
...
-        {
-          attempts: 5, // retry 3 times
-          backoff: {
-            type: "exponential",
-            delay: 1000,
-          },
-          delay: 30000, // 10 seconds
-          removeOnComplete: true,
-          removeOnFail: 1_000,
-        },
```
Steffen911, line 43: We fill fallback to the defaults on the queue. No need to overwrite for each item.

worker/src/__tests__/evalService.test.ts
```diff
@@ -437,7 +488,10 @@ describe("create eval jobs", () => {
       traceId: traceId,
     };
 
-    await createTraceEvalJobs({ event: payload });
+    await createEvalJobs({ event: payload });
+
+    // Wait for .5s
+    await new Promise((resolve) => setTimeout(resolve, 500));
```
Steffen911, line 494: This ensures that we actually have a higher event_ts for the new trace.

worker/src/__tests__/evalService.test.ts
```diff
@@ -176,8 +215,103 @@ describe("eval service tests", () => {
...
+        .selectAll()
+        .where("project_id", "=", "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a")
+        .execute();
+
+      expect(jobsAfterTrace.length).toBe(1);
+      expect(jobsAfterTrace[0].project_id).toBe(
+        "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a",
+      );
+      expect(jobsAfterTrace[0].job_input_trace_id).toBe(traceId);
+      expect(jobsAfterTrace[0].job_input_dataset_item_id).toBe(datasetItemId);
```
Steffen911, line 309: @marliessophie Do we rely anywhere on the job_input_dataset_item_id being present? With my current logic, we would not set it if the jobexecution was triggered by the TraceUpsert.

If we need it, I could additionally try to fetch the corresponding datasetItemId whenever the config.target_object === 'dataset'. This shouldn't be an issue, but I would first want to now if it makes a difference.

.github/workflows/pipeline.yml
```diff
@@ -326,9 +326,15 @@ jobs:
       - name: install dependencies
         run: |
           pnpm install
+      - name: Install golang-migrate for Clickhouse migrations
```
maxdeichmann, line 329: this is on main already :)
Haha i was struggling with this one 2 days ago.

packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -51,6 +47,13 @@ export async function upsertClickhouse<
       );
       await Promise.all(
         opts.records.map((record) => {
+          // drop trailing s and pretend it's always a create.
```
maxdeichmann, line 50: why do we need this change?

Steffen911: @maxdeichmann I want to derive a valid event to store in S3 based on the ClickHouse table. For traces and scores that's easy as they become `trace-create` and `score-create`. For observations we would want `span-create`, `generation-create`, and `event-create` though. Hence, I had to become a bit more creative here.

worker/src/env.ts
```diff
@@ -91,6 +91,8 @@ const EnvSchema = z.object({
     .default(5),
   STRIPE_SECRET_KEY: z.string().optional(),
 
+  LANGFUSE_RETURN_FROM_CLICKHOUSE: z.enum(["true", "false"]).default("false"),
```
Steffen911, line 94: TODO: We need to set this on the worker in our infra repo.

worker/src/__tests__/evalService.test.ts
marliessophie: Hey @Steffen911, the `job_input_dataset_item_id` is optional in the schema, as are the `job_input_trace_id` and `job_input_observation_id`. The code needs to handle all of these possibly being undefined in general. 

However, in the case where the `config.target_object === 'dataset'` AND we are using properties of the dataset item to map into the prompt variables (e.g. dataset_item.input --> {{variable1}}), we would end up skipping mapping those variables ([see code](https://github.com/langfuse/langfuse/blob/8517670ac2f7b56c5a832ef79a9ba46192efe715/worker/src/features/evaluation/evalService.ts#L706-L748)), and would replace them with an empty string instead, which would harm the quality of the eval and suggest a user error where there was one. Fetching the `datasetItemId` in this case seems like the right course of action, would you agree given the context? 

Thanks for double checking!

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -49,14 +51,20 @@ export default withMiddlewares({
...
-        const observation = observationId
-          ? await prisma.observation.findUnique({
-              where: {
-                id: observationId,
-                projectId: auth.scope.projectId,
-              },
-            })
-          : undefined;
+        const observation =
+          env.LANGFUSE_RETURN_FROM_CLICKHOUSE === "true"
```
marliessophie, line 55: Nit, global: we could extract a small helper that returns if env variables are true or false -- less error prone generally

worker/src/features/evaluation/evalService.ts
```diff
@@ -87,225 +90,110 @@ export const createTraceEvalJobs = async ({
...
-        }
-      }
-
-      logger.debug(
-        `Eval job for config ${config.id} matched dataset run item ids ${JSON.stringify(datasetItems.map((d) => d.id))}`,
-      );
-
+    // If we matched a trace and datasetitem, we might want to create a job
+    // If this is only run for a trace, datasetItemExists always default to true.
+    if (traceExists && datasetItemExists) {
```
marliessophie, line 190: Nit: as the code is quite complex leading up to this, I think we could be clearer on this comment. The second part is helpful, I would lead with that. I would also provide more context in the comment on the flow of this code, as someone must understand: 
- that if we are dealing with an event from the trace upsert queue we don't check if dataset items exist (irrelevant), and that defaults to true 
- that if we are dealing with an event from the dataset items queue, we require the corresponding trace to be available for the eval job to be successful

web/src/pages/api/public/dataset-run-items.ts
Steffen911: Created https://linear.app/langfuse/issue/LFE-2948/consider-an-env-helper-function-to-avoid-error-prone-===-true-lookup as a tracking issue.

Steffen911:
Deploying this to staging for testing.

packages/shared/src/server/redis/evalExecutionQueue.ts
```diff
@@ -26,7 +26,7 @@ export class EvalExecutionQueue {
             defaultJobOptions: {
               removeOnComplete: true,
               removeOnFail: 10_000,
-              attempts: 2,
+              attempts: 10,
```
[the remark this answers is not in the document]
Steffen911: @maxdeichmann We essentially already have this setting as we've overwritten it in the `evalService.ts`. I'm just pulling it into the default config. Hence, I don't expect any changes in the alert frequency or behaviour due to this change.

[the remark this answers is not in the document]
maxdeichmann: ok, thx

PR #4339, opened 2024-11-20
security: upgrade cross-spawn

maxdeichmann:


pilsna:
@maxdeichmann Thanks for putting in the effort of getting packages up-to-date.
When I'm scanning the latest langfuse2 image, it still complains about cross-spawn.
Screenshot 2024-12-03 at 15 26 29
Looks like it is coming from the `prisma-kysely` package.

pilsna:
@maxdeichmann Thank you for the reply. By 'production', do you mean that it is only used at build-time?
I am working to get this image approved by the infosec team and I need a justification as to why this image is ok to use in production.

PR #4360, opened 2024-11-21
feat(evaluators): add updates

hassiebp:
> [!IMPORTANT]
> Enhance evaluator feature with edit functionality, status management, and improved analytics tracking.
> 
>   - **Behavior**:
>     - Add edit functionality to `EvaluatorDetail` with a dialog for editing evaluator configurations.
>     - Update evaluator status toggle in `DeactivateEvaluator` to switch between ACTIVE and INACTIVE states.
>     - Modify `InnerEvalConfigForm` to handle both creation and editing of evaluator configurations.
>   - **Schemas**:
>     - Introduce `CreateEvalJobSchema` and `UpdateEvalJobSchema` in `router.ts` for job creation and updates.
>     - Add `EvaluatorStatus` and `EvaluatorStatusSchema` in `types.ts` for status management.
>   - **Analytics**:
>     - Add new events `activate`, `deactivate`, and `update` to `eval_config` in `usePostHogClientCapture.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 837df87fa1a258b1ad54b85d6fa65e735b58a317. It will automatically update as commits are pushed.</sup>


web/src/features/posthog-analytics/usePostHogClientCapture.ts
```diff
@@ -59,7 +59,13 @@ const events = {
     "version_delete_submit",
   ],
   session_detail: ["publish_button_click"],
-  eval_config: ["delete", "new_form_submit", "new_form_open"],
```
marliessophie, line 62: nit/question: unsure if we "lose" historic data if we rename an event, often in analytics the event names are treated as uuids, so we'd have a new reporting after the rename. Either way, no biggy in this case, would just be good to know/keep in mind

web/src/ee/features/evals/server/router.ts
```diff
@@ -42,6 +43,27 @@ export const CreateEvalTemplate = z.object({
...
 });
 
+const CreateEvalJobSchema = z.object({
+  projectId: z.string(),
+  evalTemplateId: z.string(),
+  scoreName: z.string().min(1),
+  target: z.string(),
+  filter: z.array(singleFilter).nullable(), // re-using the filter type from the tables
+  mapping: z.array(variableMapping),
+  sampling: z.number().gte(0).lte(1),
```
marliessophie, line 53: Nit: `sampling` should be `sampling: z.number().gt(0).lte(1),`, as sampling rate of 0 does not make sense. I believe we control for this on the FE already

web/src/ee/features/evals/server/router.ts
```diff
@@ -42,6 +43,27 @@ export const CreateEvalTemplate = z.object({
...
+  delay: z.number().gte(0).default(DEFAULT_TRACE_JOB_DELAY), // 10 seconds default
+});
+
+const UpdateEvalJobSchema = z.object({
+  scoreName: z.string().min(1).optional(),
+  target: z.string().optional(),
```
marliessophie, line 59: I don't think we'll want to update the target, at least that's disabled on the UI, can we please remove it here?

web/src/ee/features/evals/server/router.ts
```diff
@@ -42,6 +43,27 @@ export const CreateEvalTemplate = z.object({
...
+  filter: z.array(singleFilter).optional(),
+  variableMapping: z.array(variableMapping).optional(),
+  sampling: z.number().gte(0).lte(1).optional(),
```
marliessophie, line 62: Nit: same as above, we should adjust to `sampling: z.number().gt(0).lte(1).optional()`

web/src/ee/features/evals/server/router.ts
```diff
@@ -545,7 +556,7 @@ export const evalRouter = createTRPCRouter({
       z.object({
```
marliessophie, line 556: Note: Once you merge main you'll see that we also call `updateEvalJob` [here](https://github.com/langfuse/langfuse/blob/882ad83aad92816d5c630571be58d0823b610056/web/src/ee/features/experiments/components/CreateExperimentsForm.tsx#L281), we will need to reformat the input slightly there too :)

web/src/ee/features/evals/components/evaluator-form.tsx
```diff
@@ -457,7 +493,7 @@ export const InnerEvalConfigForm = (props: {
   }
 
   return (
-    <Form {...form}>
+    <Form {...form} key={props.existingEvaluator?.updatedAt?.toISOString()}>
```
marliessophie, line 496: Mini-nit: why not just the evaluator id and the mode? this seems somewhat complex

web/src/ee/features/evals/components/evaluator-form.tsx
```diff
@@ -427,21 +442,42 @@ export const InnerEvalConfigForm = (props: {
...
-        sampling: values.sampling,
-        delay: values.delay * 1000, // multiply by 1k to convert to ms
-      })
+    const delay = values.delay * 1000; // convert to ms
+    const sampling = values.sampling;
+    const mapping = validatedVarMapping.data;
+    const filter = validatedFilter.data;
+    const scoreName = values.scoreName;
+
+    (props.mode === "edit" && props.existingEvaluator
```
marliessophie, line 451: This call should be awaited if we mutate async :)

web/src/ee/features/evals/components/evaluator-detail.tsx
```diff
@@ -63,87 +73,109 @@ export const EvaluatorDetail = () => {
...
+              jobConfigurationId={existingEvaluator.id}
+            />
+          }
+          cardTitleChildren={
+            <div className="flex w-full flex-row items-center justify-between">
+              <span>Evaluator configuration</span>
+              <Dialog open={isEditOpen} onOpenChange={setIsEditOpen}>
+                <DialogTrigger asChild>
+                  <Button variant="outline" className="flex items-center gap-2">
+                    <Edit className="h-5 w-5" />
```
marliessophie, line 124: Nit: this should be `h-4 w-4`, we should really just set a helper for that though, will add to my backlog

marliessophie (commented):
Thanks so much for helping out and the work on this 🎨 Looks super good! 
Left mostly nits/questions, so should be good to go super soon!

web/src/ee/features/evals/components/evaluator-detail.tsx
hassiebp: Done, thanks 👍🏾

web/src/ee/features/evals/server/router.ts
hassiebp: Done, thanks 👍🏾

web/src/ee/features/evals/server/router.ts
hassiebp: Done, thanks 👍🏾

web/src/ee/features/evals/server/router.ts
hassiebp: Done, thanks 👍🏾

web/src/features/posthog-analytics/usePostHogClientCapture.ts
hassiebp: We are breaking in such that there is no more `delete` events, but only `toggling`. In what way do you mean we are 'losing' events? 🤔

web/src/ee/features/evals/components/evaluator-form.tsx
hassiebp: There are .then and .catch callbacks attached to handle the results, so no awaiting necessary here 👍🏾

web/src/ee/features/evals/components/evaluator-form.tsx
hassiebp: We would not force a remount if id and mode are staying the same on updates of the existingEvaluator. So I added the `updatedAt` timestamp to force remount on update. However, the form did not handle that well at all, so I added a `router.reload()` to force reload the page to show consistent state

hassiebp: I'll remove that

web/src/ee/features/evals/server/router.ts
hassiebp: Done, thanks!

web/src/features/posthog-analytics/usePostHogClientCapture.ts
hassiebp: I think there is no other way than to extend by new events by having toggles now instead of deletes 👍🏾

PR #4398, opened 2024-11-24
chore: make trace deletion asynchronous

Steffen911:
## What

Move trace deletion from synchronous calls to async calls in case Redis is enabled. In that case we perform deletions by single IDs which should make the likeliness of timeouts smaller for multiple traces that are to be deleted.

----

> [!IMPORTANT]
> Introduce asynchronous trace deletion using a Redis-backed queue with fallback to synchronous deletion if the queue is not initialized.
> 
>   - **Behavior**:
>     - Introduces asynchronous trace deletion using `TraceDeleteQueue` in `traceRouter` in `traces.ts`.
>     - Falls back to synchronous deletion if `TraceDeleteQueue` is not initialized.
>   - **Queues**:
>     - Adds `TraceDelete` to `QueueName` and `QueueJobs` enums in `queues.ts`.
>     - Defines `TraceDeleteQueue` in `traceDelete.ts` with retry options and error logging.
>   - **Processors**:
>     - Implements `traceDeleteProcessor` in `traceDelete.ts` to handle trace deletion jobs.
>   - **Environment**:
>     - Adds `QUEUE_CONSUMER_TRACE_DELETE_QUEUE_IS_ENABLED` and `LANGFUSE_TRACE_DELETE_CONCURRENCY` to `env.ts` for configuration.
>   - **Misc**:
>     - Renames `TraceUpsertEventType` to `TraceQueueEventType` in multiple files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 7aa8bf2b676ff0fb1c13c21ba7ce2f0b4dce580d. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/traces.ts
```diff
@@ -658,54 +663,72 @@ export const traceRouter = createTRPCRouter({
...
 
-      await ctx.prisma.$transaction([
-        ctx.prisma.trace.deleteMany({
-          where: {
-            id: {
-              in: input.traceIds,
+      if (!traceDeleteQueue) {
+        logger.warn(
+          `TraceDeleteQueue not initialized. Try synchronous deletion for ${input.traceIds.length} traces.`,
+        );
```
Steffen911, line 669: Moved all of the existing logic into this if block. That keeps it compatible for V2 and allows us a fallback if Redis instantiation doesn't work. Can probably be dropped in the future.

PR #4399, opened 2024-11-24
chore: make project deletion asynchronous

Steffen911:
## What

- [x] Add deleted_at timestamp to mark projects as deleted.
- [x] Move project deletion processing to worker

----

> [!IMPORTANT]
> Implement asynchronous project deletion using a `deleted_at` timestamp and a worker queue to handle the deletion process.
> 
>   - **Behavior**:
>     - Add `deleted_at` timestamp to `Project` model in `types.ts` and `schema.prisma` for soft deletion.
>     - Modify project deletion logic in `projectsRouter.ts` to set `deleted_at` and enqueue deletion job.
>     - Implement `projectDeleteProcessor` in `projectDelete.ts` to handle asynchronous deletion.
>   - **Database Schema**:
>     - Add `deleted_at` column to `projects` table in `migration.sql`.
>   - **Queue and Processing**:
>     - Add `ProjectDeleteQueue` in `projectDelete.ts` and register in `getQueue.ts`.
>     - Register `projectDeleteProcessor` in `app.ts` with concurrency settings in `env.ts`.
>   - **Code Changes**:
>     - Update `organizationRouter.ts`, `projectsRouter.ts`, `membersRouter.ts`, `telemetry/index.ts`, `trpc.ts` to handle non-deleted projects.
>     - Add checks for `deleted_at` in various routers to prevent operations on deleted projects.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b57459ae06b850f93d2936238a271dd89f55ec78. It will automatically update as commits are pushed.</sup>


web/src/features/projects/server/projectsRouter.ts
```diff
@@ -105,6 +111,9 @@ export const projectsRouter = createTRPCRouter({
         action: "delete",
       });
 
+      // TODO: Set deleted_at to current timestamp instead of deleting the project
+      // Afterwards, send message to ProjectDelete queue for asynchronous cleanup.
+      // Do we want to remove all corresponding trace information or does that happen automatically using Prisma cascade?
```
Steffen911, line 116: Validate assumptions here: Do we actually cascade via Prisma or how do we ensure that everything is deleted?

Further TODO: Delete things from ClickHouse as well.

web/src/pages/api/public/projects.ts
```diff
@@ -29,6 +29,7 @@ export default async function handler(
       const projects = await prisma.project.findMany({
         where: {
           id: authCheck.scope.projectId,
+          // deletedAt: null, // here we want to include deleted projects and grey them in the UI.
```
Steffen911, line 32: TODO: Adjust in UI to show status accordingly for deleted projects.

web/src/features/organizations/components/ProjectOverview.tsx
```diff
@@ -134,16 +134,22 @@ const SingleOrganizationProjectOverview = ({
...
+                    <Link href={`/project/${project.id}/settings`}>
+                      <Settings size={16} />
+                    </Link>
+                  </Button>
+                </CardFooter>
+              ) : (
+                <CardContent>
+                  <CardDescription>Project is being deleted</CardDescription>
+                </CardContent>
+              )}
```
Steffen911, line 152: @maxdeichmann Can you check whether this logic seems reasonable?

packages/shared/src/server/queues.ts
```diff
@@ -71,6 +71,10 @@ export const TracesQueueEventSchema = z.object({
   projectId: z.string(),
   traceIds: z.array(z.string()),
 });
+export const ProjectQueueEventSchema = z.object({
```
maxdeichmann, line 74: Lets rename this to ProjectDeleteQueueEventSchema

web/src/features/projects/server/projectsRouter.ts
```diff
@@ -160,13 +193,15 @@ export const projectsRouter = createTRPCRouter({
       const project = await ctx.prisma.project.findUnique({
         where: {
           id: input.projectId,
+          deletedAt: null,
```
maxdeichmann, line 196: did you check all cases where we fetch the project from db? We need to add this condition everywhere.

maxdeichmann: also we need to make everyone from the team aware

web/types/next-auth.d.ts
```diff
@@ -44,6 +44,7 @@ declare module "next-auth" {
       projects: {
         id: PrismaProject["id"];
         name: PrismaProject["name"];
+        deletedAt: PrismaProject["deletedAt"];
```
maxdeichmann, line 47: do we have to add this here? Id prefer to not make any changes on next-auth logic.

web/src/features/projects/server/projectsRouter.ts
```diff
@@ -121,17 +132,39 @@ export const projectsRouter = createTRPCRouter({
...
+      );
+
+      await ctx.prisma.project.update({
         where: {
           id: input.projectId,
           orgId: ctx.session.orgId,
         },
+        data: {
+          deletedAt: new Date(),
+        },
```
maxdeichmann, line 147: Differently than before, we soft-delete here. I think we should delete API keys here already or only fetch API keys in auth if the projectId is not soft-deleted. Otherwise, users are still able to ingest data into their project.

web/src/features/projects/server/projectsRouter.ts
Steffen911: Yes, I updated all calls where we use prisma.project and did not find any raw queries that reference the projects table.

I'll send an update via #engineering after merging this.

worker/src/app.ts
```diff
@@ -66,6 +67,12 @@ if (env.QUEUE_CONSUMER_TRACE_DELETE_QUEUE_IS_ENABLED === "true") {
   });
 }
 
+if (env.QUEUE_CONSUMER_PROJECT_DELETE_QUEUE_IS_ENABLED === "true") {
+  WorkerManager.register(QueueName.ProjectDelete, projectDeleteProcessor, {
+    concurrency: env.LANGFUSE_PROJECT_DELETE_CONCURRENCY,
```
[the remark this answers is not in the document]
Steffen911: @maxdeichmann Bullmq usually renews the timeout as long as the processor is still ongoing. So I would expect that we don't see issues here and would try it at first. Worst-case, we are notified about an issue and fallback to the current manual deletion while improving this. So I'd keep it simple.

web/types/next-auth.d.ts
Steffen911: @maxdeichmann This seems to be the information that we use in the frontend landing page to decide how to render the projects. I.e. to make the change in `web/src/features/organizations/components/ProjectOverview.tsx` this seemed like the place I had to edit.

web/src/features/projects/server/projectsRouter.ts
Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/4399/commits/b57459ae06b850f93d2936238a271dd89f55ec78. I just saw the invalidation on Redis and did not think about Postgres. Makes perfect sense!

worker/src/queues/projectDelete.ts
```diff
@@ -0,0 +1,55 @@
...
+      where: {
+        projectId,
+      },
+    }),
+    prisma.score.deleteMany({
+      where: {
+        projectId,
+      },
+    }),
+  ]);
```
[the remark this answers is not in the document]
Steffen911: I think this goes counter to the learnings from last week where single large deletions should be more effective on ClickHouse. As projectId is the first element of the sorting key, I'd also expect that we can remove them quite efficiently.

packages/shared/src/server/queues.ts
Steffen911: I actually like it that way because we could reuse it for project create events or similar ones. There is no reason to tie the body of a projectId/orgId combination to delete events in my mind.

PR #4419, opened 2024-11-25
fix: change the default of the migration script

maxdeichmann:
> [!IMPORTANT]
> Change migration script defaults based on `CLICKHOUSE_CLUSTER_ENABLED`, affecting engine and source paths in `up.sh` and `down.sh`.
> 
>   - **Behavior**:
>     - Change default behavior in `up.sh` and `down.sh` for `CLICKHOUSE_CLUSTER_ENABLED`.
>     - When `CLICKHOUSE_CLUSTER_ENABLED` is false, use `MergeTree` engine and `unclustered` migration source.
>     - When `CLICKHOUSE_CLUSTER_ENABLED` is true, use `ReplicatedMergeTree` engine and `clustered` migration source.
>   - **Environment**:
>     - Change `CLICKHOUSE_CLUSTER_DISABLED` to `CLICKHOUSE_CLUSTER_ENABLED` in `.env.dev.example` and `.env.dev-azure.example`.
>     - Set `CLICKHOUSE_CLUSTER_ENABLED=false` in `pipeline.yml` for various jobs.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 0dc9c31d4ec9980ec999b0852b1ca34973e27303. It will automatically update as commits are pushed.</sup>


.github/workflows/pipeline.yml
```diff
@@ -340,6 +340,7 @@ jobs:
           echo "LANGFUSE_ASYNC_INGESTION_PROCESSING=true" >> .env
           echo "LANGFUSE_CACHE_API_KEY_ENABLED=true" >> .env
           echo "LANGFUSE_CACHE_PROMPT_ENABLED=true" >> .env
+          echo "CLICKHOUSE_CLUSTER_ENABLED=false" >> .env
```
Steffen911, line 343: This shouldn't be necessary as we already cover it via `cp .env.dev.example .env`.

Steffen911 (approved):

PR #4536, opened 2024-12-03
feat(cost-tracking): allow additional cost and usage keys

hassiebp:
- **allow for additional usage and cost keys**
- **update fern**

----

> [!IMPORTANT]
> Enhance cost-tracking by adding support for `usageDetails` and `costDetails`, deprecating older fields, and updating related models, queries, and tests.
> 
>   - **Behavior**:
>     - Add `usageDetails` and `costDetails` to `Observation` and `IngestionUsage` in `commons.yml` and `ingestion.yml`.
>     - Deprecate `usage` and `calculated*Cost` fields in favor of `usageDetails` and `costDetails`.
>     - Update `EventProcessor.ts` to handle new keys for token and cost calculations.
>   - **Models**:
>     - Update `UsageDetails` and `OpenAIUsageSchema` in `types.ts`.
>     - Modify `convertObservationToView` in `observations_converters.ts` to include new details.
>   - **Database**:
>     - Adjust queries in `dashboards.ts`, `observations.ts`, and `traces.ts` to use new keys.
>   - **Tests**:
>     - Update integration tests in `IngestionService.integration.test.ts` to validate new behavior.
>   - **Misc**:
>     - Update `default-model-prices.json` with new pricing details.
>     - Add `BreakdownToolTip.tsx` for UI enhancements in displaying usage and cost breakdowns.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for df52640c9a563ec6edcb63737a858035bdc44292. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/ingestion/types.ts
```diff
@@ -298,6 +306,8 @@ export const LegacyObservationBody = z.object({
   input: jsonSchema.nullish(),
   output: jsonSchema.nullish(),
   usage: usage,
+  usageDetails: UsageOrCostDetails,
+  costDetails: UsageOrCostDetails,
```
maxdeichmann, line 310: i would not add anything here anymore.

packages/shared/src/server/ingestion/types.ts
```diff
@@ -119,6 +123,8 @@ export const CreateGenerationBody = CreateSpanBody.extend({
     )
     .nullish(),
   usage: usage,
+  usageDetails: UsageOrCostDetails,
```
maxdeichmann, line 126: I am wondering whether we should add some sort of API versioning here. Then we could more easily track who is using old fields and remove them more safely.

packages/shared/src/tableDefinitions/mapObservationsTable.ts
```diff
@@ -68,7 +68,7 @@ export const observationsTableUiColumnDefinitions: UiColumnMapping[] = [
     uiTableId: "timeToFirstToken",
     clickhouseTableName: "observations",
     clickhouseSelect:
-      "if(isNull(completion_start_time), NULL,  date_diff('milliseconds', start_time, completion_start_time))",
+      "if(isNull(completion_start_time), NULL,  date_diff('seconds', start_time, completion_start_time))",
```
hassiebp, line 71: this will not work in UI if we use decimals. Change to milliseconds and convert user entry to ms

fern/apis/server/definition/ingestion.yml
```diff
@@ -312,3 +316,17 @@ types:
...
+      completion_tokens: integer
+      total_tokens: integer
+      prompt_tokens_details: optional<map<string, integer>>
+      completion_tokens_details: optional<map<string, integer>>
+
+  UsageDetails:
+    discriminated: false
+    union:
+      - map<string, integer>
+      - OpenAIUsageSchema
```
maxdeichmann, line 332: Why do we need this exactly?

maxdeichmann: Currently we try to map within the SDK to convert from Openai to Langfuse schemas. Do you want to handle all of this server side now?

packages/shared/src/server/repositories/observations.ts
```diff
@@ -976,8 +976,8 @@ export const getObservationMetricsForPrompts = async (
           prompt_version,
           min(start_time) AS first_observation,
           max(start_time) AS last_observation,
-          medianExact(usage_details['input']) AS median_input_usage,
-          medianExact(usage_details['output']) AS median_output_usage,
+          medianExact(arraySum(mapValues(mapFilter(x -> positionCaseInsensitive(x.1, 'input') > 0, usage_details)))) AS median_input_usage,
+          medianExact(arraySum(mapValues(mapFilter(x -> positionCaseInsensitive(x.1, 'output') > 0, usage_details)))) AS median_output_usage,
```
maxdeichmann, line 980: can you try this query for large users on US?

packages/shared/src/tableDefinitions/mapDashboards.ts
```diff
@@ -85,4 +85,10 @@ export const dashboardColumnDefinitions: UiColumnMapping[] = [
     uiTableId: "version",
     uiTableName: "Version",
   },
+  {
+    clickhouseTableName: "observations",
+    clickhouseSelect: "provided_model_name",
+    uiTableId: "model",
+    uiTableName: "Model",
```
maxdeichmann, line 92: can all charts deal with this filter condition?

packages/shared/src/tableDefinitions/mapObservationsTable.ts
maxdeichmann: are seconds fine here? From my experience, we need milliseconds.

web/src/features/public-api/types/observations.ts
```diff
@@ -89,6 +91,8 @@ export const transformDbToApiObservation = (
     observation;
 
   return {
+    usageDetails: {}, // Important: order matters here, in PG there are no usageDetails but in CH there are and will be written by rest
+    costDetails: {}, // Important: order matters here, in PG there are no costDetails but in CH there are and will be written by rest
```
maxdeichmann, line 95: Did you check the API docs that this here + daily metrics are the only ones we need to check?

fern/apis/server/definition/ingestion.yml
hassiebp: Yes, I moved this to server side to also support users ingesting via API and to not duplicate logic in SDKs.

packages/shared/src/tableDefinitions/mapObservationsTable.ts
hassiebp: [Fixed it in this commit](https://github.com/langfuse/langfuse/pull/4536/commits/36b5a4741d62da6a6f3cca11962598b048e9715e) to also allow both sorting and filtering by decimal second value :)

packages/shared/src/server/ingestion/types.ts
```diff
@@ -56,6 +56,62 @@ export const usage = MixedUsage.nullish()
...
+  z.number().nonnegative().nullish(),
+);
+
+const OpenAIUsageSchema = z
+  .object({
+    prompt_tokens: z.number().nonnegative(),
+    completion_tokens: z.number().nonnegative(),
+    total_tokens: z.number().nonnegative(),
+    prompt_tokens_details: z.record(z.string(), z.number().nonnegative()),
+    completion_tokens_details: z.record(z.string(), z.number().nonnegative()),
```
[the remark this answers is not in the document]
hassiebp: I removed that as legacy for `usage_details`. We will only map the OpenAI usage keys to Langfuse keys, but not support the camelCase version that I assume was added for the legacy observation schema where input tokens were named `promptTokens`.

worker/src/services/IngestionService/index.ts
```diff
@@ -931,21 +939,36 @@ export class IngestionService {
...
-        newOutputCount
-          ? newInputCount + newOutputCount
-          : (newInputCount ?? newOutputCount));
+        (Object.keys(
+          "usageDetails" in obs.body ? (obs.body.usageDetails ?? {}) : {},
+        ).length === 0
+          ? newInputCount !== undefined &&
+            newOutputCount !== undefined &&
+            newInputCount &&
+            newOutputCount
```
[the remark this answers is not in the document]
hassiebp: The first one is an existence check ("not undefined") and the second one is a truthyness check whether the value is greater than zero.

So we check first if both input and output count are provided and greater than zero, then we calculate the total, else we fall back to the inputCount or outputCount whichever was provided, if both are undefined the total is undefined.

In any case, this is not new logic, but only extended with the `usageDetails` logic

worker/src/constants/default-model-prices.json
```diff
@@ -4,10 +4,11 @@
     "model_name": "gpt-4o",
     "match_pattern": "(?i)^(gpt-4o)$",
     "created_at": "2024-05-13T23:15:07.670Z",
-    "updated_at": "2024-05-13T23:15:07.670Z",
+    "updated_at": "2024-12-03T10:12:31.000Z",
```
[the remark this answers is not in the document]
hassiebp: Yes, I'll nevertheless double check all prices again now

packages/shared/src/tableDefinitions/mapDashboards.ts
hassiebp: From what I can see this is only used in the query builder and not in the offered filter options on the dasboard page: 

CleanShot 2024-12-16 at 15 54 54@2x

Am I missing something maybe?

packages/shared/src/server/repositories/observations.ts
hassiebp: Tried this query for one of our largest US customers, query took less than 200ms:

```sql
SELECT
    type,
	SUM(arraySum (mapValues (mapFilter (x -> positionCaseInsensitive (x .1, 'input') > 0, usage_details))) AS input_usage),
	SUM(arraySum (mapValues (mapFilter (x -> positionCaseInsensitive (x .1, 'output') > 0, usage_details))) AS output_usage)
FROM
	observations FINAL
WHERE
	(type = 'GENERATION')
	AND start_time >= toDateTime ('2024-12-16')
	AND project_id = '<projectId>'
GROUP BY
	type
```

PR #4562, opened 2024-12-04
chore: user table query

Steffen911:


user_query.sql
```diff
@@ -0,0 +1,49 @@
...
+            and t.project_id = o.project_id
+    WHERE
+        o.project_id = {projectId: String }
+      AND o.trace_id in (
+        SELECT
+        id
+        from
+        traces
+        where
+        user_id IN ({userIds: Array(String) })
```
Steffen911, line 19: Does it make sense to include projectId here as well? even if we have a bloom filter, using parts of the order key should reduce the result set even further. Also, we avoid duplicates if userIds are not unique across projects.

Steffen911: Also, you may want to include the IS NOT NULL and != '' filter here in case they bring any advantages. I would actually propose that we drop them in both queries as I don't see why they would help if we have the `IN` condition anyway.

user_query.sql
```diff
@@ -0,0 +1,49 @@
+WITH observations_agg AS (
+    SELECT
+        t.user_id,
+        count(*) as obs_count,
+        sumMap(usage_details) as sum_usage_details,
+        sum(total_cost) as sum_total_cost
+    FROM
+        observations o
+            join traces t on t.id = o.trace_id
+            and t.project_id = o.project_id
```
Steffen911, line 10: I read about using argMax for deduplication at query time this morning: https://clickhouse.com/blog/postgres-to-clickhouse-data-modeling-tips#use-argmax-to-deduplicate-rows-at-query-time. In that case, it may be applicable and avoid your need to use final. As there is a grouping anyway it should be fine as well.

user_query.sql
```diff
@@ -0,0 +1,49 @@
...
+      AND o.type = 'GENERATION'
+    GROUP BY
+        t.user_id
+),
+     user_metric_data AS (
+         SELECT
+             t.user_id,
+             max(t.timestamp) as max_timestamp,
+             min(t.timestamp) as min_timestamp,
+             count(*) as trace_count
```
Steffen911, line 34: Same thing here. argMax may help to make results more correct.

user_query.sql
maxdeichmann: agree on all

PR #4575, opened 2024-12-04
feat(url): support AWS S3 signed images in trace/observation preview

marliessophie:
> [!IMPORTANT]
> Adds AWS S3 signed URL validation and updates logging in `utilities.ts`.
> 
>   - **Behavior**:
>     - Adds `isValidPresignedS3Url()` in `utilities.ts` for AWS S3 signed URL validation.
>     - Updates `isValidImageUrl()` to incorporate `isValidPresignedS3Url()` for S3 URL handling.
>     - Replaces `console.log` with `logger.info` for logging in `utilities.ts`.
>   - **Validation**:
>     - Introduces `s3UrlSchema` for AWS S3 URL structure validation.
>     - Handles HEAD and GET requests for S3 URL validation, checking for 200, 206, and 403 status codes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 111b0a8676a244c2d6d8b54c7acd38bde6860864. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/server/api/routers/utilities.ts
+++ b/web/src/server/api/routers/utilities.ts
@@ -2,6 +2,7 @@ import { createTRPCRouter, protectedProcedure } from "@/src/server/api/trpc";
 import { z } from "zod";
 import { promises as dns } from "dns";
 import { Address4, Address6 } from "ip-address";
+import { logger } from "@langfuse/shared/src/server";
 
 const IP_4_LOOPBACK_SUBNET = "127.0.0.0/8";
 const IP_4_LINK_LOCAL_SUBNET = "169.254.0.0/16";
@@ -38,7 +39,7 @@ const isPrivateIp = (ipAddress: string): boolean => {
     }
     return false;
   } catch (error) {
-    console.error("IP parsing error:", error);
+    logger.info("IP parsing error:", error);
     return false;
   }
 };
@@ -52,13 +53,13 @@ const resolveHostname = async (
   try {
     addresses4 = await dns.resolve4(hostname);
   } catch (error) {
-    console.log("IPv4 DNS resolution error:", error);
+    logger.info("IPv4 DNS resolution error:", error);
   }
 
   try {
     addresses6 = await dns.resolve6(hostname);
   } catch (error) {
-    console.log("IPv6 DNS resolution error:", error);
+    logger.info("IPv6 DNS resolution error:", error);
   }
 
   return { addresses4, addresses6 };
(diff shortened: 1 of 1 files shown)


web/src/server/api/routers/utilities.ts
```diff
@@ -81,17 +82,102 @@ const isValidAndSecureUrl = async (urlString: string): Promise<boolean> => {
...
+    "X-Amz-Algorithm": z.string().optional(),
+    "X-Amz-Credential": z.string().optional(),
+    "X-Amz-Date": z.string().optional(),
+    "X-Amz-Expires": z.string().optional(),
+    "X-Amz-SignedHeaders": z.string().optional(),
+    "X-Amz-Signature": z.string(),
+    "X-Amz-Security-Token": z.string().optional(),
+    "x-id": z.string().optional(), // Optional ID parameter, not documented in AWS docs as it is SDK specific, https://github.com/aws/aws-sdk-go-v2/blob/04e7aca073a0a7ed479aa37cad88a1cf58a979a1/service/s3/internal/customizations/presign_test.go#L35-L41
+  }),
+});
```
Steffen911, line 108: Do we also have to consider presigned URLs for Azure, GCP, ... here? Those may be relevant for self-hosters.

web/src/server/api/routers/utilities.ts
```diff
@@ -81,17 +82,102 @@ const isValidAndSecureUrl = async (urlString: string): Promise<boolean> => {
...
+    }
     return false;
   }
 };
 
 const isValidImageUrl = async (url: string): Promise<boolean> => {
   try {
+    if (await isValidPresignedS3Url(url)) {
+      return true;
+    }
```
Steffen911, line 174: Do we have to validate whether the presigned URL actually points to an image (which may make sense given the function name)? With the current logic, I assume one could also send us a presigned JS file which may make use of Remote Code Execution? 🤔 Or do we safe-parse that within the frontend?

web/src/server/api/routers/utilities.ts
```diff
@@ -81,17 +82,102 @@ const isValidAndSecureUrl = async (urlString: string): Promise<boolean> => {
...
+    } else {
+      logger.info("Unknown error:", error);
+    }
     return false;
   }
 };
 
 const isValidImageUrl = async (url: string): Promise<boolean> => {
   try {
+    if (await isValidPresignedS3Url(url)) {
```
Steffen911, line 172: Could you add a 1-2 sentence comment on why S3 presigned URLs deserve special behaviour here? The head fetch + content type check seems sound and it's not obvious why they wouldn't work with a S3 presigned URL.

marliessophie: Will do, yes 🙌

web/src/server/api/routers/utilities.ts
marliessophie: Good shout, it is validated in the front-end also, but if someone reuses this function def a requirement to validate the content type as image also server side, will double check

web/src/server/api/routers/utilities.ts
marliessophie: Not yet, I'd need to check if our current general check would catch these or not. Given no one has reported lack of support for Azure, GCP (or else) yet, I don't think we need to look into this right now :)

Steffen911 (approved):

PR #4613, opened 2024-12-06
feat: set v3 default variables

maxdeichmann:
> [!IMPORTANT]
> Set `LANGFUSE_POSTGRES_INGESTION_ENABLED` default to `false` and update related logic for Postgres score creation.
> 
>   - **Behavior**:
>     - Change default of `LANGFUSE_POSTGRES_INGESTION_ENABLED` to `false` in `env.ts`, `env.mjs`, and `worker/src/env.ts`.
>     - Conditional Postgres score creation in `scores.ts` and `evalService.ts` based on `LANGFUSE_POSTGRES_INGESTION_ENABLED`.
>   - **Comments**:
>     - Update comment in `env.mjs` to indicate removal during V3 cleanup.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 1c6098c4959b974b8cef442a6a012d7576833fd9. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -39,7 +39,7 @@ const EnvSchema = z.object({
   // TODO: Disable Postgres before go-live
   LANGFUSE_POSTGRES_INGESTION_ENABLED: z
     .enum(["true", "false"])
-    .default("true"),
+    .default("false"),
 
   LANGFUSE_INGESTION_QUEUE_DELAY_MS: z.coerce
     .number()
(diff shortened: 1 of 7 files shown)


packages/shared/src/env.ts
```diff
@@ -39,7 +39,7 @@ const EnvSchema = z.object({
   // TODO: Disable Postgres before go-live
```
Steffen911, line 39: Let's drop the comment

worker/src/features/evaluation/evalService.ts
```diff
@@ -470,12 +470,14 @@ export const evaluate = async ({
   };
 
   // TODO: Remove foreign key on jobExecutions when removing this
```
Steffen911, line 472: Remove comment

PR #4685, opened 2024-12-11
feat: add posthog integration processing

Steffen911:
## What

Add logic to process posthog integration on the worker container.

Details: https://langfuse.com/docs/analytics/posthog

----

> [!IMPORTANT]
> Add PostHog integration with processing logic, queues, and UI settings for data synchronization.
> 
>   - **Behavior**:
>     - Adds PostHog integration processing logic in `handlePostHogIntegrationProjectJob.ts` and `handlePostHogIntegrationSchedule.ts`.
>     - Introduces `postHogIntegrationProcessor` and `postHogIntegrationProcessingProcessor` in `postHogIntegrationQueue.ts`.
>     - Adds new queues `PostHogIntegrationQueue` and `PostHogIntegrationProcessingQueue` in `queues.ts`.
>     - Updates `app.ts` to register PostHog integration queues.
>   - **Models**:
>     - Adds `getTracesForPostHog`, `getGenerationsForPostHog`, and `getScoresForPostHog` functions in `repositories` to fetch data for PostHog.
>   - **UI**:
>     - Adds PostHog integration settings page in `posthog.tsx`.
>     - Updates project settings to include PostHog integration in `index.tsx`.
>   - **Config**:
>     - Adds `QUEUE_CONSUMER_POSTHOG_INTEGRATION_QUEUE_IS_ENABLED` to `env.ts`.
>     - Adds `posthog-node` dependency in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 7870774776c95f5bd5145252b693b1bc068cbcba. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/observations.ts
```diff
@@ -1165,3 +1166,81 @@ export const getTraceIdsForObservations = async (
...
+  const records = await queryClickhouse<Record<string, unknown>>({
+    query,
+    params: {
+      projectId,
+      minTimestamp: convertDateToClickhouseDateTime(minTimestamp),
+      maxTimestamp: convertDateToClickhouseDateTime(maxTimestamp),
+    },
+  });
+
+  const baseUrl = env.NEXTAUTH_URL?.replace("/api/auth", "");
```
Steffen911, line 1217: @marcklingen Any concerns about this approach? This should keep the standard URL unaffected, but handle the special case where we have a custom base path.

I replicated the logic across all three files as it's super easy to read and a util function probably obfuscates what's going on.

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1165,3 +1166,81 @@ export const getTraceIdsForObservations = async (
...
+    langfuse_project_id: projectId,
+    langfuse_user_id: record.trace_user_id || "langfuse_unknown_user",
+    langfuse_latency: record.latency,
+    langfuse_time_to_first_token: record.time_to_first_token,
+    langfuse_release: record.trace_release,
+    langfuse_version: record.version,
+    langfuse_model: record.model,
+    langfuse_level: record.level,
+    langfuse_tags: record.trace_tags,
+    langfuse_integration_version: "1.0.0",
```
Steffen911, line 1238: @marcklingen I updated the integration version here. I'm not sure how we want to handle semantic versioning here, but it feels good to just set it to 1.0.0 now that we moved it into the codebase.

web/src/pages/project/[projectId]/settings/index.tsx
```diff
@@ -144,24 +144,14 @@ const Integrations = (props: { projectId: string }) => {
...
-            </Button>
-          )}
+          <ActionButton
+            variant="secondary"
+            hasAccess={hasAccess}
+            hasEntitlement={hasEntitlement}
+            href={`/project/${props.projectId}/settings/integrations/posthog`}
+          >
+            Configure
+          </ActionButton>
```
Comment (Steffen911) [inline web/src/pages/project/[projectId]/settings/index.tsx:154]: @marcklingen Usage of the ActionButton here makes sense, right?

web/src/pages/project/[projectId]/settings/integrations/posthog.tsx
```diff
@@ -46,7 +46,7 @@ export default function PosthogIntegrationSettings() {
   return (
     <div className="md:container">
       <Header
-        title="PostHog Integration"
+        title="PostHog Integration (Beta)"
```
Comment (maxdeichmann) [inline web/src/pages/project/[projectId]/settings/integrations/posthog.tsx:49]: LOL

maxdeichmann:
Overall looking good. Lets also add observability for SLOs to this

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1165,3 +1166,81 @@ export const getTraceIdsForObservations = async (
...
+  return records.map((record) => ({
+    timestamp: record.start_time,
+    langfuse_generation_name: record.name,
+    langfuse_trace_name: record.trace_name,
+    langfuse_url: `${baseUrl}/project/${projectId}/traces/${encodeURIComponent(record.trace_id as string)}?observation=${encodeURIComponent(record.id as string)}`,
+    langfuse_id: record.id,
+    langfuse_cost_usd: record.total_cost,
+    langfuse_input_units: record.input_tokens,
+    langfuse_output_units: record.output_tokens,
+    langfuse_total_units: record.total_tokens,
```
marcklingen, line 1227: "langfuse_unit" is missing

packages/shared/src/server/repositories/observations.ts
Steffen911: @marcklingen I hope that we don't need it more often. And I'd rather have this logic explicitly in the code than in the env.ts. In the latter case it may happen that we expect the semantic on web and worker to be the same which may cause subtle bugs.

packages/shared/src/server/repositories/observations.ts
Steffen911: No strong feeling about it either. Might be nice to keep it here as it allows us to version events separately. That way it becomes immediately obvious that a version should be bumped. As it's additional effort to change and there are no strong opinions, I'll keep as is.

packages/shared/src/env.ts
```diff
@@ -5,6 +5,7 @@ const EnvSchema = z.object({
   NODE_ENV: z
     .enum(["development", "test", "production"])
     .default("development"),
+  NEXTAUTH_URL: z.string().url().optional(),
```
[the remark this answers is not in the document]
Steffen911: Done via https://github.com/langfuse/infrastructure/pull/236

web/src/pages/project/[projectId]/settings/integrations/posthog.tsx
Comment (Steffen911) [reply web/src/pages/project/[projectId]/settings/integrations/posthog.tsx:49]: IMO those can be two separate things. While the integration may still be Beta (workload, frequency, etc.) the event format can be 1.0.0

packages/shared/src/server/repositories/observations.ts
Steffen911: @marcklingen Langfuse unit is not available in the new table. I can give a dummy value, but dropping seems more intuitive.

packages/shared/src/server/redis/postHogIntegrationQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    if (PostHogIntegrationQueue.instance) {
+      logger.debug("Scheduling jobs for PostHogIntegrationQueue");
+      PostHogIntegrationQueue.instance
+        .add(
+          QueueJobs.PostHogIntegrationJob,
+          {},
+          {
+            repeat: { pattern: "0 3 * * *" }, // 3am daily
```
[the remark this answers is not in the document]
Steffen911: Will move to every 60min via https://github.com/langfuse/langfuse/pull/4685/commits/ce1165362b7b9b84e9c4a6253f3daa93a37a3ac5. Currently, we do not handle streaming from Clickhouse well with our utilities. We may do that if this consumes too many resources, but I'd try to keep it simple and adjust if it fails.

packages/shared/src/server/repositories/observations.ts
marcklingen: we can drop it

packages/shared/src/server/redis/postHogIntegrationQueue.ts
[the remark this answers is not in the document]
marcklingen: sounds good

web/src/pages/project/[projectId]/settings/integrations/posthog.tsx
Comment (marcklingen) [reply web/src/pages/project/[projectId]/settings/integrations/posthog.tsx:49]: agree, good point

packages/shared/src/server/repositories/observations.ts
marcklingen: naming here suggests "integration" versioning

i think changing to "langfuse_event_version" would be good here as well, no strong opinion

Steffen911: Adjusted in https://github.com/langfuse/langfuse/pull/4685/commits/254799ec58f793f082e30ee73d754146f08ec28d. I like it!

Gorica132:
Everywhere good

Issue #4705, opened 2024-12-12
bug: Account creation fails when SSO provider returns unexpected fields


tovarisch-ildar:
### Describe the bug

We are encountering an issue with Langfuse when using ADFS as an SSO provider. The setup includes configuring Langfuse via AUTH_CUSTOM_* environment variables. The problem arises because the ADFS response contains an unexpected field (resource) that we cannot remove or modify.

While the user is created correctly, the account linking step fails, causing the entire sign-in process to break.

### To reproduce

Set up ADFS as an SSO provider.
Configure Langfuse with the following Docker command:

docker run --name langfuse \
-e DATABASE_URL='mypostgress' \
-e NEXTAUTH_URL='myurl' \
-e NEXTAUTH_SECRET=mysecret \
-e SALT=mysalt \
-e LANGFUSE_AUTO_POSTGRES_MIGRATION_DISABLED=false \
-e LANGFUSE_LOG_LEVEL=trace \
-e ENCRYPTION_KEY=key \
-e AUTH_CUSTOM_CLIENT_ID=client-id \
-e AUTH_CUSTOM_CLIENT_SECRET=secret \
-e AUTH_CUSTOM_ISSUER='issuer' \
-e AUTH_CUSTOM_NAME='SSO' \
-e AUTH_CUSTOM_ALLOW_ACCOUNT_LINKING=true \
-p 3030:3000 \
-a STDOUT \
langfuse/langfuse:2.93

Attempt to sign in via the configured ADFS SSO provider.

### SDK and container versions

langfuse:2.93

### Additional information

A similar issue was addressed in this issue: https://github.com/orgs/langfuse/discussions/2165. I aim to implement a more universal solution by cleaning up unnecessary fields before linkAccount. Is it okay that I created a separate issue for the upcoming PR?

### Are you interested to contribute a fix for this bug?

Yes

marcklingen:
Fixed in #4728

PR #4728, opened 2024-12-16
feat(auth): add `AUTH_IGNORE_ACCOUNT_FIELDS` to sanitize IDP fields before creating an account

tovarisch-ildar:
## What does this PR do?

This PR adds field sanitization before creating an Account in the authentication process. This change is necessary because passing extraneous fields can cause sign-in via SSO to fail. Similar issues have been discussed and addressed in [this discussion](https://github.com/orgs/langfuse/discussions/2165)

## Type of change

- [ ] Bug fix for https://github.com/langfuse/langfuse/issues/4705

## Checklist

- I haven't checked if new and existing unit tests pass locally with my changes (there are some problems with running tests in GitHub Codespaces, even on the **main** branch)

----

> [!IMPORTANT]
> Sanitizes fields in `linkAccount` function to prevent SSO sign-in failures by removing extraneous fields not in the `Account` model.
> 
>   - **Behavior**:
>     - Sanitize fields in `linkAccount` function in `auth.ts` to remove extraneous fields not present in the `Account` model.
>     - Prevents SSO sign-in failures due to incompatible data fields.
>   - **Adapters**:
>     - Extend `PrismaAdapter` in `auth.ts` to include field sanitization logic for `Account` model.
>   - **Misc**:
>     - Add `Prisma` import in `auth.ts` for accessing data model information.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e4d2c2b85a2f60b0094730c1d5b8dc2cbce76af8. It will automatically update as commits are pushed.</sup>


marcklingen:
I was just starting to work on this, thanks for the contribution. Can you accept the CLA? Have you tested whether this resolves your issue?

tovarisch-ildar:
> I was just starting to work on this, thanks for the contribution. Can you accept the CLA? Have you tested whether this resolves your issue?

Yes, of course, I've accepted the CLA. I tested it in my infrastructure, and it works perfectly. However, the tests are failing (even in GitHub Codespaces on main branch), but I assume my changes shouldn't have broken anything

marcklingen:
Please see comment above regarding the CLA, apparently your local git config uses an email that is not part of your github profile

tovarisch-ildar:
> > Prisma.dmmf is an internal api and use is discouraged. I'd change this to an allowlist of available keys. I do not expect this to change often, thus, this should be maintainable with a comment in the prisma schema
> 
> I have the changes for this already locally, will push to this branch

Yeah, ok. Thanks

marcklingen:
Unfortunately, this break google auth for me when testing this locally.
```
Argument `user` is missing.
```

Implementation
```
// Extend Prisma Adapter
const prismaAdapter = PrismaAdapter(prisma);
const ALLOWED_ACCOUNT_FIELDS = [
  "id",
  "user_id",
  "type",
  "provider",
  "providerAccountId",
  "refresh_token",
  "access_token",
  "expires_at",
  "expires_in",
  "ext_expires_in",
  "token_type",
  "scope",
  "id_token",
  "session_state",
  "refresh_token_expires_in",
  "created_at",
  "user",
];
const extendedPrismaAdapter: Adapter = {
  ...prismaAdapter,
  async createUser(profile: Omit<AdapterUser, "id">) {
    if (!prismaAdapter.createUser)
      throw new Error("createUser not implemented");
    if (
      env.NEXT_PUBLIC_SIGN_UP_DISABLED === "true" ||
      env.AUTH_DISABLE_SIGNUP === "true"
    ) {
      throw new Error("Sign up is disabled.");
    }
    if (!profile.email) {
      throw new Error(
        "Cannot create db user as login profile does not contain an email: " +
          JSON.stringify(profile),
      );
    }

    const user = await prismaAdapter.createUser(profile);

    await createProjectMembershipsOnSignup(user);

    return user;
  },

  async linkAccount(data: AdapterAccount) {
    if (!prismaAdapter.linkAccount)
      throw new Error("NextAuth: prismaAdapter.linkAccount not implemented");

    // Some sso provider's return incompatible data with the nextjs-auth schema
    // (refresh_expires_in, not-before-policy in, resource).
    // So, we need to remove this data from the payload before linking an account.
    // https://github.com/langfuse/langfuse/issues/4705
    // https://github.com/nextauthjs/next-auth/issues/7655
    for (const key in data) {
      if (!ALLOWED_ACCOUNT_FIELDS.includes(key)) {
        logger.debug(
          `[Auth] Removing incompatible key from account data: ${key}`,
        );
        delete data[key];
      }
    }

    await prismaAdapter.linkAccount(data);
  },
};
```

Do you have an idea how to best fix this?

tovarisch-ildar:
Maybe just add a separate environment variable with a list of fields to delete? This approach will not break anything, and developers can manage it on their side as they wish and there will be no need to update the `ALLOWED_ACCOUNT_FIELDS` array in case of future schema updates

marcklingen:
> Maybe just add a separate environment variable with a list of fields to delete? This approach will not break anything, and developers can manage it on their side as they wish and there will be no need to update the `ALLOWED_ACCOUNT_FIELDS` array in case of future schema updates

Good idea. I suggest keeping the existing logic implemented for Keycloak and adding an optional loop over comma-separated keys that are removed from the account fields. I'd probably recommend making this a disallow list to prevent accidental removal of fields as in the google example that I encountered on localhost. During setup of a new self-hosted langfuse instance, one can then incrementally add to this array until there are no error logs left when trying to authenticate.

tovarisch-ildar:
> > Maybe just add a separate environment variable with a list of fields to delete? This approach will not break anything, and developers can manage it on their side as they wish and there will be no need to update the `ALLOWED_ACCOUNT_FIELDS` array in case of future schema updates
> 
> Good idea. I suggest keeping the existing logic implemented for Keycloak and adding an optional loop over comma-separated keys that are removed from the account fields. I'd probably recommend making this a disallow list (e.g. `AUTH_IGNORE_ACCOUNT_FIELDS=organization,level`) to prevent accidental removal of fields as in the google example that I encountered on localhost. During setup of a new self-hosted langfuse instance, one can then incrementally add to this array until there are no error logs left when trying to authenticate.
> 
> Do you think comma-separated is good enough or does it need to be a stringified json array to make sure escapes work correctly for custom args? No strong opinion here.

I think a comma-separated array should be sufficient, as a similar array is used in [AUTH_DOMAINS_WITH_SSO_ENFORCEMENT](https://github.com/tovarisch-ildar/langfuse/blob/main/web/src/features/auth-credentials/server/signupApiHandler.ts#L45)

tovarisch-ildar:
@marcklingen 
I made the changes but couldn't test them locally because the Docker image is built for version 3 and requires clickhouse. If the PR is merged, will these changes be available in version 2? We are currently unable to run version 3 in our infrastructure

tovarisch-ildar:
> > I made the changes but couldn't test them locally because the Docker image is built for version 3 and requires clickhouse. If the PR is merged, will these changes be available in version 2? We are currently unable to run version 3 in our infrastructure
> 
> Once merged, we can cherry-pick these changes for v2. Can you create an issue with details on what blocks the v3 deployment on your end? Would love to help.

It seems that version 2 will suffice for us at the moment, and we are unable to use ClickHouse right now due to internal issues. :-)

marcklingen:
> we are unable to use ClickHouse right now due to internal issues

Can you share more details? Would love to help resolve this as we have worked with many teams who run Langfuse self-hosted

marcklingen:
release here on v2 branch: https://github.com/langfuse/langfuse/releases/tag/v2.93.7

tovarisch-ildar:
Thank you very much! Alright, I'll get back to you with questions about the version 3 a bit later. Thanks again! :-)

PR #4755, opened 2024-12-17
chore(traces): drop fk on `trace_session` table

marliessophie:
Migration already ran: https://github.com/langfuse/langfuse/pull/4612, we had not updated prisma schema though.


marliessophie:
Fixed in https://github.com/langfuse/langfuse/pull/4946 instead

PR #4802, opened 2024-12-20
feat(models): add UI to add all usage types

hassiebp:
> [!IMPORTANT]
> Add comprehensive UI and server logic for managing model usage types, including creation, editing, cloning, and deletion, with pricing and tokenization features.
> 
>   - **UI Enhancements**:
>     - Add `UpsertModelFormDrawer` for creating, editing, and cloning models.
>     - Add `DeleteModelButton`, `EditModelButton`, and `CloneModelButton` for model actions.
>     - Introduce `PriceBreakdownTooltip` and `PricePreview` for displaying model pricing.
>     - Add `PriceUnitSelector` for selecting price units.
>   - **Data Handling**:
>     - Update `DataTable` to support row click actions and dynamic column visibility.
>     - Implement `usePriceUnitMultiplier` hook for managing price unit state.
>     - Enhance `useLocalStorage` for cross-tab synchronization.
>   - **Server Logic**:
>     - Add `getById`, `getAll`, and `upsert` endpoints in `models.ts` for model management.
>     - Update `traces.ts` to include `usageDetails` and `costDetails` in `ObservationReturnType`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 6955e24131f644e7bff88639ebdf7fb711b1b7dc. It will automatically update as commits are pushed.</sup>


maxdeichmann:
image

When I delete all prices and select Openai, the table below is not filled.

web/src/server/api/routers/models.ts
```diff
@@ -1,69 +1,248 @@
...
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Model not found",
+        });
+      }
+
+      return model;
+    }),
+
+  getAll: protectedProjectProcedure
```
maxdeichmann, line 68: do we have model read roles? If yes, we should add this here.

maxdeichmann: .. and above.

web/src/server/api/routers/models.ts
```diff
@@ -1,69 +1,248 @@
...
+            matchPattern,
+            tokenizerConfig,
+            tokenizerId,
+          },
+        });
+
+        await tx.price.deleteMany({
+          where: {
+            modelId: upsertedModel.id,
+          },
```
maxdeichmann, line 220: should we also add a project id condition just to be sure?

web/src/server/api/routers/models.ts
hassiebp: we only have `models:CUD` for now, read is open to all users

hassiebp:
> image
> When I delete all prices and select Openai, the table below is not filled.

This is expected, as all prices are 0 - table will adjust when non-zero prices are added

PR #5069, opened 2025-01-16
chore: remove outdated postgres executions

Steffen911:
> [!IMPORTANT]
> This pull request removes outdated PostgreSQL executions and transitions to using Clickhouse for data operations across the codebase.
> 
>   - **Behavior**:
>     - Removed PostgreSQL executions in favor of Clickhouse for data operations across multiple files.
>     - Updated data retrieval functions to use Clickhouse, affecting `getScoresForTraces`, `getObservationsViewForTrace`, and others.
>   - **API Changes**:
>     - Modified API routes in `metrics/daily.ts`, `observations/[observationId].ts`, `scores/[scoreId].ts`, and others to use Clickhouse.
>     - Updated query logic in `traces/[traceId].ts` and `sessions/[sessionId].ts` to fetch data from Clickhouse.
>   - **Utilities**:
>     - Removed `measureAndReturnApi` utility and related PostgreSQL logic.
>     - Updated `getTracesByIds`, `getObservationsById`, and similar functions to interface with Clickhouse.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 59c88a1232a2d57911a7cc11795bba0fcc46ec96. It will automatically update as commits are pushed.</sup>


web/src/pages/api/public/scores/[scoreId].ts
```diff
@@ -63,17 +44,7 @@ export default withMiddlewares({
...
-        prisma.score.deleteMany({
-          where: {
-            id: scoreId,
-            projectId: auth.scope.projectId,
-          },
-        }),
-        deleteScore(auth.scope.projectId, scoreId),
-      ]);
-
+      await deleteScore(auth.scope.projectId, scoreId);
```
Comment (Steffen911) [inline web/src/pages/api/public/scores/[scoreId].ts:47]: This will stop deleting scores from Postgres. IMO this would be the expected behaviour going forward, but unless we prune all tables self-hosters may continue to rely on this. I'm open for suggestions on whether to keep it as is or drop it.

web/src/server/api/routers/scores.ts
```diff
@@ -515,81 +372,15 @@ export const scoresRouter = createTRPCRouter({
...
-            key: composeAggregateScoreKey({ name, source, dataType }),
-            name: name,
-            source: source,
-            dataType: dataType,
-          }));
-        },
-      });
+    .query(async ({ input }) => {
+      // TODO: Why is date unused in the ClickHouse flow?
+      // const date = getDateFromOption(input.selectedTimeOption);
```
Steffen911, line 377: For some reason we don't seem to use this date in the ClickHouse execution part. I'm unsure whether it's useful or whether we can drop it. @marliessophie @maxdeichmann does one of you have an idea here?

web/src/features/datasets/server/service.ts
```diff
@@ -71,8 +70,8 @@ export const createDatasetRunsTableWithoutMetrics = async (
     createdAt: run.run_created_at,
     updatedAt: run.run_updated_at,
     // return metric fields as undefined
-    avgTotalCost: undefined,
-    avgLatency: undefined,
+    avgTotalCost: undefined as Decimal | undefined,
```
Steffen911, line 73: This makes the frontend type parsing happy and shouldn't make a difference here.

web/src/pages/api/public/scores/[scoreId].ts
Comment (maxdeichmann) [reply web/src/pages/api/public/scores/[scoreId].ts:47]: I would not delete from postgres anymore as we also do not insert anymore.

web/src/server/api/routers/scores.ts
maxdeichmann: looks like a bug to me. The option is used here (https://github.com/langfuse/langfuse/blob/6aff9b0f4a7979cc10c41ca7ac88ce9a5aa42309/web/src/features/dashboard/components/score-analytics/ScoreAnalytics.tsx#L38) for example. Can you take this up?

Steffen911: Fixed in https://github.com/langfuse/langfuse/pull/5069/commits/59c88a1232a2d57911a7cc11795bba0fcc46ec96

PR #5110, opened 2025-01-20
feat(prompts): support in UI prompt duplication

marliessophie:
> [!IMPORTANT]
> Add UI and backend support for duplicating prompts, including single and all versions, with reusable prompt name validation.
> 
>   - **Feature**:
>     - Add `DuplicatePromptButton` and `DuplicatePromptForm` components in `duplicate-prompt.tsx` to support prompt duplication in the UI.
>     - Support duplicating a single version or all versions of a prompt.
>   - **Backend**:
>     - Add `duplicatePrompt` function in `createPrompt.ts` to handle prompt duplication logic.
>     - Update `promptRouter.ts` to include `duplicatePrompt` mutation for handling duplication requests.
>   - **Validation**:
>     - Extract prompt name validation logic to `usePromptNameValidation` hook for reuse in `NewPromptForm` and `DuplicatePromptForm`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b3083108bf6a99ae118e4a6a1f6a556be959d3a2. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/prompts/components/prompt-detail.tsx
+++ b/web/src/features/prompts/components/prompt-detail.tsx
@@ -39,6 +39,7 @@ import { CreateExperimentsForm } from "@/src/ee/features/experiments/components/
 import { useState } from "react";
 import { useHasEntitlement } from "@/src/features/entitlements/hooks";
 import { showSuccessToast } from "@/src/features/notifications/showSuccessToast";
+import { DuplicatePromptButton } from "@/src/features/prompts/components/duplicate-prompt";
 
 export const PromptDetail = () => {
   const projectId = useProjectIdFromURL();
@@ -222,6 +223,15 @@ export const PromptDetail = () => {
                   <span className="ml-2">New version</span>
                 </div>
               </Button>
+            )}
+            {projectId && (
+              <DuplicatePromptButton
+                promptId={prompt.id}
+                projectId={projectId}
+                promptName={prompt.name}
+                promptVersion={prompt.version}
+                totalPromptCount={promptHistory.data.totalCount}
+              />
             )}
             <DetailPageNav
               key="nav"
--- a/web/src/features/prompts/components/prompt-detail.tsx
+++ b/web/src/features/prompts/components/prompt-detail.tsx
@@ -230,7 +230,6 @@ export const PromptDetail = () => {
                 projectId={projectId}
                 promptName={prompt.name}
                 promptVersion={prompt.version}
-                totalPromptCount={promptHistory.data.totalCount}
               />
             )}
             <DetailPageNav
--- a/web/src/features/prompts/server/actions/createPrompt.ts
+++ b/web/src/features/prompts/server/actions/createPrompt.ts
@@ -8,9 +8,22 @@ import { type PrismaClient } from "@langfuse/shared/src/db";
 import { LATEST_PROMPT_LABEL } from "@/src/features/prompts/constants";
 import { removeLabelsFromPreviousPromptVersions } from "@/src/features/prompts/server/utils/updatePromptLabels";
 import { updatePromptTagsOnAllVersions } from "@/src/features/prompts/server/utils/updatePromptTags";
-import { PromptService, redis } from "@langfuse/shared/src/server";
+import {
+  PromptContentSchema,
+  PromptService,
+  redis,
+} from "@langfuse/shared/src/server";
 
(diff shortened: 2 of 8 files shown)


web/src/features/prompts/components/prompt-detail.tsx
```diff
@@ -223,6 +224,15 @@ export const PromptDetail = () => {
                 </div>
               </Button>
             )}
+            {projectId && (
+              <DuplicatePromptButton
+                promptId={prompt.id}
+                projectId={projectId}
+                promptName={prompt.name}
+                promptVersion={prompt.version}
+                totalPromptCount={promptHistory.data.totalCount}
```
hassiebp, line 233: I believe `promptHistory.data.totalCount` here is the number of prompt _versions_ the selected prompt has. We are using that count to check against the usage limit in `DuplicatePromptButton`, which I think is to restrict the total number of prompt _names_?

web/src/features/prompts/server/actions/createPrompt.ts
```diff
@@ -96,3 +109,62 @@ export const createPrompt = async ({
...
+        config: jsonSchema.parse(prompt.config),
+        tags: prompt.tags,
+        projectId,
+        createdBy,
+      },
+    });
+  });
+
+  // Lock and invalidate cache for _all_ versions and labels of the prompt name
+  const promptService = new PromptService(prisma, redis);
```
hassiebp, line 159: Cache locking is not strictly necessary as we are creating a new prompt name here

web/src/features/prompts/server/actions/createPrompt.ts
```diff
@@ -96,3 +109,62 @@ export const createPrompt = async ({
...
+  const promptsDb = await prisma.prompt.findMany({
+    where: {
+      projectId: projectId,
+      name: existingPrompt.name,
+      version: isSingleVersion ? existingPrompt.version : undefined,
+    },
+  });
+
+  const createPrompts = promptsDb.map((prompt) => {
+    return prisma.prompt.create({
```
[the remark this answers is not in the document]
marliessophie: Sure, will adjust, good idea 👍

web/src/features/prompts/server/actions/createPrompt.ts
marliessophie: Alright, does that mean I can drop it? The less complexity the better :)

web/src/features/prompts/server/actions/createPrompt.ts
```diff
@@ -96,3 +109,62 @@ export const createPrompt = async ({
...
+  createdBy,
+  prisma,
+}: DuplicatePromptParams) => {
+  const existingPrompt = await prisma.prompt.findUnique({
+    where: {
+      id: promptId,
+      projectId: projectId,
+    },
+  });
+
```
hassiebp, line 127: We should add a backend check here too that `existingPrompt.name !== name` and that there is no existing prompt with that new name

web/src/features/prompts/server/actions/createPrompt.ts
hassiebp: Yes, we can drop it when we have added the check above that it is in fact a new prompt name 👍🏾

web/src/features/prompts/server/actions/createPrompt.ts
marliessophie: Great catch! Will add this check for create as well. Based on this had assumed v1 could not be deleted and hence constraint would be handled schema level given `@@unique([projectId, name, version])` -- good to double check

web/src/features/prompts/server/actions/createPrompt.ts
```diff
@@ -96,3 +109,62 @@ export const createPrompt = async ({
...
+      version: isSingleVersion ? existingPrompt.version : undefined,
+    },
+  });
+
+  const createPrompts = promptsDb.map((prompt) => {
+    return prisma.prompt.create({
+      data: {
+        name,
+        version: isSingleVersion ? 1 : prompt.version,
+        labels: prompt.labels,
```
hassiebp, line 147: We also need to set the `latest` label on the prompt if `isSingleVersion` is true (and deduplicate it if it already exists in `prompt.labels`)

PR #5121, opened 2025-01-20
chore: remove legacy/batched ingestion pipeline code

Steffen911:
> [!IMPORTANT]
> Remove legacy ingestion pipeline code and related configurations, scripts, and dependencies.
> 
>   - **Behavior**:
>     - Removed legacy ingestion pipeline code, including `EventProcessor.ts` and `index.ts` in `legacy` folder.
>     - Removed `legacyIngestionQueue` and related processing logic.
>     - Removed `tokenCountDelegate` from `processEventBatch()` in `processEventBatch.ts`.
>   - **Environment**:
>     - Removed `LANGFUSE_LEGACY_INGESTION_WORKER_CONCURRENCY` and `QUEUE_CONSUMER_LEGACY_INGESTION_QUEUE_IS_ENABLED` from `.env.prod.example` and `env.ts`.
>   - **Scripts**:
>     - Removed `models:migrate` script from `package.json` and `turbo.json`.
>     - Deleted `model-match.ts` script and related tests.
>   - **Dependencies**:
>     - Removed `js-tiktoken` from `web/package.json` and `worker/package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for a80274b8e40632a61bf8547e21d22bb4d4cdfca9. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/ingestion/processEventBatch.ts
```diff
@@ -230,14 +229,19 @@ export const processEventBatch = async (
...
+
+  if (event.type === eventTypes.SCORE_CREATE) {
+    return (
+      authScope.scope.accessLevel === "scores" ||
+      authScope.scope.accessLevel === "all"
+    );
+  }
+
+  return authScope.scope.accessLevel === "all";
 };
```
Steffen911, line 245: This should cover all the `.auth` logic from `packages/shared/src/server/ingestion/legacy/EventProcessor.ts`. As this touches auth a review and sanity check would be highly appreciated @maxdeichmann.

web/src/pages/api/public/events.ts
```diff
@@ -30,7 +29,7 @@ export default withMiddlewares({
       if (!event.body.id) {
         event.body.id = v4();
       }
-      const result = await processEventBatch([event], auth, tokenCount);
```
maxdeichmann, line 33: I think you can remove the js-tiktoken dependency from web now 🥳

.env.prod.example
```diff
@@ -252,7 +252,5 @@ OTEL_SERVICE_NAME="langfuse"
 # LANGFUSE_INGESTION_CLICKHOUSE_WRITE_BATCH_SIZE=
 # LANGFUSE_INGESTION_CLICKHOUSE_WRITE_INTERVAL_MS= 
 # LANGFUSE_INGESTION_CLICKHOUSE_MAX_ATTEMPTS=
-# LANGFUSE_LEGACY_INGESTION_WORKER_CONCURRENCY=
```
maxdeichmann, line 255: can you also remove them from the infra repo?

web/scripts/model-match.ts
```diff
@@ -1,260 +0,0 @@
-// Description: New model definitions in Langfuse are automatically applied to new observations.
-// You can optionally run this script to apply new model definitions to existing observations.
-// See docs: https://langfuse.com/docs/deployment/self-host#migrate-models
-// Execute: `npm run models:migrate`
```
Steffen911, line 4: @maxdeichmann Any concerns about removing this script? Otherwise, we'd have to keep the js-tiktoken dependency in the web container.

.env.prod.example
Steffen911: https://github.com/langfuse/infrastructure/pull/313

web/src/pages/api/public/events.ts
Steffen911: Removed in web and worker via https://github.com/langfuse/langfuse/pull/5121/commits/fdb21419733a3dad6fe162c50deb87946be4aa85 🙌

web/scripts/model-match.ts
maxdeichmann: Oh, ok. This is part of the upgrade from V1 to V2. Could you add a note to the upgrade doc (https://github.com/langfuse/langfuse-docs/blob/f292c51137c6f9947c471125c4702649fd0f9dd3/pages/self-hosting/upgrade-guides/upgrade-v1-to-v2.mdx#L88-L89) that the user needs to use a V2.x version to upgrade from v1 to v3?

maxdeichmann: If we do this, we can also remove that script.

packages/shared/src/server/ingestion/processEventBatch.ts
maxdeichmann: Yes. This is correct behavior.

web/scripts/model-match.ts
Steffen911: @maxdeichmann Will be added via https://github.com/langfuse/langfuse-docs/pull/1188.

worker/package.json
```diff
@@ -49,7 +49,6 @@
     "handlebars": "^4.7.8",
     "helmet": "^7.1.0",
     "ioredis": "^5.4.1",
-    "js-tiktoken": "^1.0.15",
```
maxdeichmann, line 52: Ok, we use tiktoken instead of js-tiktoken. Using js-tiktoken could be a performance improvement but I have struggled using it without errors in the worker container. Lets leave it as it is.

PR #5137, opened 2025-01-21
chore: remove postgres fallback for generations table calls

Steffen911:
Fixes the erroneous fallback introduced in https://github.com/langfuse/langfuse/pull/5123 by removing the queryClickhouse defaults.

----

> [!IMPORTANT]
> Remove Postgres fallback for generation table queries, using only Clickhouse.
> 
>   - **Behavior**:
>     - Removed Postgres fallback for generation table queries in `getAllGenerations` and `getDatabaseReadStream`.
>     - Now exclusively uses Clickhouse for these queries.
>   - **Code Removal**:
>     - Deleted `createGenerationsQuery` and `parseGetAllGenerationsInput` from `createGenerationsQuery.ts`.
>     - Removed related imports and logic in `getAllGenerationsSqlQuery.ts` and `handleBatchExportJob.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for a0b2db2055bad6a58bfe8ba2a446303b0a1170a4. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/generations/db/getAllGenerationsSqlQuery.ts
```diff
@@ -1,69 +1,32 @@
...
 } from "@langfuse/shared/src/server";
 
 export async function getAllGenerations({
   input,
   selectIOAndMetadata,
-  queryClickhouse,
 }: {
   input: GetAllGenerationsInput;
   selectIOAndMetadata: boolean;
-  queryClickhouse?: boolean;
```
Steffen911, line 22: IMO the problem was that this defaulted to a falsy value now that we did not provide it anymore. That means we'd fallback to Postgres execution which caused the errors which we observed.

marliessophie: Yes, this would explain the issue! Aligns with the rough hypothesis I had ie results in the expected erroneous execution path 👍

marliessophie (approved):
LGTM! 

Looks like we can also remove the `parseTraceAllFilters` function in a similar fashion, I'll create a PR to clean up given we have context right now and send it to you as FYI. Even more than 200 lines then wuhu 👍

PR #5197, opened 2025-01-24
feat(prompts): support `commit_message`

marliessophie:
> [!IMPORTANT]
> Add support for `commitMessage` in prompt-related schemas, models, API endpoints, and UI components, allowing optional inclusion and display of commit messages.
> 
>   - **Schema and Model Updates**:
>     - Add `commitMessage` field to `Prompt`, `CreateTextPromptRequest`, and `CreateChatPromptRequest` in `prompts.yml`.
>     - Update `Prompt` type in `types.ts` to include `commit_message`.
>     - Add `commit_message` column to `prompts` table in `migration.sql`.
>     - Update `schema.prisma` to include `commitMessage` in `Prompt` model.
>   - **API and Validation**:
>     - Update `openapi.yml` to include `commitMessage` in relevant API components.
>     - Update `CreatePromptSchema` and `CreatePromptTRPCSchema` in `validation.ts` to include `commitMessage`.
>   - **UI Components**:
>     - Add `commitMessage` field to `NewPromptForm` in `index.tsx`.
>     - Display `commitMessage` in `PromptVersionDiffDialog.tsx` and `prompt-detail.tsx`.
>     - Update `DiffViewer.tsx` to handle `oldSubLabel` and `newSubLabel` for displaying `commitMessage`.
>   - **Testing**:
>     - Add tests for `commitMessage` in `prompts.v2.servertest.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 4840648d01fb90070e68a72a913a05915db11fa5. It will automatically update as commits are pushed.</sup>


web/src/features/prompts/server/utils/validation.ts
hassiebp, line 1: Can't we not simply add `commitMessage` to `CreateTextPromptSchema` and `CreateChatPromptSchema`? 

True, in that case it will be added to the legacy schemas too, but that is not a problem as `commitMessage` is nullish. I would prefer that over having the `final` schema being an extension of the Legacy schemas, and it would simplify the change in this file

marliessophie: Tradeoff between simplicity and clarity IMO. This way it's clear which props are supported in v1 vs v2. Either way agree that the extension based on legacy reads strangely, if we keep like this we should create `BaseCreateTextPromptSchema` instead for readability. 

Not super strong feelings though, would that approach address your concern? ☺️

hassiebp (approved):

PR #5251, opened 2025-01-28
fix: patched openid-client lib to follow-redirects

### LINKED ISSUE #5119: bug: OAuth Custom Provider fails because next-auth does not follow redirects
### Describe the bug

When you use a custom provider via OAuth (Specifically Jumpcloud) - you get the SIGNIN_OAUTH_ERROR even though all the details are correct. I was expecting the login to succeed since I mentioned all the relevant details including the issuer url, client_id, client_secret and other relevant data as stated in the docs.

### To reproduce

Try to connect Jumpcloud as a custom OAuth provider.

I debugged the code a bit, and understood the issue - langfuse uses next-auth v4 for managing OIDC, which uses openid-client@5.6.5 to do so. This particular library does not follow redirects and that's the actual bug.

What it actually means? 
Say my issuer is https://oauth.id.jumpcloud.com - next-auth with the help of openid-client tries to resolve the .well-known/openid-configuration endpoint, via a GET request. If you try to "curl" it using `curl -I https://oauth.id.jumpcloud.com/.well-known/openid-configuration` you'll see that you get a 302 status code back and the json content for the well-known api.
Normally it would be ok, you can definitely work with that structure, but if you'll take a look at [this code snippet](https://github.com/panva/openid-client/blob/45c96f67ce0644bd829f61e82fba3dd8c051c89e/lib/helpers/process_response.js#L32) in the lib of openid-client@5.6.5, you'll see that the code checks for the status code to be ONLY 200, although 302/301,201,204 might be valid as well, and otherwise it throws an error.

I am using langfuse v3.10 image,
it uses next-auth v4.24.11 which uses openid-client@5.6.5

### Additional information

The resolution would be either to
- Try and upgrade to next-auth v5 (maybe they use the next major version of the openid-client library which is totally different)
- Try and install the openid-client v6 as a peer dependency and let next-auth use it instead of the older version

Yes

ahrakos:
fixes #5119 

@marcklingen 

I patched the library to make sure that only real error codes are being treated as an error.
Having said that, now when working with identity providers that return 301/302 status codes, the issuer should still be resolved.

Will try to build locally and deploy to our cluster, and keep you posted

----

> [!IMPORTANT]
> Patches `openid-client` to follow 301/302 redirects, ensuring identity providers returning these codes are resolved.
> 
>   - **Behavior**:
>     - Patches `openid-client` to follow 301/302 redirects, ensuring identity providers returning these codes are resolved.
>   - **Dependencies**:
>     - Adds `openid-client` patch to `patchedDependencies` in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 3586bbeff569fcdfbbcdb8b4c2624a10a5c51c63. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,10 +85,15 @@
     "overrides": {
       "jsonpath-plus": "10.0.7",
       "nanoid": "^3.3.8",
-      "katex": "^0.16.21"
+      "katex": "^0.16.21",
+      "openid-client": "5.6.5"
     },
     "patchedDependencies": {
-      "next-auth@4.24.11": "patches/next-auth@4.24.11.patch"
+      "next-auth@4.24.11": "patches/next-auth@4.24.11.patch",
+      "openid-client@5.6.5": "patches/openid-client@5.6.5.patch"
     }
+  },
+  "dependencies": {
+    "follow-redirects": "^1.15.9"
   }
 }
--- a/patches/openid-client@5.6.5.patch
+++ b/patches/openid-client@5.6.5.patch
@@ -1,0 +1,13 @@
+diff --git a/lib/helpers/request.js b/lib/helpers/request.js
+index d3dfda3b8291ab8fd06bc54201b098fbaa8a702a..af02ce8de4cac121f9894e2041065607e8d58fc6 100644
+--- a/lib/helpers/request.js
++++ b/lib/helpers/request.js
+@@ -1,7 +1,6 @@
+ const assert = require('assert');
+ const querystring = require('querystring');
+-const http = require('http');
+-const https = require('https');
++const { http, https } = require('follow-redirects');
+ const { once } = require('events');
+ const { URL } = require('url');
+ 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,15 +8,23 @@ overrides:
   jsonpath-plus: 10.0.7
   nanoid: ^3.3.8
   katex: ^0.16.21
+  openid-client: 5.6.5
 
 patchedDependencies:
   next-auth@4.24.11:
     hash: cczsf6i6qe2m2htirvgxjtoclu
     path: patches/next-auth@4.24.11.patch
+  openid-client@5.6.5:
+    hash: ppbxwvx6sjf4zweycc43cxgeu4
+    path: patches/openid-client@5.6.5.patch
 
 importers:
 
   .:
+    dependencies:
+      follow-redirects:
+        specifier: ^1.15.9
+        version: 1.15.9
     devDependencies:
       '@release-it/bumper':
         specifier: ^6.0.1
@@ -158,7 +166,7 @@ importers:
         version: 0.3.12(@langchain/core@0.3.37(openai@4.82.0(zod@3.23.8)))
       '@langchain/aws':
         specifier: ^0.1.3
-        version: 0.1.3(@aws-sdk/client-sso-oidc@3.668.0(@aws-sdk/client-sts@3.668.0))(@aws-sdk/client-sts@3.675.0)(@langchain/core@0.3.37(openai@4.82.0(zod@3.23.8)))
+        version: 0.1.3(@aws-sdk/client-sso-oidc@3.668.0(@aws-sdk/client-sts@3.668.0))(@aws-sdk/client-sts@3.668.0)(@langchain/core@0.3.37(openai@4.82.0(zod@3.23.8)))
       '@langchain/core':
         specifier: ^0.3.37
         version: 0.3.37(openai@4.82.0(zod@3.23.8))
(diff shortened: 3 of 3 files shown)


marcklingen (commented):
@Steffen911 this seems to be a potentially critical change, can you review this?

ahrakos:
@Steffen911 @marcklingen 
Also not sure what exactly failed here, something related to snyk configurations maybe?

Steffen911:
@ahrakos some of the tests rely on secrets to authenticate with third parties. We don't grant those for PRs coming from forks which makes them fail. Don't worry about those.

I'll take a look! 👀

patches/openid-client.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/helpers/process_response.js b/lib/helpers/process_response.js
+index 7dad56556f717a3e4cd72b4d4599af90538ab385..d9b9dc990784c6b9b306130dc901063a31226364 100644
+--- a/lib/helpers/process_response.js
++++ b/lib/helpers/process_response.js
+@@ -29,7 +29,7 @@ const isStandardBodyError = (response) => {
+ };
+ 
+ function processResponse(response, { statusCode = 200, body = true, bearer = false } = {}) {
+-  if (response.statusCode !== statusCode) {
++  if (response.statusCode >= 400) {
```
Steffen911, line 10: @ahrakos Did you debug whether those 301/302 codes contain a body? Looking at the remaining function, my assumption is that we just move the failure a couple lines down as we won't get the body that we anticipate (or no body at all). 

```javascript
const { STATUS_CODES } = require('http');
const { format } = require('util');

const { OPError } = require('../errors');
const parseWwwAuthenticate = require('./www_authenticate_parser');

const throwAuthenticateErrors = (response) => {
  const params = parseWwwAuthenticate(response.headers['www-authenticate']);

  if (params.error) {
    throw new OPError(params, response);
  }
};

const isStandardBodyError = (response) => {
  let result = false;
  try {
    let jsonbody;
    if (typeof response.body !== 'object' || Buffer.isBuffer(response.body)) {
      jsonbody = JSON.parse(response.body);
    } else {
      jsonbody = response.body;
    }
    result = typeof jsonbody.error === 'string' && jsonbody.error.length;
    if (result) Object.defineProperty(response, 'body', { value: jsonbody, configurable: true });
  } catch (err) {}

  return result;
};

function processResponse(response, { statusCode = 200, body = true, bearer = false } = {}) {
  if (response.statusCode !== statusCode) {
    if (bearer) {
      throwAuthenticateErrors(response);
    }

    if (isStandardBodyError(response)) {
      throw new OPError(response.body, response);
    }

    throw new OPError(
      {
        error: format(
          'expected %i %s, got: %i %s',
          statusCode,
          STATUS_CODES[statusCode],
          response.statusCode,
          STATUS_CODES[response.statusCode],
        ),
      },
      response,
    );
  }

  if (body && !response.body) {
    throw new OPError(
      {
        error: format(
          'expected %i %s with body but no body was returned',
          statusCode,
          STATUS_CODES[statusCode],
        ),
      },
      response,
    );
  }

  return response.body;
}

module.exports = processResponse;

```

ahrakos: @Steffen911

I am planning to test that tomorrow, but from my knowledge and testing things with curl, 301/302 just tells the client that the response is located at some other uri, and it's the client's responsibility to redirect and grab the response from there.

Having said that, I tested it with curl and it works - I will build from source tomorrow and will try to test and debug on our own cluster. 

Generally speaking, any non-error status code (< 400) should be ok in terms of resolving the issuer, although what makes sense is to see 2xx status codes.

Steffen911: @ahrakos It could be that https://github.com/langfuse/langfuse/commit/9f8793e26a55e8e67b0539d461c346fbd2d298fc affected the default behaviour for authChecks. Can you check whether it works if you set the authCheck variable for your IDP explicitly to include `state`? If no, please create a new issue and tag me so that we can follow-up.

Steffen911: Regarding the issue at hand: Yes, that's what I was afraid of. Let me know if that follow-redirects approach yields any results. 🤞

ahrakos: > 

Can you provide some docs on how to do that?
I looked a bit in the code and I couldn't infer of anything specific I should do if I have a custom idp configured.

ahrakos: + @Steffen911 
I tried to patch it, locally when i use pnpm patch, and pnpm patch-commit, it happens that it works and the lib is being patched. When I run the docker build command for web, it seems that the patch is created but without the changes, which is quite weird.

Is anything related to the dockerfile could prevent the patch from being applied?

Steffen911: For a custom IdP you'd need to set `AUTH_CUSTOM_CHECKS=state`.

Steffen911: @ahrakos No, we use the patch command for another next-auth update and there it propagates fine into the docker container. Did you commit your changes in git before building the docker image?

ahrakos: @Steffen911 in the beginning i didn't, you are right. But after, I realized that I should commit my changes, and I did it, but still no luck - I do see a folder with a hash named openid-client, meaning the patching worked, but the content of the files are the same as the original package.

By any chance, did you patch a library which was a transitive dependency? Because this is my case, and I am not sure if it supposed to work.

Steffen911: @ahrakos I think I did. I patched a specific version of a library though so perhaps that's the difference here? In your case there are multiple instances in the pnpm installation directory and you may have to patch all of them.

ahrakos: @Steffen911 
That's a bit weird, because locally when I run the patch and peek into the node_modules, I do see that the lib is patched and the change exists there. And I made sure no other instances of openid-client exist. When I build the web image locally using docker, and run the image with "sh" I find out that the patch does not include my changes. Could you assist on that?

ahrakos:
@Steffen911 @marcklingen 

Alright, finally this PR represents a working version with the patch.
The issue was I used the suggested command of `pnpm patch openid-client` instead of `pnpm patch openid-client@5.6.5` and adding this to the overrides so pnpm will point all usages of the lib to the same version.

Let me know how would you like to proceed from here.

Steffen911:
@ahrakos did you debug how it handles an empty body on a 3xx response code? Because my understanding from the code is that it would just fail later. What's the exact behaviour you see when you run this locally?

ahrakos:
> @ahrakos did you debug how it handles an empty body on a 3xx response code? Because my understanding from the code is that it would just fail later. What's the exact behaviour you see when you run this locally?

Did you look into the updated code? I no longer change the way the processResponse function, but I do add the capability of follow-redirects to the http client they use.

I did saw that now that we have a body, it is being returned as if it was a 200 status code and we do get it back normally.

Do you have any suggestion on further testing? Cuz so far, all the http requests that openid-client lib fires work with my patch.

ahrakos:
@marcklingen @Steffen911 
Any news regarding this?

package.json
```diff
@@ -85,10 +85,15 @@
...
+      "openid-client": "5.6.5"
     },
     "patchedDependencies": {
-      "next-auth@4.24.11": "patches/next-auth@4.24.11.patch"
+      "next-auth@4.24.11": "patches/next-auth@4.24.11.patch",
+      "openid-client@5.6.5": "patches/openid-client@5.6.5.patch"
     }
+  },
+  "dependencies": {
+    "follow-redirects": "^1.15.9"
```
Steffen911, line 97: Can we move this into the web package.json? Or does it need to live in the root package.json?

ahrakos: I'll try to play with it today, and update on that @Steffen911

ahrakos: @Steffen911 
Unfortunately, I wasn't able to make this work on the web package level. It seems that patches should be handled in the root level of pnpm, otherwise it does not respect that patch.

Steffen911:
@ahrakos Thank you for the contribution here. As you have a workaround in place according to https://github.com/langfuse/langfuse/issues/5119 and this doesn't seem a popular request given the low interactions on the issue, I will not merge the PR as is. Adding such patches adds too much risks in my view.

If we start to see more demand for this issue, I may change my view on this.

Sorry.

PR #5266, opened 2025-01-29
chore: upgrade prisma to version 6

Steffen911:
None of the changes in https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-6 seem to affect our usage.


maxdeichmann (dismissed):
We also need to install this prisma version in docker then.

maxdeichmann:
Did you run migrations on this?
https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-6#schema-change-for-implicit-m-n-relations-on-postgresql
I would expect schema changes.

Steffen911:
@maxdeichmann Where do we use implicit many-to-many relationships? I'm not aware of any. 
Instead, we use the org_membership and project_membership tables as explicit M to N mappings.

PR #5283, opened 2025-01-30
fix(rbac): load trace data in shared sessions

marliessophie:
> [!IMPORTANT]
> Update `enforceTraceAccess` in `trpc.ts` to allow access to public trace sessions.
> 
>   - **Behavior**:
>     - Update `enforceTraceAccess` middleware in `trpc.ts` to check if a trace session is public.
>     - Allows access to trace data if the session is public, even if the user is not a project member or admin.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 76814837208d02981a14948b6d9dfabb65a28723. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/server/api/trpc.ts
+++ b/web/src/server/api/trpc.ts
@@ -358,7 +358,26 @@ const enforceTraceAccess = t.middleware(async ({ ctx, rawInput, next }) => {
     .flatMap((org) => org.projects)
     .find(({ id }) => id === projectId);
 
-  if (!trace.public && !sessionProject && ctx.session?.user?.admin !== true) {
+  const traceSession = !!trace.sessionId
+    ? await ctx.prisma.traceSession.findFirst({
+        where: {
+          id: trace.sessionId,
+          projectId,
+        },
+        select: {
+          public: true,
+        },
+      })
+    : null;
+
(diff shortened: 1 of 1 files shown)


web/src/server/api/trpc.ts
```diff
@@ -358,7 +358,26 @@ const enforceTraceAccess = t.middleware(async ({ ctx, rawInput, next }) => {
...
+          id: trace.sessionId,
+          projectId,
+        },
+        select: {
+          public: true,
+        },
+      })
+    : null;
+
+  const isSessionPrivate = !!traceSession && !traceSession.public;
```
marcklingen, line 373: traceSession?.public===false would be simpler here

marcklingen: i think overall the logic here is not super simple to understand based on the variable naming, as this is security related, I would suggest to make this easier instead of the subsequent "(isSessionPrivate || !trace.sessionId)"

marliessophie: good shout, adjusted to make simpler 👍

PR #5284, opened 2025-01-30
fix(media-ui): fetch i/o media only for authed project member

marliessophie:
> [!IMPORTANT]
> Restrict media fetching to authenticated project members in `ObservationPreview` and `TracePreview`.
> 
>   - **Behavior**:
>     - Add `enabled: isAuthenticatedAndProjectMember` to `observationMedia` query in `ObservationPreview` to fetch media only for authenticated project members.
>     - Add `enabled: isAuthenticatedAndProjectMember` to `traceMedia` query in `TracePreview` to fetch media only for authenticated project members.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for dd5fa6bae383e335f02a5597a72af9465d987447. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/components/trace/ObservationPreview.tsx
+++ b/web/src/components/trace/ObservationPreview.tsx
@@ -89,6 +89,7 @@ export const ObservationPreview = ({
       projectId: projectId,
     },
     {
+      enabled: isAuthenticatedAndProjectMember,
       refetchOnWindowFocus: false,
       refetchOnMount: false,
       refetchOnReconnect: false,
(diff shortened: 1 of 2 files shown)


hassiebp (approved):
Thanks for the fix!

In general, I think we should allow fetching the linked media assets if the trace is marked public. I have created [this ticket](https://linear.app/langfuse/issue/LFE-3771/allow-fetching-media-on-public-traces) to allow that

PR #5314, opened 2025-01-31
feat(tables): add select all interface to tables

marliessophie:
> [!IMPORTANT]
> Introduces batch actions for tables, enabling select all functionality and actions like delete or add to annotation queue, with supporting UI components, server logic, and tests.
> 
>   - **Batch Actions**:
>     - Introduces `BatchActionType` enum with `Create` and `Delete` actions in `types.ts`.
>     - Adds `BatchActionQuerySchema` and `CreateBatchActionSchema` for batch action queries.
>     - Implements `BatchActionQueue` in `batchActionQueue.ts` for handling batch action jobs.
>     - Adds `BatchActionProcessingEventSchema` and `BatchActionQueue` to `queues.ts`.
>     - Implements `handleBatchActionJob` in `handleBatchActionJob.ts` to process batch actions.
>   - **UI Components**:
>     - Adds `DataTableSelectAllBanner` in `data-table-select-all-banner.tsx` for select all interface.
>     - Implements `TableActionMenu` and `TableActionDialog` for batch action UI in `TableActionMenu.tsx` and `TableActionDialog.tsx`.
>     - Updates `DataTableToolbar` to include multi-select functionality.
>   - **API and Server**:
>     - Adds `createBatchActionJob` in `createBatchActionJob.ts` to create batch action jobs.
>     - Updates `traceRouter` in `traces.ts` to handle batch delete actions.
>     - Adds `tableRouter` in `tableRouter.ts` for batch action progress queries.
>   - **Testing**:
>     - Adds tests for batch actions in `batchAction.test.ts` to verify chunk processing and filtered queries.
>   - **Miscellaneous**:
>     - Updates `useSessionStorage` to sync state across tabs.
>     - Adds `useSelectAll` hook for managing select all state in tables.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 87977bc34bbb0c966af3dcaecf05e21441b3eb2c. It will automatically update as commits are pushed.</sup>


worker/src/__tests__/selectAllService.test.ts
marliessophie, line 1: Still a draft, no need to review yet

web/src/features/table/server/tableRouter.ts
```diff
@@ -0,0 +1,152 @@
...
+              actionId,
+              query: JSON.stringify(query),
+              tableName,
+              cutoffCreatedAt: new Date(),
+              targetId,
+            },
+          })
+          .then((job) => {
+            if (redis && job.id) {
+              redis.sadd(`projectJobs:${projectId}:${tableName}`, job.id); // Store job ID under project-specific key
```
marliessophie, line 118: Alternative: create table for select all jobs instead.

worker/src/__tests__/selectAllService.test.ts
marliessophie, line 1: No need to review yet, still in draft

web/src/features/table/hooks/useTableActionMutations.ts
```diff
@@ -0,0 +1,61 @@
...
+    onSuccess: () => {
+      showSuccessToast({
+        title: "Traces deleted",
+        description: `Selected traces deleted.`,
+      });
+    },
+    onSettled: () => {
+      void utils.traces.all.invalidate();
+    },
+  });
```
marliessophie, line 33: This operation is also handled async via trace delete queue. Quite some duplication here but did not want to expand PR scope further.

worker/src/features/selectAll/handleSelectAllJob.ts
```diff
@@ -0,0 +1,139 @@
...
+      // Reset batch
+      batch = [];
+    }
+  }
+
+  // Process any remaining records
+  if (batch.length > 0) {
+    await processActionChunk(actionId, batch, projectId, targetId);
+  }
+
```
marliessophie, line 133: Not sure if this is the most performant/elegant solution. Open for feedback!

packages/shared/src/features/selectAll/constants.ts
marliessophie, line 1: Please note this PR adds multi-select only on traces table for now. Adding multi-select to [existing behaviour] annotation queue items (delete) and [new behaviour] traces (add to dataset), should be done in separate PR.

web/src/components/useSessionStorage.tsx
marliessophie, line 1: As stated in commit: session storage hook was not properly syncing value from session storage. Adjusted to ensure consistency between instances and tabs.

web/src/features/table/server/tableRouter.ts
```diff
@@ -0,0 +1,152 @@
...
+const generateSelectAllId = (
+  projectId: string,
+  actionId: string,
+  tableName: string,
+) => {
+  return `${projectId}-${tableName}-${actionId}`;
+};
+
+export const tableRouter = createTRPCRouter({
+  selectAll: protectedProjectProcedure
```
maxdeichmann, line 52: Why did you decide to have this as its own API endpoint? My intuition would have told me to extend the existing api routes such as traces.bookmark.

worker/src/features/selectAll/handleSelectAllJob.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (record?.id) {
+      records.push(record);
+    }
+  }
+
+  // 2. Process in chunks
+  const startingChunk = (selectAllJob.progress as number) || 0;
+  let chunkCount = 0;
+  for (let i = 0; i < records.length; i += CHUNK_SIZE) {
+    // Skip processing if this chunk is before our saved progress
```
maxdeichmann, line 111: The saved progress is only Redis state which is not persistent. What do you think about removing this logic? I would at most use the progress as progress indicator for some UI.

web/src/features/table/server/tableRouter.ts
```diff
@@ -0,0 +1,152 @@
...
+        // Notify worker
+        await selectAllQueue
+          .add(QueueJobs.SelectAllProcessingJob, {
+            id: selectAllId, // Use the selectAllId to deduplicate when the same job is sent multiple times
+            name: QueueJobs.SelectAllProcessingJob,
+            timestamp: new Date(),
+            payload: {
+              projectId,
+              actionId,
+              query: JSON.stringify(query),
```
maxdeichmann, line 110: Why do you stringify here? We could just pass the object here and then also be able to parse it properly on consumer side.

web/src/features/table/server/tableRouter.ts
maxdeichmann: This is to rate limit our backend, right? I think given the complexity here, we can drop this.

maxdeichmann: Do you think it would be fine to have only one job which is waiting / processing etc? If yes, we can set a deterministic jobId when adding a job. Redis will then discard a second job with the same id. [Docs](https://docs.bullmq.io/guide/jobs/deduplication)

web/src/features/table/hooks/useTableActionMutations.ts
```diff
@@ -0,0 +1,61 @@
+import { showSuccessToast } from "@/src/features/notifications/showSuccessToast";
+import { type ActionId } from "@langfuse/shared";
+import { api } from "@/src/utils/api";
+
+export const useTableActionMutations = (
```
maxdeichmann, line 5: I am wondering whether we are trying to abstract a bit too much here. My take would be to:
- have apis per action (e.g. delete, bookmark etc.). These APIs should be able to take batch requests
- these apis then also pass data to the queue if required
- In the FE, we have one component which is responsible for the selection mechanisms (bookmark all etc.). This API then takes a function which is called with the selection mechanism state. This function then implements the specific use-cases (e.g. calling mutation for bookmark all)

worker/src/features/selectAll/handleSelectAllJob.ts
marliessophie: So am I understanding correctly that on Redis restart or failure we would lose this state ie the progress indicator? 

Some context as to why I added this logic: I wanted to find a solution for job failing inbetween chunks, such that the already processed chunks would not be retried, as this might lead to unexpected behaviour. Essentially my idea was to use the progress to save which chunks had been processed and skip any chunks that might have been processed already. Does that make sense? Happy to learn how I could approach solving this otherwise.

packages/shared/src/features/selectAll/types.ts
```diff
@@ -0,0 +1,34 @@
+import z from "zod";
+import { singleFilter } from "../../interfaces/filters";
+import { orderBy } from "../../interfaces/orderBy";
+import { ACTION_ACCESS_MAP } from "./constants";
+
+// TODO: merge type with batch export table name
+export enum SelectAllTableName {
```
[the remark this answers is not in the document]
marliessophie: Fair! Will adjust

web/src/features/table/server/tableRouter.ts
marliessophie: Essentially the goal is to block users triggering a batch select operation from the front-end while a batch select operation is still in progress (ie job is processed async). To this end, we need to query if a job is currently in progress for the given project and table. 

Given the docs you shared it seems my understanding was wrong and job ids in redis must be unique _only across jobs that are not completed/failed_. So I assume it is safe to drop this logic and instead use a combination of `projectId-tableName`actionId` as jobId and simply query if this job is still in progress? That's much simpler, thank you

web/src/features/table/server/tableRouter.ts
marliessophie: Fair, we can simplify, thx, will adjust

web/src/features/table/hooks/useTableActionMutations.ts
marliessophie: I believe having apis per action will address most of your feedback, I will come back with a refactored suggestion.

web/src/server/api/routers/traces.ts
```diff
@@ -269,6 +273,8 @@ export const traceRouter = createTRPCRouter({
       z.object({
         traceIds: z.array(z.string()).min(1, "Minimum 1 trace_Id is required."),
         projectId: z.string(),
+        query: BatchActionQuerySchema.optional(),
+        isBatchAction: z.boolean().default(false),
       }),
     )
     .mutation(async ({ input, ctx }) => {
```
marliessophie, line 280: @maxdeichmann It seems off to me that we do not check trace entitlement access in the `trace.deleteMany` route -- is this intended? I also noticed we do not entitle cloud to delete traces, is this correct? Thanks!

marliessophie: Added the trace delete entitlement to the check. Are we sure no cloud plan is entitled to delete traces?

marliessophie:
Will add standard DD observability settings via terraform once merged.

PR #5350, opened 2025-02-03
fix(dataset-evals): reference `job.observationId` rather than `sourceObservationId`

marliessophie:
> [!IMPORTANT]
> Remove `sourceObservationId` handling in `createEvalJobs` in `evalService.ts`, using `observationId` from event directly.
> 
>   - **Behavior**:
>     - In `createEvalJobs` in `evalService.ts`, remove `sourceObservationId` from dataset item queries and logic.
>     - Use `observationId` from event directly, setting it to `undefined` if not present.
>   - **Database Queries**:
>     - Update SQL queries in `createEvalJobs` to exclude `sourceObservationId` from `dataset_items` and `dataset_run_items` tables.
>   - **Job Creation**:
>     - Set `jobInputObservationId` to `observationId` or `null` in job creation logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 5cc50865e87a5f80f886ed774bd2e83ec79ba1a7. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/worker/src/ee/evaluation/evalService.ts
+++ b/worker/src/ee/evaluation/evalService.ts
@@ -102,9 +102,7 @@ export const createEvalJobs = async ({
     );
 
     const isDatasetConfig = config.target_object === "dataset";
-    let datasetItem:
-      | { id: string; sourceObservationId: string | undefined }
-      | undefined;
+    let datasetItem: { id: string } | undefined;
     if (isDatasetConfig) {
       // If the target object is a dataset and the event type has a datasetItemId, we try to fetch it based on our filter
       if ("datasetItemId" in event && event.datasetItemId) {
@@ -115,9 +113,9 @@ export const createEvalJobs = async ({
         );
 
         const datasetItems = await prisma.$queryRaw<
-          Array<{ id: string; sourceObservationId: string | undefined }>
+          Array<{ id: string }>
         >(Prisma.sql`
-          SELECT id, source_observation_id as "sourceObservationId"
+          SELECT id,
           FROM dataset_items as di
           WHERE project_id = ${event.projectId}
             AND id = ${event.datasetItemId}
@@ -128,9 +126,9 @@ export const createEvalJobs = async ({
         // Otherwise, try to find the dataset item id from datasetRunItems.
         // Here, we can search for the traceId and projectId and should only get one result.
         const datasetItems = await prisma.$queryRaw<
-          Array<{ id: string; sourceObservationId: string | undefined }>
+          Array<{ id: string }>
         >(Prisma.sql`
-          SELECT dataset_item_id as id, observation_id as "sourceObservationId"
+          SELECT dataset_item_id as id
           FROM dataset_run_items as dri
           WHERE project_id = ${event.projectId}
           AND trace_id = ${event.traceId}
@@ -145,7 +143,7 @@ export const createEvalJobs = async ({
     const observationId =
       "observationId" in event && event.observationId
         ? event.observationId
-        : datasetItem?.sourceObservationId;
+        : undefined;
     if (observationId) {
       const observationExists = await checkObservationExists(
         event.projectId,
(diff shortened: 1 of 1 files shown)


worker/src/ee/evaluation/evalService.ts
```diff
@@ -102,9 +102,7 @@ export const createEvalJobs = async ({
     );
 
     const isDatasetConfig = config.target_object === "dataset";
-    let datasetItem:
-      | { id: string; sourceObservationId: string | undefined }
-      | undefined;
+    let datasetItem: { id: string } | undefined;
     if (isDatasetConfig) {
```
maxdeichmann, line 106: Why do we need this differentiation here in the first place?

marliessophie: When a dataset run item is added to the dataset upsert queue it is possible that the trace linked to this DRI does not exist yet but is still being processed in the ingestion queue. In this case, we wish to evaluate subject to the dataset configs after the trace has been successfully created ie in the trace upsert queue. This is why we check if the trace is linked to a dataset run item. This cases should not happen often.

worker/src/ee/evaluation/evalService.ts
```diff
@@ -145,7 +143,7 @@ export const createEvalJobs = async ({
     const observationId =
       "observationId" in event && event.observationId
         ? event.observationId
-        : datasetItem?.sourceObservationId;
+        : undefined;
     if (observationId) {
       const observationExists = await checkObservationExists(
```
maxdeichmann, line 148: Why do we need this check here?
My understanding: directly after traces / run items are created, we will cteate the job execution. Afterwards, we add it the the eval execution with delay. The delay should be long enough to ensure all data is there.

cc @Steffen911 as you refactored the code.

worker/src/ee/evaluation/evalService.ts
```diff
@@ -145,7 +143,7 @@ export const createEvalJobs = async ({
     const observationId =
       "observationId" in event && event.observationId
         ? event.observationId
-        : datasetItem?.sourceObservationId;
+        : undefined;
```
maxdeichmann, line 146: Will removing this break any logic downstream in the eval processing?

marliessophie: No. The `job_input_observation_id` is a nullable column on the schema level

maxdeichmann: Yes, this also means that we allow evals only on dataset runs on traces, not on observations. Correct?

worker/src/ee/evaluation/evalService.ts
Steffen911: @maxdeichmann The code is used for dataset and event triggered evals. If a dataset references an observation we need to check that it exists or risk having a "null pointer".

worker/src/ee/evaluation/evalService.ts
marliessophie: Correct, we will always run evals subject to the parent `trace_id`, not the `observation_id`

worker/src/ee/evaluation/evalService.ts
maxdeichmann: Not sure I understand this correctly: where would we get a null pointer exactly?
From my understanding, we can create the eval execution in this worker. In the eval execution, we will retry if the observation does not exist.

Steffen911: > we can create the eval execution in this worker. In the eval execution, we will retry if the observation does not exist.

As far as I understand that's exactly what this snippet of code does.

maxdeichmann: This is happening before the eval execution, right? Hence without the delay. I think we should hop on a quick call on this.

PR #5381, opened 2025-02-05
chore: optionally log s3 event bucket uploads in postgres

Steffen911:
> [!IMPORTANT]
> Adds optional logging of S3 event uploads to Postgres, with schema updates and function modifications to support this feature.
> 
>   - **Behavior**:
>     - Adds optional logging of S3 event uploads to Postgres, controlled by `LANGFUSE_S3_EVENT_UPLOAD_POSTGRES_LOG_ENABLED` in `env.ts`.
>     - Updates `uploadEventToS3` in `eventLog.ts` to log events to Postgres if enabled.
>   - **Models**:
>     - Adds `EventLog` model to `schema.prisma` and `types.ts`.
>     - Creates `event_log` table in `migration.sql` with relevant fields and indices.
>   - **Functions**:
>     - Refactors `processEventBatch` in `processEventBatch.ts` to use `uploadEventToS3`.
>     - Updates `upsertClickhouse` in `clickhouse.ts` to use `uploadEventToS3`.
>     - Modifies `evaluate` in `evalService.ts` to log scores to S3 and optionally to Postgres.
>   - **Tests**:
>     - Adds a skipped test in `ingestion-api.servertest.ts` for logging S3 uploads to Postgres.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f3039752968236141d7d1089378df93b2104c30e. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
+++ b/packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
@@ -8,6 +8,7 @@ CREATE TABLE "event_log" (
     "project_id" TEXT NOT NULL,
     "entity_type" TEXT NOT NULL,
     "entity_id" TEXT NOT NULL,
+    "event_id" TEXT NOT NULL,
 
     "trace_id" TEXT,
 
@@ -16,4 +17,4 @@ CREATE TABLE "event_log" (
 );
 
 -- CreateIndex
-CREATE UNIQUE INDEX "event_log_project_id_entity_type_entity_id_key" ON "event_log"("project_id", "entity_type", "entity_id");
+CREATE INDEX "event_log_project_id_entity_type_entity_id_idx" ON "event_log"("project_id", "entity_type", "entity_id");
--- a/packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
+++ b/packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
@@ -1,0 +1,19 @@
+-- CreateTable
+CREATE TABLE "event_log" (
+    "id" TEXT NOT NULL,
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "bucket_name" TEXT NOT NULL,
+    "bucket_path" TEXT NOT NULL,
+    "project_id" TEXT NOT NULL,
+    "entity_type" TEXT NOT NULL,
+    "entity_id" TEXT NOT NULL,
+
+    "trace_id" TEXT,
+
+    CONSTRAINT "event_log_pkey" PRIMARY KEY ("id"),
+    CONSTRAINT "event_log_project_id_fkey" FOREIGN KEY ("project_id") REFERENCES "projects"("id") ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "event_log_project_id_entity_type_entity_id_key" ON "event_log"("project_id", "entity_type", "entity_id");
--- a/packages/shared/prisma/schema.prisma
+++ b/packages/shared/prisma/schema.prisma
@@ -716,6 +716,7 @@ model EventLog {
   project    Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
   entityType String  @map("entity_type")
   entityId   String  @map("entity_id")
+  eventId    String  @map("event_id")
 
   // traceId is only optional, because we cannot guarantee its presence for observations.
   // We use it to support deletions on traces as we cannot cascade it down to scores and observations otherwise.
@@ -725,7 +726,7 @@ model EventLog {
   createdAt DateTime @default(now()) @map("created_at")
   updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
 
-  @@unique([projectId, entityType, entityId])
+  @@index([projectId, entityType, entityId])
   @@map("event_log")
 }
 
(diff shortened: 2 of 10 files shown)


packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+    "bucket_name" TEXT NOT NULL,
+    "bucket_path" TEXT NOT NULL,
+    "project_id" TEXT NOT NULL,
+    "entity_type" TEXT NOT NULL,
+    "entity_id" TEXT NOT NULL,
+
+    "trace_id" TEXT,
+
+    CONSTRAINT "event_log_pkey" PRIMARY KEY ("id"),
+    CONSTRAINT "event_log_project_id_fkey" FOREIGN KEY ("project_id") REFERENCES "projects"("id") ON DELETE CASCADE ON UPDATE CASCADE
```
maxdeichmann, line 15: This will have a performance penalty now on the database as we delete projects. If we have very large projects, we will likely run into timeouts here. I am wondering whether we should drop this constraint and then take care of this in the deletion case ourselves.

Steffen911: @maxdeichmann The expected flow would be that we delete from this table first (together with the S3 deletes) and then issue the project delete. That's how we did it for traces, observations and scores and IMO it's also applicable here.

packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "event_log_project_id_entity_type_entity_id_key" ON "event_log"("project_id", "entity_type", "entity_id");
```
maxdeichmann, line 19: Would you enforce this here? "entity_id" is the event_id, not the event_body_id? If a SDK retries, we would need to catch the constraint error. I somehow lean into allowing almost everything in here (as we do this on S3) and then take care of deduplication etc in the ingestion pipeline + CH

packages/shared/prisma/schema.prisma
```diff
@@ -706,6 +707,28 @@ model Events {
...
+  bucketPath String  @map("bucket_path")
+  projectId  String  @map("project_id")
+  project    Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+  entityType String  @map("entity_type")
+  entityId   String  @map("entity_id")
+
+  // traceId is only optional, because we cannot guarantee its presence for observations.
+  // We use it to support deletions on traces as we cannot cascade it down to scores and observations otherwise.
+  // We expect a traceId to be present in most cases.
+  traceId String? @map("trace_id")
```
maxdeichmann, line 723: Thanks for the comment! This would also mean that we cannot guarantee to delete everything from S3? We should be able to do this as this is one of the main reasons we build this.
For single trace deletions we could check in clickhouse all observations + scores and search for their entity Ids here.

packages/shared/src/server/utils/eventLog.ts
```diff
@@ -0,0 +1,55 @@
...
+  >,
+  data: Record<string, unknown>[],
+) => {
+  // We upload the event in an array to the S3 bucket grouped by the eventBodyId.
+  // That way we batch updates from the same invocation into a single file and reduce
+  // write operations on S3.
+  const bucketPath = `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}${event.projectId}/${event.entityType}/${event.entityId}/${event.id}.json`;
+  if (env.LANGFUSE_S3_EVENT_UPLOAD_POSTGRES_LOG_ENABLED === "true") {
+    try {
+      await prisma.eventLog.create({
```
maxdeichmann, line 39: I think we should do the extra effort and batch all events that we write from the web-ingestion container.

packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
Steffen911: Good catch. That combination cannot be unique. Will convert this to a regular BTree for fast lookups on the "List" replacement.

packages/shared/src/server/utils/eventLog.ts
Steffen911: So already do the optimization you had in mind? IMO that would add too much logic change as part of the performance experiment and we should already see significant improvements with this approach. Also, it breaks our current S3 pattern which will be hard to translate to self-hosters without this table.

packages/shared/prisma/schema.prisma
Steffen911: I think there are some legacy (V1) SDKs that do not send a traceId with observations. In those cases, we cannot (easily) delete the data. By having this column it should work for almost all cases quite efficiently without reading from ClickHouse and taking the ClickHouse result back to Postgres. So IMO having this here is quite effective for deletions and the gaps that we produce are only relevant for very outdated SDK versions.

packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
maxdeichmann: Ok, makes sense if we take care of this the way you described above.

packages/shared/prisma/migrations/20250204180200_add_event_log_table/migration.sql
maxdeichmann: Awesome, thanks!

packages/shared/src/server/utils/eventLog.ts
maxdeichmann: I think we talk about 2 different things. If one ingestion API call contains 3 events, i would expect one INSERT statement to postgres containing data for 3 rows. I would not do further batching.

Steffen911: As discussed offline - due to the elegance of single rights, we'll attempt it that way and adjust in case we run into throughput issues.

maxdeichmann: .. or cost issues.

PR #5393, opened 2025-02-05
chore(debt): Add TLS Support for Redis

minivolk:
This pull request introduces support for TLS connections to Redis

----

> [!IMPORTANT]
> Add TLS support for Redis connections with new environment variables and updated Redis client logic.
> 
>   - **Behavior**:
>     - Adds TLS support for Redis connections in `docker-compose.build.yml`, `docker-compose.redis-tls.yml`, and `docker-compose.yml` by introducing `REDIS_TLS_ENABLED`, `REDIS_TLS_CA`, `REDIS_TLS_CERT`, and `REDIS_TLS_KEY` environment variables.
>     - Updates `createNewRedisInstance` in `redis.ts` to include TLS options when `REDIS_TLS_ENABLED` is true.
>   - **Environment**:
>     - Adds `REDIS_TLS_ENABLED`, `REDIS_TLS_CA_PATH`, `REDIS_TLS_CERT_PATH`, and `REDIS_TLS_KEY_PATH` to `EnvSchema` in `env.ts` with default values.
>   - **Misc**:
>     - Adds `docker-compose.redis-tls.yml` for Redis with TLS configuration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 4b383773191adee1aec8f359aaed656ec1869928. It will automatically update as commits are pushed.</sup>


docker-compose.redis-tls.yml
```diff
@@ -0,0 +1,171 @@
+services:
```
Steffen911, line 1: Do we need a full recplication of the docker compose or can you update the existing one to support TLS options?

docker-compose.redis-tls.yml
```diff
@@ -0,0 +1,171 @@
...
+      --tls-protocols "TLSv1.2 TLSv1.3"
+    ports:
+      - 6379:6379
+    healthcheck:
+      test: ["CMD", "redis-cli", "--cert", "/certs/redis.crt", "--key", "/certs/redis.key", "--cacert", "/certs/ca.crt", "--tls", "ping"]
+      interval: 3s
+      timeout: 10s
+      retries: 10
+    volumes:
+      - ./certs:/certs
```
Steffen911, line 144: Just to confirm: The goal is to have TLS for the docker compose deployment? Because if you use an external Redis instance you could easily use `rediss://` in your REDIS_CONNECTION_URL to enforce SSH, right?

minivolk: Yes, you are right. I understand that rediss:// is already available for secure Redis connections, but I’d like to clarify.

Using rediss:// alone ensures that the connection is encrypted with TLS, but it does not guarantee client authentication unless certificates are explicitly used. If the Redis server is configured to accept TLS connections but does not enforce client certificate authentication, then anyone with network access can connect without proving their identity.
Some Redis servers are configured to enforce client certificate verification (tls-auth-clients yes). In these cases, the current implementation of rediss:// in Langfuse does not work, because the server rejects any connection that does not present a valid client certificate.

docker-compose.redis-tls.yml
minivolk: I initially considered adding everything directly into docker-compose.yml, but I decided to implement it in a separate docker-compose.redis-tls.yml file instead. The reason is that not everyone wants to use mTLS for connecting to the Redis server, but some users might still want clear guidance on how to configure it if needed.

That said, we could also merge everything into a single docker-compose.yml and simply add comments explaining what to enable if TLS is required. I’m open to either approach—let me know what you prefer!

minivolk: Agreed, docker-compose.redis-tls.yml seems unnecessary, so I'll remove it from the changes. 
Fully support the documentation approach!

packages/shared/src/server/redis/redis.ts
```diff
@@ -23,10 +24,23 @@ export const redisQueueRetryOptions: Partial<RedisOptions> = {
...
+
+  const tlsOptions = tlsEnabled
+    ? {
+        tls: {
+          ca: fs.readFileSync(env.REDIS_TLS_CA_PATH),
+          cert: fs.readFileSync(env.REDIS_TLS_CERT_PATH),
+          key: fs.readFileSync(env.REDIS_TLS_KEY_PATH),
+        },
+      }
+    : {};
```
Steffen911, line 37: @minivolk Looking at https://github.com/redis/ioredis?tab=readme-ov-file#tls-options there seem to be cases where one might only want to set the ca or only the cert. Do you think it makes sense to make each setting optional within the tls block (and also allow an empty one)? In that case I'd remove the defaults in the env.ts and use `.optional()` instead.

I'm also happy to add the changes in case you agree as I see that there's already some effort within this PR. Really appreciate the input and thoughts here! 🙂

Steffen911:
Some final thoughts on the configuration options. Overall this change looks good and ready to be merged. Highly appreciate the contribution!

packages/shared/src/server/redis/redis.ts
minivolk: Yes, thanks for catching that! I’ve pushed a fix to make all the fields optional within the TLS block. Appreciate your input on this!

Steffen911:
Documentation prepared in https://github.com/langfuse/langfuse-docs/pull/1283.

PR #5671, opened 2025-02-21
feat: process environment in ingestion pipeline

Steffen911:
> [!IMPORTANT]
> Add `environment` processing to ingestion pipeline, updating schemas, models, and tests to handle environment data.
> 
>   - **Behavior**:
>     - Add `environment` field to `CreateAnnotationScoreBase` in `types.ts`, `ScoreBase` in `scoreTypes.ts`, and `TraceBody` in `types.ts`.
>     - Add `EnvironmentName` schema in `types.ts` with validation for environment names.
>     - Update `ingestionApiSchema` in `types.ts` to include `environment`.
>     - Modify `upsertScore` in `scores.ts` to handle `environment`.
>     - Update `convertOtelSpanToIngestionEvent` in `index.ts` to set default `environment`.
>   - **Models**:
>     - Add `environment` field to `observationRecordBaseSchema`, `traceRecordBaseSchema`, and `scoreRecordBaseSchema` in `definitions.ts`.
>     - Update `convertPostgresTraceToInsert`, `convertPostgresObservationToInsert`, and `convertPostgresScoreToInsert` in `definitions.ts` to include `environment`.
>   - **Tests**:
>     - Add tests in `ingestion-api.servertest.ts` to verify `environment` handling in API requests.
>     - Update `IngestionService.integration.test.ts` to include `environment` in test cases.
>   - **Misc**:
>     - Update `createTrace`, `createObservation`, and `createScore` in `tracing-factory.ts` to set default `environment`.
>     - Modify `generateObservationsForPublicApi` and `generateScoresForPublicApi` to include `environment` in queries.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for c16c539ebbfd2977135f2c3aab6bd86389ab6304. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/ingestion/types.ts
```diff
@@ -118,6 +118,15 @@ export const UsageOrCostDetails = z
   .union([OpenAIUsageSchema, RawUsageOrCostDetails])
   .nullish();
 
+export const EnvironmentName = z
+  .string()
+  .max(40, "Maximum length is 40 characters")
+  .regex(
+    /^(?!langfuse-)[a-z0-9-_]+$/,
```
hassiebp, line 125: should we better disallow all `langfuse` prefixed environments? So `langfuse_` will also be disallowed

packages/shared/src/server/ingestion/types.ts
```diff
@@ -118,6 +118,15 @@ export const UsageOrCostDetails = z
...
+    "Only alphanumeric lower case characters, hyphens, and underscores are allowed, and it should not start with 'langfuse-'",
+  )
+  .default("default");
```
hassiebp, line 128: How are we handling existing traces / observations / scores currently? Will they be backfilled with `default`? Otherwise user will see for old records an empty environment, and `default` for all records after this change is live.

Thinking about this more, we could make a case to leave this as a `NULL` in db to reflect that we don't know which environment this record is coming from.

Steffen911: @hassiebp We backfilled everything to `default` and also use this as a fallback in ClickHouse if no environment is specified. This has the advantage that ClickHouse is more performant for non-nullable values and it also allows us to have consistent expectations around what the environment looks like.

PR #5764, opened 2025-02-27
fix: use aes-gcm 12 bytes for IV

maxdeichmann:
> [!IMPORTANT]
> Change `IV_LENGTH` to 12 bytes for AES-GCM in `index.ts`, updating `encrypt()` and `decrypt()` functions accordingly.
> 
>   - **Behavior**:
>     - Change `IV_LENGTH` from 16 to 12 in `index.ts` for AES-GCM encryption.
>     - Update `encrypt()` and `decrypt()` functions to use 12-byte IVs.
>   - **Misc**:
>     - No changes to the logic of encryption or decryption, only the IV length is adjusted.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 162779b976b6f4ecc434eb1e92e9542b7b701f86. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/encryption/index.ts
+++ b/packages/shared/src/encryption/index.ts
@@ -2,7 +2,7 @@ import crypto from "crypto";
 import { env } from "../env";
 
 const ENCRYPTION_KEY: string | undefined = env.ENCRYPTION_KEY; // Must be 256 bits (32 bytes, 64 hex characters)
-const IV_LENGTH: number = 16; // For AES, this is always 16
+const IV_LENGTH: number = 12; // For AES-GCM, this is always 16
 
 // Alternatively: openssl rand -hex 32
 export function keyGen() {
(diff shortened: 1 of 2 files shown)


packages/shared/src/encryption/index.ts
```diff
@@ -2,7 +2,7 @@ import crypto from "crypto";
 import { env } from "../env";
 
 const ENCRYPTION_KEY: string | undefined = env.ENCRYPTION_KEY; // Must be 256 bits (32 bytes, 64 hex characters)
-const IV_LENGTH: number = 16; // For AES, this is always 16
+const IV_LENGTH: number = 12; // For AES-GCM, this is always 16
```
hassiebp, line 5:
```suggestion
const IV_LENGTH: number = 12; // For AES-GCM, this is always 12
```

hassiebp (approved):

PR #5765, opened 2025-02-27
feat: add environment filter

Steffen911:
> [!IMPORTANT]
> Add environment filter feature to dashboard and tables, enabling data filtering by environment across backend and frontend components.
> 
>   - **Backend**:
>     - Add `getEnvironmentsForProjectAndTimeFilter` function in `environments.ts` to fetch distinct environments.
>     - Add `environmentFilterOptions` endpoint in `projectsRouter.ts` to provide environment options.
>     - Modify `trace-sessions.ts` to handle environment filtering in session queries.
>   - **Frontend**:
>     - Add `useEnvironmentFilter` hook in `use-environment-filter.tsx` to manage environment selection state.
>     - Update `DataTableToolbar` in `data-table-toolbar.tsx` to include environment filter UI.
>     - Integrate environment filter in various table components like `observations.tsx`, `scores.tsx`, `sessions.tsx`, and `traces.tsx`.
>     - Update dashboard components like `ChartScores.tsx`, `LatencyChart.tsx`, and `ModelUsageChart.tsx` to support environment filtering.
>   - **UI Enhancements**:
>     - Add `MultiSelect` component for environment selection in `data-table-toolbar.tsx` and `index.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e2799de7ad196f48b823f0ae8250249a3bbf3016. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/trace-sessions.ts
marliessophie, line 1: Should I review this in detail? Have not yet

marliessophie (commented):
Checking out UI locally now, overall looks good 

Comments on UI: 
- filter width is not responsive in design, UI is wanky on dashboard (filter too wide, corner of time filter cut off). You gave it a width constraint when used in the table, I would make this consistent or let it shrink to it's content dynamically. I would also make the min width smaller, maybe `100px`, right now it feels to wide. Lmk if you have questions. 
- Globally I would order the filter after the time filter, I find it especially confusing coming first before the search, wdyt? 
Screenshot 2025-03-04 at 13 26 35

packages/shared/src/server/repositories/trace-sessions.ts
Steffen911: @marliessophie No, all good. The backend logic makes sense IMO and was sufficiently covered by my tests. Thank you!

marliessophie: That's what I figured :)

web/src/components/table/use-cases/observations.tsx
```diff
@@ -177,12 +181,36 @@ export default function ObservationsTable({
       ]
     : [];
 
-  const filterState = inputFilterState.concat([
-    ...dateRangeFilter,
-    ...promptNameFilter,
-    ...promptVersionFilter,
-    ...modelIdFilter,
-  ]);
+  const environmentFilterOptions =
```
maxdeichmann, line 184: image

I think we can do better here:
- We could disable API calls .all, .count, .filterOptions as long as .environmentFilterOptions did not return
- Even better: we pull the environmentFilterOptions API call into a hook. Here we could merge data from local storage and from the API and pass it to the filter. In most cases, we would fetch envs from the localstorage (fast). If the envs change, we would have a refetch (rare). If the env endpoint did not return and the localstorage is empty, we would wait for the env api to return.

maxdeichmann: Please consult with Marlies if you think this will increase speed here.

Steffen911: @maxdeichmann I consulted with @marliessophie and did the first one. Now, we only load the other elements once environmentFilterOptions are there. I'll check what performance we achieve in prod and check for further optimizations if it's not acceptable.

Marlies said that trpc automatically caches results, i.e. there is no need to perform the additional operation via the local storage.

maxdeichmann: Yes, makes sense. Thanks!

PR #5865, opened 2025-03-05
chore: consistent casing of README files

klikstermkd:
## What does this PR do?

Fixes the link of the README for EE in `CONTRIBUTING.md` to point to the correct file: https://github.com/langfuse/langfuse/blob/main/ee/Readme.md.

Another option would be to rename the original file to `README.md` to better align with the convention of these files (being uppercase), unless it would break other places where the existing `ee/Readme.md` file is referenced. 

## Type of change

## Mandatory Tasks

----

> [!IMPORTANT]
> Fixes broken link in `CONTRIBUTING.md` to point to `ee/Readme.md`.
> 
>   - **Documentation**:
>     - Fixes broken link in `CONTRIBUTING.md` to point to `ee/Readme.md` instead of `ee/README.md`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for fe38af4a57291282a6017a22bb523bc2b6ae40fd. It will automatically update as commits are pushed.</sup>


marcklingen:
Thanks for the contribution!

> Another option would be to rename the original file to README.md to better align with the convention of these files (being uppercase), unless it would break other places where the existing ee/Readme.md file is referenced.

Agree that this should be consistent, fixed the casing across READMEs

marcklingen:
You can fix this via `git config core.ignorecase`, on macos this is `true` by default and thereby ignores these kinds of renamings

klikstermkd:
> You can fix this via `git config core.ignorecase`, on macos this is `true` by default and thereby ignores these kinds of renamings

Thanks, that one worked!

Created a follow-up PR: https://github.com/langfuse/langfuse/pull/5900

PR #5884, opened 2025-03-06
chore(batch-action): make `entitlement` prop optional

marliessophie:
> [!IMPORTANT]
> Make `entitlement` property optional in `TableAction` type and update hooks and components to handle optional entitlements.
> 
>   - **Behavior**:
>     - Add `useOptionalEntitlement` hook in `hooks.ts` to return `true` if no entitlement is provided.
>     - Replace `useHasEntitlement` with `useOptionalEntitlement` in `TableActionDialog.tsx` and `TableActionTargetOptions.tsx`.
>   - **Types**:
>     - Make `entitlement` property optional in `TableAction` type in `types.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 724e5b4c55d679eade62ae8f7cacc24136ad26a3. It will automatically update as commits are pushed.</sup>


marliessophie:
@maxdeichmann small but would like your opinion on this. 

Context: to make this interface truly reusable we need to allow users to specify project scope and entitlement. However, not all actions are gated via an entitlement. 

Question: Would you rather: 
- add a new entitlement for all self-hosted and cloud plans for `batch-actions`
- add a conditional entitlement hook as in example here that preserves react hook linked-list
- other option? 

thanks!

maxdeichmann:
> @maxdeichmann small but would like your opinion on this.
> 
> Context: to make this interface truly reusable we need to allow users to specify project scope and entitlement. However, not all actions are gated via an entitlement.
> 
> Question: Would you rather:
> 
> * add a new entitlement for all self-hosted and cloud plans for `batch-actions` and use this as default value in the regular `useHasEntitlement` hook if entitlement is undefined
> * add a conditional entitlement hook as in example here that preserves react hook linked-list
> * other option?
> 
> thanks!

Which action does not have an entitlement? Deletions? I think your solution is fine. If the batch action does not require an entitlement, we should not check for it. We are not breaking anything here, right?

PR #5982, opened 2025-03-12
fix(ui): TableWithMetadataWrapper uses contain-layout on its root div so that EvaluatorDetail overflow issue is mitigated

davidmc971:
EDIT: Previous workaround reverted. Now using contain-layout on root div of TableWithMetadataWrapper so that the EvaluatorDetail overflow issue is mitigated. Checked for side effects on the other components using TableWithMetadataWrapper and all work as expected.

### Before

image
image

### After

image

----

> [!IMPORTANT]
> Adds `contain-layout` class to `TableWithMetadataWrapper` root `div` to fix overflow issue in `EvaluatorDetail`.
> 
>   - **UI Fix**:
>     - Adds `contain-layout` class to root `div` in `TableWithMetadataWrapper` to mitigate overflow issue in `EvaluatorDetail`.
>     - Verified no side effects on other components using `TableWithMetadataWrapper`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 30ece6e49e63e5009628c6782923f231dc582ce9. It will automatically update as commits are pushed.</sup>


davidmc971:
As discussed with @marcklingen this workaround is too complex, I will look deeper and update the PR accordingly

davidmc971:
@marcklingen Just updated the PR with a much cleaner, single line fix 😁

marcklingen (commented):
@marliessophie please have a look, `contain` could be a clean solution here but can also have side-effects if i understand this correctly. You are the expert here on all things css

marliessophie:
I rebased the PR myself as a lot of the code had moved. I reviewed and `contain-layout` seems a good solution for this situation! It affects the scrolling context within the hierarchy which is exactly what we want. Typically important to be aware of side effects such as absolute positioning referring this container by default (not relevant here) and usually used with overflow-hidden (as is here). Good job!

maxdeichmann (approved):

PR #6006, opened 2025-03-13
feat: add custom-query backend

Steffen911:
> [!IMPORTANT]
> Introduces a self-serve dashboard backend POC with a new query builder for ClickHouse, including SQL injection tests and a query playground.
> 
>   - **Query Builder**:
>     - Introduces `QueryBuilder` class in `queryBuilder.ts` for building ClickHouse queries.
>     - Supports dimensions, metrics, filters, and time dimensions.
>     - Handles SQL injection prevention and parameterization.
>   - **Views and Models**:
>     - Defines views in `dataModel.ts` for `traces`, `observations`, `scores-numeric`, and `scores-categorical`.
>     - Maps legacy UI table filters to views in `dashboardUiTableToViewMapping.ts`.
>   - **API and Execution**:
>     - Adds `executeQuery` function in `dashboard-router.ts` for executing queries using the new query builder.
>     - Updates `TracesBarListChart.tsx` to use `executeQuery` for fetching data.
>   - **Testing**:
>     - Adds `queryBuilderDashboards.servertest.ts` for testing query builder against legacy functions.
>     - Adds `queryBuilderSQLI.servertest.ts` for testing SQL injection vulnerabilities.
>   - **UI**:
>     - Introduces `queryPlayground.tsx` for testing and visualizing queries.
>     - Updates `index.tsx` to integrate new query logic into the dashboard.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e524fa3afc62395479d5842403197bf9c956e7f6. It will automatically update as commits are pushed.</sup>


web/src/pages/project/[projectId]/queryPlayground.tsx
```diff
@@ -0,0 +1,130 @@
+// TODO: Remove before merge
```
Comment (Steffen911) [inline web/src/pages/project/[projectId]/queryPlayground.tsx:1]: This will be removed before merging.

web/src/features/dashboard/server/dashboard-router.ts
```diff
@@ -281,3 +297,42 @@ const extractTimeSeries = (groupBy?: z.infer<typeof groupByInterface>) => {
...
 };
+
+/**
+ * Execute a query using the QueryBuilder.
+ *
+ * @param projectId - The project ID
+ * @param query - The query configuration as defined in QueryType
+ * @returns The query result data
+ */
+export async function executeQuery(
```
Steffen911, line 308: Check if we need to propagate projectId to all joined queries.

web/src/features/query/server/dataModel.ts
```diff
@@ -0,0 +1,369 @@
...
+  measures: {
+    count: {
+      sql: "count(*)",
+      alias: "count",
+      type: "count",
+    },
+    observationsCount: {
+      sql: "uniq(observations.id)",
+      alias: "observations_count",
+      type: "count",
```
Steffen911, line 59: What do types do? Check if they have any semantic or are necessary for frontend.

web/src/features/query/server/types.ts
```diff
@@ -0,0 +1,116 @@
+import { z } from "zod";
+
+export type ViewDeclarationType = z.infer<typeof viewDeclaration>;
+export type DimensionsDeclarationType = z.infer<
+  typeof viewDeclaration
+>["dimensions"];
+
+// TODO: Do we want to type filters here to provide specific operators for strings/arrays/numbers?
+// IMO having one filter type that covers everything could be perfectly fine.
+export const filter = z.object({
```
Steffen911, line 10: Check if we can reuse the existing filter definitions.

web/src/features/dashboard/server/dashboard-router.ts
Steffen911: Behaviour according to explain indexes is exactly the same. So it doesn't seem like we get additional selectivity by including the project_id filter for sub-conditions. Hence, I'd keep it as is and rely on the filter conditions.

packages/shared/src/server/clickhouse/schema.ts
```diff
@@ -2,6 +2,11 @@ export const ClickhouseTableNames = {
   traces: "traces",
   observations: "observations",
   scores: "scores",
+
+  // Virtual tables for dashboards
+  // TODO: Check if we can do this more elegantly
+  scores_numeric: "scores_numeric",
+  scores_categorical: "scores_categorical",
```
Steffen911, line 9: @maxdeichmann What do you think about this approach to support the new "virtual" views? Do you see any risks if we allow table names here that don't exist physically?

Those get used within the FilterState logic that I reuse for the dashboards.

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -20,17 +21,50 @@ export const TracesBarListChart = ({
...
+    toTimestamp,
+    orderBy: null,
+  };
+
+  const totalTraces = api.dashboard.executeQuery.useQuery(
     {
       projectId,
-      from: "traces",
-      select: [{ column: "traceId", agg: "COUNT" }],
-      filter: timeFilter,
```
Steffen911, line 32: For some reason we ignore all other filters here?

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -20,17 +21,50 @@ export const TracesBarListChart = ({
...
+      filters.push(filter);
+    }
+  });
+
+  // Total traces query using executeQuery
+  const totalTracesQuery: QueryType = {
+    view: "traces",
+    dimensions: [],
+    metrics: [{ measure: "count", aggregation: "count" }],
+    filters,
```
Steffen911, line 57: Note: Some of the existing filters are incompatible with the new data model. If the user filters for `Trace Name` within the UI, this query will fail. I could either attempt a best-effort mapping of columns per chart type or we enable the new way to query dashboards only for admin users like we did with ClickHouse back in November.

My current plan would be to create a util function that maps `dashboardColumnDefinitions` to the new dataModel as long as we need it and use it here.

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -20,17 +21,50 @@ export const TracesBarListChart = ({
...
+      if (filter.operator === ">=" || filter.operator === ">") {
+        fromTimestamp = filter.value.toISOString();
+      } else if (filter.operator === "<=" || filter.operator === "<") {
+        toTimestamp = filter.value.toISOString();
+      }
+    } else {
+      // Keep all other filters
+      filters.push(filter);
+    }
+  });
```
Steffen911, line 50: For all of the charts we're separating timeFilters from remaining filters now. I think we should separate this logic here into a util that we can reuse across multiple chart types once we migrate others.

web/src/features/query/dataModel.ts
```diff
@@ -0,0 +1,371 @@
...
+  dimensions: {
+    id: {
+      sql: "id",
+      type: "string",
+    },
+    traceId: {
+      sql: "trace_id",
+      type: "string",
+    },
+    traceName: {
```
Steffen911, line 105: ToDo: Also add user_id, session_id, bookmarked, public, etc. from traces here as relationship dimensions.

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const viewDeclaration = z.object({
+  name: z.string(),
+  // This is the basic statement that we query from. Usually, this should be the view_name + FINAL or a more complex subquery.
+  baseCte: z.string(),
+  dimensions: z.record(
+    z.object({
+      sql: z.string(),
+      alias: z.string().optional(),
+      type: z.enum(["string", "number", "bool"]),
```
Steffen911, line 17: Not used right now, but should become relevant once we have the frontend builder to highlight applicable filter options.

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+      relationTable: z.string().optional(),
+    }),
+  ),
+  measures: z.record(
+    z.object({
+      sql: z.string(),
+      alias: z.string().optional(),
+      type: z.enum(["count", "sum", "number"]),
```
Steffen911, line 25: Same thing here. Unclear how they will remain in the final version, but pretty confident that we'll need them.

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+    toTimestamp: string,
+  ): string {
+    const from = new Date(fromTimestamp);
+    const to = new Date(toTimestamp);
+    const diffMs = to.getTime() - from.getTime();
+    const diffDays = diffMs / (1000 * 60 * 60 * 24);
+
+    // Choose appropriate granularity based on date range to get ~50 buckets
+    if (diffDays < 1) {
+      return "minute"; // Less than a day, use minutes
```
Steffen911, line 322: This may need some refinement. Perhaps minutes only makes sense for <6h.

web/src/features/dashboard/components/TracesBarListChart.tsx
maxdeichmann: This is added to the filter in the root component of the dashboard Instead of adding the times to the filters, we can also pass fromTs and toTs on.

maxdeichmann: https://github.com/langfuse/langfuse/blob/11046570f9b0b43bb27d9625049c15195367cd91/web/src/pages/project/%5BprojectId%5D/index.tsx#L141

web/src/features/dashboard/components/TracesBarListChart.tsx
maxdeichmann: I would also create an util here.

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -62,14 +103,14 @@ export const TracesBarListChart = ({
...
-      })
-    : [];
+  // Transform the data to match the expected format for the BarList
+  const transformedTraces =
+    traces.data?.map((item: any) => {
+      return {
+        name: item.name ? (item.name as string) : "Unknown",
+        value: Number(item.count_count),
+      };
+    }) ?? [];
```
maxdeichmann, line 113: This will only get more complex over time. We need am utility to do this in a generic way.

web/src/features/dashboard/server/dashboard-router.ts
```diff
@@ -281,3 +297,42 @@ const extractTimeSeries = (groupBy?: z.infer<typeof groupByInterface>) => {
...
+): Promise<Array<Record<string, unknown>>> {
+  try {
+    // Initialize query builder with ClickHouse client
+    const queryBuilder = new QueryBuilder();
+
+    // Build the query
+    const { query: compiledQuery, parameters } = queryBuilder.build(
+      query,
+      projectId,
+    );
```
maxdeichmann, line 320: please also add the tags.

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+   * ```
+   */
+  public build(
+    query: QueryType,
+    projectId: string,
+  ): { query: string; parameters: Record<string, unknown> } {
+    // Run zod validation
+    const parseResult = queryModel.safeParse(query);
+    if (!parseResult.success) {
+      throw new Error(
```
maxdeichmann, line 663: Do we need different errors for error handling?

web/src/features/dashboard/server/dashboard-router.ts
maxdeichmann: Security looks also good here.

web/src/features/query/server/queryBuilder.ts
maxdeichmann: Why did you not copy the logic that I created here?

web/src/features/query/server/dataModel.ts
maxdeichmann: What about this one here?

web/src/__tests__/queryBuilderDashboards.servertest.ts
```diff
@@ -0,0 +1,902 @@
...
+      stats.environmentCounts[trace.environment] =
+        (stats.environmentCounts[trace.environment] || 0) + 1;
+    });
+    stats.productionTraces = stats.environmentCounts["production"] || 0;
+    stats.developmentTraces = stats.environmentCounts["development"] || 0;
+    stats.stagingTraces = stats.environmentCounts["staging"] || 0;
+
+    // Count traces by name
+    traces.forEach((trace) => {
+      stats.traceCounts[trace.name] = (stats.traceCounts[trace.name] || 0) + 1;
```
maxdeichmann, line 202: there is a trace.name can be null

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+      relationTable: z.string().optional(),
+    }),
+  ),
+  tableRelations: z.record(
+    z.object({
+      name: z.string(),
+      joinCondition: z.string(),
```
maxdeichmann, line 32: I would propose to take rename all SQL strings to ...SQL. In this case it would be joinConditionSql

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+
+      // Add filters to the filter list
+      filterList.push(...fromFilter, ...toFilter);
+
+      relationJoins.push(joinStatement);
+    }
+    return relationJoins;
+  }
+
+  private buildWhereClause(
```
maxdeichmann, line 295:
```suggestion
  private buildWhereClauseSqlString(
```

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+      );
+      dimensions += `any(${timeDimensionSql}) as time_dimension,`;
+    }
+
+    return dimensions;
+  }
+
+  private buildInnerMetricsPart(appliedMetrics: any[]) {
+    return appliedMetrics.length > 0
+      ? `${appliedMetrics.map((metric) => `${metric.sql} as ${metric.alias || metric.sql}`).join(",\n")}`
```
maxdeichmann, line 392: Alias can be set by the user, correct? Does that introduce a SQL injection vector? I dont think so but also not 100% sure.

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+        return `toStartOfMonth(${sql})`;
+      default:
+        throw new Error(
+          `Invalid time granularity: ${granularity}. Must be one of minute, hour, day, week, month`,
+        );
+    }
+  }
+
+  private buildInnerDimensionsPart(
+    appliedDimensions: any[],
```
maxdeichmann, line 354: why is this any?

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+    if (hasTimeDimension) {
+      dimensions += `time_dimension,`;
+    }
+
+    return dimensions;
+  }
+
+  private buildOuterMetricsPart(appliedMetrics: any[]) {
+    return appliedMetrics.length > 0
+      ? `${appliedMetrics.map((metric) => `${this.translateAggregation(metric.aggregation)}(${metric.alias || metric.sql}) as ${metric.aggregation}_${metric.alias || metric.sql}`).join(",\n")}`
```
maxdeichmann, line 438: Here we use the alias as SQL field name. We always need to take it from definitions and cannot take it from users directly

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+} from "../types";
+import { viewDeclarations } from "@/src/features/query/dataModel";
+import {
+  FilterList,
+  createFilterFromFilterState,
+} from "@langfuse/shared/src/server";
+
+export class QueryBuilder {
+  private translateAggregation(aggregation: string): string {
+    switch (aggregation) {
```
maxdeichmann, line 17: can you add an exhaustive typecheck here

web/src/features/query/server/queryBuilder.ts
Steffen911: @maxdeichmann No, the alias is configured by us in the data model.

web/src/features/dashboard/components/TracesBarListChart.tsx
Steffen911: @maxdeichmann I think this will become generic once we have the dashboarding templates. There, we decide where which property is going to be mapped. I.e. I don't think we need to action this here.

web/src/features/dashboard/server/dashboard-router.ts
Steffen911: Added in https://github.com/langfuse/langfuse/pull/6006/commits/80cb584618be1a27223f13ace85454164657871f.

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  segments: z.array(singleFilter),
+  timeDimension: z.string(),
+});
+
+export const stringDateTime = z.string().datetime({ offset: true });
+
+export const views = z.enum([
+  "traces",
+  "observations",
+  "scores-numeric",
```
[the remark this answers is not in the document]
Steffen911: @maxdeichmann It's an idea @marcklingen and me had. The rational is that it makes aggregations straightforward on the `value` property since we don't have to consider numeric vs string/bool types. We have many cases where a Numeric vs Categorical breakdown happens anyway within the UI and this makes it possible on the data modelling.

Do you see any case where I'd query a numeric and categorical score within the same query aside from the number of total scores?

web/src/features/query/server/queryBuilder.ts
Steffen911: @maxdeichmann All errors thrown within this function are essentially "BadRequests". So I don't think that we need additional separation here.

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  ),
+  tableRelations: z.record(
+    z.object({
+      name: z.string(),
+      joinCondition: z.string(),
+      timeDimension: z.string(),
+    }),
+  ),
+  // Segments are used to apply "constant" filters to the query. For example, if we only want one type of observations.
+  segments: z.array(singleFilter),
```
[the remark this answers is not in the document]
Steffen911: Filters are supplied by users, segments are filters supplied by us during the data modelling. Their main use is to support categorical vs numerical scores.

web/src/features/query/server/queryBuilder.ts
```diff
@@ -0,0 +1,773 @@
...
+      };
+
+      // Add relation-specific timestamp filters
+      const fromFilter = createFilterFromFilterState(
+        [
+          {
+            column: relation.timeDimension,
+            operator: ">=",
+            value: new Date(query.fromTimestamp),
+            type: "datetime",
```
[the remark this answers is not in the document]
Steffen911: @maxdeichmann No, all time based filters are global for this use-case. My assumption is that if I filter a dashboard from t1 to t2, I just want to see the results for all entities within that window without any additional modifications.

web/src/features/query/types.ts
Steffen911: Updated in https://github.com/langfuse/langfuse/pull/6006/commits/ca148dffedd63162b1c1cad0554033cc198e0933

web/src/features/query/server/queryBuilder.ts
Steffen911: I don't see how this makes a big difference in understanding and if we do it, we'd have to touch many other methods that build SQL as well. So I prefer not to do it unless you have a strong preference.

web/src/features/query/server/queryBuilder.ts
Steffen911: @maxdeichmann alias does not exist on the user inputs.

web/src/features/query/server/queryBuilder.ts
Steffen911: Done in https://github.com/langfuse/langfuse/pull/6006/commits/1092295eb5be1b38892b3210f3649d92e6dc8386

web/src/features/query/server/queryBuilder.ts
Steffen911: Updated typing in https://github.com/langfuse/langfuse/pull/6006/commits/1092295eb5be1b38892b3210f3649d92e6dc8386.

web/src/features/query/types.ts
```diff
@@ -0,0 +1,102 @@
...
+    timeDimension: z
+      .object({
+        // TODO: We may want to extend this and allow custom intervals like 3h in the future.
+        // auto tries to bin the data into approximately 50 buckets given the time range
+        granularity: z.enum(["auto", "minute", "hour", "day", "week"]),
+      })
+      .nullable(),
+    fromTimestamp: stringDateTime,
+    toTimestamp: stringDateTime,
+    orderBy: z
```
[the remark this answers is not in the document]
Steffen911: @maxdeichmann I think we should stick to the `field` naming here which requires the new order by. I feel that on the API field + direction are clearer than column + order.

web/src/features/query/server/dataModel.ts
Steffen911: I'll revisit those once I built the frontend. IMO we can change the data model if we like and it doesn't have any negative impacts if we move them around.

web/src/pages/project/[projectId]/queryPlayground.tsx
Comment (Steffen911) [reply web/src/pages/project/[projectId]/queryPlayground.tsx]: Done via https://github.com/langfuse/langfuse/pull/6006/commits/a0d5f1d5cd19082d32313362a179e6aa0a80b8ca.

web/src/features/dashboard/components/TracesBarListChart.tsx
Steffen911: Added in https://github.com/langfuse/langfuse/pull/6006/commits/5cd78403bec9bd352e6ddaedcaff365065eed6c6

web/src/features/dashboard/components/TracesBarListChart.tsx
Steffen911: Done in https://github.com/langfuse/langfuse/pull/6006/commits/f60b372af040ef93101fc26d142c25a10023e2e4

web/src/features/query/server/queryBuilder.ts
Steffen911: For simplicity, I wanted to limit this to minutes, hours, days, weeks and month as we do on the granularity decision. If that turns out to be too coarse to make useful charts, I'll switch to your bucketing.

maxdeichmann: I would opt to take my solution. I think it is a bit more powerful + we reduce code duplication.

PR #6039, opened 2025-03-17
feat(evals): Extend job configuration filter options; Evaluators can run based on environments

davidmc971:
- Can now run evaluators based on environments:
  image
  image
- Fixed MultiSelect overflow issue and removed its display of empty string options (checked for side effects, all good).
  image
- Adapted layout of the MultiSelect rendering of already selected options (not optimal, but cleaner).
  image

----

> [!IMPORTANT]
> Extend evaluator filtering to include environments, update UI components, and add tests for new functionality.
> 
>   - **Behavior**:
>     - Evaluators can now be filtered and run based on environments, with the environment data being fetched and used in `tracesTableColsWithOptions()` in `tracesTable.ts` and `inner-evaluator-form.tsx`.
>     - MultiSelect component in `multi-select.tsx` updated to prevent overflow and handle empty string options.
>   - **UI Changes**:
>     - Added environment column to `tracesTable.tsx` and updated `InnerEvaluatorForm` to include environment filter options.
>     - Adjusted layout of MultiSelect rendering for selected options.
>   - **Tests**:
>     - Added `evalService.filtering.test.ts` to test evaluator creation based on various filters including environment.
>     - Updated `evalService.test.ts` to ensure evaluators are created correctly with new environment filtering.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 998dda40801091390ae72766e134222ce6c82f73. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/traces.ts
```diff
@@ -164,12 +165,16 @@ export const traceRouter = createTRPCRouter({
           projectId: input.projectId,
           filter: timestampFilter ? [timestampFilter] : [],
         }),
+        getEnvironmentsForProject({
+          projectId: input.projectId,
+        }),
```
Steffen911, line 170: Why do we need this here? There is a second call for environmentFilterOptions on the project router that does the same thing and that we can use across the application.

worker/src/__tests__/evalService.test.ts
```diff
@@ -887,6 +893,960 @@ describe("eval service tests", () => {
...
+              properties: { score: { type: "number" } },
+            },
+          },
+        });
+
+        await use(evalTemplate);
+      },
+    });
+
+    filterTest(
```
Steffen911, line 929: That's a pattern I haven't seen before. What's the idea here and why is it useful?

Could we cover this via a `it.each` instead to avoid some of the repetitive setups instead?

davidmc971:
Tests for filtering are now isolated, all 12 tests run for about 8 seconds concurrently as opposed to 22 seconds serially on my machine and fixtures take care of interdependent values that need to be isolated per test.

web/src/features/filters/hooks/useTraceFilterOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+        refetchOnWindowFocus: false,
+        refetchOnReconnect: false,
+        staleTime: Infinity,
+      },
+    );
+
+  const traceFilterOptions = useMemo(() => {
+    if (
+      !traceFilterOptionsResponse.data ||
+      !environmentFilterOptionsResponse.data
```
maxdeichmann, line 30: We can do one more refactoring here:
- traces table also fetches the environment. However, the environment API call is much faster thant the filteroptions api call.
- this function only returns if either api call returns

Can you change the behavior here to already return the environment filter if the filterOptions did not return yet? Then you could refactor the traces table.

maxdeichmann (approved):

davidmc971:
The new hook has one side-effect: The traces table has environment filtering in both the toolbar and the filter builder (but there with extra options, similar to the timestamp filtering):

image

Any thoughts on that @maxdeichmann?

maxdeichmann:
> The new hook has one side-effect: The traces table has environment filtering in both the toolbar and the filter builder (but there with extra options, similar to the timestamp filtering):
> 
> image
> 
> Any thoughts on that @maxdeichmann?

Ok, this is not great. What do you think is the best way? Remove the hook or filter out the environment form the filter drop down in the traces table? Both are fine for me.

davidmc971:
> > The new hook has one side-effect: The traces table has environment filtering in both the toolbar and the filter builder (but there with extra options, similar to the timestamp filtering):
> > image
> > Any thoughts on that @maxdeichmann?
> 
> Ok, this is not great. What do you think is the best way? Remove the hook or filter out the environment form the filter drop down in the traces table? Both are fine for me.

I took out the environment column from the traces filter builder, as this closely aligns with the other tables.

davidmc971:
I also removed the hook now as the re-use is no longer beneficial as environment fetching is no longer needed for the filter builder on the traces table.

PR #6043, opened 2025-03-17
feat(api): add new patch route for traces

Steffen911:
> [!IMPORTANT]
> Add new API endpoints for updating trace properties and managing trace tags, with corresponding schema changes and tests.
> 
>   - **API Endpoints**:
>     - Add `PATCH /api/public/traces/{traceId}` to update trace properties (`bookmarked`, `public`, `tags`).
>     - Add `POST /api/public/traces/{traceId}/tags` to add a tag to a trace.
>     - Add `DELETE /api/public/traces/{traceId}/tags/{tagId}` to remove a tag from a trace.
>   - **Schema Changes**:
>     - Add `PatchTraceV1Body`, `PostTraceTagsV1Body`, and `DeleteTraceTagV1Query` in `traces.ts`.
>   - **Implementation**:
>     - Implement new routes in `[traceId].ts`, `[traceId]/tags/index.ts`, and `[traceId]/tags/[tagId].ts`.
>     - Use `upsertTrace` and `auditLog` for trace updates and logging.
>   - **Tests**:
>     - Add tests for new endpoints in `traces-api.servertest.ts`.
>     - Cover cases for adding, removing, and updating tags, and updating trace properties.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 4e96f04e66306e255ecf09906f2b91fad997e821. It will automatically update as commits are pushed.</sup>


web/src/pages/api/public/traces/[traceId].ts
```diff
@@ -173,4 +178,81 @@ export default withMiddlewares({
...
+          resourceType: "trace",
+          resourceId: traceId,
+          action: "updateTags",
+          after: body.tags,
+          projectId: auth.scope.projectId,
+          apiKeyId: auth.scope.apiKeyId,
+          orgId: auth.scope.orgId,
+        });
+        trace.tags = body.tags;
+      }
```
Comment (Steffen911) [inline web/src/pages/api/public/traces/[traceId].ts:248]: Add additional /traces/<id>/tags/(tagId) routes for POST and DELETE to allow adding single tags and removing single tags.

web/src/pages/api/public/traces/[traceId].ts
```diff
@@ -173,4 +178,81 @@ export default withMiddlewares({
       return { message: "Trace deleted successfully" };
     },
   }),
+
+  PATCH: createAuthedAPIRoute({
```
Comment (Steffen911) [inline web/src/pages/api/public/traces/[traceId].ts:182]: Return 202 Accepted status code and document that this is processed async.

Steffen911:
Claude currently overloaded. Try again later with
> In addition to the changes in web/src/pages/api/public/traces/[traceId].ts I want to add two new endpoints. One POST /api/public/traces/[traceId]/tags
  endpoint that adds a single tag to the trace and unions it with the existing ones and a DELETE /api/public/traces/[traceId]/tags/[tagId] endpoint that
  removes a single tag from all tags on the trace. Follow the same implementation pattern as for the PATCH traces endpoint. All of those calls should
  return a 202 status code. Update the fern documentation accordingly and make sure to indicate in the description that all three endpoints perform the
  updates in the background.

web/src/pages/api/public/traces/[traceId].ts
```diff
@@ -173,4 +178,82 @@ export default withMiddlewares({
...
+          resourceType: "trace",
+          resourceId: traceId,
+          action: "updateTags",
+          after: body.tags,
+          projectId: auth.scope.projectId,
+          apiKeyId: auth.scope.apiKeyId,
+          orgId: auth.scope.orgId,
+        });
+        trace.tags = body.tags;
+      }
```
Comment (Steffen911) [inline web/src/pages/api/public/traces/[traceId].ts:249]: Note: This call _replaces_ all tags on the trace.

web/src/pages/api/public/traces/[traceId].ts
```diff
@@ -173,4 +178,82 @@ export default withMiddlewares({
...
+
```
Comment (maxdeichmann) [inline web/src/pages/api/public/traces/[traceId].ts:250]: I am not sure whether we should use auditlogs here. This could kill postgres performance.

Comment (maxdeichmann) [reply web/src/pages/api/public/traces/[traceId].ts:250]: We can keep it and see how this will play out.

web/src/pages/api/public/traces/[traceId]/tags/[tagId].ts
```diff
@@ -0,0 +1,58 @@
...
+          before: trace.tags,
+          after: updatedTags,
+          projectId: auth.scope.projectId,
+          apiKeyId: auth.scope.apiKeyId,
+          orgId: auth.scope.orgId,
+        });
+
+        // Update trace
+        trace.tags = updatedTags;
+        await upsertTrace(convertTraceDomainToClickhouse(trace));
```
Comment (maxdeichmann) [inline web/src/pages/api/public/traces/[traceId]/tags/[tagId].ts:52]: do the upsert functions upload to s3?

maxdeichmann (approved):

web/src/pages/api/public/traces/[traceId]/tags/[tagId].ts
Comment (Steffen911) [reply web/src/pages/api/public/traces/[traceId]/tags/[tagId].ts:52]: @maxdeichmann Yes, they do.

web/src/pages/api/public/traces/[traceId].ts
Comment (Steffen911) [reply web/src/pages/api/public/traces/[traceId].ts:250]: @maxdeichmann Would you exclude all tracing operations from auditLog then? Given it's insert only my assumption was that this should be fine. Also, we have this in the trpc routes that's why I included them here.

Comment (maxdeichmann) [reply web/src/pages/api/public/traces/[traceId].ts:250]: Yes, i would exclude them for now.

PR #6055, opened 2025-03-18
feat: support external S3 endpoint for batch exports

Steffen911:
> [!IMPORTANT]
> Add support for external endpoints in S3 and Azure Blob Storage services to replace internal endpoints in signed URLs if configured.
> 
>   - **Behavior**:
>     - Add `externalEndpoint` parameter to `StorageServiceFactory`, `AzureBlobStorageService`, and `S3StorageService` to replace internal endpoint in signed URLs if configured.
>     - Update `getSignedUrl` and `getSignedUploadUrl` methods in both `AzureBlobStorageService` and `S3StorageService` to use `externalEndpoint`.
>   - **Environment**:
>     - Add `LANGFUSE_S3_BATCH_EXPORT_EXTERNAL_ENDPOINT` to `env.mjs` and `env.ts`.
>   - **Tests**:
>     - Add tests in `storageservice.test.ts` to verify behavior with and without `externalEndpoint`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 48afa12aac37fd01299fa0983968f4d3773ae629. It will automatically update as commits are pushed.</sup>


web/src/env.mjs
```diff
@@ -247,6 +248,7 @@ export const env = createEnv({
     LANGFUSE_S3_MEDIA_UPLOAD_PREFIX: z.string().default(""),
     LANGFUSE_S3_MEDIA_UPLOAD_REGION: z.string().optional(),
     LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT: z.string().optional(),
+    LANGFUSE_S3_MEDIA_UPLOAD_EXTERNAL_ENDPOINT: z.string().optional(),
```
hassiebp, line 251: On second thought I think we should remove this option for media since the endpoint _must_ be publicly resolvable as else both upload (via SDK) and download (via e.g. browser) need a publicly resolvable presigned URL.

Having two options here could add confusion

packages/shared/src/server/services/StorageService.ts
```diff
@@ -51,11 +51,22 @@ export interface StorageService {
...
+   * @param params.externalEndpoint - External endpoint to replace the internal endpoint in the signed URL.
+   * @param params.region - Region in which the bucket resides
+   * @param params.forcePathStyle - Add bucket name into the path instead of the domain name. Mainly used for MinIO.
+   */
   public static getInstance(params: {
     accessKeyId: string | undefined;
     secretAccessKey: string | undefined;
     bucketName: string;
     endpoint: string | undefined;
+    externalEndpoint?: string | undefined;
```
hassiebp, line 69: Could we simplify this by getting rid of the constructor param `externalEndpoint` and simply choose on instantiation which endpoint to pass to the constructor? 

That would mean that whereever we create batch export download links we do a simple check whether an external endpoint is defined in the environment, and if yes we use that to instantiate the StorageService so the presigned URLs will use that

Steffen911: @hassiebp IMO this is more desirable as we would never use the "internal" endpoint to make calls from Langfuse. If a user wants to keep all outbound langfuse traffic within the VPC, they could supply an VPC endpoint here and then replace it with an external endpoint for access that must come from outside the VPC.

How would you suggestion differ from just supplying the external endpoint as the endpoint?

web/src/env.mjs
hassiebp: True, but my thinking was rather that in the media case we should never have 2 different ones at the current state as we at no point directly access the bucket from the LF platform service

packages/shared/src/server/services/StorageService.ts
hassiebp: Ok, makes sense! 👍  So expectation by user is that whenever `…_EXTERNAL_..` is set, presigned URLs will use the external endpoint

PR #6068, opened 2025-03-18
fix(ui): OpenAiMessageView now renders messages where content is an empty string instead of ignoring them

davidmc971:
Doing a loose null check `message.content != null` instead of a truthy existence check `!!message.content` to determine if a message should be rendered such that empty messages, which count towards the "Show x more ..." button, will also be displayed.

----

> [!IMPORTANT]
> `OpenAiMessageView` now renders messages with empty content strings by using a loose null check instead of a truthy existence check.
> 
>   - **Behavior**:
>     - `OpenAiMessageView` now renders messages with empty content strings by using `message.content != null` instead of `!!message.content`.
>     - Updates `messagesToRender` to filter messages based on `shouldRenderContent` and `shouldRenderJson`.
>     - Adjusts collapse logic to use `messagesToRender.length`.
>   - **Functions**:
>     - Adds `shouldRenderContent` and `shouldRenderJson` functions to determine message rendering conditions in `OpenAiMessageView`.
>   - **Misc**:
>     - Adds `useMemo` import in `IOPreview.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f91768b7d681094c7e1b7f314f532d8d7db648b8. It will automatically update as commits are pushed.</sup>


davidmc971:
@ellipsis-dev please update the summary

davidmc971:
@ellipsis-dev the summary is incorrect as we do not render messages with null content, rather messages that consist of an empty string content are also rendered

marliessophie (approved):
LGTM

web/src/components/trace/IOPreview.tsx
```diff
@@ -226,26 +226,44 @@ export const OpenAiMessageView: React.FC<{
...
             )
             .map((message, index) => (
               <Fragment key={index}>
-                {(!!message.content || !!message.audio) &&
+                {shouldRenderContent(message) &&
                   (shouldRenderMarkdown ? (
                     <MarkdownJsonView
                       title={message.name ?? message.role}
-                      content={message.content}
+                      content={
```
marcklingen, line 264: Do we need this? The `Message.content` is a union type, so `.length` might not work well here. Can we coalesce to an empty string if it's falsy?

davidmc971: Good point, I can do `!message.content ? "" : message.content` here

marcklingen: `message.content ?? '""'` should do the trick

davidmc971: Tested it, `message.content ?? '""'` is loose null checking and thus does not fallback on an empty string, but `message.content || '""'` is falsy and works

web/src/components/trace/IOPreview.tsx
```diff
@@ -226,26 +226,42 @@ export const OpenAiMessageView: React.FC<{
...
             )
             .map((message, index) => (
               <Fragment key={index}>
-                {(!!message.content || !!message.audio) &&
+                {shouldRenderContent(message) &&
                   (shouldRenderMarkdown ? (
                     <MarkdownJsonView
                       title={message.name ?? message.role}
-                      content={message.content}
+                      content={message.content || '""'}
```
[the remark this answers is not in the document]
davidmc971: The falsy check is indeed the intended behavior in this case

PR #6071, opened 2025-03-18
fix(ui): For Checkbox component we stop event propagation on CheckboxPrimitive.Root such that our DataTable on row click events do not trigger when selecting a row; should not have unwanted side effects as we do not expect a checkbox to propagate a click

davidmc971:
> [!IMPORTANT]
> Add `onClick` handler to `CheckboxPrimitive.Root` in `checkbox.tsx` to stop event propagation, preventing DataTable row click events when a checkbox is clicked.
> 
>   - **Behavior**:
>     - In `checkbox.tsx`, added `onClick` handler to `CheckboxPrimitive.Root` to stop event propagation.
>     - Prevents DataTable row click events from triggering when a checkbox is clicked.
>   - **Misc**:
>     - No expected side effects as checkboxes typically do not propagate click events.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for c6aa92f2f83d33f1fd9f27d1cd2c9dfedbaab385. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/components/ui/checkbox.tsx
+++ b/web/src/components/ui/checkbox.tsx
@@ -9,7 +9,7 @@ import { cn } from "@/src/utils/tailwind";
 const Checkbox = React.forwardRef<
   React.ElementRef<typeof CheckboxPrimitive.Root>,
   React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root>
->(({ className, ...props }, ref) => (
+>(({ className, onClick, ...props }, ref) => (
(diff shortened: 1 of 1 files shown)


davidmc971 (commented):
We could also make this configurable via a prop on Checkbox and either default to preventing propagation or allowing it.

PR #6075, opened 2025-03-18
fix(tests): disable --experimental-require-module (default since Node 20.19.0) since it breaks jest-based testing

davidmc971:
> [!IMPORTANT]
> Disable `--experimental-require-module` in Jest test scripts and update Node.js version to `v20` for compatibility.
> 
>   - **Testing**:
>     - Disable `--experimental-require-module` in Jest test scripts (`test`, `test-sync`, `test-client`, `test:watch`, `test:e2e:server`) in `package.json` using `cross-env` to set `NODE_OPTIONS`.
>   - **Environment**:
>     - Update Node.js version in `.nvmrc` to `v20` to ensure compatibility with Node.js 20.x versions.
>   - **Dependencies**:
>     - Add `cross-env` as a dev dependency in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for ab6ea703695542e4dc16c5b2ee8e6071ed01378e. It will automatically update as commits are pushed.</sup>


davidmc971:
Vitest seems to be jest compatible by design: https://vitest.dev/guide/migration.html#jest

Since jest is problematic here and is no longer updated frequently, I am looking into using vitest instead

PR #6247, opened 2025-03-27
feat(evals, ui): enable deletion of evaluator via UI, DeleteButton refactor

davidmc971:
> [!IMPORTANT]
> Adds evaluator deletion via UI, refactors DeleteButton for multiple entities, updates access rights, and enhances analytics.
> 
>   - **Evaluator Deletion**:
>     - Adds `deleteEvalJob` mutation to `evalRouter` in `router.ts` for deleting evaluator jobs.
>     - Implements `DeleteEvaluatorButton` in `deleteButton.tsx` to handle evaluator deletion via UI.
>     - Updates `evaluator-detail.tsx` to include `DeleteEvaluatorButton` for evaluator deletion.
>   - **UI Refactor**:
>     - Refactors `DeleteButton` in `deleteButton.tsx` to support multiple entity types (trace, dataset, evaluator).
>     - Introduces `DeleteTraceButton`, `DeleteDatasetButton`, and `DeleteEvaluatorButton` in `deleteButton.tsx`.
>     - Updates `traces.tsx`, `TracePage.tsx`, and `DatasetActionButton.tsx` to use new delete button components.
>   - **Access Rights**:
>     - Renames `evalTemplate:create` to `evalTemplate:CUD` in `projectAccessRights.ts`.
>     - Updates access checks in `router.ts` and related components to use `evalTemplate:CUD`.
>   - **Analytics**:
>     - Adds new PostHog events for evaluator deletion in `usePostHogClientCapture.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for cd117f5a4ef2a507d71df0354d50435f4892799d. It will automatically update as commits are pushed.</sup>


web/src/components/deleteButton.tsx
```diff
@@ -161,7 +177,55 @@ export function DeleteButton({
...
+                  evalTemplateId: itemId,
+                });
+                capture("eval_templates:delete_template_button_click", {
+                  source: "template detail",
+                });
+              }}
+            >
+              Delete template
+            </Button>
+          ) : null}
```
maxdeichmann, line 228: What do you think about a refactoring where we have a component per type "trace" | "dataset" | "evaluator" | "template". Then we could implement all type specific logic in their own components which instantiate a shared component.

web/src/ee/features/evals/components/eval-template-detail.tsx
```diff
@@ -77,6 +86,29 @@ export const EvalTemplateDetail = () => {
                 );
               }}
             />
+            <DropdownMenu>
```
maxdeichmann, line 89: https://github.com/langfuse/langfuse/blob/308f4d6ed7113d526528cc53b0f8e9740db1a39d/web/src/components/trace/TracePage.tsx#L133-L134
What do you think about implementing it the same way as on the traces table?

web/src/ee/features/evals/server/router.ts
```diff
@@ -933,6 +933,101 @@
...
+        where: {
+          id: evalConfigId,
+          projectId: projectId,
+        },
+      });
+
+      // TODO: cascading delete of job executions: add test that verifies that job executions are deleted
+    }),
+
+  deleteEvalTemplate: protectedProjectProcedure
```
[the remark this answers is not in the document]
davidmc971: This happens, since the cascade logic in the prisma schema is set accordingly

web/src/ee/features/evals/server/router.ts
```diff
@@ -933,6 +933,101 @@ export const evalRouter = createTRPCRouter({
...
+        resourceId: evalConfigId,
+        action: "delete",
+      });
+
+      await ctx.prisma.jobConfiguration.delete({
+        where: {
+          id: evalConfigId,
+          projectId: projectId,
+        },
+      });
```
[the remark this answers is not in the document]
davidmc971: This has been confirmed and according tests exist now

web/src/ee/features/evals/components/evaluator-detail.tsx
```diff
@@ -140,6 +148,29 @@ export const EvaluatorDetail = () => {
                 listKey="evals"
               />
             )}
+            <DropdownMenu>
```
maxdeichmann, line 151: Same as for the eval template comment. I would not add the dropdown as the remaining product does not have that.

web/src/ee/features/evals/server/router.ts
```diff
@@ -933,6 +933,101 @@ export const evalRouter = createTRPCRouter({
...
+      const existingJob = await ctx.prisma.jobConfiguration.findUnique({
+        where: {
+          id: evalConfigId,
+          projectId: projectId,
+        },
+      });
+
+      if (!existingJob) {
+        logger.warn(
+          `Job for update not found for project ${projectId} and id ${evalConfigId}`,
```
maxdeichmann, line 959: job for deletion...

web/src/ee/features/evals/server/router.ts
```diff
@@ -933,6 +933,101 @@ export const evalRouter = createTRPCRouter({
...
+      });
+
+      await ctx.prisma.jobConfiguration.delete({
+        where: {
+          id: evalConfigId,
+          projectId: projectId,
+        },
+      });
+
+      // TODO: cascading delete of job executions: add test that verifies that job executions are deleted
```
maxdeichmann, line 981: please remove the comment.

web/src/ee/features/evals/components/eval-template-detail.tsx
davidmc971: You mean without context menu but rather directly the delete button? Do you think it's better to hide the delete button if the user has no access scope to delete or to disable the button?

web/src/components/deleteButton.tsx
davidmc971: Sounds good, would you pass the component in via props or just create local components per case that then get selected?

web/src/ee/features/evals/components/eval-template-detail.tsx
```diff
@@ -77,6 +86,29 @@ export const EvalTemplateDetail = () => {
...
+                    itemId={templateId}
+                    projectId={projectId}
+                    isTableAction={false}
+                    scope="evalTemplate:CUD"
+                    invalidateFunc={() => {
+                      void utils.evals.invalidate();
+                    }}
+                    type="template"
+                    redirectUrl={`/project/${projectId}/evals/templates`}
+                    deleteConfirmation={template.data?.name}
```
maxdeichmann, line 107: We have templates with different versions. Can you add the version here?

web/src/ee/features/evals/components/evaluator-detail.tsx
```diff
@@ -55,6 +62,7 @@ export const EvaluatorDetail = () => {
   const router = useRouter();
   const projectId = router.query.projectId as string;
   const evaluatorId = router.query.evaluatorId as string;
+  const utils = api.useUtils();
```
maxdeichmann, line 66: Below, in this section:

 if (evaluator.data && evaluator.data.evalTemplate === null) {
    return <div>Evaluator not found</div>;
  }

  If we delete the template, we do not render anything anymore. I would propose to disable the sidebar then.

web/src/components/deleteButton.tsx
maxdeichmann: have one button variant per element type. This would be the base button. Buttons for each feature are their own classes.

web/src/ee/features/evals/components/eval-template-detail.tsx
maxdeichmann: As discussed: same behavior as in the traces table. Please check what happens if user is not allowed to delete how behavior is implemented there.

web/src/ee/features/evals/components/evaluator-detail.tsx
davidmc971: You mean the evaluator configuration side panel? Don't we retain the settings even if the template got deleted?

davidmc971: You mean the evaluator configuration side panel? Don't we retain the settings even if the template got deleted?

web/src/features/datasets/components/DatasetActionButton.tsx
```diff
@@ -77,6 +77,7 @@ export const DatasetActionButton = (props: DatasetActionButtonProps) => {
               variant={props.variant || "ghost"}
               size={props.size || "icon"}
               className={props.className}
+              disabled={!hasAccess}
```
maxdeichmann, line 80: we have hasaccess logic below in line 88, or not?

web/src/features/datasets/components/DatasetActionButton.tsx
```diff
@@ -97,6 +98,7 @@ export const DatasetActionButton = (props: DatasetActionButtonProps) => {
             variant={props.variant || "ghost"}
             size={props.size}
             className={props.className}
+            disabled={!hasAccess}
```
maxdeichmann, line 101: can you also add the hasaccess logic here?

melih-assessio:
as far as I have seen, evaluators can be removed from the `Running Evaluators` page but not from the `Evaluator Library` page. is that right?

erantr1:
@melih-assessio 
This is what I see as well.
@dosu what's your take on this one? can custom evaluators now be deleted from the evaluator library?

boyscout99:
+1
I would like the ability to delete evaluators from the Evaluator Library as well.

binhpageai:
I can’t delete evaluators from the Evaluator Library page.
Even when I remove a run from Running Evaluators, the old results still affect the final score on the Dataset page.
It averages the old score with the new score after I edit the evaluator prompt — which is frustrating because I only want to see the new score for the latest version.

brandonwatts:
This is still an ask, we need the ability to delete custom evaluators out of the evaluator library in addition to running evaluators.

PR #6253, opened 2025-03-27
chore: add traces domain layer

maxdeichmann:
> [!IMPORTANT]
> Introduce `TraceDomain` schema for trace data handling and update components, services, and API responses to use it.
> 
>   - **Schema**:
>     - Add `TraceDomain` schema in `domain.ts` using `zod` for trace data.
>     - Replace `Trace` type with `TraceDomain` in various components and services.
>   - **Converters**:
>     - Update `convertTraceDomainToClickhouse()` and `convertClickhouseToDomain()` in `traces_converters.ts` to use `TraceDomain`.
>   - **Services**:
>     - Modify `traces-ui-table-service.ts` to use `TraceDomain` for UI return types.
>   - **Components**:
>     - Update components like `ObservationTree.tsx`, `TracePreview.tsx`, and `TraceTimelineView.tsx` to use `TraceDomain`.
>   - **API**:
>     - Set `externalId` to `null` in API responses in `[sessionId].ts`, `[traceId].ts`, and `index.ts` for traces.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 99fb6c11ed42b6c458835c1c4e1d6d2370d211ff. It will automatically update as commits are pushed.</sup>


packages/shared/src/domain.ts
```diff
@@ -0,0 +1,30 @@
...
+  name: z.string().nullable(),
+  timestamp: z.date(),
+  environment: z.string(),
+  tags: z.array(z.string()),
+  bookmarked: z.boolean(),
+  public: z.boolean(),
+  release: z.string().nullable(),
+  version: z.string().nullable(),
+  input: jsonSchema.nullable(),
+  output: jsonSchema.nullable(),
```
Steffen911, line 21: This may cause event loop issues again if we try to validate that via zod.

packages/shared/src/domain.ts
```diff
@@ -0,0 +1,30 @@
...
+  metadata: MetadataDomain,
+  createdAt: z.date(),
+  updatedAt: z.date(),
+  sessionId: z.string().nullable(),
+  userId: z.string().nullable(),
+  projectId: z.string(),
+});
+
+export type Trace = z.infer<typeof TraceDomain>;
```
maxdeichmann, line 30: rename to TraceDomain

packages/shared/src/domain.ts
maxdeichmann: Decision: change to unknown

maxdeichmann: Yes, we should refrain from parsing this. But using this as a typescript type is good.

PR #6322, opened 2025-04-03
feat: add new widget creator page

Steffen911:
> [!IMPORTANT]
> Adds a new widget creator page with support for various chart types and filtering options, including new chart components and updated types.
> 
>   - **New Feature**:
>     - Adds a new widget creator page in `new.tsx` for creating widgets with various chart types and filtering options.
>     - Supports chart types: line, vertical bar, horizontal bar, and pie charts.
>     - Allows filtering by environment, trace name, tags, user, session, release, and version.
>   - **Chart Components**:
>     - Introduces `Chart`, `HorizontalBarChart`, `VerticalBarChart`, `LineChartTimeSeries`, `VerticalBarChartTimeSeries`, and `PieChart` components in `chart-library`.
>     - Utilizes `recharts` library for rendering charts.
>   - **Types and Models**:
>     - Updates `TableName` in `types.ts` to include `widgets`.
>     - Adds `DataPoint` and `ChartProps` types in `chart-props.ts`.
>     - Updates `dataModel.ts` and `dashboardUiTableToViewMapping.ts` to support new widget functionality.
>   - **Miscellaneous**:
>     - Updates `package.json` to include `recharts` dependency.
>     - Modifies `queryBuilder.servertest.ts` and `queryBuilderDashboards.servertest.ts` for test adjustments.
>     - Adjusts CSS in `globals.css` for chart styling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 9113e1252d3d81f8327015820c31e070e5d52140. It will automatically update as commits are pushed.</sup>


web/src/components/ui/card.tsx
```diff
@@ -1,6 +1,6 @@
-import * as React from "react";
+import * as React from "react"
```
Steffen911, line 1: These changes where made while installing the new chart lib. They seem safe enough to not affect other things within the code.

web/src/components/ui/chart.tsx
```diff
@@ -0,0 +1,365 @@
+"use client"
```
Steffen911, line 1: Added during chart installation.

web/src/styles/globals.css
```diff
@@ -73,6 +73,11 @@
     --sidebar-accent-foreground: 240 5.9% 10%;
     --sidebar-border: 220 13% 91%;
     --sidebar-ring: 217.2 91.2% 59.8%;
+
+    --chart-1: 176 60.8% 19%;
+    --chart-2: 217 91.2% 59.8%;
+    --chart-3: 43 96% 40%;
+    --chart-4: 0 72.2% 50.6%;
```
Steffen911, line 80: Those map to
- dark-green
- dark-blue
- dark-yellow
- dark-red
If there is a good way to reuse the values instead, feel free to recommend it. Same thing below for the dark mode.

web/src/features/widgets/chart-library/HorizontalBarChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+          dataKey="dimension"
+          stroke="#888888"
+          fontSize={12}
+          tickLine={false}
+          axisLine={false}
+        />
+        <Bar
+          dataKey="metric"
+          radius={[0, 4, 4, 0]}
+          className="fill-[--color-metric]"
```
Steffen911, line 53: ToDo: Check if we can just drop this.

web/src/features/widgets/chart-library/HorizontalBarChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+        />
+        <ChartTooltip
+          content={
+            <ChartTooltipContent
+              formatter={(value) =>
+                Intl.NumberFormat("en-US")
+                  .format(value as number)
+                  .toString()
```
Steffen911, line 61: May also be possible to drop this or apply the same formatting on the other charts. Will check what looks best.

web/src/features/widgets/chart-library/PieChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+          content={<ChartTooltipContent hideLabel />}
+        />
+        <Pie
+          data={chartData}
+          dataKey="value"
+          nameKey="name"
+          cx="50%"
+          cy="50%"
+          innerRadius={80}
+          outerRadius={120}
```
Steffen911, line 56: Will have to check if we can make those dynamic based on the available space. Will do with fixed values for the time being.

web/src/pages/project/[projectId]/widgets/new.tsx
```diff
@@ -0,0 +1,592 @@
...
+  const isAdmin = session.data?.user?.admin === true;
+
+  const router = useRouter();
+  const { projectId } = router.query as { projectId: string };
+
+  // State for form fields
+  const [widgetName, setWidgetName] = useState<string>("Traces");
+  const [widgetDescription, setWidgetDescription] = useState<string>(
+    "Traces grouped by name for the last 30 days.",
+  );
```
Comment (Steffen911) [inline web/src/pages/project/[projectId]/widgets/new.tsx:55]: ToDo: Review whether we want to set empty defaults and instead use placeholder to guide the user.

web/src/components/ui/card.tsx
```diff
@@ -10,77 +10,70 @@ const Card = React.forwardRef<
...
+Card.displayName = "Card"
 
 const CardHeader = React.forwardRef<
   HTMLDivElement,
   React.HTMLAttributes<HTMLDivElement>
 >(({ className, ...props }, ref) => (
   <div
     ref={ref}
-    className={cn("flex flex-col space-y-1 p-4", className)}
+    className={cn("flex flex-col space-y-1.5 p-6", className)}
```
marliessophie, line 26: The padding changes in this file will affect all cards in our application, I am unsure this is intended? I would rather simply pass a custom `className` where you use `CardHeader` (in this case `space-y-1.5 p-6` and revert the change in this file.

web/src/components/ui/card.tsx
```diff
@@ -10,77 +10,70 @@ const Card = React.forwardRef<
...
 >(({ className, ...props }, ref) => (
-  <p
+  <div
     ref={ref}
     className={cn("text-sm text-muted-foreground", className)}
     {...props}
   />
-));
-CardDescription.displayName = "CardDescription";
+))
```
marliessophie, line 56: May I ask why we made the change from a `HTMLParagraphElement` to a `HTMLDivElement`? We mainly lose some semantic meaning here as these descriptions and titles represent text. Has minor accessibility impact too. If this was changed due to HTML nesting rules I'd like to know. Likely not a huge impact, but would like to understand the context :)

web/src/components/ui/card.tsx
```diff
@@ -10,77 +10,70 @@ const Card = React.forwardRef<
...
+CardContent.displayName = "CardContent"
 
 const CardFooter = React.forwardRef<
   HTMLDivElement,
   React.HTMLAttributes<HTMLDivElement>
 >(({ className, ...props }, ref) => (
   <div
     ref={ref}
-    className={cn("flex items-center p-4 pt-0", className)}
+    className={cn("flex items-center p-6 pt-0", className)}
```
marliessophie, line 73: Same comment as above, let's not change the padding here.

web/src/features/widgets/chart-library/HorizontalBarChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <ChartContainer config={config}>
+      <BarChart
+        accessibilityLayer={accessibilityLayer}
+        data={data}
+        layout="vertical"
+      >
+        <XAxis
+          type="number"
+          stroke="#888888"
```
marliessophie, line 37: nit: I would define a color for this, something like: ` stroke="var(--chart-grid)"`. Adding `--chart-grid` to globals would be the way to go.

web/src/features/widgets/chart-library/chart-props.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export interface ChartProps {
+  data: DataPoint[];
+  config?: {
+    metric?: {
+      theme?: {
+        light: string;
+        dark: string;
+      };
```
marliessophie, line 14: Observation: Why do we need this? We seem to use `chart-1` as color everywhere? Don't mind having it either :)

web/src/pages/project/[projectId]/widgets/new.tsx
```diff
@@ -0,0 +1,592 @@
...
+                        <PieChart data={transformedData.slice(0, rowLimit)} />
+                      );
+                    default:
+                      return (
+                        <HorizontalBarChart
+                          data={transformedData.slice(0, rowLimit)}
+                        />
+                      );
+                  }
+                })()}
```
Comment (marliessophie) [inline web/src/pages/project/[projectId]/widgets/new.tsx:576]: Nit: I would extract this. Then you can do the `queryResult.data` check within the cardcontent and return either the div or this exacted component. You only need to pass transformed data to it, and the selected chart type string.

web/src/pages/project/[projectId]/widgets/new.tsx
Comment (marliessophie) [inline web/src/pages/project/[projectId]/widgets/new.tsx:1]: General comment, I see two meta points: 
- the form also internally tracks state, we are now duplicating state by holding it locally and in the form. You mentioned you would like to drop the form all together, is that correct? Then I will not further expand on this. Either way we should get rid of this duplication. We can discuss how to best approach this if helpful.
- we have lots of hooks, would be great to extract some of them to make this more readable. We could extract some hooks e.g. something like:
   - `useWidgetOptions`, could return available views, metrics, dimensions and filter options (should only need `projectId` and `viewName`
   - `useWidgetQuery`, could define the dynamic query and execute it. We can consider also including the data transformation step here but I think I would rather create a `useTransformedChartData` hook for this

web/src/pages/project/[projectId]/widgets/new.tsx
```diff
@@ -0,0 +1,592 @@
...
+
+                {/* View Selection */}
+                <div className="space-y-2">
+                  <Label htmlFor="view-select">View</Label>
+                  <Select
+                    value={selectedView}
+                    onValueChange={(value) =>
+                      setSelectedView(value as z.infer<typeof views>)
+                    }
+                  >
```
Comment (marliessophie) [inline web/src/pages/project/[projectId]/widgets/new.tsx:363]: We might want to consider loading states here. Also is `selectedView` always defined (I don't think so)? If not we should handle this case too.

web/src/features/widgets/chart-library/VerticalBarChartTimeSeries.tsx
```diff
@@ -0,0 +1,72 @@
...
+    },
+  },
+  accessibilityLayer = true,
+}) => {
+  const groupedData = useMemo(() => groupDataByTimeDimension(data), [data]);
+  const dimensions = useMemo(() => getUniqueDimensions(data), [data]);
+  const enhancedConfig = useMemo(
+    () => expandChartConfig(config, dimensions),
+    [config, dimensions],
+  );
```
marliessophie, line 38: I think `useMemo` is optional here, this calculation is not very expensive.

web/src/features/widgets/chart-library/PieChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return data.reduce((acc, curr) => acc + curr.metric, 0);
+  }, [data]);
+
+  // Transform data for PieChart
+  const chartData = useMemo(() => {
+    return data.map((item, index) => ({
+      name: item.dimension || "Unknown",
+      value: item.metric,
+      fill: `hsl(var(--chart-${(index % 4) + 1}))`,
+    }));
```
marliessophie, line 39: Nit: these could generally be extracted, then we have a clear place where all data transformations to chart data. But no need to do this, more of an observation

marliessophie (commented):
@Steffen911 in general this looks really amazing!! Left mostly nits and one more general meta comment. Let me know if chatting about this is helpful ☺️ 

Really excited for this!!

web/src/components/ui/chart.tsx
Steffen911: Correct.

web/src/pages/project/[projectId]/widgets/new.tsx
Comment (Steffen911) [reply web/src/pages/project/[projectId]/widgets/new.tsx:363]: Remove isPresent checks in other cases.

web/src/components/ui/chart.tsx
```diff
@@ -0,0 +1,365 @@
...
+    ([, config]) => config.theme || config.color
+  )
+
+  if (!colorConfig.length) {
+    return null
+  }
+
+  return (
+    <style
+      dangerouslySetInnerHTML={{
```
dahnny012, line 81: Im doing a security audit langfuse within my org and this was flagged as high risk. Do we know if thats a false positive or this is an attack vector

Steffen911: @dahnny012 This should be a false-positive. We only render values that are sourced from our source-code to adjust the chart with the selected theme.

PR #6439, opened 2025-04-13
feat: create outline of SCIM Users support

Steffen911:
> [!IMPORTANT]
> Add initial SCIM Users API support with endpoints for user management, service provider config, resource types, and schemas, including OpenAPI/Postman docs and comprehensive tests.
> 
>   - **SCIM Users API Implementation**:
>     - Adds SCIM endpoints: `ServiceProviderConfig`, `ResourceTypes`, `Schemas`, `Users` (list, create, get, delete) in `web/src/pages/api/public/scim/`.
>     - All endpoints require organization-scoped API key; project keys and invalid keys are rejected with 403/401.
>     - `Users` supports filtering by `userName`, pagination (`startIndex`, `count`), and returns SCIM-compliant user objects.
>     - `POST /Users` creates a user (with optional password), returns 409 if user exists, 400 if required fields missing.
>     - `DELETE /Users/{id}` removes user from organization (does not delete user entity).
>     - All endpoints return SCIM-compliant error responses and 405 for unsupported methods.
>   - **Documentation & SDKs**:
>     - Adds SCIM endpoints and schemas to `openapi.yml` and Postman `collection.json`.
>     - Defines SCIM types and endpoints in `fern/apis/server/definition/scim.yml`.
>   - **Testing**:
>     - Adds `scim-api.servertest.ts` with comprehensive tests for all endpoints, authentication, error cases, and user creation/deletion scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 59eb0799a5a780a6490100cca7a75d82dc6d1428. It will automatically update as commits are pushed.</sup>


fern/apis/server/definition/scim.yml
```diff
@@ -0,0 +1,199 @@
...
+
+    listUsers:
+      docs: List users in the organization (requires organization-scoped API key)
+      method: GET
+      path: /Users
+      request:
+        name: ListUsersRequest
+        query-parameters:
+          filter:
+            docs: Filter expression (e.g. userName eq "value")
```
marliessophie, line 35: _Question_: This is not an example but the only filter value, correct?

web/src/pages/api/public/scim/Users/index.ts
```diff
@@ -0,0 +1,214 @@
...
+        where: {
+          email: userName.toLowerCase(),
+        },
+        create: {
+          email: userName.toLowerCase(),
+          name: name.formatted,
+          password: password ? await hashPassword(password) : undefined,
+        },
+        update: {},
+      });
```
marliessophie, line 175: _Question_: this might fail if there is a user in a different org with this email address, no?

marliessophie (commented):
Left one question and one comment 👍

fern/apis/server/definition/scim.yml
Steffen911: @marliessophie We may add more over time. Not sure which of the expressions we'll eventually need there.

web/src/pages/api/public/scim/Users/index.ts
marliessophie: Yes but what if there is a user in another org that has the same email address? Then: 
- `existingUser` would be `null` 
- the create call would violate the global uniqueness constraint on email. 
Or am I missing something?

Steffen911: @marliessophie No, if there is a user in another org with the same email address we would:
1. Do nothing in the user upsert, i.e. keep it as is.
2. Add the user that already exists with this email to the org that requested that this user email is to be added.

We decided to never perform any updates on users based on org api key operations as we cannot guarantee any malicious accesses if we'd allow that. We may add this later and enforce some "domain ownership" verification to allow this.

PR #6449, opened 2025-04-14
feat: add endpoints to manage project and organization memberships

Steffen911:
> [!IMPORTANT]
> Add endpoints for managing project and organization memberships, including tests and Postman collection updates.
> 
>   - **Endpoints**:
>     - Add `getOrganizationMemberships` and `updateOrganizationMembership` endpoints in `memberships.yml` and `openapi.yml`.
>     - Add `getProjectMemberships` and `updateProjectMembership` endpoints in `memberships.yml` and `openapi.yml`.
>   - **Handlers**:
>     - Implement `handleGet` and `handlePut` for organization memberships in `organizations/memberships/index.ts`.
>     - Implement `handleGet` and `handlePut` for project memberships in `projects/[projectId]/memberships/index.ts`.
>   - **Tests**:
>     - Add tests for organization and project membership endpoints in `memberships-api.servertest.ts`.
>   - **Postman Collection**:
>     - Update `collection.json` to include new membership endpoints.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 175153842aa1edb63535c4c5ae5e0e73d0c9c0e2. It will automatically update as commits are pushed.</sup>


web/src/pages/api/public/organizations/memberships/index.ts
```diff
@@ -0,0 +1,160 @@
...
+      userId: membership.userId,
+      role: membership.role,
+      email: membership.user.email,
+      name: membership.user.name,
+    })),
+  });
+}
+
+// PUT - Update or create an organization membership
+async function handlePut(
```
marliessophie, line 110: _Question_: I'm confused here. This allows to create or update an organization membership, but only if the user already exists? Is that correct?

web/src/pages/api/public/projects/[projectId]/memberships/index.ts
```diff
@@ -0,0 +1,212 @@
...
+      userId: membership.userId,
+      role: membership.role,
+      email: membership.user.email,
+      name: membership.user.name,
+    })),
+  });
+}
+
+// PUT - Update or create a project membership
+async function handlePut(
```
Comment (marliessophie) [inline web/src/pages/api/public/projects/[projectId]/memberships/index.ts:149]: _Question/Issue_: why does this not have the entitlement check?

marliessophie (commented):
Left two comments, answering both would be good here to ensure we don't miss anything.

web/src/pages/api/public/organizations/memberships/index.ts
Steffen911: @marliessophie No confusion here. This is exactly what it does. The Users endpoint from the previous PR allows you to create a user and optionally assign it to an organization with `None` permissions as SCIM does not have a standardized way to handle role assignments. Therefore, we added this additional endpoint to upsert permissions for an existing user for a given organization.

web/src/pages/api/public/projects/[projectId]/memberships/index.ts
Comment (Steffen911) [reply web/src/pages/api/public/projects/[projectId]/memberships/index.ts:149]: @marliessophie We perform the entitlement check in the generic `handler` _before_ we call the handleGet and handlePut methods. Adding them in here would be redundant.

PR #6533, opened 2025-04-21
perf: soft delete event logs clickhouse

maxdeichmann:
> [!IMPORTANT]
> Introduces soft delete mechanism for event logs in ClickHouse and S3, replacing hard delete operations with soft delete across various functions and files.
> 
>   - **Behavior**:
>     - Introduces soft delete mechanism for event logs in ClickHouse and S3.
>     - Replaces hard delete operations with soft delete in `ingestionFileDeletion.ts` and related functions.
>     - Updates `processClickhouseScoreDelete` and `processClickhouseTraceDelete` to use new soft delete functions.
>   - **Environment Variables**:
>     - Adds new S3 media upload environment variables in `env.ts`.
>   - **Functions**:
>     - Adds `removeIngestionEventsFromS3AndDeleteClickhouseRefs` and related functions in `ingestionFileDeletion.ts`.
>     - Updates `getBlobStorageByProjectId...` functions in `blobStorageLog.ts` to use `FINAL` keyword for ClickHouse queries.
>   - **Misc**:
>     - Adds `dev:worker` and `dev:web` scripts in `package.json` and `turbo.json`.
>     - Removes redundant S3 client initialization in several files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for ed423989cdd0244e907f143134f4331257089cb6. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/s3/index.ts
```diff
@@ -0,0 +1,36 @@
+import { env } from "../../env";
```
Steffen911, line 1: This is nice! Good one 🙏

packages/shared/src/server/data-deletion/ingestionFileDeletion.ts
```diff
@@ -0,0 +1,108 @@
...
+  const eventLogStream = getBlobStorageByProjectIdAndEntityIds(
+    p.projectId,
+    "score",
+    p.scoreIds,
+  );
+
+  return removeIngestionEventsFromS3AndDeleteClickhouseRefs({
+    projectId: p.projectId,
+    stream: eventLogStream,
+  });
```
Steffen911, line 26:
```suggestion
  const stream = getBlobStorageByProjectIdAndEntityIds(
    p.projectId,
    "score",
    p.scoreIds,
  );

  return removeIngestionEventsFromS3AndDeleteClickhouseRefs({
    projectId: p.projectId,
    stream,
  });
```

packages/shared/src/server/data-deletion/ingestionFileDeletion.ts
```diff
@@ -0,0 +1,108 @@
...
+
+async function removeIngestionEventsFromS3AndDeleteClickhouseRefs(p: {
+  projectId: string;
+  stream: AsyncGenerator<EventLogRecordReadType>;
+}) {
+  const { projectId, stream } = p;
+
+  let batch = 0;
+
+  let blobStorageRefs: EventLogRecordReadType[] = [];
```
Steffen911, line 64: They type probably needs an update as well.

Steffen911 (commented):
Overall the approach looks good.

Two thoughts / open questions:
- We may use the ClickHouseWriter to track the deletions. As they are batched anyway, I don't see the point though and wouldn't do it.
- Do we want to delete the S3 files before or after removing the records from clickhouse? Right now, S3 goes first and I'd keep it that way as I don't see any reason to flip the order. Want to ensure that this has a few seconds of deliberation before merging though.

PR #6729, opened 2025-05-07
feat: add org API keys to organization settings

Steffen911:
> [!IMPORTANT]
> Adds organization-level API key management with access control, UI integration, and analytics tracking.
> 
>   - **Behavior**:
>     - Adds organization-level API key management in `organizationApiKeysRouter` with operations: `create`, `delete`, `updateNote`, and `byOrganizationId`.
>     - Only owners and admins can manage organization API keys; members and unauthenticated users are restricted.
>     - Integrates API key management into organization settings page (`index.tsx`).
>   - **Access Control**:
>     - Adds `organization:CRUD_apiKeys` scope to `organizationAccessRights.ts` for OWNER and ADMIN roles.
>     - Uses `throwIfNoOrganizationAccess` to enforce access control in `organizationApiKeyRouter.ts`.
>   - **UI Components**:
>     - Updates `ApiKeyList.tsx` to handle both project and organization API keys.
>     - Adds `CreateApiKeyButton` and `DeleteApiKeyButton` components for organization API keys.
>     - Updates `SetupPage.tsx` to include organization API key creation.
>   - **Analytics**:
>     - Adds `api_key_create` and `api_key_delete` events to `usePostHogClientCapture.ts` for organization settings.
>   - **Tests**:
>     - Adds `organizations-api-key-trpc.servertest.ts` to test organization API key operations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f018acb11a7d8ce693b12b1f93567180e8b52617. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/public-api/components/ApiKeyList.tsx
```diff
@@ -22,26 +22,47 @@ import {
...
+    );
+  }
+
+  const hasProjectAccess = useHasProjectAccess({
+    projectId: props.entityId,
+    scope: "apiKeys:CUD",
+  });
+  const hasOrganizationAccess = useHasOrganizationAccess({
+    organizationId: props.entityId,
+    scope: "organization:CRUD_apiKeys",
```
Steffen911, line 50: @marcklingen Do we have any convention around scope names? I just created this one as it reflects best what we intend to cover with it. Initially, I thought `organizations:update` would also be sufficient, but having a separate scope here may make more sense.

Let me know if you have any doubts about this approach.

web/src/features/public-api/components/ApiKeyList.tsx
```diff
@@ -117,27 +152,75 @@ export function ApiKeyList(props: { projectId: string }) {
...
-    projectId: props.projectId,
+
+  const hasProjectAccess = useHasProjectAccess({
+    projectId: props.entityId,
     scope: "apiKeys:CUD",
   });
+  const hasOrganizationAccess = useHasOrganizationAccess({
+    organizationId: props.entityId,
+    scope: "organization:CRUD_apiKeys",
+  });
```
Steffen911, line 176: @marliessophie What would you consider a best practice here? I perform the access check multiple times since I only parse the scope into the sub components. I could instead parse an `hasAccess` prop to remove duplication.
IMO the current approach still feels better even if more verbose as the scopes may change and the parent should not make auth decisions for the child. Would you agree?

web/src/features/public-api/components/ApiKeyList.tsx
```diff
@@ -22,26 +22,47 @@ import {
...
+  const apiKeysQuery =
+    scope === "project"
+      ? api.projectApiKeys.byProjectId.useQuery(
+          { projectId: entityId },
+          { enabled: hasAccess },
+        )
+      : api.organizationApiKeys.byOrganizationId.useQuery(
+          { orgId: entityId },
+          { enabled: hasAccess },
+        );
```
marliessophie, line 65: Hooks cannot be called conditionally, see [docs](https://react.dev/warnings/invalid-hook-call-warning). We'll need to call both hooks each time. You can further use the `enabled` prop to call only if the scope matches.

web/src/features/public-api/components/ApiKeyList.tsx
marliessophie: I agree with your assessment, would recommend to keep the more verbose approach of performing access checks at a component level.

marliessophie (commented):
Answering comment ☺️

PR #6750, opened 2025-05-08
feat: add level dimension to traces for custom dashboards

Steffen911:


Steffen911:
obsolete with events tables.

PR #6781, opened 2025-05-12
fix(ui-graph-view): fetch graph data separately

hassiebp:
> [!IMPORTANT]
> Fetches graph data separately for UI graph view, updating components and API to support new data fetching logic.
> 
>   - **Behavior**:
>     - Adds `getAgentGraphData` function in `traces.ts` to fetch graph data based on `projectId`, `traceId`, and time range.
>     - Updates `Trace` component in `index.tsx` to use `getAgentGraphData` for fetching graph data and determining graph view availability.
>     - Removes `isLanggraphTrace` utility and replaces its usage with `isGraphViewAvailable` logic.
>   - **Components**:
>     - Modifies `TraceGraphView` in `TraceGraphView.tsx` to accept `agentGraphData` prop and update graph rendering logic.
>     - Updates `Trace` component to conditionally render `TraceGraphView` based on `isGraphViewAvailable`.
>   - **API**:
>     - Adds `getAgentGraphData` query to `traces.ts` API router to fetch and parse graph data using `AgentGraphDataSchema`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 8dc1c7c8e935d36a97de639df122010967825073. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/server/api/routers/traces.ts
```diff
@@ -483,4 +488,67 @@ export const traceRouter = createTRPCRouter({
...
+
+      const records = await queryClickhouse({
+        query,
+        params: {
+          traceId,
+          projectId,
+          minStartTime,
+          maxStartTime,
+        },
+      });
```
Steffen911, line 533: Usually, we use the `repositories/traces.ts` in the worker to collect all queries. Could you move this there and only invoke a function here?

web/src/server/api/routers/traces.ts
```diff
@@ -483,4 +488,67 @@ export const traceRouter = createTRPCRouter({
...
+        projectId: z.string(),
+        traceId: z.string(),
+        observationStartTimes: z.array(z.string()),
+      }),
+    )
+    .query(async ({ input }): Promise<Required<AgentGraphDataResponse>[]> => {
+      const { traceId, projectId, observationStartTimes } = input;
+
+      const ascendingStartTimes = observationStartTimes
+        .map((t) => t.replaceAll("Z", ""))
```
Steffen911, line 504: We have a utility function to convert the ISO string to something clickhouse understands. Something like convertClickHouseTo... - I would recommend to use it here as well.

web/src/server/api/routers/traces.ts
```diff
@@ -483,4 +488,67 @@ export const traceRouter = createTRPCRouter({
...
+      const maxStartTime = ascendingStartTimes[ascendingStartTimes.length - 1];
+
+      const query = `
+          SELECT
+            id,
+            parent_observation_id,
+            metadata['langgraph_node'] AS node,
+            metadata['langgraph_step'] AS step
+          FROM
+            observations
```
Steffen911, line 517: This will cause duplicate results. You'd have to use FINAL or some other method to deduplicate if this might be a problem.

Steffen911 (commented):
You may consider a response limit. If there are 10000 observations, you'd currently return all of them.
Apart from this, performance looks reasonable to me.

web/src/server/api/routers/traces.ts
hassiebp: Duplicates are handled, let's keep this non-FINAL to opt for better performance 👍🏾

web/src/server/api/routers/traces.ts
hassiebp: thanks! done

web/src/server/api/routers/traces.ts
hassiebp: Yes, done - thanks!

PR #6907, opened 2025-05-19
feat: add official support for conditional variables

### LINKED ISSUE #6906: Add support for conditional rendering in prompt management
### Describe the feature or potential improvement

One useful extension of the new prompt composability feature would be the ability to conditionally include prompts or text in our prompts based on a variable. 

Right now I have a few prompts where I have an instruction that I only want included when a given condition is true, with either no instruction or an alternate instruction otherwise. There are two ways that I have found to accommodate that use case currently:
1. Include both instruction sets in the prompt with a one-word variable telling the ai what to focus on, which increases input size and could confuse the model
2. Store the instructions either as string literals in code or as separate prompts that are then called, and then insert the resulting instruction as a variable in the parent prompt. This makes the parent prompt quite difficult to maintain, since either way the instruction is not connected to the prompt itself, so it is not clear to editors what the variable will actually pull in, and editing the instruction requires either editing the code, defeating  the purpose of prompt management, or finding and editing a separate prompt that is not referenced in the parent.

The mustache spec has a system for rendering conditional arguments, and it is already possible to do this (if not officially supported) in the JS sdk - see my comment below

Example use case:
One of our use cases is a math tutor that helps students on our pre-written problems, some of which allow the use of a calculator. If a calculator is allowed, we want the tutor to integrate instructions about how to use the calculator on the problem in its responses, but since this is only a small part of the overall prompt it would be annoying to create and maintain two separate prompts for each version.

Other potential use cases:
- Varying instructions or capabilities based on a users' attributes (like subscription level, preferences)
- Model-specific instructions to reflect best practices for prompting for eg. reasoning vs non-reasoning models

_Originally posted by @QuinnStraus in https://github.com/langfuse/langfuse/discussions/6010_

harrygao56:
## What does this PR do?

As discussed in https://github.com/orgs/langfuse/discussions/6010

I've added support for conditionally rendering blocks/variables (see https://mustache.github.io/mustache.5.html).
Now the prompt manager will:

- Recognize conditional variables 
- Not show a linter error for variables starting with # and /
- Show a linter error when there is an unclosed conditional variable

Lmk what you think and if I should make any changes!

Fixes #6906 

## Type of change

- [x]  New feature (non-breaking change which adds functionality)

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Adds support for conditional variables in mustache syntax, updating regex and linter logic to handle and validate these cases.
> 
>   - **Behavior**:
>     - Adds support for conditional variables in mustache syntax in `stringChecks.ts` and `CodeMirrorEditor.tsx`.
>     - Linter in `CodeMirrorEditor.tsx` now recognizes and flags unclosed conditional variables as errors.
>   - **Regex**:
>     - Adds `CONDITIONAL_MUSTACHE_REGEX_START` and `buildConditionalMustacheRegex()` in `stringChecks.ts` for detecting conditional variables.
>     - Updates `extractVariables()` in `stringChecks.ts` to include conditional variables.
>   - **Validation**:
>     - Modifies `isValidVariableName()` in `stringChecks.ts` to handle variables starting with `#` and `/`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for aca980e4543966f266045e95b01a9916a113bb6b. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/utils/stringChecks.ts
+++ b/packages/shared/src/utils/stringChecks.ts
@@ -10,6 +10,16 @@ export const VARIABLE_REGEX = /^[a-zA-Z][a-zA-Z_]*$/;
 // Regex to find variables in mustache syntax
 export const MUSTACHE_REGEX = /{{([^{}]*)}}+/g;
 
+// Regex to find conditional variables (see https://mustache.github.io/mustache.5.html)
+export const CONDITIONAL_MUSTACHE_REGEX_START = /{{#([^{}]*)}}+/g;
+
+// Regex to find conditionals with a specific variable
+export const buildConditionalMustacheRegex = (variable: string, type: "start" | "end") => {
+  return type === "start"
+    ? new RegExp(`{{#${variable}}}+`, "g")
+    : new RegExp(`{{/${variable}}}+`, "g");
+};
+
 // Regex to find multiline variables
 export const MULTILINE_VARIABLE_REGEX = /{{[^}]*\n[^}]*}}/g;
 
@@ -17,7 +27,11 @@ export const MULTILINE_VARIABLE_REGEX = /{{[^}]*\n[^}]*}}/g;
 export const UNCLOSED_VARIABLE_REGEX = /{{(?![^{]*}})/g;
 
 export function isValidVariableName(variable: string): boolean {
-  return VARIABLE_REGEX.test(variable);
+  const normalizedVariable = variable.startsWith('#') || variable.startsWith('/') 
+    ? variable.slice(1) 
+    : variable;
+    
+  return VARIABLE_REGEX.test(normalizedVariable);
 }
 
 export function extractVariables(mustacheString: string): string[] {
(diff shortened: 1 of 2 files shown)


hassiebp:
Thanks for your contribution @harrygao56 - closing this, see comment here: https://github.com/langfuse/langfuse/issues/6906#issuecomment-2890085421

Also, prompt logic must be working cross platform and not rely on the current template var resolution we use in the JS SDK (mustache), but work with the same semantics in our Python SDKs and any future language runtimes we plan to support.

alecswan:
@hassiebp , it appears that there is already a dependency on the mustache syntax in the main branch: https://github.com/langfuse/langfuse/blob/main/web/src/components/editor/CodeMirrorEditor.tsx#L36

Given that and the fact that [mustache](https://mustache.github.io/) is supported by python, javascript, go and a lot of other languages, why the push back on this suggestion?

PR #7012, opened 2025-05-26
feat: webhook for prompt version changes

maxdeichmann:
> [!IMPORTANT]
> Introduces webhook support for prompt version changes with new models, tasks, and tests, updating permissions and environment configurations.
> 
>   - **Behavior**:
>     - Adds webhook support for prompt version changes, triggering on create, update, and delete events.
>     - Implements `executeWebhook` in `webhooks.ts` to handle webhook execution with retries and error handling.
>     - Adds `promptVersionProcessor` in `promptVersionProcessor.ts` to process prompt version change events.
>     - Introduces `entityChangeWorker` in `entityChangeWorker.ts` to delegate entity change events.
>   - **Models**:
>     - Adds `Action`, `Trigger`, `Automation`, and `AutomationExecution` models in `schema.prisma`.
>     - Updates `types.ts` to include new enums `ActionType` and `ActionExecutionStatus`.
>   - **Queues**:
>     - Adds `WebhookQueue` and `EntityChangeQueue` in `queues.ts`.
>     - Registers queue processors in `app.ts` for webhook and entity change processing.
>   - **Tests**:
>     - Adds tests for webhook execution and signature verification in `webhooks.test.ts` and `signature.test.ts`.
>     - Tests prompt version processing in `promptVersionProcessor.test.ts`.
>   - **Permissions**:
>     - Updates `projectAccessRights.ts` to include `automations:CUD` and `automations:read` scopes.
>   - **Environment**:
>     - Adds environment variables for queue processing concurrency in `env.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 11b22bf6cf84471e6ff9593331a88a267c85708e. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/schema.prisma
```diff
@@ -157,6 +157,10 @@ model Project {
   Dashboard              Dashboard[]
   DashboardWidget        DashboardWidget[]
   TableViewPreset        TableViewPreset[]
+  actions                Action[]
+  triggers               Trigger[]
```
Steffen911, line 161: Let's try to be consistent here with the remaining naming.
```suggestion
  Actions                Action[]
  Triggers               Trigger[]
```

packages/shared/prisma/generated/types.ts
```diff
@@ -115,6 +115,17 @@ export const DashboardWidgetChartType = {
     PIVOT_TABLE: "PIVOT_TABLE"
 } as const;
 export type DashboardWidgetChartType = (typeof DashboardWidgetChartType)[keyof typeof DashboardWidgetChartType];
+export const ActionType = {
```
Steffen911, line 118: I assume this and the migration is fully generated from the schema.prisma. So not checks here. Please run a prisma dev migration to confirm that there is no delta.

web/package.json
```diff
@@ -14,9 +14,9 @@
     "prettier": "prettier --write ./src *.{ts,js}",
     "clean": "rm -rf node_modules",
     "start": "dotenv -e ../.env -- sh -c 'NEXT_MANUAL_SIG_HANDLE=true next start'",
-    "test": "cross-env NODE_OPTIONS='--no-experimental-require-module' dotenv -e ../.env -- jest --silent false --verbose false --runInBand --detectOpenHandles --selectProjects async-server",
```
Steffen911: Is changing to verbose actually desired here?

packages/shared/src/server/repositories/index.ts
```diff
@@ -12,3 +12,4 @@ export * from "./trace-sessions";
 export * from "./scores-utils";
 export * from "./blobStorageLog";
 export * from "./environments";
+export * from "./automation-repository";
```
Steffen911, line 15: I would challenge that a repository wrapper around prisma statements is worth it. IMO those are simple enough to be used across the board. My understanding was that it's mainly the clickhouse queries that should receive a more "ORM"-like experience using the repositories.

I would vote to remove this abstraction and inline the logic.

packages/shared/prisma/schema.prisma
```diff
@@ -1214,3 +1218,108 @@ model TableViewPreset {
...
+
+  projectId String  @map("project_id")
+  project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+  type ActionType
+
+  // Configuration specific to each action type
+  config Json // Structured JSON for different action types
+  // For WEBHOOK: { version: "1.0", url: "...", method: "POST", headers: {...}, secretId: "..." }
+  // For ANNOTATION_QUEUE: { version: "1.0", queueId: "..." }
```
Steffen911, line 1235: ANNOTATION_QUEUE is out of scope here, right?

packages/shared/prisma/schema.prisma
```diff
@@ -1214,3 +1218,108 @@ model TableViewPreset {
...
+
+  // Relations
+  triggers        Automation[] // Changed to use the join table
```
Steffen911: Let's remove the comment as it doesn't make sense looking forward. Took me some time to understand what is going on there as the statement didn't make it obvious.

packages/shared/prisma/schema.prisma
```diff
@@ -1214,3 +1218,108 @@ model TableViewPreset {
...
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now()) @map("created_at")
+  updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
+  sourceId  String   @map("source_id")
+
+  triggerId String  @map("trigger_id")
+  trigger   Trigger @relation(fields: [triggerId], references: [id], onDelete: Cascade)
+
+  actionId String @map("action_id")
+  action   Action @relation(fields: [actionId], references: [id], onDelete: Cascade)
```
Steffen911, line 1310: Just putting it out there, unclear whether it is better: We can consider referencing the Automation here. Otherwise, we may be in an "underdefined" situation in case someone creates two Automations with the same triggers and actions.
Also, an AutomationExecution feels more natural in that regard.

Alternatively, we can skip the trigger here and only bind it to actions.

packages/shared/prisma/schema.prisma
```diff
@@ -1214,3 +1218,108 @@ model TableViewPreset {
...
+
+  projectId String  @map("project_id")
+  project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+  status     ActionExecutionStatus @default(PENDING) @map("status")
+  input      Json                  @map("input")
+  output     Json?                 @map("output")
```
Steffen911, line 1317: This could become potentially very large, right? And what is the "Output" in this semantic? Just a placeholder in case we need it for non-webhook actions?

packages/shared/src/domain/prompts.ts
```diff
@@ -0,0 +1,21 @@
...
+  createdBy: z.string(),
+  isActive: z.boolean().nullable(),
+  type: z.string().default("text"),
+  tags: z.array(z.string()).default([]),
+  labels: z.array(z.string()).default([]),
+  prompt: jsonSchemaNullable,
+  config: jsonSchemaNullable,
+  projectId: z.string(),
+  commitMessage: z.string().nullable(),
+});
```
Steffen911, line 19: Tbh, I'm unsure about those. It feels useful for ClickHouse to be legacy compatible, but here we're re-declaring the prisma schema in most cases. I'm concerned that parsing back and forth between them may cause a performance overhead that is unnecessary.

packages/shared/src/domain/webhooks.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const WebhookOutboundBaseSchema = z.object({
+  id: z.string(),
+  timestamp: z.coerce.date(),
+  type: z.literal("prompt"),
+  action: EventActionSchema,
+});
+
+export const PromptWebhookOutboundSchema = z
```
Steffen911, line 18: Should this be versioned?

packages/shared/src/encryption/encryption.ts
```diff
@@ -0,0 +1,60 @@
+import crypto from "crypto";
```
Steffen911, line 1: Note: just moved, not newly written.

packages/shared/src/encryption/signature.ts
```diff
@@ -0,0 +1,115 @@
+import crypto from "crypto";
```
Steffen911, line 1: I assume this is copied and well tested? On skim this looks reasonable with no red flags going up.

Steffen911 (commented):
Reviewed ~1/3rd with a focus on the DB schema. Rest will follow later.

worker/src/app.ts
```diff
@@ -338,6 +340,22 @@ if (env.QUEUE_CONSUMER_DEAD_LETTER_RETRY_QUEUE_IS_ENABLED === "true") {
...
+    concurrency: 5,
+  });
+}
+
+if (env.QUEUE_CONSUMER_PROMPT_VERSION_CHANGE_QUEUE_IS_ENABLED === "true") {
+  WorkerManager.register(
+    QueueName.PromptVersionChangeQueue,
+    promptVersionChangeQueueProcessor,
+    {
+      concurrency: 2, // Low concurrency to avoid overwhelming the database
```
Steffen911: Might make sense to already add envs here to make them configurable.

worker/src/features/promptVersionChange/promptVersionChangeWorker.ts
```diff
@@ -0,0 +1,263 @@
...
+} from "@langfuse/shared/src/server";
+import { TriggerEventSource } from "@langfuse/shared";
+import { ActionExecutionStatus, JobConfigState } from "@langfuse/shared";
+import { prisma } from "@langfuse/shared/src/db";
+import { v4 } from "uuid";
+
+/**
+ * Helper function to check if action filter matches the event action
+ */
+function actionMatches(
```
Steffen911: Consider using the InMemoryFilterService - this already has this included.

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+    const webhookConfig = actionConfig.config;
+
+    // Validate that the webhook URL uses HTTPS protocol for security
+    try {
+      const webhookUrl = new URL(webhookConfig.url);
+      if (webhookUrl.protocol !== "https:") {
+        throw new Error(
+          `Webhook URL must use HTTPS protocol for security. Received: ${webhookUrl.protocol}`,
+        );
+      }
```
Steffen911: IMO not necessary to enforce that here. I would show a warning on the client/frontend, but if someone wants to receive the call via HTTP they should be able to. As long as we don't transfer anything sensitive this makes it easy.

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+      ...webhookConfig.headers,
+    };
+
+    if (!webhookConfig.secretKey) {
+      logger.warn(
+        `Webhook config for action ${automation.action.id} has no secret key, failing webhook execution`,
+      );
+      throw new InternalServerError(
+        "Webhook config has no secret key, failing webhook execution",
+      );
```
Steffen911, line 104: IMO this should be optional as well.

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    if (webhookConfig.secretKey) {
+      try {
+        const decryptedSecret = decrypt(webhookConfig.secretKey);
+
+        const signature = createSignatureHeader(
+          webhookPayload,
+          decryptedSecret,
+        );
+        requestHeaders["Langfuse-Signature"] = signature;
```
Steffen911: Consider prefixing custom headers with x-, e.g. x-langfuse-signature

Steffen911: For Github I've also seen that they put in the "hash" that was signed.

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+            `Webhook for project ${projectId} failed with status ${res.status} and response ${responseBody}`,
+          );
+          throw new Error(
+            `Webhook for project ${projectId} failed with status ${res.status}`,
+          );
+        }
+      },
+      {
+        numOfAttempts: 4,
+      },
```
Steffen911: This means we have a backoff via the queue _and_ here? Or do we fail the queue on the first attempt? I would consider only relying on the queue here.

Steffen911: Ah I see, this is not marked as a recoverable error in the catch block. Might make sense to highlight this in a comment.

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+        automationId,
+        projectId,
+      });
+
+      logger.info(
+        `Consecutive failures: ${consecutiveFailures} for trigger ${automation.trigger.id} in project ${projectId}`,
+      );
+
+      // Check if trigger should be disabled (>= 5 consecutive failures)
+      if (consecutiveFailures >= 4) {
```
Steffen911: Mismatch between comment and implementation

packages/shared/src/server/automations/webhooks.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    if (shouldRetryJob) {
+      logger.warn(
+        `Retrying bullmq for webhook job for action ${automation.action.id}`,
+      );
+      throw error;
+    }
+
+    // Update action execution status and check if we should disable trigger
+    await prisma.$transaction(async (tx) => {
```
Steffen911, line 201: IMO no need to wrap this in a transaction. Setting to error and then disabling can be two separate operations.

packages/shared/src/server/queues.ts
```diff
@@ -308,4 +341,16 @@ export type TQueueJobTypes = {
...
+    id: string;
+    payload: WebhookInput;
+    name: QueueJobs.WebhookJob;
+  };
+  [QueueName.PromptVersionChangeQueue]: {
+    timestamp: Date;
+    id: string;
+    payload: PromptVersionChangeEventType;
+    name: QueueJobs.PromptVersionChangeJob;
+  };
```
Steffen911: I see this already exploding in the number of queues. I would recommend a single EntityChange queue with an envelope message that indicates the body and then a type specific payload. Otherwise, this will become super messy to maintain IMO.

packages/shared/src/tableDefinitions/promptsTable.ts
```diff
@@ -77,3 +74,7 @@ export function promptsTableColsWithOptions(
     return col;
   });
 }
+
+export function webhookActionFilterOptions(): ColumnDefinition[] {
+  return promptsTableCols.filter((col) => col.id === "name");
+}
```
Steffen911, line 80: This seems super odd as a util function. Do we expect to support more columns at some point?

web/src/features/automations/server/router.ts
```diff
@@ -0,0 +1,479 @@
...
+        automationId: z.string(),
+      }),
+    )
+    .query(async ({ ctx, input }) => {
+      // Check if user has at least read access to automations
+      throwIfNoProjectAccess({
+        session: ctx.session,
+        projectId: input.projectId,
+        scope: "automations:read",
+      });
```
Steffen911, line 54: Shall we make this a property of the protectedProjectProcedure? Maybe as a future refactoring? IMO handling scope checks there could make everything simpler.

web/src/features/prompts/server/routers/promptRouter.ts
```diff
@@ -775,6 +796,24 @@ export const promptRouter = createTRPCRouter({
...
+          },
+        });
+
+        // Send webhooks for ALL affected prompts
+        await Promise.all(
+          updatedPrompts.map(async (prompt) =>
+            promptChangeEventSourcing(
+              await promptService.resolvePrompt(prompt),
+              "updated",
+            ),
```
Steffen911, line 814: Is there a prisma hook that we can use? Maybe that would cover audit logs and event sourcing.

Steffen911 (commented):
I skipped all tsx files. Overall, LGTM. Main concern is the maintainability of the "fine-grained" queues. If we go more generic we could also cover something like TraceUpsert with the same flow, etc.

packages/shared/prisma/schema.prisma
maxdeichmann: Yes, exactly! Placeholder for future automation executions where we want to show the user some result.

packages/shared/src/tableDefinitions/promptsTable.ts
maxdeichmann: Yes, we might. Therefore taking the default ones and filtering for the ones we allow for webhooks.

packages/shared/src/server/automations/webhooks.ts
maxdeichmann: I will keep this one. It will be created upon webhook configuration. There should never be a case where we do not have this.

packages/shared/src/domain/prompts.ts
maxdeichmann: I think it makes sense to add more of these. E.g. we should actually add the resolution graph here which is part of the domain but not in the database.

packages/shared/src/server/repositories/index.ts
maxdeichmann: I think we need it. The config can contain custom objects. I would like to have one central place to manage that.

packages/shared/src/domain/webhooks.ts
maxdeichmann: Yes, added a version to the outbound API.

PR #7206, opened 2025-06-09
fix(prompts): prompt name validation cleared when name blurred

nimarb:
## What does this PR do?

on unfocus of the text element, the validation errors are cleared. they are cleared because the zod schema validates again with a different ruleset and removes the errors.

https://github.com/user-attachments/assets/26e2277f-ed45-4d2c-8e49-b41d33d53069


nimarb:
superseeded by #7262 which includes this fix

PR #7241, opened 2025-06-11
feat(otel-ingestion): handle long-running traces

hassiebp:
> [!IMPORTANT]
> Introduces `OtelIngestionProcessor` class to handle OpenTelemetry spans, replacing function-based approach, with updated tests and API integration.
> 
>   - **Behavior**:
>     - Introduces `OtelIngestionProcessor` class in `OtelIngestionProcessor.ts` to handle OpenTelemetry spans, replacing previous function-based approach.
>     - Manages trace deduplication and processes spans into ingestion events.
>     - Filters out shallow trace events if full trace events exist for the same traceId.
>     - Initializes seen traces from Redis on first call.
>   - **Tests**:
>     - Updates tests in `otelMapping.servertest.ts` to use `OtelIngestionProcessor`.
>     - Adds tests for span counting, timestamp conversion, and trace creation logic.
>   - **API**:
>     - Updates `index.ts` to use `OtelIngestionProcessor` for processing OTEL resource spans.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for d8a55ec94296b5363ae2cdbe99ca932f2de5db64. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/otel/server/index.ts
```diff
@@ -1,7 +1,57 @@
...
+import {
+  TraceEventType,
+  type IngestionEventType,
+} from "@langfuse/shared/src/server";
 import { randomUUID } from "crypto";
 import { ForbiddenError, ObservationLevel } from "@langfuse/shared";
 import { LangfuseOtelSpanAttributes } from "./attributes";
+import { redis } from "@langfuse/shared/src/server";
+
+export async function getTraceSeenMap(
```
hassiebp: rename to set

web/src/features/otel/server/index.ts
```diff
@@ -1,7 +1,57 @@
...
 import { ForbiddenError, ObservationLevel } from "@langfuse/shared";
 import { LangfuseOtelSpanAttributes } from "./attributes";
+import { redis } from "@langfuse/shared/src/server";
+
+export async function getSeenTracesSet(
+  resourceSpans: unknown,
+  projectId: string,
+): Promise<Set<string>> {
+  if (!redis) {
+    console.warn("Redis client not available");
```
Steffen911: Let's use the logger library here instead of console statements.

web/src/features/otel/server/index.ts
```diff
@@ -1,7 +1,57 @@
...
+            Buffer.from(span.traceId?.data ?? span.traceId).toString("hex"),
+          );
+        }
+      }
+    });
+  }
+
+  const results = await Promise.all(
+    [...traceIds].map(async (traceId) => {
+      const key = `project:${projectId}:trace:${traceId}:seen`;
```
Steffen911: I'd also add a `langfuse:` prefix. It's not necessary but may make co-uses of the redis db cleaner.

web/src/features/otel/server/index.ts
```diff
@@ -701,53 +756,69 @@ export const convertOtelSpanToIngestionEvent = (
...
         events.push({
           id: randomUUID(),
           type: "trace-create",
           timestamp: new Date(startTimeISO).toISOString(),
           body: trace,
         });
+
+        // Update the local cache of 'seen' traces to avoid multiple
+        // trace create calls per batch
+        seenTraces.add(traceId);
```
Steffen911: I'm wondering whether it makes more sense to filter the duplicates in `web/src/pages/api/public/otel/v1/traces/index.ts` in case the same trace will be created multiple times. With this approach here, we would have two trace creates if the trace was not seen _and_ hasTraceUpdates or isRootSpan are true. If the full trace arrives within a batch that would also cause unnecessary operations. Ensuring that we only have on trace-create call might be better.

This is more complex though so unclear whether it's worth it - you may have a better intuition there.

Steffen911: Also, I dislike this kind of reference updates as the mutable state may lead to odd behaviours.

web/src/__tests__/server/api/otel/otelMapping.servertest.ts
```diff
@@ -2062,7 +2090,10 @@ describe("OTel Resource Span Mapping", () => {
         };
 
         // When
-        const langfuseEvents = convertOtelSpanToIngestionEvent(resourceSpan);
+        const langfuseEvents = convertOtelSpanToIngestionEvent(
+          resourceSpan,
+          new Set(),
+        );
```
Steffen911: Can you add a test case here that confirms that traces are created if they were not seen and skipped if they are seen? Should be easy to generate and would help confirm the behaviour.

hassiebp: Added alread :)

web/src/features/otel/server/index.ts
hassiebp: Yes, we need to do this for all our other cache keys as well at some point

web/src/features/otel/server/index.ts
hassiebp: We would need some way to manage state by traceId per batch to keep track of whether a root or a trace-update containing span was in the batch for the given trace ID, and then filter out the shallow trace-create event. 

Overall what we would gain is one less trace-create event for all traces that conclude within 10 minutes and do not have trace updates on the first processed span.

My hunch is that this is worth it, as each `trace-create` event is triggering a full ingestion pipeline run. Will look into implementation.

Steffen911: @hassiebp Sounds good. In any case, no blocker from my end and this change seems good to go 🙂

PR #7262, opened 2025-06-12
feat(prompts): virtual folders for prompt management

nimarb:
> [!IMPORTANT]
> Introduces virtual folders for prompt management by interpreting slashes in prompt names as folder separators, with updates to validation, UI, and backend logic.
> 
>   - **Behavior**:
>     - Introduces virtual folders for prompts by interpreting slashes in prompt names as folder separators.
>     - Updates `PromptNameSchema` in `validation.ts` to restrict prompt names from starting or ending with slashes, and disallow consecutive slashes.
>     - Adds breadcrumb navigation in `prompts-table.tsx` to navigate through virtual folders.
>   - **Validation**:
>     - Adds `PromptNameSchema` in `validation.ts` to enforce new naming rules.
>     - Updates `NewPromptForm/validation.ts` to use `PromptNameSchema`.
>   - **UI Components**:
>     - Modifies `NewPromptForm/index.tsx` to pre-fill prompt name with folder path if available.
>     - Updates `table-link.tsx` to handle click events for folder navigation.
>     - Enhances `prompts-table.tsx` to display folders and handle navigation.
>   - **Tests**:
>     - Adds tests in `prompts.v2.servertest.ts` to verify prompt name validation and folder behavior.
>   - **Backend**:
>     - Updates `promptRouter.ts` to handle folder paths in prompt queries.
>     - Modifies `seed.ts` to include prompts with folder-like names for testing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 690aef26a5f5c54d99c1377e9032ab2dc5737368. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

With prompt folders users with many prompts can easily organize & find them.
The folders are virtual and inferred if there is a slash in the prompt name.

image


maxdeichmann:
The folder row has a different height than the prompt row

image

packages/shared/prisma/seed.ts
```diff
@@ -1133,6 +1133,33 @@ export const SEED_PROMPTS = [
...
+  },
+  {
+    id: `folder-prompt-1`,
+    createdBy: "user-1",
+    prompt: "Folder prompt 1 content",
+    name: "folder/prompt-1",
+    version: 1,
+    labels: ["production", "latest"],
+    tags: ["tag1", "tag2"],
+  },
```
maxdeichmann, line 1162: Thanks for adding these! super helpful!

packages/shared/src/features/prompts/validation.ts
```diff
@@ -0,0 +1,16 @@
...
+  .string()
+  .min(1, "Enter a name")
+  .regex(/^[^|]*$/, "Prompt name cannot contain '|' character")
+  .regex(/^[a-zA-Z0-9_\-/. ]+$/, "Name must be alphanumeric with optional underscores, hyphens, periods, slashes or spaces")
+  .regex(/^[^/]/, "Name cannot start with a slash")
+  .regex(/^(?!.*\/\/)/, "Name cannot contain consecutive slashes")
+  .regex(/^.*[^/]$/, "Name cannot end with a slash")
+  .transform((s) => s.trim())
+  .refine((s) => s.length > 0, "Name cannot be empty")
+  .refine((name) => name !== "new", "Prompt name cannot be 'new'");
```
maxdeichmann, line 15: Did you check that all current prompts in production align with this?

web/src/__tests__/prompts.v2.servertest.ts
```diff
@@ -696,6 +696,76 @@ describe("/api/public/v2/prompts API Endpoint", () => {
...
+          "multiple...dots",
+          "name with spaces",
+          "multiple   spaces",
+        ];
+
+        for (const name of validNames) {
+          await testValidName(name);
+        }
+      });
+    });
```
maxdeichmann, line 756: Thanks!!

web/src/features/prompts/hooks/usePromptNameValidation.tsx
```diff
@@ -13,23 +13,19 @@ export const usePromptNameValidation = ({
...
       .includes(currentName);
 
     if (!isNewPrompt) {
       form.setError("name", { message: "Prompt name already exists." });
-    } else if (currentName === "new") {
-      form.setError("name", { message: "Prompt name cannot be 'new'" });
-    } else if (!/^[a-zA-Z0-9_\-.]+$/.test(currentName)) {
-      form.setError("name", {
-        message:
-          "Name must be alphanumeric with optional underscores, hyphens, or periods",
```
maxdeichmann, line 29: what happened to this validation?

packages/shared/src/features/prompts/validation.ts
nimarb: > Did you check that all current prompts in production align with this?

Actually, i just found non-complying prompts which contained a `]` character for example.
I will drop the alphanumeric requirement, and just block an initial slash, double slash and trailing slash (i didn't find any of those).

web/src/features/prompts/hooks/usePromptNameValidation.tsx
nimarb: > what happened to this validation?

It was not really applied, see #7206

packages/shared/src/features/prompts/validation.ts
maxdeichmann: Great, thanks!

web/src/features/prompts/components/prompts-table.tsx
```diff
@@ -136,10 +265,38 @@ export function PromptTable() {
...
+            />
+          );
+        }
+
         return name ? (
           <TableLink
-            path={`/project/${projectId}/prompts/${encodeURIComponent(name)}`}
+            path={`/project/${projectId}/prompts/${rowData.id}`}
             value={name}
+            icon={name}
```
maxdeichmann: What is this name icon? I think we can remove this.

web/src/features/prompts/components/prompts-table.tsx
```diff
@@ -136,10 +265,38 @@ export function PromptTable() {
...
+              }}
+              title={displayName || ""}
+            />
+          );
+        }
+
         return name ? (
           <TableLink
-            path={`/project/${projectId}/prompts/${encodeURIComponent(name)}`}
+            path={`/project/${projectId}/prompts/${rowData.id}`}
```
maxdeichmann: Cant we just leave this one as is?

web/src/features/prompts/components/prompts-table.tsx
```diff
@@ -223,14 +387,63 @@ export function PromptTable() {
       header: "Actions",
       size: 70,
       cell: (row) => {
-        const name = row.row.original.name;
-        return <DeletePrompt promptName={name} />;
+        if (isFolder(row.row.original)) return null;
+
+        const promptId = row.row.original.id;
+        return <DeletePrompt promptName={promptId} />;
```
maxdeichmann, line 392: One more question: why did you change everything to promptId here? Before it was name and when i test the PR, name is also used. Something seems off here.

nimarb: When displaying the values (e.g. a prompt name), `name` is used. Then, when we're in a folder, `name` is only a part of the entire prompt name (e.g. `folder/then/prompt` -> `then/prompt` in the `folder`). For the deletion to work, we need the entire string `folder/then/prompt`, which is still in the `id`).  Initially, it was a bug that broke the delete button.

Could also re-work that to compute the display value on the fly and make `name` and `id` identical again, wdyt?

maxdeichmann: Ok, got it. I was confused here as i thought that the id is the id the uuid from the postgres id column but we re-assign it in line 147. We should probably have better naming here as this will only be confusing going forward. If you have an idea for that, lets create a follow up PR on this one so we get this out of the door.

web/src/features/prompts/server/routers/promptRouter.ts
```diff
@@ -87,11 +88,15 @@ export const promptRouter = createTRPCRouter({
         "prompts",
       );
 
+      const pathFilter = input.pathPrefix
+        ? Prisma.sql` AND (p.name LIKE ${input.pathPrefix + '/%'} OR p.name = ${input.pathPrefix})`
```
hassiebp, line 92: to check: what do we do here for undefined / null pathPrefix

hassiebp:
prompt name `hassieb/prompts/system` does not load in UI

CleanShot 2025-06-16 at 12 15 44@2x

web/src/features/prompts/server/routers/promptRouter.ts
nimarb: if `input.pathPrefix` is empty, `pathFilter` will resolve to `Prisma.empty`(line 93)  -> so we never filter for an empty slash for example

PR #7546, opened 2025-06-27
feat: add pivot table widget to self-service dashboards

coffee4tw:
## What does this PR do?

Adds a new chart type, pivot table, that takes 0-N (max 2 currently) dimensions and 0-N (max 10 currently) metrics and displays them in a table to add to user-defined dashboards.

<div>
    <a href="https://www.loom.com/share/e0f4df8c84f94c378566daa6f8cb9c7a">
      <p>Introducing the New Pivot Table Feature in Our Widget Creation Tool! 📊 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/e0f4df8c84f94c378566daa6f8cb9c7a">
      
    </a>
  </div>

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Add pivot table widget to dashboards, supporting up to 2 dimensions and 10 metrics, with comprehensive UI and backend changes.
> 
>   - **Behavior**:
>     - Adds `PIVOT_TABLE` as a new chart type in `DashboardWidgetChartType`.
>     - Supports up to 2 dimensions and 10 metrics, with subtotals and grand totals.
>   - **Components**:
>     - Adds `PivotTable` component in `PivotTable.tsx` for rendering pivot tables.
>     - Updates `Chart.tsx` to handle `PIVOT_TABLE` type.
>     - Enhances `WidgetForm.tsx` to support pivot table configuration.
>   - **Utilities**:
>     - Implements `transformToPivotTable()` in `pivot-table-utils.ts` for data transformation.
>     - Adds utility functions for handling dimensions and metrics in `pivot-table-utils.ts`.
>   - **Tests**:
>     - Adds unit tests in `pivot-table-utils.servertest.ts` for data transformation.
>     - Adds component tests in `pivot-table.clienttest.tsx` for `PivotTable` rendering.
>     - Adds integration tests in `dashboard-router-pivot-table.servertest.ts` for query execution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b386f907685471180218bda271f634bd9f8a3dd4. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/services/DashboardService/types.ts
```diff
@@ -33,6 +34,11 @@ export const HistogramChartConfig = BaseTotalValueChartConfig.extend({
   bins: z.number().int().min(1).max(100).optional().default(10),
 });
 
+export const PivotTableChartConfig = BaseTotalValueChartConfig.extend({
+  type: z.literal("PIVOT_TABLE"),
+  dimensions: z.array(z.string()).max(MAX_PIVOT_TABLE_DIMENSIONS), // Configurable max dimensions for pivot tables
```
Steffen911: @coffee4tw Why do we need this extension of the ChartConfig? We already store an array of dimensions within the dashboard_widgets table (see database screenshot). To me this feels redundant.
image

web/src/features/query/server/queryBuilder.ts
```diff
@@ -30,9 +30,17 @@ type AppliedMetricType = {
...
+  private chartConfig?: {
+    bins?: number;
+    row_limit?: number;
+    dimensions?: string[];
+  };
+
+  constructor(chartConfig?: {
+    bins?: number;
+    row_limit?: number;
+    dimensions?: string[];
```
Steffen911: It seems like you're not using the new `dimensions` property anywhere within the query builder. Is that intentional?

web/src/features/widgets/components/WidgetForm.tsx
```diff
@@ -166,16 +190,58 @@ export function WidgetForm({
...
+      : // Default to single metric for pivot tables
+        [
+          {
+            id: `${initialValues.aggregation}_${initialValues.measure}`,
+            measure: initialValues.measure,
+            aggregation: initialValues.aggregation,
+            label: `${startCase(initialValues.aggregation)} ${startCase(initialValues.measure)}`,
+          },
+        ],
+  );
```
Steffen911: The initial loading doesn't seem to behave as expected, at least on my end. Only the first metric is loaded and shown even though the database has `[{"agg": "avg", "measure": "value"}, {"agg": "p90", "measure": "value"}, {"agg": "p95", "measure": "value"}]` as the current `metrics` state. You can reproduce it by adding a pivot table with multiple metrics, adding it to a dashboard, and then navigating back to the WidgetForm using the `Edit` button.

image

web/src/features/widgets/components/WidgetForm.tsx
```diff
@@ -441,31 +637,49 @@ export function WidgetForm({
...
+                    const val = item[dimensionField];
+                    if (typeof val === "string") return val;
+                    if (val === null || val === undefined || val === "")
+                      return "n/a";
+                    if (Array.isArray(val)) return val.join(", ");
+                    return String(val);
+                  })()
+                : startCase(
+                    metricField === "count_count" ? "Count" : metricField,
+                  ),
```
Steffen911: Currently, this hack does not seem to apply on pivot tables. Could you ensure that we avoid displaying `Count Count` within the UI? If you have a better way to keep the aggregation + measure pattern for counts, let me know - happy to solve this centrally and elsewhere. Just replacing it seems good enough though.

image

web/src/features/widgets/chart-library/PivotTable.tsx
```diff
@@ -0,0 +1,284 @@
+/**
```
Steffen911, line 1: Could we move the sub-totals to the top of each group and the full-total to the top? I believe in a dashboarding tool it's more useful to see them first and not scroll to the bottom.

web/src/features/widgets/chart-library/PivotTable.tsx
```diff
@@ -0,0 +1,284 @@
...
+      // Show up to 2 decimal places, removing trailing zeros
+      return value.toLocaleString(undefined, {
+        minimumFractionDigits: 0,
+        maximumFractionDigits: 2,
+      });
+    }
+  }
+
+  return String(value ?? "");
+}
```
Steffen911: You could use the numberFormatter utilities from web/src/utils/numbers.ts here instead.

web/src/features/query/server/queryBuilder.ts
coffee4tw: Good point, no that's not intentional. I added it following the Histogram example, before I realized the query already supports multiple dimensions out of the box and never removed it.

packages/shared/src/server/services/DashboardService/types.ts
coffee4tw: Hmm that's a reasonable argument, to me the dimensions belonged more to the chart than to the widget (similar to the rowLimit) but it is duplicated this way. I'll remove it.

web/src/features/widgets/chart-library/PivotTable.tsx
coffee4tw: Thanks for that pointer!

web/src/features/widgets/chart-library/PivotTable.tsx
coffee4tw: Sure, fine with me.

web/src/features/widgets/components/WidgetForm.tsx
coffee4tw: Good catch. This actually doesn't work for dimensions either after removing that from the chart config. Fixing both.

coffee4tw: Works as intended now.

web/src/features/widgets/components/WidgetForm.tsx
coffee4tw: Okay I moved this to a centralized utility function. I needed to add another one anyway to make the multi-metric naming more sensible. Let me know if that works for you now.

PR #7573, opened 2025-06-30
chore(prompt-experiments): set langfuse-native environment

marliessophie:
> [!IMPORTANT]
> Enhance event processing by introducing internal environment handling and refactor schema creation for Langfuse events.
> 
>   - **Behavior**:
>     - Introduces `isLangfuseInternal` option in `processEventBatch()` in `processEventBatch.ts` to handle internal Langfuse events.
>     - Updates `fetchLLMCompletion()` in `fetchLLMCompletion.ts` to use `isLangfuseInternal` for event processing.
>     - Adds `PROMPT_EXPERIMENT_ENVIRONMENT` constant in `types.ts` for internal environment setting.
>   - **Schemas**:
>     - Refactors schema creation in `types.ts` to support public and internal environments using `createIngestionEventSchema()`.
>     - Deprecates `ingestionEvent` in favor of `createIngestionEventSchema()`.
>   - **Tests**:
>     - Updates tests in `otelMapping.servertest.ts` and `IngestionService.integration.test.ts` to use `createIngestionEventSchema()`.
>     - Modifies `experimentsService.test.ts` to check for correct environment setting in traces.
>   - **Misc**:
>     - Bumps `langfuse-langchain` version to `3.38.3` in `package.json`.
>     - Removes duplicate export in `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 82948a43aa4c90053a61c1ece30c5fc8b24ea29a. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


marliessophie:
WARNING: do not merge before making required changes in TS SDK and upgrading PR to new version.

packages/shared/src/server/llm/fetchLLMCompletion.ts
```diff
@@ -145,6 +145,9 @@ export async function fetchLLMCompletion(
         await processEventBatch(
```
hassiebp, line 149: I think now that we are adding three more positional args to processEventBatch, we should move to a single config object rather than 5 positional args that in turn need comments to show what they are for / reduce bug risk

marliessophie: Agree, planned this too :) Should have made more clear this is still a draft :)

packages/shared/src/server/ingestion/types.ts
```diff
@@ -553,6 +553,34 @@ export const ingestionEvent = z.discriminatedUnion("type", [
...
+      const isEnvironmentRegexError =
+        issue.path?.includes("environment") &&
+        issue.message === ENVIRONMENT_NAME_REGEX_ERROR_MESSAGE;
+      return !isEnvironmentRegexError;
+    });
+
+    // Clear all issues and add back only the non-environment regex ones
+    ctx.issues.length = 0;
+    ctx.issues.push(...filteredIssues);
+  });
```
marliessophie: @maxdeichmann as discussed this feels a bit "unschön" - I will read docs and see if there is a cleaner zod-native solution for discriminated unions. If you have a better idea off the bat would love to hear it! Performance penalties should be negligible as >99.9% of all events should simply use the ingestionEvent schema with `isLangfuseInternal = false`.

marliessophie: Did not find a better approach from docs - would suggest timeboxing another 30 minutes for research and otherwise merge current solution, wdyt @maxdeichmann? :)

maxdeichmann: Given this is the most core object in Langfuse, we should try to find a nice solution here. Having tech debt on that part does not feel great.

packages/shared/src/server/index.ts
```diff
@@ -20,7 +20,6 @@ export * from "./clickhouse/schema";
 export * from "./repositories/definitions";
 export * from "../server/ingestion/types";
 export * from "./ingestion/processEventBatch";
-export * from "../server/ingestion/types";
```
marliessophie, line 24: Doublicated import statement

PR #7689, opened 2025-07-04
chore: create background migrations for traces amt tables

Steffen911:
> [!IMPORTANT]
> Introduces background migrations for traces using AggregatingMergeTrees, updates ingestion and writing processes, and modifies environment configurations and tests to support the new structure.
> 
>   - **Background Migrations**:
>     - Introduces `migrateTracesToTracesAMTs.ts` for migrating traces to AMTs.
>     - Creates `traces_null`, `traces_all_amt`, `traces_7d_amt`, `traces_30d_amt` tables and corresponding materialized views.
>   - **Environment Configurations**:
>     - Adds `CLICKHOUSE_CLUSTER_ENABLED` to `env.ts`.
>     - Updates `.env.dev-redis-cluster.example` with new experiment flags.
>   - **Ingestion and Writing**:
>     - Updates `IngestionService` to process and write to `traces_null`.
>     - Modifies `ClickhouseWriter` to handle `traces_null` table.
>   - **Testing**:
>     - Updates integration tests in `IngestionService.integration.test.ts` to validate new ingestion and writing logic.
>   - **Documentation**:
>     - Updates `README.md` in `IngestionService` to reflect changes in table structures and experimental setup.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 0b155796ccf32e482abb4b7cadd6811dc3b411d2. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -36,7 +36,7 @@ const getS3StorageServiceClient = (bucketName: string): StorageService => {
 export async function upsertClickhouse<
   T extends Record<string, unknown>,
 >(opts: {
-  table: "scores" | "traces" | "observations";
+  table: "scores" | "traces" | "observations" | "traces_mt";
```
Steffen911: Switch to traces_null table.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -1121,47 +1201,200 @@ export const deleteTracesOlderThanDays = async (
...
-      type: "trace",
-      kind: "delete",
-      projectId,
+    newExecution: async (input) => {
+      await Promise.all([
+        // Delete from traces
+        await commandClickhouse({
+          query: `
+            DELETE FROM traces
+            WHERE project_id = {projectId: String};
```
maxdeichmann, line 1341: Wow i did not think about this fall-out. This will become quite expensive for us.

worker/src/__tests__/evalService.test.ts
```diff
@@ -729,6 +729,7 @@ describe("eval service tests", () => {
       await upsertTrace({
         id: traceId,
         project_id: "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a",
+        user_id: "b",
```
maxdeichmann, line 732: intended?

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
```diff
@@ -0,0 +1,303 @@
...
+-- Create the all AMT
+CREATE TABLE traces_all_amt ON CLUSTER default
+(    
+    -- Identifiers
+    `project_id`         String,
+    `id`                 String,
+    `timestamp`          SimpleAggregateFunction(min, DateTime64(3)),  -- Backward compatibility: redundant with start_time
+    `start_time`         SimpleAggregateFunction(min, DateTime64(3)),
+    `end_time`           SimpleAggregateFunction(max, Nullable(DateTime64(3))),
+    `name`               SimpleAggregateFunction(anyLast, Nullable(String)),
```
maxdeichmann, line 50: Why is this nullable?

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
```diff
@@ -0,0 +1,303 @@
...
+
+    -- Metadata properties
+    `metadata`           SimpleAggregateFunction(minMap, Map(String, String)),
```
maxdeichmann: what does this minmap do here?

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
```diff
@@ -0,0 +1,303 @@
...
+    `user_id`            SimpleAggregateFunction(anyLast, Nullable(String)),
+    `session_id`         SimpleAggregateFunction(anyLast, Nullable(String)),
```
maxdeichmann, line 55: what was the reason here again not to take argMax?

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
```diff
@@ -0,0 +1,303 @@
...
+    `version`            SimpleAggregateFunction(anyLast, Nullable(String)),
+    `release`            SimpleAggregateFunction(anyLast, Nullable(String)),
+
+    -- UI properties
+    `bookmarked`         AggregateFunction(argMax, Nullable(Bool), DateTime64(3)),
+    `public`             AggregateFunction(argMax, Nullable(Bool), DateTime64(3)),
+
+    -- Aggregations
+    `observation_ids`    SimpleAggregateFunction(groupUniqArrayArray, Array(String)),
+    `score_ids`          SimpleAggregateFunction(groupUniqArrayArray, Array(String)),
```
maxdeichmann, line 67: what do we need these IDs for? This will be large in storage at some point.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
maxdeichmann: AggregateFunctions do not work woth skip indices, correct?

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
```diff
@@ -0,0 +1,303 @@
...
+    sumMap(t0.cost_details)                                                                    as cost_details,
+    sumMap(t0.usage_details)                                                                   as usage_details,
+
+    -- Input/Output
+    argMaxState(t0.input, if(t0.input <> '', t0.event_ts, toDateTime64(0, 3)))                 as input,
+    argMaxState(t0.output, if(t0.output <> '', t0.event_ts, toDateTime64(0, 3)))               as output,
+
+    min(t0.created_at)                                                                         as created_at,
+    max(t0.updated_at)                                                                         as updated_at
+FROM traces_mt t0
```
maxdeichmann: thy "t0"? this is uncommon naming, no?

maxdeichmann:
Why do we have convertObservationToTraceMt? I thought we write from the observations CH table via MV to the AMT.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -204,7 +204,7 @@ export const checkTraceExists = async ({
           t.project_id as project_id,
           -- Add a timestamp alias to ensure we can filter on it
           t.start_time as timestamp
-        FROM ${traceAmt} t
+        FROM ${traceAmt} t FINAL
```
maxdeichmann, line 207: Are all these changes intended here?

packages/shared/src/server/repositories/traces.ts
Steffen911: @maxdeichmann Theoretically, we could limit this to the traces_all_amt and use the default expiry of the 30d and 7d ones to make it cheaper. But that will probably be tough to share with users.
So with the current scope we'd have to delete from three tables.

worker/src/__tests__/evalService.test.ts
Steffen911: Yes, the test relied on some implicit behavior where this overwrote the user_id to "empty" as it wasn't set. Now we test explicitly that a change in the user_id triggers a cancellation.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Otherwise, a non-present value would overwrite the result to an empty string. To use anyLast effectively, we must keep the columns nullable.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: It takes the minimum value per key. So if someone sends us
```
{ foo: bar, bar: baz}
```
followed by
```
{ foo: baa } // earlier in alphabet than bar 
```
if would result in
```
{ foo: baa, bar: baz }
```

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Yes, we need to use simpleAggregateFunctions to add skip indices on columns.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Only if the cardinality per trace is insanely high. It's nice to keep them as we share those in the public API and that way, we can avoid joins that otherwise need to happen.
Can be dropped just as well though.

packages/shared/src/server/repositories/traces.ts
Steffen911: Yes, I added FINAL for initial simplicity during the implementation with the goal to focus on optimizations where necessary.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Yes, will update it when moving to the null table.

Steffen911:
> Why do we have convertObservationToTraceMt? I thought we write from the observations CH table via MV to the AMT.

It's a way to easily track the observation IDs we've seen (same for scores). Adding a MV for it felt like an overkill, but could be added later if we want to change this behaviour. Here, I use it, because I expect that we'll have further properties like cost and also some trace props that we'll derive from observations in the futuer.

maxdeichmann:
I am not sure why we need this:

We create the traces table based on the observations table with an MV within clickhouse. Why do we need to insert into the MT again here?

packages/shared/src/server/repositories/traces.ts
maxdeichmann: Ok

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
maxdeichmann: Ok, this feels like random sorting for our users. I am not sure we decided on this together, did we? Lets take this async as this is not a great ux.

packages/shared/src/server/repositories/traces.ts
maxdeichmann: Yes, agree. Otherwise this will remain in the UI for the next x days which is not acceptable.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
maxdeichmann: Ok, would it be possible to:

* set the column non nullable with a default value
* in the MV check if "", then defaultvalue, otherwise value

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
maxdeichmann: The entire column across the db will just get super large + the MV needs to always mutate this column. I have the intuition that this will break at some point.

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Use maxMap instead

packages/shared/clickhouse/migrations/clustered/0023_traces_aggregating_merge_trees.up.sql
Steffen911: Add clear comment in eng channel about this usage

Issue #7693, opened 2025-07-05
DELETE prompt public API


albcunha:
I saw this discussion regarding langfuse not exposing a delete method for prompt api:

_Thanks for raising this! Unfortunately right now, prompts cannot be deleted via api. This was an error on the changelog post which I just [fixed](https://github.com/langfuse/langfuse-docs/commit/477ae8b124d7ea635fa33e640bbf1d619b23e9e5)._

_Would you like to delete complete prompts via the API? Can you open a separate idea post for this to track this feature requests? Also happy to support a contribution if you are interested_

_Originally posted by @marcklingen in https://github.com/langfuse/langfuse/discussions/3503#discussioncomment-10775461_

I forked the project and made my version of and I'm willing to make a PR to add to main code base. So I'm opening this issue to discuss it. 

# The idea:
**Add the delete method into these files:**
- fern/apis/server/definition/prompts.yml
- web/public/generated/api/openapi.yml

**Add a deletePromptNameHandler function into promptNameHandler**
It would handle the requests from the api.
- web/src/features/prompts/server/handlers/promptNameHandler.ts** 

**add a deletePrompt.ts into prompts server actions**
They would pretty much copy the logic from web/src/features/prompts/server/routers/promptRouter.ts
 into this file:
- web/src/features/prompts/server/actions/deletePrompt.ts

**Add tests**
Add tests to this file:
- web/src/__tests__/async/prompts.v2.servertest.ts

# Issues I'm seeing
**Only full delete or a more complete method?**
We can just delete the full prompt and its labels and versions or add a filter using:
1. one label
1. one label + version

**URL and how to pass prompt information (Most important)**
I thought of uising the same url of GET prompt, this one: /api/public/v2/prompts/{promptName}​
If the user want to pass label or version to filter the delete, he coul pass it as queryparam, in the same way this method does. The problem is that now **prompt management accept folders**, so we would have these options:
1. pass it as path parameter and warn the user to user %2F as encoding for `/` folder division, or we could use an approach similar to the private trpc method used on the web application, in which we could pass {promptName} (or revamp all other methods to use all-catch path recognition from next-js
2. pass promptName as an queryparam, something like path=folder%2Fpromptname
3. pass promptName, label and version as payloas using json
4. Instead of using promptName, we use promptId, but it would make sense to allow promptId into to /api/public/v2/prompts/{promptName}​.

I think the best option is 1 or 3, so I'd like to know if you are interested on the PR.

[...]

PR #7704, opened 2025-07-06
Add DELETE Method to Prompt Public API

saisharan0103:
Summary

This Pull Request introduces a DELETE endpoint to Langfuse's public Prompt API, enabling deletion of prompts with optional filtering by label and version.

Changes Implemented

API Definition:

Defined DELETE method in fern/apis/server/definition/prompts.yml.

Included optional label and version query parameters.

OpenAPI Documentation:

Updated web/public/generated/api/openapi.yml with the DELETE endpoint documentation, clearly outlining the parameters and expected responses.

Backend Implementation:

Created a new handler (deletePromptNameHandler) in web/src/features/prompts/server/handlers/promptNameHandler.ts.

Implemented action logic (deletePrompt) in web/src/features/prompts/server/actions/deletePrompt.ts to handle deletion, caching, and concurrency controls.

Testing:

Added comprehensive unit tests in web/src/tests/async/prompts.v2.servertest.ts covering both unfiltered deletions and deletions filtered by label/version.

Impact

Enables users to programmatically manage prompts, improving usability and completeness of the Langfuse API.

Related Issues

Resolves issue #7693

Reviewer Notes

Verify encoding of path parameters, especially handling of folders (%2F).

Ensure the deletion process aligns with overall project architecture.

Checklist

Thank you for reviewing this PR!

### DIFF
--- a/fern/apis/server/definition/prompts.yml
+++ b/fern/apis/server/definition/prompts.yml
@@ -2,6 +2,7 @@
 imports:
   commons: ./commons.yml
   pagination: ./utils/pagination.yml
+  scim: ./scim.yml
 service:
   auth: true
   base-path: /api/public/v2
@@ -56,6 +57,25 @@ service:
       path: /prompts
       request: CreatePromptRequest
       response: Prompt
+
+    delete:
+      docs: Delete a prompt or specific versions
+      method: DELETE
+      path: /prompts/{promptName}
+      path-parameters:
+        promptName:
+          type: string
+          docs: The name of the prompt
+      request:
+        name: DeletePromptRequest
+        query-parameters:
+          label:
+            type: optional<string>
+            docs: Optional label of the prompt to delete
+          version:
+            type: optional<integer>
+            docs: Optional version of the prompt to delete
+      response: scim.EmptyResponse
 
 types:
   PromptMetaListResponse:
--- a/web/src/features/prompts/server/actions/deletePrompt.ts
+++ b/web/src/features/prompts/server/actions/deletePrompt.ts
@@ -1,0 +1,47 @@
+import { InvalidRequestError, LangfuseNotFoundError } from "@langfuse/shared";
+import { prisma } from "@langfuse/shared/src/db";
+import { PromptService, redis, logger } from "@langfuse/shared/src/server";
+
+export type DeletePromptParams = {
+  promptName: string;
+  projectId: string;
+  version?: number | null;
+  label?: string;
+};
+
+export const deletePrompt = async (params: DeletePromptParams) => {
+  const { promptName, projectId, version, label } = params;
+
+  if (version && label) {
+    throw new InvalidRequestError("Cannot specify both version and label");
+  }
+
+  const where = {
+    projectId,
+    name: promptName,
+    ...(version ? { version } : {}),
+    ...(label ? { labels: { has: label } } : {}),
+  };
+
+  const prompts = await prisma.prompt.findMany({ where });
+
+  if (prompts.length === 0) {
+    throw new LangfuseNotFoundError("Prompt not found");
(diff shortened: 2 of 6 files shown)


saisharan0103:
Yes @nimarb , 
Accepted the CLA
Thanks.

fern/apis/server/definition/prompts.yml
```diff
@@ -59,6 +60,25 @@ service:
...
+      path-parameters:
+        promptName:
+          type: string
+          docs: The name of the prompt
+      request:
+        name: DeletePromptRequest
+        query-parameters:
+          label:
+            type: optional<string>
+            docs: Optional label of the prompt to delete
```
hassiebp, line 76: Probably should be more specific here that the underlying prompt is deleted, not the label of the prompt (wording is slightly ambiguous here)

hassiebp: Also, this will delete _all_ prompt versions with that label

web/src/features/prompts/server/actions/deletePrompt.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw new InvalidRequestError(
+      `Other prompts are depending on prompt versions you are trying to delete:\n\n${dependencyMessages}\n\nPlease delete the dependent prompts first.`,
+    );
+  }
+
+  const promptService = new PromptService(prisma, redis);
+
+  try {
+    await promptService.lockCache({ projectId, promptName });
+    await promptService.invalidateCache({ projectId, promptName });
```
hassiebp, line 82: Looks like redis invalidation is in place 👍🏾

web/src/features/prompts/server/actions/deletePrompt.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    await prisma.prompt.deleteMany({
+      where: { projectId, id: { in: prompts.map((p) => p.id) } },
+    });
```
hassiebp, line 86: If the `latest` label was deleted, we have to reattach it to the prompt version with the latest `createdAt`

fern/apis/server/definition/prompts.yml
hassiebp: We should also make clear if when neither `label` nor `version` is defined, _all_ prompt versions are deleted.

web/src/features/prompts/server/actions/deletePrompt.ts
```diff
@@ -0,0 +1,93 @@
...
+    label,
+    prompts: providedPrompts,
+  } = params;
+
+  if (version && label) {
+    throw new InvalidRequestError("Cannot specify both version and label");
+  }
+
+  let prompts = providedPrompts;
+  if (!prompts) {
```
hassiebp, line 27: is this necessary? prompts are already fetched and passed in `promptNameHandler`

Also clearer naming would be `promptVersions`

web/src/features/prompts/server/actions/deletePrompt.ts
```diff
@@ -0,0 +1,93 @@
...
+      p."version" AS "parent_version",
+      pd."child_version" AS "child_version",
+      pd."child_label" AS "child_label"
+    FROM
+      prompt_dependencies pd
+      INNER JOIN prompts p ON p.id = pd.parent_id
+    WHERE
+      p.project_id = ${projectId}
+      AND pd.project_id = ${projectId}
+      AND pd.child_name = ${promptName}
```
hassiebp, line 62: I think the dependend check here is not correct - we need to check that there is no entry in `prompt_dependencies` where the `parent_id` is one of the prompt version IDs that are to be deleted

The filter here is checking for when `promptName` is a _child_

PR #7807, opened 2025-07-10
feat: automation secrets

coffee4tw:
## What does this PR do?

<div>
    <a href="https://www.loom.com/share/9eac47163f764b7e8f2a8d1177788d1f">
      <p>Managing Webhook Secrets: Creation and Update Process - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/9eac47163f764b7e8f2a8d1177788d1f">
      
    </a>
  </div>

Adds the ability to mark automation headers as secret, which stores them encrypted and only returns an obfuscated version to the UI.
Ensures that create and update API operations do not return encrypted fields.

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Adds support for managing secret webhook headers by encrypting them, handling their display values, and ensuring they are not exposed in API responses.
> 
>   - **Behavior**:
>     - Adds `secretHeaderKeys` and `displayHeaderValues` to `WebhookActionConfigSchema` in `automations.ts` for managing secret headers.
>     - Updates `getActionByIdWithSecrets` in `automation-repository.ts` to decrypt secret headers before execution.
>     - Modifies `processWebhookActionConfig` in `webhookHelpers.ts` to encrypt secret headers and generate display values.
>     - Ensures API responses do not expose secret headers by using `convertToSafeWebhookConfig` in `router.ts`.
>   - **UI**:
>     - Adds `isSecret` toggle for headers in `WebhookActionForm.tsx` to mark headers as secret.
>     - Updates `WebhookActionHandler.tsx` to handle secret headers in form data.
>   - **Tests**:
>     - Adds tests in `automations-trpc.servertest.ts` to verify secret headers are not exposed in API responses.
>     - Includes integration tests in `webhooks.test.ts` to ensure correct handling of secret headers during webhook execution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 67eb97b77ce0ea2d4a234917629a60ec58c0ac9f. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/automation-repository.ts
```diff
@@ -35,12 +36,23 @@ export const getActionByIdWithSecrets = async ({
...
+        decryptedRequestHeaders: {},
+      };
+
   return {
     ...actionConfig,
     config: {
       type: config.type,
       url: config.url,
-      headers: config.headers,
+      headers: decryptedHeaders,
```
maxdeichmann: this one can be removed now, right?

web/src/features/automations/server/webhookHelpers.ts
```diff
@@ -40,24 +55,100 @@ export async function processWebhookActionConfig({
...
+    }
+
+    // If value is empty, preserve existing value if it exists
+    if (headerObj.value.trim() === "" && existingHeader) {
+      finalRequestHeaders[key] = existingHeader;
+    } else if (headerObj.value.trim() !== "") {
+      // Only process non-empty values
+      finalRequestHeaders[key] = headerObj;
+    }
+    // If value is empty and no existing header, skip it (effectively removing it)
```
maxdeichmann: For the length of string checks: we could also do this on the zod level. This way it will direclty be rejected from the API and from the form.

packages/shared/src/server/repositories/automation-repository.ts
coffee4tw: No, the executeWebhook still uses that field. We can take it out and modify that code to explicitly use requetHeaders though if you think that makes it clearer.

web/src/features/automations/server/webhookHelpers.ts
coffee4tw: Ah, I didn't think of that. Will propose a change!

coffee4tw: Actually, I don't see how because it's not just dependent on the current object but the previous values. The check is effectively for flipping the secret bool. Otherwise we want to allow empty values because they preserve the existing value.

maxdeichmann: Ah ok, thanks for the explanation.

PR #7816, opened 2025-07-11
feat(table): persist column sizes in local storage

marliessophie:
> [!IMPORTANT]
> Add feature to persist table column sizes in local storage using a new `useColumnSizing` hook and `tableName` prop in `DataTable`.
> 
>   - **Behavior**:
>     - Persist table column widths to `localStorage` using `useColumnSizing` hook in `data-table.tsx`.
>     - Introduces `tableName` prop in `DataTable` component to uniquely identify tables.
>   - **Hooks**:
>     - New `useColumnSizing` hook in `useColumnSizing.ts` to manage column sizing state with debounced saving to `localStorage`.
>   - **Components**:
>     - Add `tableName` prop to `DataTable` instances in `models.tsx`, `observations.tsx`, and `score-configs.tsx` ... and 20 other places.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 1694dd2bc9788c3602be8459e66abc5c76601086. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/components/table/use-cases/observations.tsx
+++ b/web/src/components/table/use-cases/observations.tsx
@@ -995,6 +995,7 @@ export default function ObservationsTable({
         }}
       />
       <DataTable
+        tableName={"observations"}
         columns={columns}
         peekView={peekConfig}
         data={
--- a/web/src/components/table/data-table.tsx
+++ b/web/src/components/table/data-table.tsx
@@ -38,6 +38,7 @@ import { type PeekViewProps } from "@/src/components/table/peek/hooks/usePeekVie
 import { usePeekView } from "@/src/components/table/peek/hooks/usePeekView";
 import { isEqual } from "lodash";
 import { useRouter } from "next/router";
+import { useColumnSizing } from "@/src/components/table/hooks/useColumnSizing";
 
 interface DataTableProps<TData, TValue> {
   columns: LangfuseColumnDef<TData, TValue>[];
@@ -65,6 +66,7 @@ interface DataTableProps<TData, TValue> {
   peekView?: PeekViewProps<TData>;
   pinFirstColumn?: boolean;
   hidePagination?: boolean;
+  tableName: string;
 }
 
 export interface AsyncTableData<T> {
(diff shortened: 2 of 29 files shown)


web/src/components/table/use-cases/observations.tsx
```diff
@@ -995,6 +995,7 @@ export default function ObservationsTable({
         }}
       />
       <DataTable
```
marliessophie, line 997: Please note we will attempt a wider data-table refactor. We have multiple declarations of table name types I did not want to add to. It would have been a too large detour to decouple types in the scope of this PR. Have created a follow up ticket to be addressed soon.

web/src/components/table/use-cases/observations.tsx
```diff
@@ -995,6 +995,7 @@ export default function ObservationsTable({
...
+        tableName={"observations"}
```
maxdeichmann, line 998: Do you think we should have a small announcement in the engineering channel so that we can use this across features?

marliessophie: I don't understand, could you please rephrase @maxdeichmann?

maxdeichmann: I mean: we are now adding this table name. Before, we used different keys to access local storage for table heigth, column visibility etc. We should have one place in the code base where we define the name per table. I guess that this should be the one going forward or not?

marliessophie: Please see my [comment](https://github.com/langfuse/langfuse/pull/7816#discussion_r2200127871) that I had posted pre-review, this is the answer to your question. After we complete the refactor, I will share PR with short brief in the eng channel.

maxdeichmann: Yes, thanks for the comment. I was only wondering whether we should let everyone know that we should be using this as of now. But as you are the one working on it the most, i think we can just proceed like that.

PR #7817, opened 2025-07-11
chore: add projectId to prices table

Steffen911:
> [!IMPORTANT]
> Add `projectId` to `prices` table and update related schema, types, API, and tests.
> 
>   - **Database Schema**:
>     - Add `project_id` column to `prices` table in `migration.sql`.
>     - Add foreign key constraint on `project_id` referencing `projects` table.
>     - Backfill `project_id` in `prices` table using `model_id`.
>   - **Prisma Schema**:
>     - Update `Price` model in `schema.prisma` to include `projectId` and relation to `Project`.
>     - Update `Project` model to include relation to `Price`.
>   - **TypeScript Types**:
>     - Add `project_id` to `Price` type in `types.ts`.
>   - **API Changes**:
>     - Update `index.ts` to include `projectId` when creating prices.
>     - Update `models.ts` to handle `projectId` in `upsert` and `getAll` functions.
>   - **Tests**:
>     - Update `model-definitions.servertest.ts` to include `projectId` in price creation.
>     - Update `IngestionService.integration.test.ts` and `calculateTokenCost.unit.test.ts` to handle `projectId` in price-related tests.
>   - **Misc**:
>     - Update `.claude/settings.local.json` to allow `ls` and `mkdir` commands.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 92b957cb6ba29861f9bd9b750f45d196c0c22550. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


.claude/settings.local.json
hassiebp, line 1: I think we should exclude this file from version control as it tracks the local allowlist for claude code that might not be intended to be shared across devs.

Instead we can delete this file and add pattern *.local.*  to gitignore

packages/shared/prisma/migrations/20250711105322_prices_add_project_id/migration.sql
```diff
@@ -0,0 +1,13 @@
...
+
+-- AddForeignKey
+ALTER TABLE "prices"
+    ADD CONSTRAINT "prices_project_id_fkey" FOREIGN KEY ("project_id") REFERENCES "projects" ("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- BackfillData
+UPDATE "prices"
+SET "project_id" = (SELECT "models"."project_id"
+                    FROM "models"
+                    WHERE "models"."id" = "prices"."model_id");
```
hassiebp, line 13: Just to double check, this will handle NULLs correctly in `model.project_id` , right?

Steffen911: @hassiebp Essentially, it will overwrite them to "NULL", so yes.

PR #7912, opened 2025-07-16
feat(dataset-run-items): migrate from postgres to clickhouse; write to CH API-only 

marliessophie:
> [!IMPORTANT]
> Migrate dataset run items from Postgres to ClickHouse, updating schemas, ingestion services, and APIs to support the new storage system.
> 
>   - **Migration**:
>     - Migrate dataset run items from Postgres to ClickHouse using `MigrateDatasetRunItemsFromPostgresToClickhouse` in `migrateDatasetRunItemsFromPostgresToClickhouse.ts`.
>     - Add `executeWithDatasetRunItemsStrategy()` in `datasetExecution.ts` for dual-write strategy.
>   - **Schema and Conversion**:
>     - Add `DatasetRunItemRecordInsertType` and `DatasetRunItemRecordReadType` in `definitions.ts`.
>     - Add conversion functions in `dataset-run-items-converters.ts`.
>   - **Ingestion and Writing**:
>     - Update `IngestionService` to process dataset run item events and write to ClickHouse.
>     - Add `createDatasetRunItemsCh()` in `clickhouse-helpers.ts` for inserting dataset run items.
>   - **API and Utilities**:
>     - Update API routes in `dataset-router.ts` and `dataset-run-items.ts` to handle dataset run items.
>     - Add `createOrFetchDatasetRun()` in `dataset-runs.ts` for run management.
>   - **Testing and Documentation**:
>     - Add tests in `datasets-ui-table.servertest.ts` for dataset run items.
>     - Update `README.md` with experimental setup for dataset run items.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for bc8550f4a8346b28d9ab542b3e7861d20bdbef76. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/datasets/server/service.ts
```diff
@@ -58,7 +58,6 @@ export const createDatasetRunsTableWithoutMetrics = async (
   const runs = await getDatasetRunsFromPostgres(input);
 
   return runs.map(({ run_items, ...run }) => ({
-    ...run,
```
marliessophie: _Note:_ I figured this was a mistake before hence removed it.

packages/shared/src/server/ingestion/processEventBatch.ts
```diff
@@ -313,6 +320,8 @@ const isAuthorized = (
     );
   }
 
+  // TODO: Add authorization for dataset run items?
```
marliessophie: _Note:_ We have a specific authorization for scores, why? Makes me wonder if we need one for DRIs

worker/src/features/experiments/experimentServiceClickhouse.ts
marliessophie, line 1: _Note_: previously whenever parts of the run configuration were invalid (prompt was deleted, api key for provider not found etc), we would update the dataset run metadata with the error. This was a temporary solution to begin with. Now during the course of this refactor, we wanted to move away from this behaviour. First of all, these error cases might possibly occur during the course of an experiment. Let's take an example: we run an experiment on 1000 items in a given dataset. The first X are successful, then we run into api provider rate limits. After some delay, we attempt the job again (same logic as in eval service). Now, the user might have deleted their prompt/llm api keys linked to the experiment in the meantime. Previously, we would have just updated the metadata on the run and exited the job. Now, we create dataset run items for the experiment anyways, adding the error on run item level. We create a DRI for each item in the dataset, but skip all for which there exits a DRI already. For implementation details, please see `createAllDatasetRunItemsWithConfigError`. Happy to discuss if there are questions. 

In terms of magnitude we have cases in the low 10s across environments where this has happened - so it's not very common.

worker/src/backgroundMigrations/migrateDatasetRunItemsFromPostgresToClickhouse.ts
marliessophie, line 1: _Note:_ Please pay special attention that this works as intended.

packages/shared/scripts/seeder/utils/clickhouse-builder.ts
```diff
@@ -42,6 +44,16 @@ export class ClickHouseQueryBuilder {
...
 
+  /**
+   * Creates INSERT query for dataset run items data using VALUES syntax.
+   * Use for: Small datasets, dataset run items that link to postgres data (e.g. dataset runs)
+   */
+  async executeDatasetRunItemsInsert(
+    datasetRunItems: DatasetRunItemRecordInsertType[],
+  ): Promise<InsertResult> {
+    return await createDatasetRunItemsCh(datasetRunItems);
+  }
```
maxdeichmann, line 55: This function is one line long. Please remove this abstraction.

packages/shared/scripts/seeder/utils/data-generators.ts
```diff
@@ -60,6 +64,49 @@ export class DataGenerator {
...
+      dataset_run_metadata: { key: "value" },
+      dataset_item_id: generateDatasetItemId(
+        input.datasetName,
+        input.itemIndex,
+        projectId,
+        input.runNumber || 0,
+      ),
+      dataset_item_input: input.item.input,
+      dataset_item_expected_output: input.item.output,
+    });
```
maxdeichmann, line 107: Do we know that for these datase_item_ids and trace_ids, data exists in postgres?

packages/shared/scripts/seeder/utils/clickhouse-builder.ts
marliessophie: Just followed the pattern in this file. As this is seeder only code, think it's OK to keep. If you have strong feelings can remove all abstractions in this file.

packages/shared/src/server/ingestion/types.ts
```diff
@@ -595,6 +621,7 @@ const createAllIngestionSchemas = (
     generationCreateEvent,
     generationUpdateEvent,
     scoreEvent,
+    datasetRunItemCreateEvent,
```
maxdeichmann, line 633: Now, this is part of the public ingestion API schema, right? Can we only have this internally?

packages/shared/src/server/ingestion/processEventBatch.ts
maxdeichmann: No, we do not need anything. Scores can be ingested from our JS front-end only SDK which only sends the public key as auth. We needed to build this as we cannot bake secret keys in the front-end code of our users.

packages/shared/src/server/repositories/dataset-run-items-converters.ts
```diff
@@ -0,0 +1,83 @@
...
+    datasetItemId: row.dataset_item_id,
+    datasetItemInput: row.dataset_item_input,
+    datasetItemExpectedOutput: row.dataset_item_expected_output,
+    datasetItemMetadata: parseMetadataCHRecordToDomain(
+      row.dataset_item_metadata,
+    ),
+    createdAt: new Date(row.created_at),
+    updatedAt: new Date(row.updated_at),
+    datasetId: row.dataset_id,
+    error: row.error ?? null,
```
maxdeichmann, line 84: For all timestamps in this function, please use: parseClickhouseUTCDateTimeFormat

worker/src/services/IngestionService/index.ts
```diff
@@ -87,6 +92,17 @@ const immutableEntityKeys: {
     "created_at",
     "environment",
   ],
+  // TODO: review immutable keys for dataset run items
```
maxdeichmann: is there anything left to do here?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  if (!datasetId && !datasetName) {
+    return {
+      success: false,
+      error: "Dataset id or dataset name is required",
+    };
+  }
+
+  let datasetIdToUse = datasetId;
```
maxdeichmann, line 56: Can you just use datasetId everywhere?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+  if (!datasetId && datasetName) {
+    const dataset = await prisma.dataset.findFirst({
+      where: {
+        name: datasetName,
+        projectId,
+      },
+    });
```
maxdeichmann: Why do we have this condition? Couldnt we just fetch by id here?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+      error: string;
+    };
+
+export const validateDatasetRunAndFetch = async (params: {
+  projectId: string;
+  datasetId?: string;
+  datasetName?: string;
+  runName?: string;
+  runId?: string;
+}): Promise<ValidateDatasetRunAndFetchReturn> => {
```
maxdeichmann: Why do we have all these nullables?

packages/shared/src/server/repositories/dataset-run-items.ts
maxdeichmann: Couldnt we make this much simpler by something like:

const datasetIdToUse = datasetId ?? (await prisma.dataset.findUnique({
      where: {
        name: datasetName,
        projectId,
      },
    }))

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+      error:
+        "Dataset item not found for the given project, dataset id and item id or is not active",
+    };
+  }
+
+  return {
+    success: true,
+    datasetItem: datasetItem,
+  };
+};
```
maxdeichmann: Can we remove this function entirely? I think this is just a prisma call and does not validate anything.

packages/shared/src/server/repositories/dataset-run-items.ts
maxdeichmann: In the event.body from the ingestion pipeline we have this type:

```
body: {
    traceId: string;
    datasetId: string;
    runId: string;
    datasetItemId: string;
    id?: string | null | undefined;
    observationId?: string | null | undefined;
    error?: string | null | undefined;
    input?: string | number | boolean | JsonNested[] | {
        [key: string]: JsonNested;
    } | null | undefined;
    expectedOutput?: string | number | boolean | JsonNested[] | {
        [key: string]: JsonNested;
    } | null | undefined;
    createdAt?: string | null | undefined;
}
```

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+const isUniqueConstraintError = (error: any): boolean => {
+  return (
+    error.code === "P2002" || // Prisma unique constraint
+    error.message?.includes("duplicate key") ||
+    error.message?.includes("UNIQUE constraint") ||
+    error.message?.includes("violates unique constraint")
+  );
+};
+
+export const getDatasetRunItemsByRunId = async ({
```
maxdeichmann: This is unused, right?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -0,0 +1,231 @@
...
+    };
+  }
+
+  return {
+    success: true,
+    datasetItem: datasetItem,
+  };
+};
+
+export const createOrFetchDatasetRun = async ({
```
maxdeichmann: This is just copy-paste, right?

maxdeichmann: If yes, i would have jsut left it where it was as this just makes the change larger and the repository here is about dri and not dr

web/src/features/datasets/server/dataset-router.ts
```diff
@@ -811,7 +813,9 @@ export const datasetRouter = createTRPCRouter({
 
       return;
     }),
-
+  /**
+   * @deprecated Use runItemsByRunId or runItemsByItemId instead
+   */
```
maxdeichmann: This here is trpc. Can we remove this already?

web/src/features/datasets/server/service.ts
maxdeichmann: Why did you make this change?

packages/shared/src/server/repositories/dataset-run-items.ts
maxdeichmann: I just realized this function is also used in e.g. the public API to delete. I think this abstraction only adds confusion. I would keep the delete endpoint as is and have an implementation for the  ingestion pipeline here.

web/src/pages/api/public/datasets/[name]/runs/[runName].ts
```diff
@@ -66,45 +67,34 @@ export default withMiddlewares({
     rateLimitResource: "datasets",
     fn: async ({ query, auth }) => {
       // First get the dataset run to check if it exists
-      const datasetRuns = await prisma.datasetRuns.findMany({
```
Comment (maxdeichmann) [inline web/src/pages/api/public/datasets/[name]/runs/[runName].ts]: I think we can revert the changes here.

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -2,129 +2,190 @@ import { prisma } from "@langfuse/shared/src/db";
...
+    fn: async ({ body, auth, res }) => {
+      return await executeWithDatasetRunItemsStrategy({
+        input: body,
+        operationType: DatasetRunItemsOperationType.WRITE,
+        postgresExecution: async (bodyInput: typeof body) => {
+          /**************
+           * VALIDATION *
+           **************/
+
+          const result = await validateCreateDatasetRunItemBodyAndFetch(
```
maxdeichmann: Can we also remove the validation here to the old code?

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -2,129 +2,190 @@ import { prisma } from "@langfuse/shared/src/db";
...
+            throw new LangfuseNotFoundError(result.error);
+          }
+
+          const { datasetItem, traceId, observationId } = result;
+
+          /********************
+           *   RUN CREATION    *
+           ********************/
+
+          const run = await createOrFetchDatasetRun({
```
maxdeichmann: Does it make sense to only wrap the run creation in the helper function executeWithDatasetRunItemsStrategy? This way the change is way more focused.

packages/shared/src/domain/dataset-run-items.ts
```diff
@@ -0,0 +1,28 @@
...
+import { MetadataDomain } from "./traces";
+
+export const DatasetRunItemSchema = z.object({
+  id: z.string(),
+  projectId: z.string(),
+  datasetRunId: z.string(),
+  datasetItemId: z.string(),
+  datasetId: z.string(),
+  traceId: z.string(),
+  observationId: z.string().nullable(),
```
Steffen911, line 12: How do those actually work? Do we expect a deterministic traceId when creating the item or do they arrive via updates?

It seems odd that we cannot track the `output` here, but the traceId.

packages/shared/src/env.ts
```diff
@@ -108,7 +108,13 @@ const EnvSchema = z.object({
...
   LANGFUSE_CLICKHOUSE_QUERY_MAX_ATTEMPTS: z.coerce.number().default(3), // Maximum attempts for socket hang up errors
   LANGFUSE_SKIP_S3_LIST_FOR_OBSERVATIONS_PROJECT_IDS: z.string().optional(),
-
+  // Dataset Run Items Migration Environment Variables
+  LANGFUSE_DATASET_RUN_ITEMS_WRITE_CH: z
+    .enum(["true", "false"])
+    .default("false"),
+  LANGFUSE_DATASET_RUN_ITEMS_READ_CH: z
+    .enum(["true", "false"])
+    .default("false"),
```
Steffen911: You could make their use explicit with `LANGFUSE_EXPERIMENT`. That way users don't get ideas if they search through the code. I've seen a couple of experimental flags being used when doing the migration in November. 😄

packages/shared/src/server/dataset-run-items/datasetExecution.ts
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable no-unused-vars */
```
Steffen911: I'd prefer a line-by-line disable here.

worker/src/backgroundMigrations/migrateDatasetRunItemsFromPostgresToClickhouse.ts
Steffen911: LGTM!

worker/src/services/IngestionService/index.ts
```diff
@@ -135,9 +151,80 @@ export class IngestionService {
...
+    const finalDatasetRunItemRecords: (DatasetRunItemRecordInsertType | null)[] =
+      await Promise.all(
+        datasetRunItemEventList.map(async (event: DatasetRunItemEventType) => {
+          const enrichedItem = await enrichedDatasetRunItem({
+            body: event.body,
+            runItemId: entityId,
+            projectId,
+          });
+
+          if (!enrichedItem) return null;
```
Steffen911: You can also just use flatMap here and return `[]` for empty items. That way you essentially drop the case where elements can be null within the result.

web/src/features/public-api/server/dataset-run-items.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  let finalTraceId = traceId;
+
+  // Backwards compatibility: historically, dataset run items were linked to observations, not traces
+  if (!traceId && observationId) {
+    const observation = await getObservationById({
+      id: observationId,
+      projectId,
+      fetchWithInputOutput: true,
+    });
```
Steffen911: This is also slightly at odds with what we discussed - if we expect that the trace/observation isn't complete on dataset run item upsert this query should almost always return `null`. Can we somehow check if this is still being used?

web/src/features/public-api/server/dataset-run-items.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const datasetItem = await prisma.datasetItem.findUnique({
+    where: {
+      id_projectId: {
+        projectId,
+        id: datasetItemId,
+      },
+      status: "ACTIVE",
+    },
+  });
```
Steffen911: This will return the full body including input/expected output. If it should only be an existence check, we can reduce network and IO by selecting the identifiers only or returning 1 if found.
In case we expect this to be called hundreds of times during an execution this may help significantly in terms of performance.

worker/src/services/IngestionService/index.ts
```diff
@@ -135,9 +151,80 @@ export class IngestionService {
...
+    if (datasetRunItemEventList.length === 0) return;
+
+    const finalDatasetRunItemRecords: (DatasetRunItemRecordInsertType | null)[] =
+      await Promise.all(
+        datasetRunItemEventList.map(async (event: DatasetRunItemEventType) => {
+          const enrichedItem = await enrichedDatasetRunItem({
+            body: event.body,
+            runItemId: entityId,
+            projectId,
+          });
```
Steffen911: I agree with Max's points from above. The call here seems to be mainly a prisma wrapper that adds an unnecessary abstraction. I'd drop it and inline the logic.

packages/shared/src/server/ingestion/processEventBatch.ts
```diff
@@ -256,11 +256,18 @@ export const processEventBatch = async (
...
+
+      const isOtelOrSkipS3Project =
         authCheck.scope.projectId !== null &&
-        (projectIdsToSkipS3List.includes(authCheck.scope.projectId) ||
-          source === "otel");
+        (source === "otel" ||
+          projectIdsToSkipS3List.includes(authCheck.scope.projectId));
+
+      const shouldSkipS3List =
+        isDatasetRunItemEvent || (isObservationEvent && isOtelOrSkipS3Project);
```
Steffen911, line 270: At this point this may need a comment or a dedicated function with an explanation of what is going on.

packages/shared/src/server/ingestion/enrichedDatasetRunItem.ts
```diff
@@ -0,0 +1,55 @@
+import { DatasetRunItemDomain } from "../../domain/dataset-run-items";
```
Steffen911: IMO we should drop the whole file and inline the behaviour.

web/src/features/datasets/server/service.ts
marliessophie: It says in the note comment above why -- "I figured this was a mistake"

packages/shared/src/domain/dataset-run-items.ts
marliessophie: We expect a deterministic traceId when creating the item. 

Re tracking output - I'd refer back to our conversation with Max. This would require breaking changes on API/SDK.

packages/shared/src/env.ts
marliessophie: Thanks for the advice! Renamed ✅

packages/shared/src/server/ingestion/types.ts
marliessophie: Would require more logic. I did not add it to the public API types. Also, I don't see why we would not allow ingestion at some point in the future. If you feel this being internal is a requirement rather than a nice to have let's quickly chat. Otherwise I'd suggest to leave as is.

web/src/features/datasets/server/dataset-router.ts
marliessophie: Only when we add the CH writes. Removing deprecated flag from this PR.

packages/shared/src/server/repositories/dataset-run-items.ts
marliessophie: Yes, removed.

web/src/features/public-api/server/dataset-run-items.ts
marliessophie: Good shout!

packages/shared/src/server/repositories/dataset-run-items.ts
marliessophie: I moved it. I added the unique constraint handing, can delete if preferred.

packages/shared/src/server/repositories/dataset-run-items.ts
marliessophie: Agree. Also noticed when re-reading code.

worker/src/services/IngestionService/index.ts
marliessophie: Yes, a question to you: technically all keys on DRIs are immutable, should I include all of them here?

web/src/features/public-api/server/dataset-run-items.ts
marliessophie: We don't have logs but to my knowledge no one has complained about hitting this case.  

Context: This was built for backwards compatibility where there still might be some people posting DRI with traceId NOT included in body but only observation Id. In this case to retrieve the traceId we need to fetch the observation. 

Maybe it's time to deprecate the legacy behaviour ie reject any requests without a traceId.

web/src/pages/api/public/dataset-run-items.ts
```diff
@@ -65,66 +71,126 @@ export default withMiddlewares({
...
-          description: runDescription ?? undefined,
-          datasetId: datasetItem.datasetId,
-          metadata: metadata ?? undefined,
-          projectId: auth.scope.projectId,
-        },
-        update: {
-          metadata: metadata ?? undefined,
-          description: runDescription ?? undefined,
-        },
+      const run = await createOrFetchDatasetRun({
```
marliessophie, line 81: Note to self: check this

worker/src/services/IngestionService/index.ts
maxdeichmann: Probably this makes sense then.

packages/shared/src/server/ingestion/types.ts
marliessophie: Sync: invest 10 minutes.

web/src/features/public-api/server/dataset-run-items.ts
marliessophie: Will not address in this PR.

PR #7996, opened 2025-07-21
fix(sessions): infer session env from trace

Steffen911:
> [!IMPORTANT]
> In `IngestionService`, `processTraceEventList` now infers and updates session environment from trace data during `traceSession.upsert`.
> 
>   - **Behavior**:
>     - In `IngestionService`, `processTraceEventList` now includes `environment` in `create` and `update` operations for `traceSession.upsert`.
>     - Ensures session environment is inferred from trace data.
>   - **Misc**:
>     - Removed empty `update` object in `traceSession.upsert` call.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 559a19cbffb706fe0a5e2eb7a87509d29650ba75. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/worker/src/services/IngestionService/index.ts
+++ b/worker/src/services/IngestionService/index.ts
@@ -310,8 +310,11 @@ export class IngestionService {
           create: {
             id: finalTraceRecord.session_id,
             projectId,
+            environment: finalTraceRecord.environment,
           },
-          update: {},
+          update: {
+            environment: finalTraceRecord.environment,
+          },
         });
       } catch (e) {
         if (


mikeherring:
@Steffen911 I have a question on this - if different environments happen to use the same `session_id` for traces, is the upsert going to now merge those traces into one session? That doesn't seem right. I would expect sessions to only include traces from one environment. Or does `session_id` have to be unique across all environments?

PR #8179, opened 2025-07-29
feat(dataset-run-items): support reads for public APIs 

marliessophie:
> [!IMPORTANT]
> Add support for reading dataset run items via public APIs using Clickhouse, enhancing data retrieval with filtering, ordering, and pagination.
> 
>   - **Public API Enhancements**:
>     - Added `generateDatasetRunItemsForPublicApi` and `getDatasetRunItemsCountForPublicApi` in `dataset-run-items.ts` to fetch dataset run items and their count using Clickhouse.
>     - Updated `GET /dataset-run-items` and `GET /datasets/{name}/runs/{runName}` endpoints to use new functions for Clickhouse execution.
>   - **Repository Changes**:
>     - Added `getDatasetRunItemsByDatasetIdCh` and `getDatasetRunItemsCountByDatasetIdCh` in `dataset-run-items.ts` for Clickhouse queries.
>     - Implemented `getDatasetRunItemsTableInternal` to handle internal query logic with filtering and ordering.
>   - **Schema and Table Definitions**:
>     - Added `dataset_run_items` to `ClickhouseTableNames` in `schema.ts`.
>     - Created `mapDatasetRunItemsTable.ts` for UI column mappings.
>   - **Transformations**:
>     - Added `transformDbDatasetRunItemToAPIDatasetRunItemCh` in `datasets.ts` for transforming Clickhouse data to API format.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for c65de563173ebf96ee8f9372e968da4e94dd3938. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/public-api/types/datasets.ts
```diff
@@ -91,6 +92,21 @@ export const transformDbDatasetRunItemToAPIDatasetRunItemPg = (
 ): z.infer<typeof APIDatasetRunItem> =>
   removeObjectKeys(dbDatasetRunItem, ["projectId"]);
 
+export const transformDbDatasetRunItemToAPIDatasetRunItemCh = (
```
maxdeichmann, line 95: I think for the transformations we should have two steps:
- in the repo: transform from the CH record type to the domain type
- in the API: transform from the Domain to the api result

This way we have these things decoupled and we can re-use the repo functions for different use-cases as well. Otherwise we will start to implement different repo functions for different use cases. This should be as specidic as possible.

web/src/features/public-api/types/datasets.ts
```diff
@@ -11,6 +11,7 @@ import {
   removeObjectKeys,
 } from "@langfuse/shared";
 import { z } from "zod/v4";
+import { type DatasetRunItemDomain } from "../../../../../packages/shared/dist/src/domain/dataset-run-items";
```
maxdeichmann: This import here is broken.

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -1,6 +1,147 @@
...
+          dri.dataset_run_created_at as dataset_run_created_at,
+          dri.dataset_item_input as dataset_item_input,
+          dri.dataset_item_expected_output as dataset_item_expected_output,
+          dri.dataset_item_metadata as dataset_item_metadata,
+          dri.is_deleted as is_deleted,
+          dri.event_ts as event_ts`;
+      default:
+        throw new Error(`Unknown select type: ${opts.select}`);
+    }
+  })();
```
maxdeichmann: why is this a function that we immmediately execute?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -1,6 +1,147 @@
...
+      ...appliedFilter.params,
+    },
+    tags: {
+      ...(opts.tags ?? {}),
+      feature: "datasets",
+      type: "dataset-run-items",
+      projectId,
+      datasetId,
+    },
+    // TODO: do I need to add clickhouseConfigs here?
```
maxdeichmann: pls remove

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -1,6 +1,153 @@
...
+    FROM dataset_run_items dri 
+    WHERE ${appliedFilter.query}
+    -- We pull only the latest version of each dataset run item and assume a 1:1 mapping between dataset run items and dataset items.
+    AND dri.event_ts = (
+      SELECT argMax(event_ts, event_ts) 
+      FROM dataset_run_items dri2 
+      WHERE dri2.project_id = dri.project_id 
+        AND dri2.dataset_item_id = dri.dataset_item_id
+        AND dri2.dataset_run_id = dri.dataset_run_id
+    )
```
marliessophie: I'm hoping there is a more performant way to achieve this: I need to pull only the newest dataset run item for each combination of [datasetId, datasetRunId, datasetItemId] in all queries. How can I achieve this? We will not want to enforce this on a schema level as supporting multiple run items in the same dataset, run combination for the same dataset item is a requirement for repetitions. But currently, it's causing inconstancies.

packages/shared/src/server/repositories/dataset-run-items.ts
marliessophie: makes no sense, will remove

packages/shared/src/server/repositories/dataset-run-items.ts
maxdeichmann: You can do this using: 

```
    ORDER BY event_ts DESC
    LIMIT 1 BY id, project_id
```

maxdeichmann: We do have this a couple of times in our code base. Alternatively, you can use FINAL which should deduplicate all duplicates, right?

maxdeichmann: Uniqueness in CH is eventually guaranteed by the primary key: (project_id, dataset_id, dataset_run_id, id). Hence using final will make the query simpler for now. Given the size of the table this might be good enough.

web/src/features/public-api/server/dataset-run-items.ts
```diff
@@ -0,0 +1,68 @@
...
+    FROM dataset_run_items dri
+    WHERE project_id = {projectId: String}
+    AND dataset_id = {datasetId: String}
+    AND dataset_run_id = {runId: String}
+  `;
+
+  const records = await queryClickhouse<{ count: string }>({
+    query,
+    params: { projectId, datasetId, runId },
+  });
```
maxdeichmann, line 66: Didnt you have a count function in the repository already?

maxdeichmann: The `internal` function in the repo supports count i think.

marliessophie: yes likely copy paste error from previous commit, thanks for spotting this!

PR #8207, opened 2025-07-30
feat(dataset-run-items): reads router

marliessophie:
> [!IMPORTANT]
> Add new dataset run items reading feature with metrics and scores support, updating types, functions, and UI components.
> 
>   - **New Types and Functions**:
>     - Add `DatasetRunsMetricsTableQuery`, `DatasetRunsMetrics`, and `DatasetRunsMetricsRecordType` in `dataset-run-items.ts`.
>     - Add `convertDatasetRunsMetricsRecord()` in `dataset-run-items.ts`.
>     - Add `getDatasetRunsTableInternal()` and `getDatasetRunsTableMetricsCh()` in `dataset-run-items.ts`.
>     - Add `getTraceScoresForDatasetRuns()` in `scores.ts`.
>   - **Dataset Router Enhancements**:
>     - Add `runsByDatasetId` and `runsByDatasetIdMetrics` queries in `dataset-router.ts` to fetch dataset runs and their metrics.
>     - Add `runitemsByRunIdOrItemId` query in `dataset-router.ts` to fetch run items by run ID or item ID.
>   - **UI Updates**:
>     - Update `DatasetRunsTable.tsx` to display new metrics and scores.
>     - Modify `convertToTableRow()` in `DatasetRunsTable.tsx` to handle new data fields.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for a65b4fbf454fc9c8086f2f59aef7d0cc753800c4. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/repositories/scores.ts
```diff
@@ -284,6 +284,53 @@ export const getScoresForDatasetRuns = async <
   return rows.map(convertToScore);
 };
 
+export const getTraceScoresForDatasetRuns = async (
```
marliessophie, line 287: max, would love for you to confirm something: previously, it looked like we used the `getTraceScoresFromTempTable` which fetches all scores that match the DRIs trace id. 

Technically, if a DRI was linked to both a trace and an observation, I would have expected us to fetch only scores where trace and observation id are equal, would you agree? We should implement the legacy way in this PR regardless, but wanted to surface my confusion here and validate against your thinking. Thanks!

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -23,6 +27,59 @@ type DatasetRunItemsTableQuery = {
...
+  dataset_run_description: string;
+  dataset_run_metadata: any;
+  dataset_run_created_at: string;
+  dataset_id: string;
+  count_run_items: number;
+  avg_latency_seconds: number;
+  avg_total_cost: number;
+};
+
+const convertDatasetRunsMetricsRecord = (
```
marliessophie, line 60: Move to converters

web/src/features/datasets/server/dataset-router.ts
```diff
@@ -266,42 +314,150 @@ export const datasetRouter = createTRPCRouter({
...
+            }),
+          ]);
 
-      return {
-        totalRuns,
-        runs,
-      };
+          return {
+            totalRuns,
+            runs,
```
maxdeichmann, line 365: We do not include the traceid, observationid, ri_id etc here anymore?

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -45,6 +102,150 @@ const getProjectDatasetIdDefaultFilter = (
...
+        o.id,
+        o.trace_id,
+        o.project_id,
+        o.start_time,
+        o.end_time,
+        o.total_cost
+      FROM observations o FINAL
+      WHERE o.project_id = {projectId: String}
+        AND o.start_time >= (
+          SELECT min(dri.dataset_run_created_at) - INTERVAL 1 DAY
```
maxdeichmann, line 137: For these intervals we have constants in our code base such as TRACE_TO_OBSERVATIONS_INTERVAL. Please add this as well. This way we have an overview which time cut-offs we guarantee for our users.

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -45,6 +102,150 @@ const getProjectDatasetIdDefaultFilter = (
...
+      FROM observations o FINAL
+      WHERE o.project_id = {projectId: String}
+        AND o.start_time >= (
+          SELECT min(dri.dataset_run_created_at) - INTERVAL 1 DAY 
+          FROM dataset_run_items dri 
+          WHERE dri.project_id = {projectId: String} 
+            AND dri.dataset_id = {datasetId: String}
+        )
+        AND o.start_time <= (
+          SELECT max(dri.dataset_run_created_at) + INTERVAL 1 DAY
```
maxdeichmann, line 143: Please also add this constant. Good judgement here on the upper bound.

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -45,6 +102,150 @@ const getProjectDatasetIdDefaultFilter = (
...
+        AND dri.project_id = of.project_id
+        AND dri.trace_id = of.trace_id
+      WHERE dri.dataset_id = {datasetId: String}
+        AND dri.observation_id IS NOT NULL  -- Only for observation-level dataset run items
+    )
+    SELECT DISTINCT
+      dri.dataset_run_id as dataset_run_id,
+      dri.project_id as project_id,
+      dri.dataset_id as dataset_id,
+      any(dri.dataset_run_name) as dataset_run_name,
```
maxdeichmann, line 181: I would default to anyLast here as any seems random. I know this will basically never happen but i think some "order" is good practice.

maxdeichmann: I just realized: doesnt it make sense to take these from the dataset_run in postgres? This way the user has a more deterministic experience as updates are propagated eventually in clickhohuse via the ingestion pipeline but will be instant in PG.

packages/shared/src/server/repositories/dataset-run-items.ts
```diff
@@ -45,6 +102,150 @@ const getProjectDatasetIdDefaultFilter = (
...
+      any(dri.dataset_run_description) as dataset_run_description,
+      any(dri.dataset_run_metadata) as dataset_run_metadata,
+      any(dri.dataset_run_created_at) as dataset_run_created_at,
+      count(DISTINCT dri.project_id, dri.dataset_id, dri.dataset_run_id, dri.dataset_item_id) as count_run_items,
+      
+      -- Latency metrics (priority: observation > trace)
+      AVG(CASE
+        WHEN dri.observation_id IS NOT NULL AND od.latency_ms IS NOT NULL
+        THEN od.latency_ms / 1000.0
+        ELSE COALESCE(ta.latency_ms / 1000.0, 0)
```
maxdeichmann, line 188: Are you sure we want to default to 0 here? Null could be an alternative here.

packages/shared/src/server/repositories/dataset-run-items.ts
marliessophie: sync: kill metadata

PR #8350, opened 2025-08-06
chore: blob storage export reliability

maxdeichmann:


hcadioli:
hey @maxdeichmann! thank you so much for this.
I have a Blob integration setup with GCS but the scheduled exports don't work.
I always have to trigger this manually.

What's curious is that the scheduled exports run normally in my develop environment that has much less data.

PR #8470, opened 2025-08-12
fix: implement a recursive version of getChannels

deejay1:
## What does this PR do?

SlackService.getChannels only fetched 200 channels, which after filtering might get even less. This doesn't work for large organizations, as there are usually more than 200 active channels, with possibly more archived (which counted into the limit)

Related discussion https://github.com/orgs/langfuse/discussions/8359

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Implements recursive channel fetching in `SlackService` to handle Slack API pagination and retrieve all channels.
> 
>   - **Behavior**:
>     - Implements `getChannelsRecursive()` in `SlackService.ts` to fetch all channels using Slack API's cursor-based pagination.
>     - Updates `getChannels()` to use `getChannelsRecursive()` for complete channel retrieval.
>   - **Error Handling**:
>     - Logs errors in `getChannelsRecursive()` if channel fetching fails.
>     - Throws detailed errors in `getChannelsRecursive()` and `getChannels()` for better debugging.
>   - **Logging**:
>     - Adds debug logging in `getChannels()` to log the number of channels retrieved.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for d9e0a7ea9ede4796fd0dca43581e6a9e67b56934. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/services/SlackService.ts
+++ b/packages/shared/src/server/services/SlackService.ts
@@ -89,11 +89,8 @@ export function parseSlackInstallationMetadata(
 export class SlackService {
   private static instance: SlackService | null = null;
   private readonly installer: InstallProvider;
-  // Total number of records to fetch from Slack
-  private readonly fetchLimit: number;
 
   private constructor() {
-    this.fetchLimit = env.SLACK_FETCH_LIMIT;
     this.installer = new InstallProvider({
       clientId: env.SLACK_CLIENT_ID!,
       clientSecret: env.SLACK_CLIENT_SECRET!,
@@ -307,14 +304,11 @@ export class SlackService {
     cursor?: string,
     fetchedRecords: number = 0,
   ): Promise<SlackChannel[]> {
-    // Limit of records per page
-    const pageLimit = 200;
-
     try {
       const result = await client.conversations.list({
         exclude_archived: true,
         types: "public_channel",
-        limit: pageLimit,
+        limit: 200,
         cursor: cursor,
       });
 
@@ -332,7 +326,10 @@ export class SlackService {
       );
 
       const nextCursor = result.response_metadata?.next_cursor;
-      if (nextCursor && fetchedRecords + channels.length < this.fetchLimit) {
+      if (
+        nextCursor &&
+        fetchedRecords + channels.length < env.SLACK_FETCH_LIMIT
+      ) {
         try {
           const nextPageChannels = await this.getChannelsRecursive(
             client,
--- a/packages/shared/src/server/services/SlackService.ts
+++ b/packages/shared/src/server/services/SlackService.ts
@@ -89,8 +89,11 @@ export function parseSlackInstallationMetadata(
 export class SlackService {
   private static instance: SlackService | null = null;
   private readonly installer: InstallProvider;
+  // Total number of records to fetch from Slack
+  private readonly fetchLimit: number;
 
   private constructor() {
+    this.fetchLimit = env.SLACK_FETCH_LIMIT;
     this.installer = new InstallProvider({
       clientId: env.SLACK_CLIENT_ID!,
       clientSecret: env.SLACK_CLIENT_SECRET!,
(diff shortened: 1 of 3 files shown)


bgalek (approved):

adowsky (approved):

deejay1:
@Steffen911 could you take a look? I can make a non recursive version if needed of course.

packages/shared/src/server/services/SlackService.ts
```diff
@@ -318,6 +323,30 @@ export class SlackService {
...
+      }
+
+      return channels;
+    } catch (error) {
+      logger.error("Failed to fetch channels recursively", { error, cursor });
+      throw new Error(
+        `Failed to fetch channels: ${error instanceof Error ? error.message : "Unknown error"}`,
+      );
+    }
+  }
```
Steffen911: Two ideas I'd like to add here:
1. You could add a try/catch within the `if (nextCursor)` block that logs an error, but returns the channels result. That way, we optimize for returning something over nothing if something deep-down in the stack errors. It will also handle errors where the maximum stack size is exceeded gracefully.
2. I would prefer to still have some upper bound on the channels we return. Having an env variable that defaults to a value like 1000 seems reasonable here. If that is not feasible, we'll have to look into searches via the API instead as I believe the result won't be useful after a certain threshold is reached.

I have strong feelings about the first point and would love your thoughts on the second.

Thank you for contributing this!

deejay1: I think both things should be done now

adowsky (approved):

web/src/env.mjs
```diff
@@ -273,6 +273,11 @@ export const env = createEnv({
     SLACK_CLIENT_ID: z.string().optional(),
     SLACK_CLIENT_SECRET: z.string().optional(),
     SLACK_STATE_SECRET: z.string().optional(),
+    SLACK_FETCH_LIMIT: z
+      .number({ description: "How many records should be fetched from Slack, before we give up" })
+      .positive()
+      .optional()
+      .default(1_000),
```
Steffen911: It seems like you never consume this within the code.

deejay1: 🤦 I actually forgot about it!

deejay1: Fixed

packages/shared/src/server/services/SlackService.ts
```diff
@@ -88,9 +88,12 @@ export function parseSlackInstallationMetadata(
  */
 export class SlackService {
   private static instance: SlackService | null = null;
-  private installer: InstallProvider;
+  private readonly installer: InstallProvider;
+  // Total number of records to fetch from Slack
+  private readonly fetchLimit: number;
```
Steffen911: Why do you reassign this to a private variable instead of directly consuming the env variable? To me this feels very redundant.

web/src/env.mjs
```diff
@@ -273,6 +273,11 @@ export const env = createEnv({
     SLACK_CLIENT_ID: z.string().optional(),
     SLACK_CLIENT_SECRET: z.string().optional(),
     SLACK_STATE_SECRET: z.string().optional(),
+    SLACK_FETCH_LIMIT: z
```
Steffen911: I believe we can remove the modifications to this file as it's not being consumed.

packages/shared/src/server/services/SlackService.ts
deejay1: to much Java coding ;)

web/src/env.mjs
deejay1: looks awfully close to the `env.ts` code, but as I'm not that fluent (candidate for understatement of the year ;) in this code base, I'll remove it ;)

Steffen911 (approved):

bgalek (approved):

PR #8551, opened 2025-08-15
perf: add caching for clickhouse dataset run items in createEvalJob

Steffen911:
> [!IMPORTANT]
> Adds caching for dataset run items in `createEvalJobs` to optimize performance, with updates to related functions and tests.
> 
>   - **Caching**:
>     - Adds caching for dataset run items in `createEvalJobs` in `evalService.ts`.
>     - Uses `getDatasetItemIdsByTraceIdCh` to fetch dataset item IDs and cache them for evaluation optimization.
>   - **Behavior**:
>     - Updates `getDatasetItemIdsByTraceIdCh` to return dataset ID along with item ID in `dataset-run-items.ts`.
>     - Modifies `getTraceById` in `traces.ts` to use optimized aggregation functions.
>   - **Tests**:
>     - Adds tests in `evalService.test.ts` and `batchAction.test.ts` to verify caching behavior and ensure correct job creation.
>   - **Misc**:
>     - Adds `mapDatasetRunItemFilterColumn` to `mapDatasetRunItemsTable.ts` for in-memory filtering.
>     - Updates `.env.dev-redis-cluster.example` with new environment variables for dataset run items caching.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b61fe6052656b5fa6863d332e7ab4e5dfbacdcf7. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Steffen911:
@marliessophie Could you take a look at the failing web async tests? I believe they depend on some logic that hasn't transitioned completely to the ClickHouse based dataset run items.

worker/src/features/evaluation/evalService.ts
```diff
@@ -240,6 +241,41 @@ export const createEvalJobs = async ({
     }
   }
 
+  const datasetConfigs = configs.filter((c) => c.target_object === "dataset");
+  let cachedDatasetItemIds: { id: string; datasetId: string }[] | null = null;
+  if (datasetConfigs.length > 1) {
```
Steffen911, line 249: TODO: We should attempt to parallelize the two cache calls. This will increase throughput on the container.

Steffen911:
Note: I've manually applied the following index to improve the lookups:
```
ALTER TABLE dataset_run_items_rmt ADD INDEX IF NOT EXISTS idx_trace_id trace_id TYPE bloom_filter(0.001) GRANULARITY 1;
ALTER TABLE dataset_run_items_rmt MATERIALIZE INDEX IF EXISTS idx_trace_id;
```
This should improve things for now, but we need to move it to migrations.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -738,27 +738,28 @@ export const getTraceById = async ({
       const query = `
         SELECT
           id,
-          name as name,
-          user_id as user_id,
-          metadata as metadata,
-          release as release,
-          version as version,
+          anyLast(name) as name,
```
Steffen911, line 741: Partially unrelated patch, but necessary to ensure that reads don't use stale information.

Steffen911: Impact of the change:
image

image

🤯

worker/src/features/evaluation/evalService.ts
maxdeichmann: Parllelization also increases load on Clickhouse which then slows down queries. Not sure how much parallelization will be beneficial here.

Steffen911: Yes, I'll start to merge this without to keep things simple. Will add a comment that this could be useful.

Steffen911: Added note in https://github.com/langfuse/langfuse/pull/8551/commits/90affa95c6a564a7fb2ea53fbbd38fd411561b16.

PR #8580, opened 2025-08-18
feat(datasets): allow copying items within source dataset

leoweigand:
- allow duplicating items (target dataset can be  source now)
- streamline ux copy a bit

https://github.com/user-attachments/assets/1e7eb0a4-8a63-4ed1-8e61-43691bb03a4c

----

> [!IMPORTANT]
> Allow copying items within the same dataset and improve UX in `NewDatasetItemForm` and `NewDatasetItemFromExistingObject`.
> 
>   - **Behavior**:
>     - Allow copying items within the same dataset by setting `currentDatasetId` in `NewDatasetItemForm` and `NewDatasetItemFromExistingObject`.
>     - Update button text in `NewDatasetItemForm` to reflect the number of selected datasets.
>   - **UI/UX**:
>     - Rename "Datasets" to "Target datasets" in `NewDatasetItemForm`.
>     - Change dialog title to "Add item to datasets" in `NewDatasetItemFromExistingObject`.
>   - **Code**:
>     - Replace `blockedDatasetIds` with `currentDatasetId` in `NewDatasetItemForm` and `NewDatasetItemFromExistingObject`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 28413082d5523714131f9ceaaf104e6730dac760. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/datasets/components/NewDatasetItemForm.tsx
```diff
@@ -336,9 +336,12 @@ export const NewDatasetItemForm = (props: {
               type="submit"
               loading={createManyDatasetItemsMutation.isLoading}
               className="w-full"
-              disabled={form.watch("datasetIds").length === 0}
+              disabled={selectedDatasetCount === 0}
             >
-              Add to dataset{form.watch("datasetIds").length > 1 ? "s" : ""}
+              Copy item
```
marliessophie: Would you say this is "copy" or "add"? To me "copy" would be be to copy over without editing, but trust your judgement here on UX :)

marliessophie (commented):
Left small comment on UX :) Also would suggest adding a quick success toast and closing modal when completed. 

Did not check backend code, but assume this propagates correctly? Eg we need to make sure not to accidentally upsert when adding to the same dataset.

web/src/features/datasets/components/NewDatasetItemForm.tsx
leoweigand: This was a bit of back and forth for me as well but "Add to.." made even less sense to me since you can create a copy within the same dataset where the item you're copying already exists.

edit: just realised we use this dialog for other add to datsets buttons, so I reverted some changes

PR #8615, opened 2025-08-20
perf: split batch to be processed on string length errors in CH writer

Steffen911:
> [!IMPORTANT]
> `ClickhouseWriter` now splits batches to handle "invalid string length" errors, with tests added for this behavior.
> 
>   - **Behavior**:
>     - `ClickhouseWriter` now handles "invalid string length" errors by splitting the batch and retrying.
>     - If batch size is 1, falls back to truncation to prevent infinite loops.
>   - **Functions**:
>     - Adds `isStringLengthError()` to identify string length errors in `index.ts`.
>     - Adds `handleStringLengthError()` to split batches on string length errors in `index.ts`.
>   - **Tests**:
>     - Adds test for batch splitting on string length errors in `ClickhouseWriter.unit.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 39caf4268705dea4f85a9937086a7d2cbcc4a9a8. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


worker/src/services/ClickhouseWriter/index.ts
```diff
@@ -134,14 +134,57 @@ export class ClickhouseWriter {
...
+  private isStringLengthError(error: unknown): boolean {
+    if (!error || typeof error !== "object") return false;
+
+    const errorMessage = (error as Error).message?.toLowerCase() || "";
+
+    // Node.js string size errors
+    return errorMessage.includes("invalid string length");
+  }
+
+  private handleStringLengthError<T extends TableName>(
```
maxdeichmann, line 159: I would add a description here: we cut the array to write in half and write half of the original array

worker/src/services/ClickhouseWriter/index.ts
```diff
@@ -269,6 +313,37 @@ export class ClickhouseWriter {
...
+              const { retryItems, requeueItems } = this.handleStringLengthError(
+                tableName,
+                queueItems,
+              );
+
+              // Update records to write with only the retry items
+              recordsToWrite = retryItems.map((item) => item.data);
+              queueItems = retryItems;
+
+              // Prepend requeue items to the front of the queue to maintain order
```
maxdeichmann: This is not entirely correct: according to my understanding, the writer can have multiple "write" processes active at the same time. Therefore we cannot guarantee any order here. This is best effort but the comment is misleading.

worker/src/services/ClickhouseWriter/index.ts
Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/8615/commits/7a5d8812bc2821711e1ef1fb92718bd6ed6a727d

worker/src/services/ClickhouseWriter/index.ts
Steffen911: Addressed in https://github.com/langfuse/langfuse/pull/8615/commits/7a5d8812bc2821711e1ef1fb92718bd6ed6a727d

PR #8617, opened 2025-08-20
fix(posthog-integration): skip processing of person profiles for anon…

marcklingen:
> [!IMPORTANT]
> Improve PostHog integration by handling anonymous user events with `null` IDs and skipping person profile processing for better performance.
> 
>   - **Behavior**:
>     - Change default `langfuse_user_id` to `null` for anonymous users in `getGenerationsForPostHog`, `getScoresForPostHog`, and `getTracesForPostHog`.
>     - Add `$process_person_profile: false` for anonymous events in `getGenerationsForPostHog`, `getScoresForPostHog`, and `getTracesForPostHog`.
>     - Use random UUID for anonymous users in `processPostHogTraces`, `processPostHogGenerations`, and `processPostHogScores` in `handlePostHogIntegrationProjectJob.ts`.
>   - **Files**:
>     - `observations.ts`, `scores.ts`, `traces.ts`: Modify user ID handling and add anonymous event flag.
>     - `handlePostHogIntegrationProjectJob.ts`: Update event processing logic for PostHog integration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for c2852a4252bdee7289fb3474dd9b107edf9c2bcc. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


nickbest-ph (approved):
Hi @marcklingen, this is Nick from the Ingestion Team at Posthog.

Skipping person processing will substantially increase the rate that we are able to ingest events for anonymous ids, so this approach looks right to me!

packages/shared/src/server/repositories/traces.ts
```diff
@@ -1875,7 +1876,7 @@ export const getTracesForPostHog = async function* (
       langfuse_count_observations: record.observation_count,
       langfuse_session_id: record.session_id,
       langfuse_project_id: projectId,
-      langfuse_user_id: record.user_id || "langfuse_unknown_user",
+      langfuse_user_id: record.user_id || POSTHOG_ANON_USER_ID,
```
nickbest-ph: Posthog's ingestion is partitioned by distinct id. If you don't care about linking all these events to the same person profile/distinct id, then making the anonymous distinct id unique will further improve ingestion throughput by distributing these events across all ingestion partitions.

nickbest-ph: With the important caveat that the above is only helpful if person processing is disabled for the events

marcklingen: thank you, very helpful

marcklingen: addressed in https://github.com/langfuse/langfuse/pull/8617/commits/0bccd41b9ec76c8915237e347403462e568a8be2

PR #8680, opened 2025-08-22
feat(datasets): full text search for dataset items

marliessophie:
> [!IMPORTANT]
> Adds full-text search for dataset items by ID, content, and metadata with case-insensitive and special character handling, including pagination and extensive testing.
> 
>   - **Behavior**:
>     - Adds full-text search for dataset items by ID, content, and metadata in `dataset-service.servertest.ts`.
>     - Supports case-insensitive searches and special character handling.
>     - Implements pagination for search results.
>   - **Components**:
>     - Integrates search functionality in `DatasetItemsTable.tsx` with `useFullTextSearch` hook.
>     - Configures search options and debouncing for search queries.
>   - **Server**:
>     - Updates `dataset-router.ts` to handle search queries and types.
>     - Modifies `fetchDatasetItems` in `service.ts` to include search filters and query generation.
>   - **Tests**:
>     - Extensive tests added in `dataset-service.servertest.ts` for search functionality, including edge cases and performance.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 8736b6965543089143d9d8ad992783c630d812df. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

Further, this PR simplifies the logic in `itemsByDatasetId` where we: 
- historically: used to verify the existence of source trace id and source observation id against our CH database 
- now: we always show links. If user clicks on a trace / observation link, they will not get a "trace has been deleted or is still being processed" link but will end up in a 404

CleanShot 2025-08-22 at 12 54 45@2x


web/src/features/datasets/server/service.ts
```diff
@@ -470,12 +473,47 @@ const generateDatasetItemQuery = ({
   WHERE di.project_id = ${projectId}
   AND di.dataset_id = ${datasetId}
   ${filterCondition}
+  ${searchFilter}
   ${orderByClause}
   LIMIT ${limit} OFFSET ${page * limit}
  `;
 };
 
+const buildDatasetItemSearchFilter = (
```
maxdeichmann, line 482: We keep on re-defining this one. Do you think it might make sense to abstract this or is this too different for all the Postgres FTS use cases we have?

marliessophie: Makes sense, thank you. Will scope out of this pull request.

PR #8768, opened 2025-08-27
perf: remove traces env filter form observations table

maxdeichmann:


dahnny012:
Is there a reason this was removed?

maxdeichmann:
> Is there a reason this was removed?

Yes, performance improvement. Did this break anything for you?

PR #8885, opened 2025-09-03
chore(peek): remove polling logic and dependency on table state from peek implementation  

marliessophie:
> [!IMPORTANT]
> Refactor peek view implementation by removing polling logic, introducing `usePeekNavigation` for navigation, and updating components to use the new system.
> 
>   - **Behavior**:
>     - Removes polling logic and dependency on table state from peek implementation in `data-table.tsx` and `peek.tsx`.
>     - Introduces `usePeekNavigation` hook in `usePeekNavigation.ts` for managing peek view navigation.
>     - Refactors `DataTablePeekViewProps` to include new navigation and event handling methods.
>   - **Hooks**:
>     - Adds `usePeekNavigation` in `usePeekNavigation.ts` for managing peek view navigation.
>     - Adds `usePeekCompareDetail` in `usePeekCompareDetail.ts` for handling trace comparison details.
>     - Adds `usePeekRunsCompareData` in `usePeekRunsCompareData.ts` for fetching run comparison data.
>     - Removes deprecated hooks: `usePeekView`, `useDatasetComparePeekState`, `useObservationPeekState`, `useTracePeekState`, and others.
>   - **Components**:
>     - Updates `DataTable` in `data-table.tsx` to use new peek view props and navigation.
>     - Updates `TablePeekView` in `peek.tsx` to handle new navigation and event logic.
>     - Refactors components like `PeekDatasetCompareDetail`, `PeekViewEvaluatorConfigDetail`, `PeekViewEvaluatorTemplateDetail`, `PeekViewObservationDetail`, and `PeekViewTraceDetail` to use new navigation system.
>   - **Misc**:
>     - Updates `dataset-router.ts` to support new query parameters for dataset run items.
>     - Adds `urlSearchParamsToQuery` utility in `navigation.ts` for converting URLSearchParams to query objects.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 25e2d474a4fa7192d3892f782440c491740ea879. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

edit by Leo: fixes LFE-6243 as well


web/src/components/table/peek/hooks/usePeekExpandObservationDetail.ts
```diff
@@ -0,0 +1,25 @@
...
+  const { projectId, peek } = router.query;
+
+  const expandPeek = (openInNewTab: boolean) => {
+    const url = new URL(window.location.href);
+    const params = new URLSearchParams(url.search);
+    const traceId = params.get("traceId");
+    const timestamp = params.get("timestamp");
+    const display = params.get("display") ?? "details";
+
+    const pathname = `/project/${projectId}/traces/${encodeURIComponent(traceId as string)}?timestamp=${timestamp}&display=${display}&observation=${peek as string}`;
```
leoweigand: Shouldn't we preserve other query params here that may be set, like filters?

web/src/components/table/data-table.tsx
```diff
@@ -390,13 +379,7 @@ export function DataTable<TData extends object, TValue>({
         </div>
         <div className="grow"></div>
       </div>
-      {peekView && (
-        <TablePeekView
-          peekView={peekView}
-          row={peekRow}
-          selectedRowId={peekViewId}
```
leoweigand, line 443: good stuff

web/src/components/table/use-cases/traces.tsx
```diff
@@ -1018,8 +1022,8 @@ export default function TracesTable({
       peekEventOptions: {
         ignoredSelectors: ['[role="checkbox"]', '[aria-label="bookmark"]'],
       },
-      onOpenChange: setPeekView,
-      onExpand: expandPeek,
+      // TODO: onExpand is missing
```
leoweigand: this seems important, will it be added back before you merge it?

web/src/components/table/peek/hooks/usePeekNavigation.ts
leoweigand, line 1: I find the function signature for this hook a bit surprising considering what it does, would have expected something more like this:

```ts
interface UsePeekViewProps {
  itemQueryKey: string; // eg "observation",
  clearParamsOnClose: string[];
}

interface UsePeekViewReturn {
  openPeekView: (itemId: string, additionalParams?: Record<string,string>) => void;
  closePeekView: () => void;
  isOpen: boolean;              // for convenience, may be unneeded
  currentItemId: string | null; // for convenience, may be unneeded
}

function usePeekView(props: UsePeekViewProps): UsePeekViewReturn {...}
```

- `onOpenChange` implied that the function is an event handler that is called when the peekView open state changes but really, it's the function _to open / close_ the view.
- `itemQueryKey`: i would not set multiple params to the item id but just make customisable which param the hook uses to determine the state, so that e.g. `observation` can be used instead of `peek`
- functionality to set / remove additional params makes sense to me but feels a bit simpler this way, not sure if easy to change
- `getNavigationPath`–I didn't see this function being used anywhere but maybe I missed something 🤔

web/src/components/table/data-table.tsx
marliessophie: I laughed at this. hard.

web/src/components/table/use-cases/traces.tsx
marliessophie: yes! sorry, my comment here should have been clearer

web/src/components/table/peek/hooks/usePeekNavigation.ts
marliessophie: Appreciate this! Reworked most of this. Put a pin in itemQueryKey, as this is a further improvement we might be able to do.

PR #9073, opened 2025-09-11
fix(peek): safe parse runs url parameter in compare data hook

marliessophie:
> [!IMPORTANT]
> Introduces `safeParseUrlParamToArray` in `usePeekRunsCompareData.ts` to ensure `runs` parameter is always an array, handling string inputs.
> 
>   - **Behavior**:
>     - Introduces `safeParseUrlParamToArray` function in `usePeekRunsCompareData.ts` to ensure `runs` is always an array.
>     - Handles cases where `runs` URL parameter is a string by converting it to an array.
>   - **Hook**:
>     - Updates `usePeekRunsCompareData` to use `safeParseUrlParamToArray` for `runs` parameter.
>     - Ensures `datasetRunIds` in `runItems` query is correctly formatted as an array.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 205ab742863addb9946dffdc58e71949f876cfcf. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/components/table/peek/hooks/usePeekRunsCompareData.ts
+++ b/web/src/components/table/peek/hooks/usePeekRunsCompareData.ts
@@ -6,7 +6,15 @@ type UsePeekRunsCompareDataProps = {
   datasetItemId?: string;
   traceId?: string;
   timestamp?: Date;
-  runs?: string[];
+  runs?: string[] | string;
+};
+
+// Ensure runs is always an array - handle case where URL param comes as string
+const safeParseUrlParamToArray = (
+  runs?: string[] | string,
+): string[] | undefined => {
+  if (!runs) return undefined;
+  return Array.isArray(runs) ? runs : [runs];
 };
 
 export const usePeekRunsCompareData = ({
@@ -17,6 +25,8 @@ export const usePeekRunsCompareData = ({
   datasetItemId,
   runs,
 }: UsePeekRunsCompareDataProps) => {
+  const parsedRuns = safeParseUrlParamToArray(runs);
+
   const trace = api.traces.byIdWithObservationsAndScores.useQuery(
     {
       traceId: traceId as string,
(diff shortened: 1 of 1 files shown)


web/src/components/table/peek/hooks/usePeekRunsCompareData.ts
```diff
@@ -6,7 +6,15 @@ type UsePeekRunsCompareDataProps = {
...
-  runs?: string[];
+  runs?: string[] | string;
+};
+
+// Ensure runs is always an array - handle case where URL param comes as string
+const safeParseUrlParamToArray = (
+  runs?: string[] | string,
+): string[] | undefined => {
+  if (!runs) return undefined;
+  return Array.isArray(runs) ? runs : [runs];
```
maxdeichmann, line 17: @marliessophie i just saw this on Linear.  Wouldnt it be cleaner to add this logic to default this into an array here: https://github.com/langfuse/langfuse/blob/26045d089c866ec800664e5a7dd396b863aa0138/web/src/components/table/peek/peek-dataset-compare-detail.tsx#L73?

This way, the usePeekRunsCompareData keeps on getting a string array only instead of strings. Runs (plural) as string is confusing IMO. 

Also we default the runs in PeekDatasetCompareDetail to an array or undefined. This could also break downstream if this may also be a string.

Alternatively, you could also make sure that whoever creates an compare URL, always creates an array and never a string.

marliessophie: Thanks for the feedback, will open a second PR to address. Good to stop the bleeding first.

PR #9082, opened 2025-09-11
fix(otel): Make sure we do not process empty otel events

sumerman:
On some codepaths `processEventBatch` can be called on an empty `input`. e.g.  when `otelIngestionQueueProcessor` processes spans without any trace metadata. This behavior is generally benign because `processEventBatch` will just call a bunch of flat-maps and reduce functions on an empty array. Yet it is somewhat confusing from the o11y perspective: calls to `processEventBatch` with `batch_size = 0` is nothing but noise. This PR adds an early exit which should get rid of the noise and spare a couple of cycles along the way :) 

----

> [!IMPORTANT]
> Adds early exit in `processEventBatch` for empty inputs to reduce noise and adds a test for this behavior.
> 
>   - **Behavior**:
>     - Adds early exit in `processEventBatch` in `processEventBatch.ts` for empty `input` arrays, returning empty `successes` and `errors`.
>   - **Tests**:
>     - Adds test in `processEventBatch.test.ts` to verify early exit behavior for empty `input` arrays, ensuring no `UnauthorizedError` is thrown and results are empty.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 2db7b3d6e93f18cbbfbd8bdd76d1a7708e9b522c. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/worker/src/__tests__/processEventBatch.empty.test.ts
+++ b/worker/src/__tests__/processEventBatch.empty.test.ts
@@ -1,0 +1,25 @@
+import { describe, it, expect, assert } from "vitest";
+import { processEventBatch } from "@langfuse/shared/src/server";
+
+describe("processEventBatch - empty batch", () => {
+  it("returns early on empty input", async () => {
+    // Auth check with missing projectId will cause an exeption unless
+    // there is an early return in processEventBatch
+    const authCheck = {
+      validKey: true as const,
+      scope: {
+        projectId: null,
+        accessLevel: "project" as const,
+      },
+    };
+
+    assert.doesNotThrow(
+      async () => await processEventBatch([], authCheck, {}),
+      "UnauthorizedError",
+    );
+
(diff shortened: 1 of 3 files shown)


worker/src/__tests__/processEventBatch.empty.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, it, expect, assert } from "vitest";
```
Steffen911: I would rename this file to processEventBatch.test.ts and use it for potential extensions in the future. Otherwise, I believe it will become too granular at scale.

Steffen911 (approved):
Small nitpick on naming. Otherwise LGTM

PR #9174, opened 2025-09-17
fix(llm): OpenAI llm completion reasoning model checks and token adjustments

marliessophie:
> [!IMPORTANT]
> Adjusts token handling in `fetchLLMCompletion()` based on OpenAI reasoning model checks and adds model classification in `types.ts`.
> 
>   - **Behavior**:
>     - Adjusts token handling in `fetchLLMCompletion()` in `fetchLLMCompletion.ts` to use `maxCompletionTokens` for reasoning models and `maxTokens` otherwise.
>     - Introduces `isOpenAIReasoningModel()` in `types.ts` to determine if a model is a reasoning model.
>   - **Models**:
>     - Adds `openAIModelToReasoning` map in `types.ts` to classify models as reasoning or non-reasoning.
>   - **Misc**:
>     - Adds `OpenAIModel` type in `types.ts` for model type safety.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 4dfb52bd6b9a4a34ddd442f1e3efbd9cea350264. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

Relevant notes from OpenAI: https://community.openai.com/t/why-was-max-tokens-changed-to-max-completion-tokens/938077/4

### DIFF
--- a/packages/shared/src/server/llm/types.ts
+++ b/packages/shared/src/server/llm/types.ts
@@ -333,6 +333,57 @@ export const openAIModels = [
   "gpt-3.5-turbo",
 ] as const;
 
+type OpenAIReasoningMap = Record<OpenAIModel, boolean>;
+export const openAIModelToReasoning: OpenAIReasoningMap = {
+  // reasoning models
+  "gpt-5": true,
+  "gpt-5-2025-08-07": true,
+  "gpt-5-mini": true,
+  "gpt-5-mini-2025-08-07": true,
+  "gpt-5-nano": true,
+  "gpt-5-nano-2025-08-07": true,
+  o3: true,
+  "o3-2025-04-16": true,
+  "o4-mini": true,
+  "o4-mini-2025-04-16": true,
+  "o3-mini": true,
+  "o3-mini-2025-01-31": true,
+  "o1-preview": true,
+  "o1-preview-2024-09-12": true,
+  "o1-mini": true,
+  "o1-mini-2024-09-12": true,
+  // non-reasoning models
+  "gpt-4.5-preview": false,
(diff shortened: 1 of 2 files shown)


marliessophie:
@hassiebp as this has historically been your domain, please skim over the code to make sure I am not missing any relevant context. Thanks.

packages/shared/src/server/llm/types.ts
```diff
@@ -333,6 +333,57 @@ export const openAIModels = [
   "gpt-3.5-turbo",
 ] as const;
 
+type OpenAIReasoningMap = Record<OpenAIModel, boolean>;
+export const openAIModelToReasoning: OpenAIReasoningMap = {
```
hassiebp, line 337: I think we should simplify this to a simple array of known reasoning models and the check should be `reasoningModels.includes(model)`. By that we only need to keep in mind to add reasoning models here and not all models incl. non-reasoning models

hassiebp:
Looks good to me! Only suggested change is to simplify the reasoning model map to a simple list, makes the list clearer and also easier to maintain

packages/shared/src/server/llm/types.ts
marliessophie: The reason why I prefer the current approach is that anyone who adds to the OpenAIModel array will be forced to create an entry in the openAIModelToReasoning map by typescript. There is much to remember when adding new models as it is, I would prefer to keep as is to not rely on the developer to remember adding another entry should it be a reasoning model.

PR #9191, opened 2025-09-17
fix: improve error message for CH memory issues on public API

sumerman:
- This PR attempts to catch and wrap errors originating from ClickHouse in order provide guidance to users.
- Error codes between public api and UI are different. UI has handling for a non-standard `524` that suits this case better. While in the public API I chose to stick with a more common HTTP error code.
- I added a suite that exercises overall `withMiddlewares` error handling behaviour.

----

> [!IMPORTANT]
> Improves ClickHouse memory error handling by introducing `ClickHouseResourceError` and updating error handling logic in `clickhouse.ts`, `withMiddlewares.ts`, and `trpc.ts`, with corresponding tests added.
> 
>   - **Error Handling**:
>     - Introduces `ClickHouseResourceError` class in `clickhouse.ts` to handle memory, overcommit, and timeout errors.
>     - Updates `queryClickhouse`, `queryClickhouseStream`, and `commandClickhouse` in `clickhouse.ts` to wrap errors using `ClickHouseResourceError.wrapIfResourceError()`.
>     - Modifies `withMiddlewares.ts` to handle `ClickHouseResourceError` by returning a 524 status code with advice message.
>     - Updates `trpc.ts` middleware to handle `ClickHouseResourceError` by setting a `TIMEOUT` error code with advice message.
>   - **Testing**:
>     - Adds `clickhouse-resource-errors.servertest.ts` to test error transformation for ClickHouse errors.
>     - Adds `withMiddlewares.servertest.ts` to test error handling in `withMiddlewares` function.
>   - **Misc**:
>     - Adds advice message for ClickHouse errors in `withMiddlewares.ts` and `trpc.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b39813ba45bd7a1ee74186792c47aadfa5defe34. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/repositories/clickhouse.ts
+++ b/packages/shared/src/server/repositories/clickhouse.ts
@@ -18,7 +18,6 @@ import {
 
 /**
  * Custom error class for ClickHouse resource-related errors
- * Provides actionable advice for users
  */
 // Error type configuration map
 const ERROR_TYPE_CONFIG: Record<
@@ -38,22 +37,21 @@ const ERROR_TYPE_CONFIG: Record<
   },
 };
 
-const ERROR_ADVICE_MESSAGE = [
-  "Database resource limit exceeded. Please use more specific filters or a shorter time range. We are continuously improving our API performance.",
-  "See https://langfuse.com/docs/api-and-data-platform/features/public-api for more details.",
-].join("\n");
-
 type ErrorType = keyof typeof ERROR_TYPE_CONFIG;
 
 export class ClickHouseResourceError extends Error {
+  static ERROR_ADVICE_MESSAGE = [
+    "Database resource limit exceeded.",
+    "Please use more specific filters or a shorter time range.",
+    "We are continuously improving our API performance.",
+  ].join(" ");
+
   public readonly errorType: ErrorType;
-  public readonly displayMessage: string;
 
   constructor(errType: ErrorType, originalError: Error) {
     super(originalError.message, { cause: originalError });
     this.name = "ClickHouseResourceError";
     this.errorType = errType;
-    this.displayMessage = ERROR_ADVICE_MESSAGE;
     // Preserve the original stack trace if available
     if (originalError.stack) {
       this.stack = originalError.stack;
--- a/packages/shared/src/server/repositories/clickhouse.ts
+++ b/packages/shared/src/server/repositories/clickhouse.ts
@@ -16,6 +16,72 @@ import {
   StorageServiceFactory,
 } from "../services/StorageService";
 
+/**
+ * Custom error class for ClickHouse resource-related errors
+ * Provides actionable advice for users
+ */
+// Error type configuration map
+const ERROR_TYPE_CONFIG: Record<
+  "MEMORY_LIMIT" | "OVERCOMMIT" | "TIMEOUT",
+  {
+    discriminators: string[];
+  }
+> = {
+  MEMORY_LIMIT: {
+    discriminators: ["memory limit exceeded"],
+  },
+  OVERCOMMIT: {
+    discriminators: ["OvercommitTracker"],
+  },
+  TIMEOUT: {
+    discriminators: ["Timeout", "timeout", "timed out"],
+  },
+};
+
+const ERROR_ADVICE_MESSAGE = [
+  "Database resource limit exceeded. Please use more specific filters or a shorter time range. We are continuously improving our API performance.",
+  "See https://langfuse.com/docs/api-and-data-platform/features/public-api for more details.",
+].join("\n");
+
+type ErrorType = keyof typeof ERROR_TYPE_CONFIG;
+
+export class ClickHouseResourceError extends Error {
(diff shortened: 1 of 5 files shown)


packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -16,6 +16,72 @@ import {
...
+  OVERCOMMIT: {
+    discriminators: ["OvercommitTracker"],
+  },
+  TIMEOUT: {
+    discriminators: ["Timeout", "timeout", "timed out"],
+  },
+};
+
+const ERROR_ADVICE_MESSAGE = [
+  "Database resource limit exceeded. Please use more specific filters or a shorter time range.",
```
maxdeichmann: "Database resource limit exceeded. Please use more specific filters or a shorter time range."
```suggestion
  "Database resource limit exceeded. Please use more specific filters or a shorter time range. We are continuously improving our API performance.",
```

web/src/features/public-api/server/withMiddlewares.ts
```diff
@@ -73,6 +74,22 @@ export function withMiddlewares(handlers: Handlers) {
...
+        if (error instanceof ClickHouseResourceError) {
+          const resourceError = error as ClickHouseResourceError;
+
+          logger.error("ClickHouse resource limit exceeded", {
+            errorType: resourceError.errorType,
+            originalMessage: error.message,
+            displayMessage: resourceError.displayMessage,
+          });
+
+          return res.status(400).json({
```
maxdeichmann: I would still keep this as 5xx as this is our responsibility to fix. Do you agree or would you keep the 400?

sumerman: How do you feel about exposing non-standard `524` in the public API? That would be my preferred option.

maxdeichmann: Good for me.

maxdeichmann:
One more thought: we should show the public API docs only for errors in the public API or drop the reference to the docs if easier.

sumerman:
> One more thought: we should show the public API docs only for errors in the public API or drop the reference to the docs if easier.

From what I understand so far, tRPC doesn't go through `withMiddlewares`. What else is there other than public API and tRPC?

packages/shared/src/server/repositories/clickhouse.ts
```diff
@@ -198,14 +261,20 @@ export async function* queryClickhouseStream<T>(opts: {
           }
         }
         return res;
-      },
-    );
+      })
+      .catch((error) => {
+        // Transform resource errors to provide actionable advice
+        throw ClickHouseResourceError.wrapIfResourceError(error as Error);
+      });
```
Steffen911, line 268: Slight addition while we're touching the error handling. It seems that ClickHouse (inconsistently) returns errors as records in the result instead of raising them: https://github.com/ClickHouse/clickhouse-js/issues/332, e.g.
```
[
  { e: 0, s: '}' },
  {
    exception: "Code: 395. DB::Exception: There was an error in the stream!: while executing 'FUNCTION throwIf(equals(__table1.number, 2_UInt8) :: 3, 'There was an error in the stream!'_String :: 2) -> throwIf(equals(__table1.number, 2_UInt8), 'There was an error in the stream!'_String) UInt8 : 0'. (FUNCTION_THROW_IF_VALUE_IS_NON_ZERO) (version 24.8.3.59 (official build))"
  }
]
```
We've got a recommendation from their support to scan through the rows that were produced whether they are an exception (https://github.com/ClickHouse/clickhouse-js/blob/main/examples/node/select_json_each_row_with_progress.ts#L91). Do you think we could inject that here in the clickhouse specific error handler and also raise something if it's part of the regular result set?

If this is too much scope-creep for this purpose, we can also move it into a separate change.

Steffen911:
> From what I understand so far, tRPC doesn't go through withMiddlewares. What else is there other than public API and tRPC?

I think this would cover sufficient cases. After a week of this being in place we could just scan through our spans and see if there are any that still raised this up to the user.

maxdeichmann:
We resolved this in person.

packages/shared/src/server/repositories/clickhouse.ts
sumerman: Happy to take that, but IMO this should go into its own PR

sumerman: Interesting, https://github.com/langfuse/langfuse/pull/9191/files#diff-57a6f92be65e66e4005527d26c2337d4404114408c5380c0ceaebe1c571be273R87 should have caught this 🤔

sumerman: https://github.com/langfuse/langfuse/pull/9238

PR #9359, opened 2025-09-26
feat(auth): add support for Keycloak scope and ID token configuration

### LINKED ISSUE #9186: bug: OIDC Keycloak: No way to specify custom scopes; custom OIDC provider not compatible with DB schema
### Describe the bug

When attempting to connect Langfuse to a Keycloak instance via OIDC, only the default/predefined scopes are supported. There is currently no option to specify additional scopes (e.g., for roles or extended user info), as is possible with the custom OIDC provider in NextAuth.

At the same time, using the custom OIDC provider is not a viable workaround, because the returned fields from Keycloak (such as refresh_expires_in and others) do not match the Prisma DB schema defined in Langfuse. This results in errors like:

`Unknown argument `refresh_expires_in`. Available options are marked with ?.`

Expected behavior:

- It should be possible to define additional scopes when configuring Keycloak/OIDC in Langfuse, similar to the flexibility provided by the custom OIDC provider.
- Alternatively, the returned fields from custom OIDC providers should be filtered/mapped to match the existing Prisma DB schema, or the schema should be updated accordingly.

### Steps to reproduce

1. Configure Langfuse to use Keycloak with non-default scopes via OIDC -> Login broken
3. Use custom OIDC configuration with keycloak and custom scopes -> Login broken

Self-hosted

 3.110.0

-

### Additional information

The fix itself should be a few LOC to add a custom scope AUTH_KEYCLOAK_SCOPE (as well as AUTH_KEYCLOAK_ID_TOKEN when we're already changing it), but I am not sure of the implications of this change for the full application.

Yes

freinold:
## What does this PR do?

NextAuth Keycloak Provider optionally supports custom scopes and user_info endpoint, just like custom auth provider did.  

Fixes #9186

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Adds support for Keycloak scope and ID token configuration in `env.mjs` and `auth.ts`.
> 
>   - **Behavior**:
>     - Adds support for `AUTH_KEYCLOAK_SCOPE` and `AUTH_KEYCLOAK_ID_TOKEN` in `env.mjs` for Keycloak configuration.
>     - Updates `KeycloakProvider` in `auth.ts` to use `AUTH_KEYCLOAK_SCOPE` and `AUTH_KEYCLOAK_ID_TOKEN`.
>   - **Environment Variables**:
>     - Introduces `AUTH_KEYCLOAK_SCOPE` and `AUTH_KEYCLOAK_ID_TOKEN` in `env.mjs`.
>   - **Providers**:
>     - Modifies `KeycloakProvider` in `auth.ts` to include `idToken` and `authorization` parameters.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 217964519de8b116339dff06a6f3b85d0cb089c5. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/env.mjs
```diff
@@ -125,6 +125,8 @@ export const env = createEnv({
     AUTH_KEYCLOAK_ALLOW_ACCOUNT_LINKING: z.enum(["true", "false"]).optional(),
     AUTH_KEYCLOAK_CLIENT_AUTH_METHOD: zAuthMethod,
     AUTH_KEYCLOAK_CHECKS: zAuthChecks,
+    AUTH_KEYCLOAK_SCOPE: z.string().optional(),
+    AUTH_KEYCLOAK_ID_TOKEN: z.enum(["true", "false"]).optional(),
```
Steffen911: @freinold Do you think it makes more sense to set a `default` here and make it `true`? That could simplify the odd negation check for the `idToken` property in `auth.ts`.

Steffen911 (commented):
Thank you a lot for adding this!

web/src/env.mjs
freinold: I'm open to it, I just copied and changed the env vars used for the custom OIDC provider.
These probably should be changed as well then.

Steffen911: @freinold Yes, I agree. Could you incorporate those changes here?

freinold: @Steffen911 incorporated the changes like you asked to, good now?

Steffen911 (approved):

PR #9426, opened 2025-09-30
feat(billing): Enforce Usage Threshold on Hobby Plan (DRAFT)

FroeMic:
_// integration branch for feature_

**Notes**
- (!) Requires migration for deployment, because of new columns in organization table
- How to run tests for billing cycle: From web folder `pnpm test-sync -- billingCycleHelpers.servertest`
- How to run test for worker: From worker foler `pnpm run test usageAggregation` and `pnpm run test thresholdProcessing    `

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit fe34ef01833cbe767d19c043cb65e05e45fb422a. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

----

> [!IMPORTANT]
> Enforces usage thresholds on Hobby Plan with new database columns, email notifications, API blocking, and cache invalidation, along with comprehensive tests.
> 
>   - **Behavior**:
>     - Enforces usage thresholds on Hobby Plan with new columns in `Organization` model (`cloud_billing_cycle_anchor`, `cloud_billing_cycle_updated_at`, `cloud_current_cycle_usage`, `cloud_free_tier_usage_threshold_state`).
>     - Adds email notifications for usage warnings and suspensions in `usageThresholdWarning` and `usageThresholdSuspension`.
>     - Blocks API access when usage exceeds thresholds in `ingestion.ts`, `media/index.ts`, and `otel/v1/traces/index.ts`.
>   - **Cache Invalidation**:
>     - Adds `invalidateCachedApiKeys`, `invalidateCachedOrgApiKeys`, and `invalidateCachedProjectApiKeys` functions in `apiAuth.ts`.
>     - Invalidates cache when org state changes in `thresholdProcessing.ts`.
>   - **Testing**:
>     - Adds tests for threshold processing in `thresholdProcessing.test.ts`.
>     - Adds tests for usage aggregation in `usageAggregation.test.ts`.
>     - Adds tests for cache invalidation in `usageThresholdCacheInvalidation.test.ts`.
>   - **Misc**:
>     - Adds `cloudFreeTierUsageThresholdQueue.ts` for processing usage threshold jobs.
>     - Updates environment variables in `env.ts` for new features.
>     - Adds `send-test-threshold-emails.ts` script for testing email notifications.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b1db9cc2ef8e52bf5173597a8a84fb740870eb6f. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/migrations/20250930100520_add_billing_cycle_columns/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- AlterTable
+ALTER TABLE "organizations" ADD COLUMN     "billing_cycle_anchor" TIMESTAMP(3),
+ADD COLUMN     "billing_cycle_last_updated_at" TIMESTAMP(3),
+ADD COLUMN     "billing_cycle_last_usage" INTEGER;
+
+-- CreateIndex
+CREATE INDEX "organizations_billing_cycle_anchor_idx" ON "organizations"("billing_cycle_anchor");
```
marcklingen: add "concurrently", split to separate migration file in order for this to work

packages/shared/prisma/migrations/20250930143723_change_billing_cycle_to_timestamptz/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "organizations" ALTER COLUMN "billing_cycle_anchor" SET DATA TYPE TIMESTAMPTZ(3),
```
marcklingen: please merge the migrations into one migration

packages/shared/prisma/migrations/20250930100520_add_billing_cycle_columns/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- AlterTable
+ALTER TABLE "organizations" ADD COLUMN     "billing_cycle_anchor" TIMESTAMP(3),
```
marcklingen: default the anchor to created at and now() for new rows?

packages/shared/prisma/schema.prisma
```diff
@@ -104,19 +104,24 @@ model VerificationToken {
...
-  MembershipInvitation    MembershipInvitation[]
-  ApiKey                  ApiKey[]
-  surveys                 Survey[]
+  id                        String                   @id @default(cuid())
+  name                      String
+  createdAt                 DateTime                 @default(now()) @map("created_at")
+  updatedAt                 DateTime                 @default(now()) @updatedAt @map("updated_at")
+  cloudConfig               Json?                    @map("cloud_config") // Langfuse Cloud, for zod schema see @/src/features/organizations/utils/cloudConfigSchema
+  metadata                  Json?
+  billingCycleAnchor        DateTime?                @map("billing_cycle_anchor") @db.Timestamptz(3)
```
marcklingen: why do we use db.Timestamptz here? we have always used prisma DateTime

packages/shared/src/server/queues.ts
```diff
@@ -259,6 +259,7 @@ export enum QueueName {
   IngestionQueue = "ingestion-queue", // Process single events with S3-merge
   IngestionSecondaryQueue = "secondary-ingestion-queue", // Separates high priority + high throughput projects from other projects.
   CloudUsageMeteringQueue = "cloud-usage-metering-queue",
+  UsageThresholdQueue = "usage-threshold-queue",
```
marcklingen: CloudFreeUsageThresholdQueue?

packages/shared/src/server/utils/billingCycleHelpers.ts
```diff
@@ -0,0 +1,116 @@
...
+      23,
+      59,
+      59,
+      999,
+    ),
+  );
+}
+
+/**
+ * Get billing cycle anchor with fallback to createdAt
```
marcklingen, line 33: IMO we can cut this logic this by making billingCycleAnchor non-nullable
- on creation of the column, set to createdAt
- on newly created rows, default to now()

packages/shared/src/server/utils/billingCycleHelpers.ts
```diff
@@ -0,0 +1,116 @@
+import { getDaysInMonth, subMonths } from "date-fns";
```
marcklingen, line 1: does it make sense to call this "billingcycle", can be confusing as this is only used for free plans, also the "billing cycle" attribute on orgs is only relevant for free plans. consider renaming to make this more clear/maintainable?

eg
- freePlanUsageReset? (idea, do not really like it)

packages/shared/src/server/redis/usageThresholdQueue.ts
```diff
@@ -0,0 +1,63 @@
...
+  redisQueueRetryOptions,
+  getQueuePrefix,
+} from "./redis";
+import { logger } from "../logger";
+
+export class UsageThresholdQueue {
+  private static instance: Queue | null = null;
+
+  public static getInstance(): Queue | null {
+    if (!env.STRIPE_SECRET_KEY) {
```
marcklingen: why does this depend on stripe? if the goal here is to check whether this is langfuse cloud, I'd rely on env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION

packages/shared/src/server/redis/usageThresholdQueue.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (UsageThresholdQueue.instance) {
+      return UsageThresholdQueue.instance;
+    }
+
+    const newRedis = createNewRedisInstance({
+      enableOfflineQueue: false,
+      ...redisQueueRetryOptions,
+    });
+
+    UsageThresholdQueue.instance = newRedis
```
marcklingen: @maxdeichmann can you review the queue configuration?

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -0,0 +1,94 @@
...
+import {
+  usageThresholdDbCronJobName,
+  UsageThresholdDbCronJobStates,
+} from "./constants";
+import { Job } from "bullmq";
+import { processUsageAggregationForAllOrgs } from "./usageAggregation";
+
+export const handleUsageThresholdJob = async (job: Job) => {
+  // Get cron job, create if it does not exist
+  const cron = await prisma.cronJobs.upsert({
```
marcklingen: @maxdeichmann I built this in the past for locking for tasks that need to happen exactly once as we did not trust our queueing to execute exactly once.

Is this still a potential problem? If yes, does it make sense to abstract this once?
For this job, running it multiple times should not be a problem if we lock for every notification that we send out.

worker/src/ee/usageThresholds/usageAggregation.ts
```diff
@@ -0,0 +1,257 @@
...
+ *    - Aggregate to org level
+ *    - Accumulate into running totals
+ *    - Process orgs whose billing cycle starts on THIS day
+ *
+ * @param referenceDate - Date to process from (default: now)
+ * @param onProgress - Optional callback to report progress (0.0 to 1.0)
+ */
+export async function processUsageAggregationForAllOrgs(
+  referenceDate: Date = new Date(),
+  onProgress?: (progress: number) => void | Promise<void>, // eslint-disable-line no-unused-vars
```
marcklingen, line 167: es-klint disable seems unnecessary here

worker/src/ee/usageThresholds/usageAggregation.ts
```diff
@@ -0,0 +1,257 @@
...
+): Promise<void> {
+  // Normalize referenceDate to UTC end of day
+  const normalizedReferenceDate = endOfDayUTC(referenceDate);
+
+  // Setup
```
marcklingen, line 172: imo would be much easier to read without extracting it into separate functions, difficult for me to understand what is going on here

worker/src/ee/usageThresholds/usageAggregation.ts
```diff
@@ -0,0 +1,257 @@
...
+  const orgs = await prisma.organization.findMany();
+
+  return orgs.map((org) => parseDbOrg(org));
+}
+
+/**
+ * Calculate billing cycle start date for all organizations
+ * relative to a reference date (typically "now")
+ */
+function calculateBillingStartsForAllOrgs(
```
marcklingen: see comment on getBillingCycleStart, seems unnecessary if we set the default correctly

worker/src/ee/usageThresholds/thresholdProcessing.ts
```diff
@@ -0,0 +1,285 @@
...
+ * @param cumulativeUsage - Total usage for the billing cycle
+ */
+export async function processThresholds(
+  org: ParsedOrganization,
+  cumulativeUsage: number,
+): Promise<void> {
+  // 0. GTM-1465: Check if enforcement is enabled
+  if (env.LANGFUSE_USAGE_THRESHOLD_ENFORCEMENT_ENABLED !== "true") {
+    logger.info(
+      `[USAGE THRESHOLDS] Enforcement disabled via feature flag for org ${org.id}, tracking usage only`,
```
marcklingen: this is not a feature flag for an org id but global, IMO should be checked at beginning of job and not here

worker/src/ee/usageThresholds/constants.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Usage threshold constants for free-tier enforcement
+ */
+
+/**
+ * Maximum events allowed in free plan (50,000)
+ * Copied from web billing constants to avoid cross-package dependency
+ */
+export const MAX_EVENTS_FREE_PLAN = 50_000;
```
marcklingen, line 9: move to single global constant if possible

worker/src/ee/usageThresholds/thresholdProcessing.ts
```diff
@@ -0,0 +1,285 @@
...
+    lastProcessedUsage < BLOCKING_THRESHOLD &&
+    cumulativeUsage >= BLOCKING_THRESHOLD
+  ) {
+    shouldBlock = true;
+  }
+
+  // 4. Send email - blocking email takes precedence
+  let usageState: string | null = null;
+
+  if (shouldBlock) {
```
marcklingen: i think there is a difference between should block and should notify about block

- to make this more safe, I would always update usage state to blocked if cumulativeUsage >= BLOCKING_THRESHOLD
- we should notify the user if this state has changed (previous postgres state, postgres state now)

thereby we should be
- more race condition safe
- send out the right notifications in cases where we might change the BLOCKING_THRESHOLD constant

worker/src/ee/usageThresholds/thresholdProcessing.ts
```diff
@@ -0,0 +1,285 @@
...
+  await prisma.organization.update({
+    where: { id: org.id },
+    data: {
+      billingCycleLastUsage: cumulativeUsage,
+      billingCycleLastUpdatedAt: new Date(), // Stored as UTC in timestamptz column
+      billingCycleUsageState: usageState,
+    },
+  });
+
+  // 6. Invalidate API key cache if blocking state changed
```
marcklingen, line 404: could do this whenever the state changes, no matter the before and after values

packages/shared/src/server/auth/invalidateApiKeys.ts
```diff
@@ -0,0 +1,110 @@
...
+ * e.g. after key rotation, revocation, or entitlement/plan changes.
+ *
+ * Behavior:
+ * - Skips keys without a `fastHashedSecretKey`
+ * - No-ops when Redis is not configured
+ *
+ * @param apiKeys - List of API key records to invalidate
+ * @param identifier - Context string for logging (e.g., org or project identifier)
+ */
+export async function invalidate(apiKeys: ApiKey[], identifier: string) {
```
marcklingen: invalidate is not a good name, needs to be more specific now that it is moved out of the apiAuth file

marcklingen: this does not invalidate the key, it only invalidates the cache of the key, can be misunderstood

packages/shared/src/server/auth/invalidateApiKeys.ts
```diff
@@ -0,0 +1,110 @@
+import { prisma } from "../../db";
+import { redis, safeMultiDel } from "..";
+import { logger } from "../logger";
+
+import { type ApiKey } from "../../db";
+
+/**
+ * Invalidate specific API keys from Redis cache
```
marcklingen: concerns moving this logic from api auth to here?

web/src/features/public-api/server/apiAuth.ts
```diff
@@ -37,51 +39,15 @@ export class ApiAuthService {
   // - when projects move across organisations, the orgId in the API key cache needs to be updated
   // - when the plan of the org changes, the plan in the API key cache needs to be updated as well
   async invalidate(apiKeys: ApiKey[], identifier: string) {
```
marcklingen: why wrap these functions here, just use them right away in the code below

packages/shared/prisma/schema.prisma
```diff
@@ -104,19 +104,24 @@ model VerificationToken {
...
+  billingCycleLastUpdatedAt DateTime?                @map("billing_cycle_last_updated_at") @db.Timestamptz(3)
+  billingCycleLastUsage     Int?                     @map("billing_cycle_last_usage")
+  billingCycleUsageState    String?                  @map("billing_cycle_usage_state")
```
marcklingen: do we do anything with usagestate warning? seems like we could cut complexity and just make this a flag
- `freePlanAboveUsageLimit:bool`
- `freePlanUsageResets/Started:Date` (started is probably easier)
- `freePlanUsageInCurrentCycle:Date`
- `freePlanUsageLastUpdatedAt:Date`

marcklingen: i think "billing cycle" is confusing as this is only used for free plans

web/src/features/public-api/server/apiAuth.ts
```diff
@@ -427,6 +395,7 @@ export class ApiAuthService {
           createdAt: Date;
           updatedAt: Date;
           cloudConfig: Prisma.JsonValue;
+          billingCycleUsageState: string | null;
```
marcklingen: if we turned this into a bool (see note in schema), the state in redis would just include "isBlocked" which i think would be more readable for those who have not touched this change

web/src/ee/features/usage-thresholds/services/setBillingCycleAnchor.ts
```diff
@@ -0,0 +1,29 @@
+import { prisma } from "@langfuse/shared/src/db";
+import { startOfDayUTC } from "@langfuse/shared/src/server";
+
+/**
+ * Get organization create data with billing cycle anchor set to start of day UTC
+ */
+export function getOrgCreateDataWithAnchor<T extends Record<string, unknown>>(
```
marcklingen: can we remove this and make this db logic by setting anchor to now() when creating new row? needing to wrap the body with this will lead to problems in the future where we forget about this

web/src/ee/features/billing/server/stripeWebhookHandler.ts
```diff
@@ -532,6 +536,16 @@ async function handleSubscriptionChanged(
...
 
+    // Set billing cycle anchor on first paid subscription from Stripe
+    if (action === "created" && subscription.billing_cycle_anchor) {
+      // Convert unix timestamp (seconds) to Date object
+      const anchorDate = new Date(subscription.billing_cycle_anchor * 1000);
+      await updateOrgBillingCycleAnchor(parsedOrg.id, anchorDate);
+    }
+
+    // Invalidate API keys in Redis for it to be updated
+    await invalidateOrgApiKeys(parsedOrg.id);
```
marcklingen: why do we now need invalidateOrgApiKeys in if and else path here instead of only once below if/else as before?

web/src/ee/features/admin-api/server/organizations/index.ts
```diff
@@ -54,10 +55,10 @@ export async function handleCreateOrganization(
 
   // Create the organization in the database
   const organization = await prisma.organization.create({
-    data: {
+    data: getOrgCreateDataWithAnchor({
```
marcklingen: see note on getOrgCreateDataWithAnchor, i think we should cut this and bne able to do this on db level to make sure we cannot create orgs that are in an invalid state

packages/shared/package.json
```diff
@@ -78,6 +78,7 @@
     "@types/bcryptjs": "^2.4.6",
     "bcryptjs": "^2.4.3",
     "bullmq": "^5.34.10",
+    "date-fns": "^3.3.1",
```
marcklingen, line 81: i think we had a dependency on date-fns before and I think Max removed it for some reasons, pls check with him if this is a problem

@maxdeichmann

packages/shared/src/server/services/email/usageThresholdSuspension/UsageThresholdSuspensionEmailTemplate.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+            <Hr className="border border-solid border-[#eaeaea] my-[26px] mx-0 w-full" />
+
+            <Section>
+              <Text className="text-[#666666] text-[12px] leading-[24px]">
+                This urgent notification was sent to {receiverEmail} regarding
+                ingestion suspension for &quot;{organizationName}&quot;.
+              </Text>
+              <Text className="text-[#666666] text-[12px] leading-[24px]">
+                Need immediate help? Contact us at{" "}
```
marcklingen: we no longer want to push support@ contact option but rather prompt people to conatct us via the new support drawer. I think adding a "reply-to" to this email and ask people to reply if they have questions is a good idea -> thereby we at least have the context of this email

worker/src/queues/cloudFreeTierUsageThresholdQueue.ts
```diff
@@ -0,0 +1,36 @@
...
+import { usageThresholdDbCronJobName } from "../ee/usageThresholds/constants";
+import { UsageThresholdDbCronJobStates } from "../ee/usageThresholds/constants";
+import { prisma } from "@langfuse/shared/src/db";
+
+export const usageThresholdQueueProcessor: Processor = async (job) => {
+  if (job.name === QueueJobs.UsageThresholdJob) {
+    logger.info("Executing Usage Threshold Job", job.data);
+    try {
+      return await handleUsageThresholdJob(job);
+    } catch (error) {
```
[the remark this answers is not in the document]
FroeMic: @Steffen911 I am using the same pattern here as `cloudUsageMeteringQueue.ts` to queue another job in catch. Is this advisable?

packages/shared/prisma/migrations/20250930100520_add_billing_cycle_columns/migration.sql
FroeMic: Would advise against it. We explicitly deal with null values in code to account for existing subscriptions. Default values could be filled on write by accident with arbitrary dates and/or we would need to run migrations to backfill.

packages/shared/prisma/schema.prisma
FroeMic: DateTime has no notion of timezones, so we have no guarantee whether the time is UTC or another timezone. Since we write startOfDay timestamp (time value zeroed out) this leads to edge cases, where we actually write a 22:00 value of the day before (in German timezone) and cause a one-off error. 

tl'dr: use of timestamptz solves this problem and makes working with datetimes much more robuts

packages/shared/src/server/queues.ts
FroeMic: FreeTierUsageThresholdQueue

web/src/ee/features/usage-thresholds/services/setBillingCycleAnchor.ts
FroeMic: Removed the wrapper and explicitly set the database property.

Note: Forgetting will not lead to issues, since we fallback to createdAt for null values

web/src/ee/features/billing/server/stripeWebhookHandler.ts
FroeMic: I'd prefer having "things that conceptually belong together co-located".

If we place the statement inside the conditions, we have to type it twice, but the code in both branches ends, without having to jump back out. In my opinion it is much clearer, that way what happens, particularly in the first conditional branch.

FroeMic: Style choice - and no hard opinions here. If you would like it to change, I will do so!

packages/shared/prisma/migrations/20250930100520_add_billing_cycle_columns/migration.sql
marcklingen: let's keep these threads unresolved until we are aligned

IMO backfill when now once adding the col is very easy + transactional

afterwards the column is always non-null

all new records could get a default now() on prisma or db level and thus cannot be forgotten

packages/shared/prisma/schema.prisma
marcklingen: @maxdeichmann can you have a look into this?

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
maxdeichmann: If possible i would try to build jobs idempotent (e.g. running CH queries to get daily billable events) and only "lock" e.g. the db to make sure we only send out the email once. The "execute the job exactly once" logic is pretty complex IMO.

web/src/ee/features/usage-thresholds/services/setBillingCycleAnchor.ts
marcklingen: see note on db logic above, IMO we could also cut the defaulting on createdAt on read if we set this for every row

FroeMic: Setting it on every row would require backfilling all existing organizations and – to avoid out-of-synch data with paid plans – doing so in connection with stripe (if not, we have drift in the db)

FroeMic: I think the ROI is low, given that there are exactly two places in code, where we create an organization and missing it is already handled.

packages/shared/package.json
FroeMic: on it

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "organizations" ADD COLUMN     "billing_cycle_usage_state" TEXT;
```
Steffen911: Do we want to use multiple columns here? An alternative could be a single JSON with all relevant properties.

packages/shared/package.json
maxdeichmann: We currently have this dep in the worker container. I think we are fine if this does not add vulnerabilities.

packages/shared/prisma/schema.prisma
FroeMic: resolved (talked to max) – adding a comment for clarification to prisma

packages/shared/src/server/auth/invalidateApiKeys.ts
```diff
@@ -0,0 +1,110 @@
+import { prisma } from "../../db";
+import { redis, safeMultiDel } from "..";
+import { logger } from "../logger";
+
+import { type ApiKey } from "../../db";
+
+/**
```
Steffen911, line 7: I feel like bundling the rejection with auth is odd. IMO we could have a separate lookup + cache for this behaviour which would remove the need to invalidate API keys from Redis from the worker.

Having a separate condition check could make the behaviour more obvious, especially since it's only used for ingestion. E.g. prompt management should never need the relevant new property.

packages/shared/package.json
FroeMic: resolved – (talked to Max) – can remain in

packages/shared/src/server/auth/types.ts
```diff
@@ -16,6 +16,7 @@ const ApiKeyBaseSchema = z.object({
   orgId: z.string(),
   plan: z.enum(plans as unknown as [string, ...string[]]),
   rateLimitOverrides: CloudConfigRateLimit.nullish(),
+  isIngestionSuspended: z.boolean(),
```
Steffen911, line 19: As said in a previous comment, I'd decouple this from auth and make it a separate cache + condition check.

packages/shared/src/server/redis/usageThresholdQueue.ts
```diff
@@ -0,0 +1,63 @@
...
+        QueueJobs.UsageThresholdJob,
+        {},
+        {
+          // Run at minute 35 of every hour (30 minutes after cloudUsageMetering at :05)
+          repeat: { pattern: "35 * * * *" },
+        },
+      );
+
+      UsageThresholdQueue.instance.add(QueueJobs.UsageThresholdJob, {}, {});
+    }
```
Steffen911: Would it make sense to trigger this queue in the final step of the Cloud Usage Metering job? This adds a timing dependency to it which feels unnecessary. Instead, we could just run it right after the billing job - optionally with a delay.

IMO we should reduce the number of cron schedules as this makes the system harder to understand.

Steffen911: We have the full org context in the billing job, so we could actually trigger one usage threshold check job per (free) organization. This will make each job more reliable, too.

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1635,3 +1635,62 @@ export const getGenerationsForPostHog = async function* (
...
+}: {
+  startDate: Date;
+  endDate: Date;
+}) => {
+  const query = `
+    SELECT
+      count(*) as count,
+      project_id,
+      toDate(start_time) as date
+    FROM observations
```
Steffen911, line 1672: If you do this without FINAL, you may have double-counting. Alternatively, you can count `uniq(id)` on the observations table to get something (more) accurate. There is also uniqExact I believe.

packages/shared/src/server/queues.ts
```diff
@@ -429,4 +431,10 @@ export type TQueueJobTypes = {
     payload: EntityChangeEventType;
     name: QueueJobs.EntityChangeJob;
   };
+  [QueueName.UsageThresholdQueue]: {
+    timestamp: Date;
+    id: string;
+    payload: Record<string, never>;
```
Steffen911: Why have a payload if it doesn't contain values?

worker/src/ee/usageThresholds/handleUsageThresholdJob.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  logger.info("[USAGE THRESHOLDS] Job started");
+
+  try {
+    // Call the main usage aggregation function from GTM-1461
+    await processUsageAggregationForAllOrgs(new Date(), async (progress) => {
+      // Update job progress to prevent staleness
+      await job.updateProgress(progress * 100); // BullMQ expects 0-100
+    });
```
Steffen911: I would love for this to become a per-organization job instead of one big batch.

worker/src/queues/cloudFreeTierUsageThresholdQueue.ts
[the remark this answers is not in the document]
Steffen911: I think Max and Marc have comments on the cronJob pattern. If possible, let's kill it and then also the corresponding error handling.

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
FroeMic: yes, we do! :-D

Always a mess to build custom zod.schema.parse logic around retrieval and end up without the guarantues of a database. 

What advantage would you see otherwise?

worker/src/ee/usageThresholds/handleUsageThresholdJob.ts
FroeMic: Great point! For now we settled on a batch job, to reduce load on clickhouse.

packages/shared/src/server/queues.ts
FroeMic: removed

packages/shared/src/server/repositories/observations.ts
FroeMic: by design

packages/shared/src/server/redis/usageThresholdQueue.ts
FroeMic: noted for future

packages/shared/src/server/auth/types.ts
FroeMic: as discussed: will keep for now

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
FroeMic: will keep for now – since clickhouse pressure is a priority, we can keep it as is

worker/src/ee/usageThresholds/usageAggregation.ts
FroeMic: removed

worker/src/ee/usageThresholds/thresholdProcessing.ts
FroeMic: We check this here, so we can have this run in prod before enforcing any usage limits

worker/src/ee/usageThresholds/thresholdProcessing.ts
FroeMic: refactored

packages/shared/src/server/auth/invalidateApiKeys.ts
FroeMic: refactored

packages/shared/src/server/auth/invalidateApiKeys.ts
FroeMic: if we wanted to reuse the functionality in both web and worker package we needed to extract it to shared

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -6,8 +6,21 @@ import {
 } from "./constants";
 import { Job } from "bullmq";
 import { processUsageAggregationForAllOrgs } from "./usageAggregation";
+import { backfillBillingCycleAnchors } from "./backfillBillingCycleAnchors";
 
 export const handleUsageThresholdJob = async (job: Job) => {
+  // TECH DEBT: Backfill billing cycle anchors for organizations without one
+  // TODO: Remove this call once all organizations have been backfilled (target: Q2 2025)
```
marcklingen: nit: q2 2025

web/src/features/public-api/server/apiAuth.ts
```diff
@@ -9,7 +9,7 @@ import {
   logger,
   instrumentAsync,
   addUserToSpan,
-  invalidate as invalidateShared,
+  invalidateCachedApiKeys as invalidateCachedApiKeysShared,
   invalidateOrgApiKeys as invalidateOrgApiKeysShared,
```
marcklingen: "cached" also makes sense for the other invalidate methods here

worker/src/ee/usageThresholds/backfillBillingCycleAnchors.ts
```diff
@@ -0,0 +1,229 @@
...
+  errors: number;
+}> {
+  const startTime = Date.now();
+  logger.info("Starting billing cycle anchor backfill");
+
+  try {
+    // Find all organizations with null billingCycleAnchor
+    const orgsToBackfill = await prisma.organization.findMany({
+      where: {
+        billingCycleAnchor: null,
```
marcklingen: i'd assume that `ALTER TABLE "organizations" ADD COLUMN     "billing_cycle_anchor" TIMESTAMP(3) DEFAULT CURRENT_TIMESTAMP` will auto-backfill to now() for all existing rows and thus this wont trigger at all

i'd suggest to add a migration.sql that
1. if no active subscription is in postgres, sets billing_cycle_anchor to created_at
2. if one is in postgres, sets it to null -> this job then can pick it up

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
marcklingen: fair point from steffen

i think we could quickly chat on whether it actually makes sense to move this into cloudConfig as we store all data that is irrelevant for self-hosted usually there

(let's not resolve threads while they are undecided)

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
```diff
@@ -119,10 +119,16 @@ export const handleCloudUsageMeteringJob = async (job: Job) => {
...
-  ).map(({ projects, ...org }) => ({
-    ...parseDbOrg(org),
-    projectIds: projects.map((p) => p.id),
-  }));
+  )
+    .map(({ projects, ...org }) => ({
+      ...parseDbOrg(org),
+      projectIds: projects.map((p) => p.id),
+    }))
+    .filter(
```
marcklingen: filter here not needed, we filter on customerId above in postgres and want to run this job for every org that has a customerId linked in stripe

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -89,7 +99,83 @@ export const handleUsageThresholdJob = async (job: Job) => {
       },
     });
 
-    logger.info("[USAGE THRESHOLDS] Job completed successfully");
+    // Record DataDog metrics
```
marcklingen: discuss with @maxdeichmann, the number of gauges here does not seem to be required
- we can track exceptions in datadog for what we want to get alerts on

packages/shared/src/server/queues.ts
```diff
@@ -429,4 +431,9 @@ export type TQueueJobTypes = {
     payload: EntityChangeEventType;
     name: QueueJobs.EntityChangeJob;
   };
+  [QueueName.FreeTierUsageThresholdQueue]: {
+    timestamp: Date;
+    id: string;
+    name: QueueJobs.UsageThresholdJob;
```
marcklingen: please add freetier to the job as well

web/src/features/public-api/server/apiAuth.ts
```diff
@@ -36,52 +38,16 @@ export class ApiAuthService {
...
-    }
-
-    if (this.redis) {
-      logger.info(`Invalidating API keys in redis for ${identifier}`);
-      const keysToDelete = filteredHashKeys.map((hash) =>
-        this.createRedisKey(hash),
-      );
-      await safeMultiDel(this.redis, keysToDelete);
-    }
+  async invalidateCachedApiKeys(apiKeys: ApiKey[], identifier: string) {
```
marcklingen, line 41: wrapping of the functions here seems unnecessary, pls remove

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
FroeMic: Let's do so once you are in the office – imo we do not gain anything by essentially reimplementing ORM logic ourselves while giving up on the guarantues that the database provides us. 

I don't fully understand the mental model behind the cloudConfig in the first place – so curious to understand.

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
FroeMic: Paid Sub -> Cancelled -> Hobby Plan leads to a case where metered events are sent to stripe that do not have an active subscription. Consequently, the number of metered events in stripe does not reflect the number of "paid metered events".

Would recommend being strict here.

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
FroeMic: removed

worker/src/ee/usageThresholds/backfillBillingCycleAnchors.ts
FroeMic: fixed

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
FroeMic: @marcklingen -> just let me know your decision and I will implement here

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
FroeMic: @marcklingen  After talking with Steffen my understanding is that we tell self-hosters that they don't need to worry about "cloudConfig" but the rest is relevant for them.

I see three ways to resolve this

1. Reuse cloudConfig for thresholds 
2. Create one new column cloudUsageThresholds as jsonb and fold current columnns in there
3. Keep current columns and rename columns to `cloudUsageCycle ...` to signal cloud relation
4. Create new table cloudConfiguration and move current columns (but to cloudConfig for now) over there. 

My preferred options would be (3) or (4).

Steffen911: I like 4 best, but would not disagree with any of the others.

packages/shared/prisma/schema.prisma
```diff
@@ -104,19 +104,24 @@ model VerificationToken {
...
+  aiFeaturesEnabled         Boolean                  @default(false) @map("ai_features_enabled")
+  organizationMemberships   OrganizationMembership[]
+  projects                  Project[]
+  MembershipInvitation      MembershipInvitation[]
+  ApiKey                    ApiKey[]
+  surveys                   Survey[]
 
+  @@index([billingCycleAnchor])
```
Steffen911: IMO table is small enough to handle a full table scan anytime for this. Probably postgres would fallback to it anyway instead of using the index. I'd vote to drop and only add it if we see this being a considerable load.

packages/shared/src/server/redis/usageThresholdQueue.ts
```diff
@@ -0,0 +1,68 @@
...
+          repeat: { pattern: "35 * * * *" },
+        },
+      );
+
+      UsageThresholdQueue.instance.add(
+        QueueJobs.FreeTierUsageThresholdJob,
+        {},
+        {},
+      );
+    }
```
Steffen911: As discussed, let's drop this and schedule a job from the cloudUsageMetering job. If you happen to disagree, also good, but it seems to fit your target vision.

Having some mermaid chart of flow chart of those billing jobs would also be useful.

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1635,3 +1635,62 @@ export const getGenerationsForPostHog = async function* (
     };
   }
 };
+
+/**
+ * Get observation counts grouped by project and day within a date range.
+ *
+ * Returns one row per project per day with the count of observations started on that day.
+ * Uses half-open interval [startDate, endDate) for filtering based on start_time.
```
Steffen911, line 1643: Please add a note on why we can skip FINAL in these queries.

packages/shared/src/server/queues.ts
```diff
@@ -259,6 +259,7 @@ export enum QueueName {
   IngestionQueue = "ingestion-queue", // Process single events with S3-merge
   IngestionSecondaryQueue = "secondary-ingestion-queue", // Separates high priority + high throughput projects from other projects.
   CloudUsageMeteringQueue = "cloud-usage-metering-queue",
+  FreeTierUsageThresholdQueue = "free-tier-usage-threshold-queue",
```
Steffen911: Please align the queue name (packages/shared/src/server/redis/usageThresholdQueue.ts) with this name here. The class should be the same as the enum ideally.

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
```diff
@@ -119,10 +119,16 @@ export const handleCloudUsageMeteringJob = async (job: Job) => {
...
+      (org) =>
+        org.cloudConfig?.stripe?.activeSubscriptionId !== undefined &&
+        org.cloudConfig?.stripe?.activeSubscriptionId !== null,
+    ); // only process active subscriptions
```
Steffen911: Can we move this into the postgres query? If not, easy.

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
```diff
@@ -218,24 +228,50 @@ export const handleCloudUsageMeteringJob = async (job: Job) => {
...
+    },
+  );
+  recordGauge(
+    "langfuse.queue.cloud_usage_metering_queue.skipped_orgs",
+    countSkippedOrgs,
+    {
+      unit: "organizations",
+    },
+  );
+  recordGauge(
```
Steffen911: I feel like an increment (count) would be the more natural metric choice here: https://docs.datadoghq.com/metrics/types/?tab=count#metric-types. Please take a look into the metric type definitions and decide yourself what to choose.

Adding the metrics is a non-issue IMO.

worker/src/ee/usageThresholds/backfillBillingCycleAnchors.ts
```diff
@@ -0,0 +1,229 @@
...
+ * TECH DEBT: This function backfills billing cycle anchors for organizations.
+ * TODO: Remove this function once all organizations have been backfilled
+ *
+ * Background:
+ * - Organizations created before the billing cycle anchor feature need to be backfilled
+ * - For hobby plan users: set anchor to organization createdAt
+ * - For paid plan users: fetch from Stripe subscription and use billing_cycle_anchor
+ *
+ * This function should be removed once all existing organizations have been migrated.
+ */
```
Steffen911: This is a _great_ case for our background migrations (https://github.com/langfuse/langfuse/tree/main/worker/src/backgroundMigrations). They will run on startup in the worker container exactly once per deployment. Alternatively, if this is cloud only anyway, just execute the script manually via https://github.com/langfuse/langfuse/tree/main/worker/src/scripts. 

Here, it makes little sense.

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -0,0 +1,204 @@
...
+  // Get cron job, create if it does not exist
+  const cron = await prisma.cronJobs.upsert({
+    where: { name: usageThresholdDbCronJobName },
+    create: {
+      name: usageThresholdDbCronJobName,
+      state: UsageThresholdDbCronJobStates.Queued,
+      lastRun: null,
+    },
+    update: {},
+  });
```
Steffen911: As discussed, just set a global concurrency in the app.ts as there are no bad things that happen if this runs twice for any reason. Let's remove this complexity.

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -0,0 +1,204 @@
...
+        jobStartedAt: null,
+      },
+    });
+
+    // Record DataDog metrics
+    recordGauge(
+      "langfuse.queue.usage_threshold_queue.total_orgs",
+      stats.totalOrgs,
+      { unit: "organizations" },
+    );
```
Steffen911: Here, counts feel more intuitive again. If you find those useful, it's fine to keep them as is.

Steffen911: You could also submit the "same" metric, but set different labels. That would allow for breakdowns by label which will have the same effect but is slightly cheaper.

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -0,0 +1,204 @@
...
+      stats.blockingEmailsSent,
+      { unit: "emails" },
+    );
+
+    // Number of emails that failed to send in this job run
+    recordGauge(
+      "langfuse.queue.usage_threshold_queue.email_failures",
+      stats.emailFailures,
+      { unit: "emails" },
+    );
```
Steffen911: 100% a count metric. Just do a `recordIncrement` per email attempt instead of pushing the aggregations here.

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
```diff
@@ -0,0 +1,204 @@
...
+      backfillSuccessful,
+      { unit: "organizations" },
+    );
+
+    // Number of organizations that failed during backfill process
+    recordGauge(
+      "langfuse.queue.usage_threshold_queue.backfill_errors",
+      backfillErrors,
+      { unit: "organizations" },
+    );
```
Steffen911: There shouldn't be any reason to deploy this and run it multiple times.

worker/src/ee/usageThresholds/thresholdProcessing.ts
```diff
@@ -0,0 +1,350 @@
...
+      logger.warn(
+        `[USAGE THRESHOLDS] No admin/owner emails found for org ${org.id}`,
+      );
+      return { emailSent: false, emailFailed: false };
+    }
+
+    // Generate billing URL
+    const billingUrl = env.NEXTAUTH_URL
+      ? `${env.NEXTAUTH_URL}/organization/${org.id}/settings/billing`
+      : `https://cloud.langfuse.com/organization/${org.id}/settings/billing`;
```
Steffen911: This should never be undefined or falsy in cloud. I'd rather raise an error than linking to the wrong page.

worker/src/ee/usageThresholds/thresholdProcessing.ts
```diff
@@ -0,0 +1,350 @@
...
+      error,
+    );
+    emailFailed = true;
+  }
+
+  return { emailSent, emailFailed };
+}
+
+/**
+ * GTM-1464: Send blocking notification email
```
Steffen911: No need for internal ticket codes.

worker/src/ee/usageThresholds/usageAggregation.ts
```diff
@@ -0,0 +1,307 @@
...
+  obsCounts: Array<{ count: number; projectId: string; date: string }>,
+  scoreCounts: Array<{ count: number; projectId: string; date: string }>,
+  projectToOrgMap: ProjectToOrgMap,
+): UsageByOrg {
+  const orgCounts: UsageByOrg = {};
+
+  // Aggregate traces
+  for (const row of traceCounts) {
+    const orgId = projectToOrgMap[row.projectId];
+    if (!orgId) continue;
```
Steffen911: Shouldn't this raise an error? Or at least log a warning?

worker/src/ee/usageThresholds/usageAggregation.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    // Fetch counts from Clickhouse (3 queries for ALL projects)
+    const [traceCounts, obsCounts, scoreCounts] = await Promise.all([
+      getTraceCountsByProjectAndDay({ startDate: dayStart, endDate: dayEnd }),
+      getObservationCountsByProjectAndDay({
+        startDate: dayStart,
+        endDate: dayEnd,
+      }),
+      getScoreCountsByProjectAndDay({ startDate: dayStart, endDate: dayEnd }),
+    ]);
```
Steffen911, line 229: Why do you do this query per day instead of one big query that scans the entire month? Results should be the same, right?

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
marcklingen: chatted with Max about this yesterday night as I was also wondering if this matters, we came to conclusion that it does not really matter as database is an implementation detail and we focus more on cloud. Prefixing with cloud (as in 3) seems to be the easiest solution, no need to collapse them all into cloudConfig or manage joins in separate table

worker/src/ee/usageThresholds/usageAggregation.ts
FroeMic: For a detailed answer -> Max or Marc

My understanding from their conversation was that this makes the clickhouse query cheap and avoids one big spike. Instead we make several spaced out calls that are each cheap to perform.

From a domain logic perspective: The issue we need to deal with, is that we have up to 31 different periods we need to fetch: For all orgs with a billing cycle starting yesterday, we need to aggregate events from Today -> Yesterday. Then for all starting the day before from Today to The day before yesterday and so forth. 

Consequently, we always need up to 31 queries. 

We optimize the clickhouse cost, by only fetching 1 day and then aggegrating in the job worker. So instead of fetching [1 day, 2 days, ... 31 days] from clickhouse we fetch [1 day, 1 day ... 1day].

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
FroeMic: three it is!

packages/shared/src/server/queues.ts
FroeMic: now: `CloudFreeTierUsageThresholdQueue`

packages/shared/prisma/migrations/20250930144939_add_billing_cycle_usage_state/migration.sql
FroeMic: done

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
FroeMic: discussed with Steffen -> adressed under his comments

worker/src/ee/usageThresholds/backfillBillingCycleAnchors.ts
FroeMic: ->background job

worker/src/ee/usageThresholds/handleCloudFreeTierUsageThresholdJob.ts
FroeMic: moved to background job

packages/shared/src/server/redis/usageThresholdQueue.ts
FroeMic: will keep it as a cron job for now

worker/src/ee/cloudUsageMetering/handleCloudUsageMeteringJob.ts
FroeMic: reverted for now to not block merging

worker/src/ee/usageThresholds/usageAggregation.ts
FroeMic: marking this as resolved

packages/shared/prisma/migrations/20251001161539_organization_cloud_billing_cycle_columns/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+ADD COLUMN     "cloud_free_tier_usage_threshold_state" TEXT;
+
+-- Backfill cloud_billing_cycle_anchor for existing organizations:
+-- Step 1: Set to NULL for orgs WITH active subscriptions (will be backfilled from Stripe by worker job)
+UPDATE "organizations"
+SET "cloud_billing_cycle_anchor" = NULL;
+
+-- Step 2: Set to created_at for orgs WITHOUT active subscriptions (free tier orgs)
+UPDATE "organizations"
+SET "cloud_billing_cycle_anchor" = "created_at"
```
Steffen911, line 14: Did you test this? I'm always slightly uncertain about the quoting behaviour in postgres.

packages/shared/prisma/migrations/20251002153814_add_backfill_billing_cycle_anchors_background_migration/migration.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO background_migrations (id, name, script, args)
+VALUES ('2025-0102-backfill-billing-cycle-anchors', '20250102_backfill_billing_cycle_anchors', '20250102_backfillBillingCycleAnchors', '{}');
```
Steffen911: We usually use a UUID for the id and there is no need to have the date in the script name.

worker/src/backgroundMigrations/backfillBillingCycleAnchors.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (!env.STRIPE_SECRET_KEY) {
+      return {
+        valid: false,
+        invalidReason:
+          "Migration requires Stripe integration in cloud environment (STRIPE_SECRET_KEY not set)",
+      };
+    }
+
+    return { valid: true, invalidReason: undefined };
+  }
```
Steffen911, line 79: Theoretically, it would make sense to check that the new columns exist in the DB to avoid race-conditions between the prisma migrations and the background migrations. As we have control on cloud this should be acceptable if we deploy it correctly.

worker/src/ee/usageThresholds/constants.ts
```diff
@@ -0,0 +1,37 @@
...
+export const NOTIFICATION_THRESHOLDS = [
+  MAX_EVENTS_FREE_PLAN, // 50,000
+  MAX_EVENTS_FREE_PLAN * 2, // 100,000
+] as const;
+
+/**
+ * Blocking threshold triggers API blocking
+ * When reached, ingestion endpoints are blocked for the organization
+ */
+export const BLOCKING_THRESHOLD = MAX_EVENTS_FREE_PLAN * 4; // 200,000
```
Steffen911: Let's consider setting this to 1M.

worker/src/env.ts
```diff
@@ -16,6 +16,12 @@ const EnvSchema = z.object({
 
   NEXTAUTH_URL: z.string().optional(),
 
+  NEXT_PUBLIC_LANGFUSE_CLOUD_REGION: z
+    .enum(["US", "EU", "STAGING", "DEV", "HIPAA"])
+    .optional(),
```
Steffen911, line 21: We may need to make an infra change to set this before we can deploy.

packages/shared/prisma/migrations/20251001161539_organization_cloud_billing_cycle_columns/migration.sql
FroeMic: yes, tested it locally and worked as expected

worker/src/ee/usageThresholds/constants.ts
FroeMic: @Steffen911 -> current discussion seems to converge on 250k. 

See: https://linear.app/langfuse/issue/GTM-1488/discuss-final-thresholds

worker/src/backgroundMigrations/backfillBillingCycleAnchors.ts
FroeMic: added

packages/shared/prisma/migrations/20251002153814_add_backfill_billing_cycle_anchors_background_migration/migration.sql
FroeMic: updated

PR #9437, opened 2025-09-30
feat(prompt-experiments): allow for structured output

hassiebp:
> [!IMPORTANT]
> Adds structured output support for prompt experiments, allowing JSON schema definitions and processing in both frontend and backend.
> 
>   - **Behavior**:
>     - Adds support for structured output in prompt experiments by allowing users to define JSON schemas (`PromptExperimentsForm.tsx`, `CreateOrEditLLMSchemaDialog.tsx`).
>     - Validates structured output schema selection in `PromptExperimentsForm.tsx`.
>     - Updates `fetchLLMCompletion` to handle structured output schemas (`fetchLLMCompletion.ts`).
>   - **Schema Handling**:
>     - Adds `structured_output_schema` to `ExperimentMetadataSchema` in `types.ts`.
>     - Updates `CreateExperimentData` and `createExperiment` mutation to include `structuredOutputSchema` (`types.ts`, `router.ts`).
>   - **LLM Call**:
>     - Modifies `callLLM` and `processLLMCall` to support structured output schemas (`utilities.ts`, `experimentServiceClickhouse.ts`).
>   - **Misc**:
>     - Changes `LLMJSONSchema` to use `z.any()` instead of `z.unknown()` in `types.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 950a4b04a8842893c2a4eadd0d3ede1550d19afe. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/experiments/components/PromptExperimentsForm.tsx
+++ b/web/src/features/experiments/components/PromptExperimentsForm.tsx
@@ -18,6 +18,7 @@ import {
   InputCommandList,
   InputCommand,
   InputCommandItem,
+  InputCommandSeparator,
 } from "@/src/components/ui/input-command";
 import {
   Select,
@@ -38,7 +39,11 @@ import {
   Info,
   CircleCheck,
   Loader2,
+  PlusIcon,
+  PencilIcon,
 } from "lucide-react";
+import { CreateOrEditLLMSchemaDialog } from "@/src/features/playground/page/components/CreateOrEditLLMSchemaDialog";
+import { type LlmSchema } from "@langfuse/shared";
 import { api } from "@/src/utils/api";
 import {
   Card,
@@ -116,11 +121,21 @@ export const PromptExperimentsForm = ({
   const [selectedPromptVersion, setSelectedPromptVersion] = useState<
     number | null
   >(promptDefault?.version ?? null);
+  const [selectedSchema, setSelectedSchema] = useState<LlmSchema | null>(null);
+  const [schemaPopoverOpen, setSchemaPopoverOpen] = useState(false);
 
   const hasEvalReadAccess = useHasProjectAccess({
     projectId,
     scope: "evalJob:read",
   });
+
+  const savedSchemas = api.llmSchemas.getAll.useQuery(
+    { projectId },
+    {
+      enabled: Boolean(projectId),
+      staleTime: 1000 * 60 * 5, // 5 minutes
+    },
+  );
 
   const form = useForm({
     resolver: zodResolver(CreateExperimentData),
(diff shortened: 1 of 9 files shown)


worker/src/features/utils/utilities.ts
```diff
@@ -81,6 +86,7 @@ export async function callLLM(
   provider: string,
   model: string,
   traceParams?: TraceParams,
+  structuredOutputSchema?: LlmSchema,
```
marliessophie, line 89: I find current code hard to follow. We already have a callStructuredLLM function (see line 51), which takes very similar inputs and calls fetchLLMCompletion with structured output schema but handles completion parsing differently. Would suggest we combine into one callLLM function with conditional structured output input and conditional completion parsing, wdyt?

web/src/features/playground/page/components/CreateOrEditLLMSchemaDialog.tsx
```diff
@@ -95,7 +95,13 @@ export const CreateOrEditLLMSchemaDialog: React.FC<
     }
   }, [existingLlmSchema, form, props.defaultValues]);
 
-  async function onSubmit(values: FormValues) {
+  async function onSubmit(
+    values: FormValues,
+    event?: React.BaseSyntheticEvent,
+  ) {
+    event?.preventDefault();
+    event?.stopPropagation();
```
marliessophie, line 103: why do we need these here? We define stopPropagation on line 174 already. Would co-locate preventDefault and stopPropagation also.

worker/src/features/utils/utilities.ts
marliessophie: Also I believe in some cases the structuredOutputSchema is passed as a z.object or as z.record<string, any>, is this intended? Clearly works with both :)

web/src/features/experiments/components/PromptExperimentsForm.tsx
marliessophie, line 1: did not review this, should I? :)

web/src/features/playground/page/components/CreateOrEditLLMSchemaDialog.tsx
hassiebp: done!

web/src/features/experiments/components/PromptExperimentsForm.tsx
hassiebp: not needed, thanks!

worker/src/features/utils/utilities.ts
hassiebp: The subtle difference here is exactly in the output parsing. For the `callStructuredLLM`, we are parsing the output with the zod schema that is provided. Thus we guarantee that the return value follows the schema provided. 

For `callLLM`, we only return the plain string returned by the LLM without additional parsing. Since the user is providing a JSON schema only, we cannot 'parse' the result without some more extensive logic to transform a JSON schema to a zod schema to then use for parsing. 

Thus, in this case I opted to simply reuse the `callLLM` function to always return a string, and optionally allow passing a JSON Schema but not making any guarantees on the parsed output.

marliessophie: I see. Adding this context as a docstring to both functions and improving the function names would help here. Worried future us might be confused. 

You could also consider a strategy pattern that parses the completion conditionally. But up to you, just ideas :)

PR #9608, opened 2025-10-08
chore: add a delayed ingestion queue to insert everything to new events table

Steffen911:


Steffen911:
superseded with dual write approach.

PR #9615, opened 2025-10-08
feat(auth): add account settings with self-serve account deletion and display name update 

FroeMic:
# What's New

User's can now update their display name or delete their account from a new account settings page.

Account deletion is only possible if one of these conditions is true

- the user is not part of any organization
- the user is not owner of any organization
- the user is owner of an organization, but there is another owner

In case the user wants to delete the the account, they first need to transfer ownership or delete organizations with the existing flows.

CleanShot 2025-10-08 at 20 31 12@2x

----

> [!IMPORTANT]
> Adds account settings page for display name update and account deletion with backend support and UI integration.
> 
>   - **Behavior**:
>     - Adds account settings page for users to update display name and delete account (`index.tsx`).
>     - Account deletion allowed only if user is not part of any organization, not the sole owner, or if another owner exists.
>   - **API**:
>     - Adds `userAccountRouter` with `updateDisplayName`, `checkCanDelete`, and `delete` mutations in `userAccount.ts`.
>     - Integrates `userAccountRouter` into `appRouter` in `root.ts`.
>   - **UI**:
>     - Adds "Account Settings" link to user navigation in `layout.tsx`.
>     - Updates `NavUser` in `nav-user.tsx` to handle navigation items with `href`.
>     - Integrates account settings into `CommandMenu` in `CommandMenu.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b09b55d0e7b9fc73a25f8683c804409be0cf8f8f. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/pages/account/settings/index.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  function onSubmit(values: z.infer<typeof displayNameSchema>) {
+    updateDisplayName.mutate({ name: values.name });
+  }
+
+  return (
+    <div>
+      <Header title="Display Name" />
+      <Card className="mb-4 p-3">
+        {form.getValues().name !== "" ? (
```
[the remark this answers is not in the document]
FroeMic: no

web/src/server/api/routers/userAccount.ts
```diff
@@ -0,0 +1,122 @@
...
+        name: updatedUser.name,
+      };
+    }),
+
+  delete: authenticatedProcedure.mutation(async ({ ctx }) => {
+    const userId = ctx.session.user.id;
+
+    // First, verify user can be deleted
+    const organizationMemberships =
+      await ctx.prisma.organizationMembership.findMany({
```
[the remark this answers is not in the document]
FroeMic: no

web/src/server/api/routers/userAccount.ts
```diff
@@ -0,0 +1,124 @@
...
+      return {
+        success: true,
+        name: updatedUser.name,
+      };
+    }),
+
+  delete: authenticatedProcedure.mutation(async ({ ctx }) => {
+    const userId = ctx.session.user.id;
+
+    // Wrap check and delete in a transaction to prevent race conditions
```
maxdeichmann: Transactions per se do not prevent race conditions. I think we do not have to have this here as this is very unexpected we run into this.

Steffen911:
> the user is not part of an organization

@FroeMic Why does this have to be true? The deletion action could just remove the user from all projects and organizations if they want to delete themselves.

web/src/server/api/routers/userAccount.ts
FroeMic: Is it an issue to keep it? (maintenance / db performance)?

web/src/pages/account/settings/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+    try {
+      await deleteAccount.mutateAsync();
+      showSuccessToast({
+        title: "Account Deleted",
+        description: "Your account has been successfully deleted.",
+      });
+      await new Promise((resolve) => setTimeout(resolve, 2000));
+      await signOut();
+    } catch (error) {
+      console.error(error);
```
Steffen911, line 162: I would expect an error toast here?

web/src/server/api/routers/userAccount.ts
```diff
@@ -0,0 +1,124 @@
...
+    100,
+    "Name must be at most 100 characters",
+  ),
+});
+
+export const userAccountRouter = createTRPCRouter({
+  checkCanDelete: authenticatedProcedure.query(async ({ ctx }) => {
+    const userId = ctx.session.user.id;
+
+    // Find all organizations where user is a member
```
Steffen911:
```suggestion
    // Find all organizations where user is an owner
```

FroeMic:
> > the user is not part of an organization
> 
> @FroeMic Why does this have to be true? The deletion action could just remove the user from all projects and organizations if they want to delete themselves.

(1) Clarifying: In a case, where the user in not part of ANY organization, they should be able to delete themselves.
(2) Once we delete the user from the DB it cascades to OrganizationMembership and ProjectMembership (effectively removing the users there)

The checks regarding "last owner" are there to prevent stale orgs/projects without any owner.

web/src/server/api/routers/userAccount.ts
Steffen911: I think that having this check in here seems correct in itself. If we set the transaction level to serializable it should also prevent an incorrect state if organization owners are removed concurrently. 

If we keep it, we should refactor the "can be deleted check" into a dedicated function as you're duplicating the logic from above here.

FroeMic: (1) Refactored the `checkUserCanBeDeleted` into its own function for DRYness
(2) Also made the isolation level explicit

web/src/pages/account/settings/index.tsx
FroeMic: added

PR #9684, opened 2025-10-13
fix(dataset-run-item): enhance metadata conversion to handle various data types

marliessophie:
> [!IMPORTANT]
> Enhance metadata conversion in `utils.ts` to handle various data types and streamline event processing in `experimentServiceClickhouse.ts`.
> 
>   - **Metadata Conversion**:
>     - `convertPostgresJsonToMetadataRecord` in `utils.ts` now handles `number` and `boolean` types, converting them to strings.
>     - `convertRecordValuesToString` in `utils.ts` now returns a new result object instead of modifying the input record.
>   - **Event Processing**:
>     - Removed `input` and `expectedOutput` from the event body in `processItem` in `experimentServiceClickhouse.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 23db3ce27b946ed0b3f01a64815f7246f523e705. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/worker/src/services/IngestionService/utils.ts
+++ b/worker/src/services/IngestionService/utils.ts
@@ -26,21 +26,31 @@ export const convertJsonSchemaToRecord = (
 export const convertPostgresJsonToMetadataRecord = (
   metadata: Prisma.JsonValue,
 ): Record<string, string> => {
-  return typeof metadata === "string"
-    ? { metadata: metadata }
-    : Array.isArray(metadata)
-      ? { metadata: JSON.stringify(metadata) }
-      : (metadata as Record<string, string>);
+  if (
+    typeof metadata === "string" ||
+    typeof metadata === "number" ||
+    typeof metadata === "boolean"
+  ) {
+    return { metadata: String(metadata) };
+  }
+  if (Array.isArray(metadata)) {
+    return { metadata: JSON.stringify(metadata) };
+  }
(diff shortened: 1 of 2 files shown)


worker/src/services/IngestionService/utils.ts
```diff
@@ -26,21 +26,31 @@ export const convertJsonSchemaToRecord = (
...
+  if (metadata && typeof metadata === "object") {
+    return convertRecordValuesToString(metadata as Record<string, unknown>);
+  }
+  return {};
 };
 
 export const convertRecordValuesToString = (
   record: Record<string, unknown>,
 ): Record<string, string> => {
+  const result: Record<string, string> = {};
```
marliessophie, line 48: @Steffen911 removed in-place mutation of record for maintainability (as less predictable for devs) -- any objections?

Steffen911: @marliessophie Only concern could be the total size of records as metadata might be in the MegaByte range.
Should be ok though and no objection!

marliessophie: cool, will merge as is then, thank you 👍

PR #9910, opened 2025-10-21
chore(ui): Replace select with combobox for annotations

FroeMic:
## What does this PR do?

This PR replaces the `Select` component with a `Combobox` component for categorical score configurations within the annotation UI. This significantly improves the user experience when dealing with a large number of categorical entries by introducing search and filter capabilities.

CleanShot 2025-10-23 at 12 28 27@2x

CleanShot 2025-10-23 at 12 27 56@2x

----

> [!IMPORTANT]
> Replace `Select` with `Combobox` in `AnnotationForm.tsx` to enhance categorical score configuration handling with search and filter capabilities.
> 
>   - **Behavior**:
>     - Replace `Select` with `Combobox` in `AnnotationForm.tsx` for categorical score configurations.
>     - `Combobox` provides search and filter capabilities for better handling of large categorical entries.
>   - **Components**:
>     - Add new `Combobox` component in `combobox.tsx` with options for search, placeholder, and disabled states.
>     - Adjust `CommandInput` styles in `command.tsx` for better alignment and appearance.
>   - **Misc**:
>     - Remove `Select` related imports and components from `AnnotationForm.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for a3a0860de802b96ac357e2b1c2e4be1042fc38b8. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


marliessophie:
What's the rational here? If I remember correctly the annotation drawer should render combobox if more than 3 categories and toggle group otherwise. This makes it less clicks to annotate if there are few categories, which I would see as a positive. Curious to hear reasoning :)

FroeMic:
> What's the rational here? If I remember correctly the annotation drawer should render combobox if more than 3 categories and toggle group otherwise. This makes it less clicks to annotate if there are few categories, which I would see as a positive. Curious to hear reasoning :)

@marliessophie The annotation drawer rendered a select component before (no search field). Especially when navigating through the annotation queue with a keyboard, this quickly become cumbersome. So, small ux improvement by replacing it with a combobox

web/src/components/ui/command.tsx
marliessophie, line 1: Please make sure these changes scale across command usage. We should be careful when making changes to our design system components.

web/src/components/ui/combobox.tsx
marliessophie, line 1: Overall this looks good; as elipses pointed out, `option.label` is used for comparison/state instead of `option.value`. Breaks when label ≠ value. Please rely on value rather than label for all onChange operations. Also, would suggest adjusting interface slightly to enhance ease of use for devs. Currently passing value and label that are equal feels less intuitive imo. 

**Design improvement suggestion:**
```typescript
export interface ComboboxOption<T = string> {
  value: T;           // The actual data (ID, key)
  label?: string;     // Display text (defaults to value if omitted)
  disabled?: boolean;
}
```
Makes value the single source of truth for selection state, label purely presentational.

marliessophie (commented):
Quick review, overall nice improvement. One comment and one quick research task, should be quick to address.

web/src/components/ui/combobox.tsx
FroeMic: (1) I like your generic approach. Will implement.

(2) Ad filtering by label:

From my experience, users expect to filter for what they can see, not the underlying value. 

The selected value might be an ID, an raw date, a complete object, etc. When the UI shows "red" I would like to filter for "red" not for its id.

web/src/components/ui/command.tsx
FroeMic: These where some minor UI changes that made the entire command look more visually balanced. Placing the search-icon absolutely above the input and compensating with a padding-left for the input text, make keyboard selections with focus-visible also encompass the icon instead of just highlighting the input, not encompassing the icon

CleanShot 2025-10-27 at 15 26 47@2x
CleanShot 2025-10-27 at 15 27 08@2x

web/src/components/ui/combobox.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <CommandEmpty>{emptyText}</CommandEmpty>
+            <CommandGroup>
+              {options.map((option) => (
+                <CommandItem
+                  key={option.value}
+                  value={option.label}
+                  disabled={option.disabled}
+                  onSelect={() => {
+                    if (!option.disabled && onValueChange) {
+                      onValueChange(option.label === value ? "" : option.label);
```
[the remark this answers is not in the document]
FroeMic: completed

web/src/components/ui/combobox.tsx
FroeMic: @marliessophie implemented the generic interface

web/src/features/scores/components/AnnotationForm.tsx
```diff
@@ -586,43 +580,23 @@ function InnerAnnotationForm<Target extends ScoreTarget>({
...
-                                            >
-                                              {category.label}
-                                            </SelectItem>
-                                          ),
-                                        )}
-                                      </SelectContent>
-                                    </Select>
+                                      options={categories.map((category) => ({
+                                        value: category.label,
+                                        label: category.label,
```
marliessophie: _Nit_: we can remove label here and just pass value.

web/src/components/ui/combobox.tsx
```diff
@@ -0,0 +1,138 @@
...
+                >
+                  <Check
+                    className={cn(
+                      "mr-2 h-4 w-4",
+                      isEqual(value as T | undefined, option.value)
+                        ? "opacity-100"
+                        : "opacity-0",
+                    )}
+                  />
+                  {option.label}
```
marliessophie: We need to fallback on String(option.value) if label is not passed.

marliessophie (commented):
Tiny comment then good to go! Nice combobox implementation.

web/src/features/scores/components/AnnotationForm.tsx
FroeMic: done

web/src/components/ui/combobox.tsx
FroeMic: done

PR #9928, opened 2025-10-22
feat(comments): Support @mentions in comments

FroeMic:
# What' New?

Users are now able to tag other users in the same workspace with @ mentions. 

https://github.com/user-attachments/assets/d64e84c7-267b-4039-a2bc-254fd1f33aa7

###  Features

  - Type @ to trigger autocomplete for project members
  - Keyboard navigation (arrows, Enter/Tab, Escape)
  - CMD+Enter to submit
  - Mentions render as blue badges
  - Shows first 2 matches + "and X more..."

### Permissions

  - Requires BOTH comments:CUD AND projectMembers:read
  - Enforced on client and server
  - No member list leaking without proper access

####  Database
  - New comment_mentions table
  - Soft FK to preserve mentions when users deleted

####  Backend
  - Updated create, getByObjectId endpoints
  - New searchTaggableUsers endpoint

####  Frontend
  - MentionAutocomplete component (shadcn Command)
  - useMentionAutocomplete hook for @ detection
  - Markdown parser handles @[Name](user:id) syntax

####  Escape Key Hack
 Vaul drawer listens for Escape at document level. Event propagation solutions didn't work, so we use controlled state to intercept close attempts:

```
  <Drawer open={isDrawerOpen} onOpenChange={(open) => {
    if (!open && isMentionDropdownOpen) return; // Prevent close
    setIsDrawerOpen(open);
  }}>
```

Not the cleanest solution but works reliably.

 ### Migration

  Run: `pnpm --filter=@langfuse/shared run db:migrate`

----

> [!IMPORTANT]
> Adds @mention support in comments with backend and frontend changes, including autocomplete and rendering as badges.
> 
>   - **Behavior**:
>     - Users can tag others with @mentions in comments, triggering autocomplete for project members.
>     - Mentions render as blue badges and show first 2 matches with "and X more...".
>     - Requires `comments:CUD` and `projectMembers:read` permissions.
>   - **Backend**:
>     - Adds `comment_mentions` table with soft FK.
>     - Updates `create` and `getByObjectId` endpoints in `comments.ts`.
>     - New `searchTaggableUsers` endpoint.
>   - **Frontend**:
>     - `MentionAutocomplete` component and `useMentionAutocomplete` hook for @ detection.
>     - `MarkdownViewer.tsx` updated to render mentions as badges.
>     - `CommentList.tsx` and `CommentDrawerButton.tsx` handle mention dropdown state.
>   - **Tests**:
>     - Adds `mentionParser.test.ts` for `extractUniqueMentionedUserIds` and `sanitizeMentions` functions.
>     - Tests in `comments-api.servertest.ts` ensure public API does not process mentions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 2f5eb259806a457df30de6d6192741b7c248cf7e. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/comments/lib/mentionParser.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export function extractMentionsFromMarkdown(content: string): ParsedMention[] {
+  const mentions: ParsedMention[] = [];
+  const seen = new Set<string>();
+
+  // Reset regex lastIndex
+  MENTION_REGEX.lastIndex = 0;
+
+  let match;
+  while ((match = MENTION_REGEX.exec(content)) !== null) {
```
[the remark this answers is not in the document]
FroeMic: good catch! resolved

web/src/server/api/routers/comments.ts
```diff
@@ -304,4 +382,56 @@ export const commentsRouter = createTRPCRouter({
...
+        // Require BOTH permissions
+        throwIfNoProjectAccess({
+          session: ctx.session,
+          projectId: input.projectId,
+          scope: "comments:CUD",
+        });
+        throwIfNoProjectAccess({
+          session: ctx.session,
+          projectId: input.projectId,
+          scope: "projectMembers:read",
```
maxdeichmann: Please check if you use this check everywhere where you use getUserProjectRoles. Maybe even adding this inside getUserProjectRoles may even be preferred. Risk: on the demo project people could create a comment and iterate through users and send them emails.

web/src/components/ui/MarkdownViewer.tsx
```diff
@@ -139,6 +140,16 @@ function MarkdownRenderer({
               );
             },
             a({ children, href }) {
+              // Handle mention links
+              if (href?.startsWith("user:")) {
```
marliessophie: Please extract magic string to a constant.

web/src/features/comments/components/MentionAutocomplete.tsx
```diff
@@ -0,0 +1,126 @@
...
+interface MentionAutocompleteProps {
+  users: User[];
+  isLoading: boolean;
+  selectedIndex: number;
+  onSelect: (userId: string, displayName: string) => void;
+  onClose: () => void;
+  onSelectedIndexChange: (index: number) => void;
+}
+
+export function MentionAutocomplete({
```
marliessophie, line 27: We already have user search with autocomplete for annotation queue assignments. Reuse/abstract shared logic from existing annotation assignment logic (checks access, RBAC, supports search, exclusion, pagination) instead of creating duplicate `searchTaggableUsers`.

See [UserAssignmentSection](web/src/features/annotation-queues/components/UserAssignmentSection.tsx) and [useUserSearch hook](web/src/features/annotation-queues/hooks/useUserSearch.ts) for reuse pattern.

**Recommendation**:
- Keep your mention UI (inline @ trigger, dropdown) - looks great
- Remove `searchTaggableUsers` endpoint
- Consider abstracting to shared hook if needed

Image

packages/shared/prisma/generated/types.ts
```diff
@@ -352,6 +352,12 @@ export type Comment = {
   content: string;
   author_user_id: string | null;
 };
+export type CommentMention = {
```
marliessophie: **Critical:** comment mention table has no `project_id`, we absolutely must add this to every table, otherwise our security/RBAC will not work and we expose project specific user_id across projects. 

**Generally:** Data model is comment-specific now (`comment_mentions` table). My understanding from our discussion yesterday was that we will want to have notifications more broadly (eg for annotation queue assignments) across the platform. Should this be correct, I'd suggest to consider a generic `notifications` table with `object_type` + `object_id` pattern (used elsewhere: annotation queues, comments). Would enable mentions on other objects. Structure:
- Rename to `notifications`
- `mentioned_user_id` → `user_id`
- Add `type` enum: 'mention', (extensible later)
- Reuse existing object type enums where possible

Generally my personal preference is to loop on data model in Linear first before implementation. I have found this to be quite effective. Easy to tag relevant people and get buy-in on decisions quickly. Just a personal preference of course, you do you!

web/src/features/comments/lib/mentionParser.ts
marliessophie, line 1: Add unit tests for mention parser (`mentionParser.ts`). Critical for security - validates regex, sanitization, social engineering prevention.

web/src/pages/api/public/comments/index.ts
marliessophie, line 1: Question: Why support mentions via public API? Would love to understand your rational for adding this before reviewing in detail. 

I do have a couple concerns of not (initially) making this a UI only feature. I am a little worried about users being able to pass plain names and us resolving userIds from this via API. Maybe I'm being too paranoid here but the way I see it I could send lots of API requests creating comments if I have project comment access (only), and then just loop over all kinds of plain names and this way find out who else has access to the project. 

Plus adding later seems like a two way door to me. Should we not support via UI, we could just not resolve the comment as a mention, or strip the magic "@"; no super strong take here. 

UI: "Charmant" discovery: 
- User types `@` → autocomplete → select → notify
- Public API is programmatic, not interactive
- No public API endpoint exposes project members (cannot discover who to mention)

Current implementation: [index.ts:36-60](web/src/pages/api/public/comments/index.ts#L36-L60) processes mentions, validates against all project members.

Security concern: Implicit permission escalation
- Public API has `comments:CUD` but not `projectMembers:read`
- Mention validation grants implicit member access
- Creates enumeration vector

web/src/server/api/routers/comments.ts
```diff
@@ -38,22 +45,76 @@ export const commentsRouter = createTRPCRouter({
           });
         }
 
-        const comment = await ctx.prisma.comment.create({
-          data: {
+        // Extract mentions from content (server-side, authoritative)
+        const mentionsInContent = extractMentionsFromMarkdown(input.content);
```
marliessophie: Couple of points: 

Question: Unused `mentionedUserIds` in Schema

`mentionedUserIds` in `CreateCommentData` schema unused - client sends, server ignores, extracts from content instead. This is confusing. Choose one:

- **Server-side only**: remove from schema 
- **Use client hints**: validate client IDs match extracted

Suggestion: Reduce Overfetching Project Members

`getUserProjectRoles` with empty filters fetches **all project members**. Pass user ID filter:

```typescript
filterCondition: [
  {
    column: "userId",
    operator: "any of",
    value: mentionsInContent.map((m) => m.userId),
    type: "stringOptions",
  },
];
```

Benefits:

- Reduces DB load
- Validates only mentioned users exist
- More efficient query

web/src/server/api/routers/comments.ts
```diff
@@ -151,28 +212,45 @@ export const commentsRouter = createTRPCRouter({
...
+          u.image AS "authorUserImage",
+          u.name AS "authorUserName",
+          COALESCE(
+            json_agg(cm.mentioned_user_id) FILTER (WHERE cm.mentioned_user_id IS NOT NULL),
+            '[]'
+          )::text as "mentionedUserIds"
         FROM comments c
         LEFT JOIN users u ON u.id = c.author_user_id AND u.id in (SELECT user_id FROM organization_memberships WHERE org_id = ${ctx.session.orgId})
-        WHERE 
+        LEFT JOIN comment_mentions cm ON cm.comment_id = c.id
```
marliessophie: **Critical:** you **must** always include `project_id` in the join condition. We can never forget this, otherwise we might run risk of exposing user data to other projects.

web/src/server/api/routers/comments.ts
```diff
@@ -304,4 +382,56 @@ export const commentsRouter = createTRPCRouter({
         });
       }
     }),
+  searchTaggableUsers: protectedProjectProcedure
```
marliessophie: Given improvement suggestions around re-using annotation assignment logic I think we might be able to drop this route entirely.

web/src/server/api/routers/comments.ts
marliessophie: Agree with Max.

marliessophie (commented):
Overall: Great UI! Excited for this feature. A couple open points around data model, API design and some refactor suggestions to reuse existing code. Feel free to huddle should you have questions.

web/src/components/ui/MarkdownViewer.tsx
FroeMic: done

web/src/features/comments/components/MentionAutocomplete.tsx
FroeMic: Refactored to use existing ` api.members.byProjectId` trpc route

-> Keeping open until you resolve

web/src/features/comments/lib/mentionParser.ts
FroeMic: Added unit tests

web/src/pages/api/public/comments/index.ts
```diff
@@ -27,12 +33,57 @@ export default withMiddlewares({
...
-          id: v4(),
+      // Extract mentions from content (server-side, authoritative)
+      const mentionsInContent = extractMentionsFromMarkdown(body.content);
+
+      // Sanitize mentions and get valid user IDs
+      let sanitizedContent = body.content;
+      let validMentionedUserIds: string[] = [];
+
+      if (mentionsInContent.length > 0) {
+        // Fetch project members for validation and normalization
```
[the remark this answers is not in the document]
FroeMic: remove public api

web/src/pages/api/public/comments/index.ts
FroeMic: removed public api and created an issue with the tradeoffs from this discussion for any future implementation: https://linear.app/langfuse/issue/LF-1909/comments-public-api-support-for-mentions

packages/shared/prisma/generated/types.ts
FroeMic: Agreed! I think it is a valid point. 

I have completely remove the persistence of comment mentions. If/When we build the Notification Inbox we can properly implement it then.

web/src/server/api/routers/comments.ts
FroeMic: Removed the comments-mention table

web/src/server/api/routers/comments.ts
FroeMic: dropped this route in favor of reusing the `getUserProjectRoles`

web/src/server/api/routers/comments.ts
FroeMic: We now only use it once in the `commentCreateRoute` with a `throwIfNoProjectAccess(scope: "projectMembers:read")` check before.

The other occurrence was replace by using the already existing ` api.members.byProjectId` query from the `membersRouter` which also checks: `throwIfNoProjectAccess(scope: "projectMembers:read")` 

As such the "read" change will be concentrated there.

web/src/server/api/routers/comments.ts
FroeMic: Great point!

-> Server-side extraction; removed mentionedUserIds 

-> added filters as suggested

packages/shared/package.json
```diff
@@ -25,6 +25,10 @@
       "import": "./dist/src/server/auth/apiKeys.js",
       "require": "./dist/src/server/auth/apiKeys.js"
     },
+    "./src/features/comments/mentionParser": {
```
marliessophie: Why do we need this export here? My understanding is this file holds only exports for cross-package usage of general purpose functions (eg encryption, API keys). Comment specific logic should not be included here.

web/src/server/api/routers/comments.ts
```diff
@@ -53,7 +105,10 @@ export const commentsRouter = createTRPCRouter({
           resourceType: "comment",
           resourceId: comment.id,
           action: "create",
-          after: comment,
+          after: {
+            ...comment,
+            mentionedUserIds: validMentionedUserIds,
```
marliessophie: Let's please revert this change; we no longer save`mentionedUserIds` on the schema. Audit log should record shape of object as it is written; we are deviating from that.

web/src/server/api/routers/comments.ts
```diff
@@ -38,10 +45,55 @@ export const commentsRouter = createTRPCRouter({
           });
         }
 
+        // Extract mentions from content (server-side, authoritative)
+        const mentionsInContent = extractMentionsFromMarkdown(input.content);
+
+        // Sanitize mentions and get valid user IDs
+        let sanitizedContent = input.content;
+        let validMentionedUserIds: string[] = [];
```
marliessophie: Let's drop `validMentionedUserIds`, it's not used.

packages/shared/src/features/comments/mentionParser.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  return mentions;
+}
+
+/**
+ * Extract just user IDs for simpler validation
+ */
+export function extractMentionedUserIds(content: string): string[] {
```
marliessophie: This is dead code, it's never used. Please remove.

web/src/server/api/routers/comments.ts
```diff
@@ -38,10 +45,55 @@ export const commentsRouter = createTRPCRouter({
...
+          // Fetch only the mentioned users - more efficient than fetching all project members
+          const projectMembers = await getUserProjectRoles({
+            projectId: input.projectId,
+            orgId: ctx.session.orgId,
+            searchFilter: Prisma.empty,
+            filterCondition: [
+              {
+                column: "userId",
+                operator: "any of",
+                value: Array.from(
```
marliessophie: We can skip deduplication, `extractMentionsFromMarkdown` already returns unique userIds.

packages/shared/src/features/comments/mentionParser.ts
```diff
@@ -0,0 +1,170 @@
...
+  userId: string;
+  displayName: string;
+}
+
+/**
+ * Extract mentions from markdown content
+ * @param content - Markdown string with embedded mentions
+ * @returns Array of unique mentioned users
+ */
+export function extractMentionsFromMarkdown(content: string): ParsedMention[] {
```
marliessophie: Couple thoughts: 
- **Deduplication:** Function already deduplicates, but name doesn't reflect this. Rename to `extractUniqueMentionedUserIds` or similar. Nit: display name is never used, can simply return array of uniqueIds. 
- **Potential ReDoS Vulnerability:** Global regex with `exec()` in loop maintains state via `lastIndex`. While bounded quantifiers prevent catastrophic backtracking, stateful regex can behave unexpectedly on malformed input. Current code resets `lastIndex` twice (lines 34, 117) which suggests awareness of this issue. Consider consolidating. _Note: code flagged this, I am no expert here._

web/src/features/comments/lib/mentionParser.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @example
+ * // Valid mention gets normalized display name
+ * sanitizeMentions("Hey @[FakeAdmin](user:alice123)", [{id: "alice123", name: "Alice Smith", email: "..."}])
+ * // Returns: { sanitizedContent: "Hey @[Alice Smith](user:alice123)", validMentionedUserIds: ["alice123"] }
+ *
+ * // Invalid mention gets stripped to plain text
+ * sanitizeMentions("Hey @[Someone](user:invalid)", [{id: "alice123", name: "Alice Smith", email: "..."}])
+ * // Returns: { sanitizedContent: "Hey Someone", validMentionedUserIds: [] }
+ */
+export function sanitizeMentions(
```
marliessophie, line 86: This feels overcomplicated. 

**Current approach:**
1. Extracts mentions with indices
2. Reverses array (why?)
3. Loops with string concatenation (O(n×m) where m = content length)
4. Replaces display names with DB names (unnecessary?)
5. Reverses validUserIds again

**Questions:**
- Why replace display names? Client already sent correct name. When would `userId` be valid but `displayName` wrong?
- Why not partition valid/invalid mentions, then single replace pass?
- Why preserve order of validMentionedUserIds? Only used for audit logging (comments.ts:112) where order doesn't matter.

Unless I am missing something, we could just do something like: 
```ts
export function sanitizeMentions(
  content: string,
  projectMembers: ProjectMember[],
): SanitizeMentionsResult {
  const validUserIds = projectMembers.map(({id)} => id); // or could just pass list of validUserIds
  const mentioned = new Set<string>();

  const sanitized = content.replace(
    MENTION_REGEX,
    (match, displayName, userId) => {
      if (validUserIds.has(userId)) {
        mentioned.add(userId);
        return match; // Keep mention as-is
      }
      return displayName; // Strip markdown, keep text
    }
  );

  return {
    sanitizedContent: sanitized,
    validMentionedUserIds: [...mentioned]
  };
}
```

web/src/features/comments/hooks/useMentionAutocomplete.ts
marliessophie, line 1: Suggestion: Please reuseuseUserSearch, lines 21-39 are almost identical. Access check is also missing from this hook, we cannot rely on people to pass access check via "enabled" prop.

web/src/features/comments/CommentList.tsx
```diff
@@ -38,16 +49,20 @@ export function CommentList({
...
   objectType: CommentObjectType;
   cardView?: boolean;
   className?: string;
   onDraftChange?: (hasDraft: boolean) => void;
+  onMentionDropdownChange?: (isOpen: boolean) => void;
 }) {
   const session = useSession();
   const [textareaKey, setTextareaKey] = useState(0);
+  const [cursorPosition, setCursorPosition] = useState(0);
+  const [textareaValue, setTextareaValue] = useState("");
```
marliessophie: Why duplicate content state management? React hook form already tracks this for us?

web/src/features/comments/CommentList.tsx
```diff
@@ -170,9 +245,50 @@ export function CommentList({
   }
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLTextAreaElement>) => {
```
marliessophie, line 368: Can we simplify this function? Looks rather complex for what it does.

marliessophie (commented):
Left some comments, pls lmk in case anything is not clear.

packages/shared/package.json
FroeMic: We need this to run the mentionParser test.

The shared package has not testing setup, so (discussed with Nimar) the easiest way to test packages there is to do so in the worker package.

Without this line, the import in the testing setup fails.

web/src/server/api/routers/comments.ts
FroeMic: ok!

web/src/features/comments/lib/mentionParser.ts
FroeMic: ### Questions:

  1. **"Why reverse?" -** To maintain correct string indices. When you replace text, indices shift. By processing  from end to start, later replacements don't affect earlier indices.
  2. **"Why replace display names?"** - Security concern: Social engineering prevention. A malicious user could send:  @[SYSTEM ADMINISTRATOR](user:normalUserId).
    3. **"Why preserve order?"** - You're right that for audit logging, order probably doesn't matter. This is likely over-engineering.

packages/shared/src/features/comments/mentionParser.ts
FroeMic: renamed!

FroeMic: ### Regex ReDos:

Had it security-reviewed by claude. The explanation matches my understanding of the Regex. Not an export though either.

 **1. ReDoS Vulnerability & lastIndex Management**

  The concern here is valid but perhaps overstated. Let me break down what's actually happening:

  **The Issue:**
  - Global regex (/pattern/g) with exec() maintains state via lastIndex
  - If lastIndex isn't reset between uses, subsequent calls can start searching from unexpected positions
  - This can lead to bugs where the same regex used on different strings produces inconsistent results

  **Why This Isn't a ReDoS Vulnerability:**
  The reviewer's note says "code flagged this" - likely from a static analysis tool. However, this is not
  actually a ReDoS (Regular Expression Denial of Service) vulnerability because:

  1. Bounded quantifiers: {1,100} and {1,30} prevent catastrophic backtracking
  2. No nested quantifiers: Pattern like (a+)+ would be vulnerable, but we have [^[\\]]{1,100} which is safe
  3. Character class negation: [^[\\]] matches exactly one character that isn't [ or ] - no backtracking
  possible

  **Why Reset Twice:**
  // extractUniqueMentionedUserIds
  MENTION_REGEX.lastIndex = 0;  // Line 33

  // sanitizeMentions
  MENTION_REGEX.lastIndex = 0;  // Line 107

  This is actually correct and necessary because:
  - MENTION_REGEX is a module-level singleton shared across all function calls
  - If Function A uses the regex, lastIndex is left at some position
  - When Function B runs next, it would start searching from that position instead of 0
  - Each function needs to reset it to ensure deterministic behavior

  **Is This a Problem?**
  Not really - it's defensive programming. The "consolidation" suggestion would mean creating the regex fresh
  each time or using a factory function, which has tradeoffs:

  // Option 1: Factory (overkill for this)
  function getMentionRegex() {
    return new RegExp(...);  // New instance each time
  }

  // Option 2: String.replace with regex (stateless)
  content.replace(MENTION_REGEX, ...)  // No lastIndex issues

  // Current approach: Reset before each use
  MENTION_REGEX.lastIndex = 0;  // Clear, explicit, fast

  The current approach is fine. The resets are in different functions that may be called independently, so both
   are needed.

web/src/server/api/routers/comments.ts
FroeMic: removed

web/src/server/api/routers/comments.ts
FroeMic: removed

packages/shared/src/features/comments/mentionParser.ts
FroeMic: removed

web/src/server/api/routers/comments.ts
FroeMic: removed

web/src/features/comments/CommentList.tsx
FroeMic: Not really – this is essential for the autocomplete popover navigation to work.

- Type "@ mich", arrow-up, arrow-up, enter to select the third suggestion in the @mention popover.

web/src/features/comments/CommentList.tsx
FroeMic: done

web/src/features/comments/lib/mentionParser.ts
FroeMic: added simplified version

web/src/features/comments/hooks/useMentionAutocomplete.ts
FroeMic: added a version re-using useUserResearch

web/src/features/comments/lib/mentionParser.ts
```diff
@@ -0,0 +1,129 @@
...
+  `@\\[([^[\\]]{1,100})\\]\\(${MENTION_USER_PREFIX}([a-z0-9_-]{1,30})\\)`,
+  "gi",
+);
+
+/**
+ * Extract unique mentioned user IDs from markdown content
+ * @param content - Markdown string with embedded mentions in format @[Display Name](user:userId)
+ * @returns Array of unique user IDs (deduplicated, order preserved, max 50 mentions)
+ */
+export function extractUniqueMentionedUserIds(content: string): string[] {
```
marliessophie, line 28: _Nit:_ order is not required, could drop duplicate tracking of userIds

marliessophie (approved):
PENDING: please fix broken tests/comment mention API implementation. Once this is fixed, we should be able to merge. 

Otherwise: Code looks good now. Left one nit, but feel free to skip; non-blocking.

web/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "declaration": false,
     "declarationMap": false,
     "outDir": "dist",
+    "rootDir": ".",
```
marliessophie: Please remove this.

FroeMic: Removed. 

As observational sidenote: both the tsConfig in `/shared/package` and `/ee` have the rootDir property configured (not changed)

PR #9951, opened 2025-10-22
feat(comments): Support send email notifications on comment mention

FroeMic:
# What's New?

Implements email notifications when users are @mentioned in comments.

CleanShot 2025-10-23 at 00 05 54

CleanShot 2025-10-23 at 00 25 49@2x

  ### Summary

  - ✅ Email notifications when mentioned in comments (traces, observations, prompts, sessions)
  - ✅ Async queue-based delivery (BullMQ/Redis, non-blocking)
  - ✅ Per-user, per-project notification preferences
  - ✅ Settings UI for managing notifications
  - ✅ Deep linking from emails to specific comments

  ### Key Changes

  **Database**
  - Added `NotificationPreference` model with `EMAIL` channel and `COMMENT_MENTION` type
  - Default: notifications enabled for all users

  **Worker**
  - Notification queue processor with comment mention handler
  - Checks user preferences, fetches data from Prisma/ClickHouse
  - Sends emails via nodemailer with React Email template
  - Smart URL construction:
    - Observations: Link to trace page with `?observation={id}`
    - Prompts: Use URL-encoded name + version
    - Traces/Sessions: Standard URL pattern

  **Frontend**
  - Notification settings page with email toggle
  - Auto-open comment drawer from deep links
  - Fixed duplicate drawer rendering (conditional render with `useIsMobile`)
  - Fixed drawer reopening after manual close (ref-based flag)

  **API**
  - Comments router enqueues notification jobs
  - Notification preferences router for settings management

  ### Configuration

  Existing SMTP required: `EMAIL_FROM_ADDRESS`, `SMTP_CONNECTION_URL`, `NEXTAUTH_URL`

  New optional env var: `QUEUE_CONSUMER_NOTIFICATION_QUEUE_IS_ENABLED` (default: `true`)

----

> [!IMPORTANT]
> Adds email notifications for comment mentions with async processing, user preferences, and frontend management.
> 
>   - **Behavior**:
>     - Email notifications for @mentions in comments across various object types (traces, observations, prompts, sessions).
>     - Asynchronous delivery using BullMQ/Redis.
>     - Per-user, per-project notification preferences with a UI for management.
>     - Deep linking from emails to specific comments.
>   - **Database**:
>     - Added `NotificationPreference` model for managing notification settings.
>     - Default settings enable notifications for all users.
>   - **Backend**:
>     - `notificationQueueProcessor` in `notificationQueue.ts` processes comment mention notifications.
>     - `handleCommentMentionNotification` in `commentMentionHandler.ts` constructs and sends emails.
>     - `commentsRouter` and `notificationPreferencesRouter` handle comment creation and notification preference management.
>   - **Frontend**:
>     - `NotificationSettings` component for managing email notification preferences.
>     - `CommentDrawerButton` auto-opens for deep links and manages comment interactions.
>     - `PromptDetail` and `PromptHistoryNode` updated to display comment counts and handle comment interactions.
>   - **Misc**:
>     - Added `stripMarkdown` utility in `markdown.ts` to clean comment previews.
>     - Updated `tailwind.config.ts` to include emoji picker styles.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f4b8c12ba28b7a71cb89cf9bd0ff570526a41929. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/components/trace/index.tsx
marliessophie, line 1: Do changes in this file have something to do with notifications?

packages/shared/prisma/schema.prisma
```diff
@@ -699,6 +700,34 @@ enum CommentObjectType {
...
+}
+
+enum NotificationType {
+  COMMENT_MENTION
+  // Extend by adding: COMMENT_REPLY, COMMENT_NEW, EVAL_COMPLETE, EXPORT_READY
+}
+
+model NotificationPreference {
+  id        String              @id @default(cuid())
+  userId    String              @map("user_id")
```
marliessophie: We are missing a reference to the users table. Generally I would expect the following to be true:

- If a project is deleted: all notification preferences referencing project are deleted
- If a user is deleted: notification preference referencing user are deleted
- If a project/org_memebership is deleted:  notification preference referencing user, project_id combination is deleted

Follow up question: can emails be changed? If yes how does this flow through the data model? If an email is changed we should also consider killing the queued email job. 

In ALL of the above cases we should no longer send emails. We should cancel all jobs with given user_id, project_id combination. Up to you how you solve this, but it's critical we get the resulting behaviour right. It would break trust with our users if we eg send an email to a user who has been removed from the project.

packages/shared/prisma/schema.prisma
```diff
@@ -699,6 +700,34 @@ enum CommentObjectType {
...
+
+  channel NotificationChannel
+  type    NotificationType
+  enabled Boolean             @default(true)
+
+  createdAt DateTime @default(now()) @map("created_at")
+  updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
+
+  @@unique([userId, projectId, channel, type])
+  @@index([userId, projectId])
```
marliessophie: Index is redundant and can be removed. In PostgreSQL, the unique index on `[userId, projectId, channel, type]` can be used efficiently for queries that filter by the leftmost columns due to left-prefix matching.

packages/shared/src/server/services/email/commentMention/sendCommentMentionEmail.ts
```diff
@@ -0,0 +1,66 @@
...
+      from: {
+        address: env.EMAIL_FROM_ADDRESS,
+        name: "Langfuse",
+      },
+      subject: `${authorName} mentioned you in project ${projectName}`,
+      html: htmlTemplate,
+    });
+
+    logger.info(
+      `Comment mention email sent to ${mentionedUserEmail} for comment in project ${projectName}`,
```
marliessophie: **Critical:** Please be careful **_never_** to log PII, this creates compliance issues. We cannot log user email addresses ever, and even project names might be sensitive.

packages/shared/src/server/services/email/commentMention/sendCommentMentionEmail.ts
```diff
@@ -0,0 +1,66 @@
...
+      }),
+    );
+
+    await mailer.sendMail({
+      to: mentionedUserEmail,
+      from: {
+        address: env.EMAIL_FROM_ADDRESS,
+        name: "Langfuse",
+      },
+      subject: `${authorName} mentioned you in project ${projectName}`,
```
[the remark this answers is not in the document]
marliessophie: Did you check if Nodemailer sanitizes this? Also I briefly checked that user name is validated to not contain html elements, are we sure this was always the case?

web/src/features/comments/CommentList.tsx
```diff
@@ -60,11 +61,23 @@ export function CommentList({
...
   const session = useSession();
+  const router = useRouter();
   const [textareaKey, setTextareaKey] = useState(0);
   const [cursorPosition, setCursorPosition] = useState(0);
   const [textareaValue, setTextareaValue] = useState("");
   const commentsContainerRef = useRef<HTMLDivElement>(null);
   const textareaRef = useRef<HTMLTextAreaElement | null>(null);
+
+  // Extract comment ID from hash for highlighting
+  const highlightedCommentId = useMemo(() => {
```
marliessophie, line 79: This seems complex, what purpose does it serve; do we really need this?

web/src/features/notifications/components/NotificationSettings.tsx
```diff
@@ -0,0 +1,128 @@
...
+    { enabled: Boolean(projectId) },
+  );
+
+  const updatePreference = api.notificationPreferences.update.useMutation({
+    onSuccess: () => {
+      refetch();
+    },
+  });
+
+  const handleToggle = async (
```
marliessophie: Nit: this can just take enabled boolean as parameter.

packages/shared/prisma/migrations/20251022143653_add_notification_preferences/migration.sql
marliessophie, line 1: Will review once suggested changes to data model addressed

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+          mentionedUserEmail: mentionedUser.email,
+          authorName,
+          projectName: comment.project.name,
+          commentPreview,
+          commentLink,
+          settingsLink,
+        });
+
+        logger.info(
+          `Comment mention email sent to ${mentionedUser.email} for comment ${commentId}`,
```
marliessophie: **Critical**: another PII log, please be really careful never to log PII. We can never log any emails.

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+const GetPreferencesInput = z.object({
+  projectId: z.string(),
+});
+
+export const notificationPreferencesRouter = createTRPCRouter({
+  getForProject: protectedProjectProcedure
+    .input(GetPreferencesInput)
+    .query(async ({ input, ctx }) => {
+      try {
+        throwIfNoProjectAccess({
```
marliessophie: Please also add access check to front-end where route is called.

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+                ? preferencesMap.get("EMAIL:COMMENT_MENTION")!
+                : true, // Default: enabled
+          },
+          // Future preferences can be added here
+          // {
+          //   channel: "EMAIL" as const,
+          //   type: "COMMENT_REPLY" as const,
+          //   enabled: preferencesMap.get("EMAIL:COMMENT_REPLY") ?? true,
+          // },
+        ];
```
marliessophie: this feels overly complex, why not just do something like: 
```
const preferences = await ctx.prisma.notificationPreference.find({
          where: {
            userId,
            projectId: input.projectId,
            channel: "EMAIL",
            type: "COMMENT_MENTION"
          },
        });

const allPreferences = [
          {
            channel: "EMAIL" as const,
            type: "COMMENT_MENTION" as const,
            enabled: preferences.enabled, 
          },
],
return allPreferences;

```

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+        logger.error("Failed to fetch notification preferences", error);
+        if (error instanceof TRPCError) {
+          throw error;
+        }
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to fetch notification preferences.",
+        });
+      }
+    }),
```
marliessophie: You can just remove this we have global trpc handling in middleware :) It will also send proper trpc error toast.

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+            userId_projectId_channel_type: {
+              userId,
+              projectId: input.projectId,
+              channel: input.channel,
+              type: input.type,
+            },
+          },
+          update: {
+            enabled: input.enabled,
+            updatedAt: new Date(),
```
marliessophie: will be auto-populated :)

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+          create: {
+            userId,
+            projectId: input.projectId,
+            channel: input.channel,
+            type: input.type,
+            enabled: input.enabled,
+          },
+        });
+
+        logger.info(
```
marliessophie: rather than logging this, let's please create an audit log. See other trpc routes for examples.

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,136 @@
...
+      } catch (error) {
+        logger.error("Failed to update notification preference", error);
+        if (error instanceof TRPCError) {
+          throw error;
+        }
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to update notification preference.",
+        });
+      }
```
marliessophie: See comment above, can be removed.

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+  const { commentId, projectId, mentionedUserIds } = payload;
+
+  logger.info(
+    `Processing comment mention notification for comment ${commentId} in project ${projectId}`,
+  );
+
+  try {
+    // Fetch comment details with author and project information
+    const comment = await prisma.comment.findUnique({
+      where: { id: commentId },
```
marliessophie: **Critical**: all queries must always include projectId. Otherwise we will expose user data cross projects. The join of project makes this even more dangerous as this might expose another projects comment AND  project name.

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (!comment) {
+      logger.warn(
+        `Comment ${commentId} not found. Skipping notification processing.`,
+      );
+      return;
+    }
+
+    // Fetch author information
+    const author = comment.authorUserId
+      ? await prisma.user.findUnique({
```
marliessophie: **Critical**: As the lookup from line 26 likely returns comment from another project we are now finding user from another project too. To be extra safe, good practice to join org memberships -> organisation.projects where some id is equal to projectId.   

Note: IMO Ideally we just skip the author user name/email and start off with just sending an email to the user that they were mentioned, but not by whom. I think this is fine.

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+            where: {
+              orgId_userId: {
+                orgId: project.orgId,
+                userId,
+              },
+            },
+          });
+
+          if (!orgMembership) {
+            logger.info(
```
marliessophie: This is core logic, can we please extract/reuse logic in `members.byProjectId` to not reimplement this? Could just have helper with optional filter condition and return array of users and total count; we could pass mentionedUserIds as filter condition to check only for these ids.

marliessophie (changes requested):
Adding review. Please pay special attention to security comments.

web/src/components/trace/index.tsx
FroeMic: Yes – previously the traces page would mount two drawer components, since we would only hide the mobile one. 

When auto-opening via URL params, this would cause two drawers to open.

FroeMic: Essentially this change replaces "md:hidden" with isMobile() to toggle between sheet and drawer

packages/shared/src/server/services/email/commentMention/sendCommentMentionEmail.ts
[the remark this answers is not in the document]
FroeMic: Could you clarify what you mean by that?

web/src/features/comments/CommentList.tsx
FroeMic: See screenshot in PR:

- It adds a blue board around the selected comment
- I think it is a useful UX improvement, in situations where there are a lot of comments on an object

packages/shared/prisma/schema.prisma
FroeMic: ### Updated schema as proposed.

- Added FK constraint to users table with CASCADE delete
- added User relation, removed redundant index (moved to
  separate CONCURRENTLY migration)
- and enhanced comment in notification handler to clarify cascade delete behavior
 
All deletion scenarios now properly handled.
- no job cancellation neccessary
- defensive checks in handler don't send email if user / project doesn't exist

FroeMic: ### Email Change Handling

  - No special handling needed - NotificationPreference references stable userId, not email
  - Emails are resolved from User.email at send-time

packages/shared/prisma/schema.prisma
FroeMic: Removed ` @@index([userId, projectId])` index

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: Refactored to log userId instead

packages/shared/src/server/services/email/commentMention/sendCommentMentionEmail.ts
FroeMic: Refactored to log userId instead

packages/shared/src/server/services/email/commentMention/sendCommentMentionEmail.ts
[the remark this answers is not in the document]
FroeMic: **Generally, these names should be safe:**

```
  // In userAccount.ts line 11-16
  const updateDisplayNameSchema = z.object({
    name: StringNoHTML.min(1, "Name cannot be empty").max(100, "Name must be at most 100 characters"),
  });

  // StringNoHTML definition (line 91-93 in zod.ts):
  export const StringNoHTML = z.string().refine((val) => !htmlRegex.test(val), {
    message: "Text cannot contain HTML tags",
  });
```

But I now extracted a function to sanitize all fields

web/src/server/api/routers/notificationPreferences.ts
FroeMic: simplified based on your suggestion

web/src/features/notifications/components/NotificationSettings.tsx
FroeMic: updated

web/src/server/api/routers/notificationPreferences.ts
FroeMic: added

web/src/server/api/routers/notificationPreferences.ts
FroeMic: removed

web/src/server/api/routers/notificationPreferences.ts
FroeMic: removed

web/src/server/api/routers/notificationPreferences.ts
FroeMic: added auditLog

web/src/server/api/routers/notificationPreferences.ts
FroeMic: removed

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: I disagree with this notion.

I understand the general defensiveness regarding PII, but for a comment tag to be useful in a work collaboration context, it makes a big difference who tagged a user: your boss vs the intern is a substantial difference.

I don't think we should degrade value for the user

FroeMic: As for "who mentioned" – I think we can rely on the write operation to check that the tagging user has the right to tag said user. 

Consequently we should be in the clear here

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: Good point! Have updated it

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: I extracted the logic into the shared packages. This commit is maybe worth an extra look, since a bunch of files have changed (mostly imports though).

https://github.com/langfuse/langfuse/pull/9951/commits/95fba2a7579ebc10e012655f858e0f33a1915b38

dahnny012:
How would it work for self host folks? Generally i've used simple email service(SES in AWS) to send emails on behalf of a domain i own.

FroeMic:
> How would it work for self host folks? Generally i've used simple email service(SES in AWS) to send emails on behalf of a domain i own.

@dahnny012 – thanks for your interest! Once we roll this feature out (likely next week), you will be able to also use it in a self-hosted deployment.

You can have a look at this guide, how to set up transactional email: https://langfuse.com/self-hosting/configuration/transactional-emails

packages/shared/prisma/schema.prisma
maxdeichmann: @marliessophie / @FroeMic did you handle the cases of both: project_membership and org_membership revoked?

packages/shared/package.json
```diff
@@ -29,6 +29,11 @@
       "import": "./dist/src/features/comments/mentionParser.js",
       "require": "./dist/src/features/comments/mentionParser.js"
     },
+    "./src/features/rbac": {
+      "types": "./dist/src/features/rbac/index.d.ts",
+      "import": "./dist/src/features/rbac/index.js",
+      "require": "./dist/src/features/rbac/index.js"
+    },
```
maxdeichmann: @FroeMic why do we need a new export here?

web/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "declaration": false,
     "declarationMap": false,
     "outDir": "dist",
+    "rootDir": ".",
```
maxdeichmann, line 7: Why is this required?

packages/shared/package.json
FroeMic: @maxdeichmann 

1. To avoid duplicate code for fetching the users of a workspace behind secured and authorized function, marlies pointed out that it would be good to reuse the `getUserProjectRoles.ts` function.
2. This required moving it into the `shared` packages
3. The shared packages, however, has not testing instrumentation set up
4. For the tests to run in the web container this export is needed

packages/shared/prisma/schema.prisma
FroeMic: @maxdeichmann Yes we do, at runtime when sending the notification:

In `commentMentionHandler` (line 57) we call `getUserProjectRoles`, which fetches all users with the mentioned ids that have access to the project based on either their organization of project-membership. 

This function has been part of the codebase and is tested in [generateUserProjectRolesQuery.servertest.ts](https://github.com/langfuse/langfuse/pull/9951/files#diff-06327fdaeedbe7c78a8c7928a49ad867a01088026190b18ff18f0b68bfba2441). No modifications were made to these test-cases as part of this PR.

If mentioned users do not have access to the project at the time the notification is sent (might happen in very rare cases due to a race condition; VERY unlikely given the current implementation), their notification will be skipped.

web/tsconfig.json
FroeMic: @maxdeichmann After extracting `getUserProjectRoles.ts` to the shared package the build would fail do to some TS import resolve errors for the root dir. This resolved the issue. Good to go?

packages/shared/prisma/migrations/20251022143654_add_notification_preferences_redundant_index/migration.sql
```diff
@@ -0,0 +1,3 @@
+
+-- CreateIndex
+CREATE UNIQUE INDEX CONCURRENTLY "notification_preferences_user_id_project_id_channel_type_key" ON "notification_preferences"("user_id", "project_id", "channel", "type");
```
marliessophie, line 3: FYI: You do not need a concurrent index when creating empty tables. But do not really hurt, feel free to leave as is.

web/tsconfig.json
marliessophie: I am pretty sure this would cause issues and is not required. The issues you saw were possibly caused by not having rebuilt shared package or missing/incorrect export in `packages/shared/src/features/rbac/index.ts`. Please review this, we import from the shared package in web all the time, I don't see why that would suddenly cause issues.

worker/src/queues/notificationQueue.ts
```diff
@@ -0,0 +1,33 @@
...
+import { logger, QueueName, TQueueJobTypes } from "@langfuse/shared/src/server";
+import { handleCommentMentionNotification } from "../features/notifications/commentMentionHandler";
+
+export const notificationQueueProcessor = async (
+  job: Job<TQueueJobTypes[QueueName.NotificationQueue]>,
+) => {
+  try {
+    const { type, ...payload } = job.data.payload;
+
+    logger.info("Processing notification job", { type, jobId: job.id });
```
marliessophie: _Nit_: would suggest including project_id in log, otherwise we cannot identify the project that faced issues.

worker/src/queues/notificationQueue.ts
marliessophie, line 1: Generally queue related: will show you how to add graphs to DD for observability once we merge this. All configs are terraformed.

web/src/features/notifications/components/NotificationSettings.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const handleToggle = async (enabled: boolean) => {
+    setIsSaving(true);
+    try {
+      await updatePreference.mutateAsync({
+        projectId,
+        channel: "EMAIL",
+        type: "COMMENT_MENTION",
+        enabled,
+      });
+    } catch (error) {
```
marliessophie: _Nit_: pls remove try catch block and `console.error` (this emits an issue to sentry which we do not want), all logging is handled automatically via trpc, which logs to DD _and_ shows error toast to user with actionable error message or information we have been notified and are working on fix (if internal server error).

packages/shared/package.json
marliessophie: I commented on this [in the base PR](https://github.com/langfuse/langfuse/pull/9928/files#r2459439563) too, I am not convinced this is needed, please rebuild shared package and make sure you are exporting all required functions. If the test still fails, let's briefly chat sync to move fast. Applies also to mentionParser export too (let's handle both in this PR).

packages/shared/src/features/rbac/usersTableCols.ts
```diff
@@ -0,0 +1,17 @@
+import { type ColumnDefinition } from "@langfuse/shared";
+
+export const usersTableCols: ColumnDefinition[] = [
```
marliessophie, line 3: You did not delete the file in web? Please keep only one version of this file to not create unexpected drift. Shared should be fine to import from for all current usages.

packages/shared/src/features/rbac/getUserProjectRoles.ts
marliessophie, line 1: **Critical**: You did not delete the file in web? Please keep only one version of this file to not create unexpected drift. Shared should be fine to import from for all current usages. This file has some core security logic, we cannot afford drift.

web/src/server/api/routers/comments.ts
```diff
@@ -104,6 +106,32 @@ export const commentsRouter = createTRPCRouter({
           after: comment,
         });
 
+        // Enqueue notification job for mentioned users
+        if (mentionedUserIds.length > 0) {
```
marliessophie: why not 
- use `sanitizationResult.validMentionedUserIds` most importantly on line 121
- nit: include in if condition following line 56

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,118 @@
...
+  createTRPCRouter,
+  protectedProjectProcedure,
+} from "@/src/server/api/trpc";
+import { auditLog } from "@/src/features/audit-logs/auditLog";
+
+// Currently only EMAIL and COMMENT_MENTION are supported
+// Future channels: IN_APP, SLACK
+// Future types: COMMENT_REPLY, COMMENT_NEW, EVAL_COMPLETE, EXPORT_READY
+const NotificationChannelEnum = z.enum(["EMAIL"]);
+const NotificationTypeEnum = z.enum(["COMMENT_MENTION"]);
```
marliessophie: Should import from shared. This ensures trpc input schema and DB schema remain in sync.

web/src/server/api/routers/notificationPreferences.ts
```diff
@@ -0,0 +1,118 @@
...
+          channel: "EMAIL" as const,
+          type: "COMMENT_MENTION" as const,
+          enabled: preference?.enabled ?? true, // Default: enabled
+        },
+      ];
+
+      return allPreferences;
+    }),
+
+  update: protectedProjectProcedure
```
marliessophie, line 61: Question: why upsert rather than update?

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+import {
+  logger,
+  sendCommentMentionEmail,
+  getObservationById,
+} from "@langfuse/shared/src/server";
+import { prisma } from "@langfuse/shared/src/db";
+import { Prisma } from "@langfuse/shared";
+import { getUserProjectRoles } from "@langfuse/shared/src/features/rbac";
+
+type CommentMentionPayload = {
```
marliessophie: _Nit_: why redefine this?

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    if (!comment) {
+      logger.warn(
+        `Comment ${commentId} not found in project ${projectId}. Skipping notification processing.`,
+      );
+      return;
+    }
+
+    // Fetch all users (author + mentioned users) with single efficient query
+    // This replaces ~18-24 individual database queries with 1 SQL query
```
marliessophie: Let's please not add AI-generated refactor comments in our production code.

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+          comment.content.length > 500
+            ? comment.content.substring(0, 497) + "..."
+            : comment.content;
+
+        // Strip markdown link syntax from mentions for plain text display
+        // Convert @[DisplayName](user:userId) to @DisplayName
+        commentPreview = commentPreview.replace(
+          /@\[([^\]]+)\]\(user:[^)]+\)/g,
+          "@$1",
+        );
```
marliessophie: _Nit_: this does not happen for N users in loop.

worker/src/features/notifications/commentMentionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+          }
+          case "TRACE":
+          case "SESSION":
+          default: {
+            // For traces and sessions, use standard URL pattern
+            commentLink = `${baseUrl}/project/${encodeURIComponent(projectId)}/${encodeURIComponent(comment.objectType.toLowerCase())}s/${encodeURIComponent(comment.objectId)}?${commonParams}#comment-${encodeURIComponent(commentId)}`;
+            break;
+          }
+        }
+
```
marliessophie: _Nit_: can we please extract to helper function that returns returns comment link or null (if reference object not found).

packages/shared/prisma/schema.prisma
marliessophie: Generally: Runtime checks in `handleCommentMentionNotification` validate all access conditions at job processing time: the handler fetches the comment (fails if project/comment deleted), queries current project members via `getUserProjectRoles()` (excludes deleted users or removed memberships), and checks notification preferences (respects disabled settings). This ensures emails are never sent to users who no longer have access, even if the job was queued earlier when access was valid. As more notification types are added, these validation patterns should be abstracted into reusable functions. 

- Note: Job sends to new email, not email at mention time. This is expected imo. 
- One shortcoming: when project/org membership for user is deleted notification preference will continue to exist. We check `getUserProjectRoles` in queue job so not email would be sent, but we'd have orphaned record (ie project_id, user_id combination that is non-sensical). Will not block this PR because of this, but would be great to add a comment.

packages/shared/prisma/migrations/20251022143653_add_notification_preferences/migration.sql
marliessophie, line 1: Have you run postgres migrations before? If not, let's please do this together.

marliessophie (commented):
Good news, all security related comments look good now. Mostly nits and a couple of file structure comments.

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: done

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: moved outside of loop

web/src/server/api/routers/notificationPreferences.ts
FroeMic: Imo easier than two separate functions to handle from client. We can just display the defaults in the client and once the user changes them for the first time, we upsert. No need for complicated checking whether the preference actually exists.

worker/src/features/notifications/commentMentionHandler.ts
FroeMic: done

FroeMic: now imported

web/src/server/api/routers/notificationPreferences.ts
FroeMic: done

web/src/server/api/routers/comments.ts
FroeMic: done

packages/shared/src/features/rbac/getUserProjectRoles.ts
FroeMic: done

packages/shared/src/features/rbac/usersTableCols.ts
FroeMic: deleted

web/src/features/notifications/components/NotificationSettings.tsx
FroeMic: removed

packages/shared/prisma/migrations/20251022143654_add_notification_preferences_redundant_index/migration.sql
FroeMic: oke

packages/shared/prisma/migrations/20251022143653_add_notification_preferences/migration.sql
FroeMic: yes – I think I can handle it, but we can also do it together if you want :)

worker/src/queues/notificationQueue.ts
FroeMic: Did it once already. Will open a PR for this

worker/src/queues/notificationQueue.ts
FroeMic: done

packages/shared/prisma/schema.prisma
marliessophie: cc @maxdeichmann, pls give 👍 so we can close this. Better safe than sorry; thanks.

web/package.json
```diff
@@ -33,6 +33,7 @@
     "@dnd-kit/modifiers": "^7.0.0",
     "@dnd-kit/sortable": "^8.0.0",
     "@dnd-kit/utilities": "^3.2.2",
+    "@ferrucc-io/emoji-picker": "^0.0.47",
```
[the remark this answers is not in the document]
FroeMic: no

packages/shared/prisma/schema.prisma
maxdeichmann: LGTM, thanks!

PR #9953, opened 2025-10-22
feat(comments): Add comment quicksearch bar

FroeMic:
## What's New?

Introduced a client-side quicksearch bar to the comment list to quickly filter comments by text

CleanShot 2025-10-23 at 01 00 03@2x

----

> [!IMPORTANT]
> Add quicksearch bar and emoji reactions to comments, updating database, API, and UI components accordingly.
> 
>   - **Features**:
>     - Add quicksearch bar to `CommentList` for filtering comments by text.
>     - Introduce comment reactions with emoji support in `CommentList` and `ReactionBar`.
>   - **Database**:
>     - Add `CommentReaction` model to `schema.prisma` and `types.ts`.
>     - Create `comment_reactions` table and index in `migration.sql` files.
>   - **API**:
>     - Add `commentReactionsRouter` to handle adding, removing, and listing reactions in `commentReactions.ts`.
>     - Update `commentsRouter` to include comment count queries in `comments.ts`.
>   - **UI Components**:
>     - Update `CommentDrawerButton`, `CommentList`, `ReactionBar`, and `ReactionPicker` to support new features.
>     - Add `@ferrucc-io/emoji-picker` to `package.json` for emoji selection.
>   - **Misc**:
>     - Add `stripMarkdown` utility function in `markdown.ts` for processing comment content.
>     - Update `tailwind.config.ts` to include emoji picker styles.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 7804ee080dc819d7b93b11c759d5573a3638dac4. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/features/comments/CommentDrawerButton.tsx
```diff
@@ -140,7 +140,16 @@ export function CommentDrawerButton({
         </Button>
       </DrawerTrigger>
       <DrawerContent overlayClassName="bg-primary/10">
-        <div className="mx-auto flex h-full w-full flex-col overflow-hidden md:max-h-full">
+        <div
+          className="mx-auto flex h-full w-full flex-col overflow-hidden md:max-h-full"
+          tabIndex={-1}
+          ref={(el) => {
+            // Auto-focus drawer content when it opens
```
marliessophie: Personal opinion: I prefer drawer content _not_ to be focused when drawer opens; I think we can skip this. Also it's not consistent with other drawers. Will not die on this hill though.

web/src/features/comments/CommentList.tsx
```diff
@@ -265,6 +300,36 @@ export function CommentList({
     }));
   }, [comments.data]);
 
+  // Helper function to strip markdown formatting for search
+  const stripMarkdown = (text: string): string => {
```
marliessophie: _Nit_: can we extract this to some util file please? Or at least out of component.

marliessophie (commented):
Two small comments, non-blocking.

web/src/features/comments/CommentList.tsx
```diff
@@ -265,6 +300,36 @@ export function CommentList({
...
+
+  // Client-side filtering based on search query
+  const filteredComments = useMemo(() => {
+    if (!searchQuery.trim()) {
+      return commentsWithFormattedTimestamp;
+    }
+
+    const query = searchQuery.toLowerCase();
+    return commentsWithFormattedTimestamp?.filter((comment) => {
+      const strippedContent = stripMarkdown(comment.content).toLowerCase();
```
marliessophie: _Nit_: `stripMarkdown` runs on every render. Would suggest to memoize or debounce for performance.

web/src/features/comments/CommentDrawerButton.tsx
FroeMic: If we do not focus the drawer, then CMD+F  will be captured by the browser, not the quicksearchbar. 

My intention: I believe it would be expected behavior for the drawer to focus after opening, so keyboard navigation (tab tab tab) captures the right input elements. 

Curious to hear: What are your thoughts on why we shouldn't do it?

web/src/features/comments/CommentList.tsx
FroeMic: Since we are on react-19, I would advise against using useMemo or useCallback and let the new react compiler do its thing? 

Weakly held opinion, see https://dev.to/joodi/react-19-memoization-is-usememo-usecallback-no-longer-necessary-3ifn

I am happy to refactor, but I think it is worth getting to a common understanding so I can adhere to coding standard in the future

web/src/features/comments/CommentList.tsx
FroeMic: done

web/src/features/comments/CommentDrawerButton.tsx
FroeMic: Note: I disabled an focus-visible css style on the drawer to void the no-so-nice border when focused via keyboard

marliessophie: This solution works for me!

web/src/features/comments/CommentList.tsx
marliessophie: From my understanding, when using NextJS the new react compiler is an experimental feature that needs to be opted-in the config; we never did this. So we still need to use memo etc for now. 

Separate discussion should we want to try out the new react compiler soon, lmk should you see a case for this.

FroeMic: will memoize

PR #10029, opened 2025-10-27
feat(api): Traces public API v1 re-implementation

sumerman:
builds on https://github.com/langfuse/langfuse/pull/10027 

----

> [!IMPORTANT]
> Re-implements Traces public API v1 to use the events table, introducing new query builders and updating tests for dual-path support.
> 
>   - **Behavior**:
>     - Re-implements Traces public API v1 to use the events table instead of the legacy traces table.
>     - Adds `useEventsTable` parameter to toggle between events and traces table.
>     - Updates `getTracesFromEventsTableForPublicApi` and `getTracesCountFromEventsTableForPublicApi` to handle traces from events table.
>   - **Query Builders**:
>     - Introduces `EventsAggregationQueryBuilder` and `CTEQueryBuilder` in `event-query-builder.ts` for building complex queries with CTEs.
>     - Modifies `eventsTracesAggregation` and `eventsTracesScoresAggregation` in `query-fragments.ts` to support new query structure.
>   - **API Changes**:
>     - Updates `GET /api/public/traces` in `traces.ts` to use new query builders and handle `useEventsTable` parameter.
>     - Adjusts `GetTracesV1Query` and `GetTracesV1Response` in `types/traces.ts` to support new fields and parameters.
>   - **Tests**:
>     - Adds dual-path tests in `traces-api.servertest.ts` to cover both legacy and new implementations.
>     - Ensures tests validate behavior with and without `useEventsTable` parameter.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for fc97bf562002d30269a74a4bd42f363f7969238e. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
   projectId: "project_id",
 
   // Aggregated fields
-  name: "argMaxIf(name, event_ts, isNull(parent_span_id)) AS name",
+  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
   timestamp: "min(start_time) as timestamp",
-  environment: "argMax(environment, event_ts) AS environment",
-  version: "argMax(version, event_ts) AS version",
-  session_id: "argMax(session_id, event_ts) AS session_id",
-  user_id: "argMax(user_id, event_ts) AS user_id",
+  environment:
+    "argMaxIf(environment, event_ts, environment <> '') AS environment",
+  version: "argMaxIf(version, event_ts, version <> '') AS version",
+  session_id: "argMaxIf(session_id, event_ts, session_id <> '') AS session_id",
+  user_id: "argMaxIf(user_id, event_ts, user_id <> '') AS user_id",
   input: "argMax(input, event_ts) AS input",
   output: "argMax(output, event_ts) AS output",
   metadata: "argMax(metadata, event_ts) AS metadata",
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -48,7 +48,7 @@ const EVENTS_FIELDS = {
   // I/O & metadata fields
   input: "e.input",
   output: "e.output",
-  metadata: "e.metadata",
+  metadata: "mapFromArrays(e.metadata_names, e.metadata_values) as metadata",
 
   // Calculated fields
   latency:
@@ -149,7 +149,8 @@ const EVENTS_AGGREGATION_FIELDS = {
   user_id: "argMaxIf(user_id, event_ts, user_id <> '') AS user_id",
   input: "argMaxIf(input, event_ts, parent_span_id = '') AS input",
   output: "argMaxIf(output, event_ts, parent_span_id = '') AS output",
-  metadata: "argMaxIf(metadata, event_ts, parent_span_id = '') AS metadata",
+  metadata:
+    "argMaxIf(mapFromArrays(e.metadata_names, e.metadata_values), event_ts, parent_span_id = '') AS metadata",
   created_at: "min(created_at) AS created_at",
   updated_at: "max(updated_at) AS updated_at",
   total_cost: "sum(total_cost) AS total_cost",
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -140,23 +140,23 @@ const EVENTS_AGGREGATION_FIELDS = {
   projectId: "project_id",
 
   // Aggregated fields
-  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
+  name: "argMaxIf(name, event_ts, parent_span_id = '') AS name",
   timestamp: "min(start_time) as timestamp",
   environment:
     "argMaxIf(environment, event_ts, environment <> '') AS environment",
   version: "argMaxIf(version, event_ts, version <> '') AS version",
   session_id: "argMaxIf(session_id, event_ts, session_id <> '') AS session_id",
   user_id: "argMaxIf(user_id, event_ts, user_id <> '') AS user_id",
-  input: "argMax(input, event_ts) AS input",
-  output: "argMax(output, event_ts) AS output",
-  metadata: "argMax(metadata, event_ts) AS metadata",
+  input: "argMaxIf(input, event_ts, parent_span_id = '') AS input",
+  output: "argMaxIf(output, event_ts, parent_span_id = '') AS output",
+  metadata: "argMaxIf(metadata, event_ts, parent_span_id = '') AS metadata",
   created_at: "min(created_at) AS created_at",
   updated_at: "max(updated_at) AS updated_at",
   total_cost: "sum(total_cost) AS total_cost",
   latency_milliseconds:
-    "date_diff('millisecond', least(min(start_time), min(end_time)), greatest(max(start_time), max(end_time))) AS latency_milliseconds",
+    "date_diff('millisecond', min(start_time), greatest(max(start_time), max(end_time))) AS latency_milliseconds",
   observation_ids:
-    "groupUniqArrayIf(span_id, isNotNull(span_id) AND span_id != '') AS observation_ids",
+    "groupUniqArrayIf(span_id, span_id <> '') AS observation_ids",
 
   // Legacy fields for backward compatibility
   tags: "array() AS tags",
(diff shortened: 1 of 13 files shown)


packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
   projectId: "project_id",
 
   // Aggregated fields
-  name: "argMaxIf(name, event_ts, isNull(parent_span_id)) AS name",
+  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
```
Steffen911: I think you can drop the isNull here as we'll most likely make the field non-nullable.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
...
   timestamp: "min(start_time) as timestamp",
-  environment: "argMax(environment, event_ts) AS environment",
-  version: "argMax(version, event_ts) AS version",
-  session_id: "argMax(session_id, event_ts) AS session_id",
-  user_id: "argMax(user_id, event_ts) AS user_id",
+  environment:
+    "argMaxIf(environment, event_ts, environment <> '') AS environment",
+  version: "argMaxIf(version, event_ts, version <> '') AS version",
+  session_id: "argMaxIf(session_id, event_ts, session_id <> '') AS session_id",
+  user_id: "argMaxIf(user_id, event_ts, user_id <> '') AS user_id",
```
Steffen911, line 149: I think it would make sense to apply the setup for `name` to input/output/metadata as well via `argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') `. Otherwise, we'd probably just use the last-seen span data?

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
...
+  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
   timestamp: "min(start_time) as timestamp",
-  environment: "argMax(environment, event_ts) AS environment",
-  version: "argMax(version, event_ts) AS version",
-  session_id: "argMax(session_id, event_ts) AS session_id",
-  user_id: "argMax(user_id, event_ts) AS user_id",
+  environment:
+    "argMaxIf(environment, event_ts, environment <> '') AS environment",
+  version: "argMaxIf(version, event_ts, version <> '') AS version",
+  session_id: "argMaxIf(session_id, event_ts, session_id <> '') AS session_id",
```
Steffen911, line 148: Further below you have the latency calculation. IMO min(end_time) is not necessary as start_time is non-nullable.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
...
-  name: "argMaxIf(name, event_ts, isNull(parent_span_id)) AS name",
+  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
   timestamp: "min(start_time) as timestamp",
-  environment: "argMax(environment, event_ts) AS environment",
-  version: "argMax(version, event_ts) AS version",
-  session_id: "argMax(session_id, event_ts) AS session_id",
-  user_id: "argMax(user_id, event_ts) AS user_id",
+  environment:
+    "argMaxIf(environment, event_ts, environment <> '') AS environment",
+  version: "argMaxIf(version, event_ts, version <> '') AS version",
```
Steffen911, line 147: Also there is a new metadata key `trace_tags` going forward. You could use it for the legacy mapping and deconstruct it back into the expected field.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -140,12 +140,13 @@ const EVENTS_AGGREGATION_FIELDS = {
...
 
   // Aggregated fields
-  name: "argMaxIf(name, event_ts, isNull(parent_span_id)) AS name",
+  name: "argMaxIf(name, event_ts, isNull(parent_span_id) OR parent_span_id = '') AS name",
   timestamp: "min(start_time) as timestamp",
-  environment: "argMax(environment, event_ts) AS environment",
-  version: "argMax(version, event_ts) AS version",
-  session_id: "argMax(session_id, event_ts) AS session_id",
-  user_id: "argMax(user_id, event_ts) AS user_id",
+  environment:
```
Steffen911, line 145: You can also move to the `truncated` version of the IO fields further below.

Steffen911 (approved):
Added a couple of notes on the data model. API logic itself LGTM 🙌

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: Ah, thanks. Forgot about it after a rebase.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: I don't think this is an expected behaviour in v1 APIs. Will punt on it for now.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: I don't see anything in the current dev-tables script that would spell "tags" 🤔

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: I will switch to truncated for the observation ui table though, there seems to be an option already

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: Won't do in this PR. Figuring out tags/public/bookmarking... is my next order of business anyway.

byteccwu:
When using the openapi interface (/api/public/traces) for querying, if latency is used as a filter parameter, an SQL exception will occur. Is this due to the program not supporting it or a bug in the program?

filter: [{"column":"latency","operator":">=","value":0,"type":"number"},{"column":"latency","operator":"<=","value":1.9,"type":"number"}]

exception:Unknown expression or function identifier `o.latency_milliseconds` in scope SELECT count() AS count FROM traces AS t WHERE (project_id = 'cmha81n1o000pnt085m2zhyyq') AND ((o.latency_milliseconds \/ 1000) >= _CAST('0', 'Decimal64(3)')) AND ((o.latency_milliseconds \/ 1000) <= _CAST('1.9', 'Decimal64(3)')) AND (t.environment IN _CAST(['dev'], 'Array(String)')).

sumerman:
@byteccwu 

> When using the openapi interface (/api/public/traces) for querying, if latency is used as a filter parameter, an SQL exception will occur. Is this due to the program not supporting it or a bug in the program?
> 
> filter: [{"column":"latency","operator":">=","value":0,"type":"number"},{"column":"latency","operator":"<=","value":1.9,"type":"number"}]
> 
> exception:Unknown expression or function identifier `o.latency_milliseconds` in scope SELECT count() AS count FROM traces AS t WHERE (project_id = 'cmha81n1o000pnt085m2zhyyq') AND ((o.latency_milliseconds / 1000) >= _CAST('0', 'Decimal64(3)')) AND ((o.latency_milliseconds / 1000) <= _CAST('1.9', 'Decimal64(3)')) AND (t.environment IN _CAST(['dev'], 'Array(String)')).

Looks like bug. May I ask you to provide a complete set of http parameters? E.g. did you try to use `useEventsTable` that I am introducing in this PR?

byteccwu:
> @byteccwu
> 
> > When using the openapi interface (/api/public/traces) for querying, if latency is used as a filter parameter, an SQL exception will occur. Is this due to the program not supporting it or a bug in the program?
> > filter: [{"column":"latency","operator":">=","value":0,"type":"number"},{"column":"latency","operator":"<=","value":1.9,"type":"number"}]
> > exception:Unknown expression or function identifier `o.latency_milliseconds` in scope SELECT count() AS count FROM traces AS t WHERE (project_id = 'cmha81n1o000pnt085m2zhyyq') AND ((o.latency_milliseconds / 1000) >= _CAST('0', 'Decimal64(3)')) AND ((o.latency_milliseconds / 1000) <= _CAST('1.9', 'Decimal64(3)')) AND (t.environment IN _CAST(['dev'], 'Array(String)')).
> 
> Looks like bug. May I ask you to provide a complete set of http parameters? E.g. did you try to use `useEventsTable` that I am introducing in this PR?

> The useEventsTable will throw an exception whether it is True or False。
Request Param: /api/public/traces?page=1&limit=10&environment=dev&filter=[{"type": "number","column": "latency","operator": ">","value": 0.5},{"type": "number","column": "latency","operator": "<","value": 5}]

PR #10091, opened 2025-10-29
chore: V2 of don't run llm tests unless required to avoid throttling on CI

sumerman:
> [!IMPORTANT]
> Optimize CI/CD pipeline to run LLM connection tests only when necessary, reducing CI throttling.
> 
>   - **CI/CD Pipeline**:
>     - Adds `llm_connections_changed` output in `pre-job` to track changes in `packages/shared/package.json` and `packages/shared/src/server/llm/fetchLLMCompletion.ts`.
>     - Modifies `test-worker-llm-connections` job to run only if LLM connections change, a tag is pushed, or the workflow is manually triggered.
>   - **Job Outputs**:
>     - Adds `success` output in `all-ci-passed` to determine if all jobs passed or were skipped.
>     - Updates `push-docker-image` job to check `success` output before proceeding with deployment.
>   - **Misc**:
>     - Adds `Set check result as an output` step in `all-ci-passed` to determine overall job success.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 439b7e26032197199146e915651a726a0b27150c. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -21,12 +21,28 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
+      llm_connections_changed: ${{ steps.filter.outputs.llm_connections }}
     timeout-minutes: 15
+    permissions:
+      pull-requests: read
     steps:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@v5
         with:
           do_not_skip: '["workflow_dispatch"]'
+      - uses: actions/checkout@v4
+        with:
+          # Recommended for paths-filter action
+          # may save additional git fetch roundtrip if
+          # merge-base is found within latest N commits
+          fetch-depth: 20
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yml
```diff
@@ -370,7 +386,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - pre-job
-    if: needs.pre-job.outputs.should_skip != 'true'
+    if: needs.pre-job.outputs.should_skip != 'true' && (needs.pre-job.outputs.llm_connections_changed == 'true' || github.event_name == 'workflow_dispatch')
```
nimarb: i would also run them on every tagged release, just to be sure that stuff works in case sth in the target API or similar broke:

```
startsWith(github.ref, 'refs/tags/') || (needs.pre-job.outputs.should_skip != 'true' && (needs.pre-job.outputs.llm_connections_changed == 'true' || github.event_name == 'workflow_dispatch'))
```

PR #10178, opened 2025-11-04
feat: bookmarked dual updates / writes in events table

sumerman:
- **feat:** expand events table definition adding public & bookmarked
- **feat:** updating backfill code to copy bookmarks and public
- **feat:** Trace bookmark propagation onto corresponding events

----

> [!IMPORTANT]
> Add `bookmarked` and `public` fields to events table, update related queries and tests, and introduce environment flag for enabling these features.
> 
>   - **Events Table Schema**:
>     - Add `bookmarked` and `public` fields to `events` table in `dev-tables.sh` and `backfillEventsHistoric.ts`.
>     - Update `eventRecordBaseSchema` in `definitions.ts` to include `bookmarked` and `public`.
>   - **Query and Functionality Updates**:
>     - Modify `getTraceByIdFromEventsTable` in `events.ts` to handle `bookmarked` and `public` fields.
>     - Update `updateEvents` function in `events.ts` to allow updating `bookmarked` and `public` fields.
>     - Adjust `eventsTracesAggregation` in `query-fragments.ts` to include new fields.
>   - **Environment and Configuration**:
>     - Add `LANGFUSE_ENABLE_EVENTS_TABLE_FLAGS` to `.env.dev.example` and `env.mjs`.
>   - **Testing**:
>     - Add tests in `event-repository.servertest.ts` and `traces-trpc.servertest.ts` to verify bookmarking functionality and public flag handling.
>   - **Miscellaneous**:
>     - Minor comment and logging updates in `query-fragments.ts` and `event-query-builder.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for fb645ddb1cc128065591af801aca4ce59e64bd82. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/__tests__/async/repositories/event-repository.servertest.ts
```diff
@@ -1501,4 +1503,98 @@ describe("Clickhouse Events Repository Test", () => {
       await prisma.model.delete({ where: { id: modelId } });
     });
   });
+
+  maybe("Update methods", () => {
```
maxdeichmann, line 1507: What is maybe doing?

web/src/server/api/routers/traces.ts
```diff
@@ -412,7 +414,24 @@ export const traceRouter = createTRPCRouter({
...
+              ? { rootEventOnly: true }
+              : {};
+            promises.push(
+              updateEvents(
+                input.projectId,
+                { traceIds: [clickhouseTrace.id], ...rootEventOnly },
+                { bookmarked: input.bookmarked },
+              ),
+            );
+          }
```
maxdeichmann, line 428: I think we need the same logic also for public

web/src/__tests__/async/repositories/event-repository.servertest.ts
sumerman: It is defined at the top.

```typescript
const maybe =
  env.LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS === "true"
    ? describe
    : describe.skip;

```

sumerman: We have this to pass azure suite that doesn't have `events` table

web/src/server/api/routers/traces.ts
sumerman: It's coming :)

packages/shared/clickhouse/scripts/dev-tables.sh
```diff
@@ -225,6 +225,8 @@ CREATE TABLE IF NOT EXISTS events
       updated_at DateTime64(6) DEFAULT now(),
       event_ts DateTime64(6),
       is_deleted UInt8,
+      bookmarked Bool DEFAULT false,
+      public Bool DEFAULT false,
```
Steffen911: I would move those up next to version etc. That way we keep track of similar kinds of properties in the same part of the DDL.

packages/shared/clickhouse/scripts/dev-tables.sh
```diff
@@ -254,7 +256,9 @@ CREATE TABLE IF NOT EXISTS events
     index_granularity = 8192,
     index_granularity_bytes = '64Mi', -- Default 10MiB. Avoid small granules due to large rows.
     min_rows_for_wide_part = 0,
-    min_bytes_for_wide_part = 0;
+    min_bytes_for_wide_part = 0,
+    enable_block_number_column = 1,
+    enable_block_offset_column = 1;
```
Steffen911, line 262: Great catch!

packages/shared/clickhouse/scripts/dev-tables.sh
```diff
@@ -325,12 +324,15 @@ clickhouse client \
...
          ''                                                                            AS event_raw,
          0                                                                             AS event_bytes,
-         created_at,
-         updated_at,
-         event_ts,
-         is_deleted
-  FROM observations
-  WHERE (is_deleted = 0);
+         t.public                                                                      AS public,
+         t.bookmarked AND (o.parent_observation_id IS NULL OR o.parent_observation_id == '') AS bookmarked,
```
Steffen911: What is the idea here? Why do we need the conditional statement?

packages/shared/src/server/repositories/events.ts
```diff
@@ -447,6 +450,125 @@ const getObservationByIdFromEventsTableInternal = async ({
...
+
+  // Handle input/output with truncation
+  if (renderingProps.truncated) {
+    queryBuilder
+      .select(
+        `leftUTF8(t.input_truncated, ${env.LANGFUSE_SERVER_SIDE_IO_CHAR_LIMIT}) as input`,
+      )
+      .select(
+        `leftUTF8(t.output_truncated, ${env.LANGFUSE_SERVER_SIDE_IO_CHAR_LIMIT}) as output`,
+      );
```
Steffen911, line 523: IMO we shouldn't truncate here again on ClickHouse. I would assume that the truncation that happens on write is sufficient.

sumerman: is there any downside for this extra truncation though? This logic matches traces table behaviour.

packages/shared/clickhouse/scripts/dev-tables.sh
sumerman: The logic: 
- if a trace public make all related events public, later, should we change auth check logic to be on the event basis, this should hold.
- bookmarked traces become bookmarked root spans.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -159,10 +165,11 @@ const EVENTS_AGGREGATION_FIELDS = {
   observation_ids:
     "groupUniqArrayIf(span_id, span_id <> '') AS observation_ids",
 
+  bookmarked:
+    "argMaxIf(bookmarked, event_ts, parent_span_id = '') AS bookmarked",
+  public: "argMaxIf(public, event_ts, parent_span_id = '') AS public",
```
Steffen911, line 170: This will become `max()`, right?

packages/shared/src/server/repositories/events.ts
Steffen911: It may have additional CPU overhead on reads compared to reading naively. As this is a read and can be changed at any time it's non blocking for a merge IMO.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: Yes.

PR #10277, opened 2025-11-07
fix(db-schema): make schema match prod db state (json instead of jsonb)

nimarb:
> [!IMPORTANT]
> Update `inputSchema` and `expectedOutputSchema` fields in `Dataset` model to use `@db.Json` in `schema.prisma`.
> 
>   - **Schema Update**:
>     - In `schema.prisma`, update `inputSchema` and `expectedOutputSchema` fields in `Dataset` model to use `@db.Json` to match production database state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 35b4c9a4f56db9f6c247937323a47f122d4362c4. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/prisma/schema.prisma
+++ b/packages/shared/prisma/schema.prisma
@@ -598,8 +598,8 @@ model Dataset {
   metadata                Json?
   remoteExperimentUrl     String?       @map("remote_experiment_url")
   remoteExperimentPayload Json?         @map("remote_experiment_payload")
-  inputSchema             Json?         @map("input_schema")
-  expectedOutputSchema    Json?         @map("expected_output_schema")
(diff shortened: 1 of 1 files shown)


nimarb:
@hassiebp , FYI: we created input_schema and expected_output_schema columns as json type, but prisma's default for Json? is JSONB, causing schema drift that generated an unwanted migration. fixed by adding @db.Json annotations to the schema to explicitly match the production type, preventin new migrations.

Issue #10290, opened 2025-11-07
bug: Some alternative S3 compatible storages also require to pass the region for the client


deejay1:
### Describe the bug

Some storage types also require to set the region for them to work properly not only stock S3 (for example some ceph versions)

### Steps to reproduce

1. Have a ceph based S3 storage
2. Create an S3 compatible blob storage export integration
3. Try to run validate
4. See the whole web instance die

### Langfuse Cloud or self-hosted?

Self-hosted

### If self-hosted, what version are you running?

3.128.0

### Additional information

```javascript
[...]

PR #10297, opened 2025-11-08
fix(ui): allow to specify region for S3 compatible storage

deejay1:
## What does this PR do?

Show the Region field for S3 and compatible storage, because some backend require to specify the region

Relates to #10290

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Show 'Region' field for both AWS S3 and S3-compatible storage in `blobstorage.tsx`.
> 
>   - **Behavior**:
>     - Modify `BlobStorageIntegrationSettingsForm` in `blobstorage.tsx` to show 'Region' field for both AWS S3 and S3-compatible storage.
>     - Update 'Region' field description to differentiate between AWS S3 and S3-compatible storage.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 67a5b5d63f921f91bc20c47ef3785bcc9cfda8ae. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx
```diff
@@ -356,7 +356,11 @@ const BlobStorageIntegrationSettingsForm = ({
                 <FormControl>
                   <Input {...field} />
                 </FormControl>
-                <FormDescription>AWS region (e.g., us-east-1)</FormDescription>
+                <FormDescription>
+                  {integrationType === "S3"
+                    ? "AWS region (e.g., us-east-1)"
+                    : "S3 compatible storage region"}
+                </FormDescription>
```
Comment (Steffen911) [inline web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx:363]: @deejay1 Have you tested what happens if a user doesn't provide the region? Do we fallback to `auto` in this case?

Comment (deejay1) [reply web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx:363]: I'm looking at it now, as far as we managed to find, aws-s3 sdk requires te region to be set as soon as you have a custom endpoint specified

Comment (deejay1) [reply web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx:363]: @Steffen911 when you don't specify the region it acts exactly like now - kills the web server due to passing nothing to the region parameter in the S3 client, maybe we should set `auto` as the default?

Comment (Steffen911) [reply web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx:363]: @deejay1 Yes, this seems reasonable. Thank you!

Steffen911 (approved):

PR #10465, opened 2025-11-13
feat(api): public api v2 basic functionality

sumerman:
> [!IMPORTANT]
> Introduces public API v2 for observations with selective field fetching and cursor-based pagination, updating query builders, converters, and tests.
> 
>   - **API Enhancements**:
>     - Introduces `getObservationsV2FromEventsTableForPublicApi` for v2 API with selective field fetching and cursor-based pagination.
>     - Adds `OBSERVATION_FIELD_GROUPS` and `ObservationFieldGroup` for field group management in `events.ts`.
>     - Implements cursor-based pagination in `observations.ts` and `index.ts`.
>   - **Query Builders**:
>     - Updates `EventsQueryBuilder` and `CTEQueryBuilder` to support new field sets and cursor logic.
>     - Adds `applyCursorPagination` and `applyOrderByForObservationsQuery` functions in `events.ts`.
>   - **Converters**:
>     - Modifies `convertObservationPartial` and `convertEventsObservation` to handle partial observations.
>     - Adds `ensureObservationCoreFields` for core field validation in `observations_converters.ts`.
>   - **Tests**:
>     - Adds `observations-api-v2.servertest.ts` for testing v2 API features like field selection and pagination.
>     - Updates existing tests in `observations-api.servertest.ts` to ensure compatibility with new API changes.
>   - **Miscellaneous**:
>     - Updates `mapEventsTable.ts` to include new field mappings.
>     - Refactors `getObservationsFromEventsTableForPublicApi` to use new query logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 8218f76d233c8474d7d7b94f0e143fe27399a794. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -136,7 +136,7 @@ const FIELD_SETS = {
   byIdTimestamps: ["createdAt", "updatedAt", "eventTs"],
 
   // Public API v2 field sets (field groups for selective fetching)
-  publicApiCore: [
+  core: [
     "id",
     "traceId",
     "startTime",
@@ -145,7 +145,7 @@ const FIELD_SETS = {
     "parentObservationId",
     "type",
   ],
-  publicApiBasic: [
+  basic: [
     "name",
     "level",
     "statusMessage",
@@ -156,19 +156,17 @@ const FIELD_SETS = {
     "userId",
     "sessionId",
   ],
-  publicApiTime: ["completionStartTime", "createdAt", "updatedAt"],
-  publicApiIo: ["input", "output"],
-  publicApiMetadata: ["metadata"],
-  publicApiModel: ["providedModelName", "internalModelId", "modelParameters"],
-  publicApiUsage: [
+  time: ["completionStartTime", "createdAt", "updatedAt"],
+  model: ["providedModelName", "internalModelId", "modelParameters"],
+  usage: [
     "usageDetails",
     "costDetails",
     "totalCost",
     "providedUsageDetails",
     "providedCostDetails",
   ],
-  publicApiPrompt: ["promptId", "promptName", "promptVersion"],
-  publicApiMetrics: ["latency", "timeToFirstToken"],
+  prompt: ["promptId", "promptName", "promptVersion"],
+  metrics: ["latency", "timeToFirstToken"],
 } as const;
 
 export type FieldSetName = keyof typeof FIELD_SETS;
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -260,10 +260,13 @@ abstract class AbstractQueryBuilder {
    * Add LIMIT and OFFSET
    */
   limit(limit?: number, offset?: number): this {
-    if (limit !== undefined && offset !== undefined) {
+    if (limit !== undefined && offset !== undefined && offset > 0) {
       this.limitClause = "LIMIT {limit: Int32} OFFSET {offset: Int32}";
       this.params.limit = limit;
       this.params.offset = offset;
+    } else if (limit !== undefined) {
+      this.limitClause = "LIMIT {limit: Int32}";
+      this.params.limit = limit;
     } else {
       this.limitClause = "";
     }
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -134,7 +134,44 @@ const FIELD_SETS = {
   ],
   byIdPrompt: ["promptId", "promptName", "promptVersion"],
   byIdTimestamps: ["createdAt", "updatedAt", "eventTs"],
+
+  // Public API v2 field sets (field groups for selective fetching)
+  publicApiCore: [
+    "id",
+    "traceId",
+    "startTime",
+    "endTime",
+    "projectId",
+    "parentObservationId",
+    "type",
+  ],
+  publicApiBasic: [
+    "name",
+    "level",
+    "statusMessage",
+    "version",
+    "environment",
+    "bookmarked",
+    "public",
+    "userId",
+    "sessionId",
+  ],
+  publicApiTime: ["completionStartTime", "createdAt", "updatedAt"],
+  publicApiIo: ["input", "output"],
+  publicApiMetadata: ["metadata"],
+  publicApiModel: ["providedModelName", "internalModelId", "modelParameters"],
+  publicApiUsage: [
+    "usageDetails",
+    "costDetails",
+    "totalCost",
+    "providedUsageDetails",
+    "providedCostDetails",
+  ],
+  publicApiPrompt: ["promptId", "promptName", "promptVersion"],
+  publicApiMetrics: ["latency", "timeToFirstToken"],
 } as const;
+
+export type FieldSetName = keyof typeof FIELD_SETS;
 
 /**
  * Aggregation fields for trace-level queries
@@ -483,7 +520,7 @@ export class EventsQueryBuilder extends BaseEventsQueryBuilder<
   /**
    * Add SELECT fields from predefined field sets
    */
-  selectFieldSet(...setNames: Array<keyof typeof FIELD_SETS>): this {
+  selectFieldSet(...setNames: Array<FieldSetName>): this {
     setNames
       .flatMap((s) => {
         return FIELD_SETS[s];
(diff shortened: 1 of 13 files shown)


packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -132,8 +132,43 @@ const FIELD_SETS = {
...
+    "statusMessage",
+    "version",
+    "environment",
+    "bookmarked",
+    "public",
+    "userId",
+    "sessionId",
+  ],
+  time: ["completionStartTime", "createdAt", "updatedAt"],
+  model: ["providedModelName", "internalModelId", "modelParameters"],
```
maxdeichmann, line 160: Are you sure you want to move API specific selectors into the query builder? IMO we should have a logical separation from the API layer and query builder layer.

packages/shared/src/server/repositories/events.ts
```diff
@@ -56,44 +59,98 @@ type ObservationsTableQueryResultWitouhtTraceFields = Omit<
   "trace_tags" | "trace_name" | "trace_user_id"
 >;
 
+type ObservationV2 = Partial<Observation & ObservationPriceFields> &
```
maxdeichmann: I feel we should keep the observation public api types in the api layer

packages/shared/src/server/repositories/events.ts
```diff
@@ -102,12 +159,12 @@ const enrichObservationsWithModelData = async (
       totalPrice:
         model?.Price?.find((m) => m.usageType === "total")?.price ?? null,
     };
-  });
-};
+  }) as Array<R>;
```
maxdeichmann: Why do you need the type override here?

sumerman: Because the return type is conditional and this is the only way to make it typecheck

packages/shared/src/server/repositories/events.ts
sumerman: This was my initial reaction, but then I realized I reference these in `shared` and most of API lives in `web`. What would the best place to shove them into?

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: On the one hand I agree, on the other I don't like that we have practically identical definitions in dozens of places. I will think how can I make this nicer.

web/src/features/public-api/types/observations.ts
```diff
@@ -205,3 +210,123 @@ export const GetObservationV1Query = z.object({
...
+  // Parsing behavior
+  parseIoAsJson: z
+    .union([z.literal("true"), z.literal("false")])
+    .transform((val) => val === "true")
+    .default(false),
+  // Filters
+  topLevelOnly: z
+    .union([z.literal("true"), z.literal("false")])
+    .optional()
+    .transform((val) => (val === undefined ? undefined : val === "true")),
```
maxdeichmann: IMO we should drop the toplevel for now. We cannot guarantee that the toplevel observation of a user has always no parentObservationId. This happens in case someone declares a dedicated otel tracerprovider in their applications: https://langfuse.com/docs/observability/sdk/python/advanced-usage#isolated-tracerprovider

web/src/features/public-api/types/observations.ts
```diff
@@ -205,3 +210,123 @@ export const GetObservationV1Query = z.object({
...
+          ? cursor.lastStartTimeTo.toISOString()
+          : cursor.lastStartTimeTo,
+      lastTraceId: cursor.lastTraceId,
+      lastId: cursor.lastId,
+    }),
+  ).toString("base64");
+};
+
+// GET /v2/observations
+export const GetObservationsV2Query = z.object({
```
maxdeichmann, line 277: Please create the API docs also in fern. See here: https://github.com/langfuse/langfuse/blob/56651a18d9fcfb9ad90dbcf1e14a4c417400b8d4/fern/apis/server/definition/observations.yml#L1

web/src/features/public-api/types/observations.ts
```diff
@@ -205,3 +210,123 @@ export const GetObservationV1Query = z.object({
...
+      return v
+        .split(",")
+        .map((f) => f.trim())
+        .filter((f) =>
+          OBSERVATION_FIELD_GROUPS.includes(f as ObservationFieldGroup),
+        );
+    })
+    .pipe(z.array(z.enum(OBSERVATION_FIELD_GROUPS)).nullable()),
+  // Pagination
+  limit: z.coerce.number().nonnegative().lte(10000).default(50),
```
maxdeichmann: lets keep a max as low as 1000 or so? This way we protect us from event loop issues.

web/src/features/public-api/types/observations.ts
```diff
@@ -205,3 +210,123 @@ export const GetObservationV1Query = z.object({
...
 });
 export const GetObservationV1Response = APIObservation;
+
+/**
+ * Cursor schema for v2 observations pagination
+ * Encodes the position in the result set using the table's ordering:
+ * (start_time, xxHash32(trace_id), span_id)
+ */
+export const ObservationsCursorV2 = z.object({
+  lastStartTimeTo: z.coerce.date(),
```
maxdeichmann, line 226: This is the last startTmie that was returned in the previous API call, right?

web/src/features/public-api/types/observations.ts
```diff
@@ -205,3 +210,123 @@ export const GetObservationV1Query = z.object({
...
+        .split(",")
+        .map((f) => f.trim())
+        .filter((f) =>
+          OBSERVATION_FIELD_GROUPS.includes(f as ObservationFieldGroup),
+        );
+    })
+    .pipe(z.array(z.enum(OBSERVATION_FIELD_GROUPS)).nullable()),
+  // Pagination
+  limit: z.coerce.number().nonnegative().lte(10000).default(50),
+  withCursor: EncodedObservationsCursorV2.optional(),
```
maxdeichmann: Maybe `cursor` only?

packages/shared/src/server/repositories/events.ts
Steffen911: Maybe we log as a refactoring idea to create a new `shared/types` space? And in there we can have subfolders for API, data model, etc. with corresponding mapper functions. IMO that would be the perfect separation here and make mappers and types accessible anywhere.

Steffen911: What do both of you think? And should be log this as a good improvement?

packages/shared/src/server/repositories/events.ts
```diff
@@ -56,44 +59,98 @@ type ObservationsTableQueryResultWitouhtTraceFields = Omit<
...
+ * @param parseIoAsJson - Whether to parse input/output as JSON
+ * @param requestedFields - Field groups for V2 API (undefined/null = V1 API, returns complete observations)
  */
-const enrichObservationsWithModelData = async (
+async function enrichObservationsWithModelData<
+  F extends ObservationFieldGroup[] | null,
+  R = F extends ObservationFieldGroup[]
+    ? ObservationV2
+    : Observation & ObservationPriceFields,
+>(
```
Steffen911: Nice!

packages/shared/src/server/repositories/events.ts
sumerman: I agree in general. Right now there are splits between `domain`, validation types, etc. In the scope of this PR this should probably go into `domain`.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
maxdeichmann: Yes, agree. We somehow need to tame the "definitions" overload in our codebase.

web/src/features/public-api/types/observations.ts
sumerman: I don't think we should merge them yet (this is not going to be live until events double writes / double reads / backfill are fully out). But I did prepare a separate PR https://github.com/langfuse/langfuse/pull/10547

web/src/features/public-api/types/observations.ts
sumerman: Yes. Start time from the last data item returned by the API. But it should be completely opaque to the end user.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: I would like to take it out of scope of this PR.

sumerman: I also have this draft of the next iteration of query builder which I would like to return to https://github.com/langfuse/langfuse/compare/main...valeriy/lfe-6995-declarative-builder#diff-067705a7093d59a7d405c7947459b16d2febb975f78d0172e5e4665c4addc08cR1

PR #10574, opened 2025-11-19
feat: adding MutationMonitor component that pauses/resumes TraceDelete based on mutations count

sumerman:


worker/src/features/mutation-monitoring/mutationMonitor.ts
```diff
@@ -0,0 +1,243 @@
...
+  /**
+   * Check mutations and schedule the next check (ensures sequential execution)
+   */
+  private static async checkMutationsAndScheduleNext(): Promise<void> {
+    try {
+      await this.checkMutations();
+    } finally {
+      // Always schedule next check, even if current check failed
+      this.scheduleNextCheck();
+    }
```
Steffen911, line 88: Could this raise an uncaught exception in case the checkMutations raises anything? Or does the finally catch this sufficiently?

sumerman: I am manually catching errors pretty much everywhere else, but this is a good one. thanks! https://github.com/langfuse/langfuse/pull/10578

PR #10588, opened 2025-11-19
fix(ingestion): allow environment field to be updated during trace merging

### LINKED ISSUE #10436: bug: environment always set to default
### Describe the bug

Despite that i have tried setting `LANGFUSE_TRACING_ENVIRONMENT` to `staging` or even directly passing `environment='staging'` when initialising `Langfuse()`, the traces on the UI are bind to the `default` environment. 

### Steps to reproduce

1. Set `LANGFUSE_TRACING_ENVIRONMENT` to `staging`
2.  Decorate a generation function with `@observe`

or 

```python3
from langfuse import Langfuse

client = Langfuse(
    ....,
    environment='staging'
)

@observe
def my_func():
     ....

```

Traces on the UI are bind to the `default` environment. 

Langfuse Cloud

_No response_

Python client, `v3.9.2`. 

### Additional information

_No response_

No

4thel00z:
## Description

Fixes #10436

This PR fixes a bug where the `environment` field was always set to `default` even when explicitly specified as `staging` or other values in the Python SDK.

## Root Cause

The `environment` field was incorrectly marked as **immutable** in the `IngestionService`. This meant that once it was set (even implicitly to `default`), subsequent events couldn't update it.

## Solution

Removed `environment` from the `immutableEntityKeys` arrays for:
- Traces
- Scores  
- Observations

This allows the environment field to be updated during the merge process while still defaulting to `default` when not specified.

## Testing

Added comprehensive regression tests in `worker/src/__tests__/regressions/issue_10436_environment.test.ts`:

✅ All 4 tests passing:
1. Environment can be updated from `default` → `staging`
2. Environment defaults to `default` when not specified
3. Environment preserves `staging` when explicitly set from the start
4. Environment preserves `production` when explicitly set

## Commits

1. **test(ingestion): add regression tests for environment field updates** - Tests added (1 failing, 3 passing)
2. **fix(ingestion): remove environment from immutable keys to allow updates** - Fix applied (all 4 tests passing)

The fix is minimal and focused, changing only 10 lines to remove the immutability constraint.

### DIFF
--- a/worker/src/__tests__/regressions/issue_10436_environment.test.ts
+++ b/worker/src/__tests__/regressions/issue_10436_environment.test.ts
@@ -279,4 +279,62 @@ describe("Issue #10436: Environment ingestion bug", () => {
     // Should preserve the explicitly set production environment
     expect(traces[0].data.environment).toBe("production");
   });
+
+  it("should preserve early environment when later events have undefined environment (Steffen911 concern)", async () => {
+    const traceId = uuidv4();
+    const projectId = "project-123";
+    const now = new Date();
+
+    // Event 1: Early event with explicit staging environment
+    const traceCreate = {
+      id: uuidv4(),
+      type: eventTypes.TRACE_CREATE,
+      timestamp: now.toISOString(),
+      body: {
+        id: traceId,
+        name: "trace-early-env",
+        environment: "staging",
+      },
+    };
+
+    // Event 2: Later event without environment field (undefined)
+    const traceUpdate1 = {
+      id: uuidv4(),
+      type: eventTypes.TRACE_CREATE,
+      timestamp: new Date(now.getTime() + 100).toISOString(),
+      body: {
+        id: traceId,
+        name: "trace-early-env-updated",
+        // environment is undefined here
+      },
+    };
+
+    // Event 3: Another update without environment
+    const traceUpdate2 = {
+      id: uuidv4(),
+      type: eventTypes.TRACE_CREATE,
+      timestamp: new Date(now.getTime() + 200).toISOString(),
+      body: {
+        id: traceId,
+        output: { result: "done" },
+        // environment is undefined here too
+      },
+    };
+
+    await ingestionService.mergeAndWrite(
+      "trace",
+      projectId,
+      traceId,
+      now,
+      [traceCreate, traceUpdate1, traceUpdate2] as any[],
+      false,
+    );
+
+    const traces = mockClickhouseWriter.queue[TableName.Traces];
+    expect(traces).toHaveLength(1);
+
+    // Should preserve the original "staging" environment from the early event
+    // because later events have undefined environment (not explicit "default")
+    expect(traces[0].data.environment).toBe("staging");
+  });
 });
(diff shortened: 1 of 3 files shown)


worker/src/services/IngestionService/index.ts
```diff
@@ -162,28 +162,20 @@ const immutableEntityKeys: {
   [TableName.Observations]: (keyof ObservationRecordInsertType)[];
   [TableName.DatasetRunItems]: (keyof DatasetRunItemRecordInsertType)[];
 } = {
-  [TableName.Traces]: [
-    "id",
-    "project_id",
-    "timestamp",
-    "created_at",
-    "environment",
```
Steffen911, line 170: One problem that we've observed is that sometimes an "early" event has the environment set with subsequent events missing it. In that situation, it can happen that the original "correct" environment is being overwritten.

Could you check if your patch addresses this correctly as well?

4thel00z: Hey @Steffen911,
so for my understanding:

You want me to check if a theoretical stream of events as such:

```
ev0,env="foo"
ev1,env=""
ev2,env=""
ev3,env=""
```
results in```env=""```, correct ?

alphamt: Added one additional testcase to address this concern and one integration test to make sure that we did not mock too much.

alphamt: Find here an image describing the fix
image

4thel00z: Find attached an overview of different other conditions that might occur on the ingestion path. Some are related to other bugs I have seen in the issue tracker.

Some might be actual genuine behaviour (Postgres write):

Ingestion Bugs Part 1
Ingestion Bugs Part 2
Ingestion Bugs Part 3
Ingestion Bugs Part 3

worker/src/__tests__/regressions/issue_10436_environment.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    const traceUpdate2 = {
+      id: uuidv4(),
+      type: eventTypes.TRACE_CREATE,
+      timestamp: new Date(now.getTime() + 200).toISOString(),
+      body: {
+        id: traceId,
+        output: { result: "done" },
+        // environment is undefined here too
+      },
+    };
```
Steffen911, line 322: This assumption does not hold. The environment cannot be undefined since the environment schema defaults to `default`:
```
const PublicEnvironmentName = z
  .string()
  .toLowerCase()
  .max(40, "Maximum length is 40 characters")
  .regex(/^(?!langfuse)[a-z0-9-_]+$/, ENVIRONMENT_NAME_REGEX_ERROR_MESSAGE)
  .default("default");
``` 
Therefore, a later environment without an explicit value _via the API_ should always be overwritten.

Steffen911 (changes requested):
I think the current behaviour is just as good as the proposal given that updates in both directions can be unintentional. We strongly encourage users to configure their environment centrally within the SDKs or using OTel and to me this seems like the most reliable way to achieve the correct result.

Hence, I'll go ahead and close this.

Thank you a lot for giving this a shot and the extensive documentation of your thoughts.

PR #10606, opened 2025-11-20
feat(model-prices): add model pricing tiers

hassiebp:


web/src/__tests__/async/model-pricing-tiers.servertest.ts
```diff
@@ -0,0 +1,1023 @@
...
+        "Invalid regex syntax",
+      );
+    });
+
+    it("should reject patterns that may cause catastrophic backtracking", () => {
+      // Known unsafe pattern (exponential backtracking)
+      expect(() => validateRegexPattern("(a+)+b")).toThrow(
+        "catastrophic backtracking",
+      );
+      expect(() => validateRegexPattern("(x+x+)+y")).toThrow(
```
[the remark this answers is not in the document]
hassiebp: This is a test case that tests exactly for non-performant regex detection.

fern/apis/server/definition/commons.yml
```diff
@@ -479,14 +488,210 @@ types:
...
+      Pricing tier definition with conditional pricing based on usage thresholds.
+
+      Pricing tiers enable accurate cost tracking for LLM providers that charge different rates based on usage patterns.
+      For example, some providers charge higher rates when context size exceeds certain thresholds.
+
+      How tier matching works:
+      1. Tiers are evaluated in ascending priority order (priority 1 before priority 2, etc.)
+      2. The first tier where ALL conditions match is selected
+      3. If no conditional tiers match, the default tier is used as a fallback
+      4. The default tier has priority 0 and no conditions
```
Steffen911, line 573: Priority 0 doesn't make sense theoretically. It would need priority infinity.

packages/shared/clickhouse/migrations/clustered/0031_add_usage_pricing_tier_columns.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE observations ON CLUSTER default ADD COLUMN usage_pricing_tier_id Nullable(String) AFTER total_cost;
+ALTER TABLE observations ON CLUSTER default ADD COLUMN usage_pricing_tier_name Nullable(String) AFTER usage_pricing_tier_id;
```
Steffen911, line 2: Can the tier names change with the ID remaining the same? This could produce discrepancies here and might be risky.

packages/shared/prisma/migrations/20251120105316_add_pricing_tiers/migration.sql
```diff
@@ -0,0 +1,59 @@
...
+  SELECT DISTINCT model_id
+  FROM prices
+  WHERE pricing_tier_id IS NULL
+) AS distinct_models
+ON CONFLICT (id) DO NOTHING;
+
+-- Data Migration: Link existing prices to their default tiers using deterministic IDs
+UPDATE prices
+SET pricing_tier_id = model_id || '_tier_default'
+WHERE pricing_tier_id IS NULL;
```
Steffen911, line 49: Let's consider doing that in a separate migration. That way, we split data movement errors from DDL errors. Do we expect this to be fast even for thousands of records?

packages/shared/prisma/migrations/20251120105316_add_pricing_tiers/migration.sql
```diff
@@ -0,0 +1,59 @@
...
+
+-- AlterTable: Make pricing_tier_id NOT NULL
+ALTER TABLE "prices" ALTER COLUMN "pricing_tier_id" SET NOT NULL;
+
+-- AddForeignKey
+ALTER TABLE "prices" ADD CONSTRAINT "prices_pricing_tier_id_fkey" FOREIGN KEY ("pricing_tier_id") REFERENCES "pricing_tiers"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- DropIndex
+DROP INDEX IF EXISTS "prices_model_id_usage_type_key";
```
Steffen911, line 58: Is this intentional?

packages/shared/src/features/model-pricing/validation.ts
```diff
@@ -0,0 +1,251 @@
...
+      "Pattern may cause catastrophic backtracking. Please simplify your regex.",
+    );
+  }
+}
+
+/**
+ * Pricing tier condition schema (shared across API and tRPC)
+ */
+export const PricingTierConditionSchema = z.object({
+  usageDetailPattern: z
```
Steffen911, line 45: Is the naming expected? What is usageDetail in this case?

packages/shared/src/features/model-pricing/validation.ts
```diff
@@ -0,0 +1,251 @@
...
+export const PricingTierInputSchema = z.object({
+  name: z
+    .string()
+    .min(1, "Name cannot be empty")
+    .max(100, "Name exceeds maximum length of 100 characters"),
+  isDefault: z.boolean().default(false),
+  priority: z
+    .number()
+    .int("Priority must be an integer")
+    .min(0, "Priority must be non-negative")
```
Steffen911, line 83: I would force 1+ here as 0 should only be done in the isDefault case. Perhaps we can combine that smartly somehow.
Otherwise, all good.

packages/shared/prisma/migrations/20251120105316_add_pricing_tiers/migration.sql
```diff
@@ -0,0 +1,59 @@
...
+
+-- AddForeignKey
+ALTER TABLE "pricing_tiers" ADD CONSTRAINT "pricing_tiers_model_id_fkey" FOREIGN KEY ("model_id") REFERENCES "models"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AlterTable
+ALTER TABLE "prices" ADD COLUMN "pricing_tier_id" TEXT;
+
+-- Data Migration: Create default pricing tiers with DETERMINISTIC IDs
+-- Pattern: {model_id}_tier_default
+INSERT INTO "pricing_tiers" (id, model_id, name, is_default, priority, conditions)
```
Steffen911, line 31: As discussed, let's move those statements into a dedicated transaction block with BEGIN and COMMIT.

fern/apis/server/definition/commons.yml
hassiebp: Yes, but value infinity comes with gotchas to handle across our stack. thus chose value zero as reserved for default.

packages/shared/clickhouse/migrations/clustered/0031_add_usage_pricing_tier_columns.up.sql
hassiebp: Yes, in this case we would maintain the tier name at the moment of ingestion. I think this is a defensible standpoint.

packages/shared/src/features/model-pricing/validation.ts
hassiebp: This is expected, this stands for a regex pattern agains we match usage detail keys and sum up their values. `usageDetailKeyPattern` would be more elaborate but opted for simpler `usageDetailPattern`

packages/shared/src/features/model-pricing/validation.ts
hassiebp: The `UpsertModelSchema` and `PostModelV1Schema` have both in the refine step the `validatePricingTier` function that verifies that only the default can be zero and does full validation on the _list_ of pricing tiers

PR #10622, opened 2025-11-20
refactor(layout): modernize layout architecture and fix publishable path access

FroeMic:
## Summary

Complete rewrite of the master layout component to improve maintainability while fixing critical bugs preventing unauthenticated access to shared traces and sessions.

## Problem

The original `layout.tsx` (493 lines) was a "God Component" handling 8+ responsibilities:
- Authentication state management
- Authorization checks (RBAC, entitlements)
- Navigation filtering and processing
- Routing and redirects
- Metadata management
- Feature flag evaluation
- Project access verification
- Layout variant selection

This made it difficult to:
- Test individual concerns in isolation
- Understand the flow of logic
- Make changes without unintended side effects
- Debug authentication/authorization issues

Additionally, two critical bugs existed:
1. **Project access check blocked all unauthenticated users** on publishable paths (traces, sessions)
2. **Layout rendering crashed** when unauthenticated users accessed publishable paths

### Architecture Refactor

Replaced the monolithic layout with a modular architecture:

**Custom Hooks (Single Responsibility):**
- `useAuthSession` - Session management with retry logic
- `useLayoutConfiguration` - Determines layout variant based on path/session
- `useAuthGuard` - Authentication checks and redirect logic
- `useProjectAccess` - Project membership verification
- `useFilteredNavigation` - Navigation filtering with RBAC/entitlements
- `useLayoutMetadata` - Page titles and favicon generation

**Utilities:**
- `pathClassification.ts` - Path categorization (publishable, auth, public)
- `navigationFilters.ts` - Composable, testable filter functions
- `navigationFilters.types.ts` - Type definitions

**Layout Variants:**
- `LoadingLayout` - Loading/redirecting states
- `UnauthenticatedLayout` - Sign-in/sign-up pages
- `MinimalLayout` - Onboarding, public routes, publishable paths (unauthenticated)
- `AuthenticatedLayout` - Full app with sidebar and navigation

### Bug Fixes

**1. Publishable Path Access**
- Updated project access check to only run for authenticated users on non-publishable routes
- Publishable paths (traces, sessions) now accessible to anyone with the link

**2. Layout Rendering Crash**
- Added conditional rendering: unauthenticated users on publishable paths get `MinimalLayout`
- Prevents null session data from being passed to `AuthenticatedLayout`

**3. Security Integration**
- Integrated `getSafeRedirectPath` utility from security fix branch
- Replaced DOMPurify + manual validation with comprehensive redirect protection
- Prevents open redirect attacks (protocol-relative URLs, absolute URLs, JS URIs)

### Files Changed
- **17 files**: +1,130 insertions, -486 deletions
- Deleted: `layout.tsx` (493 lines)
- Created: 15 new focused files in `app-layout/` directory
- Modified: `_app.tsx` (feature flag integration)

### Key Improvements

**Maintainability:**
- ✅ Reduced complexity through separation of concerns
- ✅ Each hook handles one responsibility
- ✅ Composable, pure filter functions
- ✅ Clear data flow and dependencies

**Testability:**
- ✅ All hooks are independently testable
- ✅ Filter functions are pure (input → output, no side effects)
- ✅ Memoization prevents unnecessary recalculations

**Security:**
- ✅ Integrated secure redirect validation
- ✅ Prevents open redirect vulnerabilities
- ✅ Proper basePath handling

**UX:**
- ✅ Unauthenticated users can view shared traces/sessions
- ✅ Clean minimal UI for public content
- ✅ No crashes on publishable paths

### Manual Testing Performed
- ✅ Unauthenticated access to shared trace URLs
- ✅ Unauthenticated access to shared session URLs
- ✅ Authenticated user navigation and RBAC
- ✅ Sign-in/sign-up redirects with targetPath
- ✅ Project access denied for unauthorized users
- ✅ Feature flags and entitlements filtering
- ✅ Build passes with no TypeScript errors

### Test Scenarios
1. **Logged out + shared trace URL** → Shows MinimalLayout with trace content
2. **Logged out + non-publishable URL** → Redirects to sign-in with targetPath
3. **Logged in + project member** → Shows AuthenticatedLayout with full navigation
4. **Logged in + non-member** → Shows "Unauthorized" error page
5. **Auth page + authenticated** → Redirects to targetPath or home

## Migration Notes

- Feature flag enabled by default in `_app.tsx`
- Original `layout.tsx` deleted, replaced by `app-layout/` directory
- No API changes - all existing routes continue to work
- Session management and RBAC logic unchanged

## Related

- Fixes layout maintainability issues
- Resolves shared trace/session access bugs
- Integrates security fix from #10621

----

> [!IMPORTANT]
> Refactor layout architecture to improve maintainability and fix unauthenticated access issues, introducing modular components and custom hooks.
> 
>   - **Architecture Refactor**:
>     - Replaced monolithic `layout.tsx` with modular architecture in `app-layout/`.
>     - Introduced custom hooks: `useAuthSession`, `useLayoutConfiguration`, `useAuthGuard`, `useProjectAccess`, `useFilteredNavigation`, `useLayoutMetadata`.
>     - Added utilities: `pathClassification.ts`, `navigationFilters.ts`, `navigationFilters.types.ts`.
>     - Created layout variants: `LoadingLayout`, `UnauthenticatedLayout`, `MinimalLayout`, `AuthenticatedLayout`.
>   - **Bug Fixes**:
>     - Fixed unauthenticated access to publishable paths (traces, sessions).
>     - Prevented layout rendering crash for unauthenticated users on publishable paths.
>     - Integrated `getSafeRedirectPath` for secure redirects.
>   - **Files Changed**:
>     - Deleted `layout.tsx` (493 lines).
>     - Created 15 new files in `app-layout/`.
>     - Modified `_app.tsx` for feature flag integration.
>   - **Testing**:
>     - Manual tests for unauthenticated access, navigation, and redirects.
>     - Test scenarios for various user states and paths.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 9dbda3adf4bf13e9874e2a60dd9668810dab7060. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/components/layouts/app-layout/components/ResizableContent.tsx
marliessophie, line 1: Generally seems rather low risk hence only skimmed over. Please make sure to test this on your phone when it's on staging before deploying to prod. Sometimes behaves differently on mobile device vs resized chrome.

web/src/components/layouts/app-layout/components/ResizableContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Resizable content for support drawer on the right side of the screen (desktop).
+ * On mobile, renders a Drawer instead of a resizable sidebar.
+ *
+ * Key optimization: Always renders ResizablePanelGroup on desktop to prevent
+ * remounting children when drawer opens/closes. Uses refs for programmatic control.
+ */
+export function ResizableContent({ children }: PropsWithChildren) {
+  const { open, setOpen } = useSupportDrawer();
+  const isDesktop = useMediaQuery({ query: "(min-width: 768px)" });
```
marliessophie: We have `useIsMobile` hook, please reuse if possible.

web/src/components/layouts/app-layout/hooks/useAuthSession.ts
```diff
@@ -0,0 +1,49 @@
...
+  const session = useSession();
+
+  useEffect(() => {
+    if (session.status === "unauthenticated" && retryCount < MAX_RETRIES) {
+      const fetchSession = async () => {
+        try {
+          await getSession({ broadcast: true });
+        } catch (error) {
+          console.error("Error fetching session:", error);
+          // Don't throw - let it retry
```
marliessophie: We do not throw as we are retrying an ambiguous state that might resolve, correct?
What causes this behaviour, the api/auth/session endpoint? If so, I'd add a comment here.

web/src/components/layouts/app-layout/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // Authenticated layout
+  const handleSignOut = () => {
+    void signOut({ redirect: false });
+  };
+
+  return (
+    <AuthenticatedLayout
+      session={session.data!}
```
marliessophie: Is this type assertion safe?

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,97 @@
...
+      { name: "Theme", onClick: () => {}, content: <ThemeToggle /> },
+      { name: "Sign out", onClick: onSignOut },
+    ],
+  };
+
+  return (
+    <>
+      <Head>
+        <title>{metadata.title}</title>
+        <link rel="icon" href={metadata.faviconPath} />
```
marliessophie: Are we safe to drop sizes?

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <SidebarProvider>
+          <div className="flex h-dvh w-full flex-col">
+            <PaymentBanner />
+            <div className="flex min-h-0 flex-1 pt-banner-offset">
+              <AppSidebar
+                navItems={navigation.mainNavigation}
+                secondaryNavItems={navigation.secondaryNavigation}
+                userNavProps={userNavProps}
+              />
+              <SidebarInset>
```
marliessophie: Can you please verify we can drop all tailwind css classes here? Unsure what impact this has but assume they used to be there for a reason?

web/src/components/layouts/app-layout/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+      navigation={navigation}
+      metadata={metadata}
+      onSignOut={handleSignOut}
```
marliessophie, line 140: Could you please provide more context on how the new sign out works? This is one of the largest differences in functionality.

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * - Dynamic page metadata
+ */
+export function AuthenticatedLayout({
+  children,
+  session,
+  navigation,
+  metadata,
+  onSignOut,
+}: AuthenticatedLayoutProps) {
+  const user = session.user!;
```
marliessophie: Unsure about this type assertion?

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { ResizableContent } from "../components/ResizableContent";
+import { ThemeToggle } from "@/src/features/theming/ThemeToggle";
+import type { Session } from "next-auth";
+
+type AuthenticatedLayoutProps = PropsWithChildren<{
+  session: Session;
+  navigation: {
+    mainNavigation: any;
+    secondaryNavigation: any;
+    navigation: any[];
```
marliessophie: Can we do better with typing here? Also the naming is not ideal, navigation.navigation is confusing (I know I initially wrote this so this is a self-roast).

web/src/components/layouts/app-layout/utils/navigationFilters.ts
```diff
@@ -0,0 +1,216 @@
...
+    // If all children were filtered out, hide parent too
+    if (filteredItems.length === 0) {
+      return null;
+    }
+
+    filtered = { ...filtered, items: filteredItems };
+  }
+
+  return filtered;
+}
```
marliessophie, line 207: What do you think of doing something like the below? Easier to read imo

```
// Apply filters in sequence - chain short-circuits on first null
  const filterChain = [
    filters.projectScope,
    filters.organizationScope,
    filters.uiCustomization,
    filters.featureFlags,
    filters.entitlements,
    filters.projectRbac,
    filters.organizationRbac,
    (r: Route) => filters.customShow(r, ctx, organization),
  ];

  let filtered: Route | null = route;
  for (const filter of filterChain) {
    filtered = filter(filtered, ctx);
    if (!filtered) return null;
  }
```

web/src/components/layouts/app-layout/utils/navigationFilters.ts
```diff
@@ -0,0 +1,216 @@
...
+    if (!route.organizationRbacScope || !ctx.routerOrganizationId) return route;
+
+    const hasScope = hasOrganizationAccess({
+      session: ctx.session,
+      organizationId: ctx.routerOrganizationId,
+      scope: route.organizationRbacScope,
+    });
+
+    return hasScope ? route : null;
+  },
```
marliessophie, line 145: These all include a `cloudAdmin` check, why was this removed?

web/src/components/layouts/app-layout/hooks/useFilteredNavigation.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * Returns navigation split into main/secondary sections with active states
+ *
+ * @param session - Current user session
+ * @param organization - Current organization object
+ * @returns Processed navigation with main, secondary, and flattened arrays
+ */
+export function useFilteredNavigation(
+  session: Session | null,
+  organization: any,
```
marliessophie: Let's please avoid using any as typing where we can.

marliessophie (commented):
I got through half the code, need to finish rest tomorrow. This is critical code, need to be diligent. Please do not start addressing feedback before I get a chance to read through all code.

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @returns Object with hasAccess boolean and projectId
+ */
+export function useProjectAccess(session: Session | null) {
+  const router = useRouter();
+  const routerProjectId = router.query.projectId as string | undefined;
+
+  return useMemo(() => {
+    // No project in URL - access granted
+    if (!routerProjectId) {
+      return { hasAccess: true, projectId: undefined };
```
marliessophie, line 40: Intuitively it does not make sense to return hasAccess true if there is no project in the url, which project does the user have access to then?

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    // Demo project is always accessible
+    if (routerProjectId === env.NEXT_PUBLIC_DEMO_PROJECT_ID) {
+      return { hasAccess: true, projectId: routerProjectId };
+    }
```
marliessophie: This is new, are we sure we are protected against unauthenticated users? Possible that this hook is reused, so should be clear here.

web/src/components/layouts/app-layout/utils/pathClassification.ts
```diff
@@ -0,0 +1,88 @@
...
+  const isPublicPath = pathname.startsWith("/public/");
+  const isWithoutNavigation = PATH_CONSTANTS.withoutNavigation.some((path) =>
+    pathname.startsWith(path),
+  );
+  const isUnauthenticated = PATH_CONSTANTS.unauthenticated.some((path) =>
+    pathname.startsWith(path),
+  );
+  const isPublishable = PATH_CONSTANTS.publishable.some(
+    (path) =>
+      pathname === path || pathname.startsWith(path.replace("[projectId]", "")),
```
marliessophie: Can you explain why we do pathname.startsWith(path.replace("[projectId]", "") here? Thanks!

web/src/components/layouts/app-layout/utils/pathClassification.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+
+  // Determine if navigation should be hidden
+  const hideNavigation =
+    isPublicPath || isWithoutNavigation || isUnauthenticated;
+
+  // Determine variant (will be overridden by session status in hook)
```
marliessophie: I find this tight coupling hard to understand. why not just determine the variant in `useLayoutConfiguration`? This seems cleaner. Also I don't think the name "variant" is specific enough.

web/src/components/layouts/app-layout/hooks/useLayoutMetadata.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  return useMemo(() => {
+    const basePath = env.NEXT_PUBLIC_BASE_PATH ?? "";
+
+    // Determine page title from active route
+    const title = activePathName ? `${activePathName} | Langfuse` : "Langfuse";
+
+    // Use dev favicon in DEV region for visual distinction
+    const faviconPath =
+      region === "DEV" ? `${basePath}/icon-dev.svg` : `${basePath}/icon.svg`;
```
marliessophie, line 36: did we replace the pngs with svgs? cannot find the change in the PR but likely missing it

marliessophie: Also, the size indication was dropped here, making sure this was intended?

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+import { useRouter } from "next/router";
+import { useMemo } from "react";
+import { env } from "@/src/env.mjs";
+import { PATH_CONSTANTS } from "../utils/pathClassification";
+import { getSafeRedirectPath } from "@/src/utils/redirect";
+import type { useSession } from "next-auth/react";
+
+export type AuthGuardState =
```
marliessophie: See comment below, let's find one consistent variable name for auth states (maybe even single hook to return them; unsure). We use a mix of type and variant at the minute.

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (session.status === "loading") {
+      return { type: "loading", message: "Loading" };
+    }
+
+    const isUnauthPath = PATH_CONSTANTS.unauthenticated.some((p) =>
+      pathname.startsWith(p),
+    );
+    const isPublishable = PATH_CONSTANTS.publishable.some(
+      (p) =>
+        pathname === p || pathname.startsWith(p.replace("[projectId]", "")),
```
marliessophie: Again do not understand the second right side of the OR condition, this seems new. Could you explain please?

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    // Invalid user - has session but no DB user
+    // This can happen if user was deleted from DB but still has valid JWT
+    if (
+      session.data &&
+      session.data.user === null &&
+      !isUnauthPath &&
+      !isPublishable
```
marliessophie: Shouldn't this check for public? Used to do `router.pathname.startsWith("/public/");`

Please make sure no boolean checks are lost along the way.

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+      return { type: "sign-out", message: "Redirecting" };
+    }
+
+    // Unauthenticated user trying to access protected route
+    if (
+      session.status === "unauthenticated" &&
+      !isUnauthPath &&
+      !isPublishable
```
marliessophie: Same question, public check missing?

web/src/components/layouts/app-layout/hooks/useFilteredNavigation.ts
marliessophie, line 1: I think this file has some issues, here's a quick claude summary: 
Filters routes once, then re-iterates original ROUTES with O(n²) pathname lookup. Breaks nested routes.
**Fix:** `return filteredRoutes.map(mapRouteToNavigationItem);`

## HIGH: index.tsx (102-115)
`hideNavigation` check before `variant` check → auth pages render MinimalLayout instead of UnauthenticatedLayout. Lines 111-115 unreachable.

## MEDIUM: pathClassification.ts (48)
`path.replace("[projectId]", "")` creates `/project//traces/` (double slash), breaks publishable path matching.

marliessophie (commented):
Done with review, lmk if any questions! 

Thanks for cleaning this up, was indeed very suboptimal code.

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
FroeMic: The purpose of this hook in the context of the Layout component is to

(1) check whether we are in a projecct/[projectId] subroute
(2) if so, limit access on the client to those where the user has access

If we are in a routes without a projectId, this hook should generally not block it since we might be on a public page or organization page etc. 

=> add clarifying comment, and maybe rename so it is clear it should not be used somewhere else?

web/src/components/layouts/app-layout/components/ResizableContent.tsx
FroeMic: Replaced useMediaQuery from react-responsive with the existing useIsMobile
  hook from @/src/hooks/use-mobile. Will test on mobile device before deploying to prod.

web/src/components/layouts/app-layout/hooks/useAuthSession.ts
FroeMic: Added clarifying comment explaining that the /api/auth/session endpoint  occasionally fails with transient errors, and by not throwing we allow the retry logic to attempt recovery on the next iteration.

Note: this is an historical implementation I kept as-is

web/src/components/layouts/app-layout/index.tsx
FroeMic: Replaced the non-null assertion with an explicit runtime guard that returns
  <LoadingLayout> if session.data is unexpectedly null. This should never happen due to the auth
  guards above, but TypeScript is now satisfied without using !.

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
FroeMic: Restored proper favicon handling with both SVG and PNG variants. Now includes
   type="image/svg+xml" for modern browsers and a PNG fallback with sizes="256x256" for broader
  compatibility.

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
FroeMic: Replaced the non-null assertion with an explicit guard that returns null if
  user is unexpectedly missing. Added comment explaining that AuthenticatedLayout is only rendered
  after auth checks pass in AppLayout, guaranteeing session.user exists.

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
FroeMic: Replaced all any types with proper GroupedNavigation and NavigationItem[]
  types imported from the routes utilities. The confusing navigation.navigation naming is inherited
   from processNavigation - I've preserved the original API for now but added proper types. A
  follow-up refactor could rename this to something like navigation.items.

web/src/components/layouts/app-layout/utils/navigationFilters.ts
FroeMic: Refactored applyFiltersToRoute to use the filter chain pattern as suggested.
  Much cleaner and easier to read.

web/src/components/layouts/app-layout/hooks/useFilteredNavigation.ts
FroeMic: Added proper Organization type definition (User["organizations"][number] |  null | undefined) and updated the function signature to use it.

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
FroeMic: Added comprehensive JSDoc comments explaining the hook's purpose. When  routerProjectId is undefined, hasAccess: true is correct because this hook is specifically  designed to block access to project routes the user doesn't have permission for - it's not meant  to validate access to non-project routes (org pages, public pages, etc.).

FroeMic:
```
/**
 * Checks if the authenticated user has access to the project specified in the URL.
 * Used by AppLayout to show an error page when a user tries to access a project
 * they don't have permission for.
 *
 * Note: When routerProjectId is undefined (no project in URL), returns hasAccess=true
 * because this hook is only meant to block access to specific project routes,
 * not to validate access to non-project routes (org pages, public pages, etc.).
 *
 * @param session - Current user session (must be authenticated for meaningful check)
 * @returns Object with hasAccess boolean and projectId
 */
```

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
FroeMic: Added clarifying comment explaining this is safe because: (1) the demo  project contains only sample data, (2) API routes still enforce proper authentication for write  operations, and (3) this hook is client-side UX protection, not a security measure.

web/src/components/layouts/app-layout/utils/pathClassification.ts
FroeMic: The original pattern had a bug creating double slashes (/project//traces/). Replaced with a proper regex-based approach that converts path patterns like  /project/[projectId]/traces/[traceId] to regex patterns matching /project/[^/]+/traces/. Added  clear comments explaining the logic.

web/src/components/layouts/app-layout/utils/pathClassification.ts
FroeMic: Refactored to decouple path classification from variant determination:
 
  - classifyPath now returns PathClassification with isAuthPage, hideNavigation, isPublishable
  - useLayoutConfiguration handles all variant logic based on path classification + session state
  - Added clear JSDoc documenting the layout decision logic

web/src/components/layouts/app-layout/hooks/useLayoutMetadata.ts
FroeMic: The SVGs already exist in public (icon.svg). Now using both: SVG with  type="image/svg+xml" for modern browsers, PNG fallback with sizes="256x256" for broader  compatibility. Apple touch icon remains PNG.

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
FroeMic: Renamed AuthGuardState to AuthGuardResult and changed type to action for
  clarity. Now consistently uses "action" for auth guard states and "variant" for layout types.

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
FroeMic: Fixed the pathname matching to use proper regex (same fix as pathClassification.ts) and added the missing isPublicPath check. Public paths (/public/*) are now  properly excluded from authentication requirements.

FroeMic: new:

```
    const isPublishable = PATH_CONSTANTS.publishable.some((path) => {
      // Case 1: Exact match (e.g., pathname === "/auth/reset-password")
      if (pathname === path) return true;

      // Case 2: Prefix match for dynamic routes
      // Example: path = "/project/[projectId]/traces/[traceId]"
      //   -> pathPrefix = "/project/[^/]+/traces" (last segment removed, params converted to regex)
      //   -> matches pathname like "/project/abc123/traces/xyz789"
      // This allows shared trace/session links to be accessed without authentication
      const pathPrefix = path
        .split("/")
        .slice(0, -1)
        .join("/")
        .replace(/\[([^\]]+)\]/g, "[^/]+");
      const prefixRegex = new RegExp(`^${pathPrefix}/`);
      return prefixRegex.test(pathname);
    });
```

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
FroeMic: The `isPublicPath` check is present at line 52 and included in both condition checks (lines 80 and 90). 

```
  const isPublicPath = pathname.startsWith("/public/");
```

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
FroeMic: The `isPublicPath` check is present at line 52 and included in both condition checks (lines 80 and 90). 

```
  const isPublicPath = pathname.startsWith("/public/");
```

web/src/components/layouts/app-layout/hooks/useFilteredNavigation.ts
FroeMic: Reordered the checks so variant === "unauthenticated" is
  checked first, before hideNavigation. Auth pages now correctly render UnauthenticatedLayout.

FroeMic: Replaced the simple string replacement with a proper  regex-based approach that converts path patterns to regex segment matchers. No more double  slashes.

web/src/components/layouts/app-layout/index.tsx
FroeMic: Fixed to match the original behavior.

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
FroeMic: Good catch! The CSS classes were accidentally dropped. I've restored the original classes

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <SidebarProvider>
+          <div className="flex h-dvh w-full flex-col">
+            <PaymentBanner />
+            <div className="flex min-h-0 flex-1 pt-banner-offset">
+              <AppSidebar
+                navItems={navigation.mainNavigation}
+                secondaryNavItems={navigation.secondaryNavigation}
+                userNavProps={userNavProps}
+              />
+              <SidebarInset className="h-screen-with-banner max-w-full md:peer-data-[state=collapsed]:w-[calc(100vw-var(--sidebar-width-icon))] md:peer-data-[state=expanded]:w-[calc(100vw-var(--sidebar-width))]">
```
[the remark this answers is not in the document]
FroeMic: intended as is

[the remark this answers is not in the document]
FroeMic: no that works as is

web/src/components/layouts/app-layout/hooks/useAuthSession.ts
marliessophie: Makes sense. 

For correctness: previously this threw an error I think?

web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
marliessophie: sounds good, thx

web/src/components/layouts/app-layout/utils/pathClassification.ts
marliessophie: Sure, thanks. Let's make sure to test this though or even add unit test to avoid such problems in the future.

web/src/components/layouts/app-layout/hooks/useAuthGuard.ts
marliessophie: I still don't see it in either place :( But if you are sure it's there I trust you.

web/src/components/layouts/app-layout/hooks/useFilteredNavigation.ts
marliessophie: Great! So issues and performance are fixed then I presume?

web/src/components/layouts/app-layout/hooks/useProjectAccess.ts
marliessophie: I do not know how we handled this before, could you please double check to be authenticated was not a requirement to view demo project? The API routes I believe largely require an authed user for reads also.

web/src/components/layouts/app-layout/utils/pathClassification.ts
marliessophie: Cool. 

I will say still not a fan of the naming "variant" but up to you.

marliessophie (commented):
Completed another re-review ie went through all comments. Did not read all code again, should I?
Also I did not test this, please make sure to be very through with testing on this one :)

PR #10629, opened 2025-11-21
feat(filters): add comment filtering to traces, sessions, and observations

FroeMic:
## Summary
- Adds comment filtering capability to traces, sessions, and observations tables
- Users can now filter by comment count (>=, <=, =, >, <) and comment content (contains, does not contain, starts with, ends with)
- Uses PostgreSQL full-text search with GIN index for efficient comment content searches
- Two-phase query pattern: queries PostgreSQL for matching object IDs, then injects as filter into main query

## Changes
- New `commentFilterHelpers.ts` with reusable filtering logic for all object types
- Added `processCommentFilters()` to traces, sessions, and observations routers
- Added `commentCount` and `commentContent` column definitions to table definitions
- Added filter facets to traces-config.ts, sessions-config.ts, observations-config.ts
- Updated batch export warnings to note that comment filters aren't included in exports
- Added database migration for GIN index on comments content column

## Test plan
- [x] Server tests for traces comment filtering (9 tests)
- [x] Server tests for sessions comment filtering (7 tests)
- [x] Server tests for observations comment filtering (7 tests)
- [ ] Manual testing of filter UI in traces/sessions/observations pages

----

> [!IMPORTANT]
> Adds comment filtering to traces, sessions, and observations with support for count and content filters, including database and code updates for efficient querying.
> 
>   - **Behavior**:
>     - Adds comment filtering to traces, sessions, and observations, allowing filtering by comment count and content.
>     - Supports operators: `>=`, `<=`, `=`, `>`, `<` for count and `contains`, `does not contain`, `starts with`, `ends with` for content.
>     - Uses PostgreSQL full-text search with GIN index for content filtering.
>     - Two-phase query: fetches matching IDs from PostgreSQL, then filters main query.
>   - **Code Changes**:
>     - New `commentFilterHelpers.ts` for reusable logic.
>     - Updates to `traces.ts`, `sessions.ts`, `getAllQueries.ts` to apply comment filters.
>     - Adds `commentCount` and `commentContent` to table definitions in `eventsTable.ts`, `observationsTable.ts`, `sessionsView.ts`.
>     - Database migration for GIN index on comments content.
>   - **Tests**:
>     - Adds server tests for comment filtering in `traces`, `sessions`, `observations`.
>     - Tests cover count, content, combined filters, and edge cases.
>   - **Misc**:
>     - Updates export warnings in `BatchExportTableButton.tsx` to note comment filters are not included in exports.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for d25a5da0c05fd158c66e5301d645f543c3ef0865. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/migrations/20251126000000_add_comment_search_indexes/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Full-text search index for comment content filtering
+-- This enables efficient text search queries on comment content
+-- Uses PostgreSQL's built-in full-text search with GIN index
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_comments_content_gin
+ON comments USING gin(to_tsvector('english', content));
```
maxdeichmann, line 5: I think marlies merged some migrations today. You need to bump this to the the latest one.

packages/shared/src/observationsTable.ts
```diff
@@ -11,8 +11,9 @@ export const observationsTableCols: ColumnDefinition[] = [
   {
     name: "ID",
     id: "id",
-    type: "string",
+    type: "stringOptions",
```
maxdeichmann, line 15: Are you sure about this change?

web/src/features/comments/server/commentFilterHelpers.ts
```diff
@@ -0,0 +1,331 @@
...
+  value: number;
+}): Promise<string[]> {
+  // Validate operator to prevent SQL injection
+  const validOperators: CommentCountOperator[] = [
+    ">=",
+    "<=",
+    "=",
+    ">",
+    "<",
+    "!=",
```
maxdeichmann: We have such operators already defined somewhere. Do we really need to do this again?

web/src/features/comments/server/commentFilterHelpers.ts
```diff
@@ -0,0 +1,331 @@
...
+    SELECT object_id
+    FROM comments
+    WHERE project_id = ${projectId} AND object_type = ${objectType}::"CommentObjectType"
+    GROUP BY object_id
+    HAVING COUNT(*) ${Prisma.raw(operator)} ${value}
+  `;
+
+  const results = await prisma.$queryRaw<{ object_id: string }[]>(rawQuery);
+  return results.map((r) => r.object_id);
+}
```
maxdeichmann: These are classic repository functions. Please move into repo.

web/src/features/comments/server/commentFilterHelpers.ts
```diff
@@ -0,0 +1,331 @@
...
+  searchQuery,
+  operator = "contains",
+}: {
+  prisma: PrismaClient;
+  projectId: string;
+  objectType: CommentObjectType;
+  searchQuery: string;
+  operator?: CommentContentOperator;
+}): Promise<string[]> {
+  if (operator === "contains") {
```
maxdeichmann: We have one search abstraction already built for e.g. prompts. Can you check whether you can reuse here?

web/src/server/api/routers/generations/getAllQueries.ts
```diff
@@ -18,19 +19,74 @@ export type GetAllGenerationsInput = z.infer<typeof GetAllGenerationsInput>;
...
+        }
+
+        // Inject matching observation IDs as filter
+        filterWithComments.push({
+          type: "stringOptions",
+          operator: "any of",
+          column: "id",
+          value: matchingObjectIds,
+        });
+      }
```
maxdeichmann, line 55: You copy paste the same logic over and over. Can this be abstracted?

packages/shared/prisma/migrations/20251126000000_add_comment_search_indexes/migration.sql
FroeMic: renamed

web/src/features/comments/server/commentFilterHelpers.ts
FroeMic: Done. Created packages/shared/src/server/repositories/comments.ts with:
  - getObjectIdsByCommentCount()
  - getObjectIdsByCommentContent()
  - Type exports: CommentObjectType, CommentCountOperator, CommentContentOperator

  Exported from packages/shared/src/server/repositories/index.ts.

web/src/features/comments/server/commentFilterHelpers.ts
FroeMic: Refactored to reference the existing filterOperators from
  packages/shared/src/interfaces/filters.ts. The CommentContentOperator now maps directly to
  filterOperators.string.

  For CommentCountOperator, I kept a custom type that extends filterOperators.number with "!="
  since that operator isn't in the standard set but is useful for comment count filtering (e.g.,
   "show items with != 0 comments"). Added a comment explaining this.

web/src/features/comments/server/commentFilterHelpers.ts
FroeMic: Reviewed the prompt search pattern (buildPromptSearchFilter in promptRouter.ts). It uses ILIKE
   for text matching.

  The comment search implementation follows the same pattern for "starts with", "ends with", and
   "does not contain" operators. For the "contains" operator, I kept the PostgreSQL full-text
  search with GIN index since it's more performant for content searches across potentially large
   comment volumes. This hybrid approach is intentional - we get the benefits of full-text
  search for the most common operation while maintaining consistency with existing patterns for
  other operators.

packages/shared/src/observationsTable.ts
FroeMic: Yes, this change is intentional. The stringOptions type is needed to support the "any of"
  filter operator when injecting matching object IDs from comment filters. Without this, the
  comment filtering feature can't work because we push a filter like:
  { type: "stringOptions", operator: "any of", column: "id", value: matchingObjectIds }
  Added an inline comment to explain this.

web/src/server/api/routers/generations/getAllQueries.ts
FroeMic: Done. Created applyCommentFilters() helper in commentFilterHelpers.ts that encapsulates the
  entire pattern:

FroeMic: // Before: ~25 lines per router procedure
  ```
  const { updatedFilterState, matchingObjectIds } = await processCommentFilters({...});
  let filterWithComments = updatedFilterState;
  if (matchingObjectIds !== null) {
    if (matchingObjectIds.length === 0) { return { traces: [] }; }
    filterWithComments.push({ type: "stringOptions", operator: "any of", column: "id", value:
  matchingObjectIds });
  }
  ```
vs

  // After: ~8 lines per router procedure
```
  const { filterState, hasNoMatches } = await applyCommentFilters({...});
  if (hasNoMatches) { return { traces: [] }; }
  // Use filterState directly
```
  Updated all three routers (traces.ts, sessions.ts, getAllQueries.ts) to use the new
  abstraction.

packages/shared/src/observationsTable.ts
FroeMic: Triple-Checked and this change seems to be safe.

- **UI Filter Builder:** Users would now be able to select multiple IDs using "any of" / "none of" operators instead of just using string operators like "contains". **The ID column was and still is not filterable via the UI**
- **Saved Views:** Any saved views with ID filters – if they were created in the past – will continue to work since the type in the URL query string determines how it's decoded, not the column definition.
- **API Compatibility:** Public API endpoints that accept filters should work fine since they validate against the filter schemas, not against column types directly.

PR #10698, opened 2025-11-25
chore(dataset-versioning): add dual-write for dataset-item-events

marliessophie:
> [!IMPORTANT]
> Adds dual-write strategy for dataset item events and refactors dataset item management by removing `DatasetItemManager` class, updating API routes and tests accordingly.
> 
>   - **Behavior**:
>     - Introduces dual-write strategy for dataset item events in `executeWithDatasetServiceStrategy.ts`, writing to both stateful and versioned implementations.
>     - Adds environment variables `LANGFUSE_DATASET_SERVICE_WRITE_TO_VERSIONED_IMPLEMENTATION` and `LANGFUSE_DATASET_SERVICE_READ_FROM_VERSIONED_IMPLEMENTATION` in `env.ts`.
>   - **Refactoring**:
>     - Removes `DatasetItemManager` class, replaces with standalone functions `createDatasetItem`, `upsertDatasetItem`, `deleteDatasetItem`, and `createManyDatasetItems` in `dataset-items.ts`.
>     - Updates `DatasetItemValidator` to use `validateAndNormalize()` instead of `preparePayload()`.
>   - **API and Tests**:
>     - Updates API routes in `dataset-router.ts` and `pages/api/public/dataset-items` to use new dataset item functions.
>     - Modifies tests in `datasets-schema-validation.servertest.ts` to align with refactored validation logic.
>     - Adds `datasetItemEvent` deletion in `pruneDatabase()` in `test-utils.ts` and `utils.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 5c70e0ef86b19f4cf73fdb0a3ef8933528281c85. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -203,39 +269,79 @@ export class DatasetItemManager {
...
+    // 5. Prepare full item data for writing
+    const itemData = {
+      itemId: itemId,
+      input: itemPayload.input,
+      expectedOutput: itemPayload.expectedOutput,
+      metadata: itemPayload.metadata,
+      sourceTraceId: mergedItemData.sourceTraceId ?? undefined,
+      sourceObservationId: mergedItemData.sourceObservationId ?? undefined,
+      status: mergedItemData.status ?? undefined,
+    };
```
maxdeichmann: why do we need the ?? undefined here? Is it possible to return e.g. `string|undefined` also from `preparePayload` ? I assume that it returns  `string|null`

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -140,13 +146,48 @@ export class DatasetItemManager {
 
     const datasetItem = {
       ...result.datasetItems[0],
+      id: result.datasetItems[0].itemId,
```
maxdeichmann: I think having all these static functions is strange in typescript. I would only expect these if the manager is something like a singleton or needs lots of configuration for setup. In this case this is only a wrapper around functions, which is already taken care of by the JS module of which they are part of.

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -169,7 +210,13 @@ export class DatasetItemManager {
       normalizeOpts?: { sanitizeControlChars?: boolean };
       validateOpts: { normalizeUndefinedToNull?: boolean };
     } & IdOrName,
-  ): Promise<{ success: true; datasetItem: DatasetItem } | PayloadError> {
+  ): Promise<
+    | {
+        success: true;
```
maxdeichmann: Is there a reason to pass around the "success" variables? I would expect this function to just throw exceptions and the APIs to take these exceptions and return to the user.

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -182,7 +229,26 @@ export class DatasetItemManager {
...
+
+    // 3. Merge incoming data with existing data
+    // For fields where props value is undefined, use existing value
+    const mergedItemData = existingItem
+      ? this.mergeItemData(existingItem, props)
+      : props;
+
+    // 4. Validate merged payload
     const validator = new DatasetItemValidator({
       inputSchema: dataset.inputSchema as Record<string, unknown> | null,
```
maxdeichmann: Why do we need this casting?

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -192,9 +258,9 @@ export class DatasetItemManager {
     });
 
     const itemPayload = validator.preparePayload({
```
maxdeichmann: I think there can be better naming than "preparePayload" -- what is this function doing?

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
```diff
@@ -203,39 +269,79 @@ export class DatasetItemManager {
...
+        //   datasetId: res.datasetId,
+        //   status: res.status ?? DatasetStatus.ACTIVE,
+        //   input: res.input,
+        //   expectedOutput: res.expectedOutput,
+        //   metadata: res.metadata,
+        //   sourceTraceId: res.sourceTraceId,
+        //   sourceObservationId: res.sourceObservationId,
+        //   createdAt: res.createdAt ?? new Date(),
+        //   updatedAt: res.createdAt ?? new Date(),
+        // };
```
maxdeichmann: Pls. remove

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
marliessophie: Expectation of `ajv.compile` under the hood; artifact of dataset schema implementation. Would prefer to keep as is.

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
marliessophie: removed it.

packages/shared/src/server/services/DatasetService/DatasetItemManager.ts
marliessophie: we don't, removed.

PR #10768, opened 2025-11-28
chore: get events from s3

maxdeichmann:
> [!IMPORTANT]
> Introduces S3 recovery feature with a new queue, API endpoint, and job processor for handling recovery tasks.
> 
>   - **S3 Recovery Queue**:
>     - Adds `S3RecoveryQueue` class in `s3RecoveryQueue.ts` to manage S3 recovery jobs.
>     - Introduces `S3RecoveryJobSchema` in `queues.ts` for job validation.
>     - Registers `S3RecoveryQueue` in `getQueue.ts` and `index.ts`.
>   - **API Endpoint**:
>     - Adds `s3-recovery/index.ts` to handle POST requests for S3 recovery jobs.
>     - Validates requests using `S3RecoveryJobSchema` and queues jobs in `S3RecoveryQueue`.
>   - **Worker and Environment**:
>     - Implements `s3RecoveryQueueProcessor` in `s3RecoveryQueue.ts` to process jobs.
>     - Registers processor in `app.ts` with concurrency control.
>     - Adds `QUEUE_CONSUMER_S3_RECOVERY_QUEUE_IS_ENABLED` to `env.ts` for feature toggle.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 04f4cec84344ec71deae00939ce0e489e280f5a0. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/worker/src/queues/s3RecoveryQueue.ts
+++ b/worker/src/queues/s3RecoveryQueue.ts
@@ -1,0 +1,196 @@
+import { Job } from "bullmq";
+import { Readable } from "stream";
+import { randomUUID } from "crypto";
+import {
+  logger,
+  QueueName,
+  TQueueJobTypes,
+  getS3EventStorageClient,
+  StorageServiceFactory,
+  traceException,
+} from "@langfuse/shared/src/server";
+import { transformStreamToCsv } from "@langfuse/shared/src/server/utils/transforms/transformStreamToCsv";
+import { env } from "../env";
+
+interface TimeframeFilter {
+  startDate: Date;
+  endDate: Date;
+}
+
+/**
+ * Generate all possible S3 path prefixes for a given project and timeframe
+ * Path structure: {PREFIX}otel/{projectId}/{YYYY}/{MM}/{DD}/{HH}/{mm}/
+ */
+function generateS3Prefixes(
+  projectId: string,
+  timeframe: TimeframeFilter,
+): string[] {
+  const prefixes: string[] = [];
+  const prefix = env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX || "";
+
+  const current = new Date(timeframe.startDate);
+  const end = new Date(timeframe.endDate);
+
+  while (current <= end) {
+    const year = current.getFullYear();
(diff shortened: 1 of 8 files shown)


worker/src/queues/s3RecoveryQueue.ts
```diff
@@ -0,0 +1,196 @@
...
+      try {
+        logger.info(`Starting to list files for ${prefixes.length} prefixes`);
+
+        // Process prefixes in batches to avoid overwhelming S3
+        for (let i = 0; i < prefixes.length; i += BATCH_SIZE) {
+          const batch = prefixes.slice(i, i + BATCH_SIZE);
+
+          const batchResults = await Promise.all(
+            batch.map(async (prefix) => {
+              return await storageClient.listFiles(prefix);
```
Steffen911, line 78: ListFiles has a cap of 1000 results. If a client sends us more than 1000 requests in a single minute this would break down. IMO we'd be better off in generating that list from S3 access logs and using that for replay instead if possible.

Alternatively, we need to add a recursive solution here that checks if there is a "next" token.

worker/src/queues/s3RecoveryQueue.ts
```diff
@@ -0,0 +1,196 @@
...
+  logger.info(`Starting streaming upload to CSV: ${fileName}`);
+
+  // Create the S3 file stream
+  const fileStream = createS3FileStream(prefixes);
+
+  // Transform to CSV
+  const csvStream = fileStream.pipe(transformStreamToCsv());
+
+  // Upload to S3
+  const storageClient = StorageServiceFactory.getInstance({
```
Steffen911, line 127: Is this actually meant to be merged like this or a workaround?

maxdeichmann: 1. This PR was vibe coded
2. Yes, we should add something like this as a permanent solution to our code base. Hassieb took this over.

PR #10787, opened 2025-11-29
fix: otel event replay endpoint

hassiebp:
- **temp: add otel-ingestion-replay endpoint**
- **push**

----

> [!IMPORTANT]
> Adds OTEL ingestion replay endpoint, utility functions, and a script for batch processing OTEL events from S3.
> 
>   - **API Endpoint**:
>     - Adds `otel-ingestion-replay.ts` to handle OTEL ingestion replay requests, validating input and queuing jobs.
>   - **Utility Functions**:
>     - Adds `generateOtelS3Prefixes()` in `otelS3Utils.ts` for generating S3 prefixes based on time range and granularity.
>     - Adds `listFilesPaginated()` to `StorageService` for paginated file listing in S3, Azure, and Google Cloud.
>   - **Scripts**:
>     - Adds `otel-ingestion-replay` script in `worker/src/scripts/otel-ingestion-replay/index.ts` for batch processing OTEL events from S3.
>   - **Misc**:
>     - Removes `validatePublicKey()` from `OtelIngestionProcessor.ts`.
>     - Updates `package.json` to include new script and dependencies.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 39d1f3778f5b9a9f24e12c0b710f849fc930e700. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/otel/OtelIngestionProcessor.ts
```diff
@@ -955,6 +955,9 @@ export class OtelIngestionProcessor {
     scopeAttributes: Record<string, unknown>,
     resourceAttributes: Record<string, unknown>,
   ): void {
+    // Disable public key validation for replay
+    return;
```
maxdeichmann: @hassiebp will you roll this back or how do we deal with this going forward?

hassiebp: We will roll this back, this PR is never to be merged :)

packages/shared/src/server/s3/index.ts
```diff
@@ -38,3 +40,38 @@ export const getS3EventStorageClient = (bucketName: string): StorageService => {
...
+      forcePathStyle: env.LANGFUSE_S3_EVENT_UPLOAD_FORCE_PATH_STYLE === "true",
+    });
+  }
+
+  return {
+    client: s3EventStorageRawClient,
+    bucketName: env.LANGFUSE_S3_EVENT_UPLOAD_BUCKET,
+    prefix: env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX,
+  };
+};
```
maxdeichmann, line 54: I feel like this is a re-definition. We ususally instantiate S3 like this: https://github.com/langfuse/langfuse/blob/ffd324826755c09c441b966674c6cd0a157a6c90/packages/shared/src/server/repositories/clickhouse.ts#L86

web/src/pages/api/admin/otel-ingestion-replay.ts
```diff
@@ -0,0 +1,194 @@
...
+} from "@langfuse/shared/src/server";
+import { AdminApiAuthService } from "@/src/ee/features/admin-api/server/adminApiAuth";
+
+/*
+This API route is used to replay OTEL ingestion events from S3 for a specific project
+within a given time range. It lists S3 objects and queues them for reprocessing.
+*/
+
+const RequestSchema = z.object({
+  projectId: z.string(),
```
maxdeichmann, line 23: We could also make this a list of projectids

web/src/pages/api/admin/otel-ingestion-replay.ts
```diff
@@ -0,0 +1,194 @@
...
+      return res.status(405).json({ error: "Method Not Allowed" });
+    }
+
+    // Auth check - allowed on Langfuse Cloud
+    if (
+      !AdminApiAuthService.handleAdminAuth(req, res, {
+        isAllowedOnLangfuseCloud: true,
+      })
+    ) {
+      return;
```
maxdeichmann, line 48: We should return an error here

web/src/pages/api/admin/otel-ingestion-replay.ts
```diff
@@ -0,0 +1,194 @@
...
+          span.setAttribute("error", "queue_unavailable");
+          return res
+            .status(500)
+            .json({ error: "Failed to get OtelIngestionQueue" });
+        }
+
+        // Queue jobs in batches of 10,000
+        let totalQueued = 0;
+
+        for (let i = 0; i < allFiles.length; i += BATCH_SIZE) {
```
maxdeichmann, line 141: allFiled may grow too large

web/src/pages/api/admin/otel-ingestion-replay.ts
hassiebp: I think this should be controlled by the caller of the endpoint for better control

web/src/pages/api/admin/otel-ingestion-replay.ts
hassiebp: already handled by AdminApiAuthService.handleAdminAuth

web/src/pages/api/admin/otel-ingestion-replay.ts
hassiebp: allFiles is limited to 1_000_000 entries, see line 112

maxdeichmann:
IMO we should get this into a mergable PR that we can roll out:
- We need this functionality for disaster recovery. This needs to work at scale. Hence I would move the processing to the worker side and make sure it can work with many "allFiles"
- We also need to think about the publicKey situation. We do not have a chance to get the key in the processing that was used to ingest the data in the first place. Is the project_id from the S3 path safe to use? In this case we could use that to make the publickey check on the worker side.

PR #10817, opened 2025-12-01
chore(dataset-versioning): add version cols to dataset items model

marliessophie:
> [!IMPORTANT]
> Add versioning columns to `DatasetItem` model and update related schema, migration, and test files.
> 
>   - **Models**:
>     - Add `sys_id`, `valid_from`, and `is_deleted` columns to `DatasetItem` in `schema.prisma`.
>     - Update `status` column in `DatasetItem` to allow null values.
>   - **Migrations**:
>     - Add migration `20251201095227_dataset_items_add_version_cols` to add new columns and modify `status` column in `dataset_items` table.
>   - **Types**:
>     - Update `DatasetItem` type in `types.ts` to include new columns and nullable `status`.
>   - **Tests**:
>     - Update `evalService.test.ts` to include `sys_id` in `dataset_items` insertions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for bc7ca61bdca39813bbebec2e5937d557ee2afe79. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


worker/src/__tests__/evalService.test.ts
marliessophie, line 1: Explicitly setting sys_id required here as we are inserting into table directly without using standard prisma handler, `.create`

packages/shared/prisma/migrations/20251201095227_dataset_items_add_version_cols/migration.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+-- Step 1: Add columns
+ALTER TABLE "dataset_items" ADD COLUMN     "is_deleted" BOOLEAN NOT NULL DEFAULT false,
+ADD COLUMN     "sys_id" TEXT,
```
maxdeichmann: Is the @default(cid) taken care off by prisma? If not, this migration may break as we 1) ingest nullable data, 2) backfill, 3) have concurrent writes without sys_id set, 4) apply the constraint. We would fail in step 4.

marliessophie: Technically yes, but it's an application level constraint, not a DB constraint. 

We could do even better by changing this line to something like (applies to all INSERT statements): 
`ADD COLUMN     "sys_id" TEXT DEFAULT md5(random()::text || clock_timestamp()::text)::uuid::text,` 

and simply doing: 
`ALTER COLUMN "sys_id" DROP DEFAULT;`
in the next migration once application code is live. 

Will quickly verify if this works as expected. If it does, let's go with DB-level default for duration of migration as this is the safest option. We can rely on prismas cid going forward.

PR #10619, opened 2025-11-20
chore(dataset-versioning): add dataset item events table 

marliessophie:
> [!IMPORTANT]
> Add `dataset_item_events` table and related schema, migrations, and seeder updates for dataset item event tracking.
> 
>   - **Database Schema**:
>     - Add `dataset_item_events` table in `migration.sql` with fields for `id`, `item_id`, `project_id`, `dataset_id`, `status`, `input`, `expected_output`, `metadata`, `source_trace_id`, `source_observation_id`, `created_at`, `deleted_at`.
>     - Add foreign key constraint on `dataset_id` and `project_id` referencing `datasets` table.
>     - Add index on `project_id`, `dataset_id`, `item_id`, `created_at` in `add_idx_dataset_item_events/migration.sql`.
>   - **Prisma Schema**:
>     - Add `DatasetItemEvent` model in `schema.prisma` with fields matching the new table.
>     - Update `Dataset` model to include relation to `DatasetItemEvent`.
>   - **Seeder Scripts**:
>     - Add `seedDatasetVersions` function in `seed-dataset-versions.ts` to generate test data for `dataset_item_events`.
>     - Update `seed-postgres.ts` to call `seedDatasetVersions` for seeding data.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 506c85994cb5d2a58a3b21b90cec7e0b27c28776. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #10845, opened 2025-12-02
chore(dataset-versioning): READ path

marliessophie:
> [!IMPORTANT]
> Enhances dataset item handling with new versioning and retrieval methods, refactoring existing logic for improved efficiency and consistency.
> 
>   - **Behavior**:
>     - Adds unique index on `dataset_items` for `id`, `project_id`, and `valid_from` in `migration.sql`.
>     - Introduces `getDatasetItemsByLatest` and `getDatasetItemById` functions in `dataset-items.ts` for fetching latest dataset items and specific items by ID.
>     - Updates `validateAllDatasetItems` in `schemaValidation.ts` to use new dataset item retrieval methods.
>   - **Refactoring**:
>     - Moves dataset item filtering logic to `dataset-items-columns.ts`.
>     - Refactors dataset item retrieval in `dataset-router.ts` and `experimentServiceClickhouse.ts` to use new functions.
>     - Replaces direct database queries with `getDatasetItemsByLatest` in multiple files for consistency.
>   - **Testing**:
>     - Updates tests in `datasets-api.servertest.ts` and `datasets-schema-validation.servertest.ts` to align with new dataset item retrieval logic.
>   - **Misc**:
>     - Removes `types.ts` from `DatasetService` and integrates types into `dataset-items.ts`.
>     - Adjusts `getDatabaseReadStreamPaginated` in `getDatabaseReadStream.ts` to use `getDatasetItemsByLatest` for dataset item streaming.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f6f7abc98c9f27f867f997d84a46b1783fd9d15b. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/datasets/schemaValidation.ts
```diff
@@ -43,30 +47,23 @@ export async function validateAllDatasetItems(params: {
   expectedOutputSchema: Record<string, unknown> | null;
   prisma: PrismaClient;
 }): Promise<ValidationResult> {
-  const { datasetId, projectId, inputSchema, expectedOutputSchema, prisma } =
-    params;
+  const { datasetId, projectId, inputSchema, expectedOutputSchema } = params;
```
maxdeichmann, line 48: why do we need to pass prisma as a function argument? We could just import it.

worker/src/features/evaluation/evalService.ts
```diff
@@ -418,24 +419,42 @@ export const createEvalJobs = async ({
       | { id: string; observationId: string | null }
       | undefined;
     if (isDatasetConfig) {
-      const condition = tableColumnsToSqlFilterAndPrefix(
+      // Validate filter condition
+      tableColumnsToSqlFilterAndPrefix(
```
maxdeichmann: does this one validate? IMO this creates the SQL with the filters. Or do you take the underlying mechanism to check for filter options as a validation step only here?

worker/src/features/evaluation/evalService.ts
```diff
@@ -418,24 +419,42 @@ export const createEvalJobs = async ({
...
+          const allowedDatasetIds =
+            datasetIdFilter?.type === "stringOptions" &&
+            isArray(datasetIdFilter.value)
+              ? datasetIdFilter.value
+              : null;
+
+          const matchesFilter =
+            !allowedDatasetIds ||
+            allowedDatasetIds.length === 0 ||
+            allowedDatasetIds.includes(item.datasetId);
```
maxdeichmann: why do you do this logic on the application layer. Isnt it easier to have the dataset filter in sql? This would feel "simpler" at least as a reader.

worker/src/features/evaluation/evalService.ts
marliessophie: > do you take the underlying mechanism to check for filter options as a validation step only here

Yes. Exactly. Will make comment more explicit.

packages/shared/src/server/datasets/schemaValidation.ts
marliessophie: Sure, can change.

worker/src/features/database-read-stream/getDatabaseReadStream.ts
```diff
@@ -38,7 +37,7 @@ const tableNameToTimeFilterColumn: Record<BatchTableNames, string> = {
   traces: "timestamp",
   observations: "startTime",
   dataset_run_items: "createdAt",
-  dataset_items: "createdAt",
+  dataset_items: "createdAt", // TODO: flip to validFrom once we write in new format
```
marliessophie, line 40: No regression caused here. Created_at and valid_from both default to the same timestamp. Can simply switch in the same PR when we flip to write in new format.

worker/src/__tests__/batchExport.test.ts
```diff
@@ -1176,10 +1179,10 @@ describe("batch export test suite", () => {
       cutoffCreatedAt: new Date(Date.now() + 1000 * 60 * 60 * 24),
       filter: [
         {
-          type: "string",
```
marliessophie, line 1179: Mistake in the test file previously.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,974 @@ export async function createManyDatasetItems(props: {
...
+    data: preparedItems,
   });
 
   return {
     success: true,
     datasetItems: preparedItems,
   };
 }
+
+// dedupe
```
maxdeichmann: What is this?

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,974 @@ export async function createManyDatasetItems(props: {
...
+/**
+ * Raw database row for getting latest dataset items from dataset_items table
+ */
+type QueryGetLatestDatasetItemRow = {
+  id: string;
+  project_id: string;
+  dataset_id: string;
+  input?: any;
+  expected_output?: any;
+  metadata?: any;
```
maxdeichmann: Why do we have any here? We have the arbitrary json type for all other entities as well

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,974 @@ export async function createManyDatasetItems(props: {
...
+  sourceTraceId?: string;
+  sourceObservationId?: string;
+  createdAt: Date;
+}[];
+
+/**
+ * Type for bulk dataset item validation errors
+ * Used when validating multiple items before creation (e.g., CSV upload)
+ */
+export type CreateManyValidationError = {
```
maxdeichmann, line 622: Shouldnt this be an exception?

maxdeichmann: Same for PayloadError above.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,974 @@ export async function createManyDatasetItems(props: {
...
     datasetItems: preparedItems,
   };
 }
+
+// dedupe
+
+type IdOrName = { datasetId: string } | { datasetName: string };
+
+export type PayloadError = {
+  success: false;
```
maxdeichmann, line 588: When is an error a success?

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Deleted, sorry.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Will adjust.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: No, as comment states this is used in a couple places to display to the user what validation errors against dataset schema occured eg when uploading dataset as csv.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Never, this is used as a discriminated union type. Will define as discriminated union directly, rather than two individual types. More explicit.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+        operator: "=",
+        value: options.sourceTraceId,
+      });
+    }
+  }
+
+  if (options.sourceObservationId !== undefined) {
+    if (options.sourceObservationId === null) {
+      filterState.push({
+        type: "null",
```
maxdeichmann: Different behavior based on undefined/null may lead to bugs down the line as it is easy to gloss over it when dealing with undefined | null types. Sure this is handled everywhere?

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+
+  return filterState;
+}
+
+/**
+ * Converts FilterState to Prisma where clause for STATEFUL mode.
+ * This is a temporary helper for the deprecated STATEFUL implementation.
+ * VERSIONED mode uses SQL directly and doesn't need this.
+ */
+function buildPrismaWhereFromFilterState(filterState: FilterState): any {
```
maxdeichmann, line 753: Dont we have tableColumnsToSqlFilterAndPrefix for this?

maxdeichmann: Also any as typing is dangerous here given the where conditions need to match always.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+}
+
+/**
+ * Builds SQL search filter for full-text search on dataset items.
+ * Applies ILIKE search on id, input, expectedOutput, and metadata fields.
+ * Returns Prisma.empty if no search query provided.
+ *
+ * @param tableAlias - The table alias to use (default 'di' for dataset items)
+ */
+function buildDatasetItemSearchCondition(
```
maxdeichmann, line 806: We have so many implementations of FTS on PG in our code base. Can you take one of the existing implementations?

maxdeichmann: ... maybe the other ones do not match here. I am not sure.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+  return searchConditions.length > 0
+    ? Prisma.sql` AND (${Prisma.join(searchConditions, " OR ")})`
+    : Prisma.empty;
+}
+
+/**
+ * Builds SQL query for STATEFUL dataset items with search support.
+ * Simple direct query without version logic.
+ */
+function buildStatefulDatasetItemsQuery(
```
maxdeichmann, line 845: What is stateful?

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+
+  const datasetNameField = includeDatasetName
+    ? Prisma.sql`, d.name as dataset_name`
+    : Prisma.empty;
+
+  const filterCondition = tableColumnsToSqlFilterAndPrefix(
+    filter,
+    datasetItemsFilterCols,
+    "dataset_item_events",
+  );
```
maxdeichmann, line 871: - could this also contain dataset filters? If yes, we would need to join with datasets. Not sure if this is the case.
- do we still have the "dataset_item_events" table in postgres?

maxdeichmann: IMO we should revert this PR: https://github.com/langfuse/langfuse/issues/10619

marliessophie: > IMO we should revert this PR: https://github.com/langfuse/langfuse/pull/10619

Agree, that's planned.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+        di.source_observation_id,
+        di.status,
+        di.created_at,
+        di.updated_at,
+        di.valid_from,
+        di.is_deleted
+      FROM dataset_items di
+      WHERE di.project_id = ${projectId}
+      ${filterCondition}
+      ORDER BY di.id, di.project_id, di.valid_from DESC
```
maxdeichmann: No effect on the functionality but i would first order by project and then by id

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Just the name for the current mutable non-versioned datasets implementation.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: I have a PR to clean this up. Will unify functionality in different PR.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: I can rewrite, yes. Better to do in raw SQL and reuse `tableColumnsToSqlFilterAndPrefix`

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+/**
+ * Builds the SQL query for counting latest dataset items grouped by dataset_id.
+ */
+function buildDatasetItemsLatestCountGroupedQuery(
+  projectId: string,
+  datasetIds: string[],
+): Prisma.Sql {
+  return Prisma.sql`
+    WITH latest_items AS (
+      SELECT DISTINCT ON (di.id, di.project_id)
```
maxdeichmann: Nit: you could also implement the latest items as postgres view. Thereby you would may save yourself complexity in queries.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+      WHERE di.project_id = ${projectId}
+        AND di.dataset_id = ANY(${datasetIds})
+      ORDER BY di.id, di.project_id, di.valid_from DESC
+    )
+    SELECT
+      di.dataset_id,
+      COUNT(*) as count
+    FROM latest_items di
+    WHERE di.is_deleted = false
+      AND di.status = 'ACTIVE'
```
maxdeichmann, line 1105: This here is ACTIVE vs. ARCHIVED, right?

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+        expectedOutput: row.expected_output ?? null,
+        metadata: row.metadata ?? null,
+      }
+    : base;
+
+  const withDatasetName = includeDatasetName
+    ? { ...withIO, datasetName: row.dataset_name! }
+    : withIO;
+
+  return withDatasetName as any;
```
maxdeichmann, line 1151: why do you need `as any`?

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+}
+
+/**
+ * Counts dataset items at a specific version timestamp.
+ * Uses the same logic as getDatasetItemsByVersion but returns only the count.
```
maxdeichmann: This description does not match the implementation

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+
+  const items = result.map((row) =>
+    convertLatestRowToDomain(
+      row,
+      params.includeIO,
+      params.includeDatasetName ?? false,
+    ),
+  );
+
+  return items as any;
```
maxdeichmann, line 1200: Same comment w.r.t. any as above. General question: do we really need to check for I/O/M everywhere? Do we do this today already? We may need this but this also drives complexity.

maxdeichmann: Edit: yes, we need it

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+          },
+          ...(props.datasetId ? { datasetId: props.datasetId } : {}),
+          ...(status === "ACTIVE" && { status: DatasetStatus.ACTIVE }),
+        },
+      });
+      return item ? toDomainType(item, includeIO) : null;
+    },
+    [Implementation.VERSIONED]: async () => {
+      // NEW: Get latest version from dataset_items by validFrom DESC
+      const item = await prisma.datasetItem.findFirst({
```
maxdeichmann: Why dont you re-use one of the functions `getDatasetItemsByLatestInternal` above? May reduces risk of errors.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -518,11 +539,972 @@ export async function createManyDatasetItems(props: {
...
+
+      // Get latest version first, then check if it matches status and is not deleted
+      if (!item || item.isDeleted) {
+        return null;
+      }
+
+      // Check status filter AFTER getting latest version
+      if (status === "ACTIVE" && item.status !== DatasetStatus.ACTIVE) {
+        return null;
+      }
```
maxdeichmann: Shouldnt this be dealt with on SQL level?

packages/shared/src/server/repositories/dataset-items.ts
maxdeichmann: Edit: you could maybe use `getDatasetItemsByLatest` and drop this function?

packages/shared/src/server/repositories/dataset-items.ts
maxdeichmann: Is this only used in "STATEFUL" code path? In that case we will therow this away anyways right? If this is the case no need to clean up and rather focus on deelting it :D

web/src/features/datasets/server/dataset-router.ts
```diff
@@ -635,35 +654,26 @@ export const datasetRouter = createTRPCRouter({
...
-          sourceObservationId: true,
-          createdAt: true,
-          updatedAt: true,
-        },
+    .query(async ({ input }) => {
+      const item = await getDatasetItemById({
+        projectId: input.projectId,
+        datasetItemId: input.datasetItemId,
+        datasetId: input.datasetId,
+        status: "ALL",
```
maxdeichmann: What does "ALL" mean here?

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Even better!

web/src/pages/api/public/datasets.ts
```diff
@@ -68,17 +72,6 @@ export default withMiddlewares({
...
-            where: {
-              status: "ACTIVE",
-            },
-            select: {
-              id: true,
-            },
-            orderBy: {
-              createdAt: "desc",
-            },
-          },
```
maxdeichmann, line 81: If latest_dataset_items was a view we could directly join it here instead of making the second db call. No need to change.

worker/src/features/evaluation/evalService.ts
```diff
@@ -428,15 +428,22 @@ export const createEvalJobs = async ({
...
           FROM dataset_items as di
           WHERE project_id = ${event.projectId}
             AND id = ${event.datasetItemId}
             ${condition}
+          ORDER BY valid_from DESC
+          LIMIT 1
         `);
-        datasetItem = datasetItems.shift();
+        const latestDatasetItem = datasetItems.shift();
+        if (latestDatasetItem && latestDatasetItem.is_deleted === false) {
```
maxdeichmann: Why not adding teh condition on is_deleted to SQL?

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Will keep as is as I will re-rewrite 
- getDatasetItemsByLatestInternal -> getDatasetItemsByVersionInternal which is then called by getDatasetItemsByLatest with version=now() and for any other version.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Yes, and we do this for heavy IO for traces/observations too.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: > could this also contain dataset filters? If yes, we would need to join with datasets. Not sure if this is the case.

No, I checked.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Correct. I matched previous behaviour for dataset item count; only accounted for active items.

packages/shared/src/server/repositories/dataset-items.ts
maxdeichmann: Additional bike thought: benefit of pg view depends heavily on how often you need the latest DI throughout our backend logic. If we will almost always have a time filter, then it may does not make sense. Or at least needs more research on how to implement

marliessophie: Interesting idea. Logged as timeboxed potential improvement.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: TypeScript cannot prove that the runtime conditional logic (includeIO ? ... : ...) produces a value that matches the type-level conditional return type (IncludeIO extends true ? ... : ...), so the type assertion tells TypeScript "trust me, the runtime value matches the declared return type."

web/src/features/datasets/server/dataset-router.ts
marliessophie: Changes such that passing now status will return all items regardless of status.

packages/shared/src/server/repositories/dataset-items.ts
marliessophie: Changed function signature to explicitly accept a `sourceObservationIdIsNull` parameter. Much easier to understand.

PR #10923, opened 2025-12-04
fix(prompts): delete prompt API hanldes versions correctly

nimarb:
> [!IMPORTANT]
> Fix delete prompt API to correctly handle version and label dependencies, updating documentation and adding tests for verification.
> 
>   - **Behavior**:
>     - Update `delete` operation in `prompts.yml` and `openapi.yml` to clarify deletion behavior when neither version nor label is specified.
>     - Change response code from `200` to `204` for successful deletions in `openapi.yml`.
>   - **Functionality**:
>     - Modify `deletePrompt` in `deletePrompt.ts` to handle dependencies correctly, blocking deletion if other prompts depend on the version or label being deleted.
>     - Reattach "latest" label to the highest remaining version if the current "latest" version is deleted.
>   - **Tests**:
>     - Add tests in `prompts.v2.servertest.ts` to verify correct handling of mixed version and label dependencies, and reattaching "latest" label.
>     - Ensure tests cover scenarios where deletion should be blocked due to dependencies.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for a4f5dac01b70fd82d692cc0ec5d137d4cc95817c. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/prompts/server/actions/deletePrompt.ts
+++ b/web/src/features/prompts/server/actions/deletePrompt.ts
@@ -7,39 +7,21 @@ export type DeletePromptParams = {
   projectId: string;
   version?: number | null;
   label?: string;
-  prompts?: Prompt[]; // Optional: if already fetched, avoid duplicate query
+  promptVersions: Prompt[];
 };
 
 export const deletePrompt = async (params: DeletePromptParams) => {
-  const {
-    promptName,
-    projectId,
-    version,
-    label,
-    prompts: providedPrompts,
-  } = params;
+  const { promptName, projectId, version, label, promptVersions } = params;
 
   if (version && label) {
     throw new InvalidRequestError("Cannot specify both version and label");
   }
 
-  let prompts = providedPrompts;
-  if (!prompts) {
-    const where = {
-      projectId,
-      name: promptName,
-      ...(version ? { version } : {}),
-      ...(label ? { labels: { has: label } } : {}),
-    };
-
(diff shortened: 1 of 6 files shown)


nimarb:
follow-up to #7704 , cc @hassiebp

web/src/features/prompts/server/actions/deletePrompt.ts
hassiebp: why do we need the inner join here?

web/src/features/prompts/server/handlers/promptNameHandler.ts
```diff
@@ -92,13 +92,13 @@ const deletePromptNameHandler = async (
     });
   }
 
-  // Delete prompts (pass fetched prompts to avoid duplicate query)
+  // Delete prompt versions
   await deletePrompt({
```
hassiebp, line 96: We are still not handling the case if the prompt that is being deleted has the latest label, we will have to reattach it to the latest prompt after the deletion

web/src/features/prompts/server/actions/deletePrompt.ts
nimarb: removed it, was mostly to show better error messages but not that important i agree

PR #10944, opened 2025-12-05
chore(dataset-versioning): add unique idx on [id, project_id, valid_from]

marliessophie:
> [!IMPORTANT]
> Adds a unique index on `dataset_items` for `[id, project_id, valid_from]` to ensure data integrity, with changes in migration and Prisma schema.
> 
>   - **Database Schema**:
>     - Adds a unique index on `dataset_items` table for columns `[id, project_id, valid_from]` in `migration.sql`.
>     - Updates `schema.prisma` to include `@@unique([id, projectId, validFrom])` in `DatasetItem` model.
>   - **Warnings**:
>     - Migration will fail if duplicates exist in the specified columns.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for c67046579a04f566bb2e3867dc2f725711753054. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/generated/types.ts
```diff
@@ -430,7 +430,7 @@ export type DatasetItem = {
   dataset_id: string;
   created_at: Generated<Timestamp>;
   updated_at: Generated<Timestamp>;
-  sys_id: Generated<string | null>;
+  sys_id: string | null;
```
maxdeichmann: didnt we remove this?

packages/shared/prisma/migrations/20251209133946_dataset_items_create_idx_id_project_id_valid_from/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[id,project_id,valid_from]` on the table `dataset_items` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- CreateIndex
+CREATE UNIQUE INDEX CONCURRENTLY "dataset_items_id_project_id_valid_from_key" ON "dataset_items"("id", "project_id", "valid_from");
```
maxdeichmann: this will be heavy on the database. Please keep high timeout setting for the migration

packages/shared/prisma/schema.prisma
```diff
@@ -628,11 +628,12 @@ model DatasetItem {
   createdAt           DateTime          @default(now()) @map("created_at")
   updatedAt           DateTime          @default(now()) @updatedAt @map("updated_at")
   // dataset version cols
-  sysId               String?           @default(dbgenerated()) @map("sys_id")
+  sysId               String?           @map("sys_id")
   validFrom           DateTime          @default(now()) @map("valid_from")
   isDeleted           Boolean           @default(false) @map("is_deleted")
 
   @@id([id, projectId])
+  @@unique([id, projectId, validFrom])
```
maxdeichmann, line 637: what is the difference between primary index and unique index in postgres?

packages/shared/prisma/generated/types.ts
marliessophie: Not yet. Will do in clean up migration. Dropping default first as default computation is non-trivial.

packages/shared/prisma/schema.prisma
marliessophie: TLDR no immediate functional difference. 
Differences: 
- Quantity: You strictly get one Primary Key per table (to identify the row), whereas you can have multiple Unique Indexes (to enforce business rules like unique emails).
- Nullability: Primary Keys strictly forbid NULL values; Unique Indexes allow NULL values (and because Postgres treats NULLs as distinct, you can have multiple rows with NULL in a unique column).

To my knowledge, the database will still function if we drop the pk and keep only this unique constraint, as stated this should be functionally equivalent. However, it might be confusing for devs and possibly cause downstream problems with ORMs or replication.

packages/shared/prisma/generated/types.ts
maxdeichmann: Can you explain this to me? The default only applies to newly inserted data.

packages/shared/prisma/schema.prisma
maxdeichmann: Ok then we should change the primary key. In the current implementation the unique constraint will not allow multiple inserts per id as the primary key is still set.

packages/shared/prisma/migrations/20251209133946_dataset_items_create_idx_id_project_id_valid_from/migration.sql
maxdeichmann: Is it possible to do an "if not exists"? This may be helpful in case the migration fails but the index was still applied on the database

marliessophie: Good shout. Adding.

packages/shared/prisma/schema.prisma
marliessophie: See message in slack re release plan. Should answer this.

PR #10982, opened 2025-12-08
chore(dataset-versioning): switch pk 

marliessophie:
> [!IMPORTANT]
> Switches primary key of `dataset_items` table to include `validFrom` in both migration and Prisma schema.
> 
>   - **Database Migration**:
>     - Changes primary key of `dataset_items` table in `migration.sql` to use `id`, `projectId`, and `validFrom`.
>   - **Prisma Schema**:
>     - Updates `DatasetItem` model in `schema.prisma` to set `@@id([id, projectId, validFrom])` as the primary key.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 7eefff0fd8bcac7f71618b38773ac679ef41c1a8. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/migrations/20251209204006_dataset_items_switch_pk/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - The primary key for the `dataset_items` table will be changed. If it partially fails, the table could be left without primary key constraint.
+
+*/
+-- AlterTable
+ALTER TABLE "dataset_items" DROP CONSTRAINT "dataset_items_pkey",
+ADD CONSTRAINT "dataset_items_pkey" PRIMARY KEY ("id", "project_id", "valid_from");
```
maxdeichmann: Cant we just drop this one? This will be expensive again.

maxdeichmann: We already have this as a normal constraint.

marliessophie: Same conversation no? 

https://github.com/langfuse/langfuse/pull/10944#discussion_r2602736299

maxdeichmann: Yes, agree. We should have this as a primary key. Otherwise, prisma will likely not work good enough for us.

PR #11031, opened 2025-12-10
feat(tracing): filter observations by tool calls

nimarb:
> [!IMPORTANT]
> Enhances tracing by extracting and processing tool definitions and calls from observations, updating data models and ingestion logic.
> 
>   - **Behavior**:
>     - Adds extraction of tool definitions and calls from observation input/output in `extractToolsFromObservation()`.
>     - Updates `IngestionService` to process and store tool data in `processObservationEventList()`.
>     - Enhances `getObservationStream()` to include tool data in the output.
>   - **Data Models**:
>     - Adds `toolDefinitions`, `toolCalls`, and `toolCallNames` fields to observation records.
>   - **Tests**:
>     - Adds comprehensive tests for tool extraction logic in `extractToolsBackend.test.ts`.
>   - **Misc**:
>     - Updates ESLint rules in `.eslintrc.js` to allow unused vars prefixed with `_`.
>     - Adds new path mappings in `package.json` for `chatml` utilities.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 53143e7000ca38380b74bee0b050af5dc2778e3a. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/clickhouse/migrations/unclustered/0033_tool_columns.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE observations ADD COLUMN tool_definitions Map(String, String) DEFAULT map();
```
sumerman: Migrations themselves look alright.

A few considerations:
1. What happened to the arrays idea?
2. I would very strongly suggest splitting migration commits from the rest of this PR.
3. You would have to add these columns to the `events` table as well and coordinate with @Steffen911 on double-writes.

nimarb: thanks @sumerman !
1. chatted with hassieb about it, and came up with this schema for which we only need 2 columns but still should be fast. we only loose the ordering of the tool calls in case we don't use the index in the datamodel. but that is fine.
2. will do, thanks !
3. 👍 , would do that after the merge, or rather do it before merging ?

sumerman: 3. better check with @Steffen911 my understanding is that we can go live with it any moment now and it would be a lot simpler if we didn't have to backfill the new columns.

PR #11060, opened 2025-12-11
feat: introducing a single-level SELECt optimization in queryBuilder.

sumerman:
> [!IMPORTANT]
> Introduces single-level SELECT optimization in `queryBuilder` to enhance query performance, with shadow testing for result verification and new environment configuration for control.
> 
>   - **Behavior**:
>     - Introduces single-level SELECT optimization in `queryBuilder.ts` to improve query performance by simplifying SQL queries when applicable.
>     - Adds shadow testing in `queryExecutor.ts` to compare results of optimized and non-optimized queries.
>     - New environment variable `LANGFUSE_ENABLE_QUERY_OPTIMIZATION_SHADOW_TEST` in `env.mjs` to control shadow testing.
>   - **Query Building**:
>     - Adds `canUseSingleLevelQuery()` and `buildSingleLevelSelect()` in `queryBuilder.ts` to determine and construct single-level queries.
>     - Modifies `build()` in `queryBuilder.ts` to choose between single-level and two-level query construction based on optimization applicability.
>   - **Testing**:
>     - Adds tests in `queryBuilder.servertest.ts` to verify single-level optimization produces correct results and handles various cases, including complex aggregations and relation tables.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for b1358f5a988f425d498f504398134cbae0a1f8c0. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/query/server/queryBuilder.ts
+++ b/web/src/features/query/server/queryBuilder.ts
@@ -32,6 +32,7 @@ type AppliedMetricType = {
   aggregation: z.infer<typeof metricAggregations>;
   alias?: string;
   relationTable?: string;
+  aggs?: Record<string, string>;
 };
 
 export class QueryBuilder {
@@ -118,6 +119,7 @@ export class QueryBuilder {
       return {
         ...view.measures[metric.measure],
         aggregation: metric.aggregation,
+        aggs: view.measures[metric.measure].aggs,
       };
     });
   }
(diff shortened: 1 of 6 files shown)


web/src/features/query/server/queryBuilder.ts
```diff
@@ -374,6 +376,39 @@ export class QueryBuilder {
     return relationTables;
   }
 
+  private canUseSingleLevelQuery(
```
Steffen911, line 379: Is it necessary to check whether there's a relationTable definition? I think no, but double checking.

web/src/features/query/server/queryBuilder.ts
```diff
@@ -374,6 +376,39 @@ export class QueryBuilder {
...
+  }
+
+  private substituteAggTemplates(
+    sql: string,
+    aggs: Record<string, string>,
+  ): string {
+    let result = sql;
+    // Replace each ${aggN} placeholder with its corresponding aggregation function
+    for (const [placeholder, aggFunc] of Object.entries(aggs)) {
+      const pattern = new RegExp(`\\$\\{${placeholder}\\}`, "g");
```
Steffen911: You can also use some pattern that's easier to regex 😄

web/src/features/query/server/queryBuilder.ts
sumerman: I thought about it and it shouldn't play a role it itself. What matters is 1-1 or 1-N relations and that should be reflected in the type of aggregations we perform.

sumerman: E.g. as far as my understanding goes scores view always has 1 score to 1 trace / observation. It may get weird with events where a single score row may be pulling multiple events if we chose to join by `trace_id` only.

PR #11408, opened 2026-01-06
fix: a workaround for socket hangup issues in our delete pipeline.

sumerman:
> [!IMPORTANT]
> Introduces mutation monitoring for ClickHouse DELETE operations to address socket hangup issues, with updates to environment variables and delete functions, and adds a new integration test.
> 
>   - **Behavior**:
>     - Introduces mutation monitoring for ClickHouse DELETE operations using `executeWithMutationMonitoring()` in `mutationWaiter.ts`.
>     - Adds environment variables in `env.ts` for async deletion tracking and mutation monitoring.
>     - Updates delete functions in `events.ts`, `observations.ts`, `scores.ts`, and `traces.ts` to use mutation monitoring if enabled.
>   - **Testing**:
>     - Adds `mutationWaiter.test.ts` to test DELETE operation with mutation monitoring.
>   - **Misc**:
>     - Removes redundant query status polling functions from `backfillEventsHistoric.ts` and `backfillEventsHistoricFromParts.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 804d19dd6d77486a1531a78718d1926fc8041f4e. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -170,10 +170,14 @@ const EnvSchema = z.object({
   LANGFUSE_CLICKHOUSE_QUERY_MAX_ATTEMPTS: z.coerce.number().default(3), // Maximum attempts for socket hang up errors
   // Mutation monitoring for ClickHouse DELETE operations
   LANGFUSE_MUTATION_MONITOR_ENABLED: z.enum(["true", "false"]).default("false"),
-  LANGFUSE_MUTATION_POLL_INTERVAL_MS: z.coerce
+  // Async deletion tracking - uses AbortController + query_id polling for DELETE operations
+  LANGFUSE_ASYNC_DELETE_TRACKING_ENABLED: z
+    .enum(["true", "false"])
+    .default("false"),
+  LANGFUSE_ASYNC_DELETE_TRACKING_POLL_INTERVAL_MS: z.coerce
     .number()
     .positive()
-    .default(600_000), // 10 minutes
+    .default(60_000), // 1 minute
   LANGFUSE_SKIP_S3_LIST_FOR_OBSERVATIONS_PROJECT_IDS: z.string().optional(),
   LANGFUSE_INGESTION_PROCESSING_SAMPLED_PROJECTS: z
     .string()
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -168,6 +168,12 @@ const EnvSchema = z.object({
     .default(80e6), // 80MB
   LANGFUSE_CLICKHOUSE_DELETION_TIMEOUT_MS: z.coerce.number().default(600_000), // 10 minutes
   LANGFUSE_CLICKHOUSE_QUERY_MAX_ATTEMPTS: z.coerce.number().default(3), // Maximum attempts for socket hang up errors
+  // Mutation monitoring for ClickHouse DELETE operations
+  LANGFUSE_MUTATION_MONITOR_ENABLED: z.enum(["true", "false"]).default("false"),
+  LANGFUSE_MUTATION_POLL_INTERVAL_MS: z.coerce
+    .number()
+    .positive()
+    .default(600_000), // 10 minutes
   LANGFUSE_SKIP_S3_LIST_FOR_OBSERVATIONS_PROJECT_IDS: z.string().optional(),
   LANGFUSE_INGESTION_PROCESSING_SAMPLED_PROJECTS: z
     .string()
--- a/packages/shared/src/server/clickhouse/mutationWaiter.ts
+++ b/packages/shared/src/server/clickhouse/mutationWaiter.ts
@@ -1,11 +1,11 @@
 import { randomUUID } from "crypto";
-import { env } from "../../env";
 import { logger } from "../logger";
-import { clickhouseClient, convertDateToClickhouseDateTime } from "./client";
-import { NodeClickHouseClientConfigOptions } from "@clickhouse/client/dist/config";
 import { ClickHouseSettings } from "@clickhouse/client";
 import { instrumentAsync } from "../instrumentation";
 import { SpanKind } from "@opentelemetry/api";
+import { pollQueryStatus, getQueryError, sleep } from "./queryTracking";
+import { commandClickhouse } from "../repositories/clickhouse";
+import { env } from "../../env";
 
 /**
  * Options for executing a DELETE command with mutation monitoring.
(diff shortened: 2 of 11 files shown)


packages/shared/src/server/clickhouse/mutationWaiter.ts
```diff
@@ -0,0 +1,205 @@
...
+    AND command LIKE {uuidPattern: String}
+    AND create_time >= toDateTime({startTime: DateTime64(3)})
+  ORDER BY create_time DESC
+  LIMIT 1
+`;
+
+const MUTATION_NOT_VISIBLE_RETRY_LIMIT = 5;
+const MUTATION_NOT_VISIBLE_RETRY_DELAY_MS = 1000;
+
+function sleep(ms: number): Promise<void> {
```
maxdeichmann: we have a librrary in our code base called back-off. This may be helpful here as well.

sumerman: Thanks. IDK if we truly would want to back off here. I would rather poll at regular intervals.

packages/shared/src/env.ts
```diff
@@ -163,6 +163,16 @@ const EnvSchema = z.object({
     .default(80e6), // 80MB
   LANGFUSE_CLICKHOUSE_DELETION_TIMEOUT_MS: z.coerce.number().default(600_000), // 10 minutes
   LANGFUSE_CLICKHOUSE_QUERY_MAX_ATTEMPTS: z.coerce.number().default(3), // Maximum attempts for socket hang up errors
+  // Mutation monitoring for ClickHouse DELETE operations
+  LANGFUSE_MUTATION_MONITOR_ENABLED: z.enum(["true", "false"]).default("false"),
+  LANGFUSE_MUTATION_POLL_INTERVAL_MS: z.coerce
+    .number()
+    .positive()
+    .default(600_000), // 10 minutes
```
Steffen911: I would start with 1min. Polling is cheap enough and that should get us faster turnaround times.

packages/shared/src/server/repositories/traces.ts
```diff
@@ -954,14 +955,26 @@ export const deleteTracesByProjectId = async (projectId: string) => {
...
-      await commandClickhouse({
-        query: query,
-        params: input.params,
-        clickhouseConfigs: {
-          request_timeout: env.LANGFUSE_CLICKHOUSE_DELETION_TIMEOUT_MS,
-        },
-        tags: input.tags,
-      });
+
+      if (env.LANGFUSE_ASYNC_DELETE_TRACKING_ENABLED === "true") {
```
Steffen911, line 959: Do you deliberately exclude the other delete statements in this file? E.g. the timestamp based and the project based deletes above?

packages/shared/src/server/clickhouse/queryTracking.ts
```diff
@@ -0,0 +1,127 @@
...
+    clickhouseConfigs: {
+      request_timeout: 60_000,
+    },
+    clickhouseSettings: {
+      skip_unavailable_shards: 1,
+    },
+    tags: {
+      feature: "query-tracking",
+      operation: "pollQueryStatus-processes",
+    },
```
Steffen911: Propagating custom tags here would be nice.

packages/shared/src/server/clickhouse/mutationWaiter.ts
```diff
@@ -0,0 +1,175 @@
...
+        tableName: opts.tableName,
+        timeoutMs: opts.timeoutMs,
+        pollIntervalMs: opts.pollIntervalMs,
+      });
+
+      // Track command errors (non-abort)
+      let commandError: Error | undefined;
+
+      // Fire the DELETE command (blocks until mutation completes, but we'll abort the HTTP connection)
+      const commandPromise = clickhouseClient(opts.clickhouseConfigs).command({
```
Steffen911: I think using our utility here could be useful, e.g.
```
    // Fire the query with abort signal. The query will continue on the server
    // even after we abort the HTTP connection.
    const queryPromise = commandClickhouse({
      query,
      tags: {
        feature: "background-migration",
        operation: "fireQuery",
        queryId,
      },
      // clickhouseConfigs: {
      //   request_timeout: timeoutMs,
      // },
      clickhouseSettings: {
        // send_progress_in_http_headers: 1,
        // http_headers_progress_interval_ms: "30000",
        ...retrySettings,
      },
      abortSignal: abortController.signal,
    });
```
If you don't use it, make sure to explicitly overwrite the request_timeout, send_progress_in_http_headers, and http_headers_progress_interval_ms seconds. Otherwise, they may interrupt the execution if there is no client listening anymore.

packages/shared/src/server/clickhouse/mutationWaiter.ts
```diff
@@ -0,0 +1,175 @@
...
+        abort_signal: abortController.signal,
+      });
+
+      // Track errors - suppress expected abort errors, capture others
+      commandPromise.catch((err) => {
+        const isAbortError =
+          err?.name === "AbortError" || err?.message?.includes("aborted");
+        if (!isAbortError) {
+          commandError = err;
+          logger.debug(`Mutation ${queryId} promise rejected: ${err?.message}`);
```
Steffen911:
```suggestion
          logger.info(`Mutation ${queryId} promise rejected: ${err?.message}`);
```

packages/shared/src/server/clickhouse/mutationWaiter.ts
```diff
@@ -0,0 +1,175 @@
...
+async function waitForQueryCompletion(
+  queryId: string,
+  timeoutMs: number,
+  pollIntervalMs: number,
+  abortController?: AbortController,
+): Promise<void> {
+  const startTime = Date.now();
+  let notFoundCount = 0;
+
+  while (Date.now() - startTime < timeoutMs) {
```
Steffen911, line 40: Interesting approach. Intuitively I would've used something like setInterval and then have an abort or exit clause somewhere. Or, alternatively, have a singleton that tracks all queries and polls them jointly would be nice.

I don't see anything odd with this approach though, so should be good.

packages/shared/src/server/repositories/traces.ts
sumerman: Yes, for now should only cover project deletions. If it works I will apply for other lengthy deletion types.

packages/shared/src/server/clickhouse/mutationWaiter.ts
sumerman: Sure, `commandClickhouse` should be just fine here. But the custom logic around that seems to actually live in `clickhouseClient`, so it would have been triggered anyway.

packages/shared/src/server/clickhouse/mutationWaiter.ts
sumerman: I was pondering a singleton, but ultimately went for the dumbest approach.

PR #11547, opened 2026-01-14
feat(evals): add single observation evals

hassiebp:
> [!IMPORTANT]
> Adds single observation evaluation feature with schema, processing logic, and integration into OTEL ingestion and S3 storage.
> 
>   - **Behavior**:
>     - Adds `observationForEvalSchema` in `observationForEval.ts` for single observation evaluations.
>     - Introduces `convertEventRecordToObservationForEval()` to convert event records to observation evals.
>     - Implements `scheduleObservationEvals()` and `processObservationEval()` for scheduling and processing observation evals.
>     - Integrates observation evals with OTEL ingestion in `otelIngestionQueue.ts`.
>   - **Infrastructure**:
>     - Adds `LLMAsJudgeExecutionQueue` for handling observation eval jobs.
>     - Implements `createObservationEvalSchedulerDeps()` and `createObservationEvalProcessorDeps()` for dependency management.
>     - Updates `IngestionService` to support event record creation and writing.
>   - **Testing**:
>     - Extensive tests for observation evals in `observationEval` directory.
>     - Tests for OTEL to observation eval conversion in `otelToObservationForEval.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 39b15fdf0af7f2b24d2b0582a621ebe36225d49f. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/prisma/migrations/20260113120000_add_filter_target_to_job_configurations/migration.sql
```diff
@@ -0,0 +1,23 @@
...
+SET "filter_target" = CASE
+  WHEN "target_object" = 'dataset' THEN 'DATASET'::"JobConfigFilterTarget"
+  ELSE 'TRACE'::"JobConfigFilterTarget"
+END;
+
+-- Make column non-nullable after migration
+ALTER TABLE "job_configurations" ALTER COLUMN "filter_target" SET NOT NULL;
+
+-- Add index for efficient filtering by project, filter_target, and status
+CREATE INDEX IF NOT EXISTS "job_configurations_project_id_filter_target_status_idx"
```
hassiebp: Table is small enough for this to run without concurrently in a few milliseconds

packages/shared/src/features/evals/types.ts
```diff
@@ -157,3 +157,37 @@ export const JobTimeScopeZod = z.enum(["NEW", "EXISTING"]);
...
+    id: "observation",
+    display: "Observation",
+    availableColumns: [
+      {
+        name: "Metadata",
+        id: "metadata",
+        type: "stringObject",
+      },
+      { name: "Input", id: "input" },
+      { name: "Output", id: "output" },
```
hassiebp: todo: expand with experiment fields

packages/shared/src/tableDefinitions/tracesTable.ts
```diff
@@ -211,6 +211,91 @@ export const datasetOnlyCols: ColumnDefinition[] = [datasetCol];
 
 export const evalTraceTableCols: ColumnDefinition[] = tracesOnlyCols;
 export const evalDatasetFormFilterCols: ColumnDefinition[] = datasetOnlyCols;
+
+// Columns for observation-based eval filtering.
+// These are evaluated in-memory against the processed observation record.
+// Includes both observation-level fields and trace-level fields extracted from OTEL attributes.
+export const evalObservationFilterCols: ColumnDefinition[] = [
```
hassiebp: todo: expand to allow filtering for experiments

hassiebp:
@ellipsis-dev update PR description

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
```diff
@@ -0,0 +1,89 @@
...
+          jobInputObservationId,
+        },
+        select: { id: true },
+      });
+
+      return existing;
+    },
+
+    uploadObservationToS3: async (params) => {
+      const path = `${env.LANGFUSE_S3_EVENT_UPLOAD_PREFIX}evals/${params.projectId}/observations/${params.observationId}.json`;
```
hassiebp, line 51: @Steffen911 Please let me know whether this is a good path for the observation eval JSONs

packages/shared/src/features/evals/observationForEval.ts
```diff
@@ -0,0 +1,164 @@
+import { z } from "zod/v4";
+import { DEFAULT_TRACE_ENVIRONMENT } from "../../server/ingestion/types";
+import { type EventRecordBaseType } from "../../server/repositories/definitions";
+import { ObservationLevel } from "../../domain";
+
+const flexibleUsageCostSchema = z.record(z.string(), z.number().nullable());
+
+export const observationForEvalSchema = z.object({
```
maxdeichmann, line 8: Why dont you derive this from the observations domain?

packages/shared/src/server/otel/OtelIngestionProcessor.ts
```diff
@@ -287,6 +290,24 @@ export class OtelIngestionProcessor {
                   );
                 }
 
+                let toolDefinitions = undefined;
```
maxdeichmann, line 293: why do you need to change anything about tool definitions in your PR?

packages/shared/src/server/evalJobConfigCache.ts
```diff
@@ -10,73 +10,98 @@ import { redis } from "./";
...
  */
 
-const NO_JOB_CONFIG_PREFIX = "langfuse:eval:no-job-configs";
+/** Cache types for different eval job configuration targets */
+type EvalConfigCacheType = "trace" | "event";
+
+const CACHE_PREFIXES: Record<EvalConfigCacheType, string> = {
+  trace: "langfuse:eval:no-job-configs",
+  event: "langfuse:eval:no-event-eval-configs",
+};
```
maxdeichmann, line 19: What about dataset level configs here?

packages/shared/src/server/evalJobConfigCache.ts
```diff
@@ -10,73 +10,98 @@ import { redis } from "./";
...
+export const clearNoJobConfigsCache = (projectId: string) =>
+  clearNoEvalConfigsCache(projectId, "trace");
+
+// Event-targeted eval config cache (targetObject: "event")
+export const hasNoObservationEvalConfigsCache = (projectId: string) =>
+  hasNoEvalConfigsCache(projectId, "event");
+export const setNoObservationEvalConfigsCache = (projectId: string) =>
+  setNoEvalConfigsCache(projectId, "event");
+export const clearNoObservationEvalConfigsCache = (projectId: string) =>
+  clearNoEvalConfigsCache(projectId, "event");
```
maxdeichmann: Functions with one line function calls are a bit much overhead for anyone who wants to use the functions at a later point in time

packages/shared/src/features/evals/types.ts
```diff
@@ -1,5 +1,17 @@
...
+  TRACE: "trace",
+  DATASET: "dataset",
+  EVENT: "event",
+  EXPERIMENT: "experiment",
+} as const;
+
+export type EvalTargetObject =
+  (typeof EvalTargetObject)[keyof typeof EvalTargetObject];
+
+export const EvalTargetObjectSchema = z.enum(["trace", "dataset", "event"]);
```
Steffen911: Why is the ObjectSchema a subset of the target objects? I.e. where did experiment go?

Or is the idea that we have the two "legacy" modes and a single new mode?

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -236,43 +244,123 @@ export const otelIngestionQueueProcessor: Processor = async (
...
     // the highest possible throughput. Therefore, we start with a Promise.all.
     // If necessary, we may use a for each instead.
-    await Promise.all(
-      [
-        // Process traces
-        processEventBatch(traces, auth, {
-          delay: 0,
-          source: "otel",
-          forwardToEventsTable: shouldForwardToEventsTable,
-        }),
```
Steffen911, line 246: Why did you move this one into a separate statement? This just adds a bit of nesting, right?

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -236,43 +244,123 @@ export const otelIngestionQueueProcessor: Processor = async (
...
+      return;
+    }
+
+    // Create scheduler deps only if we have eval configs
+    const evalSchedulerDeps = hasEvalConfigs
+      ? createObservationEvalSchedulerDeps()
+      : null;
+
+    // Process each event independently
+    for (const eventInput of eventInputs) {
```
Steffen911: Why don't we do this in a Promise.all? This could take a long time as each call is blocking with the await, right?

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -236,43 +244,123 @@ export const otelIngestionQueueProcessor: Processor = async (
...
-      useDirectEventWrite
-    ) {
+      useDirectEventWrite;
+
+    const evalConfigs = await fetchObservationEvalConfigs(projectId).catch(
+      (error) => {
+        traceException(error);
+        logger.warn(
+          `Failed to fetch observation eval configs for project ${projectId}`,
+          error,
```
maxdeichmann, line 295: Ok for me to not fail ingestion pipeline for this error. Yet, we need observability on this.

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "@langfuse/shared/src/server";
+import { env } from "../../../env";
+import { getEvalS3StorageClient } from "../s3StorageClient";
+import { type ObservationEvalSchedulerDeps } from "./types";
+
+/**
+ * Creates production dependencies for the observation eval scheduler.
+ * Wires up real implementations for Prisma, S3, and BullMQ.
+ */
+export function createObservationEvalSchedulerDeps(): ObservationEvalSchedulerDeps {
```
maxdeichmann, line 15: Is this a pattern we want to have? Looks like functional dependency injection.

worker/src/services/IngestionService/index.ts
```diff
@@ -450,10 +457,34 @@ export class IngestionService {
...
+   *
+   * @param eventData - The event data to write
+   * @param fileKey - The file key where the raw event data is stored
+   */
+  public async writeEvent(
+    eventData: EventInput,
+    fileKey: string,
+  ): Promise<void> {
+    const eventRecord = await this.createEventRecord(eventData, fileKey);
+    this.writeEventRecord(eventRecord);
```
Steffen911: This doesn't seem to have any invocations?

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
maxdeichmann: We need to ensure the following:
- Data needs to be deleted when projects are deleted
- Data retention also needs to pick this up

IMO we could keep it on this filepath and have a very strict ttl on S3 on this path (e.g. 2 hours). All evals which did not succeed by that time would fail. This is also problematic for evals which failed for whatever reason and we retry them manually. Hence having ~7 days woudl be favorable. In this case, we should then also have these objects be part of the data retention / project deletion processing. For this, the path should probably be /proj_id/evals ...

worker/src/features/evaluation/observationEval/fetchObservationEvalConfigs.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Fetches active observation eval configs for a project.
+ *
+ * Uses a cache to avoid unnecessary database queries:
+ * - If cached as "no configs", returns empty array immediately
+ * - If cache miss, queries database and caches result if empty
+ *
+ * @param projectId - The project ID to fetch configs for
+ * @returns Array of observation eval configs (empty if none exist)
+ */
+export async function fetchObservationEvalConfigs(
```
Steffen911, line 20: IMO at this point, we should also consider caching the configs themselves and invalidate them eventually. As they're running more in the critical path of the otel-ingestion-queue, where additional time may stall our ingestion pipeline, having an awaited database call for each project with a config is most likely too expensive.

We can either do a cache with a 5min duration or a longer one and add invalidation logic. I'm pretty sure that this will be insufficient to keep otel-ingestion-queue throughput high.

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
Steffen911, line 15: Why did you decide to make this thing a function?

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
Steffen911: @hassiebp I like that this would allow for reuse between multiple configs. So +1 on the setup.

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  if (!job) {
+    logger.info(
+      `Job execution ${event.jobExecutionId} not found. It may have been deleted.`,
+    );
+
+    return;
+  }
+
+  if (job.status === JobExecutionStatus.CANCELLED) {
```
maxdeichmann: IMO we can drop the cancellation logic here. Events are immutable and there should not be a late update causing a cancellation.

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  // Process each matching config
+  for (const matchingConfig of matchingConfigs) {
+    try {
+      await processMatchingConfig({
+        observation,
+        matchingConfig,
+        observationS3Path,
+        schedulerDeps,
+      });
```
Steffen911: We could consider moving all of these into a Promise.all together to reduce waiting times.

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  if (existingJob) {
+    logger.debug("Job already exists for observation and config", {
+      configId: matchingConfig.id,
+      observationId: observation.span_id,
+      existingJobId: existingJob.id,
+    });
+
+    return;
+  }
```
Steffen911: Can we track an recordIncrement how often that happens? With single span evals, I'd hope that this is an extremely rare case and that we can drop this lookup eventually to reduce reads from this large table.
Potentially, we can optimistically write and ON CONFLICT do nothing?

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -236,43 +244,123 @@ export const otelIngestionQueueProcessor: Processor = async (
...
+        );
+
+        continue;
+      }
+
+      // Step 2: Schedule observation evals (independent of event writes)
+      if (hasEvalConfigs && evalSchedulerDeps) {
+        try {
+          const observation =
+            convertEventRecordToObservationForEval(eventRecord);
```
maxdeichmann: This here is parsing again. IMO we should check the entire ingestion pipeline again for double parsing of events

worker/src/features/evaluation/observationEval/shouldSampleObservation.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @returns true if the observation should be sampled, false otherwise.
+ */
+export function shouldSampleObservation(params: ShouldSampleParams): boolean {
+  const { samplingRate, randomValue = Math.random() } = params;
+
+  if (samplingRate >= 1) return true;
+  if (samplingRate <= 0) return false;
+
+  return randomValue < samplingRate;
+}
```
Steffen911, line 22: I would rather have our zod schema reject values that are >1 and <0 and skip this.

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+async function processMatchingConfig(
+  params: ProcessConfigParams,
+): Promise<void> {
+  const { observation, matchingConfig, observationS3Path, schedulerDeps } =
+    params;
+
+  // Check deduplication (job already exists?)
+  const existingJob = await schedulerDeps.findExistingJobExecution({
```
maxdeichmann: IMO we should drop this and go via unique jobId

maxdeichmann: If we want to move the job executions to CH at some point, we cannot read form the table like this.

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
Steffen911: Is this still a thing? I didn't find any cancellation logic in the otel-ingestion-queue part. And with immutable events that shouldn't happen anyway.

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
```diff
@@ -0,0 +1,172 @@
...
+    logger.debug(`Job ${job.id} was cancelled, deleting execution record.`);
+
+    await prisma.jobExecution.delete({
+      where: {
+        id: job.id,
+        projectId: event.projectId,
+      },
+    });
```
Steffen911: Even if it's cancelled, I'd rather keep it around. Otherwise, this will be insanely hard to debug if jobs just disappear. Do we do that in other cases? If yes, we should remove this behaviour too. Perhaps that would have kept me more sane in debugging things before Christmas.

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
```diff
@@ -0,0 +1,172 @@
...
+    return;
+  }
+
+  // Fetch job configuration
+  const config = await prisma.jobConfiguration.findFirst({
+    where: {
+      id: job.jobConfigurationId,
+      projectId: event.projectId,
+    },
+  });
```
Steffen911: Consider reusing the cache ones the actual configs reside in there.

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
```diff
@@ -0,0 +1,172 @@
...
+    );
+  }
+
+  // Fetch eval template
+  const template = await prisma.evalTemplate.findFirst({
+    where: {
+      id: config.evalTemplateId,
+      OR: [{ projectId: event.projectId }, { projectId: null }],
+    },
+  });
```
Steffen911: Can't we do that lookup using Foreign Keys from the jobConfig table? That would reduce queries. Also, I thought we wanted to move to prompts for this instead of sticking with evalTemplates.

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
hassiebp: TODO: add TTL of 3 days to this path

packages/shared/src/features/evals/observationForEval.ts
hassiebp: `evalRecordBaseSchema` is not enforced on ingestion. Some transformations are handled by Clickhouse SDK / Clickhouse itself, i.e. metadata is a `Record<string, object>` on ingestion and not as enforced by schema a `Record<string, string>` . There are some more types where on ingestion the object is actually differently typed then `eventRecordBaseSchema`. Thus I use a more loosely typed schema here to get typing right and move the more detailed parsing (if necessary) to the business logic.

packages/shared/src/server/otel/OtelIngestionProcessor.ts
hassiebp: Tool calls were missing from the direct write path. Thus added them here.

packages/shared/src/server/evalJobConfigCache.ts
hassiebp: Actually these were not exhaustively named, the `trace` prefix is for the legacy `trace` and `dataset` experience, and `event` is for configs in the `event` and `experiment` experience. will fix

worker/src/queues/otelIngestionQueue.ts
hassiebp: Does the `traceException` already handles the observability on this by setting the span status accordingly?

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
hassiebp: I added the dependency injection for the core scheduling and processing functions to make the testing easier and more comprehensive

packages/shared/src/features/evals/types.ts
hassiebp: Fixed this, thanks

packages/shared/src/server/evalJobConfigCache.ts
hassiebp: agree, fixed this

worker/src/features/evaluation/observationEval/fetchObservationEvalConfigs.ts
hassiebp: Agree - I'd suggest this is a follow-on task

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
hassiebp: removed it, thanks

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
hassiebp: Adding caching will be tackled in follow-up task

worker/src/features/evaluation/observationEval/observationEvalProcessor.ts
hassiebp: Moving to prompts is a separate, much larger feature that is not in scope here. Will optimize the query for the template though 👍

worker/src/queues/otelIngestionQueue.ts
hassiebp: parsing is with z.unknown() for expensive fields (esp. metadata) as discussed in our call

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
hassiebp: Removed this as discussed

worker/src/features/evaluation/observationEval/shouldSampleObservation.ts
hassiebp: we have this also in our zod schema, this is just as a double cover in case funciton is reused

worker/src/queues/otelIngestionQueue.ts
hassiebp: I found this slightly more comprehendible - no strong feelings and happy to revert if you push on this

worker/src/queues/otelIngestionQueue.ts
Steffen911: @hassiebp Yes, traceException should set the span status to failed without raising an error for bullmq.

packages/shared/src/server/redis/llmAsJudgeExecutionQueue.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    LLMAsJudgeExecutionQueue.instance = newRedis
+      ? new Queue<TQueueJobTypes[QueueName.LLMAsJudgeExecution]>(
+          QueueName.LLMAsJudgeExecution,
+          {
+            connection: newRedis,
+            prefix: getQueuePrefix(QueueName.LLMAsJudgeExecution),
+            defaultJobOptions: {
+              removeOnComplete: 10_000, // important for job deduplication
```
Steffen911, line 33: I think we're not actively setting a `jobId` when creating the job in `enqueueEvalJob`. I.e. there is no deduplication happening within bullMq. I would set removeOnComplete to true as this serves no purpose.

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
```diff
@@ -0,0 +1,99 @@
...
+        where: {
+          projectId,
+          jobConfigurationId,
+          jobInputObservationId,
+        },
+        select: { id: true },
+      });
+
+      return existing;
+    },
```
Steffen911: Is this still being invoked? I think we dropped it, right?

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
```diff
@@ -0,0 +1,177 @@
...
+        matchingConfig,
+        observationS3Path,
+        schedulerDeps,
+      }).catch((error) => {
+        logger.error("Failed to process observation eval config", {
+          configId: matchingConfig.id,
+          observationId: observation.span_id,
+          projectId: observation.project_id,
+          error,
+        });
```
Steffen911, line 97: This implies that there are no retries at all, right? If something fails, we log it and move on? Is this intentional?

Steffen911 (approved):

packages/shared/src/server/redis/llmAsJudgeExecutionQueue.ts
hassiebp: We are setting the `jobId` here to the deterministic `jobExectionId` which is basically a hash ob job configuration id and observation id:

https://github.com/langfuse/langfuse/blob/single-observation-evals/worker/src/features/evaluation/observationEval/createSchedulerDeps.ts#L84-L84

worker/src/features/evaluation/observationEval/createSchedulerDeps.ts
hassiebp: correct, forgot to clean up

worker/src/features/evaluation/observationEval/scheduleObservationEvals.ts
hassiebp: Correct, if the write to postgres for the job execution record fails or if the enqueue of the job fails, we do not retry. Would you add retry logic for this?

Steffen911: @hassiebp Maybe add a metric and an alert for this? That way, we don't have to actively check how regularly this happens?

packages/shared/src/server/redis/llmAsJudgeExecutionQueue.ts
Steffen911: @hassiebp This is just part of the payload. ID is a dedicated field (https://docs.bullmq.io/guide/jobs/job-ids)

PR #11743, opened 2026-01-27
fix(dataset-compare): adjust trace query to include timestamp and disable retries

marliessophie:


web/src/features/datasets/components/DatasetAggregateTableCell.tsx
```diff
@@ -77,14 +78,10 @@ const DatasetAggregateCellContent = ({
     },
     {
       enabled: value.observation !== undefined,
-      trpc: {
-        context: {
-          skipBatch: true,
-        },
```
maxdeichmann, line 83: now all trace.byid APIs are batched into a single REST call. This may result in too large bodies and server side errors.

marliessophie: We have a global default setting that disabled batching for all queries no? Making the setting redundant. Pls see: https://github.com/langfuse/langfuse/blob/main/web/src/utils/api.ts#L199

PR #11745, opened 2026-01-27
chore: debounce trace deletion

sumerman:


sumerman:
With `lightweight_update` deletes it is PostgreSQL that is becoming the bottleneck and this debouncing technique although makes sense overall has the downside of putting load squarely on PostgreSQL.

PR #11763, opened 2026-01-28
fix(widgets): fix Add filter button not responding in InlineFilterBuilder

Steffen911:
## Summary
- Fixed the "Add filter" button in widget form not responding when clicked
- The issue was caused by a race condition where the `useEffect` sync was overwriting newly added empty filter rows
- Now only calls `onChange` when valid filters have actually changed, preventing unnecessary sync triggers

## Root Cause
When clicking "Add filter":
1. A new empty filter row was added to `wipFilterState`
2. `onChange` was called with valid filters (excluding the empty row)
3. This triggered the `useEffect` that syncs `wipFilterState` with `filterState`
4. The sync immediately removed the new empty row before it could render

## Test plan
- [ ] Open the widget form (create new widget or edit existing)
- [ ] Click "Add filter" button
- [ ] Verify a new empty filter row appears
- [ ] Fill in the filter details and verify it persists

----

> [!IMPORTANT]
> Fixes 'Add filter' button in `InlineFilterBuilder` by ensuring `onChange` is only called when valid filters change, preventing race condition.
> 
>   - **Behavior**:
>     - Fixes 'Add filter' button in `InlineFilterBuilder` not responding due to race condition.
>     - `onChange` in `setWipFilterState` now only called if valid filters change, preventing `useEffect` from overwriting new empty filter rows.
>   - **Functions**:
>     - Modifies `setWipFilterState` in `InlineFilterBuilder` to compare `validFilters` and `prevValidFilters` before calling `onChange`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 962acfae556d50642fafc6167567e230d9abef64. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/filters/components/filter-builder.tsx
+++ b/web/src/features/filters/components/filter-builder.tsx
@@ -289,7 +289,16 @@ export function InlineFilterBuilder({
       const validFilters = newState.filter(
         (f) => singleFilter.safeParse(f).success,
       ) as FilterState;
-      onChange(validFilters);
+
+      // Only call onChange if valid filters actually changed to avoid
+      // triggering the useEffect sync which would discard WIP filter rows
+      const prevValidFilters = prev.filter(
+        (f) => singleFilter.safeParse(f).success,
+      ) as FilterState;
+      if (JSON.stringify(validFilters) !== JSON.stringify(prevValidFilters)) {
+        onChange(validFilters);
+      }
+
       return newState;
     });
   };


Steffen911:
Initial plan output: 

> I found the issue. Looking at the InlineFilterBuilder component in                    
>   /web/src/features/filters/components/filter-builder.tsx, there's a race condition     
>   caused by the useEffect that syncs the internal wipFilterState with the external      
>   filterState prop.                                                                     
>                                                                                         
>   Root Cause Analysis:                                                                  
>                                                                                         
>   1. When you click "Add filter", addNewFilter() is called which adds an empty filter to
>    wipFilterState                                                                       
>   2. The setWipFilterState function then calls onChange(validFilters) - but the new     
>   empty filter isn't valid (it has no column selected), so validFilters is the same as  
>   before                                                                                
>   3. However, onChange is called with a new array reference (even though the content is 
>   the same)                                                                             
>   4. This triggers the useEffect at lines 280-282:                                      
>   useEffect(() => {                                                                     
>     _setWipFilterState(filterState);                                                    
>   }, [filterState]);                                                                    
>   5. The useEffect immediately overwrites wipFilterState with the external filterState, 
>   which doesn't include the new empty filter row - effectively removing it before it can
>    be rendered                                                                          
>                                                                                         
>   The Fix:                                                                              
>                                                                                         
>   The setWipFilterState function should only call onChange when the valid filters have  
>   actually changed. Here's how to fix it:

PR #11858, opened 2026-02-03
fix(prompt-automations): filter events by trigger eventActions array

ximzzzzz:
Previously, prompt automation triggers ignored the eventActions array configured via the GUI, causing automations set to only trigger on "created" events to also fire on "updated" events.

## What does this PR do?

Fixes prompt automation triggers to properly filter events based on the eventActions array configured via the GUI.
 
**Problem**: Prompt automation triggers ignored the eventActions array and only checked the filter field. This caused automations set to trigger only on "created" events to also fire on "updated" events when a new prompt version was created (due to label changes on previous versions).

 **Solution**: Added a check in promptVersionProcessor to verify that the event action is included in the trigger's eventActions array before processing. If not included, the trigger is skipped.

Fixes #9630 

## Type of change

## Mandatory Tasks

## Checklist
- [x] I have read the https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md
 -  [x] My code follows the style guidelines of this project (pnpm run format)
-  [x] I have commented my code, particularly in hard-to-understand areas
  - [x] I have checked if my PR needs changes to the documentation
  - [x] I have checked if my changes generate no new warnings (npm run lint)
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [x] I have checked if new and existing unit tests pass locally with my changes

### DIFF
--- a/web/src/__tests__/async/prompts.v2.servertest.ts
+++ b/web/src/__tests__/async/prompts.v2.servertest.ts
@@ -69,7 +69,7 @@ const setupTriggerAndAction = async (projectId: string) => {
       id: v4(),
       projectId: projectId,
       eventSource: "prompt",
-      eventActions: ["updated"],
+      eventActions: ["created", "updated"],
       filter: [],
       status: "ACTIVE",
     },
--- a/worker/src/__tests__/promptVersionProcessor.test.ts
+++ b/worker/src/__tests__/promptVersionProcessor.test.ts
@@ -4,12 +4,10 @@ import {
   ActionExecutionStatus,
   JobConfigState,
   TriggerEventSource,
-  TriggerEventAction,
   PromptType,
 } from "@langfuse/shared";
 import {
   createOrgProjectAndApiKey,
-  redis,
   EntityChangeEventType,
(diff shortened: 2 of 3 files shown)


web/src/__tests__/async/prompts.v2.servertest.ts
```diff
@@ -69,7 +69,7 @@ const setupTriggerAndAction = async (projectId: string) => {
       id: v4(),
       projectId: projectId,
       eventSource: "prompt",
-      eventActions: ["updated"],
+      eventActions: ["created", "updated"],
```
ximzzzzz, line 72: people passed this before as eventActions is ignored. but after i fix code, code works as intended. 
if there is only "updated" eventActions, created test will fail(that's why i failed CI). so test script has to be modified.

marksalpeter:
Thanks for the contribution @ximzzzzz! We went a slightly different route when fixing this in order to maximize backwards compatibility with the older style of filter parameters, you basically had the right idea. 

I'm going to close this for now because the fix is already merged.

PR #11873, opened 2026-02-04
fix: an attempt to allocate a little less in our web containers and workers

sumerman:
> [!IMPORTANT]
> Optimized record conversion and JSON parsing to reduce memory allocation and improve performance in web containers and workers.
> 
>   - **Performance Optimization**:
>     - Introduced `convertNumericRecord` in `observations_converters.ts` to replace `Object.fromEntries` for converting records to numbers, improving performance.
>     - Replaced `Object.fromEntries` with `convertNumericRecord` for `usageDetails`, `costDetails`, and `providedCostDetails` in `convertObservationPartial()`.
>   - **JSON Parsing**:
>     - Added `UNSAFE_NUMBER_PATTERN` in `json.ts` to detect numbers that might lose precision with native `JSON.parse`.
>     - Updated `parseJsonPrioritised()` to use native `JSON.parse` if no unsafe numbers are detected, otherwise use `lossless-json` for precision.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for dcadd3c57a6d76b57ba8c80915bee4039e7a7e3b. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/utils/json.ts
+++ b/packages/shared/src/utils/json.ts
@@ -381,10 +381,25 @@ export function deepParseJsonIterative(
   return rootEntry.output;
 }
 
+/**
+ * Pattern to detect JSON numbers that might lose precision with native JSON.parse.
+ *
+ * Catches:
+ * 1. [\d.]{13,} - 13+ characters of digits/dots (conservative threshold for ~12+ significant digits)
+ * 2. \d[eE] - scientific notation (always use lossless-json for safety)
+ */
+const UNSAFE_NUMBER_PATTERN = /[\d.]{13,}|\d[eE]/;
+
 export const parseJsonPrioritised = (
   json: string,
 ): JsonNested | string | undefined => {
   try {
+    // Fast path: use native JSON.parse if no potentially unsafe numbers detected
+    if (!UNSAFE_NUMBER_PATTERN.test(json)) {
+      return JSON.parse(json) as JsonNested;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/shared/src/utils/json.ts
```diff
@@ -381,10 +381,26 @@ export function deepParseJsonIterative(
...
 
+/**
+ * Pattern to detect JSON numbers that might lose precision with native JSON.parse.
+ *
+ * Catches:
+ * 1. \d{14,} - 14+ consecutive digits (extra safe margin before MAX_SAFE_INTEGER's 16 digits)
+ * 2. \d[eE] - scientific notation (digit followed by e/E, e.g., 1e5, 2.5E10)
+ * 3. \.\d{14,} - 14+ digits after decimal point (extra safe margin)
+ */
+const UNSAFE_NUMBER_PATTERN = /\d{14,}|\d[eE]|\.\d{14,}/;
```
Steffen911: Would we expect a problem if a number has 13 digits before the decimal point and 13 after? Or is that still within the safety margins you define here?

sumerman: Good point. It is not in the margin. any string of numbers and a dot that is this long can be problematic.

PR #11916, opened 2026-02-06
chore(home-dashboards): replace tremor with recharts

coffee4tw:
## What does this PR do?

Introduce a beta toggle that switched tremor based dashboards to recharts in preparation of removing tremor as a dependency completely.

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Introduces a beta toggle to switch from Tremor to Recharts for dashboard charts, adds new chart types, and updates components to support Recharts.
> 
>   - **Behavior**:
>     - Introduces a beta toggle to switch from Tremor to Recharts for dashboard charts in `index.tsx`.
>     - Adds support for `AREA_TIME_SERIES` in `types.ts`, `schema.prisma`, and `migration.sql`.
>     - Updates `Chart` component to handle new chart types and configurations in `Chart.tsx`.
>   - **Components**:
>     - Replaces Tremor charts with Recharts in `LatencyChart.tsx`, `ModelUsageChart.tsx`, `TracesBarListChart.tsx`, `TracesTimeSeriesChart.tsx`, `UserChart.tsx`, `CategoricalScoreChart.tsx`, `NumericScoreHistogram.tsx`, `NumericScoreTimeSeriesChart.tsx`, `ScoreAnalytics.tsx`.
>     - Adds `AreaChartTimeSeries.tsx` for area chart support.
>     - Updates `HorizontalBarChart.tsx`, `LineChartTimeSeries.tsx`, `PieChart.tsx`, `VerticalBarChart.tsx`, `VerticalBarChartTimeSeries.tsx` for Recharts.
>   - **Utilities**:
>     - Adds `legacy-chart-adapters.ts` for data conversion to Recharts format.
>     - Updates `getColorsForCategories.tsx` to include more colors.
>   - **Misc**:
>     - Updates `globals.css` for chart color variables.
>     - Removes `TabTimeSeriesChart.tsx` and updates `TabsComponent.tsx` for new charting logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e10580f0dec4f78dc85073da17e83c837a096ca9. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


nimarb:
off the top of my head:

- nice!
- can we get the same colour for the charts as before?
- can we add displaying of the value on the horizontal bar charts as before?
- how does it look like in dark mode?
- the tooltips: can we all make them native rechart based style?
- what about the tremor based charts e.g. in the dashboards scores etc?

(feel free to reply on slack), thanks !

web/src/features/dashboard/components/TracesBarListChart.tsx
```diff
@@ -98,6 +102,10 @@ export const TracesBarListChart = ({
     ? transformedTraces.slice(0, maxNumberOfEntries.expanded)
     : transformedTraces.slice(0, maxNumberOfEntries.collapsed);
 
+  // Height scales with bar count so each bar keeps the same height when expanding (legacy behavior)
```
nimarb: what does legacy behaviour mean here - what are we trying to do

web/src/features/dashboard/lib/tremorv4-recharts-chart-adapters.ts
```diff
@@ -0,0 +1,145 @@
+import type { DataPoint } from "@/src/features/widgets/chart-library/chart-props";
+import type { TimeSeriesChartDataPoint } from "@/src/features/dashboard/components/BaseTimeSeriesChart";
+import type { ChartBin } from "@/src/features/scores/types";
+import {
+  dashboardDateRangeAggregationSettings,
+  type DashboardDateRangeAggregationOption,
+} from "@/src/utils/date-range-utils";
+
```
nimarb, line 13: can we add a top level comment here why this file is needed and what it does? also, lets not call it legacy because that doesn't mean anything. lets rather say tremorv4-chart-adapter or similar to make it specific

web/src/features/widgets/chart-library/AreaChartTimeSeries.tsx
```diff
@@ -0,0 +1,139 @@
...
+  getUniqueDimensions,
+  groupDataByTimeDimension,
+} from "@/src/features/widgets/chart-library/utils";
+import { compactNumberFormatter } from "@/src/utils/numbers";
+import { cn } from "@/src/utils/tailwind";
+
+/**
+ * AreaChartTimeSeries component
+ * Same data shape as LineChartTimeSeries; uses Recharts Area for filled series.
+ */
```
nimarb: in general, we can remove those types of comments. for agents, they just cause context bloat and don't add any information

web/src/features/widgets/chart-library/chart-props.ts
```diff
@@ -6,8 +6,18 @@ export interface DataPoint {
...
   config?: ChartConfig;
   accessibilityLayer?: boolean;
+  /** Optional formatter for tooltip values (e.g. USD, compact number). */
+  valueFormatter?: (value: number) => string;
+  /** Legend placement. 'above' = horizontal legend row above the chart. */
+  legendPosition?: LegendPosition;
+  /** Show numeric value to the right of each bar (horizontal bar chart). */
+  showValueLabels?: boolean;
+  /** Show dots on line/area data points. Default true. */
+  showDataPointDots?: boolean;
```
nimarb: can remove those comments i think

web/src/features/widgets/chart-library/Chart.tsx
```diff
@@ -33,43 +37,84 @@ export const Chart = ({
...
+      const time_dimension = Number.isNaN(parsed.getTime())
+        ? item.time_dimension
+        : parsed.toLocaleTimeString("en-US", {
+            year: "2-digit",
+            month: "numeric",
+            day: "numeric",
+            hour: "2-digit",
+            minute: "2-digit",
+          });
+      return { ...item, time_dimension };
```
nimarb, line 83: did you test if this works with another locale as well? i,e, is correct if you set your browser to a different locale

web/src/features/widgets/chart-library/LineChartTimeSeries.tsx
```diff
@@ -24,43 +30,116 @@ export const LineChartTimeSeries: React.FC<ChartProps> = ({
...
-            type="monotone"
-            dataKey={dimension}
-            strokeWidth={2}
-            dot={true}
-            activeDot={{ r: 6, strokeWidth: 0 }}
-            stroke={`hsl(var(--chart-${(index % 4) + 1}))`}
+    <div className="flex h-full w-full min-w-0 flex-col">
+      {legendPosition === "above" && dimensions.length > 0 && (
+        <div className="min-w-0 shrink-0 overflow-x-auto overflow-y-hidden pb-3">
+          <div className="flex w-max min-w-full flex-nowrap justify-end gap-4">
```
nimarb: lots of tailwind classes here - do we really need all of those? seems excessive

web/src/features/widgets/chart-library/LineChartTimeSeries.tsx
```diff
@@ -24,43 +30,116 @@ export const LineChartTimeSeries: React.FC<ChartProps> = ({
...
+                  }
+                >
+                  <div
+                    className="h-2 w-2 shrink-0 rounded-[2px]"
+                    style={{
+                      backgroundColor: `hsl(var(--chart-${(index % 8) + 1}))`,
+                    }}
+                  />
+                  <span className="text-muted-foreground">{dimension}</span>
+                </button>
```
nimarb, line 82: are we intentionally using a custom button here?

web/src/features/widgets/chart-library/PieChart.tsx
```diff
@@ -31,7 +37,7 @@ export const PieChart: React.FC<ChartProps> = ({
     return data.map((item, index) => ({
       name: item.dimension || "Unknown",
       value: item.metric,
-      fill: `hsl(var(--chart-${(index % 4) + 1}))`,
+      fill: `hsl(var(--chart-${(index % 8) + 1}))`,
```
nimarb, line 41: why  a different split here for fill colour

web/src/features/widgets/chart-library/VerticalBarChartTimeSeries.tsx
```diff
@@ -49,14 +55,23 @@ export const VerticalBarChartTimeSeries: React.FC<ChartProps> = ({
           <Bar
             key={dimension}
             dataKey={dimension}
-            stroke={`hsl(var(--chart-${(index % 4) + 1}))`}
-            fill={`hsl(var(--chart-${(index % 4) + 1}))`}
+            stroke={`hsl(var(--chart-${(index % 8) + 1}))`}
+            fill={`hsl(var(--chart-${(index % 8) + 1}))`}
             // Stack bars if there are multiple dimensions
```
nimarb, line 62: why does this change

web/src/styles/globals.css
```diff
@@ -74,10 +74,15 @@
...
-    --chart-4: var(--dark-red);
+    /* Legacy chart palette (matches getColorsForCategories: indigo, cyan, zinc, purple, yellow, red, lime, pink) */
+    --chart-1: 239 84% 58%;
+    --chart-2: 188 94% 43%;
+    --chart-3: 240 4% 46%;
+    --chart-4: 271 91% 65%;
+    --chart-5: 48 96% 47%;
+    --chart-6: 0 84% 60%;
+    --chart-7: 84 81% 44%;
+    --chart-8: 330 81% 60%;
```
nimarb: what does legacy mean here -> can we delete those values after we have finally deployed without beta toggle?

web/src/styles/globals.css
```diff
@@ -166,10 +171,15 @@
...
-    --chart-4: var(--dark-red);
+    /* Legacy chart palette (same as light for consistent chart colors) */
+    --chart-1: 239 84% 58%;
+    --chart-2: 188 94% 43%;
+    --chart-3: 240 4% 46%;
+    --chart-4: 271 91% 65%;
+    --chart-5: 48 96% 47%;
+    --chart-6: 0 84% 60%;
+    --chart-7: 84 81% 44%;
+    --chart-8: 330 81% 60%;
```
nimarb: same

web/src/features/widgets/chart-library/Chart.tsx
coffee4tw: I hadn't but just did and it works fine for the respective time zones. Formatting is still "en-US" of course since we specify that. Do we want to be less geo-specific?

web/src/features/widgets/chart-library/LineChartTimeSeries.tsx
coffee4tw: I agree and I kind of hate it. I replaced/removed two but the rest seems necessary. I am certainly no expert though and I would love to bring this whole thing into the recharts native Legend but couldn't get that to work with the same interactivity and the existing chart.tsx/ChartLegendContent.

So here is what each of them do:

Here’s what each remaining class does and why it’s there:

---

**Line 51 – outer wrapper:**  
`className="flex min-w-0 flex-col size-full"`

| Class | What it does | Why we need it |
|-------|----------------|----------------|
| `flex` | Makes this div a flex container. | So we can stack the legend and chart vertically and control how they share space. |
| `flex-col` | Flex direction is column (stack children vertically). | Puts the legend above and the chart below. |
| `min-w-0` | Sets `min-width: 0` on this flex item. | Flex items default to `min-width: auto`, so they won’t shrink below their content. In a narrow card, that can force horizontal overflow. `min-w-0` lets this container shrink so the layout can stay within the card and the inner scroll (on the legend) can work. |
| `size-full` | Sets `width: 100%` and `height: 100%`. | So this chart block fills the space it’s given (e.g. the card content area). |

---

**Line 53 – legend scroll container:**  
`className="min-w-0 shrink-0 overflow-x-auto pb-3"`

| Class | What it does | Why we need it |
|-------|----------------|----------------|
| `min-w-0` | Same as above: allows this flex child to shrink. | The parent is a flex column. Without this, this div wouldn’t shrink below the width of its content, so the horizontal scroll area wouldn’t get a bounded width and `overflow-x-auto` wouldn’t kick in properly. |
| `shrink-0` | Sets `flex-shrink: 0`. | Keeps the legend row from being squashed when vertical space is tight. The chart (with `flex-1`) is the one that should shrink; the legend keeps its height. |
| `overflow-x-auto` | Horizontal scroll when content is wider than the container. | When there are many series, the legend items can overflow. This lets the user scroll the legend horizontally instead of breaking the layout. |
| `pb-3` | Padding below the legend. | Adds space between the legend and the chart. |

---

**Line 54 – legend items row:**  
`className="flex min-w-full w-max flex-nowrap justify-end gap-4"`

| Class | What it does | Why we need it |
|-------|----------------|----------------|
| `flex` | Flex container for the legend buttons. | Puts the legend items in a row. |
| `flex-nowrap` | Prevents wrapping to a new line. | Keeps all items on one row so we get a single horizontal strip that can scroll, instead of wrapping. |
| `justify-end` | Aligns items to the end of the row (right in LTR). | Puts the legend on the right side of the chart. |
| `gap-4` | Space between each legend item. | Avoids the buttons touching each other. |
| `min-w-full` | Sets `min-width: 100%`. | Makes this row at least as wide as its parent. So when there are only a few items, the row is still full width and `justify-end` actually moves them to the right. Without it, the row would only be as wide as the buttons and they’d sit on the left. |
| `w-max` | Sets `width: max-content`. | Makes the row as wide as its content when there are many items. So the row can grow wider than the container, the parent gets a horizontal scrollbar (`overflow-x-auto`), and we don’t truncate or wrap the legend. |

`min-w-full` and `w-max` together mean: “Be at least as wide as the container (so `justify-end` works when there are few items), but if content is wider, grow with it (so many items scroll horizontally).”

web/src/features/widgets/chart-library/LineChartTimeSeries.tsx
coffee4tw: Yeah, it's not really a button though but the clickable legend item aka data line in the chart.
Open to better suggestions.

web/src/features/widgets/chart-library/PieChart.tsx
coffee4tw: Previously the recharts based charts were only using 4 defined colors, the tremor based ones had 8. When I replaced recharts ones with the tremor color palette I added more colors for the charts to use and so they should do that.

web/src/features/widgets/chart-library/VerticalBarChartTimeSeries.tsx
coffee4tw: Same as above.

web/src/styles/globals.css
coffee4tw: Removed the comment. No, these are the colors for the recharts based charts, they just changed to adhere to the tremor based charts color palette.

PR #12048, opened 2026-02-16
chore(events-table): more than 100 observations per trace through tRPC

nimarb:
> [!IMPORTANT]
> Removes 100 observation limit per trace, sets new limit to 10,000, and updates error handling in related components.
> 
>   - **Behavior**:
>     - Removes 100 observation limit per trace in `getObservationsForTraceFromEventsTable` in `events.ts`.
>     - Introduces `MAX_OBSERVATIONS_PER_TRACE` set to 10,000 in `events.ts`.
>     - If trace exceeds 10,000 observations, throws `BAD_REQUEST` error in `eventsRouter.ts`.
>   - **Components**:
>     - `TracePage.tsx`: Displays error page if `eventsData.error` is present, indicating too many observations.
>   - **Hooks**:
>     - `useEventsTraceData.ts`: Uses `byTraceId` query to fetch observations, handles `BAD_REQUEST` error.
>     - `useEventsTableData.ts`: Removes comment about fetching observations, no functional change.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 6586bcdcba33a077cc47874eb87d2a4909ac0be6. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/shared/src/server/repositories/events.ts
+++ b/packages/shared/src/server/repositories/events.ts
@@ -49,7 +49,6 @@ import {
   queryClickhouse,
   queryClickhouseStream,
 } from "./clickhouse";
-import { TRACE_TO_OBSERVATIONS_INTERVAL } from "./constants";
 import { ObservationRecordReadType, TraceRecordReadType } from "./definitions";
 import type { AnalyticsObservationEvent } from "../analytics-integrations/types";
 import {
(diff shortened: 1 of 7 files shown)


web/src/features/events/server/eventsRouter.ts
```diff
@@ -176,6 +179,44 @@ export const eventsRouter = createTRPCRouter({
...
+          const { observations, totalCount } =
+            await getObservationsForTraceFromEventsTable({
+              projectId: ctx.session.projectId,
+              traceId: input.traceId,
+              timestamp: input.timestamp,
+            });
+
+          if (totalCount > MAX_OBSERVATIONS_PER_TRACE) {
+            throw new TRPCError({
+              code: "BAD_REQUEST",
```
sumerman: Not sure if bad request is the best UX here. I would rather indicate that there are even more entries, but we aren't going to load them.

packages/shared/src/server/repositories/events.ts
```diff
@@ -283,6 +284,71 @@ const TRACES_ORDER_BY_COLUMNS = TRACES_FROM_EVENTS_UI_COLUMN_DEFINITIONS.filter(
...
+  const totalCount = records.length;
+
+  const withModelData = await enrichObservationsWithModelData(
+    records.slice(0, MAX_OBSERVATIONS_PER_TRACE),
+    projectId,
+    false,
+    null,
+  );
+
+  const observations = await enrichObservationsWithTraceFields(withModelData);
```
sumerman, line 336: This is expensive. I would actually show 10K + warning, but should you decide to stick with bad request it would be best to avoid calling `enrichObservationsWithTraceFields`.

web/src/features/events/server/eventsRouter.ts
```diff
@@ -176,6 +179,44 @@ export const eventsRouter = createTRPCRouter({
...
+              message: `Trace has ${totalCount} observations, exceeding the maximum of ${MAX_OBSERVATIONS_PER_TRACE} for the trace detail view.`,
```
sumerman:
```suggestion
              message: `Trace exceeds the maximum of ${MAX_OBSERVATIONS_PER_TRACE} for the trace detail view.`,
``` 

otherwise it's misleading, it'll always be 10_001

packages/shared/src/server/repositories/events.ts
```diff
@@ -283,6 +284,71 @@ const TRACES_ORDER_BY_COLUMNS = TRACES_FROM_EVENTS_UI_COLUMN_DEFINITIONS.filter(
   queryPrefix: "t", // Use 't' prefix because we're selecting from traces CTE
 }));
 
+// TODO: remove limit?
+export const MAX_OBSERVATIONS_PER_TRACE = 10_000;
```
sumerman, line 287: I would get it down to an even smaller value like 1k. If there's a common complaint we should work on paginating.

nimarb: i'd start off with 10k and rather implement pagination if this breaks stuff. shouldn't be many traces. currently, there is no limit at all and works fine usualy, also on traces with 12k obs for example.

PR #12206, opened 2026-02-23
fix(filter-sidebar): use positive matching for arrayOptions checkbox filter

### LINKED ISSUE #12202: bug: Tag sidebar filter checkboxes produce incorrect results due to "none of" operator inversion
### Describe the bug

Hi, thanks for building such a great project!

I found an issue with the sidebar tag filter when using checkboxes to select/deselect tags.
My understanding of the SOME/ALL toggle (verified in clickhouse-filter.ts):
- SOME (any of): OR — show traces that have at least one of the selected tags
- ALL (all of): AND — show traces that have all of the selected tags

When using checkboxes to filter tags, the results are incorrect. The root cause is in useSidebarFilterState.tsx — the applySelection function (lines 602-631)
stores the filter as "none of" [deselected tags] instead of "any of" [selected tags]. These are not equivalent for array columns because a trace can have multiple tags.

Example: Available tags: [1, 2, 3, 4, 5]. User checks only tag 1 via checkboxes.
| trace tags     | Current: none of [2,3,4,5] | Expected: any of [1] |
|---------------|----------------------------|----------------------|
| [1]           | shown                      | shown                |
| [1, 2]        | hidden (has 2)             | shown                |
| [] (no tags)  | shown (no 2,3,4,5)         | hidden               |

The same issue applies when multiple tags are checked — e.g., checking [1, 2, 3] creates none of [4, 5], which excludes traces like [1, 4] that should match.
Note: clicking a tag label (not checkbox) triggers "Only" mode which correctly uses "any of", but this only allows single-tag selection — there is no way to correctly filter by multiple tags via OR/AND using the checkbox UI.

Interestingly, there is a workaround that works correctly:
- Switch the toggle to ALL first (all checkboxes become unchecked)
- Switch back to SOME
- Then check tags via checkboxes → filtering works correctly

This works because toggling ALL → SOME creates an existing "any of" filter internally, which causes applySelection to skip the broken "none of" inversion code path and use "any of" with selected values directly.
Related: #11606 (may share the same root cause)

If I've misunderstood anything about the intended behavior, please feel free to leave a comment — happy to discuss!

### Steps to reproduce

Broken flow:
1. Navigate to the Traces page with traces that have various tags
2. In the sidebar "Tags" filter, all tags are checked by default
3. Uncheck tags using checkboxes until only one or a few remain
4. Observe: traces with no tags appear in results, and traces that have both a selected and deselected tag are excluded

Working flow (workaround):
1. Switch toggle to ALL (checkboxes clear)
2. Switch toggle back to SOME
3. Check desired tags via checkboxes
4. Filtering works correctly

Self-hosted

3.155.0

This is a UI-only bug in the sidebar filter component, not related to any SDK

#### working flow result
Image

#### broken flow result
Image

Yes

gnsals0904:
## What does this PR do?

Fix incorrect filtering when selecting multiple tags via sidebar checkbox.

When using checkbox toggle on tag filters, the `applySelection` function used `"none of"` inversion (storing deselected items). This is semantically wrong for array columns like tags, because a trace with tags `[tag-1, tag-3]` matches `"any of [tag-1, tag-2]"` but is incorrectly excluded by `"none of [tag-3, tag-4, tag-5]"` (since it contains `tag-3`).

Extract `resolveCheckboxOperator` as a pure function that uses `"any of"` positive matching for `arrayOptions` and preserves the existing `"none of"` inversion for single-valued `stringOptions`.

Fixes #12202 

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Fixes incorrect filtering for arrayOptions in `useSidebarFilterState.tsx` by introducing `resolveCheckboxOperator` for correct operator logic, with tests added in `filter-integration.clienttest.ts`.
> 
>   - **Behavior**:
>     - Fixes incorrect filtering for arrayOptions in `useSidebarFilterState.tsx` by using positive matching ("any of") instead of inversion ("none of").
>     - Introduces `resolveCheckboxOperator` to determine operators for checkbox filters, ensuring correct logic for arrayOptions and stringOptions.
>   - **Tests**:
>     - Adds tests for `resolveCheckboxOperator` in `filter-integration.clienttest.ts` to verify correct operator logic for arrayOptions and stringOptions.
>   - **Misc**:
>     - Refactors `applySelection` in `useSidebarFilterState.tsx` to use `resolveCheckboxOperator` for determining filter operators.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 63352b846eaca85329a0ec08307cb02b2f43c048. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/web/src/features/filters/filter-integration.clienttest.ts
+++ b/web/src/features/filters/filter-integration.clienttest.ts
@@ -18,7 +18,10 @@ import { traceFilterConfig } from "./config/traces-config";
 import { observationFilterConfig } from "./config/observations-config";
 import { transformFiltersForBackend } from "./lib/filter-transform";
 import { sessionFilterConfig } from "./config/sessions-config";
-import { decodeAndNormalizeFilters } from "./hooks/useSidebarFilterState";
+import {
+  decodeAndNormalizeFilters,
+  resolveCheckboxOperator,
+} from "./hooks/useSidebarFilterState";
 
 // Helper to simulate complete URL flow
 function simulateUrlFlow(filters: FilterState): FilterState {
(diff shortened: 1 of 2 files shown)


nimarb:
thank you contribution, especially for the analysis and clear articulation of the issue + tests @gnsals0904 !

PR #12215, opened 2026-02-23
fix(evals): parse metadata values in observation-level evaluators

### LINKED ISSUE #12125: bug: Observation-level evaluators serialize nested metadata values as strings instead of parsing them as JSON objects
### Describe the bug

When using observation-level LLM-as-a-Judge evaluators (migrated from trace-level per the [migration guide](https://langfuse.com/faq/all/llm-as-a-judge-migration)), JSONPath extraction on metadata returns nested objects as serialized JSON strings instead of navigable objects. The same JSONPath works correctly with trace-level evaluators.

**Expected behavior:** JSONPath on nested metadata objects should work the same way in observation-level and trace-level evaluators.

**Actual behavior:** Observation-level evaluators serialize nested metadata values as JSON strings instead of keeping them as navigable objects. JSONPath cannot traverse into them.

### Steps to reproduce

1. Create a generation observation with a nested object in metadata:

```python
@observe(as_type="generation")
async def my_function():
    langfuse.update_current_generation(
        metadata={
            "nested_object": {
                "key": "value"
            }
        }
    )
```

This produces metadata with the following structure:

```json
{
  "nested_object": {
    "key": "value"
  }
}
```

2. Create two evaluators with the same variable mapping:
   - **Trace-level evaluator** — Source: `metadata`, JSONPath: `$.nested_object`
   - **Observation-level evaluator** — Source: `metadata`, JSONPath: `$.nested_object`

3. Run both evaluators on the same trace/observation.

### Result with `$.nested_object` (whole nested object)

| Evaluator level | Result |
|---|---|
| Trace-level | `[{"key": "value"}]` — parsed object |
| Observation-level | `["{\"key\": \"value\"}"]` — serialized string |

*(screenshot with real data)*
Trace-level:
Image

Observation-level:
Image

### Result with `$.nested_object.key` (navigating into nested value)

| Evaluator level | Result |
|---|---|
| Trace-level | Returns `"value"` correctly |
| Observation-level | Returns nothing — cannot navigate into a string |

### Note on configuration preview

The preview shown during evaluator setup **correctly** resolves the variable in both cases. The issue only manifests during actual evaluation execution.

*(screenshot from observation level evaluator)*
Image
Image

Langfuse Cloud

_No response_

- Langfuse Cloud v3.154.1
- Langfuse Python SDK v3.12.0
- Using `@observe(as_type="generation")` decorator
- Metadata set via `langfuse.update_current_generation(metadata={...})`

### Root cause analysis

I traced the issue to a difference in how metadata values are processed between the two evaluation paths.

**Trace-level path:** Trace metadata goes through [`parseMetadataCHRecordToDomain`](https://github.com/langfuse/langfuse/blob/main/packages/shared/src/server/utils/metadata_conversion.ts), which calls `parseJsonPrioritised()` on **each metadata value individually**. This converts string-encoded JSON values back into objects before JSONPath is applied:

```
{"input_variables": "{\"text\":\"...\"}"} → {"input_variables": {"text": "..."}}
```

**Observation-level path:** [`convertEventRecordToObservationForEval`](https://github.com/langfuse/langfuse/blob/main/packages/shared/src/features/evals/observationForEval.ts) only runs `observationForEvalSchema.parse(record)`, which validates the shape but does not parse individual metadata values. They remain as serialized strings.

Then in [`extractObservationVariables`](https://github.com/langfuse/langfuse/blob/main/worker/src/features/evaluation/observationEval/extractObservationVariables.ts), `applyJsonSelector` does `typeof value === "string" ? JSON.parse(value) : value`. This handles the case where the **entire** metadata field is a string, but since `observation.metadata` is already an object (with string *values*), it passes through unparsed. The nested string values are never deserialized.

### Suggested fix

Add a metadata parsing step (equivalent to `parseMetadataCHRecordToDomain`) in the observation evaluation path — either in `convertEventRecordToObservationForEval` or before JSONPath is applied in `extractObservationVariables`.

### Related issues

- #5917 — Fixed the same JSONPath/serialization issue for the trace-level evaluator path
- #8151 — Feature request that introduced observation-level evaluators

Yes

cristiangb13:
## What does this PR do?

Observation-level evaluators fail to extract nested metadata values via JSONPath. ClickHouse stores metadata values as serialized JSON strings (`Record<string, string>`), but `convertEventRecordToObservationForEval` passes them through without parsing — so JSONPath expressions like `$.metadata.input_variables.company_name` cannot navigate into nested objects.

The trace-level code path already handles this correctly with `parseMetadataCHRecordToDomain`. This PR applies the same parsing to the observation-level path in `convertEventRecordToObservationForEval`.

Fixes https://github.com/langfuse/langfuse/issues/12125

## Type of change

## Mandatory Tasks

## Checklist

- [x] I have read the [contributing guide](https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md)
- [x] My code follows the style guidelines of this project (`pnpm run format`)
- [x] New and existing unit tests pass locally with my changes

----

> [!IMPORTANT]
> Fixes metadata parsing in `convertEventRecordToObservationForEval` to support JSONPath navigation into nested objects, with tests added for validation.
> 
>   - **Behavior**:
>     - Fixes metadata parsing in `convertEventRecordToObservationForEval` in `observationForEval.ts` to allow JSONPath navigation into nested objects.
>     - Uses `parseMetadataCHRecordToDomain` for parsing metadata.
>   - **Tests**:
>     - Adds tests in `observationForEvalSchema.test.ts` to verify JSON string metadata parsing, simple string preservation, and null handling.
>     - Tests JSONPath navigation with `extractObservationVariables`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for edde861668728b740f6427d1272947e53da39d28. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


tminot18-web:
Nice fix — aligning the observation-level path with parseMetadataCHRecordToDomain makes sense and brings parity with the trace-level behavior.

One thought: since both paths now depend on identical metadata parsing semantics, would it make sense to add a parity-style test that asserts trace-level and observation-level evaluators produce identical results for nested metadata cases (objects, arrays, invalid JSON, etc.)?

That might help prevent future drift if one path changes independently.

wochinge:
@cristiangb13 Thanks for the contribution! 

I just discovered that this is already fixed on `main` - probably due to a mix of #12394 and #12474 .
I'll close for now - please feel free to re-open if you think this is still necessary.

PR #12229, opened 2026-02-24
fix(filters): decode empty arrayOptions value as empty array in URL roundtrip

gnsals0904:
## What does this PR do?

Fix ALL mode tag filter returning 0 results when switching from SOME with no checkboxes selected.

When switching from SOME to ALL with no checkboxes selected, `updateOperator` creates a filter with an empty array (`value: []`). This empty array gets corrupted during the URL encoding/decoding round-trip:

1. **Encode**: The empty array is joined into an empty string, so the URL stores no value
2. **Decode**: The empty string is restored as `[""]` — an array containing one empty string element
3. **Query**: ClickHouse runs `hasAll(tags, [''])`, which matches nothing → 0 results
4. **Subsequent checkbox clicks**: The new tag is appended to the corrupted array (e.g. `["", "tag-1"]`), so results remain 0

In `decodeFiltersGeneric`, when the decoded value is an empty string for `arrayOptions` type, return `[]` instead of `[""]`. This is safe because empty strings are never valid values for array columns like tags, while `stringOptions` (e.g., trace name) still correctly allows `[""]`.

Fixes #12228

## Type of change

## Mandatory Tasks

## Checklist

----

> [!IMPORTANT]
> Fixes URL encoding/decoding bug for empty `arrayOptions` in `decodeFiltersGeneric` to prevent incorrect query results.
> 
>   - **Behavior**:
>     - Fixes bug in `decodeFiltersGeneric` in `filter-query-encoding.ts` to decode empty `arrayOptions` as `[]` instead of `['']`.
>     - Ensures `hasAll(tags, [''])` does not return 0 results by decoding to `[]`.
>   - **Tests**:
>     - Adds test case in `filter-integration.clienttest.ts` to verify round-trip encoding/decoding of empty `arrayOptions` results in `[]`.
>     - Confirms that switching from SOME to ALL with no checkboxes selected decodes correctly.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 655fc76b8363eb27cb6903e010d2fcf815d29eba. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


nimarb:
thank you !

PR #12300, opened 2026-02-26
fix(auth): support Authentik authorization URL override

markwoodford:
Fixes #11884

## Summary
- Add optional `AUTH_AUTHENTIK_AUTHORIZATION_URL` env var to allow browser redirects to use an external Authentik URL.
- Keep issuer-based server-side validation while deriving token, userinfo, and JWKS endpoints from the issuer when override mode is enabled.
- Relax `AUTH_AUTHENTIK_ISSUER` validation to allow an optional trailing slash.

## Test plan
- [ ] Configure `AUTH_AUTHENTIK_ISSUER` to an internal host and `AUTH_AUTHENTIK_AUTHORIZATION_URL` to a browser-reachable host.
- [ ] Click Authentik login and verify redirect goes to the external authorization URL.
- [ ] Complete login and verify token/userinfo/JWKS calls still work.
- [ ] Verify existing Authentik setups without the new env var behave unchanged.

----

> [!IMPORTANT]
> Add support for overriding Authentik authorization URL via new env var, while maintaining server-side validation and relaxing issuer validation.
> 
>   - **Behavior**:
>     - Add `AUTH_AUTHENTIK_AUTHORIZATION_URL` env var in `env.mjs` to override Authentik authorization URL for browser redirects.
>     - Maintain server-side validation by deriving token, userinfo, and JWKS endpoints from issuer in `auth.ts`.
>     - Relax `AUTH_AUTHENTIK_ISSUER` validation to allow optional trailing slash in `env.mjs`.
>   - **Implementation**:
>     - Modify `auth.ts` to handle new `AUTH_AUTHENTIK_AUTHORIZATION_URL` by setting custom authorization, token, userinfo, and JWKS URLs when override is enabled.
>     - Update regex for `AUTH_AUTHENTIK_ISSUER` in `env.mjs` to allow optional trailing slash.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for e27d9632e8bcf748dc0ea67f1710595782442349. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


web/src/env.mjs
```diff
@@ -133,14 +133,15 @@ export const env = createEnv({
     AUTH_AUTHENTIK_CLIENT_SECRET: z.string().optional(),
     AUTH_AUTHENTIK_ISSUER: z
       .string()
-      .regex(/^https:\/\/.+\/application\/o\/[^/]+$/, {
+      .regex(/^$|^https:\/\/.+\/application\/o\/[^/]+\/?$/, {
         message:
-          "Authentik issuer must be in format https://<domain>/application/o/<slug> without trailing slash",
+          "Authentik issuer must be in format https://<domain>/application/o/<slug> with optional trailing slash, or empty to disable",
```
Steffen911: What are cases where we want to accept an empty issuer? My expectation would be that it's always required. Is having `^$` in the regex as an option desirable?

web/src/server/auth.ts
```diff
@@ -225,22 +225,45 @@ if (
...
+      url: env.AUTH_AUTHENTIK_AUTHORIZATION_URL,
+      params: { scope: "openid email profile" },
+    };
+    authentikProvider.token = {
+      url: `${authentikBase}token/`,
+    };
+    authentikProvider.userinfo = {
+      url: `${authentikBase}userinfo/`,
+    };
+    authentikProvider.jwks_endpoint = authentikJwksUrl;
```
Steffen911, line 273: I would appreciate a comment that explains those replacements on a high level as they seem to encode some assumptions about authentik-specific URL formats and behaviours. Having a doc string at the start of the `if` block would help any developer who'd touch this in the future.

Steffen911 (commented):
Thank you a lot for creating this! 🙌

Steffen911:
Corresponding docs PR once this is merged: https://github.com/langfuse/langfuse-docs/pull/2577

web/src/server/auth.ts
markwoodford: added a comment

web/src/env.mjs
markwoodford: Reverted

Steffen911:
Could you also address the linting reports?

PR #12336, opened 2026-03-01
fix(auth): fall back to "mail" attribute in custom SSO profile

Br1an67:
## What does this PR do?

Some SSO/IdP providers (e.g. PingFederate, LDAP-backed) return the email address under the `mail` attribute instead of the standard OIDC `email` claim. This causes a "No email found in user object" error during sign-in.

This PR adds a fallback in the custom SSO provider profile function: when the configured email claim (default: `email`, configurable via `LANGFUSE_CUSTOM_SSO_EMAIL_CLAIM`) is absent from the profile object, it also checks the `mail` attribute before giving up.

Users can still explicitly set `LANGFUSE_CUSTOM_SSO_EMAIL_CLAIM=mail` as a workaround, but this fallback handles the most common case automatically.

Fixes #12334

## Type of change

## Mandatory Tasks

----

> [!IMPORTANT]
> Fixes email retrieval in `CustomSSOProvider` by falling back to `mail` attribute if `CUSTOM_EMAIL_CLAIM` is absent.
> 
>   - **Behavior**:
>     - In `customSsoProvider.ts`, `CustomSSOProvider` function now falls back to `mail` attribute if `CUSTOM_EMAIL_CLAIM` is absent in the profile.
>     - Handles cases where SSO/IdP providers return email under `mail` instead of `email`.
>   - **Configuration**:
>     - Users can still set `LANGFUSE_CUSTOM_SSO_EMAIL_CLAIM=mail` explicitly, but fallback is automatic for common cases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 31728dadca0bbc8f6913fd269c6cafda41968037. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Steffen911:
@Br1an67 Could you accept the CLA?

Br1an67:
@Steffen911 Hi! The CLA has already been signed — the CLAassistant bot confirms "All committers have signed the CLA" ✅

Regarding the `test-docker-build` failure: it's a transient CI infrastructure issue — the `wget` downloading `golang-migrate` during the Docker build got `Resource temporarily unavailable`. Not related to the PR changes. A re-run should fix it. Let me know if you'd like me to do anything else!

Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

Br1an67:
I cleaned up the branch and force-pushed it so the PR now only contains the SSO "mail" fallback change (no unrelated commits). This should also clear the CLA issue from additional committers. Could you please re-run the CLA check? Thanks!

Steffen911 (approved):

packages/shared/src/server/auth/customSsoProvider.ts
```diff
@@ -26,7 +26,8 @@ export function CustomSSOProvider<P extends CustomSSOUser>(
       return {
         id: profile[CUSTOM_SUB_CLAIM],
         name: profile[CUSTOM_NAME_CLAIM],
-        email: profile[CUSTOM_EMAIL_CLAIM],
+        // Fall back to "mail" (LDAP/AD-style) if the configured claim is absent
+        email: profile[CUSTOM_EMAIL_CLAIM] ?? profile["mail"],
```
Steffen911, line 30: I think doing this silently is weird. Can you add a fallback chain the first tries CUSTOM_EMAIL_CLAIM, then `email` as plain string in case the original env was overwritten, and thirdly falls back to `mail`?
I'd also add a info level log message everytime this fallback happens to that administrators can check whether they need to update their claim env setting.

Steffen911: @Br1an67 Do you want to take a look into this?

PR #12359, opened 2026-03-03
chore(experiments): add single experiment paginated list view

marliessophie:
> [!IMPORTANT]
> Add a paginated list view for experiment items with backend support for querying and filtering, and update the UI to display these items in a detailed table.
> 
>   - **Backend**:
>     - Add `getExperimentItemsFromEvents`, `getExperimentItemsCountFromEvents`, and `getExperimentItemMetricsFromEvents` functions in `experiments.ts` for querying experiment items.
>     - Update `router.ts` to include new endpoints for fetching experiment items and their metrics.
>     - Add `experimentItemPreAggCols` and `experimentItemPostAggCols` in `mapExperimentItemsTable.ts` for column mappings.
>   - **UI Components**:
>     - Add `ExperimentItemsTable` component in `ExperimentItemsTable.tsx` to display experiment items with features like sorting, filtering, and pagination.
>     - Add `ExperimentOverviewPanel` component in `ExperimentOverviewPanel.tsx` for displaying experiment details.
>     - Update `experiment-items-filter-config.ts` for filter configuration of experiment items.
>   - **Misc**:
>     - Add `experiment-items` to `TableName` in `types.ts` and `table-view-presets.ts`.
>     - Update `useExperimentItemsTableData` and `useExperimentItemsFilterOptions` hooks for data fetching and filter options.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for 4bf4f2f668e8ac7f77a759330fb044a4ac8a8e74. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


marliessophie:
### Backend review summary 

#### event-query-builder.ts

- Added `experimentItems` field set so experiment-item list/batch queries can select a stable, reusable projection instead of ad-hoc raw columns.
- Added `experiment_item_id` to aggregation fields (`argMaxIf`) so grouped event aggregations can preserve item identity.
- Extended `EventsAggQueryBuilder` to emit `HAVING`, which is critical for post-group constraints (e.g., “baseline must be present” after grouping by `experiment_item_id`).

#### query-fragments.ts
- Added `eventsExperimentsRootSpans(...)`: a reusable fragment for “experiment events, root-span only, optionally narrowed to experiment item IDs”.
- Extended `eventsScoresAggregation` params with `hasScoreAggregationFilters` for API symmetry/documented behavior around score-filter paths.

#### experiments.ts
- Introduced full experiment-item repository layer (count, list, batch I/O, experiment-name lookup).
- Core list/count query pattern is intentionally two-step for UI correctness:
  - Query 1 (`qualification`): compute paginated `item_id`s using per-experiment filters, grouped by `experiment_item_id`.
  - Query 2 (`data fetch`): fetch all selected experiment-items for only those qualified item IDs.
- `qualification plan` exists to translate UI “filters target specific experiment” into SQL:
  - Per-experiment filters are compiled as `(experiment_id = X AND filters_for_X)`, then OR-combined.
  - Score CTE joins are conditional (only when score filters are present).
  - Optional baseline inclusion uses `HAVING countIf(...)` after grouping, so baseline gating is applied at item level, not event row level. Baseline gating is necessary to allow user control if they wish to see items only present in baseline or across all selected runs. 
- Result shape matches UI intent: one row per item with stacked per-experiment values. Please note that our data model allowed multiple traces-per-item within each run, but our UI always limited this to showing only the one trace-per-item. In the scope of this PR, this is the desired behaviour.  
- Count query reuses the same qualification logic, keeping pagination and total count consistent.

_Please note that **UI** currently always sets a baseline but we are adding a "clear baseline" functionality in a follow up PR, so queries must run if no baseline is set_

packages/shared/src/server/repositories/experiments.ts
```diff
@@ -370,3 +377,596 @@ const getExperimentsFromEventsGeneric = async <T>(
...
+  const queryBuilder = new EventsAggQueryBuilder({
+    projectId,
+    groupByColumn: "e.experiment_item_id",
+    selectExpression: "e.experiment_item_id as item_id",
+  })
+    .whereRaw("e.span_id = e.experiment_item_root_span_id")
+    .when(hasScoreFilters, (b) =>
+      b.withCTE(
+        "scores_agg",
+        // Optionally add timestamp >= oldest_selected_experiment_start as a coarse partition prune
```
marliessophie, line 663: Valeriy: I'd like your input if you believe adding the timestamp of the oldest experiment to filter will be sufficient here or if we need to resort to building something like candidate_traces (given experiment params) first and joining scores of those candidate ids inside the score CTE?

web/src/__tests__/server/repositories/experiment-items-repository.servertest.ts
```diff
@@ -0,0 +1,1200 @@
...
+                column: "trace_scores_avg",
+                type: "numberObject",
+                key: "accuracy",
+                operator: ">",
+                value: 0.5,
+              } as FilterCondition,
+            ],
+          },
+        ],
+        orderBy: { column: "itemId", order: "ASC" },
```
sumerman: AFAICT orderBy is ignored by this function

packages/shared/src/server/queries/clickhouse-sql/query-fragments.ts
```diff
@@ -152,13 +152,17 @@ export const buildScoresAggregationCTE = (
 interface EventsScoresAggregationParams {
   projectId: string;
   startTimeFrom?: string | null;
+  hasScoreAggregationFilters?: boolean;
```
sumerman: is it actually used?

packages/shared/src/server/repositories/experiments.ts
```diff
@@ -370,3 +377,606 @@ const getExperimentsFromEventsGeneric = async <T>(
...
+    }),
+  );
+
+  return {
+    where: combineConditions(compiledFiltersByExperiment, "OR"),
+    having: isBaselineEnforced
+      ? filteredCompExperimentIds.length > 0
+        ? {
+            query: `
+          countIf(e.experiment_id = {baseExperimentId: String}) > 0
```
sumerman, line 599: I have slight preference for letting the builder assemble filters and insert `AND`.

packages/shared/src/server/repositories/experiments.ts
sumerman: Personally, I would start with just a timestamp and only go for more involved CTE if you discover edge cases or you see it is slow. I don't expect either to be a problem in `EventsAggQueryBuilder` context.

sumerman (commented):
Just a couple of nits. Thank you!

web/src/__tests__/server/repositories/experiment-items-repository.servertest.ts
marliessophie: thx! removing.

marliessophie:
Thanks so much for the review, @sumerman! Will address feedback.

PR #12360, opened 2026-03-03
fix(storage): add buffered stream uploader with per-part retry for resilient S3 uploads

sumerman:
> [!IMPORTANT]
> Introduces `BufferedStreamUploader` for resilient S3 uploads with per-part retry, updates services, and adds tests.
> 
>   - **Behavior**:
>     - Adds `BufferedStreamUploader` class for S3 uploads with per-part retry logic in `BufferedStreamUploader.ts`.
>     - Introduces `S3ChunkedUploadStrategy` in `S3ChunkedUploadStrategy.ts` for handling multipart uploads.
>     - Updates `StorageService` in `StorageService.ts` to use `BufferedStreamUploader` for buffered uploads.
>     - Adds environment variables `LANGFUSE_S3_UPLOAD_MAX_PART_ATTEMPTS` and `LANGFUSE_S3_UPLOAD_MAX_CONCURRENT_PARTS` in `env.ts`.
>   - **Tests**:
>     - Adds `bufferedStreamUploader.test.ts` to test `BufferedStreamUploader` and `S3ChunkedUploadStrategy`.
>   - **Misc**:
>     - Updates `handleBatchExportJob.ts` and `handleBlobStorageIntegrationProjectJob.ts` to use buffered uploads.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f4a248ae41606268d32af17d2439f7cd180c72c7. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/shared/src/server/services/BufferedStreamUploader.ts
```diff
@@ -0,0 +1,199 @@
...
+  maxPartAttempts: number;
+  maxConcurrentParts: number;
+  key: string; // for logging
+}
+
+// Encapsulates "first write wins" for error capture. Prevents callers from
+// accidentally overwriting a prior error — a subtle bug that would be easy to
+// introduce when multiple parts fail concurrently, even though Node's
+// single-threaded model makes the raw check-then-set technically safe today,
+// it is entirely possible to just forget to check if an error is already captured.
```
Steffen911: We could also consider errors a list and then push to that from within `capture`. Then two utility/retrieval methods may make sense: `first` and `getAll`.

packages/shared/src/server/services/BufferedStreamUploader.ts
```diff
@@ -0,0 +1,199 @@
...
+        this.completedParts.push(result);
+        logger.debug(
+          `Uploaded part ${partNumber} (${(data.byteLength / 1024 / 1024).toFixed(1)} MiB) for key ${this.params.key}`,
+        );
+      })
+      .catch((err) => {
+        this.firstError.capture(err);
+      })
+      .finally(() => {
+        this.inFlightUploads.delete(promise);
```
Steffen911: Nice that this works! I thought, we'd need a map with some consistent identifier.

packages/shared/src/server/services/BufferedStreamUploader.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.inFlightUploads.add(promise);
+  }
+
+  private async uploadPartWithRetry(
+    data: Buffer,
+    partNumber: number,
+  ): Promise<CompletedPart> {
+    return backOff(() => this.params.strategy.uploadPart(data, partNumber), {
+      numOfAttempts: this.params.maxPartAttempts,
+      startingDelay: 1000,
```
Steffen911, line 192: Is this for the first execution or for every execution beyond the first?

Steffen911 (commented):
High-level: Why do we expect this to be more resilient than `Upload`? Because we retry more often on transient errors for individual parts?

sumerman:
>High-level: Why do we expect this to be more resilient than Upload? Because we retry more often on transient errors for individual parts?

AWS SDK doesn't retry upon socket hangup errors. Which is our primary source of woes. Here we have full control over types of errors we retry.

packages/shared/src/server/services/BufferedStreamUploader.ts
sumerman: I had exact same idea, but then realized it usually one part that fails and even if there's more than one I'd expect them to have similar causes. Don't know if I want to make this more complicated.

packages/shared/src/server/services/BufferedStreamUploader.ts
sumerman: Yeah, it works by object reference. But on a second thought it is far to easy to break by accident.

packages/shared/src/server/services/BufferedStreamUploader.ts
sumerman: It is kinda weird and counterintuitive but default behaviour is to skip initial delay https://github.com/coveooss/exponential-backoff/blob/master/src/delay/delay.factory.ts#L14

We rely on default value of 100ms everywhere else.

Steffen911 (approved):

PR #12378, opened 2026-03-04
feat(ui): support form to write to Pylon

jannikmaierhoefer:
## Summary

- Write support form submissions to **Pylon** alongside Plain. When a user submits the support form, an issue is also created in Pylon with the same title, body, severity/priority mapping, tags, and custom fields (URL, question type, topic, integration type, metadata).
- Add a new **Pylon client** (`pylonClient.ts`) with helpers for creating issues, uploading attachments, building HTML body/metadata strings, and mapping severity levels to Pylon priority values.
- Add a new **attachment upload endpoint** (`/api/support/upload-attachments`) that accepts base64-encoded files from the browser, validates size/count limits, uploads them to Pylon via its attachments API, and returns the resulting URLs.
- On the frontend, file uploads to Plain S3 and Pylon now run **in parallel**; Pylon upload failures are treated as best-effort (logged but non-blocking).
- **Scoped rollout**: Pylon write is currently limited to `langfuse.com` and `clickhouse.com` email domains only.
- Fix file drop handler to **append** files to the existing list instead of replacing them.
- New `PYLON_API_KEY` env var (optional) to enable the integration.


web/src/features/support-chat/trpc/plainRouter.ts
```diff
@@ -312,6 +321,48 @@ export const plainRouter = createTRPCRouter({
...
+              { slug: "question_type", value: input.messageType },
+              { slug: "topic", value: input.topic },
+              ...(input.integrationType
+                ? [{ slug: "integration_type", value: input.integrationType }]
+                : []),
+              { slug: "metadata", value: pylonMetadata },
+            ],
+          });
+        } catch (e) {
+          logger.error("Pylon issue creation failed (best-effort)", e);
```
marcklingen, line 423: please make sure that these trigger an incident in a way that this gets fixed, support tickets failing is critical

@maxdeichmann can you help here?

web/src/features/support-chat/trpc/plainRouter.ts
```diff
@@ -312,6 +321,48 @@ export const plainRouter = createTRPCRouter({
...
+            browserMetadata: input.browserMetadata,
+          });
+          await createPylonIssue({
+            apiKey: env.PYLON_API_KEY,
+            title: pylonTitle,
+            bodyHtml: pylonBodyHtml,
+            requesterEmail: email,
+            requesterName: fullName,
+            tags: ["Langfuse"],
+            priority: mapSeverityToPylonPriority(input.severity),
```
marcklingen, line 363: probbaly we need to refine this to also incorporate plan into the priority, can be done later

web/src/features/support-chat/trpc/plainRouter.ts
```diff
@@ -312,6 +321,48 @@ export const plainRouter = createTRPCRouter({
...
+      const emailDomain = email.split("@")[1]?.toLowerCase();
+      const pylonTestDomains = ["clickhouse.com", "langfuse.com"];
+      if (env.PYLON_API_KEY && pylonTestDomains.includes(emailDomain ?? "")) {
+        try {
+          const pylonTitle = `[${uniqueId}] ${input.messageType}: ${input.topic} • ${topLevel}/${subtype}`;
+          const pylonBodyHtml = buildPylonIssueBodyHtml({
+            message: input.message,
+            requesterEmail: email,
+          });
+          const pylonMetadata = buildPylonMetadataString({
```
marcklingen, line 340: plan is not added here even though the string builder allows for this

I'd assume that this is not sufficient as thereby this metadata is mixed into a single metadata field, will be hard to do prioritization based on the plan this way

There is no data written on the account, currently this is all issue metadata, I'd suggest we also add the langfuse plan to the account/company in pylon next to the issue

marcklingen: metadata builder allows for 11 fields, only 5 are used here; I'd make them all required and allow null as a value

marcklingen (commented):
see comments, looks like a solid starting point that is good to be merged to start exprimenting with it

jannikmaierhoefer:
Thanks for the review @marcklingen, I updated the PR to also properly add the langfuse plan and show a warning toast if issue creation failed. I'll talk to max how we can also trigger an incident for failed issue creation. 

I'll merge the PR now to test it in prod.

web/src/features/support-chat/trpc/plainRouter.ts
jannikmaierhoefer: - Plan now added to issue and company field
- Metadata now uses all 11 fields

PR #12386, opened 2026-03-04
fix(ui): preserve query params in ResizableImage custom loader

MatrixA:
## Summary

- Fix `customLoader` in `ResizableImage` to use `&` instead of `?` when the URL already contains query parameters
- This fixes inline image rendering for S3/MinIO presigned URLs in self-hosted deployments

## Problem

When Langfuse generates presigned S3 URLs for media (e.g., `http://minio:9000/langfuse/media/...?X-Amz-Algorithm=...`), the `customLoader` in `resizable-image.tsx` appends `?w=640&q=75`, creating a malformed URL with two `?` characters. This causes S3 signature verification to fail, so the `<Image>` component triggers `onError` and falls back to showing a broken image icon with a link.

The `LangfuseMediaView` component correctly fetches the presigned URL and passes `shouldValidateImageSource={false}`, but the `customLoader` still corrupts the URL.

## Fix

Check if the URL already contains `?` and use `&` as separator if so:

```diff
-  return src + (width && quality ? `?w=${width}&q=${quality || 75}` : "");
+  if (!width || !quality) return src;
+  const separator = src.includes("?") ? "&" : "?";
+  return `${src}${separator}w=${width}&q=${quality || 75}`;
```

## Test plan

- [x] Self-hosted deployment with MinIO: verify presigned URL images render inline in trace detail view
- [x] Regular HTTP/HTTPS image URLs still render correctly (no query params case)
- [x] Base64 data URI images unaffected (no query params)


nimarb:
hi @MatrixA , thank you for the contirbution ! could you sign the CLA so we can merge? thank you!

MatrixA:
> hi @MatrixA , thank you for the contirbution ! could you sign the CLA so we can merge? thank you!

Thanks! I’ve just signed the CLA. Looking forward to the merge.

PR #12420, opened 2026-03-05
feat(models): add gpt-5.4

nimarb:


worker/src/constants/default-model-prices.json
```diff
@@ -3656,6 +3656,64 @@
       }
     ]
   },
+  {
+    "id": "bee3c111-fe6f-4641-8775-73ea33b29fca",
+    "modelName": "gpt-5.4",
+    "matchPattern": "(?i)^(openai\/)?(gpt-5.4)$",
```
hassiebp, line 3662: We also need to match the pattern with the date suffix (gpt-5.4-2026-03-05), see the other gpt models (we typically use a separate entry as the catch all without the version might change price in the future)

PR #12430, opened 2026-03-06
fix(Storage): use GOOGLE_CLOUD_UNIVERSE_DOMAIN

dclinchamp-mgen:
## What does this PR do?
Allow Google Storage to use env var GOOGLE_CLOUD_UNIVERSE_DOMAIN with a fallback to standard googleapis.com

Fixes #12426 

## Type of change

## Mandatory Tasks

## Checklist


packages/shared/src/server/services/StorageService.ts
```diff
@@ -799,7 +799,9 @@ class GoogleCloudStorageService implements StorageService {
       }
     } else {
       // Use default authentication (environment variables or instance metadata)
-      this.storage = new Storage();
+      this.storage = new Storage({
+        universeDomain: env.GOOGLE_CLOUD_UNIVERSE_DOMAIN || "googleapis.com",
+      });
```
Steffen911, line 894: Are there cases where it would be desirable to have it in the other `Storage()` invocations, too? I think it could lead to subtle issues if it's only supported in one of the codepaths.

I would also recommend to add the `googleapis.com` fallback as a default in the env.ts file instead of coding it here.

fmontalbano-mgen:
Hi @Steffen911 , thank you for your feedbacks. The creator of this PR is no longer in our company and we do not have access to edit this PR anymore.

I took the liberty of creating a new PR with the originals mofications and your recommendations here : https://github.com/langfuse/langfuse/pull/13453

Sorry for the inconvenience and thank you again for your help.

Steffen911:
Replaced with https://github.com/langfuse/langfuse/pull/13453

PR #12435, opened 2026-03-06
chore(public-api): move legacy endpoints to separate namespace

hassiebp:


dahnny012:
@nimarb  is this safe to consume for self hosters who are not on v4? Would we have to do any migrations?

PR #12459, opened 2026-03-09
fix(seeder): improve database seeder idempotency and fix upsert syntax

chjnett:
## Overview
This PR fixes several issues in the `seed-postgres.ts` script that caused it to fail during second or subsequent runs. It improves the seeder's idempotency, allowing developers to populate their local environments consistently.

## Changes
- **Fixed `upsert` syntax**: Removed redundant `id` fields from the `where` clause in `Prompt` and `DatasetItem` upserts, which were causing Prisma validation errors.
- **Improved Idempotency**: Converted `create` calls to `upsert` for `TraceSession` and `JobExecution`.
- **Added Duplicate Check**: Added `skipDuplicates: true` to `llmSchema.createMany` to prevent unique constraint violations on re-seeding.
- **Fixed Typo**: Corrected compound identifier `id_validFrom` to `id_projectId_validFrom` for `DatasetItem` model.

## Why this is needed?
Currently, running `pnpm run db:seed:examples` multiple times fails with unique constraint errors or Prisma validation errors. This fix unblocks developers who want to re-seed their database or add data to an existing setup without manual cleanups.

## How to Verify
1. Run `docker-compose -f docker-compose.dev.yml up -d`
2. Run `pnpm run db:seed:examples` for the first time. (Success)
3. Run `pnpm run db:seed:examples` again. (Now succeeds instead of failing)


chjnett:
Quick update: I tested this fix against both empty and existing databases, and it works as expected. This should resolve the inconsistency issues we’ve seen in the seeder. Looking forward to your feedback!

PR #12477, opened 2026-03-09
chore(dx): allow vscode debugger break point setting

wochinge:
## What does this PR do?

Fixes setting break points from VSCode.

Fixes # (issue)

https://github.com/user-attachments/assets/95949028-64b0-468c-a81c-7650845c3187

## Type of change

## Mandatory Tasks


web/package.json
```diff
@@ -16,7 +16,7 @@
     "lint:fix": "dotenv -e ../.env -- eslint . --cache --cache-location .next/cache/eslint/ --fix",
     "clean": "rm -rf node_modules",
     "start": "dotenv -e ../.env -- sh -c 'NEXT_MANUAL_SIG_HANDLE=true next start'",
-    "test": "cross-env NODE_OPTIONS='--no-experimental-require-module' dotenv -e ../.env.test -e ../.env -- jest --verbose --runInBand --selectProjects server",
+    "test": "NODE_OPTIONS=\"${NODE_OPTIONS:-} --no-experimental-require-module\" dotenv -e ../.env.test -e ../.env -- jest --verbose --runInBand --selectProjects server",
```
wochinge, line 19: `cross-env` overwrites `NODE_OPTIONS` entirely, breaking VS Code debugger attachment. We need to replace it with shell variable expansion (`${NODE_OPTIONS:-}`) to preserve existing `NODE_OPTIONS` while appending our flags. 

This is consistent with other scripts in this repo that already use POSIX shell syntax (e.g., build, start). 

Windows developers using cmd.exe would need to configure script-shell to use sh/bash (same as they already need to do for the other commands)

nimarb:
see cross-env introduction here: https://github.com/langfuse/langfuse/commit/c6864ed8

--> seems good to go. it _will_ probbaly break windows test runner support, not that big of a deal as windows users can use WSL

PR #12502, opened 2026-03-10
chore: replace separate langchain google packages with unified one

coffee4tw:
## What does this PR do?

Migrate from `@langchain/google-vertexai `+ `@langchain/google-genai `to unified `@langchain/google`

- Replace `ChatVertexAI` and `ChatGoogleGenerativeAI` with `ChatGoogle` from `@langchain/google/node`
- Bump `@langchain/core` from `^1.1.17` to `^1.1.31` (required for `@langchain/google` namespace API)
- Adapt thinking block detection from Set-based type matching to predicate-based (`block.thought === true`), matching the new package's output format
- Map `baseUrl` → `endpoint` for GoogleAIStudio via `baseUrlToEndpoint()` helper
- Add thinking model tests for GoogleAIStudio (tool calling + completion with reasoning)
- No changes to enum values, config schemas, or UI — only the LangChain runtime layer is affected

**NOTE:** the baseURL -> endpoint transformation means that some llmConnections will break, if the user has defined something like `http://proxy.internal/v1`it will be converted to `https://proxy.internal/v1/v1beta/models/...`. Any baseURLs with `https` and without duplicate paths `v1beta/models/...`will continue to work. If either are present, they will have to adjust their llmConnection settings.

## Type of change

## Mandatory Tasks


nimarb:
@coffee4tw FYI same branch but with running LLM tests here: https://github.com/langfuse/langfuse/pull/12505

Issue #11942, opened 2026-02-09
bug: filtering on metadata in dashboards returns nothing


ronancummins:
### Describe the bug

In "Dashboards", when trying to filter by metadata (adding a key and some value), no data is returned in the charts. Looking the the specific traces, they do include the specific metadata (keys and values) that should return those results. 

### Steps to reproduce

- have traces with some custom metadata viewable on the trace
- create a dashboard that shows some information about those traces
- add a filter that should filter a proportion of those traces

The filter on the metadata returns nothing.

### Langfuse Cloud or self-hosted?

Langfuse Cloud

### Are you interested in contributing a fix for this bug?

No

ronancummins:
[@dosu](https://go.dosu.dev/dosubot). The keys that I want to filter on are already top-level metadata keys. The keys appear in the traces that I can see in langfuse at the top-level. The following is what I can see in the traces view in langfuse cloud right under the Metadata panel. I can filter my dashboard on "sessionName" or "isTestUser". 

`
{
    "sessionName": "intro",
    "isTestUser": false,
    },
    "resourceAttributes": {
        "service.name": "skill-bot",
        "host.name": "skill-bot-65478c5668-fnldd",
        "host.arch": "amd64",
        "process.pid": 1,
        "process.executable.name": "node",
        "process.executable.path": "/usr/local/bin/node",
        "process.command_args": [
            "/usr/local/bin/node",
            "/usr/src/app/src/index.js"
        ],
        "process.runtime.version": "20.11.1",
        "process.runtime.name": "nodejs",
        "process.runtime.description": "Node.js",
        "process.command": "/usr/src/app/src/index.js",
        "process.owner": "root",
        "telemetry.sdk.language": "nodejs",
        "telemetry.sdk.name": "opentelemetry",
        "telemetry.sdk.version": "2.2.0"
    },
    "scope": {
        "name": "langfuse-sdk",
        "version": "4.5.0",
        "attributes": {}
    }
}`

ronancummins:
Any update on this?

Steffen911:
@ronancummins Could you share a sample of your chart configuration? And are you referring to a custom dashboard or the root dashboard in Langfuse Home?
Any screenshots of your definition would help here.

ronancummins:
@Steffen911 
Thanks. It's a custom dashboard. The following is an example of a trace with a field at the top level of the json:

[...]

PR #12528, opened 2026-03-11
fix(queryBuilder): enhance metadata filtering for observations to support trace-level keys

saurav470:
Fix: Dashboard metadata filter returns no data (#11942)
Problem
Filtering by metadata (e.g. sessionName = "00_initial") in Dashboards returned no data in charts, even when traces clearly had that metadata in the trace detail view.

Root cause
Charts like "P 50 Latency by task" and "Latency by task" use the observations view. The dashboard "Metadata" filter was applied to observation metadata (observations.metadata). Trace-level fields such as sessionName and isTestUser live on trace metadata, so the filter was evaluated on the wrong table and matched no rows.

Solution
web/src/features/query/server/queryBuilder.ts
In mapFilters(), when the filter column is metadata and the view has a traces relation (observations, scores-numeric, scores-categorical), the metadata filter is now applied on the traces table (traces.metadata[key]) instead of the base table. The existing JOIN to traces is reused, so no extra JOIN is added.

Traces view: unchanged; still filters on traces.metadata.

Observations view: metadata filter now uses traces.metadata, so trace-level keys (e.g. sessionName, isTestUser) correctly restrict observation-level charts.

Scores views: same behavior; metadata filter uses traces.metadata.


nimarb:
hi @saurav470 , thank you for the contribution! could oyu sign the CLA? thank you !

saurav470:
Hi @nimarb, I’ve signed the CLA. Thanks!

nimarb:
thanks! can you check the failed tests @saurav470 ? thank you!

saurav470:
Hi @nimarb, checking

saurav470:
Hi @nimarb

The workflows seem to be awaiting maintainer approval, so CI checks haven’t started yet.
Could you please approve the workflow run so the tests can execute?

Thanks again!

nimarb:
hi @saurav470 , thank you for this change. sorry about the late message here, there are 2 more things to consider:
1. is it clear to the user which metadata is queried here? we want to make it explicit in the UI idealely
2. can we retain the ability to query score metadata easily? if so, how? there might be people relying on this currently / want to query for score metadata going forward.

saurav470:
Thanks for the review, @nimarb.

You’re right on both points:

Clarity in the UI – Right now the filter is just labeled “Metadata,” so it’s not obvious whether we’re filtering trace, observation, or score metadata. I’d be happy to add clearer labels like “Trace metadata” vs “Score metadata” in the filter UI.

Retaining score metadata – The current change routes the generic “Metadata” filter to trace metadata for observation/score views, which fixes the dashboard bug but removes the ability to filter by score-level metadata. We could:

Add a separate “Score metadata” filter option for score views (and “Observation metadata” for observation views),
 or
Keep both: “Trace metadata” for the dashboard use case and “Score metadata” for score-level filtering.

I’m open to either implementing these changes in this PR or handling them in a follow-up. Let me know what you prefer.

brandonwatts:
@saurav470 @nimarb Any update on this? The use case for us is we have trace level metadata that I need to filter out of a 'Scores Numeric' widget. It works until I try to set up the filter then the dashboard goes blank. The trace has a top level metadata flag that I would like to use. I dont personally love the Score Metadata/Trace Metadata split as its not very intuitive but as long as there is a way to do it i'm fine with whatever. From a users perspective I would have just assumed that if a score was ran of a trace that traces metadata would be present on the score by default. I wasn't even aware scores could have their own metadata thats interesting. At least for LLM as s a judge though it doesn't make alot of sense because there is no way in the UI to instruct it to carry over certain metadata properties. I can inject them into the LLM as a judge prompt, I can use them as a filter but there is no way I can find to say "and also carry over these metadata attributes onto the score" so then all the dashboards related to scoring gets nerfed a ton.

nimarb:
@claude review

PR #12546, opened 2026-03-12
fix: avoid FINAL on traces table in traces.metrics endpoint

6matt:
Pass orderBy to getTracesTableMetrics so ClickHouse uses LIMIT 1 BY instead of FINAL for deduplication, matching traces.all behavior.

## What does this PR do?

`getTracesTableGeneric` uses `FINAL` on the traces ReplacingMergeTree when no `orderBy` is provided. The `traces.metrics` call was the only caller not passing `orderBy`, forcing ClickHouse to merge all table parts at query time.

Fixes #12545

## Type of change

## Mandatory Tasks

## Checklist


Steffen911 (approved):
Nice catch! Looking good 🙌

PR #12547, opened 2026-03-12
feat: specify UTC timezone for DateTime64 fields in clickhouse migrat…

formorter:
## What does this PR do?

This PR adds explicit UTC timezone specification to DateTime64 fields in ClickHouse tables.

The issue: when querying a ClickHouse cluster with a non-UTC timezone configuration, traces and observations often return no results due to timezone conversion errors in query filters.

This partly resolves the issue outlined in: https://github.com/orgs/langfuse/discussions/5046

Tested in a self-hosted Langfuse instance with an existing ClickHouse cluster.

## Type of change

## Mandatory Tasks

## Checklist


sumerman:
Thanks for the contribution! Unfortunately, we can't merge this as-is — editing already-applied migrations won't have any effect on existing deployments, since `golang-migrate` only tracks the migration version number and won't re-run previously applied migrations. For fresh installs, the initial table creation would differ from what existing users have, which could lead to subtle inconsistencies that are hard to debug down the line.

formorter:
Thanks for the reply @sumerman! So, if I add an extra boolean environment variable(something like LANGFUSE_UTC_MIGRATION) and create a migration that explicitly sets the timezone to UTC and they runs in startup, would that be acceptable?

I have an issue with timezone in my ClickHouse and want to make some changes without forking repository:(

sumerman:
@formorter Thanks for the suggestion — directionally, I think this sounds better than modifying a bunch of existing migrations.

A couple of questions:

1. How would an environment variable actually control whether a migration runs? I'm not fully following the mechanism you have in mind.

2. What's your ClickHouse setup that requires Langfuse itself to apply this migration, rather than configuring ClickHouse ahead of time (e.g., setting the cluster config before Langfuse is deployed)?

formorter:
> 1. How would an environment variable actually control whether a migration runs? I'm not fully following the mechanism you have in mind.

I want to make a new environment variable here:
https://github.com/langfuse/langfuse/blob/main/packages/shared/clickhouse/scripts/up.sh#L42

If this variable is enabled, I plan to change the migration source here(i will create new path clickhouse/migrations/clustered_utc):
https://github.com/langfuse/langfuse/blob/main/packages/shared/clickhouse/scripts/up.sh#L71

> 2. What's your ClickHouse setup that requires Langfuse itself to apply this migration, rather than configuring 

In my environment, we use a shared ClickHouse cluster that's not exclusive to Langfuse. It has a local timezone configured, and I’m unable to change it due to dependencies from other services.

sumerman:
> > 1. How would an environment variable actually control whether a migration runs? I'm not fully following the mechanism you have in mind.
> 
> I want to make a new environment variable here: https://github.com/langfuse/langfuse/blob/main/packages/shared/clickhouse/scripts/up.sh#L42
> 
> If this variable is enabled, I plan to change the migration source here(i will create new path clickhouse/migrations/clustered_utc): https://github.com/langfuse/langfuse/blob/main/packages/shared/clickhouse/scripts/up.sh#L71
> 
> > 2. What's your ClickHouse setup that requires Langfuse itself to apply this migration, rather than configuring
> 
> In my environment, we use a shared ClickHouse cluster that's not exclusive to Langfuse. It has a local timezone configured, and I’m unable to change it due to dependencies from other services.

Wouldn't that mean we'll have to maintain 3 sets of migrations?

formorter:
yes, it is. But if its a problem, I can try to come up with something else

formorter:
For example, in a bash script i actually can change all fields that contain DateTime64(3) to DateTime64(3, 'UTC') in migrations, if some env is true 

In up.sh, for example i can add this:
```
if [ "$LANGFUSE_UTC" = "true" ]; then
    find "$MIGRATIONS_DIR" -type f -exec sed -i 's/DateTime(3)/DateTime(3, '\''UTC'\'')/g' {} \;
fi
```

sumerman:
> For example, in a bash script i actually can change all fields that contain DateTime64(3) to DateTime64(3, 'UTC') in migrations, if some env is true
> 
> In up.sh, for example i can add this:
> 
> ```
> if [ "$LANGFUSE_UTC" = "true" ]; then
>     find "$MIGRATIONS_DIR" -type f -exec sed -i 's/DateTime(3)/DateTime(3, '\''UTC'\'')/g' {} \;
> fi
> ```

That definitely could work, but that is something you could already do without forking Langfuse, can't you?

formorter:
> > For example, in a bash script i actually can change all fields that contain DateTime64(3) to DateTime64(3, 'UTC') in migrations, if some env is true
> > In up.sh, for example i can add this:
> > ```
> > if [ "$LANGFUSE_UTC" = "true" ]; then
> >     find "$MIGRATIONS_DIR" -type f -exec sed -i 's/DateTime(3)/DateTime(3, '\''UTC'\'')/g' {} \;
> > fi
> > ```
> 
> That definitely could work, but that is something you could already do without forking Langfuse, can't you?

Yes, I can rebuild the docker image with the new up.sh if that's acceptable for the user experience, then l will do just that. But i think it would be great to have this functionality in the original image.

I could say that I'm not the only one facing this issue if looks in https://github.com/orgs/langfuse/discussions/5046
And i think it will be good to documentate it somewhere

sumerman:
@formorter you can run:

```
  ALTER TABLE traces
      MODIFY COLUMN `timestamp` DateTime64(3, 'UTC'),
      MODIFY COLUMN `created_at` DateTime64(3, 'UTC') DEFAULT now('UTC'),
      MODIFY COLUMN `updated_at` DateTime64(3, 'UTC') DEFAULT now('UTC'),
      MODIFY COLUMN `event_ts` DateTime64(3, 'UTC');
```
etc 

after you provision Langfuse, but before you ingest data. It might even work on top of already ingested data since in ClickHouse timezone is nothing more than a metadata annotation.

formorter:
Yes, actually i can, but i dont want to do a manual actions after langfuse startup

Issue #11307, opened 2025-12-24
bug: Formatted tab does not properly show pydantic-ai LLM generations


konen-theori:
### Describe the bug

Hi, I use pydantic-ai with langfuse to call and track LLMs in my project. The conversation view using `Formatted` tab is very useful, but it sometimes doesn't show the conversation properly. 

Here are the steps to reproduce:

```
import asyncio
from pydantic import BaseModel
from pydantic_ai import Agent, ModelRequest, RunContext, UserPromptPart
from pydantic_ai.models import ModelRequestParameters
from pydantic_ai.output import OutputObjectDefinition
from pydantic_ai.direct import model_request
from pydantic_ai.models.google import GoogleModel
from pydantic_ai.providers.google import GoogleProvider
from langfuse import get_client

# For loading API keys, use your own method
from dotenv import load_dotenv
load_dotenv()

langfuse_client = get_client()
langfuse_client.auth_check()

Agent.instrument_all()

class MyResponse(BaseModel):
    answer: str
    rationale: str

async def naive_request():
    response = await model_request(
        # "google-gla:gemini-2.5-flash",
        "openai-responses:o4-mini",
        [ModelRequest(parts=[UserPromptPart(content="""what is 1+1?
Output format:
```json
{
    "answer": your answer,
    "rationale": "explanation for answer"
}
```""")])],
        # model_request_parameters=ModelRequestParameters(
        #     output_mode="native",
        #     output_object=OutputObjectDefinition(
        #         json_schema=MyResponse.model_json_schema()
        #     )
        # ),
        instrument=True,
    )

asyncio.run(naive_request())
```

This looks fine on the langfuse dashboard. [[link]](https://us.cloud.langfuse.com/project/cmjjlsmpf01q5ad083m5ud2p5/traces/956feedf0fb60a252d02369311b762ee?timestamp=2025-12-24T06:01:00.786Z) However,

- if you uncomment the `model_request_parameters` then the assistant response is shown as a broken JSON table view. [[link]](https://us.cloud.langfuse.com/project/cmjjlsmpf01q5ad083m5ud2p5/traces/b44f5e2b64f199efb429cf4c305459d9?timestamp=2025-12-24T06:01:27.917Z)
[...]

PR #12552, opened 2026-03-12
chore(ChatMLAdapter): Add test cases against seed spans

ouvreboite:
Changing the current adaption logic is brittle as several adapters relies on each others (cf the various "exclusions" cases). Having stronger E2E test for the adapter would make future changes safer. 

## Edit 🆕 

This PR now focuses on implementing an E2E test suite to assert the chatML output of real-world traces/observations.
We no longer assert which adapter is picked, only the chatML result matters.
This provides both a more concrete test output and more flexibility to refactor the adapter logic in the future.

## Initial message: 
For example, by having a bank of real-world spans for each framework/provider and ensuring the correct adapter is used.

**🚧This is a WIP and I have a few questions:**
1. I'm assuming that [seeds](https://github.com/langfuse/langfuse/tree/main/packages/shared/scripts/seeder/utils/framework-traces)  are raw dump of traces/spans sent from their corresponding frameworks. And that I can extract each `observations` items and process it which the adapters. **Does that makes sense?**
2. I've defined a naive expected classification based on the name of the seeds:

| Seed | Expected adapter | Result | Details | 
|---|----|---|---|
| [agno-2025-06-11.js](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/agno-2025-06-11.json#L23) | generic  | ✅ | |
| [autogen-2025-06-06.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/autogen-2025-06-06.json#L26) | generic | ✅ | |
| [beeai-2025-08-01.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/beeai-2025-08-01.json#L68) | generic | ✅ | |
| [google-adk-2025-08-28.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/google-adk-2025-08-28.json#L68) | gemini | ❌| [Observation 2](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/google-adk-2025-08-28.json#L893) is detected as `generic` |
| [google-gemini-2025-08-01.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/google-gemini-2025-08-01.json#L68)  | gemini | ❌| [Observation 0](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/google-gemini-2025-08-01.json#L70) is detected as `generic` |
| [google-vertex-tools-langchain-2025-09-19.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/google-vertex-tools-langchain-2025-09-19.json#L16) | langgraph | ✅ | |
| [koog-2025-08-26.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/koog-2025-08-26.json#L23) | generic | ✅ | |
| [langgraph-2025-08-22.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/langgraph-2025-08-22.json#L68) | langgraph | ✅ | |
| [llamaindex-2025-06-05.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/llamaindex-2025-06-05.json#L26) | openai | ✅ | despite being called "llama", this has nothing to do with the llama models and uses gpt |
| [microsoft-agent-2025-10-14.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/microsoft-agent-2025-10-14.json#L68) | microsoft-agent | ✅ | |
| [openai-agents-2025-09-30.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/openai-agents-2025-09-30.json#L68) | openai | ✅ | |
| [openai-assistants-2024-05-29.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/openai-assistants-2024-05-29.json#L22) | openai | ❌ | [Observation 0](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/openai-assistants-2024-05-29.json#L24) is detected as `generic` |
| [pydantic-ai-tools-2025-12-04.json](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/pydantic-ai-tools-2025-12-04.json#L26) | pydantic-ai | ❌ | [Observation 0](https://github.com/langfuse/langfuse/blob/main/packages/shared/scripts/seeder/utils/framework-traces/pydantic-ai-tools-2025-12-04.json#L28) is detected as `generic` |
| | aisdk | ⚠️ | no seed seems to correspond to Vercel's AI sdk |
| | semantic-kernel | ⚠️ | no seed seems to correspond to Semantic Kernel |
3. Is it expected that some observations from google-adk, gemini, openai-assistant and pydantic-ai are classified as `generic` ?
4. There are missing seed for aisdk and semantic-kernal (given that SK has been somehow replaced by agent-framework, that may not matter much)

### What does this PR do?

Add test cases that read the seed spans and assert which chatML adapter should process them.

Related to https://github.com/langfuse/langfuse/issues/11307 and https://github.com/langfuse/langfuse/issues/12550

### Type of change

### Mandatory Tasks

### Checklist


nimarb:
hi @ouvreboite , thank you for this PR. your assumption is correct here. the `framework-traces/...` traces stem from real integrations and/are used to test the different adapters. adding a real test here also makes sense.

are you looking in fixing the adapters themselves, too?
the exclusion is a bit tricky here, because many data formats are close to one another. do you have a significantly better solution in mind here?

ouvreboite:
hi @nimarb, ultimately I would like to fix the adapters themselves, and I've toyed with a few ideas to restructure the selection logic _(ex: doing two passes (scope.name then shape detection), reordering the adapter to match the generic framework first, ...)_. But, I quickly stopped as I feared introducing side effects / breaking other adapters.

➡️ Hence this PR: I would like to reinfore the testing strategy first, to then more safely iterate on the adapter selection logic (me or anybody else).

If this logic sound good to you, I could proceed by:

- **Adding more examples**
  - From the various "framework integration examples" on langfuse's own doc (ex: [hackstack](https://langfuse.com/integrations/frameworks/haystack#step-5-view-traces-in-langfuse))?
  - And including the problematic spans from [11307](https://github.com/langfuse/langfuse/issues/11307) and [12550](https://github.com/langfuse/langfuse/issues/12550), (if [konen-theori](https://github.com/konen-theori) agree, of course)
  - Any other example would be welcomed :) 
  - I would duplicate/move those test files in a dedicated folder, to not mix them with the seed files.
- **Grandfathering the current failing tests behavior** (ex: I would assert that spans from google-adk-2025-08-28.json should be matched by the gemini **or** generic adapters). And then updating those test cases when introducing fixes for [11307](https://github.com/langfuse/langfuse/issues/11307) and [12550](https://github.com/langfuse/langfuse/issues/12550) in later PRs.

Note: At the moment my test focus only on the asserting the adapter id. I suppose I could also assert the chatML conversion result itself, but I think that could be done later to keep this PR small. And once we have fixed the few selection bugs, to avoid unnecessary noise.

nimarb:
hi @ouvreboite thank you !

approach makes sense! ideally, to merge this PR, we could already fix the adapter selection for the tests where they fail. the google ones _should_ be rendered by the gemini adapter. if too much, feel free to do the grandfathering to gemini or generic.

i would like to keep the `generic` adapter last, because per default, it would match too much otherwise.

including more (up to date) examples is good!

the adapters themselves do have a test already which is where i would do the chatml assertion.

ouvreboite:
I'll work on that and keep you posted.

ouvreboite:
@nimarb when downloading trace data (json file) from the UI, it seems some of the observations/span data is missing.

For example, if I get the vercel aisdk demo [trace](https://cloud.langfuse.com/project/cloramnkj0002jz088vzn1ja4/traces/df41e597d0a85e0d7a6ae8ebfaa70aa0) (from the [doc](https://langfuse.com/integrations/frameworks/vercel-ai-sdk#5-see-traces-in-langfuse)) and click on `Download trace as JSON`, none of the observations in the file have input/output. Their metadata is also a simple empty object. 

The tool definition (visible in the UI) is also missing.

--- edit ---

Looking at the code and network, it seems the download button basically repackage the data coming from [/api/trpc/traces.byIdWithObservationsAndScores](https://cloud.langfuse.com/api/trpc/traces.byIdWithObservationsAndScores?input={"json":{"traceId":"df41e597d0a85e0d7a6ae8ebfaa70aa0","projectId":"cloramnkj0002jz088vzn1ja4"}}) which I suppose feeds the timeline panel so do not get full observation details. And the `includeIO` is set to false in this endpoint internally. (cf [traces.ts](https://github.com/langfuse/langfuse/blob/main/web/src/server/api/routers/traces.ts#L362)). So no input/output :) 

I can still manually export the full observation details by calling [/api/trpc/observations.byId](https://cloud.langfuse.com/api/trpc/observations.byId?input={"json":{"observationId":"91a17a2274bfd23f","traceId":"df41e597d0a85e0d7a6ae8ebfaa70aa0","projectId":"cloramnkj0002jz088vzn1ja4"}}) for each observation.

nimarb:
@ouvreboite sorry about this annoying bit - can you check if you can get the full trace here? : 
2026-03-12_18 27 39-Google Chrome@2x
thanks!

worker/src/__tests__/chatml/integration.test.ts
```diff
@@ -535,18 +536,72 @@ describe("ChatML Integration", () => {
 type FileTestCase = {
   file: string;
   expected: string;
+  obsOverride: string[];
 };
 
 describe("ChatML adapter selection tests", () => {
   it.each([
-    { file: "agno-2025-06-11.json", expected: "generic" },
+    {
```
ouvreboite: At the moment there is a lot of traces that contains conflicting observations (i.e. all the trace's observations are not matched by the same adapter).

In order to having passing test, I've excluded those observations and tried to document what kind of edge case they represent.

The langgraph adapter is often involves, as its structural matching is quite broad. For example, this can match a lot of things:

```typescript
const LangGraphWrappedSchema = z.looseObject({
  messages: z.array(z.any()),
});
```

Traces coming for the openinference packages often contains troublesome observations, because they dump the raw method arguments of the client being called. So if you call `runner.run_async(new_message: "foo bar")` in your agent, openinference put `new_message: "foo bar"` as the top level observation input, regardless of what format is sent to the LLM.

ouvreboite:
@nimarb after thinking about the problem this weekend, I feel like asserting "which adapter" is selected is a bit of a dead end. Especially because a given framework can have observations with very different format, depending on when the observation is emitted in the pipeline and what is the underlying LLM provider (or LLM proxy). In addition, a lot of framework have overlapping formats, which means a mis-processed input/output can sometime result in a semi-valid result.

Ultimately, what we want is a **better and more stable mapping**, i.e. better `ChatMlMessageSchema` result for a given observation.

So I've updated the PR to focus on that: 
- A bunch of `framework-traces/*.trace.json` files which contains observations from one of the seed/example trace
- Corresponding `framework-traces/*.chatml.json` files that contains the exact expected chatml conversion result for each observation (Those are grand-fathered result, to get started)
- A test that load and map each observation and assert the deep equality against the expected chatml result

I think that makes it way easier to understand what a given change to the mapping logic yield.
As an example, this commit https://github.com/langfuse/langfuse/pull/12552/changes/05c1133ab584acdb7f9b13cae712a4e282773d4a fixes the pydantic mapping and we can clearly see the impact on the produced chatML: the pydantic observations now have their tools correctly mapped ✅.

That means the workflow to reliably fix mapping bug becomes:
1. Add the problematic trace+observations file and init the corresponding *.chatml.json file using the current behavior (there is an "update" toggle on the test to ease that)
2. Do your fix + update the expected file
3. The reviewers can easily assess what the fix actually changes, and that unrelated traces are not impacted
4. Once your fix is merged, it will be implicitly part of the test suite forever.

PS: I don't know why, but even when rebasing, many files on the /web folder seems improperly formatted and fails my precommit hook. So I've applied pnpm run format in a dedicated commit. That's why so many files are changed.

ouvreboite:
@nimarb I see that the codespell action is failing due to some typos in the seed traces themselves.
To you want me to update the `.codespellrc` file to ignore the `*.trace.json` and` *.chatml.json` files, or to fix the typo in the trace file itself?

nimarb:
@ouvreboite , ideally we leave them in for now, i just fixed the failing spelling :)!

PR #12562, opened 2026-03-12
feat(tracing-table): auto-suggest nested metadata

nimarb:


packages/shared/src/server/utils/metadata_conversion.ts
```diff
@@ -13,3 +14,33 @@ export function parseMetadataCHRecordToDomain(
...
+/**
+ * Reverses flattenJsonToPathArrays: groups dot-notation keys by their first
+ * segment back into nested objects. Only unflattens one level — the first dot
+ * separates parent from child key (child may itself contain dots).
+ *
+ * Example: { "resourceAttributes.service.name": "svc" } → { resourceAttributes: { "service.name": "svc" } }
+ *
+ * Only used for events table metadata which uses flattenJsonToPathArrays on write.
+ */
+export function unflattenMetadata(flat: MetadataDomain): MetadataDomain {
```
sumerman: will behave wrong if anyone puts . into their metadata field name

packages/shared/src/server/repositories/events.ts
```diff
@@ -2339,6 +2342,53 @@ export const getEventsGroupedByCalledToolName = async (
   return res;
 };
 
+/**
+ * Get grouped available metadata keys from events table
+ * Used for filter options
+ */
+export const getEventsGroupedByMetadataKey = async (
```
sumerman, line 2346: This is a rather reasonable way to get what you want but still very expensive on long time horizons. E.g. 30days for a big customer could be running over ten seconds. We should cache these or something. Or consider forcing only recent values.

packages/shared/src/server/utils/metadata_conversion.ts
nimarb: @sumerman yes, imo the least "bad" way of doing this though - wdyt? i don't want to start escaping `.` in strings

sumerman: why do we even have to do that?

PR #12583, opened 2026-03-13
chore: Shard additional worker queues

maxdeichmann:


packages/shared/src/server/otel/OtelIngestionProcessor.ts
```diff
@@ -189,7 +189,9 @@ export class OtelIngestionProcessor {
     ).uploadJson(fileKey, resourceSpans as Record<string, unknown>[]);
 
     // Add queue job
-    const queue = OtelIngestionQueue.getInstance({});
+    const queue = OtelIngestionQueue.getInstance({
+      shardingKey: `${this.projectId}-${fileKey}`,
+    });
```
maxdeichmann, line 194: @Steffen911 why was this not implemented yet?

Steffen911: @maxdeichmann Three things: time constraints, because it causes monitoring/operational overhead to have multiple shards, and it wasn't necessary so far.

packages/shared/src/server/redis/llmAsJudgeExecutionQueue.ts
```diff
@@ -6,46 +6,87 @@ import {
...
-                type: "exponential",
-                delay: 1000,
-              },
+    const name = `${QueueName.LLMAsJudgeExecution}${shardIndex > 0 ? `-${shardIndex}` : ""}`;
+    const queueInstance = newRedis
+      ? new Queue<TQueueJobTypes[QueueName.LLMAsJudgeExecution]>(name, {
+          connection: newRedis,
+          prefix: getQueuePrefix(name),
+          defaultJobOptions: {
+            removeOnComplete: 10_000, // important for job deduplication
```
Steffen911: Shouldn't be necessary here as we expect exactly one job per span and don't have cancellation anymore. Would vote to remove as this reduces unnecessary data in redis.

web/src/pages/api/admin/evals.ts
```diff
@@ -1,98 +0,0 @@
-import { type NextApiRequest, type NextApiResponse } from "next";
```
Steffen911, line 1: Is this drop intentional?

web/src/pages/api/admin/bullmq/index.ts
```diff
@@ -1,6 +1,10 @@
 import { type NextApiRequest, type NextApiResponse } from "next";
 import { z } from "zod/v4";
 import {
+  EvalExecutionQueue,
```
Steffen911, line 4: Could you also update the `ingestion-replay.ts` bullmq endpoint? I think it still uses the native getInstance calls.

web/src/pages/api/admin/evals.ts
maxdeichmann: Yes, it is.

PR #12618, opened 2026-03-16
fix(otel-ingestion): restrict usage_details to only generation and embedding observation types

hassiebp:


hassiebp:
@claude review

PR #12622, opened 2026-03-16
fix: use created_at for incremental exports to avoid skipping late-arriving data

jaydeerossi:
## What does this PR do?

> PR title must follow Conventional Commits, for example `feat(web): add trace filters` or `fix: handle empty dataset names`.

fixes: #12439

Alter queries for blob export to filter on created_at instead of other timestamp columns in traces, observations, and scores.

Filtering by the business timestamp of these objects can result in data missing from the exports, due to the possibility of late-arriving data, i.e. where `timestamp < created_at`.

Fixes # (issue)

## Type of change

## Mandatory Tasks


Steffen911:
Will close this until the discussion in the corresponding issue is resolved.

PR #12695, opened 2026-03-20
feat(security): implement nonce-based CSP to remove unsafe-inline

Steffen911:


Steffen911:
Currently, it seems that all nonce values are empty and the CSP header is not injected anymore. Proposal prompt for fixing:
```
We're working on PR #12695 in the langfuse/langfuse repo (branch: feat/nonce-based-csp).
Title: "feat(security): implement nonce-based CSP to remove unsafe-inline".
The PR removes 'unsafe-inline' from the CSP in web/next.config.mjs and replaces it
with a per-request nonce. It is currently broken on staging: SSO sign-in redirects
fail, and the root cause is upstream of CSP — the nonce mechanism itself isn't
wiring through.

Concrete symptom on a fresh load of /auth/sign-in on staging:

1. The HTML response contains every script and link tag with `nonce=""` (empty
   string). Example:
     <meta property="csp-nonce" content=""/>
     <script src="/_next/static/chunks/xyz.js" nonce="" defer=""></script>
   The __NEXT_DATA__ JSON also embeds `"nonce": ""`.
2. The HTTP response has NO `Content-Security-Policy` header at all.

Both symptoms together mean the middleware that's supposed to (a) generate the
nonce, (b) set the CSP response header, and (c) propagate the nonce into the
request headers for getServerSideProps / _document.tsx to read, is not executing
for this route. The page-side render code clearly expects an `x-nonce` (or similar)
request header and is falling back to "" when it's missing.

This is a Next.js Pages Router app. Source layout uses web/src/, so middleware
must live at web/src/middleware.ts (NOT web/middleware.ts). The pages are under
web/src/pages/. next.config.mjs uses `output: 'standalone'`.

Please:

1. Check out the branch and review the PR's diff. Specifically look at:
   - web/src/middleware.ts (or web/middleware.ts — confirm which) — does it exist,
     where is it, what's its `config.matcher`, what runtime does it use, what
     headers does it set on the request vs. the response, and does it return
     NextResponse.next({ request: { headers: requestHeaders } })?
   - web/src/pages/_document.tsx — does getInitialProps read the nonce off
     ctx.req.headers and pass it as a prop? Is the prop name consistent with
     what the rendered <Head nonce={...}> expects?
   - web/src/pages/_app.tsx and any per-page getServerSideProps that surface a
     `nonce` prop — confirm they read from req.headers using the same key the
     middleware sets.
   - web/next.config.mjs — verify whether the static `Content-Security-Policy`
     entry in headers() was removed (it should have been, since middleware now
     owns it) AND that nothing else is conflicting.

2. Diagnose, in this order:
   a. Path: is the middleware file at web/src/middleware.ts? If it's at
      web/middleware.ts in a src-layout project, Next won't load it.
   b. Matcher: does config.matcher actually match /auth/sign-in? Common bug is
      a regex like /((?!auth|api|_next).*)/ that accidentally excludes auth
      pages. The intended pattern is usually
      /((?!api|_next/static|_next/image|favicon.ico).*)/.
   c. Edge runtime: middleware runs on the Edge runtime by default. If it uses
      Node-only APIs (crypto.randomBytes, Buffer in certain ways, fs, etc.) it
      will throw and Next serves the page without it. Use crypto.randomUUID()
      or crypto.getRandomValues(). Check for any Node imports.
   d. Header propagation: confirm middleware does
        const requestHeaders = new Headers(request.headers)
        requestHeaders.set('x-nonce', nonce)
        const response = NextResponse.next({ request: { headers: requestHeaders } })
        response.headers.set('Content-Security-Policy', cspValue)
      Returning NextResponse.next() without the { request: { headers } } option
      means downstream getServerSideProps/_document does NOT see x-nonce.
   e. Header key consistency: middleware's set key must exactly match what
      _document.tsx / getServerSideProps reads. Lowercase header names in fetch
      land — e.g. setting 'x-nonce' and reading req.headers['x-nonce'] is fine,
      but mismatches like 'X-Nonce' vs 'x-nonce' on Node http req work, while
      Edge Headers are case-insensitive — confirm this isn't a typo issue.
   f. Standalone build: `output: 'standalone'` in next.config.mjs must include
      middleware in the Docker image. Confirm .next/server/middleware.js exists
      in the build output.

3. Fix whatever is actually broken. Most likely fix is one of: move the
   middleware file to web/src/middleware.ts, correct the matcher, swap a
   Node-only crypto call for an Edge-safe one, or add the missing
   `{ request: { headers: requestHeaders } }` to NextResponse.next().

4. Verify:
   - Run a local dev server (pnpm run dev:web) and curl /auth/sign-in. The
     response MUST include a Content-Security-Policy header containing
     `'nonce-...'` with a real value (not empty).
   - The HTML body's <meta property="csp-nonce"> and every <script nonce="..."> /
     <link nonce="..."> must contain that same nonce string, non-empty.
   - The nonce should differ across two consecutive requests (it's per-request).
   - Run pnpm --filter web run lint and pnpm --filter web run typecheck — both
     must pass.
   - Do not re-introduce 'unsafe-inline' to script-src; that defeats the PR.
   - Preserve the form-action allowlist for *.microsoftonline.com,
     login.microsoft.com, login.microsoftonline.com — SSO redirects need it.
   - Preserve the existing third-party allowlists (Sentry, Stripe, Cloudflare
     Turnstile, PostHog, Plain, S3 buckets) on script-src/connect-src/frame-src/
     style-src/img-src.

5. Once the nonce is flowing, do a sanity check for CSP violations on the
   rendered page in a browser. Note any console errors and either nonce-stamp
   the offending inline content or 'sha256-' hash it in the CSP. Do not paper
   over with 'unsafe-inline'.

Constraints:
- Follow repo conventions in /Users/steffen/Documents/langfuse/CLAUDE.md and
  web/AGENTS.md.
- Keep changes scoped to this PR's intent.
- Do not hand-edit anything under generated/, .next/, or *.next-check/.
- If you change anything in packages/shared/**, also run that package's lint.

Report back: (1) what the actual root cause was, (2) the specific change you
made, (3) confirmed CSP header value and a sample nonce from the rendered HTML
proving it's non-empty and per-request.
```

Issue #11138, opened 2025-12-15
bug: High Postgres CPU from job_executions


scherervictor:
### Describe the bug

- Every time a trace is upserted, Langfuse checks “does an eval job for this trace + evaluator already exist?”
- To answer that, it runs a query on the job_executions table that should return at most one row, but on large projects the table has millions of rows.
- With a big backlog and multiple workers, this tiny “dedupe” check is executed constantly, which:
  - keeps Postgres CPU at 100%,
  - makes eval creation very slow, and
  - causes the trace-upsert queue (redis) to grow instead of draining.

### Steps to reproduce

*Context*
- In a prod setup with high eval throughput, Postgres CPU was constantly at 100%.
- Redis bull:trace-upsert:wait had millions of jobs; eval creation (createEvalJobs) was clearly the hot path.
- The job_executions table grew to ~1.7M rows for a single project.
*Obersavation*
- Using pg_stat_activity and EXPLAIN ANALYZE showed a hot query:
```
SELECT "id"
FROM "job_executions"
WHERE "project_id" = $1
  AND "job_configuration_id" = $2
  AND "job_input_trace_id" = $3
  AND "job_input_dataset_item_id" IS NULL
  AND "job_input_observation_id" IS NULL;
```

On a large job_executions table, Postgres was doing a parallel seq/bitmap heap scan over ~1.7M rows per call, e.g.:
```
Parallel Seq Scan on job_executions
  Rows Removed by Filter: ~1,706,870
  Execution Time: ~4.3s
```

*Why existing indexes weren’t enough*

- Prisma defines @@index([projectId, jobConfigurationId, jobInputTraceId]) on job_executions.
- However, the dedupe query also has job_input_dataset_item_id IS NULL and job_input_observation_id IS NULL.
- In practice, the planner was choosing:
  - either a parallel seq scan, or
  - a bitmap heap scan on (project_id, status) that still had to visit ~1.7M rows,
 rather than a highly selective index lookup.

Even with `enable_seqscan=off`, plans still scanned ~1.7M rows, confirming the planner didn’t have a good selective index for the exact predicate.

### Langfuse Cloud or self-hosted?

Self-hosted

### If self-hosted, what version are you running?

v3.139.0

### SDK and integration versions

-

### Additional information

After creating the index started to work as expected.

### Are you interested in contributing a fix for this bug?

Yes

scherervictor:
[...]

PR #12760, opened 2026-03-24
perf(prisma): add index on job_executions.job_configuration_id

wochinge:
## What does this PR do?

Speeds up slow deletes of LLM-as-a-judge.
Reasoning and context in ticket: https://linear.app/langfuse/issue/LFE-8961/deleting-a-llm-as-a-judge-takes-forever.

The PR addresses the issue by adding an index to the foreign key which is used for the cascading delete when the evaluator is deleted.

Fixes # (issue)

## Type of change

## Mandatory Tasks

### DIFF
--- a/packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
+++ b/packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
@@ -1,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX CONCURRENTLY "job_executions_job_configuration_id_idx" ON "job_executions"("job_configuration_id");
--- a/packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
+++ b/packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
@@ -1,2 +1,2 @@
 -- CreateIndex
-CREATE INDEX CONCURRENTLY "job_executions_job_configuration_id_idx" ON "job_executions"("job_configuration_id");
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "job_executions_job_configuration_id_idx" ON "job_executions"("job_configuration_id");
--- a/packages/shared/prisma/schema.prisma
+++ b/packages/shared/prisma/schema.prisma
@@ -1025,6 +1025,7 @@ model JobExecution {
 
   @@index([projectId, jobConfigurationId, jobInputTraceId])
   @@index([projectId, jobOutputScoreId])
+  @@index([jobConfigurationId])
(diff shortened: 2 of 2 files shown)


packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX CONCURRENTLY "job_executions_job_configuration_id_idx" ON "job_executions"("job_configuration_id");
```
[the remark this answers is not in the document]
hassiebp: Please see this comment, `IF NOT EXISTS` is crucial as we will have to likely apply this manually on our prod DBs and not via the default prisma migration workflow

hassiebp (changes requested):

hassiebp:
@wochinge could you please additionally look into this issue? it could make sense to create both indices together in one manual application flow :) https://github.com/langfuse/langfuse/issues/11138

packages/shared/prisma/migrations/20260324000000_add_job_executions_job_configuration_id_idx/migration.sql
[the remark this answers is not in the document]
wochinge: done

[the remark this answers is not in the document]
wochinge: done

packages/shared/prisma/schema.prisma
```diff
@@ -1025,6 +1025,7 @@ model JobExecution {
 
   @@index([projectId, jobConfigurationId, jobInputTraceId])
   @@index([projectId, jobOutputScoreId])
+  @@index([jobConfigurationId])
```
Steffen911, line 1028: Shouldn't this be covered by @@index([projectId, jobConfigurationId, jobInputTraceId]) ? Or do we perform deletions without a projectId condition?

wochinge: the deletion is a result from the foreign key cascade and that simply checks the foreign key itself.

The alternative would be a composite foreign key for `(projectId, jobConfigurationID`). Given the size of the table, the current index addition should be faster (and hence less risk on the migration). What do you think?

Steffen911: As discussed offline - both approaches seem valid and if you formed an opinion happy to follow along :) Let me know if you need help collecting data points or applying.

wochinge: Decided together with @Steffen911 and @hassiebp to go with the index route for now (instead of the composite foreign key route) due to
- simplicity of the migration
- this is most likely a temporary setup anyway as the table should be be moved to clickhouse

PR #12782, opened 2026-03-25
fix(dashboard): restore filter column labels when editing saved widgets

Krishnachaitanyakc:
## Summary

Fixes #8282 - When editing a saved dashboard widget, filter column labels (e.g., "User", "Session") were incorrectly displayed as "Column" in the filter builder UI.

**Root cause**: When saving a widget, filter column names are mapped from UI display names to view-level identifiers via `mapLegacyUiTableFilterToView` (e.g., `"User"` -> `"userId"`, `"Session"` -> `"sessionId"`). When loading a saved widget for editing, no reverse mapping was applied. The filter builder could not match the persisted view-level name (e.g., `"userId"`) against the widget form's filter column definitions (which use `id: "user"` and `name: "User"`), so it fell back to showing "Column".

**Fix**: 
- Added `mapViewFilterToWidgetFormFilter` as a reverse mapping function in `dashboardUiTableToViewMapping.ts` that converts persisted view-level column names back to the widget form's filter column display names
- Extracted `WIDGET_FILTER_COLUMNS` as a static lookup table of all filter columns used by the widget form
- Replaced the partial ad-hoc mapping (which only handled the `"name"` column) with the comprehensive reverse mapping function

This affects all filter columns that have different view-level names, including: User, Session, Release (observations view), Version (observations view), Trace Name, Observation Name, and Score Name.

## Test plan

- [ ] Create a widget with a "User" filter on the traces view, save it, then edit -- verify the "User" label is preserved in the filter builder
- [ ] Create a widget with "Session" and "User" filters on the observations view, save and re-edit -- verify both labels are correctly displayed
- [ ] Create a widget with a "Score Value" filter on the scores-numeric view, save and re-edit -- verify the label is preserved
- [ ] Create a widget with "Release" and "Version" filters on the observations view, save and re-edit -- verify labels are correct
- [ ] Verify that creating new widgets and adding filters still works correctly (forward mapping path unchanged)
- [ ] Verify that existing saved widgets with "Environment", "Tags", or "Metadata" filters (which have matching ids) still display correctly

### DIFF
--- a/web/src/features/query/dashboardUiTableToViewMapping.ts
+++ b/web/src/features/query/dashboardUiTableToViewMapping.ts
@@ -282,3 +282,50 @@ export const mapLegacyUiTableFilterToView = (
     return [{ ...filter, column: definition.viewName }];
   });
 };
+
+/**
+ * Reverse mapping: converts persisted view-level filter column names back to
+ * the widget form's filterColumns identifiers so that the filter builder UI
+ * can match them and display the correct column label.
+ *
+ * `targetColumns` is the array of ColumnDefinition objects used by the widget
+ * form's InlineFilterBuilder. A filter column value is remapped when it does
+ * not already match any column `id` or `name` in `targetColumns` but does
+ * match a `viewName` in the view mappings – in which case we look up the
+ * corresponding `uiTableName` and then resolve it to the matching
+ * targetColumn's `id`.
+ */
+export const mapViewFilterToWidgetFormFilter = (
+  view: z.infer<typeof views>,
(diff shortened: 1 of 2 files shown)


Steffen911:
@Krishnachaitanyakc Sorry for the lack of interaction here! Could you rebase the change and then I'll take a look.

Krishnachaitanyakc:
@Steffen911 Closing this PR as the underlying issue (#8282) was fixed in main by #13281

Steffen911:
@Krishnachaitanyakc Perfect, thank you a lot!

PR #12822, opened 2026-03-26
fix(events): populate tags and latency fallback in V3 trace adapter

### LINKED ISSUE #12736: bug: <[UI Bug] V3 Preview "Faster Experience" drops Latency pill and Tags for traces ingested via REST API (Works perfectly in Stable)>
### Describe the bug
When ingesting historical traces via the REST API (`/api/public/ingestion`) using a custom JSON batch payload, the new V3 "Faster Langfuse experience" UI fails to render the Latency pill and Tags array on the right-hand panel. 

However, when toggling the Preview UI OFF (returning to the Stable UI), the Latency pill and Tags render perfectly. It appears the new UI requires an OpenTelemetry-specific envelope to render these components, causing a degradation for users relying on the REST API for asynchronous/historical ETL ingestion.

### To Reproduce
1. Ingest a trace via `POST /api/public/ingestion`.
2. Include `startTime`, `endTime`, and a `tags` array in the trace-create JSON body.
3. Open the trace with the **Preview UI enabled**. Notice the Latency and Tags are missing from the right panel (though total time appears in the middle trace tree).
4. **Toggle Preview UI OFF**. Notice the Latency grey pill and Tags appear correctly on the right panel.

### Expected behavior
The new UI should reach feature parity with the stable UI, safely rendering `tags` and calculating the Latency pill from the `timestamp` / `endTime` fields of raw REST API ingestion payloads, without strictly requiring OTel SDK wrappers.

Langfuse Cloud

_No response_

_No response_

### Additional context
Ingesting webhooks from platforms like ElevenLabs requires historical "time-travel" (logging an event that happened 5 minutes ago). Because OpenTelemetry enforces current system time, the REST API is architecturally required here.

No

shawny011717:
## Summary

Fixes #12736.

The V3 preview trace adapter hardcodes `tags: []` instead of aggregating tags from observations. Additionally, latency returns `undefined` when no `endTime` values exist — the stable backend falls back to `max(startTime)`, which this adapter now mirrors.

## Changes

- Aggregate tags from all observations via `Set` deduplication
- Add `max(startTime)` fallback for latency computation when `endTime` is null

## Impacted packages

- `web`

### DIFF
--- a/web/src/features/events/lib/eventsToTraceAdapter.ts
+++ b/web/src/features/events/lib/eventsToTraceAdapter.ts
@@ -48,9 +48,10 @@ export function adaptEventsToTraceFormat(params: {
     .filter((t): t is Date => t !== null);
   const latestEnd =
     endTimes.length > 0 ? Math.max(...endTimes.map((d) => d.getTime())) : null;
+  const latestStart = Math.max(...events.map((e) => e.startTime.getTime()));
   const latencyMs = latestEnd
     ? latestEnd - earliest.startTime.getTime()
-    : undefined;
+    : latestStart - earliest.startTime.getTime() || undefined;
 
   // Create synthetic trace from observations
   const trace: SyntheticTrace = {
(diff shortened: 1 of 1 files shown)


shawny011717:
Gentle ping — this one's been waiting ~3 weeks for review. Happy to rebase or tweak if anything stands out. 🙏

PR #12823, opened 2026-03-26
fix(dashboard): restrict aggregations for uniqueness measures

shawny011717:
## Summary

Fixes #12273.

Dashboard widgets using "Unique User Ids" or "Unique Session Ids" with "Count" aggregation display total trace count instead of the actual distinct count. This happens because the inner query already computes `uniq(user_id)` per group (always yielding an integer), and wrapping that with `count()` in the outer query just counts rows.

The fix introduces a `uniqueCount` measure type that permits all numeric aggregations (sum, avg, max, min, percentiles, etc.) except `count`, which is semantically incorrect for pre-aggregated uniqueness measures. This type is applied to all `uniqueUserIds` and `uniqueSessionIds` measures across both v1 and v2 view definitions.

## Changes

- Add `uniqueCount` branch to `getValidAggregationsForMeasureType` in `types.ts` that filters out `count`
- Change measure type from `integer`/`string` to `uniqueCount` for `uniqueUserIds` and `uniqueSessionIds` in all views (`traceView`, `eventsTracesView`, `eventsObservationsView`)
- Add test coverage for the new type in `queryBuilder.servertest.ts`

## Impacted packages

- `web`

### DIFF
--- a/web/src/__tests__/server/queryBuilder.servertest.ts
+++ b/web/src/__tests__/server/queryBuilder.servertest.ts
@@ -4355,6 +4355,8 @@ describe("getValidAggregationsForMeasureType", () => {
     ["integer", allAggs],
     ["decimal", allAggs],
     ["number", allAggs],
+    // uniqueCount → all aggregations except count
+    ["uniqueCount", allAggs - 1],
     // Non-numeric / missing → restricted
     ["string", restricted.length],
     ["boolean", restricted.length],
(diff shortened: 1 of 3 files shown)


amit-gupta-:
@sumerman Can we priortise this fix ? This issue has been there for a long time and blocks the basic metrics to track Daily Active User count.

sumerman:
Hi @shawny011717 @amit-gupta-  I deliberately chose not to apply this change to v1 definitions because there are cases when widgets already built on top of v1 definitions might break for existing users. This will be fixed in Langfuse v4 with v2 definitions

PR #12826, opened 2026-03-26
fix(playground): make ctrl+enter shortcut work on mac

wochinge:
### Motivation

- Ensure the global "Run All" keyboard shortcut reliably triggers across input focus contexts and different keyboard layouts, and provide a clearer shortcut hint for macOS users.

### Description

- Update `useCommandEnter` to use `event.key === "Enter"` instead of `event.code` and add `preventDefault()`/`stopPropagation()` to avoid duplicate handling and unwanted default behavior. 
- Attach the keydown listener in the capture phase by passing `true` to `addEventListener`/`removeEventListener` so the shortcut fires even when focus is inside nested elements. 
- Add `isMac` detection to the playground page and render a keyboard hint that shows `⌘ + Enter` on macOS or `Ctrl + Enter` on other platforms in the Run All button. 
- Import `useEffect`/`useState` in the page component and adjust the Run All button markup to include a `kbd` element with appropriate styling.

### Testing

- Ran TypeScript type-check and a local build, which completed successfully. 
- Executed the existing automated unit test suite (`yarn test`), and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69c5127e196c83278148aafb565704b9)

### DIFF
--- a/web/src/features/playground/page/hooks/useCommandEnter.ts
+++ b/web/src/features/playground/page/hooks/useCommandEnter.ts
@@ -5,12 +5,12 @@ export default function useCommandEnter(
   callback: () => Promise<void>,
 ) {
   useEffect(() => {
+    const isMac = window.navigator.userAgent.includes("Mac");
+
     function handleKeyDown(event: KeyboardEvent) {
-      if (
-        isEnabled &&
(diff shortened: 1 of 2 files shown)


wochinge:
@claude review

PR #12844, opened 2026-03-26
chore(dx): upgrade to pnpm v10

nimarb:


nimarb:
@maxdeichmann please briefly check if i missed anything, works well locally so far for me as well as staging / CI

PR #12922, opened 2026-03-31
fix: stricter validaiton of cost and usage details on OTEL ingest path

sumerman:


sumerman:
@claude review

web/src/pages/api/public/otel/v1/traces/index.ts
```diff
@@ -166,6 +166,17 @@ export default withMiddlewares({
         ingestionVersion,
       });
 
+      // Validate span attributes before queueing — reject the entire batch if
+      // any span has an invalid cost_details or usage_details attribute.
+      const validationError = processor.validateResourceSpans(resourceSpans);
+      if (validationError) {
+        logger.warn(
+          `Rejected OTEL batch for project ${auth.scope.projectId}: ${validationError.attribute} on span ${validationError.spanId} is invalid (${validationError.reason})`,
+        );
```
Steffen911, line 175: Let's also add a metric to track how often this happens and for which span sizes. I think that would help our debugging and allow us to alert if this occurs too frequently.

web/src/pages/api/public/otel/v1/traces/index.ts
```diff
@@ -166,6 +166,17 @@ export default withMiddlewares({
         ingestionVersion,
       });
 
+      // Validate span attributes before queueing — reject the entire batch if
+      // any span has an invalid cost_details or usage_details attribute.
```
Steffen911, line 170: Did you discuss this behavior with Max or Marc? My thinking is that we may want to ignore cost or usage details in this case instead of rejecting the entire batch, as this could lead to silent, strange errors. In the past, we have preferred failing open rather than failing closed. Frameworks may add or start to support new formats, and rejecting those instead of handling them could be problematic.

packages/shared/src/server/otel/OtelIngestionProcessor.ts
```diff
@@ -219,6 +226,75 @@ export class OtelIngestionProcessor {
...
+            // Sums string attribute value lengths (input, output, metadata are all stringValues).
+            // Avoids JSON.stringify which would be prohibitively expensive on oversized spans.
+            const estimatedBytes = this.estimateSpanBytes(span);
+            if (estimatedBytes > maxSpanBytes) {
+              return {
+                spanId,
+                attribute: "span",
+                reason: "span_too_large",
+                message: `Span exceeds maximum size of ${maxSpanBytes} bytes`,
+              };
```
Steffen911, line 271: In the ClickHouse writer, we often try to truncate the span or the inputs and outputs up to a reasonable default, such as one megabyte, instead of dropping the entire thing. If I understand this flow correctly, dropping the entire batch due to one outsized span is not ideal. I think one outsized span shouldn't cause the entire process batch to be dropped; instead, we should fail open.

PR #12961, opened 2026-04-02
fix(web): Improve search highlighting in CodeMirrorEditor

bezbac:
## What does this PR do?

This PR updates the search functionality by introducing a custom search decoration instead of reusing the CodeMirror selection, and styles this decoration using a new CSS variable.

I think it makes sense to introduce a new color here rather than reuse an existing one, since search highlighting should be visually distinct. It is typically a bright yellow, which doesn’t have many other practical uses.

The new CSS variable names are `find-match-selected-background`, `find-match-selected-foreground`, and `find-match-background`. I reviewed naming conventions used in other tools and tried to align with them - for example, VS Code uses `editor.findMatchBackground` and `editor.findMatchHighlightBackground`: https://code.visualstudio.com/api/references/theme-color#editor-colors

The PR also removes the location message from the `MessageSearchToolbar` component, as there was not enough space to display the text properly. This made it illegible and caused the component to look unbalanced.

Loom Video: https://www.loom.com/share/65ab290172c8499eb685247048fac385

## Type of change

## Mandatory Tasks


web/src/components/editor/CodeMirrorEditor.tsx
```diff
@@ -174,18 +310,30 @@ export function applyCodeMirrorSearchQuery(
...
+  let current = cursor.next();
+  while (!current.done) {
+    matchRanges.push(current.value);
+    current = cursor.next();
+  }
+
+  view.dispatch({
+    effects: setSearchHighlightMarks.of(matchRanges),
   });
 }
```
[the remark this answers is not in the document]
bezbac: This is incorrect, in the current code the `search()` extension does not produce any highlights at all. The `controller.ts` used `selectCodeMirrorRange` to modify the selection. The selection was used instead of search match marks. The new extension (`searchHighlightingSupport`) was introduced so we wouldn't need to abuse the selection.

web/src/components/editor/CodeMirrorEditor.tsx
```diff
@@ -195,14 +343,24 @@ export function selectCodeMirrorRange(
...
 
   view.dispatch({
-    selection: {
-      anchor: range.from,
-      head: range.to,
-    },
+    effects: setSelectedSearchHighlightMark.of(range),
     scrollIntoView: true,
   });
 }
```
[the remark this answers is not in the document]
bezbac: I replaced `scrollIntoView: true` with `EditorView.scrollIntoView(range.from)`

web/src/components/editor/CodeMirrorEditor.tsx
```diff
@@ -277,6 +438,16 @@ export function CodeMirrorEditor({
...
+        // Update search match highlight styles
+        EditorView.theme({
+          ".cm-searchMatch": {
+            backgroundColor: "hsl(var(--find-match-background))",
+          },
+          ".cm-searchMatch.cm-searchMatch-selected": {
+            backgroundColor: "hsl(var(--find-match-selected-background))",
+            color: "hsl(var(--find-match-selected-foreground))",
+          },
+        }),
```
marliessophie, line 450: When I run locally, the CSS variable `--find-match-selected-background` isn't resolving - it's returning empty/invalid, so the browser falls back to transparent.

Most likely cause: CodeMirror's `EditorView.theme()` creates styles that are injected separately, and the CSS variables from your globals.css :root aren't being inherited in that context.

web/src/components/editor/CodeMirrorEditor.tsx
marliessophie, line 1: Global comment: when I add a message, then search, then edit the message and type another occurrence of the word I had searched, it is not highlighted though the counter is increased. Ideally, the match I typed is highlighted too. I understand this might have performance implications.

marliessophie (commented):
Directionally looks great, thanks for adding! 

Added two comments.

web/src/components/editor/CodeMirrorEditor.tsx
bezbac: This was resolved for me by adding the `--color-[NAME]` variables to the `@theme` in `global.css`. I'm not 100% certain why this fixes it but maybe the variable was previously pruned as unused?

Commit: a4b0519c978e1ba8c8e2064bbfa13ddb90c710f5

web/src/components/editor/CodeMirrorEditor.tsx
bezbac: Fixed in 2ffa382056999def4993c24eefb369d5370a9bd6

web/src/components/editor/CodeMirrorEditor.tsx
bezbac: Apparently this was not resolved by the shared commit, Nimar ran into the same issue: https://linear.app/langfuse/issue/LFE-8894/promptplayground-make-search-highlight-colour-more-promising#comment-9f7bc373.

I sat together with him to debug this and we think it's an issue with the dev build, we didn't encounter the issue when running via `pnpm build` and `pnpm start`.

marliessophie: I see. If you handle this issue together that sounds sensible to me 👍 Thanks for update. Feel free to self-resolve whenever you deem appropriate.

web/src/components/editor/CodeMirrorEditor.tsx
```diff
@@ -277,6 +438,16 @@ export function CodeMirrorEditor({
             outline: "none",
           },
         }),
+        // Update search match highlight styles
+        EditorView.theme({
+          ".cm-searchMatch.cm-searchMatch": {
```
nimarb, line 443: @bezbac , making it more specific here fixed the dev mode

nimarb (approved):

PR #12977, opened 2026-04-04
feat(cache): add local l1 cache for model match

hassiebp:
## Summary
- add a shared TTL-only process-local L1 cache primitive backed by `lru-cache`
- use the new local cache for hot model-match and pricing-tier reads while keeping Redis as L2 and preserving Redis invalidation
- add model-match config knobs, example env docs, and targeted shared/worker tests for local hits, negative caching, TTL, and single-flight behavior

## Why
Model matching sits directly on the hot ingestion cost path and currently pays repeated Redis round-trips. This change reduces Redis traffic during spikes by allowing very short-lived process-local reads for stable model metadata, without adding container-local invalidation complexity.

## Impacted Packages
- `packages/shared`
- `worker`
- `web`

## Validation
- `git diff --check`

## Validation Notes
- Local Node-based lint/tests were not run in this shell because `node` is not available in the environment.
- CI has been rerun on the latest branch updates.


hassiebp:
@claude review

hassiebp:
@claude review

hassiebp:
@claude review

hassiebp:
@claude review

PR #13037, opened 2026-04-08
fix(evals): prevent llm-as-a-judge queue stalls

hassiebp:
## Summary
- add a hard runtime timeout around Vertex/Google AI Studio LLM calls so hanging provider/auth requests cannot occupy LLM-as-a-judge worker slots indefinitely
- only mark eval and experiment executions as `DELAYED` when a retry was actually re-enqueued; otherwise fall through to terminal error handling
- add regression coverage for both the retry-state behavior and the new Vertex timeout behavior

## Root cause
The affected LLM-as-a-judge queue could stall when Google Vertex/Auth requests hung without an app-level abort, leaving all worker slots occupied while BullMQ still considered the workers healthy. Separately, retryable errors older than the 24h retry window could still be written as `DELAYED` even when no retry job was queued, leaving stale rows behind.

## Impact
- prevents queue starvation in the Vertex-backed LLM-as-a-judge path
- avoids permanent `DELAYED` job execution rows when retries are skipped
- keeps retry bookkeeping explicit for eval and experiment queues

## Validation
- `pnpm --filter worker run test src/queues/__tests__/llmAsJudgeExecutionQueueProcessor.test.ts src/__tests__/fetchLLMCompletionTimeout.test.ts`
- `pnpm --filter worker exec eslint --no-warn-ignored src/queues/evalQueue.ts src/features/utils/retry-handler.ts src/queues/experimentQueue.ts src/queues/__tests__/llmAsJudgeExecutionQueueProcessor.test.ts src/__tests__/fetchLLMCompletionTimeout.test.ts`
- `pnpm --filter @langfuse/shared exec eslint --no-warn-ignored src/server/llm/fetchLLMCompletion.ts`

## Notes
- full `pnpm --filter @langfuse/shared run typecheck` is still failing in this worktree because of unrelated existing Prisma/generated-client issues outside this patch


packages/shared/src/server/llm/fetchLLMCompletion.ts
```diff
@@ -109,6 +109,45 @@ type FetchLLMCompletionParams = LLMCompletionParams & {
   tools?: LLMToolDefinition[];
 };
 
+const RUNTIME_TIMEOUT_ADAPTERS = new Set([
+  LLMAdapter.VertexAI,
+  LLMAdapter.GoogleAIStudio,
+]);
+
+async function executeWithRuntimeTimeout<T>({
```
hassiebp: move this to packages/shared/src/server/llm/utils.ts

worker/src/queues/experimentQueue.ts
```diff
@@ -32,6 +32,8 @@ export const experimentCreateQueueProcessor = async (
         delayFn: delayInMs,
       });
 
+      if (retryResult.outcome === "scheduled") return;
```
hassiebp, line 35: this condition has no impact

PR #13044, opened 2026-04-08
feat(experiments): direct-write prompt experiment root events

hassiebp:
## What changed
- centralized internal direct event writes in the shared internal tracing flow so `fetchLLMCompletion` can handle events-table writes from one place
- added a shared internal-trace materializer that turns processed lifecycle events into final trace-rooted event inputs for both prompt experiments and non-experiment internal traces
- moved prompt experiments to pass experiment context into `traceSinkParams` instead of owning worker-side direct-write orchestration
- enabled the same direct-write path for LLM-as-a-judge traces
- kept the worker-side persistence logic in one reusable sink that still uses `createEventRecord()` and `writeEventRecord()`

## Why
The direct-write logic had started to spread into prompt experiments only, while other `fetchLLMCompletion` internal traces such as LLM-as-a-judge still only used legacy ingestion. That made the behavior inconsistent and made it harder to reason about duplication safety and root semantics.

## Duplicate handling
- when internal direct-write is enabled, the shared tracing handler now calls legacy ingestion with `forwardToEventsTable: false`
- this avoids happy-path duplication through the regular staging and propagation flow, including for LLM-as-a-judge traces
- prompt-experiment backfill remains unchanged as a delayed fallback path; on the happy path the direct write lands first and the later backfill skips the trace via the existing `events_core` anti-join

## Impact
- prompt experiments and judge traces now share the same internal-tracing direct-write path
- prompt experiments still schedule experiment evals from the trace-root observation
- the direct-write decision and staging suppression now live in one place instead of being spread across worker callsites

## Packages
- `packages/shared`
- `worker`

## Validation
- `worker/node_modules/.bin/vitest run worker/src/features/internal-tracing/__tests__/getInternalTracingHandler.test.ts worker/src/features/experiments/__tests__/scheduleExperimentEvals.test.ts worker/src/features/evaluation/__tests__/evalExecutionDeps.test.ts`

## Notes
- `pnpm --filter @langfuse/shared run typecheck` is still failing because of a pre-existing unrelated error in `packages/shared/scripts/seeder/seed-postgres.ts` (`TS2322` around `ScoreConfigDataType`)


worker/src/features/experiments/scheduleExperimentEvals.ts
```diff
@@ -13,83 +7,20 @@ import {
...
- * any configured observation-level evals. It mirrors the eval scheduling that
- * happens in the OTEL ingestion queue for externally-ingested observations.
- *
- * @param params.projectId - The project ID
- * @param params.traceId - The trace ID for the experiment run
- * @param params.datasetItem - The dataset item being evaluated
- * @param params.config - The prompt experiment configuration
- * @param params.generationDetails - The generation details extracted from traced events
- */
 export async function scheduleExperimentObservationEvals(
```
hassiebp, line 13: can we get rid of this entirely now and use the regular schedule observation eval functions?

marliessophie:
## PR Notes: Prompt Experiment Direct-Write Deduplication

### What changed
- **Simplified direct write**: Prompt experiment roots now emit `span_id = traceId` (original observation ID) — no ID remapping needed
- **Environment-based filtering**: Continuous backfill job now excludes traces/observations with `environment = 'langfuse-prompt-experiment'` via WHERE clause in `getRelevantTraces` and `getRelevantObservations`

https://github.com/langfuse/langfuse/commit/94d4e08169a3e430daee80a45ba644949a686307

- [Minor] **Interface cleanup**: Renamed `eventsTableWrite` → `eventsWriter`, simplified shared contract by encapsulating callback logic inside the factory

### Why this matters

- **ClickHouse deduplication is not viable** — ReplacingMergeTree deduplication can take months to collapse duplicates, making it unsuitable for preventing duplicate experiment data
- Instead, we prevent duplicates at the source:
  1. **Direct write** writes events with original IDs for prompt experiments
  2. **Continuous backfill** skips these traces/observations entirely via environment filter
  3. **Historic backfill** (`backfillExperimentsHistoric.ts`) remains unchanged — it processes all historic data since those weren't direct-written
- This approach is simpler and guarantees no duplicates rather than relying on eventual ClickHouse deduplication

#### Accepted variance: 2-span vs 3-span structure

- Direct write produces 2 spans (`{traceId}` → generation); backfill produces 3 spans (`t-{traceId}` → `{traceId}` → generation)
- Cause: LangChain emits trace-create and span-create with same ID — direct write merges them, backfill reads from separate tables
- No functional impact: `experiment_item_root_span_id` queries work identically

### Self-hosted safety

- **Env var guard**: `createInternalEventsWriter()` now does not execute the events-write when `LANGFUSE_EXPERIMENT_INSERT_INTO_EVENTS_TABLE !== "true"`, preventing direct event writes for self-hosters without the events table enabled.

### Existing behavior preserved

- Direct-write failures are swallowed and logged (same pattern as OTEL direct-write pipeline)
- For prompt experiments, backfill remains a delayed fallback if direct-write fails before any row lands
- For LLM-as-a-judge, no backfill safety net exists (unchanged)

### LLM-as-a-judge

- No concerns — there is no backfill for LLM-as-a-judge traces, hence no parity assumption required

packages/shared/src/server/llm/internalTraceEvents.ts
```diff
@@ -0,0 +1,456 @@
...
+        type: getSnapshotType(event.type),
+        metadata: {},
+      } satisfies InternalTraceSnapshot);
+
+    snapshots.set(spanId, mergeSnapshotEvent(existingSnapshot, event));
+  }
+
+  const rootSpanId = isExperiment ? `t-${traceId}` : originalRootId;
+  const finalSnapshots = isExperiment
+    ? remapSnapshotsForExperiment({
```
hassiebp: why are we remapping only for experiments?

hassiebp: ah I believe we cannot do this as for LLM-as-a-judge we need the executionTraceId predetermined for the linking on the score

worker/src/features/experiments/experimentServiceClickhouse.ts
```diff
@@ -202,9 +191,24 @@ async function processLLMCall(
...
+        description: config.datasetRun.description,
+        datasetId: datasetItem.datasetId,
+        itemId: datasetItem.id,
+        itemVersion: convertDateToClickhouseDateTime(datasetItem.validFrom),
+        itemExpectedOutput: datasetItem.expectedOutput,
+        itemMetadata: asRecord(datasetItem.metadata),
+      },
+      onRootEventRecordReady: async (rootEventRecord) => {
+        await scheduleExperimentObservationEvals({
+          observation: convertEventRecordToObservationForEval(rootEventRecord),
```
hassiebp, line 208: I think we have an issue here with self hosters that use experiment target evals but are not yet on the events table.

The internal tracing handler will return a root span here with ID in the shape `t-{traceId}` and the score will reference that observation, but that observation does not exist in v3 so the score is orphaned.

Could you please think about this case?

hassiebp: Quick fix would be to only remap in `internalTraceEvents` if we actually write to events table :) Please add good comments so we don't forget why we added that in the future

marliessophie: But they will have an observation of the ID `traceId` no? Can we just use that instead?

worker/src/features/eventPropagation/handleExperimentBackfill.ts
```diff
@@ -231,6 +231,7 @@ export async function getRelevantObservations(
       AND o.trace_id IN {traceIds: Array(String)}
       AND o.start_time >= {minTime: DateTime64(3)} - interval 4 hour
       AND o.start_time <= {maxTime: DateTime64(3)} + interval 7 day
+      AND coalesce(o.environment, '') != 'langfuse-prompt-experiment'
     ORDER BY o.event_ts DESC
```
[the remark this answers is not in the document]
marliessophie: that's expected. we rely only on the direct-write going forward.

PR #13047, opened 2026-04-09
fix(evals): make evaluation prompt read-only in view-only template mode

pratimapatel2008:
## What does this PR do?

Non-editable CodeMirror instances (e.g. LLM-as-a-Judge **Evaluator Library** in view-only mode) still accepted paste because CodeMirror 6 treats `EditorView.editable` separately from `EditorState.readOnly`. This adds `EditorState.readOnly.of(true)` whenever `editable={false}` on the shared `CodeMirrorEditor`.

## Impacted packages
- `web` (`web/src/components/editor/CodeMirrorEditor.tsx`)

Fixes # 
The issue is that in LLM-as-a-Judge(evals) it allows to paste anything in the "prompt" section even though it is a view-only page. 

## Type of change

## Mandatory Tasks

## Checklist


marliessophie (approved):
LGTM, thanks for the contribution!

PR #13055, opened 2026-04-09
fix(web): require secret key for LLM test base URL changes

hassiebp:
## Summary
- block test-update calls from reusing the stored secret when the requested `baseURL` changes without a new `secretKey`
- prevent stored `extraHeaders` from being reused when `baseURL` changes, unless new headers are explicitly supplied in the same request
- add focused server tests covering the blocked case, same-host reuse, and changed-host header dropping

## Why
Changing the base URL during an LLM connection test can redirect sensitive credentials to a different endpoint. The original fix prevented silent reuse of the stored `secretKey`, but stored encrypted `extraHeaders` could still be decrypted and forwarded to an attacker-controlled `baseURL`. This update treats both pieces of sensitive state consistently.

## Impact
- updating and testing an LLM connection with the same base URL still reuses the stored secret and stored extra headers for the test
- changing the base URL during `testUpdate` now requires a non-empty `secretKey` in the same request
- changing the base URL during `testUpdate` no longer reuses stored `extraHeaders`; callers must explicitly provide any headers they want sent to the new endpoint
- project members without `llmApiKeys:create` remain blocked from this test path

## Impacted packages
- `web`

## Verification
- `git diff --check`
- attempted `pnpm --filter web run test --testPathPatterns="src/__tests__/server/llm-api-key.servertest.ts"`
- the Jest run is currently blocked in this Codex shell because `node` is not available on `PATH`, so the `dotenv` shim cannot start Jest


hassiebp:
@claude review

PR #13057, opened 2026-04-09
fix(web): Create new `TableCellWithCopyButton` for `ApiKeyList`

bezbac:
## What does this PR do?

This PR improves the API key table in the web app by introducing a dedicated `TableCellWithCopyButton` component. Instead of rendering the public key through `CodeView`, the table now uses a purpose-built table cell with an inline copy action, which is a better fit for truncated table content and preserves focus correctly after copy interactions.

To support this, the PR also extracts copy-state handling into a reusable `useCopyToClipboard` hook and updates `CodeJsonViewer` to use the same shared logic. The new hook is covered by client tests for temporary success state handling, timeout resets, and rerender behavior.

Screenshot 2026-04-09 at 13 39 19

## Type of change

## Mandatory Tasks

### DIFF
--- a/web/src/hooks/useCopyToClipboard.clienttest.ts
+++ b/web/src/hooks/useCopyToClipboard.clienttest.ts
@@ -1,0 +1,112 @@
+import { act, renderHook } from "@testing-library/react";
+import { useCopyToClipboard } from "@/src/hooks/useCopyToClipboard";
+
+describe("useCopyToClipboard", () => {
+  const originalClipboard = navigator.clipboard;
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+    Object.defineProperty(navigator, "clipboard", {
+      configurable: true,
+      value: {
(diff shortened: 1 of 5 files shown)


web/src/hooks/useCopyToClipboard.clienttest.ts
nimarb, line 1: while the tests are true, i don't think we need them. they test very straight forward things here. i often write those tests for dev but then not commit them, mostly due to test performance reasons.

we could argue to keep the 1st test, it kinda makes sense but is also very narrow.

wdyt?

nimarb (approved):

web/src/hooks/useCopyToClipboard.clienttest.ts
bezbac: Done in 51ef7e2718dd070833fcf897de31c3281dccbcd0

PR #13073, opened 2026-04-09
fix(llm-connections): validate new LLM base URLs

hassiebp:
## Summary
This PR adds server-side validation for newly provided or changed LLM connection `baseURL` values.

It blocks loopback, private, link-local, and other internal network targets by default, while letting self-hosted operators explicitly allow trusted internal hosts or ranges through environment-based allowlists.

## What changed
- added a shared LLM connection base URL validator in `@langfuse/shared`
- applied validation to LLM connection create, update, test, and `testUpdate` flows
- applied the same validation to the public `PUT /api/public/llm-connections` route
- added self-hosted allowlist env vars for hosts, IPs, and CIDR ranges
- documented the new env vars in the example env files
- added regression coverage for internal/test/public API validation behavior

## Important behavior
This intentionally validates only newly provided or changed `baseURL` values.

Existing saved `baseURL`s are not revalidated during normal LLM calls, so this does not retroactively break already stored connections.

## Why
Previously, `baseURL` validation only checked that the value was a syntactically valid URL. That meant callers could point test or saved connections at internal targets such as `localhost` or `127.0.0.1`, which is too permissive for Langfuse Cloud and risky by default for self-hosted deployments.

## Impact
- Langfuse Cloud now rejects unsafe internal targets for new or updated LLM connection base URLs
- self-hosted users can still support trusted local proxies by explicitly allowlisting them
- existing stored connections continue to work unless their base URL is changed

## Root cause
We treated `baseURL` as a string-format validation problem instead of as an outbound network target that needs SSRF-aware checks.

## Verification
- `pnpm --filter web exec eslint src/features/llm-api-key/server/router.ts src/pages/api/public/llm-connections/index.ts src/__tests__/server/llm-api-key.servertest.ts src/__tests__/server/llm-connections-api.servertest.ts`
- `CLICKHOUSE_URL=http://localhost:8123 CLICKHOUSE_USER=test CLICKHOUSE_PASSWORD=test LANGFUSE_S3_EVENT_UPLOAD_BUCKET=test pnpm --filter worker exec vitest run src/__tests__/llm-base-url-validation.test.ts`

## Notes
I also attempted the targeted web server Jest test, but it was not runnable in this environment because the required local web env bootstrap values were not set (`DATABASE_URL`, `NEXTAUTH_URL`, `SALT`, and `CLICKHOUSE_*`).

Impacted packages: `web`, `worker`, `@langfuse/shared`


hassiebp:
@claude review

PR #13125, opened 2026-04-12
fix: URL-encode ClickHouse credentials in migration scripts

Yanhu007:
## Summary

Fixes #12077

ClickHouse migration scripts (`up.sh`, `down.sh`, `drop.sh`) interpolate `CLICKHOUSE_PASSWORD` directly into the database connection URL. Passwords containing special characters (`@`, `&`, `#`, `!`, `%`, etc.) break the URL parser:

```
password=MyP@ss&Word#123!
        ↑     ↑    ↑   ↑
        host  param fragment
        sep   sep  
```

Result: `Authentication failed: password is incorrect`

## Fix

Add `url_encode()` helper using Python's `urllib.parse.quote` and encode both username and password before URL construction:

```bash
url_encode() {
    python3 -c "import urllib.parse, sys; print(urllib.parse.quote(sys.argv[1], safe=''))" "$1"
}

ENCODED_CLICKHOUSE_PASSWORD=$(url_encode "$CLICKHOUSE_PASSWORD")
ENCODED_CLICKHOUSE_USER=$(url_encode "$CLICKHOUSE_USER")
```

Applied to all 3 scripts: `up.sh`, `down.sh`, `drop.sh`.

### DIFF
--- a/packages/shared/clickhouse/scripts/down.sh
+++ b/packages/shared/clickhouse/scripts/down.sh
@@ -28,12 +28,20 @@ if [ -z "${CLICKHOUSE_CLUSTER_NAME}" ]; then
     export CLICKHOUSE_CLUSTER_NAME="default"
 fi
 
+# URL-encode a string (handles special chars like @, &, #, !, %, etc.)
+url_encode() {
+    python3 -c "import urllib.parse, sys; print(urllib.parse.quote(sys.argv[1], safe=''))" "$1"
+}
+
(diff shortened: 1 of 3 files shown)


Steffen911:
@Yanhu007 Thank you for the contribution. With https://github.com/langfuse/langfuse/pull/13186 we've added a patch that more clearly highlights if a missing encoding could be the root of the connection issues. As this should patch this sufficiently, I'll go ahead and close this.

PR #13147, opened 2026-04-14
fix: force functionCalling method for all adapters in withStructuredOutput

Arundhathi-Nair-S:
## What does this PR do?

Fixes a bug where LLM-as-a-judge evaluators fail to save when using OpenAI-compatible providers that don't support `response_format: json_schema` (e.g. Sarvam AI `sarvam-30b`).

In `fetchLLMCompletion.ts`, `structuredOutputConfig` was only set to `{ method: "functionCalling" }` for thinking-capable adapters (VertexAI, GoogleAIStudio). All other adapters received `undefined`, causing LangChain's `withStructuredOutput()` to default to `json_schema`. Providers that don't support `json_schema` silently ignore the parameter and return plain text/markdown, which Langfuse then fails to parse as JSON → `PRECONDITION_FAILED` on eval save.

**Error seen:**
```
TRPCError: Model configuration not valid for evaluation.
Unexpected token '*', "**Score:**"... is not valid JSON
PRECONDITION_FAILED
```

**Root cause** (`packages/shared/src/server/llm/fetchLLMCompletion.ts` lines 536–539):
```typescript
// BEFORE — undefined for all non-thinking adapters → LangChain defaults to json_schema
const structuredOutputConfig =
  thinkingTypes != null
    ? { method: "functionCalling" as const }
    : undefined;

// AFTER — force functionCalling for all adapters
const structuredOutputConfig = { method: "functionCalling" as const };
```

**Why this is safe:** `functionCalling` (tools + tool_choice) is more universally supported across OpenAI-compatible providers than `json_schema`. It also works correctly for thinking-capable adapters (VertexAI, GoogleAIStudio) — the `thinkingTypes` variable is still used for reasoning block extraction downstream.

## Evidence

| Method | Sarvam AI (`sarvam-30b`) result |
|--------|-------------------------------|
| `response_format: json_schema` (current) | Markdown essay in `content`, `tool_calls: null` — silently ignored |
| `tools` + `tool_choice: required` (fix) | Clean JSON in `tool_calls[].function.arguments`, `finish_reason: "tool_calls"` |

## Type of change

## Mandatory Tasks

## Checklist

- [x] My changes generate no new warnings (`pnpm run lint`) — change is a single-line simplification with no new imports or syntax


Arundhathi-Nair-S:
**Additional context for reviewers:**

We hit this bug on a self-hosted Langfuse instance (v3.162.0) using Sarvam AI (`sarvam-30b`) as an LLM-as-a-judge evaluator. The eval save consistently failed with:

```
TRPCError: Model configuration not valid for evaluation.
Unexpected token '*', "**Score:**"... is not valid JSON
PRECONDITION_FAILED
```

We confirmed the root cause by testing the two methods directly against the Sarvam API:

**With `json_schema` (what Langfuse currently sends):**
- Response: markdown essay in `content`, `tool_calls: null`
- `json_schema` is silently ignored by the provider

**With `tools` + `tool_choice: required` (functionCalling — this fix):**
- Response: clean JSON in `tool_calls[].function.arguments`
- `finish_reason: "tool_calls"` — works perfectly

This is likely to affect any OpenAI-compatible provider that doesn't fully implement `json_schema` structured output support, not just Sarvam AI.

PR #13165, opened 2026-04-14
fix: Missing trace tags in v4 table and detail view

bezbac:
## What does this PR do?

This PR fixes missing trace tags in the "Fast (Preview)" experience.

Specifically:
- restores trace tags for events table observations
- fixes the events-backed observation by-id path so it returns the same trace-level fields as the legacy path which are rendered in the `Preview` tab to match the v3 details layout

Table:
Screenshot 2026-04-14 at 15 37 04

Details:
Screenshot 2026-04-14 at 15 37 13

## Type of change

## Mandatory Tasks


packages/shared/src/server/repositories/events.ts
```diff
@@ -79,7 +79,14 @@ import { parseMetadataCHRecordToDomain } from "../utils/metadata_conversion";
...
   "trace_tags" | "trace_name" | "trace_user_id"
->;
+> & {
+  tags?: string[];
+};
+
+type EventsObservationWithPriceAndRawTags = EventsObservation &
+  ObservationPriceFields & {
+    tags?: string[];
+  };
```
Steffen911: I would propose to extend EventsObservations directly instead of including tags in here.

packages/shared/src/server/repositories/events.ts
```diff
@@ -163,6 +170,7 @@ async function enrichObservationsWithModelData(
 
     const enriched = {
       ...converted,
+      ...(!isV2 ? { tags: o.tags ?? [] } : {}),
```
Steffen911: Can we always include that?

packages/shared/src/server/repositories/events.ts
```diff
@@ -589,6 +601,7 @@ export const getObservationByIdFromEventsTable = async ({
   traceId,
   renderingProps = DEFAULT_RENDERING_PROPS,
   preferredClickhouseService,
+  includeTraceTags = true,
```
Steffen911: Do we need this flag? I would just always include them if we have them anyway in the query result and the table.

packages/shared/src/server/repositories/events.ts
```diff
@@ -1090,12 +1110,19 @@ export const getObservationsFromEventsTableForPublicApi = async (
...
     opts.projectId,
     opts.parseIoAsJson ?? true, // V1 API: default to parsing JSON (backwards compatibility)
     null, // V1 API: no field groups, return complete observations
   );
+
+  return observations.map((observation) => {
+    // Tags should not be returned in the Public API
+    const { tags: _tags, ...observationWithoutRawTags } = observation;
+    return observationWithoutRawTags;
+  });
```
Steffen911: Why do we have to map _tags here? Can we do that on the query level?

web/src/pages/api/public/observations/[observationId].ts
```diff
@@ -31,6 +31,7 @@ export default withMiddlewares({
             id: query.observationId,
             projectId: auth.scope.projectId,
             fetchWithInputOutput: true,
+            includeTraceTags: false,
```
Comment (Steffen911) [inline web/src/pages/api/public/observations/[observationId].ts]: Can we just include them and ignore if necessary? IMO shouldn't make much of a difference if we add this as a new field to the API.

web/src/server/api/routers/traces.ts
```diff
@@ -82,6 +82,8 @@ export type ObservationReturnTypeWithMetadata = Omit<
 > & {
   traceId: string;
   metadata: string | null;
+  traceName?: string | null;
+  traceTags?: string[];
   // optional, because in v4 an observation can have those properties
   userId?: string | null;
   sessionId?: string | null;
```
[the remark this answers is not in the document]
bezbac: The updateTags trpc route is not used in the codebase (0 results for `api.traces.updateTags`)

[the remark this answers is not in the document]
bezbac: Created a linear issue to remove the route: https://linear.app/langfuse/issue/LFE-9285/remove-apitracesupdatetags

packages/shared/src/server/repositories/events.ts
bezbac: By extending `EventsObservations` directly as suggested in https://github.com/langfuse/langfuse/pull/13165#discussion_r3086150586, it would be weird to now exclude it here. I removed the exclusion completely and now drop the field in `transformDbToApiObservation`

web/src/components/trace2/components/ObservationDetailView/ObservationDetailView.tsx
```diff
@@ -397,6 +398,23 @@ export function ObservationDetailView({
...
+                <div
+                  className={`flex flex-wrap gap-x-1 gap-y-1 px-2 pb-2 ${currentView !== "pretty" ? "shrink-0" : ""}`}
+                >
+                  <TagList
+                    selectedTags={observation.traceTags}
+                    isLoading={false}
+                  />
+                </div>
+              </>
+            )}
```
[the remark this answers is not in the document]
bezbac: Resolved in a95f955f83f867e217ec226e383d2e6c879fe08c

web/src/components/trace2/components/ObservationDetailView/ObservationDetailView.tsx
```diff
@@ -397,6 +398,24 @@ export function ObservationDetailView({
...
+                  className={`flex flex-wrap gap-x-1 gap-y-1 px-2 pb-2 ${currentView !== "pretty" ? "shrink-0" : ""}`}
+                >
+                  <TagList
+                    selectedTags={observation.traceTags}
+                    isLoading={false}
+                    viewOnly
+                  />
+                </div>
+              </>
+            )}
```
nimarb: wdyt about showing them on each observation vs only on the top level trace? i'd only show them on the top level, as we do now. results in a small discrepancy between table view and detail view but i think its ok

web/src/features/events/components/EventsTable.tsx
```diff
@@ -1022,14 +1022,15 @@ export default function ObservationsEventsTable({
...
+          traceTags &&
+          traceTags.length > 0 && (
             <div
               className={cn(
                 "flex gap-x-2 gap-y-1",
                 rowHeight !== "s" && "flex-wrap",
               )}
             >
-              <TagList selectedTags={traceTags} isLoading={false} />
+              <TagList selectedTags={traceTags} isLoading={false} viewOnly />
```
nimarb: technically correct, but i think the `viewOnly` rendering looks a bit ugly, wdyt? click on table row is hijacked anyways, so we can just omit the viewOnly.

web/src/components/trace2/components/ObservationDetailView/ObservationDetailView.tsx
```diff
@@ -397,6 +398,24 @@ export function ObservationDetailView({
...
+                >
+                  Tags
+                </div>
+                <div
+                  className={`flex flex-wrap gap-x-1 gap-y-1 px-2 pb-2 ${currentView !== "pretty" ? "shrink-0" : ""}`}
+                >
+                  <TagList
+                    selectedTags={observation.traceTags}
+                    isLoading={false}
+                    viewOnly
```
nimarb: technically correct, but i think the `viewOnly` rendering looks a bit ugly, wdyt? curosor is suggestive of interaction but interaction itself is already not possible

web/src/features/public-api/types/observations.ts
```diff
@@ -163,7 +168,10 @@ export const transformDbToApiObservation = (
 
     public: _public,
     ...rest
-  } = observation as EventsObservation & ObservationPriceFields;
+  } = observation as EventsObservation &
+    ObservationPriceFields & {
+      traceTags?: string[];
```
nimarb, line 177: lets add a comment here why we need this (ie because we renamed `tags` to `traceTags` and thus doesnt match db type)

web/src/components/trace2/components/ObservationDetailView/ObservationDetailView.tsx
bezbac: I would also keep them on the top level as this is what the users are used too. We can always move them to each individual observation later if that makes sense.

web/src/features/events/components/EventsTable.tsx
bezbac: I dropped [a95f955](https://github.com/langfuse/langfuse/pull/13165/commits/a95f955f83f867e217ec226e383d2e6c879fe08c)

web/src/components/trace2/components/ObservationDetailView/ObservationDetailView.tsx
bezbac: I dropped [a95f955](https://github.com/langfuse/langfuse/pull/13165/commits/a95f955f83f867e217ec226e383d2e6c879fe08c)

web/src/features/public-api/types/observations.ts
bezbac: Done in e99e7c3831c36b09c25be9ffd4526c56dbe41304

packages/shared/src/server/repositories/observations_converters.ts
```diff
@@ -411,6 +411,7 @@ export function convertEventsObservation(
     userId: record.user_id ?? null,
     sessionId: record.session_id ?? null,
     traceName: record.trace_name ?? null,
+    tags: record.tags ?? [],
     bookmarked: record.bookmarked,
```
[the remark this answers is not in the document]
bezbac: This was asked for https://github.com/langfuse/langfuse/pull/13165#discussion_r3086152410

PR #13171, opened 2026-04-14
feat(web): add in-process L1 cache for API key auth lookups

sumerman:


sumerman:
Unfortunately ineffective

PR #13215, opened 2026-04-16
fix: block private IPv6 variants in image URL validation

### LINKED ISSUE #13213: bug: validateImgUrl allows IPv6 ULA and IPv4-mapped private addresses
### Describe the bug

`validateImgUrl` still accepts internal/private targets when a hostname resolves to certain IPv6 addresses that are not treated as private by `isPrivateIp`.

At the moment, the IPv6 branch only blocks:
- loopback (`::1`)
- link-local (`fe80::/10` via `isLinkLocal()`)

It does **not** block at least these private/internal cases:
- IPv6 Unique Local Addresses (`fc00::/7`), e.g. `fd12::1`
- IPv4-mapped private IPv6 addresses, e.g. `::ffff:10.0.0.1`

That means the image URL validation path can still return `{ isValid: true }` for internal targets even after the fix from `#13207`.

### Steps to reproduce

I validated this locally against `v3.167.4` with a focused server test that mocks DNS resolution.

Minimal reproduction logic:

1. Create a logged-in tRPC caller for `utilities.validateImgUrl`
2. Mock `dns.resolve4` / `dns.resolve6`
3. Mock `fetch` to return `200` with `content-type: image/png`
4. Call `validateImgUrl` with a hostname that resolves only to a private IPv6 target

Example cases:

```ts
mockedResolve4.mockResolvedValue([]);
mockedResolve6.mockResolvedValue(["fd12::1"]);

const result = await caller.utilities.validateImgUrl(
  "https://ula-only.example.test/image.png",
);
// expected: { isValid: false }
// actual:   { isValid: true }
```

```ts
mockedResolve4.mockResolvedValue([]);
mockedResolve6.mockResolvedValue(["::ffff:10.0.0.1"]);

const result = await caller.utilities.validateImgUrl(
  "https://mapped-private.example.test/image.png",
);
// expected: { isValid: false }
// actual:   { isValid: true }
```

I also validated the helper behavior directly with the same `ip-address` dependency currently used by the repo:

- `::1 => true`
- `fd12::1 => false`
- `::ffff:10.0.0.1 => false`
- `10.0.0.1 => true`

So the gap is specifically in IPv6 private-address coverage.

Self-hosted

`v3.167.4`

_No response_

### Additional information

I also checked the current `main` branch source after `#13207`, and the IPv6 helper logic still appears to only use:

```ts
address.isLinkLocal() || address.isLoopback()
```

so this looks like a remaining SSRF-style validation gap rather than a release-only artifact.

My suggestion would be to extend the IPv6 private-address detection to cover at least:
- `fc00::/7`
- IPv4-mapped private IPv6 addresses (e.g. by mapping to IPv4 and reusing the existing IPv4 private-range checks)

Yes

shaun0927:
Fixes #13213

## Summary
- block IPv6 Unique Local Addresses (`fc00::/7`) in `validateImgUrl`
- block IPv4-mapped private IPv6 addresses by reusing the existing IPv4 private-range checks
- add focused server regressions for both missed cases

## Why this scope
Recent accepted security fixes in this area were merged as small, tightly scoped patches with direct tests (for example `#13207`). I kept this change limited to the existing `validateImgUrl` security boundary instead of refactoring the broader URL-validation flow.

## Impacted packages
- `web` — `utilities.ts` image URL validation and targeted server regression tests

## Test plan
- `pnpm --filter web run test --testPathPatterns='utilities-router.servertest.ts'`
- `pnpm --filter web exec dotenv -e ../.env -- eslint src/server/api/routers/utilities.ts src/__tests__/server/utilities-router.servertest.ts --max-warnings 0`

## Notes
- The targeted test run in my local environment logs a missing local Postgres instance during the shared Jest setup/teardown path, but the focused test body still runs and passes.
- This PR intentionally does not revisit the already-fixed dual-stack bypass from `#13207`; it only closes the remaining IPv6 private-address gap.

### DIFF
--- a/web/src/__tests__/server/utilities-router.servertest.ts
+++ b/web/src/__tests__/server/utilities-router.servertest.ts
@@ -1,0 +1,67 @@
+/** @jest-environment node */
+
+jest.mock("dns", () => ({
+  promises: {
+    resolve4: jest.fn(),
+    resolve6: jest.fn(),
+  },
+}));
+
+import { promises as dns } from "dns";
+import { appRouter } from "@/src/server/api/root";
+import { createInnerTRPCContext } from "@/src/server/api/trpc";
+
+const mockedResolve4 = jest.mocked(dns.resolve4);
+const mockedResolve6 = jest.mocked(dns.resolve6);
+
+describe("utilities.validateImgUrl", () => {
+  const caller = appRouter.createCaller(
+    createInnerTRPCContext({
+      session: {
+        expires: "1",
+        user: {
+          id: "user-1",
+          organizations: [],
+          featureFlags: {},
+          admin: false,
+        },
+      } as any,
+      headers: {},
+    }),
+  );
+
+  beforeEach(() => {
+    mockedResolve4.mockReset();
+    mockedResolve6.mockReset();
+    global.fetch = jest.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      headers: {
+        get: (name: string) =>
+          name.toLowerCase() === "content-type" ? "image/png" : null,
+      },
+    }) as any;
+  });
+
+  it("should reject hostnames that resolve to an IPv6 unique local address", async () => {
+    mockedResolve4.mockResolvedValue([]);
+    mockedResolve6.mockResolvedValue(["fd12::1"]);
+
+    await expect(
+      caller.utilities.validateImgUrl(
+        "https://ula-only.example.test/image.png",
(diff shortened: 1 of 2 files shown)


Steffen911:
@shaun0927 This should be patched in https://github.com/langfuse/langfuse/commit/aecb6ef2becbcbdd74176b5500453061ae923441. If something is missing from your perspective, please let me know.

PR #13221, opened 2026-04-17
feat(experiments): add enabled toggle for remote dataset run trigger

toyayuto:
## What does this PR do?

Adds an `enabled` boolean flag to the remote dataset run trigger so users can **temporarily disable** the trigger without losing the configured URL and default payload.

### Motivation

Today, as soon as a remote experiment URL is set on a dataset, every Run click in the UI fires a `fetch` to that URL. If the URL is unreachable (intentionally, e.g. the team is not using remote triggers right now), users see a `Failed to trigger remote experiment / Failed to connect to remote server` error toast on every run.

The only way to silence this today is to delete the URL entirely — which is lossy (users have to re-enter the URL and payload every time they re-enable the feature).

This PR adds a simple switch in the `Edit remote dataset run trigger` dialog that:

- Hides the `Run` button in the experiment dialog while keeping the `Edit` (gear) button visible
- Skips the remote HTTP call on the server side, returning `{ success: true, skipped: true }` instead of attempting to reach a disabled endpoint

### Impacted packages

- `packages/shared` — Prisma schema + migration for `Dataset.remote_experiment_enabled`
- `web` — `upsertRemoteExperiment` / `getRemoteExperiment` / `triggerRemoteExperiment` tRPC procedures, `RemoteExperimentUpsertForm` UI toggle, `CreateExperimentsForm` Run-button gating

### Changes

**Schema**

- `packages/shared/prisma/schema.prisma` — new `remoteExperimentEnabled Boolean @default(true)` column on `Dataset`
- `packages/shared/prisma/migrations/20260417000000_add_remote_experiment_enabled_to_datasets/migration.sql` — additive migration, safe for existing installations (defaults to `true`, i.e. existing datasets keep working exactly as before)

**Server**

- `web/src/features/datasets/server/dataset-router.ts`
  - `upsertRemoteExperiment`: accepts optional `enabled: boolean` input
  - `getRemoteExperiment`: returns `enabled` alongside `url` and `payload`
  - `triggerRemoteExperiment`: early-returns `{ success: true, skipped: true }` when `remoteExperimentEnabled === false` (guards the case where the flag is set to `false` but the UI still somehow calls the endpoint)
- `web/src/features/datasets/server/actions/createDataset.ts` — `updateDataset` accepts optional `remoteExperimentEnabled`

**UI**

- `web/src/features/experiments/components/RemoteExperimentUpsertForm.tsx` — adds a `Switch` bound to the new `enabled` field (defaults to `true` for backward compatibility)
- `web/src/features/experiments/components/CreateExperimentsForm.tsx` — hides the `Run` button when `enabled === false`, keeps the gear icon accessible so users can re-enable

### Backward compatibility

- Default is `true` at both the column (`DEFAULT true`) and the form (`existingRemoteExperiment?.enabled ?? true`) levels — existing datasets behave identically.
- `triggerRemoteExperiment` returns `success: true` when skipped, so the UI success toast still shows (no user-facing behavior change for enabled triggers).

### Notes for reviewers

- Open to alternative UX: e.g. disabled-but-visible Run button vs. hidden Run button. The current PR hides it to match the user intent of "I don't want to accidentally trigger this."
- The server-side skip is defensive — the UI already hides the button when disabled, but I wanted to avoid relying solely on UI checks.

## Type of change

## Mandatory Tasks


nimarb:
@claude review

nimarb:
@toyayuto can you check why the test here is failing? thank you!

toyayuto:
> @toyayuto can you check why the test here is failing? thank you!

@nimarb 
  Thanks for the nudge! Root cause: I had forgotten to add `remoteExperimentEnabled`
  to the `APIDataset` Zod response schema (`web/src/features/public-api/types/datasets.ts`),
  so its `.strict()` mode was rejecting the new column in all 24
  `datasets-schema-validation.servertest.ts` cases with `unrecognized_keys`.
  Fixed in 34526d035.

  Also addressed the three Claude review comments in a follow-up commit ea15d1286
  (reset enabled on delete, handle skipped toast, fix allDatasets type).
  CI is re-running now.

nimarb:
@toyayuto i think build still fails -- thank ytou !

toyayuto:
Hi @nimarb — sorry for the repeated CI cycles. I've now verified locally that CI should pass:

  **Root causes of the previous two failures:**
  1. `APIDataset` response schema was missing `remoteExperimentEnabled` → fixed in 34526d0
  2. List dataset handlers (`GET /v2/datasets`, `GET /datasets`) used Prisma `select` that omitted `remoteExperimentEnabled` → fixed in bfdf6d7

  **Local verification on bfdf6d7:**
  - `pnpm --filter web run typecheck`: no `remoteExperimentEnabled`-related errors
  - `datasets-schema-validation.servertest.ts`: 46/46 pass
  - All four `transformDbDatasetToAPIDataset` call sites (v1/v2 × get/list) reviewed for select narrowing — no remaining gaps

  Could you approve the CI run when you have a moment? Thanks!

marliessophie:
Hey @toyayuto - thanks for adding this change. As we know that your team heavily relies on such functionality, we'll get this live. I will merge to my own branch, make sure to run all migrations & test and merge into main branch. Once on main branch, it will be included in next open source release.

Issue #10972, opened 2025-12-08
Support configurable ensure_ascii parameter in EventSerializer


homejim:
# Describe the feature or potential improvement
When integrating Langfuse with LangGraph, I noticed that all reported data is ASCII-encoded, which causes Chinese characters (and other non-ASCII characters) to not display properly.
After investigation, I found that the EventSerializer class in langfuse/_utils/serializer.py has ensure_ascii=True set by default in its JSON serialization.

#Current Behavior
Chinese characters and other non-ASCII text are encoded as \uXXXX escape sequences
This makes the data difficult to read in the Langfuse UI when working with non-English languages
Example: "你好" is displayed as "\u4f60\u597d"

# Expected Behavior

Chinese characters and other Unicode characters should be displayed in their original form
Users should have the option to configure the ensure_ascii parameter

# Proposed Solution
Would it be possible to:

Add a configuration option to control the ensure_ascii parameter in EventSerializer
Consider setting ensure_ascii=False as the default, or
Provide an environment variable or initialization parameter to configure this behavior

<img width="722" height="43" alt="Image" src="https://github.com/user-attachments/assets/d0444134-6932-4987-ab4a-789aff71f35f" />

nimarb:
Hi @homejim , thank you for the report ! it looks like your screenshot is taken from the trace table view, is that correct?

out of curiosity, how do you instrument langfuse ? usually, the content should be escaped client side and thus the unicode sent as `\\uXXXX`. then, we display proper characters.

homejim:
> out of curiosity, how do you instrument langfuse ? usually, the content should be escaped client side and thus the unicode sent as `\\uXXXX`. then, we display proper characters.

@nimarb Thank you for your reply!

1、Yes, my screenshot is from the traces table view. When I click into a trace, the content does display correctly. However, issues still exist regarding list display (e.g., in table rows) and search functionality (e.g., searching for non-ASCII characters).
2、We are using a self-hosted Langfuse instance. In our LangGraph integration, we report traces to Langfuse via the callback mechanism.

nimarb:
@homejim , sorry for the late reply here. could you send me an example trace to nimar at langfuse dot com ? I am interested in seeing the details here, because this issue should've been fixed (confirmed) for most users: https://github.com/langfuse/langfuse/issues/9526

what pages did you take the UI screenshots from? is the display broken in all pages or just select ones? thank you for looking into this.

Br1an67:
[...]

PR #13223, opened 2026-04-17
feat(ui): decode unicode escapes in PrettyJsonView for trace detail

toyayuto:
## What does this PR do?

Applies `decodeUnicodeEscapesOnly()` to the parsed JSON rendered by `PrettyJsonView`, so that `\uXXXX` sequences produced by the Python SDK's `json.dumps(..., ensure_ascii=True)` render as their original characters (Japanese, Chinese, Korean, emoji, ...) in the **trace detail view**.

This is a follow-up to #12882 (batch export pipeline) and #9686 (IOTableCell in trace table). Together these three changes ensure non-ASCII content ingested with `ensure_ascii=True` renders correctly everywhere a user encounters it in Langfuse — the trace list, the trace detail view, and exported CSV/JSON/JSONL files.

Refs #10972

### Motivation

The Python SDK defaults to `ensure_ascii=True` when serialising `input` / `output` / `metadata`, so payloads like `"以上"` are stored in ClickHouse as `"\u4ee5\u4e0a"`. After #9686 and #12882, these escaped sequences render correctly in the trace table and in exports, but the trace detail view (`PrettyJsonView`) still shows the raw `\uXXXX` form, which is hard to read for users of non-Latin languages.

This PR closes that remaining gap by decoding `\uXXXX` escapes recursively on the parsed JSON before it is handed to the table renderer and the raw JSON renderer.

### Changes

**`web/src/components/ui/PrettyJsonView.tsx`** (+36 / -4)

- Import `decodeUnicodeEscapesOnly` from `@/src/utils/unicode` (the shared-package re-export already used by `IOTableCell`).
- Add a small exported helper `decodeUnicodeInJson(value)` that walks the parsed JSON tree and decodes `\uXXXX` in every string leaf. Arrays and objects are traversed recursively; non-string primitives are returned as-is.
- Apply `decodeUnicodeInJson` inside the `parsedJson` `useMemo`, so all three input paths benefit:
  - `props.parsedJson` (pre-parsed by the `useParsedObservation` hook)
  - `props.json` when already an object
  - `props.json` when a string (after `deepParseJson`)
- Pass the decoded `parsedJson` to the nested `<JSONView>` (raw JSON tab) instead of `props.json`, so both the pretty view and the raw view show decoded characters.

**`web/src/components/ui/PrettyJsonView.clienttest.ts`** (new, +93)

- 10 unit tests for `decodeUnicodeInJson`: primitive passthrough, string decoding, greedy double-escape (`\\uXXXX`) form, recursive decoding through arrays / nested objects, surrogate-pair emoji, and mixed already-decoded input. Style matches the existing `unicode.clienttest.ts` (from #12882).

### Decoder behavior (for reviewers)

- Uses **greedy mode** (`decodeUnicodeEscapesOnly(value, true)`) to match `IOTableCell.tsx` — this covers both `\uXXXX` (single-escaped) and `\\uXXXX` (double-escaped) forms that can appear depending on the ingest path.
- Only `\uXXXX` escapes are touched. `\n`, `\t`, `\"`, `\\` etc. are left untouched.
- Invalid / truncated escapes are preserved literally (existing behavior of `decodeUnicodeEscapesOnly`, already covered by `web/src/utils/unicode.clienttest.ts`).
- Early-return on strings with no `\\`, so the common "already-decoded payload" case has near-zero overhead.

### Backward compatibility

- No schema or API changes.
- Payloads that do not contain `\uXXXX` sequences are unaffected (fast-path early return).
- Payloads that do contain `\uXXXX` will now display as the decoded character. This matches what `IOTableCell` and batch export already do, so the trace detail view is finally consistent with the rest of the UI.

### Notes for reviewers

- Kept the helper (`decodeUnicodeInJson`) exported from `PrettyJsonView.tsx` to minimize the change surface. If reviewers prefer, it can be lifted into `packages/shared/src/utils/unicode.ts` alongside `decodeUnicodeEscapesOnly` (happy to do this in the same PR).
- `decodeUnicodeInJson` allocates new objects/arrays only when the tree is actually traversed; for already-decoded payloads the string-level early return keeps the overhead minimal, but on very large parsed trees we still walk the tree once. If this becomes a concern I can add a short-circuit that scans the top-level string first.

## Type of change

## Mandatory Tasks

---

<h3>Greptile Summary</h3>

This PR extends the `\uXXXX` Unicode-escape decoding already applied in `IOTableCell` and batch export to the `PrettyJsonView` component, closing the last UI surface where traces ingested via Python SDK's `ensure_ascii=True` showed raw escape sequences instead of the original characters. The implementation introduces `decodeUnicodeInJson` — a recursive tree-walker that applies `decodeUnicodeEscapesOnly` (greedy mode) to every string leaf — and applies it in the `parsedJson` memo across all three input paths, including threading the decoded value through to the inner `JSONView`.

Safe to merge; the single remaining finding is a P2 edge case that does not affect the common ingestion paths.

All remaining findings are P2. The core fix is correct, consistent with the existing IOTableCell approach, and backed by a solid unit-test suite. The only gap is that object keys with \uXXXX sequences are not decoded alongside values — an unlikely edge case that still leaves behaviour better than before this PR.

web/src/components/ui/PrettyJsonView.tsx — object key decoding in decodeUnicodeInJson

| Filename | Overview |
|----------|----------|
| web/src/components/ui/PrettyJsonView.tsx | Adds decodeUnicodeInJson helper and applies it in parsedJson useMemo across all three input paths; also threads the decoded value through to JSONView. Minor gap: object keys are not decoded alongside values. |
| web/src/components/ui/PrettyJsonView.clienttest.ts | New unit tests covering strings, arrays, deep objects, surrogate pairs, greedy double-escape, mixed input, and primitive passthrough — comprehensive for the implemented behaviour. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[PrettyJsonView receives props.json / props.parsedJson] --> B{Which input path?}
    B -->|props.parsedJson provided| C[Use parsedJson directly]
    B -->|props.isParsing true| D[Return null - show loading]
    B -->|props.json is object| E[Use object directly]
    B -->|props.json is string| F[deepParseJson with size/depth limits]
    C --> G[decodeUnicodeInJson]
    E --> G
    F --> G
    G --> H[parsedJson - decoded Unicode]
    H --> I[Pretty Table View]
    H --> J["JSONView (raw tab)\nreceives parsedJson ?? props.json"]
    D --> K["JSONView fallback to props.json\n(transient loading state)"]
```

`````markdown
Path: web/src/components/ui/PrettyJsonView.tsx
Line: 96-104

Comment:
**Object keys are not decoded**

`decodeUnicodeInJson` decodes string *values* recursively but leaves object *keys* unchanged. If a double-encoded payload produces a key like `"\u4f60\u597d"` (literal backslash-u characters), the key stays escaped while its value is decoded, producing an inconsistency in the rendered output.

Decoding the key alongside the value would be straightforward:

```suggestion
  if (value !== null && typeof value === "object") {
    const result: Record<string, unknown> = {};
    for (const [k, v] of Object.entries(value as Record<string, unknown>)) {
      result[decodeUnicodeEscapesOnly(k, true)] = decodeUnicodeInJson(v);
    }
    return result;
  }
```

Keys containing `\uXXXX` sequences are rare in practice, so this won't affect most traces — but it's worth keeping the behaviour consistent with value decoding.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test(ui): add unit tests for decodeUnico..."](https://github.com/langfuse/langfuse/commit/7f189e89cfd2d10fe3483ed454289ef7397e8600) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28746814)</sub>

### CODE REVIEW

**Comment by nimarb** [comment]
@claude review

File: web/src/components/ui/PrettyJsonView.tsx [hunk 1]
```diff
@@ -82,6 +83,32 @@ const SYSTEM_TITLES = ["system", "Input"];
...
+ * structure. Used so that traces ingested via Python SDK's json.dumps(ensure_ascii=True)
+ * display non-ASCII characters (e.g. Japanese, Chinese) correctly in the trace detail
+ * view. Mirrors the approach used in IOTableCell and batch export (see PR #12882).
+ *
+ * Uses greedy mode to handle both \uXXXX (single-escaped) and \\uXXXX (double-escaped)
+ * forms that can appear depending on the ingest path.
+ */
+export function decodeUnicodeInJson(value: unknown): unknown {
+  if (typeof value === "string") {
+    return decodeUnicodeEscapesOnly(value, true);
```
nimarb, line 106: for very large jsons, will we run into "Maximum call stack size exceeded" errors here? if so, should be a loop. We also must introduce a max length after we stop processing, as we have very large jsons which could crash the browser tab potentially.

toyayuto:
@nimarb 
Thanks for the quick review — pushed two follow-up commits covering all three findings:

  1. `975c1373b` — Replaced the recursion with an explicit-stack iterative walk, and added `DECODE_UNICODE_MAX_NODES` (50,000) and `DECODE_UNICODE_MAX_DEPTH` (200) caps so very large / deeply
  nested payloads can't blow the stack or freeze the tab. Tests cover both boundaries.

  2. `2ee48ff01` — `structuredClone` the decoded payload (via a `useMemo`) before handing it to `JSONView`, so JSONView's internal `deepParseJson` can't mutate the shared `rawChildData` tree.
  Also apply `decodeUnicodeEscapesOnly` to object keys, not just values. Two new tests for key decoding.

  14/14 PrettyJsonView unit tests pass locally. Happy to address anything else that comes up!

PR #13225, opened 2026-04-17
fix(evals): fix score filtering on evaluator runs page

wochinge:
## Summary

- Remove broken score filter for now.
- This is currently considered the best solution as score data is stored in ClickHouse and job executions in Postgres. It's also already on the roadmap to move the job execution data to postgres as well. Unfortunately there is currently no way to filter across both databases in an reliable and performant way


web/src/features/evals/server/router.ts
```diff
@@ -1604,7 +1608,6 @@ const generateExecutionsQuery = (
    ${select}
    FROM job_executions je
    LEFT JOIN traces t ON je.job_input_trace_id = t.id AND je.project_id = t.project_id
```
marliessophie: Traces no longer live in postgres, let's remove this.

web/src/features/evals/server/router.ts
marliessophie: We'll need to drop this & UI column too.

web/src/features/evals/server/router.ts
```diff
@@ -1346,8 +1346,12 @@ export const evalRouter = createTRPCRouter({
         scope: "evalJobExecution:read",
       });
 
+      // Strip scoreValue filters — the score filter was removed from the UI
+      // but bookmarked URLs may still include it.
+      const filters = input.filter.filter((f) => f.column !== "scoreValue");
```
marliessophie: Pls at least define constant list of "deprecated filter columns"

web/src/features/filters/config/eval-logs-config.ts
marliessophie: Pls rm, will not work as requires tracing data

web/src/server/api/definitions/evalExecutionsTable.ts
marliessophie: Pls rm, requires traces data (CH)

marliessophie (commented):
Few more lines of code to remove, then done

marliessophie (approved):
LGTM 🚀

PR #13274, opened 2026-04-21
fix(events): use release field for trace release in events table adapter

### LINKED ISSUE #13273: bug: eventsToTraceAdapter copies version into release field for SDK v5+ traces
## Summary

When using `@langfuse/otel` v5.x, Langfuse routes ingestion through the **direct events-table write path** (triggered when the `x-langfuse-sdk-version` header is `>= 5.0.0` for JS). In this path, trace metadata is synthesized client-side from the events table by `adaptEventsToTraceFormat` in `web/src/features/events/lib/eventsToTraceAdapter.ts`.

On **line 84**, both `release` and `version` are set from the same source field:

```typescript
release: earliest.version ?? null,   // ← BUG: should be earliest.release
version: earliest.version ?? null,
```

This means `trace.release` always shows the `langfuse.version` span attribute value instead of the `langfuse.release` value, making release-based filtering and analytics in the Langfuse UI incorrect for all SDK v5+ users who set both fields.

## Root cause

`release` and `version` are **separate** OTel span attributes (`langfuse.release` and `langfuse.version`) and separate columns in the ClickHouse events table. `OtelIngestionProcessor.ts` correctly maps them independently. But `eventsToTraceAdapter.ts` never reads `earliest.release` — it only reads `earliest.version` for both fields.

Additionally, `release` was not included in the `base` / `baseWithoutTools` / `byIdBase` field sets in `EventsQueryBuilder`, so the column was never selected from ClickHouse even though it is defined in `EVENTS_FIELDS` and exists in the events table schema.

## Affected versions

All users of `@langfuse/otel` >= 5.0.0 or any SDK that sends `x-langfuse-sdk-version >= 5.0.0` with both `langfuse.release` and `langfuse.version` span attributes.

## Fix

Three-file change (see linked PR):

1. `eventsToTraceAdapter.ts` line 84: `release: earliest.release ?? null`
2. `EventsObservationSchema` in `domain/observations.ts`: add `release: z.string().nullable().optional()`
3. `event-query-builder.ts`: add `"release"` to `base`, `baseWithoutTools`, and `byIdBase` field sets so the column is actually selected from ClickHouse

AlissonRS:
## Summary

Fixes #13273

When using `@langfuse/otel` v5.x (SDK v5+ direct events-table write path), `adaptEventsToTraceFormat` was setting both `release` and `version` on the synthetic trace from `earliest.version`:

```typescript
// Before (buggy)
release: earliest.version ?? null,
version: earliest.version ?? null,
```

This overwrote the correct `langfuse.release` span attribute value with the `langfuse.version` value, making release-based filtering and analytics wrong for all SDK v5+ users who set both fields.

The fix is deeper than just the adapter line: `release` was never actually selected from ClickHouse (missing from all `FIELD_SETS`), not in the TypeScript record schema (`eventsObservationRecordReadSchema`), and not mapped in the observation converter (`convertEventsObservation`).

## Changes

**`web/src/features/events/lib/eventsToTraceAdapter.ts`**
- Line 84: `earliest.version` → `earliest.release` for the `release` field

**`packages/shared/src/domain/observations.ts`**
- Added `release: z.string().nullable().optional()` to `EventsObservationSchema` so the TypeScript type carries the field through to the adapter

**`packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts`**
- Added `"release"` to `base`, `baseWithoutTools`, and `byIdBase` field sets in `FIELD_SETS` so the column is actually selected from ClickHouse (`e.release as release` was defined in `EVENTS_FIELDS` but never included in any field set used for trace/observation queries)

**`packages/shared/src/server/repositories/definitions.ts`**
- Added `release: z.string().nullish()` to `eventsObservationRecordReadSchema` so the field is present in `EventsObservationRecordReadType` and not stripped by the schema

**`packages/shared/src/server/repositories/observations_converters.ts`**
- Added `release: record.release ?? null` mapping in `convertEventsObservation` so the field flows from raw ClickHouse record to the domain `EventsObservation` object

**`web/src/__tests__/server/repositories/event-repository.servertest.ts`**
- Added test confirming `release` is returned in the result set from `getObservationsWithModelDataFromEventsTable`
- Added test confirming `release` is returned when fetching an observation by id via `getObservationByIdFromEventsTable`

## Test plan

- [ ] Set `langfuse.release = "1.0.0"` and `langfuse.version = "V2"` span attributes (via `LangfuseSpanProcessor({ release: "1.0.0" })` + `propagateAttributes({ version: "V2" })`)
- [ ] Send a trace using `@langfuse/otel` >= 5.0.0 (triggers events-table path)
- [ ] Open the trace in Langfuse UI → confirm `release = "1.0.0"` and `version = "V2"` are shown independently
- [ ] Filter traces by release in the traces table → confirm the filter works correctly


Steffen911:
@AlissonRS Thank you a lot for this contribution! Could you sign the CLA?
And in addition, do you think you could add a test to the event-repository.servertest.ts that confirms whether the `release` property is part of the result set after your changes?

AlissonRS:
Hi @Steffen911 thanks for the quick response!

I signed the CLA, added a couple tests, and had to fix a couple other locations to properly return the `release` field and make the tests pass, thanks for calling that out.

Steffen911:
@AlissonRS Thank you a lot! I think we're getting close. 
It seems that there are some requests that don't expect the `release` property in their body. I think there are a few zod definitions that may require an optional release property. Adding that to the expected API response body should work well for the test cases that fail.

AlissonRS:
Hi @Steffen911 just got back to this today, sorry for being absent. Could you check again?

Steffen911:
Thank you so much @AlissonRS !

PR #13309, opened 2026-04-22
fix(observations): resolve `t-<traceId>` ids on the classic read path

wochinge:
## Summary

Fixes [LFE-9379](https://linear.app/langfuse/issue/LFE-9379/ui-bug-incorrect-source-observation-id-when-adding-trace-to-dataset).

## Failure path

1. User traces with v3 SDK → we insert a `t-` observation in the events table as the replacement for the v3 root trace.
2. User uses v4 in the UI and adds the item to the dataset (which stamps the `t-` id into `sourceObservationId`).
3. User then uses the v3 SDK to retrieve the dataset and access the `sourceObservationId` (`t-…`) → since it's the old SDK, it uses the old tables (`useEventsTable: false`) that of course don't have it → 404.

## Fix

Teach the classic read helpers (`getObservationById`, `checkObservationExists`, `getTraceIdsForObservations`) to fall back to the events reader when the classic lookup misses **and** the id is `t-`-prefixed. Classic-first preserves user-ingested real spans whose id happens to start with `t-`; the events fallback only fires on a genuine miss.

- `TRACE_OBSERVATION_ID_PREFIX` + `isTraceObservationId` introduced in `definitions.ts`, adopted in `convertTraceToStagingObservation` and the two worker backfill sites that previously used the literal.
- New `getTraceIdsForObservationIdsFromEventsTable` helper for the plural-lookup fallback.
- Dead code removed: `getObservationsById`, `getLatencyAndTotalCostForObservations` (singular) — zero callers repo-wide.

## Why we didn't touch the write path

We considered preventing the UI from stamping `t-<traceId>` into `source_observation_id` in the first place (and backfilling existing rows to `NULL`). We deliberately kept the write path as-is for two reasons:

- **Avoids a data migration.** No backfill needed for existing `dataset_item` rows with `source_observation_id LIKE 't-%'`; they resolve correctly through the fixed read path.
- **More robust across surfaces.** Other features also let users interact with the synthetic trace-as-span node (e.g. scoring on it produces a score with `observation_id = t-<traceId>`). Fixing the read path uniformly means those cross-cutting cases keep their linkage instead of being quietly stripped.

---

<h3>Greptile Summary</h3>

This PR fixes a 404 regression where v3 SDK clients calling `GET /api/public/observations/t-<traceId>` on the classic read path got a not-found error because the synthetic `t-` observation only lives in the events table. The fix teaches `getObservationById`, `checkObservationExists`, and `getTraceIdsForObservations` to fall back to the events table when the classic lookup misses and the ID is `t-`-prefixed, preserving classic-first priority so user-ingested spans that happen to start with `t-` are returned without fallback. Dead code (`getObservationsById`, `getLatencyAndTotalCostForObservations`) is also cleaned up.

Safe to merge — fix is narrowly scoped, classic-first ordering is preserved, and all three read helpers have new test coverage.

No P0/P1 findings. The fallback only fires on a genuine miss for `t-`-prefixed IDs, so no regression risk for existing observation lookups. Dead code removal has zero callers. Tests cover the resolution path, the 404 path, and the classic-priority edge case.

| Filename | Overview |
|----------|----------|
| packages/shared/src/server/repositories/definitions.ts | Introduces `TRACE_OBSERVATION_ID_PREFIX` constant and `isTraceObservationId` helper; replaces the hardcoded `t-` literal in `convertTraceToStagingObservation`. |
| packages/shared/src/server/repositories/events.ts | Adds `getTraceIdsForObservationIdsFromEventsTable` and `checkObservationExistsInEventsTable`; `OBSERVATIONS_TO_TRACE_INTERVAL` import correctly moved to top of file. |
| packages/shared/src/server/repositories/observations.ts | Classic-first fallback added to `checkObservationExists`, `getObservationById`, and `getTraceIdsForObservations`; dead code (`getObservationsById`, `getLatencyAndTotalCostForObservations`) removed. |
| web/src/__tests__/server/observation-api.servertest.ts | Adds three new server tests covering: synthetic `t-<traceId>` resolution, 404 on missing synthetic id, and classic-first priority for user-ingested `t-`-prefixed spans. |
| web/src/__tests__/server/repositories/event-repository.servertest.ts | New tests for `getTraceIdsForObservationIdsFromEventsTable` and `checkObservationExistsInEventsTable`, including the cross-midnight lookback window case. |
| web/src/__tests__/server/repositories/observation-repository.servertest.ts | Adds repository-level tests for synthetic ID resolution, missing-ID no-data-invention, and cross-midnight existence checks. |
| worker/src/backgroundMigrations/backfillExperimentsHistoric.ts | Replaces hardcoded `t-${dri.trace_id}` with `${TRACE_OBSERVATION_ID_PREFIX}${dri.trace_id}` to use the shared constant. |
| worker/src/features/eventPropagation/handleExperimentBackfill.ts | Same constant adoption as `backfillExperimentsHistoric.ts`; no functional change. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client as v3 SDK Client
    participant API as GET /observations/:id
    participant Classic as observations table (classic)
    participant Events as events_full table

    Client->>API: GET /observations/t-<traceId>?useEventsTable=false
    API->>Classic: getObservationByIdInternal(id)
    Classic-->>API: [] (not found — t- id lives in events)
    API->>API: mapped.length === 0 && isTraceObservationId(id)?
    Note over API: YES → fall back
    API->>Events: getObservationByIdFromEventsTable(id)
    Events-->>API: observation record
    API-->>Client: 200 OK

    Note over Client,Events: Classic-first: if a real span with t- prefix exists it is returned before fallback fires
```

<sub>Reviews (1): Last reviewed commit: ["fix(observations): resolve \`t-&lt;traceId&gt;\`..."](https://github.com/langfuse/langfuse/commit/6a96bc9285c29381a894625e140bfa1ea513a669) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29285519)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/definitions.ts
+++ b/packages/shared/src/server/repositories/definitions.ts
@@ -1,6 +1,17 @@
 import z from "zod";
 import { DEFAULT_TRACE_ENVIRONMENT } from "../ingestion/types";
 
+/**
+ * Prefix the ingestion pipeline adds to a trace id when representing the trace
+ * as a synthetic SPAN observation in the events pipeline (see
+ * `convertTraceToStagingObservation`). Read-side lookups of `t-<traceId>` must
+ * resolve against the events table, not the classic observations table.
+ */
+export const TRACE_OBSERVATION_ID_PREFIX = "t-" as const;
+
+export const isTraceObservationId = (id: string): boolean =>
+  id.startsWith(TRACE_OBSERVATION_ID_PREFIX);
+
 export const clickhouseStringDateSchema = z
   .string()
   // clickhouse stores UTC like '2024-05-23 18:33:41.602000'
@@ -385,7 +396,7 @@ export const convertTraceToStagingObservation = (
 ): ObservationBatchStagingRecordInsertType => {
   return {
     // Identity - trace acts as its own span. Modify traceId to avoid cases where users set spanId = traceId.
-    id: `t-${traceRecord.id}`,
+    id: `${TRACE_OBSERVATION_ID_PREFIX}${traceRecord.id}`,
     trace_id: traceRecord.id,
     project_id: traceRecord.project_id,
 
--- a/packages/shared/src/server/repositories/events.ts
+++ b/packages/shared/src/server/repositories/events.ts
@@ -668,7 +668,7 @@ export const getTraceIdsForObservationIdsFromEventsTable = async ({
       SELECT
         span_id as id,
         trace_id
-      FROM events_full
+      FROM events_core
       WHERE project_id = {projectId: String}
         AND span_id IN ({observationIds: Array(String)})
       ORDER BY event_ts DESC
@@ -702,7 +702,7 @@ export const checkObservationExistsInEventsTable = async ({
   const rows = await queryClickhouse<{ id: string }>({
     query: `
       SELECT span_id as id
-      FROM events_full
+      FROM events_core
       WHERE project_id = {projectId: String}
         AND span_id = {observationId: String}
         ${startTime ? `AND start_time >= {startTime: DateTime64(3)} - ${OBSERVATIONS_TO_TRACE_INTERVAL}` : ""}
--- a/packages/shared/src/server/repositories/events.ts
+++ b/packages/shared/src/server/repositories/events.ts
@@ -75,6 +75,7 @@ import { UiColumnMappings } from "../../tableDefinitions";
 import { eventsTableCols } from "../../eventsTable";
 import { tracesTableCols } from "../../tableDefinitions/tracesTable";
 import { parseMetadataCHRecordToDomain } from "../utils/metadata_conversion";
+import { OBSERVATIONS_TO_TRACE_INTERVAL } from "./constants";
 
 type ObservationsTableQueryResultWitouhtTraceFields = Omit<
   ObservationsTableQueryResult,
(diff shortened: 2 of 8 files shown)


packages/shared/src/server/repositories/definitions.ts
```diff
@@ -1,6 +1,17 @@
 import z from "zod";
 import { DEFAULT_TRACE_ENVIRONMENT } from "../ingestion/types";
 
+/**
+ * Prefix the ingestion pipeline adds to a trace id when representing the trace
+ * as a synthetic SPAN observation in the events pipeline (see
+ * `convertTraceToStagingObservation`). Read-side lookups of `t-<traceId>` must
+ * resolve against the events table, not the classic observations table.
+ */
+export const TRACE_OBSERVATION_ID_PREFIX = "t-" as const;
```
wochinge, line 10: just to avoid the magic string

packages/shared/src/server/repositories/observations.ts
```diff
@@ -385,58 +407,6 @@ export const getObservationById = async ({
   return mapped.shift();
 };
```
wochinge, line 387: was unsued

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1518,12 +1471,6 @@ export const getObservationMetricsForPrompts = async (
     params: {
       projectId,
       promptIds,
-      ...(fromTimestamp
-        ? { fromTimestamp: convertDateToClickhouseDateTime(fromTimestamp) }
-        : {}),
-      ...(toTimestamp
-        ? { toTimestamp: convertDateToClickhouseDateTime(toTimestamp) }
```
wochinge, line 1525: was unsued

holyweeper (approved):

packages/shared/src/server/repositories/events.ts
```diff
@@ -651,6 +652,82 @@ export const getObservationByIdFromEventsTable = async ({
...
+  const rows = await queryClickhouse<{ id: string; trace_id: string }>({
+    query: `
+      SELECT
+        span_id as id,
+        trace_id
+      FROM events_core
+      WHERE project_id = {projectId: String}
+        AND span_id IN ({observationIds: Array(String)})
+      ORDER BY event_ts DESC
+      LIMIT 1 BY id
```
marliessophie, line 675: Did you test this query on large projects? There is a bloom filter on span_id, but unsure this suffices. Ideally we can filter on trace_id and start_time range.  

Generally: We'll want to use `EventsQueryBuilder` for all events queries. Also please check if there is no pre-exiting query we can use. I'd imagine this already exists.

packages/shared/src/server/repositories/events.ts
```diff
@@ -651,6 +652,82 @@ export const getObservationByIdFromEventsTable = async ({
...
+}): Promise<boolean> => {
+  const rows = await queryClickhouse<{ id: string }>({
+    query: `
+      SELECT span_id as id
+      FROM events_core
+      WHERE project_id = {projectId: String}
+        AND span_id = {observationId: String}
+        ${startTime ? `AND start_time >= {startTime: DateTime64(3)} - ${OBSERVATIONS_TO_TRACE_INTERVAL}` : ""}
+      ORDER BY event_ts DESC
+      LIMIT 1 BY id
```
marliessophie, line 710: We'll want to use `EventsQueryBuilder` for all queries. Also please check if there is no pre-exiting query we can use. I would imagine we already have such a query?

packages/shared/src/server/repositories/observations.ts
```diff
@@ -93,7 +97,14 @@ export const checkObservationExists = async (
     },
   });
 
-  return rows.length > 0;
+  if (rows.length > 0) return true;
+  return isTraceObservationId(id)
+    ? checkObservationExistsInEventsTable({
```
marliessophie, line 102: We'll need to check env variable (you use it in tests), to not acidently hit a case where a self hoster happens to run into this (only if they have such id format by coincidence), such that the app does not crash bc we try to read from events table that does not exist.

packages/shared/src/server/repositories/observations.ts
```diff
@@ -371,6 +382,17 @@ export const getObservationById = async ({
     );
   });
   if (mapped.length === 0) {
+    if (isTraceObservationId(id)) {
+      return getObservationByIdFromEventsTable({
```
marliessophie, line 386: We'll need to check env variable (you use it in tests), to not acidently hit a case where a self hoster happens to run into this (only if they have such id format by coincidence), such that the app does not crash bc we try to read from events table that does not exist.

packages/shared/src/server/repositories/observations.ts
```diff
@@ -1797,10 +1701,20 @@ export const getTraceIdsForObservations = async (
...
-  return rows.map((row) => ({
-    id: row.id,
-    traceId: row.trace_id,
-  }));
+  const found = new Set(rows.map((r) => r.id));
+  const missingSyntheticIds = observationIds.filter(
+    (id) => !found.has(id) && isTraceObservationId(id),
+  );
+
+  const syntheticFallbacks = await getTraceIdsForObservationIdsFromEventsTable({
```
marliessophie, line 1709: We'll need to check env variable (you use it in tests), to not acidently hit a case where a self hoster happens to run into this (only if they have such id format by coincidence), such that the app does not crash bc we try to read from events table that does not exist.

marliessophie (commented):
I'm unsure how to best handle such cases actually. I think directionally there are a couple options 
- Option 1: read observations with trace prefix schema from events table (your recommendation) 
- Option 2: strip the prefix and attempt to read from traces table instead 
- other? 

Personally I'd prefer if we swiftly align a couple mins sync with steffen (or valeriy) on approach. Would you like to shoot either a message or should I? 

Generally we should document such behaviour -- both internally and externally. It's unlikely, though not impossible, that some user is sending us custom id's that are prefixed with `t-`.

PR #13342, opened 2026-04-24
fix(otel): serialize int64 fields as strings, not protobufjs Long internals

### LINKED ISSUE #13295: bug: OTel ingestion serializes int64 as protobufjs Long internals after protobufjs 7.4 → 7.5 upgrade
### Describe the bug

After upgrading to 3.169.0 from 3.155.1 (specifically PR #13232 which bumped protobufjs from 7.4.0 to 7.5.5), all `int64` fields in OTLP/protobuf ingestion (e.g. intValue, startTimeUnixNano, endTimeUnixNano, timeUnixNano) are serialized as `protobufjs` internal Long objects (`{"low": <int>, "high": <int>, "unsigned": <bool>}`) instead of the previous numeric form. Anything downstream that reads the S3-staged OTLP payload (most notably the server-side ingestion masking callback) receives this shape and breaks if it expects OTLP/JSON-conformant int64 (decimal string per spec, or at least a number).

The endpoint code at `web/src/pages/api/public/otel/v1/traces/index.ts` calls `.toObject(parsed)` without a longs option:

```javascript
  resourceSpans = $root.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest
    .toObject(parsed)
    .resourceSpans;
```

In protobufjs 7.4, `util.Long` was not initialized in our build (no long.js resolvable from the protobufjs install location), so `int64` fell back to plain JS number. In protobufjs 7.5, an internal fallback Long class is created when long.js cannot be resolved, but that class does not have a `toJSON()` method — so `JSON.stringify` emits the raw enumerable properties {low, high, unsigned} and they get persisted to S3 / forwarded to the masking callback verbatim.

This is a silent regression: spans still ingest, but downstream consumers that strictly type-check int64 (e.g. pydantic `int`) start failing.

### Steps to reproduce

Reproducible directly inside the official images (no langfuse setup required):
```bash
  docker run --rm --entrypoint sh langfuse/langfuse:3.155.1 -c 'cd /app/web && node -e "
  const pb = require(\"protobufjs\");
  const root = pb.Root.fromJSON({ nested: { T: { fields: { v: { type: \"int64\", id: 1 } } } } });
  const T = root.lookupType(\"T\");
  const dec = T.decode(T.encode({ v: 1234567890 }).finish());
  console.log(JSON.stringify(T.toObject(dec)));
  "'
  # → {"v":1234567890}

  docker run --rm --entrypoint sh langfuse/langfuse:3.169.0 -c 'cd /app && node -e "
  const pb = require(\"/app/node_modules/.pnpm/protobufjs@7.5.5/node_modules/protobufjs\");
  const root = pb.Root.fromJSON({ nested: { T: { fields: { v: { type: \"int64\", id: 1 } } } } });
  const T = root.lookupType(\"T\");
  const dec = T.decode(T.encode({ v: 1234567890 }).finish());
  console.log(JSON.stringify(T.toObject(dec)));
  "'
  # → {"v":{"low":1234567890,"high":0,"unsigned":false}}
```

- End-to-end repro:
  1. Configure server-side ingestion masking (LANGFUSE_INGESTION_MASKING_CALLBACK_URL=...) on a 3.169.x deployment with EE license.
  2. Send a trace via OTLP/protobuf (e.g. OpenTelemetry Java agent or any Python SDK using OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf or grpc).
  3. Inspect the body received by the masking callback — intValue and timestamps arrive as {low,high,unsigned} rather than number/string.

  We bisected our own ingestion against the rolling-upgrade timestamp (S3-staged OTel payload bucket) and confirmed the transition is exactly the 3.155.1 → 3.169.0 upgrade for the same SDK and same scope (no client-side change
  involved).

Self-hosted

3.169.0 (regressed); 3.155.1 (last known good)

  - Affected reproductions observed with: OpenTelemetry Java SDK 1.25.0 (with org.springframework.boot 3.1.12 instrumentation, OTLP/protobuf), and langfuse-sdk Python 3.12.0 (telemetry sdk 1.39.1).
  - Bisect on a single project shows that scope/SDK identity remained constant across the regression — only the langfuse server image changed.

### Additional information

_No response_

Yes

he-yufeng:
Fixes langfuse/langfuse#13295.

## Bug

After [#13232](<https://github.com/langfuse/langfuse/issues/13232>) bumped protobufjs from 7.4.0 → 7.5.5, every \`int64\` field in OTLP/protobuf ingestion (\`intValue\`, \`startTimeUnixNano\`, \`endTimeUnixNano\`, \`timeUnixNano\`) started serializing as the protobufjs internal Long shape:

\`\`\`json<br>{"low": 1234567890, "high": 0, "unsigned": false}<br>\`\`\`

Anything downstream that reads the S3-staged OTLP payload (most notably the server-side ingestion masking callback) receives that shape and breaks when it expects OTLP/JSON-conformant int64 (decimal string per spec, or at least a number).

## Root cause

The endpoint at \`web/src/pages/api/public/otel/v1/traces/index.ts:96\` calls \`toObject(parsed)\` with no \`longs\` option.

* On 7.4, \`util.Long\` was never initialized in our build (long.js wasn't resolvable from the protobufjs install location), so \`int64\` silently fell back to plain JS number.
* On 7.5, an internal fallback Long class is created when long.js is missing, and that class doesn't implement \`toJSON()\`. \`JSON.stringify\` then emits the raw enumerable \`{low, high, unsigned}\` properties verbatim.

Silent regression: spans still ingest, strict int64 consumers start rejecting them.

## Fix

Pass \`{ longs: String }\` to \`toObject\` so \`int64\` fields come out as decimal strings. Matches the OTLP/JSON spec and is safe for both the masking callback and the downstream writer (string form is number-compatible and parses cleanly into BigInt when needed).

## Impacted package

* \`web\`

## Verification

Reporter's repro inside the official image (no langfuse setup) shows the regression starting from the protobufjs bump. I don't have a local pnpm/ClickHouse environment to run \`pnpm --filter web run lint\` + targeted web tests end-to-end — happy to iterate if a reviewer prefers \`{ longs: Number }\` (only safe up to \`Number.MAX_SAFE_INTEGER\`, so strings are more correct for nano-second timestamps).

### Greptile Summary

This PR fixes a silent regression introduced by the protobufjs 7.4→7.5 bump ([#13232](<https://github.com/langfuse/langfuse/issues/13232>)), where `int64` fields in OTLP/protobuf ingestion began serializing as raw Long internals (`{low, high, unsigned}`) instead of usable values. The fix passes `{ longs: String }` to `toObject()`, which matches the OTLP/JSON spec requirement of decimal strings for int64 fields like `startTimeUnixNano`, `endTimeUnixNano`, and `intValue`.

### Confidence Score: 4/5

Safe to merge — the fix is minimal, targeted, and spec-compliant; the only risk is downstream consumers that previously relied on numeric int64 values.

Single-line change with a well-justified option; `{ longs: String }` is the correct OTLP/JSON representation and avoids precision loss for nanosecond timestamps that exceed Number.MAX_SAFE_INTEGER. One P2 concern: prior to protobufjs 7.5 the output was plain JS numbers, so downstream code that does arithmetic on timestamp fields without parsing them as strings first may silently break — worth verifying.

Downstream consumers of `resourceSpans` in `OtelIngestionProcessor` and the ClickHouse writer — verify they handle string-typed int64 fields for timestamps and `intValue`.

<details><summary>Important Files Changed</summary>

| Filename | Overview |
| -- | -- |
| web/src/pages/api/public/otel/v1/traces/index.ts | Adds `{ longs: String }` to `toObject()` call to fix int64 serialization regression from protobufjs 7.4→7.5 bump; correct and spec-compliant, though downstream consumers should be verified for string handling |

</details>

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
    participant OtelEndpoint as otel/v1/traces/index.ts
    participant Protobufjs as protobufjs toObject()
    participant Processor as OtelIngestionProcessor

    Client->>OtelEndpoint: POST (application/x-protobuf)
    OtelEndpoint->>Protobufjs: decode(body)
    Protobufjs-->>OtelEndpoint: parsed message
    OtelEndpoint->>Protobufjs: toObject(parsed, { longs: String })
    Note over Protobufjs: int64 fields emitted as<br/>decimal strings (spec-compliant)<br/>e.g. "1700000000000000000"
    Protobufjs-->>OtelEndpoint: resourceSpans (int64 as string)
    OtelEndpoint->>Processor: publishToOtelIngestionQueue(resourceSpans)
```

Prompt To Fix All With AI

```markdown
Path: web/src/pages/api/public/otel/v1/traces/index.ts
Line: 105

Comment:
**Downstream consumers now receive strings, not numbers, for int64 fields**

Before protobufjs 7.5, `toObject` with no `longs` option silently emitted plain JS numbers for int64 fields. With `{ longs: String }` those same fields (`startTimeUnixNano`, `endTimeUnixNano`, `timeUnixNano`, `intValue`, etc.) are now decimal strings. Any downstream code in `OtelIngestionProcessor` or the ClickHouse writer that does arithmetic, `typeof x === 'number'` guards, or numeric comparisons on these fields without first parsing them will silently produce `NaN` or wrong results.

The OTLP/JSON spec mandates strings for int64, so `{ longs: String }` is the correct long-term representation — worth verifying that the processor/writer path handles string timestamps consistently.

```

Reviews (1): Last reviewed commit: ["fix(otel): serialize int64 fields as str..."](<https://github.com/langfuse/langfuse/commit/166127b01b308fe9494b088897eac99b8808d48c>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=29585128>)


Steffen911:
@he-yufeng Could you add a test-case and sign the CLA? Otherwise, we cannot merge this.

he-yufeng:
Added a focused regression test for the protobuf path.

What changed:
- rebuilds the PR on current `main` so the diff stays small
- keeps OTLP protobuf `int64` fields serialized as decimal strings with `{ longs: String }`
- adds a Vitest case covering fixed64 timestamps so they do not leak protobufjs `{ low, high, unsigned }` internals into JSON

Verified locally:
- `prettier --write web/src/pages/api/public/otel/v1/traces/index.ts web/src/__tests__/otel-protobuf.clienttest.ts`
- `vitest run --project client src/__tests__/otel-protobuf.clienttest.ts` (1 passed)
- `eslint src/__tests__/otel-protobuf.clienttest.ts src/pages/api/public/otel/v1/traces/index.ts --max-warnings 0`
- `git diff --check`

Steffen911:
@he-yufeng You will also need to sign the CLA for this to be mergeable

he-yufeng:
The CLA check is now passing, and the focused protobuf regression test was added in the latest branch update. The current status contexts I can see are all green (license/cla, Snyk license, Snyk security). This should be ready for another review when you have time.

he-yufeng:
Gentle bump — large int64 OTEL attributes lose precision going through protobufjs here. It is a small serialization fix; I can add a regression test if that would help move it along.

Steffen911:
@claude review

he-yufeng:
Both should be in place now: the test case is in (4b77bc51, "should stage protobuf int64 fields as OTLP/JSON decimal strings" in otel-api.servertest.ts, covers the int64 decimal-string staging end to end through S3), and the CLA check shows signed on the PR. Let me know if you'd like the test shaped differently.

PR #13383, opened 2026-04-27
chore(ci): disable test sharding + speed up test suite

wochinge:
## Summary

- Removes the `tests-web` shard matrix from `.github/workflows/pipeline.yml`.
- Runs the web server Vitest project without `--shard` so each remaining matrix cell executes the full server suite.
- Only downloads headless chrome -> speed up e2e tests
- only verify types fast -> speed up indexes
- general test speed ups

## Why

This lets us compare CI behavior and runtime without the current web test sharding split.

---

<h3>Greptile Summary</h3>

This PR removes the 3-shard matrix from `tests-web` (collapsing 18 CI jobs down to 6 per push), speeds up builds by skipping duplicate Next.js TypeScript checks (covered by a new `typecheck` step in the lint job), narrows Playwright to the chromium shell only, and adds a custom `VitestCiReporter` that surfaces slow tests and retries. Alongside the CI changes, a large batch of server test files is refactored for parallel isolation: hardcoded seed project IDs are replaced with `createOrgProjectAndApiKey()`, sequential awaits are converted to `Promise.all`, and several `beforeEach` blocks promoting expensive setup are lifted to `beforeAll`.

<h3>Confidence Score: 4/5</h3>

Safe to merge; all findings are P2 quality/style concerns with no correctness-breaking changes on the critical path.

No P0/P1 issues found. The three P2 observations are: a fragile test-name guard in traces-api, blanket retry:3 that can mask flakiness, and the bcrypt sentinel in org-factory which is well-commented but warrants awareness. The CI structural changes (sharding removal, typecheck separation, chromium-only playwright) are sound.

packages/shared/src/server/test-utils/org-factory.ts (bcrypt path change) and web/src/__tests__/server/traces-api.servertest.ts (currentTestName guard).

| Filename | Overview |
|----------|----------|
| .github/workflows/pipeline.yml | Removes 3-way shard matrix (3→1 CI runs per pg/mode combination), adds NEXT_IGNORE_BUILD_ERRORS to build steps, narrows Playwright install to chromium shell only, and moves typecheck into the lint job. |
| packages/shared/src/server/test-utils/org-factory.ts | Switches test fixture API key from bcrypt (hashSecretKey) to SHA fast-hash, leaving hashedSecretKey as a non-bcrypt sentinel — legacy auth path no longer usable through this factory. |
| scripts/vitest/ci-reporter.ts | New custom Vitest reporter that prints slowest tests, slowest files (summed durations), and retried tests at end of run. |
| web/vitest.config.mts | Adds VitestCiReporter and retry: 3 on CI; retry value masks transient flakiness but is a common pattern for integration test suites. |
| turbo.json | Disables Turbo cache for db:generate (Prisma writes to node_modules, not restored by cache replays) and adds ^build dependency to typecheck. |
| web/next.config.mjs | Adds typescript.ignoreBuildErrors opt-in via NEXT_IGNORE_BUILD_ERRORS env var, controlled from CI only; production builds are unaffected. |
| web/src/__tests__/server/api-auth.servertest.ts | Legacy bcrypt tests now create dedicated keys with hashSecretKey instead of relying on createOrgProjectAndApiKey; TTL-expiry test replaces a 5-second sleep with redis.expire() for speed. |
| worker/src/__tests__/backfillValidToForDatasetItems.test.ts | Replaces single-char project IDs ("A","B") with valid UUID constants that sort lexicographically before random UUIDs, preventing interference when tests run in parallel. |
| web/src/__tests__/server/traces-api.servertest.ts | Many parallelisation improvements and beforeAll migrations; contains a fragile currentTestName.includes guard to skip outer beforeEach for an inner describe block. |
| worker/vitest.config.ts | Adds VitestCiReporter and retry: 3 on CI, matching the web vitest config. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Push / PR] --> B[pre-job]
    B --> C[lint job\n+ typecheck]
    B --> D[tests-web\nmatrix: node×pg×mode\n6 jobs total]
    B --> E[e2e tests\nchromium shell only]
    B --> F[other build jobs]

    C -->|pnpm run typecheck| G{TS errors?}
    G -- yes --> H[❌ Blocked]
    G -- no --> I[✅ Pass]

    D -->|build with\nNEXT_IGNORE_BUILD_ERRORS=true| J[vitest run --project server\nfull suite, no shard]
    D -->|deploy-mode == ''| K[vitest run --project client]

    subgraph Before_18_CI_jobs
        direction LR
        X1[mode x pg x shard 1/3]
        X2[mode x pg x shard 2/3]
        X3[mode x pg x shard 3/3]
    end

    subgraph After_6_CI_jobs
        direction LR
        Y1[mode x pg full suite]
    end
```

`````markdown
Path: web/src/__tests__/server/traces-api.servertest.ts
Line: 582-586

Comment:
**Fragile test-name guard in `beforeEach`**

Checking `currentTestName.includes("Advanced Filtering - Dual Path Tests")` to conditionally skip setup couples the lifecycle hook to a hardcoded string. If the describe block is renamed or moved, the guard silently stops matching and the outer `beforeEach` will create redundant (unused) fixtures for every inner test.

A cleaner approach is to move the outer `beforeEach` fixture setup into the outer `beforeAll`, or use a nested `describe` wrapper with its own lifecycle, so the `"Advanced Filtering"` block can declare `beforeAll` without needing to disable the parent hook by name.

---

Path: web/vitest.config.mts
Line: 13

Comment:
**Blanket `retry: 3` can mask persistent flakiness**

Setting `retry: 3` globally on CI means any test that fails three times in a row is still reported as a failure, but tests that fail 1-2 times are silently marked "flaky" and the run passes. The new `VitestCiReporter` does surface retried tests in the log, which helps, but callers that only look at the exit code won't notice growing flakiness until it becomes consistent. Consider starting with `retry: 1` to get the "don't fail on one transient error" benefit while keeping clearer signal about test health. The same concern applies to `worker/vitest.config.ts`.

---

Path: packages/shared/src/server/test-utils/org-factory.ts
Line: 52-55

Comment:
**Legacy bcrypt path no longer exercised by `createOrgProjectAndApiKey`**

`hashedSecretKey` is now set to a non-bcrypt sentinel (`test-hashed-secret-key-${v4()}`), so any test that calls `createOrgProjectAndApiKey()` and then exercises the legacy upgrade-on-first-use path (where the system falls back to bcrypt when `fastHashedSecretKey` is null) will silently get wrong behaviour — the bcrypt comparison will always fail for these keys. `api-auth.servertest.ts` was updated to create explicit legacy keys, but other tests that happen to rely on the bcrypt path through this factory will break in a non-obvious way. The comment in the code acknowledges this, but it's worth ensuring all callers have been audited.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["ci: run prisma generate without turbo ca..."](https://github.com/langfuse/langfuse/commit/c391e4f12b097237373671685fa8891ca355048d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30199198)</sub>


wochinge:
@claude review

PR #13453, opened 2026-05-04
feat(storage): add GOOGLE_CLOUD_UNIVERSE_DOMAIN configuration option for GCP blob storage

fmontalbano-mgen:
## What does this PR do?

This is a duplicate of https://github.com/langfuse/langfuse/pull/12430 since the creator of the PR is no longer in the company and nobody else can edit his work. I re-integrated his changes in this PR and added @Steffen911 recommendations.

Allow Google Storage to use env var GOOGLE_CLOUD_UNIVERSE_DOMAIN with a fallback to standard googleapis.com

Fixes #12426

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## Mandatory Tasks

- [X ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

---

<h3>Greptile Summary</h3>

This PR adds support for the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` environment variable to the `GoogleCloudStorageService`, allowing deployments targeting sovereign or private Google Cloud universes (e.g., government clouds) to configure the correct API endpoint. The new env var is registered in the Zod schema with a default of `"googleapis.com"`, preserving backward compatibility, and is threaded into all three `Storage` constructor paths.

This PR is safe to merge — the change is minimal, all Storage initialization paths are covered, and the default value preserves existing behavior.

No P0 or P1 issues found. The fix is focused: one new env var with a safe default and three symmetric constructor-call updates. No breaking changes introduced.

| Filename | Overview |
|----------|----------|
| packages/shared/src/env.ts | Adds GOOGLE_CLOUD_UNIVERSE_DOMAIN to the Zod env schema with a default of "googleapis.com", maintaining backward compatibility. |
| packages/shared/src/server/services/StorageService.ts | Threads universeDomain from env into all three Storage constructor call sites (JSON credentials, key file, and default auth) in GoogleCloudStorageService. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[GoogleCloudStorageService constructor] --> B{googleCloudCredentials set?}
    B -- Yes --> C{JSON string?}
    C -- Yes --> D["new Storage with credentials + universeDomain"]
    C -- No --> E["new Storage with keyFilename + universeDomain"]
    B -- No --> F["new Storage with universeDomain only"]
    D & E & F --> G["this.bucket = storage.bucket(bucketName)"]
    ENV["GOOGLE_CLOUD_UNIVERSE_DOMAIN env var\ndefault: googleapis.com"] -->|"env.GOOGLE_CLOUD_UNIVERSE_DOMAIN"| D
    ENV -->|"env.GOOGLE_CLOUD_UNIVERSE_DOMAIN"| E
    ENV -->|"env.GOOGLE_CLOUD_UNIVERSE_DOMAIN"| F
```

<sub>Reviews (1): Last reviewed commit: ["feat(Storage): Add universeDomain to oth..."](https://github.com/langfuse/langfuse/commit/78f86c4b4ccef5f55e7e6bd3677343dc8f11e2b4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30699028)</sub>


packages/shared/src/server/services/StorageService.ts
```diff
@@ -889,7 +891,9 @@ class GoogleCloudStorageService implements StorageService {
       }
     } else {
       // Use default authentication (environment variables or instance metadata)
-      this.storage = new Storage();
+      this.storage = new Storage({
+        universeDomain: env.GOOGLE_CLOUD_UNIVERSE_DOMAIN,
+      });
     }
```
Steffen911, line 897: @fmontalbano-mgen Apologies for the delay in review here.
One thing I'm wondering: I would expect that the instance metadata service provides this correctly. I.e. if someone is not using environment variables, but an internal metadata endpoint this addition would be breaking for them, as they now need to configure a new variable.

Can you confirm this understanding? And would it be sufficient for you if we keep this limited to the other two places and leave the "default authentication" as-is? Curious to get your thoughts here.

fmontalbano-mgen: Hi @Steffen911 , thank you for your time for your review.

From what I understood from the Google lib for the Storage instanciation (https://github.com/googleapis/nodejs-storage/blob/189663a279d85451a65614b47a748d667d7eb3db/src/storage.ts#L725), this value is not retrieved from metadata but from the constructor parameters or a default constant, so I think there will be no breaking existing configuration. (I may be wrong on that as I am not that familiar with the internal metadata endpoint)

We use the default authentication path in our use case but we could try to provide "params.googleCloudCredentials" with a default value (if possible) to avoid modifying the default authentication.

Another point : should we change the name of the new env variable "GOOGLE_CLOUD_UNIVERSE_DOMAIN" to something more specific to avoid possible conflict ? i.e : GOOGLE_CLOUD_UNIVERSE_DOMAIN => LANGFUSE_GOOGLE_CLOUD_UNIVERSE_DOMAIN

Steffen911: @fmontalbano-mgen I think it's acceptable as-is. Wanted to understand whether you have any other experience.
Will watch the CI and otherwise merge as-is.

PR #13456, opened 2026-05-04
feat(experiments): show metadata in overview

wochinge:
## Summary

- Shows experiment metadata in the experiment overview sidebar for LFE-9600.
- Pulls out `provider` / `model`, `langfuse.pr_url`, and `langfuse.github_job_url` as first-class overview fields while keeping remaining metadata available below.
- Renders remaining metadata in a collapsed-by-default Pretty JSON view to keep arbitrary metadata inspectable without overwhelming the sidebar.
- Moves Baseline / Compare with controls to a sticky top section so selection controls stay accessible while reviewing overview details.

## Linear

- https://linear.app/langfuse/issue/LFE-9600/show-metadata-in-ui

## Major Decisions

- Metadata is rendered only in the overview sidebar, not in the experiments table, because table cells are not well-suited for interacting with individual metadata values.
- Filtering of promoted metadata keys happens at the overview panel callsite so the metadata view displays only additional metadata.
- Reused the existing Pretty JSON view for arbitrary metadata to reduce bespoke UI logic and align with existing metadata displays.

## Review Focus

- UI layout of the overview sidebar, especially the sticky Baseline / Compare with controls and the collapsed metadata section.
- Whether the promoted PR / GitHub job links and remaining metadata placement feel right in the sidebar priority order.

### DIFF
--- a/web/src/features/experiments/components/ExperimentMetadataSection.tsx
+++ b/web/src/features/experiments/components/ExperimentMetadataSection.tsx
@@ -6,7 +6,6 @@ import {
 } from "@/src/components/ui/collapsible";
 import { ChevronDown, ChevronRight } from "lucide-react";
 import { useState } from "react";
-import { ExperimentOverviewSectionHeading } from "./ExperimentOverviewField";
 
 export const ExperimentMetadataSection = ({
   metadata,
@@ -20,15 +19,18 @@ export const ExperimentMetadataSection = ({
   return (
     <Collapsible open={isOpen} onOpenChange={setIsOpen}>
       <div className="border-t pt-4">
-        <CollapsibleTrigger className="flex w-full items-center justify-between text-left">
-          <ExperimentOverviewSectionHeading>
-            Metadata
-          </ExperimentOverviewSectionHeading>
-          {isOpen ? (
-            <ChevronDown className="text-muted-foreground h-4 w-4" />
-          ) : (
-            <ChevronRight className="text-muted-foreground h-4 w-4" />
-          )}
+        <CollapsibleTrigger asChild>
+          <button
+            type="button"
+            className="flex w-full items-center justify-between text-left"
+          >
+            <span className="text-sm font-medium">Metadata</span>
+            {isOpen ? (
+              <ChevronDown className="text-muted-foreground h-4 w-4" />
+            ) : (
+              <ChevronRight className="text-muted-foreground h-4 w-4" />
+            )}
+          </button>
         </CollapsibleTrigger>
         <CollapsibleContent className="mt-2">
           <PrettyJsonView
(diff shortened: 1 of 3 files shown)


web/src/features/experiments/components/ExperimentMetadataSection.tsx
marliessophie, line 1: Few design comments:

- Metadata should be part of experiment details; not below baseline/comparison selectors. It's part of experiment
- Please don't capaitalise metadata headers (lanfguse/other). This is inconsistent with other headers across the product.
- Please create a metadata header that is consistent with Name, Prompt headers. Also, how does large blob of metadata currently render? Unsure if flattening all keys is sensible. I'd likely start with rendering full metadata in log view format; feels quite native to what we'd like to show. Should we feel there is strong value in breaking out langfuse specific metadata let's: 
  - (1) break out langfuse metadata in different header and find better name -- what does it correspond to semantically? "langfuse metadata" has little semantic meaning. I like the idea of resolving urls etc if we can; so flattening these keys makes sense.  
  - (2) render other metadata as Json or Json tree view (both supported in trace preview, JSON mode beta/non-beta). Current approach will not scale to large blob of metadata. Consider even rendering the full metadata, including the keys that are resolved by us for completness. Let's make this section toggleable either way, and collapsed by default.

web/src/features/experiments/components/ExperimentMetadataSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+const isSafeHttpUrl = (value: string) => {
+  try {
+    const url = new URL(value);
+    return url.protocol === "http:" || url.protocol === "https:";
+  } catch {
+    return false;
+  }
+};
+
+const formatExperimentMetadataValue = (value: unknown): string => {
```
marliessophie: Don't we have an existing helper for this?

web/src/features/experiments/components/ExperimentMetadataSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+  if (value === undefined) return "undefined";
+
+  try {
+    return JSON.stringify(value) ?? String(value);
+  } catch {
+    return String(value);
+  }
+};
+
+const groupMetadata = (
```
marliessophie: We might not need this depending on design - let's discuss.

web/src/features/experiments/components/ExperimentMetadataSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+      // un-namespaced metadata is least structured, so keep it last.
+      if (a.id === LANGFUSE_METADATA_GROUP_ID) return -1;
+      if (b.id === LANGFUSE_METADATA_GROUP_ID) return 1;
+      if (a.id === OTHER_METADATA_GROUP_ID) return 1;
+      if (b.id === OTHER_METADATA_GROUP_ID) return -1;
+      return a.label.localeCompare(b.label);
+    });
+};
+
+const MetadataValue = ({ value }: { value: unknown }) => {
```
marliessophie: Let's consider having a <MetadataHeader> and <MetadataValue> component and using this across the board (prompt name, dataset, github action prams etc)

marliessophie (commented):
Thanks for adding! Excited about functionality. 

Mostly thoughts around design. Let's chat sync as it's likely quicker.

PR #13494, opened 2026-05-07
fix: return 503 instead of 401 when database is unreachable during auth

### LINKED ISSUE #13482: bug: API key validation returns 401 (not 5xx) when database is unreachable
### Describe the bug

When the Langfuse Postgres database is briefly unreachable (e.g. during a Kubernetes pod restart or eviction), incoming OTLP/API requests are rejected with **HTTP 401 Unauthenticated** because the API key lookup throws a Prisma `Can't reach database server` error inside the auth middleware.

This is a server-side condition, not a client auth failure. Returning 401 causes downstream OpenTelemetry collectors to classify the failure as permanent / non-retryable per the OTLP/HTTP spec and drop the spans rather than queue them for retry.

### Steps to reproduce

1. Send valid OTLP traces to `/api/public/otel/v1/traces` with a working public/secret key.
2. Make Postgres unreachable for ~30s (e.g. `kubectl delete pod langfuse-pg-0`, scale pg to 0, or block the pg port).
3. Observe responses to in-flight OTLP requests.

**Expected:** HTTP 5xx (e.g. 503 Service Unavailable or 500), so OTel and other well-behaved clients retry.
**Actual:** HTTP 401 with body like `{"message":"Unauthorized"}`. OTel logs `not retryable error: Permanent error` and drops spans.

Self-hosted

3.172.0

_No response_

### Additional information

In the API auth middleware (the `apiKey.findUnique`-based path), I think it should distinguish between:
- **Truly invalid credentials** → 401 (current behavior, keep).
- **`PrismaClientKnownRequestError` / `PrismaClientInitializationError` (DB unreachable / connection closed)** → return **503 Service Unavailable** (or 500), so retryable per OTLP/HTTP semantics.

No

abhilov23:
## What does this PR do?

When Postgres is temporarily unreachable, the catch block in `createAuthedProjectAPIRoute.ts`
falls back to `error.status || 401` since Prisma exceptions don't carry a `.status` property.
This causes OTel collectors to classify the failure as permanent and drop spans instead of retrying.

Added an `isPrismaException` check to return 503 Service Unavailable instead, consistent with
how `withMiddlewares.ts` already handles Prisma exceptions.

Fixes #13482

## Type of change

## Mandatory Tasks

## Checklist

- I have read the contributing guide
- I have checked if my changes generate no new warnings

---

<h3>Greptile Summary</h3>

This PR fixes a bug where Prisma database exceptions during API key authentication were falling through to `error.status || 401`, causing OTel collectors to treat transient database outages as permanent auth failures and drop spans instead of retrying. The fix adds an `isPrismaException` guard that returns 503 before the 401 fallback.

- **Core fix is correct** — returning 503 instead of 401 for database errors during auth is the right direction, consistent with the motivation in `withMiddlewares.ts`.
- **Missing observability** — the new branch silently discards the error without calling `traceException` or `logger.error`, unlike every other Prisma catch site in the codebase.
- **Over-broad condition** — `isPrismaException` includes `PrismaClientValidationError` and `PrismaClientKnownRequestError`, which are not service-unavailability errors and should not return 503.

<h3>Confidence Score: 3/5</h3>

The auth error path now silently drops Prisma errors with no trace or log entry, creating a blind spot in production observability.

The direction of the fix is sound, but the new branch discards all Prisma errors without calling traceException or logging — every other Prisma catch site in the codebase records the error before responding. A database connectivity issue during auth will produce a 503 with zero diagnostic signal. The guard also covers error classes that do not indicate service unavailability.

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts — the new catch branch needs traceException/logger.error calls and a more precise error-class check.

| Filename | Overview |
|----------|----------|
| web/src/features/public-api/server/createAuthedProjectAPIRoute.ts | Adds `isPrismaException` guard returning 503 in the auth catch block; missing `traceException`/logging means Prisma errors during auth are silently dropped, and the check is broader than needed. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Incoming API Request] --> B[verifyAuth]
    B --> C{Auth succeeds?}
    C -- Yes --> D[Rate limit check]
    C -- No / throws --> E{catch error}
    E --> F{isPrismaException?}
    F -- Yes --> G[503 Service Unavailable - no traceException/log]
    F -- No --> H{error.status?}
    H -- Yes --> I[Return error.status]
    H -- No --> J[Return 401]
    D --> K[Parse query / body]
    K --> L[Execute route fn]
    L --> M[Return response]

    style G fill:#f9c74f,stroke:#f3722c
    style J fill:#f9c74f,stroke:#f3722c
```

`````markdown

---

### Issue 1 of 2
web/src/features/public-api/server/createAuthedProjectAPIRoute.ts:291-294
**Missing `traceException` and logging for Prisma errors**

The equivalent path in `withMiddlewares.ts` (lines 96–98 and 140–146) always calls `traceException(error)` — and the outer `catch` calls `logger.error(error)` — before returning the response. Here the Prisma error is silently discarded: no trace is recorded, no log entry is written. Any database connectivity problem during auth will produce a 503 with zero observability signal, making it very hard to diagnose in production.

### Issue 2 of 2
web/src/features/public-api/server/createAuthedProjectAPIRoute.ts:291-294
`isPrismaException` matches five Prisma error classes, including `PrismaClientValidationError` (malformed query — a developer error) and `PrismaClientKnownRequestError` (e.g., unique constraint violation). These are not service-unavailability conditions and should not map to 503. `withMiddlewares.ts` returns 500 for all Prisma exceptions; the original issue is specifically about connection/initialization errors (`PrismaClientInitializationError`, `PrismaClientUnknownRequestError`). Scoping the check more precisely avoids misleading callers.

```suggestion
      if (
        error instanceof Prisma.PrismaClientInitializationError ||
        error instanceof Prisma.PrismaClientUnknownRequestError ||
        error instanceof Prisma.PrismaClientRustPanicError
      ) {
        res.status(503).json({ message: "Service Unavailable" });
        return;
      }
      if (isPrismaException(error)) {
        res.status(500).json({ message: "Internal Server Error" });
        return;
      }
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/db-unreacha..."](https://github.com/langfuse/langfuse/commit/83bf1bc7dfc25b7bf558403b19c47e88eb832d4a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31097182)</sub>

### CODE REVIEW

**Comment by Steffen911** [comment]
@abhilov23 Could you sign the CLA?

**Comment by Steffen911** [comment]
I think it would also be great to log an error here and capture an exception in our tracing using something like
```
logger.error("Response validation failed:", parsingResult.error);
traceException(parsingResult.error);
```

Steffen911:
@abhilov23 Could you review my previous feedback on tracing the exception?

abhilov23:
@Steffen911, sorry for the late reply. Let me take a look at that.

Steffen911:
@abhilov23 Did you get a chance to look into this and add a patch?

Steffen911:
Duplicates https://github.com/langfuse/langfuse/pull/13866

PR #13541, opened 2026-05-09
fix(queue): make BullMQ Redis version check configurable

### LINKED ISSUE #13540: Allow skipping BullMQ Redis version check for Redis-compatible services
## Problem

BullMQ performs a Redis version check during queue/worker initialization and expects Redis >= 5.0.0.

Some Redis-compatible services are functionally compatible with BullMQ but expose their own internal version through the `INFO` command. If that version is lower than 5, Langfuse workers fail to start even though the service supports the required BullMQ commands.

## Proposed solution

Add a self-hosting configuration option to skip BullMQ's Redis version check:

`LANGFUSE_BULLMQ_SKIP_REDIS_VERSION_CHECK=true`

The default remains `false` to preserve current behavior for standard Redis deployments.

## Scope

Apply the option to both BullMQ workers and queue producers, since BullMQ exposes `skipVersionCheck` on queue/worker options.

caikaijie:
## What

Adds `LANGFUSE_BULLMQ_SKIP_REDIS_VERSION_CHECK` as a self-hosting configuration option.

When enabled, Langfuse passes `skipVersionCheck: true` to BullMQ queue producers and workers through centralized BullMQ Redis option helpers.

Default remains `false`, preserving the current behavior for standard Redis/Valkey deployments.

Fixes #13540

## Why

BullMQ checks the Redis version during queue/worker initialization. Some Redis-compatible services support the BullMQ command surface used by Langfuse but expose an internal product version through `INFO`.

If that reported version is lower than 5, Langfuse workers can fail to start even though the backing service is functionally compatible.

## Scope

- Adds validated shared env config
- Centralizes BullMQ Redis options for queue producers and workers
- Applies the option to BullMQ workers
- Applies the option to BullMQ queue producers
- Keeps queue-specific `defaultJobOptions` local to each queue
- Documents the option in env examples and docker compose
- Adds regression coverage for default and enabled behavior

## Impacted packages

- `packages/shared`
- `worker`
- `web` via shared queue producers

## Verification

- `pnpm --filter worker run test bullmqVersionCheckOptions.test.ts`
- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run typecheck`
- `git diff origin/main --check`

---

This PR adds `LANGFUSE_BULLMQ_SKIP_REDIS_VERSION_CHECK` as a configurable environment variable, allowing users with Redis-compatible services that report a version below 5 via `INFO` to bypass BullMQ's startup version check. The default remains `false`, preserving existing behavior.

- Adds the env var to the Zod schema with a `"false"` default, exports a `getBullMQRedisVersionCheckOptions()` helper from `redis.ts`, and consistently spreads it into every BullMQ `Queue` and `Worker` constructor across 30+ queue files and `workerManager.ts`.
- Adds a comprehensive test suite covering the default, enabled, worker, and queue-producer paths using `vi.doMock` and dynamic imports.
</details>

Safe to merge. The change is additive and behind a flag that defaults to false, preserving all existing behavior for standard deployments.

The implementation is mechanically consistent across all 30+ queue files and the worker manager. The env var parsing follows the established Zod pattern, the helper is pure and side-effect-free, and the spread order in workerManager.ts intentionally allows per-worker overrides. Tests cover all four code paths (default, enabled, workers, queue producers) with proper module isolation via vi.doMock and vi.resetModules.

</details>

```mermaid
flowchart TD
    A[Env: LANGFUSE_BULLMQ_SKIP_REDIS_VERSION_CHECK] -->|parsed by Zod, default 'false'| B[env object in shared/src/env.ts]
    B --> C[getBullMQRedisVersionCheckOptions in redis.ts]
    C -->|=== 'true'| D["returns { skipVersionCheck: true }"]
    C -->|otherwise| E["returns {}"]
    D & E --> F[Spread into BullMQ Queue options]
    D & E --> G[Spread into BullMQ Worker options via WorkerManager.register]
    F --> H[Queue producers - 30+ queue files]
    G --> I[Workers - workerManager.ts]
    I -->|additionalOptions spread after| J[Per-worker overrides can supersede global setting]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(queue): make BullMQ Redis version ch..."](https://github.com/langfuse/langfuse/commit/249f8c28f0de548bca613b37a0e286d2aea0d258) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31434377)</sub>


Steffen911:
@caikaijie Could you sign the CLA?

Steffen911 (commented):
Is there a way to handle this more centrally on the redis invocation? The current setup is IMO very brittle due to the amount of times that the config is repeated.

caikaijie:
> Is there a way to handle this more centrally on the redis invocation? The current setup is IMO very brittle due to the amount of times that the config is repeated.

@Steffen911  Thanks for the review. I reworked this to centralize the BullMQ Redis options instead of repeating the config at each BullMQ invocation.

The diff still touches many files under `packages/shared/src/server/redis/`, but those changes are mechanical one-time replacements: each queue producer now calls `createBullMQQueueOptionsWithRedis(...)` instead of constructing the repeated Redis/BullMQ options inline. Future shared BullMQ Redis options should usually only need to be added in the centralized helper.

A few implementation choices:
- Queue producers and workers use separate helpers because queue producers keep `enableOfflineQueue: false`, while workers should not set that option.
- Queue-specific `defaultJobOptions` remain local to each queue so the shared helper only owns Redis/BullMQ connection options.
- The helpers are named `create*WithRedis` because they create a Redis connection and return `null` if that connection cannot be created.
- The new env option defaults to `false`, preserving the current behavior for standard Redis/Valkey deployments.

I also updated the regression tests to cover the default and enabled behavior, plus the queue producer and worker paths.

Steffen911 (approved):

Steffen911:
@caikaijie Change looks good in general! Could you take a look at the failing tests?

caikaijie:
@Steffen911  Thanks! I reproduced and fixed the failing test; the change only updates the test code.

The issue was that the default BullMQ options tests unintentionally inherited `REDIS_CLUSTER_ENABLED=true` from the redis-cluster CI matrix. I isolated the default cases to single-node Redis and added a narrow cluster-mode assertion for `skipVersionCheck`.

I also rebased onto the latest `main` and force-pushed with lease to keep the PR history linear.

Steffen911:
> The issue was that the default BullMQ options tests unintentionally inherited REDIS_CLUSTER_ENABLED=true from the redis-cluster CI matrix. I isolated the default cases to single-node Redis and added a narrow cluster-mode assertion for skipVersionCheck.

@caikaijie Isn't it supposed to work in clustered setups, too? Why does this make a difference or require a single-node redis setup?

caikaijie:
> > The issue was that the default BullMQ options tests unintentionally inherited REDIS_CLUSTER_ENABLED=true from the redis-cluster CI matrix. I isolated the default cases to single-node Redis and added a narrow cluster-mode assertion for skipVersionCheck.
> 
> @caikaijie Isn't it supposed to work in clustered setups, too? Why does this make a difference or require a single-node redis setup?

@Steffen911  Yes, it is expected to work in clustered setups too. This does not require a single-node Redis setup.

The option is applied as a BullMQ top-level option after the Redis connection is created, so it is independent of whether `createNewRedisInstance` returns a standalone Redis client or an ioredis Cluster instance. The clustered path still uses the hash-tagged prefix for BullMQ key slot compatibility.

I also added coverage for this path in the test named `skips Redis version checks in cluster mode`: it enables `REDIS_CLUSTER_ENABLED`, verifies that the mocked connection is a cluster, and checks that `skipVersionCheck: true` is passed together with the cluster prefix.

I’ve also rebased this PR onto the latest main. There were no conflicts, and the targeted worker test passes on my side.

Could you please take another look when you have a chance? If everything looks good, I’d appreciate it if you could merge it. Please let me know if there’s anything else I should update.

Thanks!

PR #13565, opened 2026-05-11
fix(web): use prompt model config for experiments

wochinge:
## Summary

- Use prompt config model settings when configuring an experiment from a prompt, including newly released model names that are not yet in Langfuse's built-in model list.
- Resolve prompt config providers against stored LLM connection providers first, then adapters, and fall back to model-only matching when no provider is configured.
- Preserve prompt-config generation params when provider adapter defaults are applied.

## Linear

- [LFE-9509](https://linear.app/langfuse/issue/LFE-9509/bug-cannot-run-experiments-with-newly-released-models)

## Major Decisions

- Kept runtime semantics aligned with experiment execution: the UI stores the matched LLM connection provider, while the adapter still comes from the stored connection.
- Supported flat prompt config values so configs like `{ "model": "gpt-4.1", "temperature": 0.2 }` can preselect model settings.

## Review Focus

- Provider resolution order in experiment setup: provider match, adapter match, then model-only match.
- Interaction between prompt-config params and adapter default model params.
- Scope of `useModelParams` changes for adding prompt-config models to the selectable model list.

---

This PR fixes experiment setup so that newly released models (not yet in Langfuse's built-in list) can be used when a prompt's `config` field contains model settings. It also hardens SSRF protection by making DNS-resolution failure a hard error, closing the DNS-rebinding gap for LLM base URL and image URL validation.

- **Prompt config model pre-selection**: `allPromptMeta` now returns `config`; `useExperimentPromptData` parses it via `ZodModelConfig.extend`; `useModelParams` gains a `promptConfigModel` option that injects the model into the selectable list and resolves its provider via a three-step lookup (stored provider \u2192 adapter name \u2192 model scan); a `useEffect` in `MultiStepExperimentForm` applies the resolved values to the form.
- **Adapter-default preservation**: The form effect depends on `modelParams.adapter.value` to re-apply prompt-config params after the adapter defaults are written, but this same dependency causes user-initiated provider changes to be silently reverted (see inline comment).
- **SSRF hardening**: Removes `shouldThrowIfDnsResolutionFails` flag; DNS failures now always throw; self-hosted gateways must be explicitly allowlisted.
</details>

The experiment form can be used but users who select a prompt with model config and then try to manually override the provider/model will find their change silently reverted.

The core model-resolution logic and SSRF hardening are sound. The main concern is in MultiStepExperimentForm: the useEffect that re-applies prompt config after adapter defaults uses modelParams.adapter.value as a dependency. This fires on every manual provider change, overriding the user's explicit selection without feedback. A user who wants to run the experiment with a different model than the one in the prompt config cannot do so once a prompt with model config is selected.

web/src/features/experiments/components/MultiStepExperimentForm.tsx — the useEffect dependency list at lines 229-235 needs a closer look for the adapter-change override scenario.
</details>

```mermaid
sequenceDiagram
    participant U as User
    participant Form as MultiStepExperimentForm
    participant EPD as useExperimentPromptData
    participant UMP as useModelParams
    participant API as promptRouter

    U->>Form: Select prompt with model config
    Form->>EPD: watch(promptId)
    EPD->>API: allPromptMeta.useQuery (now includes config)
    API-->>EPD: prompt meta + config
    EPD->>EPD: getPromptModelConfig(config)
    EPD-->>Form: selectedPromptModelConfig

    Form->>UMP: "useModelParams(options={promptConfigModel})"
    UMP->>UMP: resolveProviderForModel(promptConfigModel)
    note over UMP: 1) match by stored provider name<br/>2) match by adapter name<br/>3) match by model in available models
    UMP-->>Form: resolvedPromptConfigProvider, expanded providerModelCombinations

    Form->>Form: useEffect fires [promptId/config changed]
    Form->>UMP: setModelParams(promptProvider, model, modelParams)
    UMP->>UMP: adapter effect fires, applies adapter defaults
    UMP-->>Form: modelParams.adapter.value changed
    Form->>Form: useEffect re-fires [adapter.value dep]
    Form->>UMP: setModelParams again (prompt config params override adapter defaults)
```
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/experiments/components/MultiStepExperimentForm.tsx:217-235
**Adapter change re-fires effect and overrides manual provider selection**

`modelParams.adapter.value` is included as a dependency so that prompt-config params survive the adapter-defaults reset, but it has an unintended consequence: any manual provider change the user makes in the UI triggers an adapter update (via the adapter effect in `useModelParams`), which fires this effect again, which calls `setModelParams` with the *prompt's* provider/model — silently reverting the user's choice.

The failure path is: (1) Prompt with model config selected → effect pre-fills provider/model. (2) User manually picks a different provider → adapter effect fires → `modelParams.adapter.value` changes. (3) This effect re-runs with `promptIdFromHook` and `selectedPromptModelConfig` still set, resolves the same `promptProvider`, and overwrites the user's selection.

A narrower fix would be to track whether the current adapter is already consistent with the prompt config and skip the re-apply on subsequent adapter changes, or handle the adapter-default timing with a single-pass merge inside `useModelParams`.

### Issue 2 of 2
web/src/features/playground/page/hooks/useModelParams.ts:95-102
`resolvedPromptConfigProvider` is recomputed on every render by calling `resolveProviderForModel` inline. Since `resolveProviderForModel` performs a linear scan through `llmApiKeys`, wrapping this in `useMemo` avoids the repeated work and makes the dependency tracking explicit for `providerModelCombinations` and `availableModels`.

```suggestion
  const promptConfigProvider = options?.promptConfigModel?.provider;
  const promptConfigModel = options?.promptConfigModel?.model;
  const resolvedPromptConfigProvider = useMemo(
    () =>
      promptConfigModel
        ? resolveProviderForModel({
            ...(promptConfigProvider ? { provider: promptConfigProvider } : {}),
            model: promptConfigModel,
          })
        : undefined,
    [promptConfigModel, promptConfigProvider, resolveProviderForModel],
  );
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): use prompt model config for ex..."](https://github.com/langfuse/langfuse/commit/6ed0457b821069e4b18b6f9e86fef439a3209100) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31580819)</sub>

### DIFF
--- a/web/src/features/evals/hooks/useEvaluationModel.ts
+++ b/web/src/features/evals/hooks/useEvaluationModel.ts
@@ -1,4 +1,5 @@
 import { api } from "@/src/utils/api";
+import { getEnabledModelParamState } from "@/src/utils/getFinalModelParams";
 import {
   type ModelParams,
   ZodModelConfig,
@@ -32,19 +33,9 @@ export function useEvaluationModel(
         return;
       }
 
-      const modelConfig = Object.entries(parsedModelParams.data).reduce(
-        (acc, [key, value]) => {
-          return {
-            ...acc,
-            [key]: { value, enabled: true },
-          };
-        },
-        {} as UIModelParams,
-      );
-
       setModelParams((prev: UIModelParams) => ({
         ...prev,
-        ...modelConfig,
+        ...getEnabledModelParamState(parsedModelParams.data),
         provider: { value: provider, enabled: true },
         model: { value: model, enabled: true },
       }));
--- a/web/src/features/experiments/components/MultiStepExperimentForm.tsx
+++ b/web/src/features/experiments/components/MultiStepExperimentForm.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { Button } from "@/src/components/ui/button";
 import { Form } from "@/src/components/ui/form";
 import {
@@ -28,7 +28,11 @@ import { useEvaluatorDefaults } from "@/src/features/experiments/hooks/useEvalua
 import { useExperimentEvaluatorData } from "@/src/features/experiments/hooks/useExperimentEvaluatorData";
 import { useExperimentNameValidation } from "@/src/features/experiments/hooks/useExperimentNameValidation";
 import { useExperimentPromptData } from "@/src/features/experiments/hooks/useExperimentPromptData";
-import { getFinalModelParams } from "@/src/utils/getFinalModelParams";
+import {
+  getDisabledModelParamState,
+  getEnabledModelParamState,
+  getFinalModelParams,
+} from "@/src/utils/getFinalModelParams";
 import { showErrorToast } from "@/src/features/notifications/showErrorToast";
 import { Skeleton } from "@/src/components/ui/skeleton";
 import {
@@ -97,6 +101,7 @@ export const MultiStepExperimentForm = ({
   const [selectedSchemaName, setSelectedSchemaName] = useState<string | null>(
     null,
   );
+  const lastAppliedPromptConfigId = useRef<string | null>(null);
 
   const steps = [
     { id: "prompt", label: "Prompt & Model" },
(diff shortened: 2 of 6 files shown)


wochinge:
@claude review

Issue #13287, opened 2026-04-21
bug: Critical - Worker pod fails with OOM when few thousands of traces being added into Human annotation queue


tairMustafaiev:
### Describe the bug

We are having trouble with Langfuse, when on the web UI, user selects few thousands (yes, it's huge amount, but that's the requirement from the business to review and annotate almost everything) of traces and then adds them into Human annotation queue, the worker pod fails to add them in batch action. It works for 1-2k of traces, but fails after 3-5k of traces.

Pod fails with this error:

<--- Last few GCs --->  
  
[7:0x7f408b6ce000]    23260 ms: Scavenge (interleaved) 2028.1 (2039.2) -> 2027.9 (2050.0) MB, pooled: 0 MB, 42.60 / 0.00 ms  (average mu = 0.294, current mu = 0.241) allocation failure;  
[7:0x7f408b6ce000]    24899 ms: Mark-Compact (reduce) 2028.3 (2050.2) -> 2028.0 (2036.2) MB, pooled: 0 MB, 1221.73 / 0.00 ms  (+ 82.7 ms in 7 steps since start of marking, biggest step 59.0 ms, walltime since start of marking 1307 ms) (average mu = 0.266,  
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory  
----- Native stack trace -----

And then after restart of the pod:

2026-04-17T11:10:16.054Z error   Queue job batch-action-processing-job with id codemie-traces-trace-add-to-annotation-queue in batch-action-queue failed job stalled more than allowable limit
Error: job stalled more than allowable limit
    at /app/worker/node_modules/.pnpm/[bullmq@5.34.10](mailto:bullmq@5.34.10)/node_modules/bullmq/dist/cjs/classes/worker.js:811:38
    at Array.forEach (<anonymous>)
    at Worker.notifyFailedJobs (/app/worker/node_modules/.pnpm/[bullmq@5.34.10](mailto:bullmq@5.34.10)/node_modules/bullmq/dist/cjs/classes/worker.js:805:20)
    at /app/worker/node_modules/.pnpm/[bullmq@5.34.10](mailto:bullmq@5.34.10)/node_modules/bullmq/dist/cjs/classes/worker.js:800:18
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
[...]

PR #13585, opened 2026-05-12
fix(annotation): introduce conditional read from events table for batch action

marliessophie:
---

This PR fixes the batch \"add to annotation queue\" action for observations when the events-table UI flag is enabled, routing the worker to a new lightweight ClickHouse stream (`getEventsStreamForAnnotationQueue`) instead of the legacy observations table. It also removes an incorrect type cast that previously prevented the OBSERVATION object type from reaching the action dispatch.

- **New `getEventsStreamForAnnotationQueue`**: Reads `span_id` (aliased as `id`) and `trace_id` from the events table, applying the same cutoff/filter/search logic used by other event streams and deduplicating with `LIMIT BY span_id, project_id`.
- **Conditional tableName routing**: In `annotationQueueItemsRouter`, the OBSERVATION entry now selects `BatchExportTableName.Events` when `LANGFUSE_ENABLE_EVENTS_TABLE_UI === \"true\"`, falling back to the Observations table otherwise.
- **New integration tests**: Adds coverage for trace, session, and events-based observation batch add-to-queue flows; fixes a pre-existing `JSON.stringify` omission in `createEvent` helper calls.
</details>

The change is safe to merge for its intended use case; the new events-table observation path works correctly end-to-end as confirmed by the new integration tests.

The core logic — selecting the right ClickHouse stream and writing annotation queue items with the correct span_id — is sound and well-tested. The main concern is that the events-table branch in handleBatchActionJob dispatches to getEventsStreamForAnnotationQueue based solely on tableName, without an actionId guard. Any future action added to that if-block that also uses the events table would silently receive observation rows.

The stream-selection ladder in worker/src/features/batchAction/handleBatchActionJob.ts (lines 185-199) deserves a second look to consider adding an explicit actionId guard alongside the tableName check.
</details>

```mermaid
flowchart TD
    A[Batch Action Job observation-add-to-annotation-queue] --> B{LANGFUSE_ENABLE_EVENTS_TABLE_UI?}
    B -->|true| C[tableName = Events]
    B -->|false| D[tableName = Observations]
    C --> E{tableName === Events?}
    D --> F{tableName === Observations?}
    E -->|yes| G[getEventsStreamForAnnotationQueue reads span_id as id from ClickHouse events table]
    F -->|yes| H[getObservationStream reads from ClickHouse observations table]
    G --> I[Stream rows with id and traceId]
    H --> J[Stream rows with id]
    I --> K[processActionChunk processAddObservationsToQueue chunkIds = row.id = span_id]
    J --> K
    K --> L[(Postgres annotationQueueItem objectId = span_id)]
```
</details>

`````markdown

---

### Issue 1 of 2
worker/src/features/batchAction/handleBatchActionJob.ts:191-199
**Events-table routing is not guarded to observation-only actions**

The new `tableName === BatchTableNames.Events` branch routes to `getEventsStreamForAnnotationQueue` without first verifying that `actionId` is `observation-add-to-annotation-queue`. If a future action (e.g., `score-delete`) is ever invoked with `tableName === BatchTableNames.Events`, this branch will silently use the wrong stream — reading observation rows by `span_id` and handing those IDs to the wrong processor. The `trace-delete` case is already protected because it is checked by `actionId` first; the same pattern should apply here.

### Issue 2 of 2
worker/src/features/database-read-stream/event-stream.ts:700-708
**`traceId` is yielded but never consumed downstream**

`processActionChunk` calls `processAddObservationsToQueue` with only `batch.map((r) => r.id)` — the `traceId` field yielded here is never read. Emitting unused fields is harmless, but it adds unnecessary noise to the stream type and may mislead future readers into thinking `traceId` is required by the consumer.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["tests: add"](https://github.com/langfuse/langfuse/commit/e886f7716fccbcf9cfbb32308887fd6bb9bdb527) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31733445)</sub>

### DIFF
--- a/web/src/features/annotation-queues/server/annotationQueueItemsRouter.ts
+++ b/web/src/features/annotation-queues/server/annotationQueueItemsRouter.ts
@@ -59,7 +59,10 @@ const MAP_OBJECT_TYPE_TO_ACTION_PROPS: Record<
   },
   [AnnotationQueueObjectType.OBSERVATION]: {
     actionId: ActionId.ObservationAddToAnnotationQueue,
-    tableName: BatchExportTableName.Observations,
+    tableName:
+      env.LANGFUSE_ENABLE_EVENTS_TABLE_UI === "true"
+        ? BatchExportTableName.Events
+        : BatchExportTableName.Observations,
   },
 };
 
@@ -296,10 +299,7 @@ export const queueItemRouter = createTRPCRouter({
       let createdCount = 0;
 
       if (input.isBatchAction && input.query) {
-        const actionProps =
-          MAP_OBJECT_TYPE_TO_ACTION_PROPS[
-            input.objectType as "TRACE" | "SESSION"
-          ];
+        const actionProps = MAP_OBJECT_TYPE_TO_ACTION_PROPS[input.objectType];
         if (!actionProps) {
           throw new TRPCError({
             code: "BAD_REQUEST",
--- a/worker/src/__tests__/batchAction.test.ts
+++ b/worker/src/__tests__/batchAction.test.ts
@@ -672,6 +672,122 @@ describe("select all test suite", () => {
       expect(jobs).toHaveLength(0);
     });
   });
+
+  it("should add traces to annotation queue", async () => {
+    const { projectId } = await createOrgProjectAndApiKey();
+
+    const traceId1 = uuidv4();
+    const traceId2 = uuidv4();
+    const traces = [
+      createTrace({
+        project_id: projectId,
+        id: traceId1,
+        timestamp: new Date("2024-01-01").getTime(),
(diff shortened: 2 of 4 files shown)


tairMustafaiev:
Hi @marliessophie 

Could you please take a look into this issue https://github.com/langfuse/langfuse/issues/13287

I believe, your MR fixes that and problem with OOM, for thousands of traces being added into Human Annotation queue. Please confirm, if so, I'll close my MR.

marliessophie:
I believe steffen responsed to you on your PR. Please follow up there should you continue to experience issues.

PR #13602, opened 2026-05-12
feat(dashboarding): support experiment dimensions and filters in observations view

marliessophie:
# UI/UX commentary 

Added support for experiment preset, dimensions and filters. See video for current UI. An alternative UI consideration is to add "experiments" as a dropdown option in views, which can be achieved through a simple UI only alias. However, for v0 a preset is sufficient and simpler. It is a non-issue to add a view alias later as this is a forwards/backwards compatible UI only change. 

We can also consider extending presets in the future, allowing to filter for total experiment cost or avg experiment item latency. 

https://github.com/user-attachments/assets/f96f98a4-3235-44a0-b2ca-8204e6e32d0e

This PR adds experiment dimensions (`experimentName`, `experimentDatasetId`, `experimentId`, `experimentMetadata`) to the observations v2 view, surfaces them as filter and breakdown options in the widget editor, and introduces a "Presets" popover that lets users apply a canned "Experiment Data" filter in one click.

- New experiment fields are wired from `eventsObservationsView` through `dashboardUiTableToViewMapping`, `widgetFilterColumns`, and a new `queryBuilder` special-case for the array-backed `experimentMetadata` column.
- `eventsFilterOptions` (v2) and `datasets.allDatasetMeta` are fetched conditionally to populate experiment-name and dataset-name dropdown options, with v1 queries now gated behind `enabled: viewVersion === \"v1\"`.
- The "Presets" popover introduces `WIDGET_FILTER_PRESETS`, a typed constant that seeds view + filter state in a single click.
</details>

Two defects in the changed code need fixing before merge: the preset handler leaves measure/aggregation/pivot state dirty when it switches views, and the experimentMetadata special-case in queryBuilder bypasses the view-membership guard that protects every other filter column.

The preset onClick path directly calls setSelectedView without running the cleanup block that the normal view-switch Select handler executes, so a user arriving from scores-numeric will have an invalid measure on the observations query. Separately, experimentMetadata is handled in queryBuilder via a hardcoded column name with no check that the current view actually owns that column, leaving a gap in the validation fence that normally throws for unknown columns.

web/src/features/widgets/components/WidgetForm.tsx (preset onClick) and web/src/features/query/server/queryBuilder.ts (experimentMetadata branch)
</details>

```mermaid
flowchart TD
    A[User opens WidgetForm v2] --> B{Action}
    B --> C[Clicks Presets popover]
    B --> D[Changes View dropdown]
    C --> C1[setSelectedView preset.view]
    C --> C2[setUserFilterState preset.filters]
    C1 --> C3[Measure/dimension/pivot cleanup NOT triggered]
    D --> D1[onValueChange handler]
    D1 --> D2[Reset measure, aggregation, dimension]
    D1 --> D3[Filter invalid pivot metrics/dims]
    D1 --> D4[Remove score-only filters]
    C1 --> E[eventsObservationsView selected]
    D --> E
    E --> F[QueryBuilder.mapFilters]
    F --> G{Filter column?}
    G -->|in view.dimensions| H[Normal dimension path]
    G -->|experimentMetadata| I[Special case: clickhouseSelect = experiment_metadata]
    G -->|not in dimensions or special cases| J[throw InvalidRequestError]
    I --> K[No view membership check; other views would hit this branch]
```
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/widgets/components/WidgetForm.tsx:1380-1385
**Preset handler skips measure/dimension/pivot cleanup**

`setSelectedView(preset.view)` is called directly, bypassing all the cleanup logic that lives in the Select's `onValueChange` handler (lines 1409-1441). If the user is on `scores-numeric` with `selectedMeasure = "value"` and clicks "Experiment Data", the view becomes `"observations"` while `selectedMeasure` stays `"value"`. Since `"value"` is not a valid measure on the observations view, the next query will hit an `InvalidRequestError` (or silently return wrong data). The same applies to `selectedDimension`, `selectedAggregation`, and pivot-table metrics/dimensions — none of them are reset.

### Issue 2 of 2
web/src/features/query/server/queryBuilder.ts:383-387
**`experimentMetadata` special-case bypasses view-dimension validation**

`experimentMetadata` is not defined in `eventsObservationsView.dimensions`, so a client sending `column: "experimentMetadata"` on any other view (traces, scores-numeric, etc.) would skip the dimension check, skip the `"metadata"` branch, and land here — generating a SQL fragment that references `experiment_metadata_names` / `experiment_metadata_values` columns that don't exist on those tables. The guard at line 396 that would normally catch this (`throw new InvalidRequestError(...)`) is unreachable for this column. Consider adding an explicit view-membership check before the special-case branch, or defining `experimentMetadata` as a proper dimension so the normal validation path applies.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["refactor: handle metadata filter appropr..."](https://github.com/langfuse/langfuse/commit/a40243c351f6778ff68e5de27aec88143dd30e53) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31811273)</sub>

### CODE REVIEW

File: web/src/features/query/dataModel.ts [hunk 1]
```diff
@@ -1147,6 +1147,35 @@ export const eventsObservationsView: ViewDeclarationType = {
         valueAlias: "usage_value",
       },
     },
+    // Experiment dimensions (v2 only - experiment data only exists in events table)
+    experimentName: {
+      sql: "nullIf(events_observations.experiment_name, '')",
+      alias: "experimentName",
+      type: "string",
+      description: "Name of the experiment this observation belongs to.",
+      // highCardinality: true,
```
marliessophie: **Question**

`experimentName`, and `experimentDatasetId` are technically high cardinality but immensely useful for grouping 
- Are we concerned if users group by these dimensions (likely yes)
- If yes, would we be happy to expose these dimensions if and only if we ensure they can only be added in combination with filter on `experiment_id != ''`? I can work some simple UI magic in a follow up PR. 
- Either way, in this PR, I'd suggest let's add highCardinality flag and keep dimensions as they are useful for top n queries nontheless

web/src/features/query/dataModel.ts
```diff
@@ -1147,6 +1147,35 @@ export const eventsObservationsView: ViewDeclarationType = {
...
+    },
+    experimentId: {
+      sql: "nullIf(events_observations.experiment_id, '')",
+      alias: "experimentId",
+      type: "string",
+      description: "ID of the experiment this observation belongs to.",
+      uiHidden: true, // Only available as filter, not as breakdown dimension
+    },
+    experimentMetadata: {
```
marliessophie: **Question**

I noticed we don't define `metadata` as dimension and have [special handling](https://github.com/langfuse/langfuse/blob/0f08f517ee4e052a28a2ef16ffdd06814e3c59e6/web/src/features/query/server/queryBuilder.ts#L379) for it. Why is this? 

I chose to define as dimension bc experiment metadata should force view version 2 as soon as any experiment related filter is added. Tested locally. Could you please confirm this is the correct pattern to follow?

web/src/features/widgets/components/WidgetForm.tsx
```diff
@@ -1089,6 +1181,43 @@ export function WidgetForm({
...
+        (dimension) => dimension in newViewDeclaration.dimensions,
+      );
+      setPivotDimensions(validDimensions);
+    }
+
+    // Remove filters that are not valid for the new view
+    const validColumns = getValidFilterColumnIds(newView);
+    setUserFilterState((prev) =>
+      prev.filter((filter) => validColumns.has(filter.column)),
+    );
```
marliessophie, line 1230: **FYI**

Made improvement to keep only view agnostic filters on view change. More stable than hardcoded allowlist that manually strips specific columns.

web/src/features/widgets/components/WidgetForm.tsx
```diff
@@ -663,26 +693,63 @@ export function WidgetForm({
...
+    {
+      projectId,
+    },
+    { enabled: viewVersion === "v2" },
   );
-  const tagsOptions = traceFilterOptions.data?.tags || [];
-  const modelOptions = generationsFilterOptions.data?.model || [];
-  const toolNamesOptions = generationsFilterOptions.data?.toolNames || [];
+
+  // Resolve filter options based on viewVersion
```
marliessophie, line 721: **FYI** 

Note that I gated the trpc calls for filter options to the view version; meaning that if viewVersion v2 we read eg the environment options from events table and if viewVersion v1 we read them from observations table.

marliessophie:
Hey @sumerman, would appreciate eyes on this pull request, could you please take a look? 

In particular skimming over: 
- web/src/features/widgets/components/widgetFilterColumns.ts
- web/src/features/query/dataModel.ts
- web/src/features/query/dashboardUiTableToViewMapping.ts

and reading the comments I left should do it. Appreciate it, thank you!

packages/shared/src/features/query/dataModel.ts
```diff
@@ -1147,6 +1147,35 @@ export const eventsObservationsView: ViewDeclarationType = {
         valueAlias: "usage_value",
       },
     },
+    // Experiment dimensions (v2 only - experiment data only exists in events table)
+    experimentName: {
```
sumerman, line 1151: I would prefer if it wasn't at all possible for users to save widgets with these fields or use in public API.

web/src/features/query/dataModel.ts
sumerman: Yes, at the very least `highCardinality` flag should be in place

web/src/features/query/dataModel.ts
sumerman: IIRC because `metadata` is never used. It's always `metadata[key]` or even something more involved. It requires custom validations and operators. So, I would expect placing `experiment_metadata` here to break in some instances.

marliessophie: In that case, I'll remove the `experimentMetadata` dimension for now and potentially add this in a later stage PR.

packages/shared/src/features/query/dataModel.ts
marliessophie: These fields have `highCardinality: true`, which ensures validateQuery() rejects unbounded queries—using them as dimensions requires explicit `row_limit` + `orderBy` desc (top-N pattern) and they cannot be used with `timeDimension`. From my understanding, this protection applies to both the public API and saved widgets, so we believe it's sufficient.

Or am I missing something? Thanks!

sumerman: Yeah. High cardinality protections should work just fine. But I believe we settled on not allowing these fields in custom widgets for now, or am I misremembering?

marliessophie: My understanding was

* we would not allow time series graphs with bucketing by experiment_name in custom widgets
* exposing them as filters should work fine 
* high cardinality blocks using the dimensions for time series, but allows using them for total value

Essentially treating them like other correlation identifiers: session_id, user_id.

Where does our understanding diverge?

sumerman: Nevermind then. I am probably misremembering.

marliessophie: So you're okay with the current approach? Then I'd merge the PR 🙏

sumerman: yeah

PR #13623, opened 2026-05-13
feat(worker): V4 self-hosted historic backfill chain

Steffen911:

This PR implements a V4 self-hosted historic backfill chain — a five-step background migration sequence (M1–M5) that populates the new `events_full` table from existing `traces`, `observations`, and `dataset_run_items_rmt` data. It also adds V4 write-mode flags to control the dual-write/events-only transition and makes ClickHouse cluster query references configurable.

- **M1–M5 chain**: M1 creates virtual root spans from traces, M2 rewrites observations into a sort-key-optimized scratch table, M3 joins that scratch table against live traces and writes child spans to `events_full`, M4 enriches experiment-tagged spans via DRI cursor pagination, and M5 drops the scratch table once the v4 path is confirmed healthy. Each step ships dormant behind an env gate.
- **envGate mechanism**: `BackgroundMigrationManager` filters the Prisma query by active gate env vars so a dormant gated row never head-of-line blocks ungated migrations.
- **V4 flags**: `LANGFUSE_MIGRATION_V4_WRITE_MODE` and `LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR` control the write destination; boot-time validation rejects incoherent combinations.
</details>

The migration chain ships dormant and requires an explicit opt-in env variable, so the M3 query bug does not affect live traffic — but any self-hoster who activates the historic backfill gate will have M3 fail immediately on every chunk.

The M3 migration uses `t.partition_id` in the traces subquery WHERE clause. The `traces` table has no `partition_id` column — the correct virtual column is `_partition_id`. ClickHouse will throw a missing-column error on every chunk, making M3 completely non-functional for all self-hosters who activate the backfill gate. The rest of the chain (M1, M2, M4, M5), the envGate mechanism, V4 flag helpers, and `queryTracking.ts` changes are all correct.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts — specifically the traces subquery on line 295
</details>

```mermaid
flowchart TD
    subgraph Gate["Env Gates (default: dormant)"]
        G1[LANGFUSE_BACKGROUND_MIGRATION_V4_ENABLE_HISTORIC_BACKFILL]
        G2[LANGFUSE_BACKGROUND_MIGRATION_V4_DROP_PID_TID_SORTING_TABLES]
    end
    subgraph M1["M1: createRootSpansFromTraces"]
        T1[traces table] -->|INSERT per partition| EF1[events_full root spans]
    end
    subgraph M2["M2: rewriteObservationsToPidTidSorting"]
        OBS[observations] -->|INSERT per partition| SORT[observations_pid_tid_sorting]
        SORT -->|SYSTEM STOP MERGES| FROZEN[frozen scratch table]
    end
    subgraph M3["M3: backfillEventsFullFromObservations"]
        FROZEN2[observations_pid_tid_sorting per part] -->|LEFT ANY JOIN| TR[traces live]
        TR --> EF2[events_full child spans]
    end
    subgraph M4["M4: backfillEventsFullFromDatasetRunItems"]
        DRI[dataset_run_items_rmt cursor paginated] --> ENRICH[enrichSpansWithExperiment]
        ENRICH --> EF3[events_full experiment spans]
    end
    subgraph M5["M5: dropPidTidSortingTables"]
        DROP[DROP TABLE IF EXISTS observations_pid_tid_sorting SYNC]
    end
    G1 -->|activates| M1
    G1 -->|activates| M2
    G1 -->|activates| M3
    G1 -->|activates| M4
    G2 -->|activates| M5
    M1 --> M2 --> M3 --> M4 --> M5
```
</details>

`````markdown

---

### Issue 1 of 3
worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts:295-305
**Wrong column name for traces partition filter**

The subquery's WHERE clause uses `t.partition_id` (no underscore prefix), but `partition_id` is not a real column on the `traces` table. The correct virtual column is `t._partition_id`. Since the `traces` DDL (`0001_traces.up.sql`) contains no `partition_id` column, ClickHouse will throw "Missing columns: 'partition_id'" when M3 runs, halting the entire migration step. `createRootSpansFromTraces.ts` correctly uses `WHERE t._partition_id = {partition: String}` for the same pattern.

```suggestion
        FROM traces t
        WHERE t._partition_id = {partition: String}
```

### Issue 2 of 3
worker/src/backgroundMigrations/rewriteObservationsToPidTidSorting.ts:266-272
**`stopMergesOnScratchTable` is called even when chunks have failures**

`stopMergesOnScratchTable()` runs after the main loop regardless of how many todos remain in `"failed"` status. On a deployment where some partitions fail permanently, merges are frozen on `observations_pid_tid_sorting` while the table is partially populated. If the operator then re-runs M2 with `retryFailed=true`, new inserts accumulate into frozen parts, which can cause a part explosion until M5 drops the table.

### Issue 3 of 3
worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts:490-495
**`recoverInProgressTodos` in-place mutation coupling**

`recoverInProgressTodos(state.todos, ...)` modifies todo objects in-place and the wrapper then calls `this.updateState(state)` to persist. This is functionally correct today because TypeScript object references propagate the mutations. If a future refactor re-loads state inside the base function, the mutations would be silently lost. Consider documenting this coupling.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: extend eval tests"](https://github.com/langfuse/langfuse/commit/ca6227caf826404052b3ecbd4d6d7ea6b3e43327) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34586015)</sub>

### CODE REVIEW

File: packages/shared/clickhouse/scripts/dev-tables.sh [hunk 1]
```diff
@@ -407,10 +403,10 @@ SETTINGS
     enable_block_number_column = 1,
     enable_block_offset_column = 1,
     prewarm_mark_cache = 1,
-    prewarm_primary_key_cache = 1;
-    -- cache_populated_by_fetch = 1; -- Not available in OSS ClickHouse
```
sumerman, line 411: I would like to keep this documented somewhere. For our own record keeping as well as for self-hosters on CH Cloud.

packages/shared/prisma/migrations/20260521120002_add_backfill_events_full_from_observations_background_migration/migration.sql
```diff
@@ -0,0 +1,19 @@
+-- M3 of the V4 self-hosted historic backfill chain (LFE-8833).
+-- Reads `observations_pid_tid_sorting` (produced by M2) LEFT ANY JOIN with the
+-- live `traces` table and writes child events into `events_full`. Trace
+-- properties are propagated lightly (name, user_id, session_id, version,
+-- release, tags, public, bookmarked); metadata is intentionally not copied to
+-- keep the join cheap on self-hoster hardware.
```
sumerman: I think this needs to be explicitly documented in the migration docs.

packages/shared/src/server/clickhouse/client.ts
```diff
@@ -21,10 +21,7 @@ type ServiceClickhouseSettings = {
  * Remove these once we remove corresponding variables
  */
 const EVENTS_TABLE_READ_PATH_ENV_KEYS = [
-  "LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS",
-  "LANGFUSE_ENABLE_EVENTS_TABLE_UI",
-  "LANGFUSE_ENABLE_EVENTS_TABLE_FLAGS",
-  "LANGFUSE_ENABLE_EVENTS_TABLE_V2_APIS",
+  "LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN",
```
sumerman, line 28: I would prefer this to be a cleanup at v4 release time. It is merely used as a detection measure, so IMO, there is no harm in keeping a longer list here.

packages/shared/src/server/repositories/events.ts
```diff
@@ -959,6 +962,10 @@ export const getTraceByIdFromEventsTable = async ({
   renderingProps?: RenderingProps;
   clickhouseFeatureTag?: string;
   preferredClickhouseService?: PreferredClickhouseService;
+  /** When true, sets input/output columns to empty in the query to reduce database load */
+  excludeInputOutput?: boolean;
```
sumerman, line 974: On the one hand I dislike this, on the other, I see this is a crutch that is necessary until old codepaths are gone. But since it may linger for some time, I would nevertheless prefer not to have `metadata` or IO in the return type whenever `excludeInputOutput` or `excludeMetadata` are set. Or have them as Optional.

.env.dev.example
```diff
@@ -182,15 +182,16 @@ LANGFUSE_AWS_BEDROCK_REGION="eu-west-1"
 LANGFUSE_AWS_BEDROCK_MODEL="eu.anthropic.claude-3-haiku-20240307-v1:0"
 LANGFUSE_IN_APP_AGENT_AWS_PROFILE="playground"
 
+# V4 migration flags.
+# Write target: `legacy` | `dual` | `events_only`.
+LANGFUSE_MIGRATION_V4_WRITE_MODE=dual
+# Gate the V4 events_full read paths (UI, tRPC, v2 APIs, observations API).
+LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true
+# OTel ingestion behaviour: `dual_write` (SDK-version dispatch) or `direct`.
+LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=dual_write
```
sumerman, line 190: we need to explicitly ask folks to set `CLICKHOUSE_DISABLE_LAZY_MATERIALIZATION` as it becomes problematic exactly at the CH versions we set as minimal requirement.

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
```diff
@@ -288,6 +296,21 @@ export const createAuthedProjectAPIRoute = <
...
+    // when the deployment is in events_only mode — those tables are no longer
+    // populated, so the response would be stale or empty. Returning 404 keeps
+    // the surface area consistent with "this endpoint is not available here".
+    if (
+      routeConfig.rejectInEventsOnlyMode &&
+      env.LANGFUSE_MIGRATION_V4_WRITE_MODE === "events_only"
+    ) {
+      res.status(404).json({
+        message:
+          "This endpoint is not available on deployments running in V4 events_only mode.",
```
sumerman: Would be great to add a link to a docs page describing legacy -> v4 mapping.

worker/src/features/eventPropagation/handleEventPropagationJob.ts
```diff
@@ -63,13 +63,6 @@ export const handleEventPropagationJob = async (
     job.data.id,
   );
 
-  if (env.LANGFUSE_EXPERIMENT_EARLY_EXIT_EVENT_BATCH_JOB === "true") {
```
sumerman, line 66: Would love some context on why this flag was here initially.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -387,46 +396,62 @@ export const otelIngestionQueueProcessorBuilder = (
       }
 
       const writePath = useDirectEventWrite
-        ? headerBasedDirectWrite
-          ? "direct_header"
-          : "direct_scope"
+        ? envForcesDirect && !headerBasedDirectWrite
```
sumerman: This is too terse, I would strongly prefer an if or a switch statement.

Steffen911:
@claude review

worker/src/backgroundMigrations/backgroundMigrationManager.ts
```diff
@@ -37,16 +40,37 @@ export class BackgroundMigrationManager {
 
   public static async run(): Promise<void> {
     await instrumentAsync({ name: "background-migration-run" }, async () => {
+      // A migration row may declare `args.envGate = "<ENV_VAR>"` to remain dormant
+      // until the operator sets that env var to "true" (e.g. ship dormant on v3,
+      // activate on v4). We push the gate check into the findFirst predicate so
+      // dormant rows never get picked — otherwise the first dormant row in name
+      // order would head-of-line block every unrelated later migration.
+      //
+      // Gate names share the LANGFUSE_BACKGROUND_MIGRATION_ prefix so we can
```
sumerman: nit: I think detail is missing from the backgroundMigration level README

web/src/__tests__/server/traces-trpc-events-only.servertest.ts
```diff
@@ -0,0 +1,225 @@
...
+        },
+      });
+      expect(eventTraceFull).toBeDefined();
+      expect(eventTraceFull?.bookmarked).toBe(true);
+    });
+  });
+
+  // Same routing as the bookmark case, but publish throws NOT_FOUND (rewrapped
+  // as INTERNAL_SERVER_ERROR) instead of silently no-opping when the read is
+  // misrouted to the empty legacy table.
```
sumerman, line 168: What this is about I see no assertion on errors being thrown?

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+
+  startPolling(
+    pollIntervalMs: number,
+    onComplete: OnQueryCompleteCallback<T>,
+    scheduleNext: () => Promise<void>,
+    logPrefix = "[Backfill]",
+  ): void {
+    if (this.pollInterval) return;
+
+    this.pollInterval = setInterval(async () => {
```
sumerman: `setInterval` may have a weird edge case if the function body takes longer than interval there could be a backlog of polls. I believe a recursive `setTimeout` is the recommended approach. The `if (this.isPolling) return;` and ` if (this.pollInterval) return;` should be achieving similar outcome, but I find it easier to reason about `setTimeout`.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+      try {
+        for (const [queryId, todo] of this.activeQueries) {
+          try {
+            const status = await pollQueryStatus(queryId);
+
+            if (status === "completed") {
+              this.activeQueries.delete(queryId);
+              try {
+                await onComplete(todo, true);
+                await scheduleNext();
```
sumerman: Clearly it is implied that `scheduleNext` updates `activeQueries` but it is not evident from the types and not mentioned in the comments.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+              }
+            } else if (status === "failed" || status === "not_found") {
+              this.activeQueries.delete(queryId);
+              const error =
+                status === "failed"
+                  ? await getQueryError(queryId)
+                  : "Query not found in query_log";
+              try {
+                await onComplete(todo, false, error);
+                await scheduleNext();
```
sumerman: So `scheduleNext` is called inside 2 branches of a non-exhaustive `if` statement. From the current `QueryStatus` definition it is evident that we don't want to `scheduleNext` while a query is still running. But I would prefer if that were explicit and the if/else were exhaustive.

sumerman (commented):
Stage 1 of the review. Everything but `worker/src/backgroundMigrations/utils/backfillBase.ts` and BG migrations/backfill code has been reviewed.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+/**
+ * Fires a long-running ClickHouse query, confirms it's tracked in
+ * system.processes, then aborts the HTTP connection so the query continues
+ * server-side and we poll for completion via pollQueryStatus.
+ */
+export async function fireQuery({
+  query,
+  queryId,
+  params,
+  retryCount = 0,
```
sumerman: While reading this I thought that `retryCount` controls how many times this function retries the query. Instead, it looks like a retry attempt number that is used to pick from `retrySettings`. IMO should have a different name.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+    );
+    throw error;
+  }
+
+  logger.info(
+    `${logPrefix} Query ${queryId} confirmed running, aborting HTTP connection`,
+  );
+  abortController.abort();
+
+  queryPromise.catch((err) => {
```
sumerman: AFAIK .catch should be set before aborting

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+  retryWaitMs?: number;
+  logPrefix?: string;
+}
+
+/**
+ * Fires a long-running ClickHouse query, confirms it's tracked in
+ * system.processes, then aborts the HTTP connection so the query continues
+ * server-side and we poll for completion via pollQueryStatus.
+ */
+export async function fireQuery({
```
sumerman: can potentially be a general purpose helper along with `pollQueryStatus`

sumerman (commented):
worker/src/backgroundMigrations/utils/backfillBase.ts is done

packages/shared/clickhouse/scripts/dev-tables.sh
Steffen911: @sumerman I'll add a section `ClickHouse Cloud Settings` to the migration guide in our Google Doc (see https://docs.google.com/document/d/1bvz3FUFn3T4rfJ_U1qIh36FcJGfUVQwdAIOAHa3wPnA/edit?tab=t.0#heading=h.wn2wqzyr013v). Would that work for you or do you have a separate preference?

packages/shared/prisma/migrations/20260521120002_add_backfill_events_full_from_observations_background_migration/migration.sql
Steffen911: Will add a new Historic Data Migration section here: https://docs.google.com/document/d/1bvz3FUFn3T4rfJ_U1qIh36FcJGfUVQwdAIOAHa3wPnA/edit?tab=t.0#heading=h.lagiaelyw3mi.

packages/shared/src/server/repositories/events.ts
Steffen911: @sumerman Agree. I don't want to change the interface of this method, though. I've added a JSDoc comment marking this as `@deprecated` which should discourage use in most IDEs and point to the getTraceByIdFromEventsTable with a cleaner interface. I.e. will solve this via documentation which should cover this sufficiently.

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
Steffen911: @sumerman Would you link to the google doc here or rather a Langfuse docs page like https://langfuse.com/docs/v4?

worker/src/features/eventPropagation/handleEventPropagationJob.ts
Steffen911: @sumerman We wanted to have a "kill switch". But stopping queue processing would be just as effective. I.e. it was redundantly added and never served a purpose.

worker/src/backgroundMigrations/backgroundMigrationManager.ts
Steffen911: @sumerman Actually it's added there? Which detail are you missing specifically?

web/src/__tests__/server/traces-trpc-events-only.servertest.ts
Steffen911: @sumerman There should be no error thrown in this case as we should fallback correctly to the events table. I think that's more an explanation in how publish and bookmark differ than what should happen in the test.
The test case is overly explicit w.r.t to comments, but it tests a valid scenario that we may have throughout the migration phase.

worker/src/backgroundMigrations/utils/backfillBase.ts
Steffen911: @sumerman My understanding is that any subsequent interval would simply trigger the callback as well and this should be caught by returning early via `this.isPolling`.
Unless you feel strongly, I'd accept it as-is.

packages/shared/clickhouse/scripts/dev-tables.sh
sumerman: It was a code comment I don't see why it should stop being a code comment. Google doc is good for external visibility, but easier to loose when we start creating new envs. I am already not sure if this was applied in JP.

sumerman: (now it is applied)

packages/shared/src/server/repositories/events.ts
sumerman: Thank you. That works as well.

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
sumerman: Initially could be a google doc that we update later if we don't have a comprehensive docs page. Or we can just link both https://langfuse.com/docs/api-and-data-platform/features/observations-api#v2 and https://langfuse.com/docs/metrics/features/metrics-api#v2 

By the time v4 OSS release comes we should definitely make this a Langfuse docs page.

worker/src/features/eventPropagation/handleEventPropagationJob.ts
sumerman: TY

worker/src/backgroundMigrations/backgroundMigrationManager.ts
sumerman: Oh you right, my bad `LANGFUSE_BACKGROUND_MIGRATION` prefix is described there.

web/src/__tests__/server/traces-trpc-events-only.servertest.ts
sumerman: TY

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -668,25 +499,31 @@ export default class BackfillEventsHistoricFromParts implements IBackgroundMigra
...
+        // the source rows are now in the merged successor — which is not in
+        // our todo list. Abort so the operator can clear `state.chunksLoaded`
+        // (or wipe state) and re-run to pick up the merged successor.
         const partStillActive = await this.verifyPartStillActive(todo.partId);
-
         if (!partStillActive) {
           logger.error(
-            `[Backfill Events] CRITICAL: Part ${todo.partId} no longer exists after processing! ` +
-              `Data may be incomplete. Aborting migration.`,
+            `[Backfill Events Observations] CRITICAL: Part ${todo.partId} no longer active after processing — ` +
```
sumerman: This is something that belongs in "Troubleshooting & FAQ" section along with the expected steps to recover.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -311,7 +259,7 @@ export default class BackfillEventsHistoricFromParts implements IBackgroundMigra
         coalesce(t.release, '') as release,
         t.tags as tags,
         t.public as public,
-        t.bookmarked as bookmarked,
+        t.bookmarked AND (o.parent_observation_id IS NULL OR o.parent_observation_id = '') AS bookmarked,
```
sumerman: is this a "root span only" guard?

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+ *
+ * The `BackgroundMigrationManager` has no dependency model — it runs every eligible
+ * row in name order and marks each finished/failed independently, so a failed
+ * upstream step does NOT by itself stop a downstream step from running on
+ * partial data. Each downstream step calls this from its `validate()`, which
+ * routes through the manager's validation-failure path and records `failedAt`.
+ * Because the next step in turn guards on its own predecessor, a single failure
+ * transitively halts the rest of the chain.
+ */
+export async function checkPredecessorMigrationFinalized(
```
sumerman: From what I can tell recovering from this and some other failure modes requires manual state manipulation in PostgreSQL. I don't think we have it documented anywhere.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -609,16 +437,15 @@ export default class BackfillEventsHistoricFromParts implements IBackgroundMigra
         manager.stopPolling();
         state.phase = "completed";
         await this.updateState(state);
-        logger.info("[Backfill Events] All chunks completed!");
+        logger.info("[Backfill Events Observations] All parts completed!");
         return;
       }
 
-      if (manager.activeCount >= config.concurrency!) return; // At capacity
+      if (manager.activeCount >= config.concurrency!) return;
```
sumerman: nit: I find it somewhat strange that `ConcurrentQueryManager` doesn't actually control concurrency. But this has no bearing on correctness.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -706,37 +543,37 @@ export default class BackfillEventsHistoricFromParts implements IBackgroundMigra
...
       logger.info(
-        `[Backfill Events] Added recovered running query ${todo.queryId} for part ${todo.partId} to manager`,
+        `[Backfill Events Observations] Added recovered running query ${todo.queryId} for part ${todo.partId} to manager`,
       );
     }
 
-    // Schedule initial batch up to concurrency limit (minus already-running recovered queries)
     const slotsAvailable = config.concurrency! - stillRunningTodos.length;
     for (let i = 0; i < slotsAvailable; i++) {
       await scheduleNext();
```
sumerman: I believe this is a data race. As both this loop and `startPolling` may call `scheduleNext` which does `SELECT, …, UPDATE` outside of a transaction. I think we should either move state management outside of `scheduleNext` or relinquish control to `ConcurrentQueryManager` completely and `await` until aborted or done. Or remove `setInterval` from `ConcurrentQueryManager` and run its iterations from the `while` loop below.

sumerman: Asked codex to check if it exists elsewhere. Here is the result:

Yes. The same race pattern exists in two other background migrations:
createRootSpansFromTraces.ts (line 309): scheduleNext does loadState → picks first pending todo → updateState outside a transaction. It is invoked by ConcurrentQueryManager.startPolling at line 412 (line 412) and by the initial slot-fill loop at line 426 (line 426).

rewriteObservationsToPidTidSorting.ts (line 437): same shape. scheduleNext mutates JSON state at line 459 (line 459), persists it at line 463 (line 463), and is called both from polling at line 540 (line 540) and the slot-fill loop at line 554 (line 554).

The shared trigger is backfillBase.ts (line 50): the interval callback calls scheduleNext after completion/failure at line 63 (line 63) and line 79 (line 79). isPolling serializes interval ticks, but it does not serialize against external calls to the same scheduleNext.

worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts
```diff
@@ -0,0 +1,764 @@
...
+        t.public,
+        t.name AS trace_name,
+        coalesce(t.user_id, '') AS user_id,
+        coalesce(t.session_id, '') AS session_id
+      FROM traces t
+      WHERE t.project_id IN {projectIds: Array(String)}
+        AND t.id IN {traceIds: Array(String)}
+        AND t.timestamp >= {minTime: DateTime64(3)}
+        AND t.timestamp <= {maxTime: DateTime64(3)}
+      ORDER BY t.event_ts DESC
```
sumerman: shouldn't this ORDER BY be more aligned with `traces` table order or do we not expect large volumes of rows from individual executions?

sumerman (commented):
I will do another pass tomorrow over `worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts` everything else is reviewed. 

I believe there is the same data race in all migrations that rely on concurrent query manager. I am also not a fan how they all follow very similar structure that is not abstracted away. It makes reviewing harder as individual migration logic is intermingled with query dispatch logic and state management.

Steffen911:
📦 The V4 **background migrations** (data-movement logic) have been extracted into a stacked PR for focused review: #14017.

This PR now retains only the removals of the old backfill scripts and the `events_only` read-path / `LANGFUSE_MIGRATION_V4_WRITE_MODE` changes. The migration chain (Prisma migrations, `createRootSpansFromTraces` / `rewriteObservationsToPidTidSorting` / `backfillEventsFullFromObservations` / `backfillEventsFullFromDatasetRunItems` / `dropPidTidSortingTables`, `BackgroundMigrationManager` env-gate support, and `LANGFUSE_BACKGROUND_MIGRATION_*` gates) lives in #14017, stacked on top of this branch.

The split is lossless — base + #14017 is byte-for-byte identical to the prior combined branch HEAD.

.env.dev.example
Steffen911: Will flip the default for it. Add to scaling page that this can be enabled for speedup on version xx.y

worker/src/backgroundMigrations/utils/backfillBase.ts
Steffen911: attemptNumber

.env.dev.example
Steffen911: @sumerman I've flipped the default and added a note in https://docs.google.com/document/d/1bvz3FUFn3T4rfJ_U1qIh36FcJGfUVQwdAIOAHa3wPnA/edit?tab=t.0#heading=h.wn2wqzyr013v that this is available from 26.4 onwards and can be opted-in.

worker/src/backgroundMigrations/utils/backfillBase.ts
Steffen911: Moved

worker/src/backgroundMigrations/utils/backfillBase.ts
Steffen911: Moved.

Steffen911:
@claude review

PR #13633, opened 2026-05-14
fix(datasets): guard prototype keys in dataset json

Steffen911:
## Summary
- reject dangerous JSON keys in dataset metadata and schemas before writes
- strip dangerous keys from existing dataset rows before tRPC list responses are serialized
- add server tests for rejected writes and legacy-row serialization behavior

## Linear
- LFE-9780

## Verification
- git commit hook: pnpm run format:check && pnpm run lint

## Not run locally
- pnpm --filter web test src/__tests__/server/datasets-trpc.servertest.ts: blocked by missing local test env variables (DATABASE_URL, NEXTAUTH_URL, SALT, CLICKHOUSE_URL, CLICKHOUSE_USER, CLICKHOUSE_PASSWORD).
- pnpm --filter web run typecheck: blocked in the fresh workspace before Prisma generation/shared build artifacts were available.

This replacement PR is based directly on main and contains only the dataset JSON guard commit.

### DIFF
--- a/web/src/features/datasets/server/actions/createDataset.ts
+++ b/web/src/features/datasets/server/actions/createDataset.ts
@@ -1,5 +1,6 @@
 import {
   DatasetNameSchema,
+  findDangerousJsonKeyPath,
   InvalidRequestError,
   Prisma,
 } from "@langfuse/shared";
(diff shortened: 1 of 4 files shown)


web/src/features/datasets/server/actions/createDataset.ts
```diff
@@ -45,6 +68,8 @@ export const upsertDataset = async ({
     );
   }
 
+  assertNoDangerousDatasetJsonKeys(input);
```
hassiebp, line 71: We would need to assert this for dataset item metadata as well

web/src/features/datasets/server/dataset-router.ts
```diff
@@ -307,6 +308,8 @@ export const datasetRouter = createTRPCRouter({
           expectedOutputSchema: true,
         },
       });
+
+      return stripDangerousJsonKeys(datasets);
```
hassiebp, line 312: We would need to strip for public API request as well, right?

hassiebp (approved):

marliessophie (commented):
Thanks! 

Write (CU): I'd like to ask if we can generalize this fix to all `Record<string, any>` data ingested into the platform; rather than adding it to all CU dataset / dataset item routes? 

Read: keep logic read-time serializer (as you suggested).

Steffen911:
This doesn't seem like the way to go as we may want a more generic solution. Will close.

PR #13636, opened 2026-05-14
feat(onboarding): land-users-in-default-project

annabellscha:
## Summary

This PR makes first-time Cloud onboarding feel complete by auto-provisioning a starter organization and starter project for eligible new users, then routing them into the tracing empty state of that project.

## What changed

- auto-create a starter organization and starter project for new Cloud users who do not already belong to a real non-demo org
- keep invited users from getting duplicate starter resources by processing invitations before starter provisioning
- reuse a shared onboarding service to:
  - detect whether a user already has a real non-demo org
  - create starter resources only when needed
  - resolve the correct onboarding redirect target
- extract org-creation permission checks into a shared `canCreateOrganizations` helper and reuse it from auth/session logic and starter provisioning
- change `/project/[projectId]` so empty projects redirect to `/project/[projectId]/traces`
- preserve the existing dashboard home behavior for non-empty projects
- show a one-time invite-members prompt on onboarding-created starter projects by reusing the existing invite dialog on the traces page
- only clear that invite prompt after the server successfully persists the change

## Behavior / edge cases

- invited users do not get an extra starter org/project
- demo org membership does not count as a real org for starter provisioning
- enterprise/self-host org-creation restrictions are still respected
- existing accessible projects route through `/project/[id]`, so they keep the normal home-vs-traces behavior
- manual setup and the onboarding survey are unchanged in this PR

## Impacted packages

- `web`

This PR auto-provisions a starter organization and project for eligible new Cloud users at sign-up time and redirects them through the onboarding survey directly into the tracing empty state, replacing the prior behavior of landing on `/` after the survey.

- **Provisioning at signup**: `createProjectMembershipsOnSignup` now calls `provisionStarterOrganizationForNewUser` (inside a serialized `SELECT FOR UPDATE` transaction) after processing invitations, preventing duplicate starter resources for invited users and concurrent signups.
- **Onboarding router**: A new `onboarding.complete` tRPC mutation resolves the correct redirect target from existing DB memberships; a new `consumeStarterProjectInvitePrompt` mutation clears the one-time invite-dialog flag.
- **Project home routing**: `/project/[id]` now runs a GSSP that redirects empty projects to `/traces` and falls back to a DB membership check for stale sessions.
</details>

Safe to merge after addressing the invite dialog close behavior; all provisioning, auth, and routing logic is well-guarded.

The starter provisioning, onboarding routing, and GSSP changes are well-structured and guarded. The one actionable defect is in traces/index.tsx: the controlled open prop on CreateProjectMemberButton is derived from hasHandledStarterInvitePrompt, which only flips to true inside the finally block of an async network call. This means the invite dialog stays visually stuck open while the consumeStarterProjectInvitePrompt request is in-flight.

web/src/pages/project/[projectId]/traces/index.tsx - the invite dialog close timing
</details>

```mermaid
flowchart TD
    A[New user signs up] --> B[createProjectMembershipsOnSignup]
    B --> C[processMembershipInvitations]
    C --> D{isCloudDeployment and new user?}
    D -- Yes --> E[provisionStarterOrganizationForNewUser]
    E --> F[SELECT FOR UPDATE user row]
    F --> G{real org memberships count}
    G -- zero --> H[create org and project in transaction]
    H --> I[write audit logs]
    G -- greater than zero --> J[skip provisioning]
    D -- No --> J
    J --> K[Session issued]
    I --> K
    K --> L[User reaches OnboardingSurvey]
    L --> M[onboarding.complete mutation]
    M --> N{starter project found in DB?}
    N -- Yes --> O[redirect to traces URL]
    N -- No, can create orgs --> P[redirect to setup]
    N -- No, cannot create orgs --> Q[redirect to home]
    O --> R[updateSession to refresh JWT]
    R --> S[navigate to traces page]
    S --> T{showInviteMembersPrompt active?}
    T -- Yes --> U[open invite dialog]
    U --> V[user dismisses]
    V --> W[consumeStarterProjectInvitePrompt]
    W --> X[set flag to false in DB]
    X --> Y[close dialog]
    T -- No --> Z[render traces normally]
```
</details>

`````markdown

---

### Issue 1 of 1
web/src/pages/project/[projectId]/traces/index.tsx:75-87
The invite dialog stays visually open while the network call runs. `CreateProjectMemberButton` is a controlled Dialog (`open={showStarterProjectInvitePrompt}`), so in Radix UI controlled mode the dialog won't close until the parent sets `open` to `false`. `showStarterProjectInvitePrompt` only becomes `false` after `setHasHandledStarterInvitePrompt(true)` is called — which happens in the `finally` block, i.e., only after `consumeStarterProjectInvitePrompt` and `updateSession()` both complete. Depending on network latency this can be several hundred milliseconds where the dialog appears stuck open after the user dismisses it.

```suggestion
  const starterProjectInvitePrompt =
    showStarterProjectInvitePrompt && organization && project ? (
      <CreateProjectMemberButton
        orgId={organization.id}
        project={{ id: project.id, name: project.name }}
        hideTrigger
        open={showStarterProjectInvitePrompt}
        onOpenChange={(open) => {
          if (!open) {
            setHasHandledStarterInvitePrompt(true);
            void handleConsumeStarterProjectInvitePrompt();
          }
        }}
      />
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;main&#39; into feature/mkt-213..."](https://github.com/langfuse/langfuse/commit/4c8d63a6fa80f8e176007cd8f26a3942173c5e3f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32765889)</sub>

### CODE REVIEW

**Comment by bezbac** [comment]
@greptileai

File: web/src/server/auth.ts [hunk 1]
```diff
@@ -668,7 +648,7 @@ const extendedPrismaAdapter: Adapter = {
     // This is idempotent - won't duplicate or overwrite existing memberships
     const user = await prisma.user.findUnique({
       where: { id: data.userId },
-      select: { id: true, email: true },
+      select: { id: true, email: true, name: true },
```
bezbac, line 653: I'm not sure if this should have changed.

web/src/features/auth/lib/createProjectMembershipsOnSignup.ts
```diff
@@ -157,6 +160,13 @@ export async function createProjectMembershipsOnSignup(
     if (user.email) await processMembershipInvitations(user.email, user.id);
 
     if (isCloudDeployment && (options?.userWasJustCreated || isNewUser)) {
+      await provisionStarterOrganizationForNewUser({
+        prisma,
+        userId: user.id,
+        userName: user.name,
+        canCreateOrganizations: canCreateOrganizations(user.email),
+      });
```
bezbac, line 175: I think we should only create a starter organization and project if the user was not invited. If I understand it correctly, the invite logic is handled in `processMembershipInvitations`. Maybe it makes sense to return some information from `processMembershipInvitations` so that `provisionStarterOrganizationForNewUser` does not get called if the user was invited?

web/src/features/onboarding/components/OnboardingSurvey.tsx
```diff
@@ -11,16 +12,47 @@ import {
...
+import { api } from "@/src/utils/api";
 import { useSurveyForm } from "../hooks/useSurveyForm";
 import type { SurveyFormData } from "../lib/surveyTypes";
 
 export function OnboardingSurvey() {
   const router = useRouter();
+  const { update: updateSession } = useSession();
   const { form, handleSubmit } = useSurveyForm();
+  const completeOnboardingMutation = api.onboarding.complete.useMutation();
+  const [isFinishingOnboarding, setIsFinishingOnboarding] = useState(false);
```
bezbac: Nit: There is a `useWatchedPromiseCallback` that you can use to manage the `isFinishingOnboarding` loading state for you

web/src/pages/project/[projectId]/traces/index.tsx
```diff
@@ -11,12 +12,80 @@ import {
...
   const router = useRouter();
   const projectId = router.query.projectId as string;
+  const { data: session, update: updateSession } = useSession();
   const { isBetaEnabled, isInitializing } = useV4Beta();
-  const { project } = useQueryProject();
+  const { project, organization } = useQueryProject();
+  const [hasHandledStarterInvitePrompt, setHasHandledStarterInvitePrompt] =
+    useState(false);
+  const [isConsumingStarterInvitePrompt, setIsConsumingStarterInvitePrompt] =
+    useState(false);
```
Comment (bezbac) [inline web/src/pages/project/[projectId]/traces/index.tsx]: Same nitpick as before, you could use `useWatchedPromiseCallback` to avoid having to manage the loading state yourself. It's fine either way as we have a bunch of places where we are handling it like we do here.

web/src/pages/project/[projectId]/index.tsx
```diff
@@ -354,3 +358,46 @@ export default function Dashboard() {
     </DashboardQuerySchedulerProvider>
   );
 }
+
+export const getServerSideProps: GetServerSideProps = async (context) => {
```
Comment (bezbac) [inline web/src/pages/project/[projectId]/index.tsx]: Why was this added? This could potentially be problematic as it will do an additional server side lookup for every project route URL visit. Maybe there is a better solution to accoplish what this is supposed to do.

web/src/pages/project/[projectId]/traces/index.tsx
```diff
@@ -11,12 +12,80 @@ import {
...
+        project={{ id: project.id, name: project.name }}
+        hideTrigger
+        open={showStarterProjectInvitePrompt}
+        onOpenChange={(open) => {
+          if (!open) {
+            void handleConsumeStarterProjectInvitePrompt();
+          }
+        }}
+      />
+    ) : null;
```
Comment (bezbac) [inline web/src/pages/project/[projectId]/traces/index.tsx]: I think it might make sense to abstract some of the added code here into a new component. I think the `isConsumingStarterInvitePrompt` and `handleConsumeStarterProjectInvitePrompt` should not live top level in the page.

web/src/features/rbac/components/CreateProjectMemberButton.tsx
```diff
@@ -50,9 +50,21 @@ const formSchema = z.object({
...
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(false);
+  const isControlled = props.open !== undefined;
+  const open = isControlled ? props.open : uncontrolledOpen;
+  const setOpen = (nextOpen: boolean) => {
+    if (!isControlled) {
+      setUncontrolledOpen(nextOpen);
+    }
+
+    props.onOpenChange?.(nextOpen);
+  };
```
bezbac: It's a bit awkward that this button now has both a controlled and uncontrolled API in the same component. Maybe reconsider if you actually need to change this or if the previous state was fine and if you _need_ to change it, make the base component `CreateProjectMemberButton` controlled by default and move the state into a wrapper e.g. `UncontrolledCreateProjectMemberButton` which would hold the state. That way the API is a lot cleaner.

web/src/features/onboarding/lib/starterProjectMetadata.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  return {
+    ...metadataRecord,
+    [ONBOARDING_STARTER_PROJECT_METADATA_KEY]: {
+      ...starterMetadata,
+      showInviteMembersPrompt: false,
+    },
+  };
+};
```
bezbac: Do we have other places in the code where we have these "metadata" fields? I'm not sure if using a normal database column might be better for this.

web/src/features/onboarding/server/onboardingRouter.ts
```diff
@@ -0,0 +1,90 @@
...
+            metadata: project.metadata,
+            userId: ctx.session.user.id,
+          }),
+        },
+      });
+
+      return {
+        updated: true,
+      };
+    }),
```
bezbac: Why don't we just "consume" the prompt when a user is invited? Why is this a separate thing?

web/src/features/onboarding/server/onboardingService.ts
```diff
@@ -0,0 +1,243 @@
...
+  canCreateOrganizations,
+}: {
+  prisma: PrismaClient;
+  userId: string;
+  userName?: string | null;
+  canCreateOrganizations: boolean;
+}) => {
+  if (!canCreateOrganizations) {
+    return null;
+  }
```
bezbac: Do we need to call pass `canCreateOrganizations` just to return early? I think it would be better simply not to call `provisionStarterOrganizationForNewUser` in the first place if `canCreateOrganizations` is false. That way you avoid this indirection and can remove the argument.

web/src/features/onboarding/server/onboardingService.ts
```diff
@@ -0,0 +1,243 @@
...
+    : {}),
+});
+
+export const getRealOrganizationMemberships = async ({
+  prisma,
+  userId,
+}: {
+  prisma: Pick<PrismaClient, "organizationMembership">;
+  userId: string;
+}): Promise<RealOrganizationMembership[]> =>
```
bezbac: You don't need this `: Promise<RealOrganizationMembership[]>` return type. Typescript can infer it.

web/src/features/onboarding/server/onboardingService.ts
```diff
@@ -0,0 +1,243 @@
...
+        ],
+      },
+    },
+  },
+} satisfies Prisma.OrganizationMembershipInclude;
+
+export type RealOrganizationMembership =
+  Prisma.OrganizationMembershipGetPayload<{
+    include: typeof realOrganizationMembershipInclude;
+  }>;
```
bezbac: Consider using ReturnType<typeof getRealOrganizationMemberships> here so you don't need to specify this large and brittle type. It would also allow you to inline `realOrganizationMembershipInclude` which makes the entire file more readable.

web/src/features/onboarding/server/onboardingService.ts
```diff
@@ -0,0 +1,243 @@
...
+    `;
+
+    const realOrganizationMembershipCount =
+      await tx.organizationMembership.count({
+        where: getRealOrganizationMembershipWhere(userId),
+      });
+
+    if (realOrganizationMembershipCount > 0) {
+      return null;
+    }
```
bezbac, line 356: I'm not sure if this is possible but maybe this could re-use `getRealOrganizationMemberships`. That way you can inline `getRealOrganizationMembershipWhere`.

web/src/server/auth.ts
annabellscha: name: true is needed here because this user object is passed into createProjectMembershipsOnSignup.ts, and that flow may auto-create a starter organization for a user who signs in through linkAccount.

The starter-org creation logic in onboardingService.ts uses user.name to derive a personalized org name via getStarterOrganizationName(...), e.g. Taylor's Organization, with My Organization only as a fallback.

PR #13665, opened 2026-05-15
fix(web): preserve trace URL filters when opening shared links in new tab

pratimapatel2008:
## What does this PR do?
When the traces table (and other tables using useTableViewManager) loaded without viewId in the URL but with a filter query param, bootstrap still restored a saved view from session storage (or applied the project's default view). That path called setFilters with the saved view’s filters and overwrote URL-driven sidebar filters, so copying a ?filter=… link and opening it in a new tab could lose filters while session still held a last-used viewId.

This change skips saved-view bootstrap when the router has a non-empty filter and no viewId, so the URL remains the source of truth for shareable filter links. viewId permalinks are unchanged.

Adds a Vitest regression in savedViewImplicitEnvironmentRegression.clienttest.tsx.

Fixes # (issue)

Before Fix: https://www.loom.com/share/0715f73156854fa8a1c926cf59bace4a
After Fix: https://www.loom.com/share/86f59f77734646c192af07048164b9ab

## Type of change

## Mandatory Tasks

## Checklist

This PR fixes a bug where opening a shared `?filter=…` link in a new tab could silently lose its filters when the browser session still held a `storedViewId` from a previous visit. The bootstrap effect in `useTableViewManager` now exits early — preserving the URL as the source of truth — when `router.query.filter` is present and no `viewId` is in the URL.

- **`useTableViewManager.ts`**: Adds `firstRouterQueryString` helper and a new early-return guard in the bootstrap effect; adds `router.query.filter` to the effect dependency array.
- **`savedViewImplicitEnvironmentRegression.clienttest.tsx`**: Adds a Vitest regression suite that seeds session storage with a saved-view ID, sets a `filter` query param on the router mock, and asserts the URL filter is preserved.
</details>

The change is a well-scoped early-return in the bootstrap effect that guards URL filter links against being overwritten by session-restored views; it does not affect the viewId permalink path.

The fix correctly inserts the guard before both session-storage and default-view bootstrap paths, and the new test validates the targeted regression. The only gap is that the test suite doesn't exercise the scenario where the project has a defaultViewId (as opposed to a session storedViewId), leaving a slight blind spot in the Priority 2 bootstrap path.

The test file savedViewImplicitEnvironmentRegression.clienttest.tsx would benefit from an additional scenario covering the default-view path.
</details>

```mermaid
flowchart TD
    A([Bootstrap effect fires]) --> B{disabled or\nisInitialized?}
    B -- Yes --> Z([return / no-op])
    B -- No --> C{router ready?}
    C -- No --> Z
    C -- Yes --> D{selectedViewId\nin URL?}
    D -- Yes, non-system --> E([Wait for getById query])
    D -- Yes, system preset --> F([Clear viewId, return])
    D -- No --> G{router.query.filter\nnon-empty? NEW}
    G -- Yes --> H([setIsInitialized + setIsLoading=false\nURL filter wins])
    G -- No --> I{storedViewId\nin session?}
    I -- Yes --> J([setSelectedViewId = storedViewId])
    I -- No --> K{defaultViewId\nloading?}
    K -- Yes --> Z
    K -- No, exists --> L([setSelectedViewId = defaultViewId])
    K -- No, null --> M([setIsInitialized + setIsLoading=false])
```
</details>

`````markdown

---

### Issue 1 of 2
web/src/components/table/table-view-presets/hooks/useTableViewManager.ts:180
The `explicitFilterQuery.length > 0` check is redundant: a truthy string is by definition non-empty, so the condition collapses to just the truthiness guard and the `!selectedViewId` check.

```suggestion
    if (explicitFilterQuery && !selectedViewId) {
```

### Issue 2 of 2
web/src/features/filters/savedViewImplicitEnvironmentRegression.clienttest.tsx:504-570
Missing test coverage for the default-view path: the new test suite only seeds session storage with a `storedViewId` (`view-1`) but sets `mockGetDefaultUseQuery` to return `null`. A second scenario where session storage is empty but the project has a `defaultViewId` would confirm that the same bootstrap skip protects against overwrite via the Priority 2 path as well.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): preserve trace URL filters whe..."](https://github.com/langfuse/langfuse/commit/4bc1321e595df8f49a5351adad49305525a0362e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32363879)</sub>


nimarb:
@claude review

PR #13677, opened 2026-05-18
feat(monitors): add Monitor Prisma schema and migration

marksalpeter:
## Summary
The Monitoring & Alerting feature ([RFC](https://linear.app/langfuse/document/rfc-monitoring-and-alerting-653cc4349c29)) needs a Postgres schema and migration before any service, scheduler, or tRPC code can land.

To achieve this we:
- Added the `Monitor` Prisma model and four supporting enums (`MonitorThresholdOperator`, `MonitorSeverity`, `MonitorStatus`, `MonitorView`). `MonitorSeverity` includes `UNKNOWN` as the row-insert default — distinct from `OK` ("evaluated and healthy") and `NO_DATA` ("evaluated and no data found"), per [RFC §Severity State Machine](https://linear.app/langfuse/document/rfc-monitoring-and-alerting-653cc4349c29#severity-state-machine). `MonitorView` is a dedicated subset of data sources (`OBSERVATIONS`, `SCORES_NUMERIC`, `SCORES_CATEGORICAL`) — the query builder is shared logic but the entity-level view enum is feature-owned.
- Wired the inverse relation `monitors` on `Project` with `onDelete: Cascade`, and `createdByUser` / `updatedByUser` FKs on `User` with `onDelete: SetNull`, mirroring `DashboardWidget` / `Dashboard` / `TableViewPreset`.
- Added indexes for the three hot read paths: project-scoped listing (`project_id`), processor batch read (`scheduler_batch_id`), and the scheduler tick range scan (`next_run_at, scheduler_batch_id`).

**Note**: schema-only PR. Zod schemas, `MonitorService`, tRPC router, scheduler, queue processor, and UI ship in follow-up tickets under the same project. The scheduler tick's modulo predicate (`scheduler_batch_id % N = K`) is not sargable against a B-tree, so the composite index only accelerates the `next_run_at` range scan and covers the modulo as an index-only filter — it does not narrow by shard. The optimal sharded index (expression index, generated `shard_index` column, or per-shard partials) will be decided alongside [LFE-9812](https://linear.app/langfuse/issue/LFE-9812/add-monitorscheduler-with-slot-based-batching) (MonitorScheduler) where the workload can be measured.

Fixes [LFE-9807](https://linear.app/langfuse/issue/LFE-9807/add-monitor-prisma-schema-and-migration)

## Verification

- [x] `pnpm --filter @langfuse/shared run db:generate`
- [x] `pnpm --filter @langfuse/shared run typecheck`
- [x] `pnpm --filter @langfuse/shared exec prisma migrate deploy` — migration applies cleanly forward on a populated dev DB after rollback
- [x] `Project.monitors`, `User.monitorsCreated`, `User.monitorsUpdated` inverse relations visible on the generated PrismaClient

## Test plan

- [x] Apply the migration on a populated DB; confirm no errors.
- [x] Create a `Monitor`, delete its `Project`, confirm the monitor cascades (count 1 → 0 after `DELETE FROM projects`).
- [x] Create a `Monitor` with `createdBy = $userId`, delete the user, confirm `monitors.created_by` is `NULL` (SetNull verified).
- [x] Insert a `Monitor` without setting `severity`; row stores `UNKNOWN` (and `status` defaults to `ACTIVE`).
- [x] Confirm `MonitorView` enum contains exactly `OBSERVATIONS`, `SCORES_NUMERIC`, `SCORES_CATEGORICAL`. Attempted `INSERT ... view='TRACES'` rejected with `ERROR: invalid input value for enum "MonitorView": "TRACES"`.
- [x] `EXPLAIN ANALYZE` the scheduler tick query (`WHERE next_run_at <= now() AND scheduler_batch_id % N = K`) on 10k seeded rows: planner picks `Index Scan using monitors_scheduler_tick_idx`, with `Index Cond: (next_run_at <= now())` and `Filter: ((scheduler_batch_id % 10) = 3)` — exactly the sargable/non-sargable split described above.

### DIFF
--- a/packages/shared/prisma/migrations/20260518075046_add_monitors/migration.sql
+++ b/packages/shared/prisma/migrations/20260518075046_add_monitors/migration.sql
@@ -1,0 +1,53 @@
+-- CreateEnum
+CREATE TYPE "MonitorThresholdOperator" AS ENUM ('GT', 'GTE', 'LT', 'LTE', 'EQ', 'NEQ');
+
+-- CreateEnum
+CREATE TYPE "MonitorSeverity" AS ENUM ('OK', 'WARNING', 'ALERT', 'NO_DATA');
+
+-- CreateEnum
+CREATE TYPE "MonitorStatus" AS ENUM ('ACTIVE', 'PAUSED', 'ERROR_BAD_QUERY');
+
+-- CreateTable
+CREATE TABLE "monitors" (
+    "id" TEXT NOT NULL,
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "created_by" TEXT,
+    "updated_by" TEXT,
+    "project_id" TEXT NOT NULL,
+    "view" "DashboardWidgetViews" NOT NULL,
+    "filters" JSONB NOT NULL,
+    "metric" JSONB NOT NULL,
+    "window_ms" BIGINT NOT NULL,
+    "cadence_ms" BIGINT NOT NULL,
+    "threshold_operator" "MonitorThresholdOperator" NOT NULL,
+    "alert_threshold" DECIMAL(65,30) NOT NULL,
+    "warning_threshold" DECIMAL(65,30),
+    "severity" "MonitorSeverity" NOT NULL DEFAULT 'OK',
+    "severity_changed_at" TIMESTAMP(3),
+    "no_data" JSONB NOT NULL,
+    "renotify" JSONB NOT NULL,
+    "status" "MonitorStatus" NOT NULL DEFAULT 'ACTIVE',
+    "scheduler_batch_id" BIGINT NOT NULL,
+    "next_run_at" TIMESTAMP(3) NOT NULL,
+    "last_published_run_at" TIMESTAMP(3),
+    "last_completed_run_at" TIMESTAMP(3),
+    "name" TEXT NOT NULL,
+    "message" TEXT NOT NULL DEFAULT '',
+    "tags" TEXT[] DEFAULT ARRAY[]::TEXT[],
+    "alerted_at" TIMESTAMP(3),
+
+    CONSTRAINT "monitors_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "monitors_project_id_idx" ON "monitors"("project_id");
+
+-- CreateIndex
+CREATE INDEX "monitors_scheduler_batch_id_idx" ON "monitors"("scheduler_batch_id");
+
+-- CreateIndex
+CREATE INDEX "monitors_scheduler_tick_idx" ON "monitors"("next_run_at", "scheduler_batch_id");
+
+-- AddForeignKey
+ALTER TABLE "monitors" ADD CONSTRAINT "monitors_project_id_fkey" FOREIGN KEY ("project_id") REFERENCES "projects"("id") ON DELETE CASCADE ON UPDATE CASCADE;
--- a/packages/shared/prisma/migrations/20260518092907_add_monitors/migration.sql
+++ b/packages/shared/prisma/migrations/20260518092907_add_monitors/migration.sql
@@ -53,4 +53,10 @@ CREATE INDEX "monitors_scheduler_batch_id_idx" ON "monitors"("scheduler_batch_id
 CREATE INDEX "monitors_scheduler_tick_idx" ON "monitors"("next_run_at", "scheduler_batch_id");
 
 -- AddForeignKey
+ALTER TABLE "monitors" ADD CONSTRAINT "monitors_created_by_fkey" FOREIGN KEY ("created_by") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "monitors" ADD CONSTRAINT "monitors_updated_by_fkey" FOREIGN KEY ("updated_by") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+-- AddForeignKey
 ALTER TABLE "monitors" ADD CONSTRAINT "monitors_project_id_fkey" FOREIGN KEY ("project_id") REFERENCES "projects"("id") ON DELETE CASCADE ON UPDATE CASCADE;
--- a/packages/shared/prisma/schema.prisma
+++ b/packages/shared/prisma/schema.prisma
@@ -70,6 +70,8 @@ model User {
   dashboardUpdated          Dashboard[]                 @relation("UpdatedByUser")
   tableViewPresetCreated    TableViewPreset[]           @relation("CreatedByUser")
   tableViewPresetUpdated    TableViewPreset[]           @relation("UpdatedByUser")
+  monitorsCreated           Monitor[]                   @relation("CreatedByUser")
+  monitorsUpdated           Monitor[]                   @relation("UpdatedByUser")
   annotationQueueAssignment AnnotationQueueAssignment[]
   surveys                   Survey[]
   commentReactions          CommentReaction[]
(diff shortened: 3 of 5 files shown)


packages/shared/prisma/migrations/20260518075046_add_monitors/migration.sql
```diff
@@ -0,0 +1,53 @@
...
+    "project_id" TEXT NOT NULL,
+    "view" "DashboardWidgetViews" NOT NULL,
+    "filters" JSONB NOT NULL,
+    "metric" JSONB NOT NULL,
+    "window_ms" BIGINT NOT NULL,
+    "cadence_ms" BIGINT NOT NULL,
+    "threshold_operator" "MonitorThresholdOperator" NOT NULL,
+    "alert_threshold" DECIMAL(65,30) NOT NULL,
+    "warning_threshold" DECIMAL(65,30),
+    "severity" "MonitorSeverity" NOT NULL DEFAULT 'OK',
```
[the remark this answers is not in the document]
marksalpeter: Same point as the schema thread above — pushing back. The migration default tracks the Prisma schema default by design, and the schema default of `OK` is required by the RFC's cold-start emit rule.

packages/shared/prisma/migrations/20260518092907_add_monitors/migration.sql
```diff
@@ -0,0 +1,53 @@
...
+);
+
+-- CreateIndex
+CREATE INDEX "monitors_project_id_idx" ON "monitors"("project_id");
+
+-- CreateIndex
+CREATE INDEX "monitors_scheduler_batch_id_idx" ON "monitors"("scheduler_batch_id");
+
+-- CreateIndex
+CREATE INDEX "monitors_scheduler_tick_idx" ON "monitors"("next_run_at", "scheduler_batch_id");
```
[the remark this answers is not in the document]
marksalpeter: Correct on the sargability — `scheduler_batch_id % N = K` is a row-level Filter, not an index-narrowing seek. The composite still earns its keep on the `next_run_at` range scan and as a covering column for the modulo filter (index-only scan), but it does not act like a sharded scheduler index.

Updating the PR description to drop the "shard-pruning" implication. The optimal shard index (generated `shard_index` column, expression index, or per-shard partials) depends on whether `LANGFUSE_MONITOR_SCHEDULERS` is fixed or dynamic at runtime — that's a scheduler-PR decision, not a schema-PR one. Will revisit alongside [LFE-9812](https://linear.app/langfuse/issue/LFE-9812/add-monitorscheduler-with-slot-based-batching) (MonitorScheduler), where the actual tick query and workload can be measured.

packages/shared/prisma/migrations/20260518075046_add_monitors/migration.sql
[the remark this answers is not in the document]
marksalpeter: Actually we're revising the rfc to have an explicit UNKNOWN type. Good note.

packages/shared/prisma/schema.prisma
```diff
@@ -1577,6 +1578,77 @@ model AutomationExecution {
...
+
+// Scheduler evaluates only ACTIVE; any other state is skipped.
+enum MonitorStatus {
+  ACTIVE
+  PAUSED
+  ERROR_BAD_QUERY
+}
+
+model Monitor {
+  id        String   @id @default(uuid())
```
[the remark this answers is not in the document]
marksalpeter: Fixed in a946bdaf5. You're right — the repo convention is overwhelmingly cuid, and the uuid choice came from a memory rule about populated-table column adds that doesn't apply to a brand-new table.

packages/shared/prisma/schema.prisma
```diff
@@ -1577,6 +1578,77 @@ model AutomationExecution {
...
+  createdAt DateTime @default(now()) @map("created_at")
+  updatedAt DateTime @default(now()) @updatedAt @map("updated_at")
+  createdBy String?  @map("created_by")
+  updatedBy String?  @map("updated_by")
```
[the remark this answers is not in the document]
marksalpeter: Fixed in a946bdaf5. Wired `createdByUser`/`updatedByUser` relations with `onDelete: SetNull` and added the inverse `monitorsCreated`/`monitorsUpdated` arrays on `User`, matching the DashboardWidget/Dashboard/TableViewPreset pattern. Migration now adds the two FK constraints.

packages/shared/prisma/schema.prisma
```diff
@@ -1577,6 +1580,86 @@ model AutomationExecution {
...
+  project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+  // === Query shape (mirrors DashboardWidget, minus TRACES) ===
+  view    MonitorView
+  filters Json
+  metric  Json
+
+  // === Monitor Evaluation ===
+  windowMs          BigInt                   @map("window_ms")
+  cadenceMs         BigInt                   @map("cadence_ms")
```
Steffen911, line 1633: I feel like regular ints should be fine here. Not sure how they are stored underneath, but I cannot think of a window or cadence in MS that would exceed a regular Int32 or Int64.

packages/shared/prisma/schema.prisma
```diff
@@ -1577,6 +1580,86 @@ model AutomationExecution {
...
+  thresholdOperator MonitorThresholdOperator @map("threshold_operator")
+  alertThreshold    Decimal                  @map("alert_threshold")
+  warningThreshold  Decimal?                 @map("warning_threshold")
+
+  severity          MonitorSeverity @default(UNKNOWN)
+  severityChangedAt DateTime?       @map("severity_changed_at")
+
+  noData   Json @map("no_data")
+  renotify Json
```
Steffen911, line 1642: What gets stored in the Jsons here? Perhaps add a comment with the intended value shape.

packages/shared/prisma/schema.prisma
```diff
@@ -1577,6 +1580,86 @@ model AutomationExecution {
...
+
+  // === Alerts ===
+  name      String
+  message   String    @default("")
+  tags      String[]  @default([])
+  alertedAt DateTime? @map("alerted_at")
+
+  @@index([projectId])
+  @@index([schedulerBatchId])
+  @@index([nextRunAt, schedulerBatchId], map: "monitors_scheduler_tick_idx")
```
Steffen911, line 1659: Does this combination make sense? Why don't we have only the first or the second one at access time?

Steffen911: If we go with `schedulerBatchId` only (IMO sufficient) and expect that we query only within a single project, we should actually consider `[projectId, schedulerBatchId]` as index which would've been my first intuition.

packages/shared/prisma/schema.prisma
marksalpeter: `Int` is an `int32`. Max window would be 24-25days - not big enough.

If we go with `BigInt` / `int64`, max window is 300k years - too big but sufficient for the 1 / 2 month and custom windowing use cases that might be added.

`cadence` could probably safely be an `Int`, but just in case we open up evaluation windows and want there to be some kind of monthly cadence, I will keep it for now.

packages/shared/prisma/schema.prisma
marksalpeter: It will be in the zod. I feel like thats unambiguous enough, no? [Ref](https://linear.app/langfuse/document/rfc-monitoring-and-alerting-653cc4349c29#schema-zod-4d8a5416)

packages/shared/prisma/schema.prisma
marksalpeter: We need the `schedulerBatchId` index for the monitor queue processor. [Ref](https://linear.app/langfuse/document/rfc-monitoring-and-alerting-653cc4349c29#monitor-queue-processor-04970bb3)
```
SELECT … FROM monitors
WHERE scheduler_batch_id = $batchId
FOR UPDATE SKIP LOCKED;
```

We need an index on `nextRunAt` for the monitor scheduler itsself. Including `schedulerBatchId` prevents pg from needing a second heap fetch [Ref](https://linear.app/langfuse/document/rfc-monitoring-and-alerting-653cc4349c29#example-sql-94d9505d) 
```
SELECT … FROM monitors
WHERE next_run_at <= $tick
  AND (scheduler_batch_id % $N) = $K
FOR UPDATE SKIP LOCKED;
```
From the explain:
```
Index Scan using monitors_scheduler_tick_idx
  Index Cond: (next_run_at <= now())          ← seek
  Filter: ((scheduler_batch_id % 10) = 3)     ← evaluated from index tuple
  Rows Removed by Filter: 45                  ← these never touched the heap
  Buffers: shared hit=4                       ← only 4 pages, not 50+
```

PR #13686, opened 2026-05-18
feat(experiments): support charting in experiment list

marliessophie:
### Scope 

https://github.com/user-attachments/assets/13f8cdbc-617b-4af3-a1b4-7aa1a84c7317

#### Note 
Trace-level score dashboarding on the /experiments page is intentionally out of scope in this PR. Observation-level and run-level scores are supported; trace-level scores need a dedicated, performant scores.trace_id to selected experiment events mapping. We will consider adding this in a follow up PR. 

This PR adds a collapsible charts section to the experiment list view, showing cost, latency, and score metrics grouped by experiment name using a new `entityDimension` concept in the query builder.

- **New `entityDimension` query feature**: extends `QueryBuilder` and `QueryType` to support categorical bucketing (e.g., by `experimentName`) as an alternative to `timeDimension`. The two are validated as mutually exclusive, and `WITH FILL` is correctly skipped for entity dimensions.
- **Three-level score filter options**: `getExperimentItemsFilterOptions` runs three parallel ClickHouse queries to discover trace-, observation-, and run-level scores available for the selected experiments, feeding both the chart metric dropdown and the sidebar filter state.
- **Dynamic chart grid**: up to 4 chart slots, each with an embedded metric selector and hover-to-delete, persisted per-project in session storage.
</details>

The chart section in ExperimentsTable passes an unstable Date reference as toTimestamp when the date range has no upper bound, causing every table re-render to re-execute all chart queries unnecessarily.

The core query-builder and repository changes are solid. The main concern is in ExperimentsTable: tableDateRange.to ?? new Date() is evaluated on every render, so any local state change such as selecting rows or toggling filters produces a fresh Date object, which propagates into each ExperimentChartSlot useMemo and triggers a new API call for every visible chart.

web/src/features/experiments/components/table/ExperimentsTable.tsx (toTimestamp prop), web/src/features/widgets/components/InlineWidget.tsx (queryId fallback)
</details>

```mermaid
sequenceDiagram
    participant ET as ExperimentsTable
    participant EIF as useExperimentItemsFilterOptions
    participant CH as ClickHouse 3x parallel
    participant ECG as ExperimentChartsGrid
    participant ECS as ExperimentChartSlot
    participant WC as WidgetContent
    participant QB as QueryBuilder

    ET->>EIF: experimentIds all rows in view
    EIF->>CH: trace-level scores query
    EIF->>CH: observation-level scores query
    EIF->>CH: run-level scores query
    CH-->>EIF: ScoreFilterOptionsRow results
    EIF-->>ET: scoreFilterOptions

    ET->>ECG: experimentIds fromTimestamp toTimestamp scoreFilterOptions
    ECG->>ECS: metricId experimentIds timestamps
    ECS->>ECS: buildWidgetConfigFromId metricId
    ECS->>WC: query with entityDimension and experimentId filter
    WC->>QB: executeQuery
    QB-->>WC: rows with entity_dimension
    WC-->>ECS: Chart rendered
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/features/experiments/components/table/ExperimentsTable.tsx:694-695
The `new Date()` fallback is evaluated on **every render** of `ExperimentsTable`. When `tableDateRange.to` is `undefined` (open-ended range), each re-render triggered by row selection, filter changes, or any other local state update produces a different `Date` object. That reference propagates through `ExperimentChartsGrid` to `ExperimentChartSlot` to the `useMemo` that builds `query` (which depends on `toTimestamp`), invalidating it and enqueuing a fresh API call for every chart slot on every unrelated interaction.

```suggestion
                    fromTimestamp={tableDateRange.from}
                    toTimestamp={tableDateRange.to ?? tableDateRange.from}
```

### Issue 2 of 3
web/src/features/widgets/components/InlineWidget.tsx:186-187
`Date.now()` is evaluated in the component body (not inside a `useMemo`) so it produces a new value on every render. Any caller that omits `schedulerId` will pass a different `queryId` on each render, which will cause `useScheduledDashboardExecuteQuery` to treat every render as a brand-new query and re-execute the API call continuously. The current experiment usage always supplies `schedulerId`, but this fallback is broken for any future consumer of `WidgetContent`.

```suggestion
      // todo: pass queryId as prop
      queryId: schedulerId ?? "inline-widget:unscheduled",
```

### Issue 3 of 3
web/src/features/widgets/hooks/useWidgetQuery.ts:2105-2112
**Duplicate interface definitions** - `WidgetMetricConfig` and `WidgetDimensionConfig` are declared identically in both this file and `InlineWidget.tsx`. If one is updated the other could silently diverge. Consider re-exporting them from `InlineWidget.tsx` or a shared types file and importing from there instead.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat: implement experiment run score fil..."](https://github.com/langfuse/langfuse/commit/22a341b8db61f3424c96824cdf99c921048bf152) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32599002)</sub>

### CODE REVIEW

File: packages/shared/src/features/query/types.ts [hunk 1]
```diff
@@ -135,6 +135,10 @@ export function getValidAggregationsForMeasureType(
   ) {
     return [...metricAggregations.options];
   }
+  // datetime supports min/max for earliest/latest timestamps
+  if (measureType === "datetime") {
+    return ["min", "max", "count", "uniq"];
+  }
```
marliessophie: **Question**

Any objections?

web/src/features/widgets/components/InlineWidget.tsx
marliessophie, line 1: **Note**

Note that we plan to re-use these components in `DashboardWidget` in follow up PR. To be done in separate PR to keep current PR scope as small as possible.

web/src/features/widgets/hooks/useWidgetQuery.ts
marliessophie, line 1: **Note**

Note that we plan to re-use these components in `DashboardWidget` in follow up PR. To be done in separate PR to keep current PR scope as small as possible.

web/src/features/experiments/components/ExperimentChartSlot.tsx
marliessophie, line 1: **Question**

Suggestion: No `DashboardQuerySchedulerProvider`: Max 4 charts, bounded queries — parallel fetch is fine. Scheduler is for large dashboards; add later only if needed.

packages/shared/src/features/query/validateQuery.ts
```diff
@@ -9,6 +9,11 @@ export type QueryValidationResult =
   | { valid: true }
   | { valid: false; reason: string };
 
+const ALLOWED_HIGH_CARDINALITY_ENTITY_DIMENSIONS = new Set([
```
marliessophie: **Question**

I whitelisted relevant high-cardinality entityDimension because they are set by non-user editable experiment charts; public metrics and the generic widget UI do NOT expose arbitrary entityDimension selection. 

Note that authenticated users can still manually call the tRPC route with hardcoded input as entityDimension: { field: "datasetRunId" } or experimentName. Because these are now allowlisted, they bypass the high-cardinality/top-N guard. I'd be fine accepting this risk, but will let you be the final judge.

packages/shared/src/features/query/dataModel.ts
marliessophie, line 1: **Question** 

I added `timestamp` and `start_time` as measures so I can order my experiment by time. I did add them as `uiHidden` so they are not exposed in custom dashboarding. Please flag any design or performance constraints, if any.

marliessophie (commented):
Hey @sumerman, thanks again for review 🙏  
As discussed, goal of this PR is to allow hard-coded time-series widgets that group by experiment_name, for scores-numeric, scores-categorical and observations. 

Please pay special attention to the performance of my widgets. You can see widget defintion in `web/src/features/experiments/constants/charts.ts`. Please note that ALL widgets take a list of max 50 expeirment_ids as filter condition.  

I'd appreciate if you could please review: 
- packages/shared/src/features/query/server/queryBuilder.ts
- packages/shared/src/features/query/types.ts
- packages/shared/src/features/query/dataModel.ts
- Optional, only if you have some time/are interessted: packages/shared/src/server/repositories/experiments.ts

packages/shared/src/features/query/server/queryBuilder.ts
```diff
@@ -531,6 +532,12 @@ export class QueryBuilder {
         relationTables.add(metric.relationTable);
       }
     });
+    if (query.entityDimension) {
```
sumerman: Intuitively, I would expect this step to happen before collectRelationTables is called and entityDim to be either among the applied dimensions or as an additional argument, if doesn't belong with applied dimensions.

More broadly, I would expect `entityDimension` to be treated more like time dimension. E.g. time dimension is thoroughly propagated across joins when applicable. Since at this point only a single table has it, writing propagation logic would be wasteful. Instead I would think through consequences of such joins and potentially block them.

sumerman: https://github.com/langfuse/langfuse/pull/13686/changes#r3258469449 highlights a similar type of issue. I would like a breakdown on how you expect `entityDimension` to interact with joins.

packages/shared/src/features/query/server/queryBuilder.ts
```diff
@@ -813,6 +849,12 @@ export class QueryBuilder {
       dimensions += `${timeDimensionSql},`;
     }
 
+    // Add entity dimension if specified - reuse unified builder with any() wrapper
+    const entityDimensionSql = this.buildEntityDimensionSql(view, query, "any");
+    if (entityDimensionSql) {
+      dimensions += `${entityDimensionSql},`;
+    }
+
```
sumerman: - this is now a longer block that repeatps completely over two call sites. I would prefer it as a standalone function. 
- even if it validates elsewhere is shouldn't be possible here to emit both `timeDimensionSql` and `entityDimensionSql`; we should throw if both are true.

packages/shared/src/features/query/dataModel.ts
sumerman: I really don't like `uiHidden` concept. Even though it is not possible to see weird combinations they are still possible to access via public API and it is possible to save something UI can't show via hand crafted tTRPC call.

sumerman: I also don't like `description: "Earliest score timestamp in the group.",`. It is tied to your feature. I can see how a timestamp-like could be generally useful dimension i.e. `max(timestamp) group by provided_model_name`.

packages/shared/src/features/query/types.ts
sumerman: what is the meaning of count and uniq on timestamps? 🤔

packages/shared/src/features/query/types.ts
```diff
@@ -167,6 +172,16 @@ export const query = z
         granularity: granularities,
       })
       .nullable(),
+    // Entity dimension for bucketing by a categorical field (e.g., experimentName).
+    // IMPORTANT: Unlike timeDimension which has implicit bucket limits (24 hours/day),
+    // entityDimension has NO cardinality guarantee. Callers MUST filter the dataset
```
sumerman: I think this should be a validation at both widget save time (the refine below) AND query building time

packages/shared/src/features/query/validateQuery.ts
sumerman: There is also public API risk and it is IMO greater. Instead of a whitelist I would rather change validation to let high cardinality measures through when there is a filter on that measure.

packages/shared/src/features/query/types.ts
marliessophie: thinking about it again, not meaningful. I'll drop count and unique.

packages/shared/src/features/query/dataModel.ts
marliessophie: I'd be happy to add them as dimensions. Agree they can be helpful in some cases. Also, then we can avoid adding to the `uiHidden` pattern. 

I'll also update descriptions to be more general. 

Edit: decided to not add them in this PR after all. Found a way to not need them for my queries. And the chart currently does not handle displaying datetime data well, so moved this out of scope.

packages/shared/src/server/repositories/experiments.ts
```diff
@@ -476,6 +478,334 @@ export const getExperimentItemsCountFromEvents = async (
...
+  // Build the join condition based on level
+  const joinCondition =
+    level === "trace"
+      ? "ON s.project_id = ee.project_id AND s.trace_id = ee.trace_id"
+      : "ON s.project_id = ee.project_id AND s.trace_id = ee.trace_id AND s.observation_id = ee.span_id";
+
+  // Compose the full query using CTEQueryBuilder
+  // Excludes CORRECTION scores by whitelisting allowed data types
+  const queryBuilder = new CTEQueryBuilder()
+    .withCTE("experiment_events", {
```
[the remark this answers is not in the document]
marliessophie: this is intended.

web/src/features/experiments/components/table/ExperimentsTable.tsx
```diff
@@ -501,6 +509,15 @@ export default function ExperimentsTable({
       : [];
   }, [experiments]);
 
+  // Get experiments from the current query result (for charts)
+  const chartExperiments = useMemo(() => {
+    return rows.map((row) => ({ id: row.id, name: row.name }));
+  }, [rows]);
```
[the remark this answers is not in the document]
marliessophie: surfacing "charts reflect current page (X of Y matching)" is an optional improvement. Not strictly needed.

packages/shared/src/features/query/server/queryBuilder.ts
marliessophie: Thanks for the feedback! 

High-level changes:
- I made `entityDimension` a first-class internal bucket in QueryBuilder, parallel to `timeDimension`, while keeping it out of the generic widget builder UI and public metrics API schema. It is currently only set by hard-coded experiment chart configs, and high-cardinality entity buckets now require a finite same-field filter.

I think if you re-read the packages/shared/src/features/query/server/queryBuilder.ts and packages/shared/src/features/query/validateQuery.ts files that should give you a good picture. 

Join contract:
- `entityDimension` reuses the selected view dimension’s SQL/relation metadata, but it is not treated as a regular dimensions[] entry. If the entity field is relation-backed, QueryBuilder includes only the relation explicitly declared on the view; it does not infer broader join paths or propagate entity semantics across joins.

packages/shared/src/features/query/server/queryBuilder.ts
```diff
@@ -51,6 +51,20 @@ type MappedFilters = {
   whereRawParts: RawSqlPart[];
 };
 
+type AppliedBucketDimension =
```
sumerman: micronit: `AppliedBucketingDimension`. "Potato, potahto", but I like it more this way. Also fine to keep it as it is.

sumerman (approved):
Love what you did here! As usual — thank your for bearing with me. Only a single naming micronit, feel free to ignore.

PR #13694, opened 2026-05-18
fix(blob-storage): harden endpoint connection validation

wochinge:
## Summary

- Adds blob storage endpoint validation for S3-compatible and Azure Blob integrations to block internal/special-use targets unless explicitly allowlisted.
- Wires S3/Azure SDK clients through connection-time DNS/IP validation to defend against DNS rebinding after save-time validation.
- Keeps self-hosted enforcement opt-in via new allowlist env vars for compatibility with internal MinIO/Azure deployments; Cloud remains strict.
- Adds focused web and worker coverage for blocked and whitelisted endpoint behavior.

## Linear

- [INT-1702](https://linear.app/langfuse/issue/INT-1702)

## Major Decisions

- Self-hosted validation activates only when blob endpoint allowlist env vars are configured, with a TODO to enforce by default in the next major release.
- GCS/OCI remain out of scope for blob integrations and are explicitly disabled on the blob export worker path.

## Review Focus

- Self-hosted rollout behavior and allowlist env naming.
- S3/Azure SDK connection-time validation wiring.
- Worker export flow validation timing and StorageService tests.

This PR hardens blob storage endpoint security by adding save-time URL validation (blocking private/special-use IP ranges and hostnames) and wiring connection-time DNS/IP validation through the S3 `NodeHttpHandler` and Azure Blob Storage `RequestPolicyFactory` to defend against DNS-rebinding after save-time checks. Cloud deployments always enforce strict validation; self-hosted deployments opt in by configuring new `LANGFUSE_BLOB_STORAGE_ENDPOINT_WHITELISTED_*` env vars.

- **New `blobStorageEndpointValidation.ts`**: centralized save-time validation called at upsert, validate-test, and as a job preflight in the worker.
- **`connection.ts` refactor**: generic `getOrCreatePolicyResource` helper backing both the Undici dispatcher cache and the new `getSecureOutboundHttpAgents` API.
- **GCS and OCI explicitly disabled** in both the worker export path and the router's validate handler.
</details>

Safe to merge for Cloud deployments; self-hosted rollout is intentionally gated behind allowlist env vars, so no existing deployments are broken.

The S3 validation path is straightforward and the integration tests exercise both the blocked and whitelisted cases end-to-end. The Azure connection-time protection relies on mutating request.agent inside a legacy RequestPolicyFactory shim — a plausible but internally-dependent mechanism that could silently stop working if @azure/storage-blob drops its @azure/core-http compatibility layer in a future upgrade.

StorageService.ts — specifically the createSecureAzureBlobRequestPolicyFactory function and how request.agent flows through the Azure SDK pipeline to the underlying HTTP transport.
</details>

```mermaid
sequenceDiagram
    participant User
    participant WebRouter as Web tRPC Router
    participant Service as service.ts
    participant Validator as blobStorageEndpointValidation
    participant DB as Database
    participant Worker as Worker Job
    participant StorageSDK as S3 / Azure SDK
    participant DNS as DNS Resolver

    Note over User,DNS: Save-time validation
    User->>WebRouter: update(endpoint, ...)
    WebRouter->>Service: upsertBlobStorageIntegration
    Service->>Validator: validateBlobStorageEndpoint(endpoint)
    Validator->>DNS: resolveHost(hostname)
    DNS-->>Validator: resolved IPs
    Validator-->>Service: blocked or allowed
    Service->>DB: upsert config (encrypted)

    Note over User,DNS: Worker export job
    Worker->>DB: load blobStorageIntegration
    Worker->>Validator: validateBlobStorageEndpoint(endpoint)
    Validator->>DNS: resolveHost
    DNS-->>Validator: resolved IPs
    Worker->>StorageSDK: StorageServiceFactory.getInstance(connectionValidation)
    loop Per table
        Worker->>StorageSDK: uploadFileBuffered
        StorageSDK->>DNS: lookup via secure agent
        DNS-->>StorageSDK: resolved IP
        StorageSDK-->>Worker: blocked or upload success
    end
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/__tests__/server/blob-storage-integration-trpc.servertest.ts:382-386
Imports placed inside the test function body violate the project rule that requires all imports to be at the top of the module. The two `await import(...)` calls fetch the same modules that are already referenced higher in the file — the direct relative path is used here only to manipulate module-level state for test isolation, but the dynamic import pattern itself is what the rule targets. Consider using `vi.mock` / `vi.spyOn`, or restructuring the test so the manipulated module reference is the same one already imported at the top of the file.

```suggestion
    it("allows endpoints when their IP is whitelisted", async () => {
      // Import should be moved to module top-level; use vi.mock/vi.spyOn for isolation
      const { env: validationEnv } =
        await import("../../../../packages/shared/src/env");
      const { validateBlobStorageEndpoint } =
        await import("../../../../packages/shared/src/server/services/blobStorageEndpointValidation");
```

### Issue 2 of 3
packages/shared/src/server/services/blobStorageEndpointValidation.ts:46-65
On Langfuse Cloud, the HTTPS-only check is positioned after `validateOutboundUrlHost`, which performs a full DNS resolution. For an `http://` endpoint pointing to a valid public host, the DNS lookup completes successfully before the HTTPS rejection is raised, spending network round-trips that aren't needed. Moving the Cloud HTTPS guard above the host-validation call avoids the wasted DNS lookup and produces the clearest error first.

```suggestion
  if (env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION && url.protocol !== "https:") {
    throw new Error(
      "Only HTTPS blob storage endpoints are allowed on Langfuse Cloud",
    );
  }

  try {
    await validateOutboundUrlHost({
      url,
      whitelist: effectiveWhitelist,
      logContext: BLOB_STORAGE_ENDPOINT_VALIDATION_LOG_CONTEXT,
      // Public IP literals are valid storage endpoints after blocklist checks and
      // should not require a reverse DNS path to exist at save time.
      shouldSkipDnsCheckForLiteralIps: true,
    });
  } catch (error) {
    const message =
      error instanceof Error ? error.message : "Validation failed";
    throw new Error(`${message}${getSelfHostedWhitelistGuidance()}`);
  }
```

### Issue 3 of 3
packages/shared/src/server/services/StorageService.ts:324-343
**Azure agent injection relies on undocumented SDK internals**

`createSecureAzureBlobRequestPolicyFactory` injects the validating HTTP agent by mutating `request.agent` inside a `RequestPolicyFactory`. This interface is from `@azure/core-http`, the legacy pipeline layer. In `@azure/storage-blob` v12.10+, `newPipeline` routes through the newer `@azure/core-rest-pipeline`; `RequestPolicyFactory` objects are adapted via a compatibility shim, and the shim's `WebResource → PipelineRequest` translation does not guarantee that `agent` is forwarded to the underlying transport. If a future SDK update drops or changes the shim, connection-time SSRF validation for Azure silently disappears without a failing test (the preflight `validateBlobStorageEndpoint` call would still pass). Consider using `@azure/core-rest-pipeline`'s first-class `PipelinePolicy` API or a custom `HttpClient` that wraps `node:https.Agent`, which is the stable extension point for injecting a custom agent.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(blob-storage): harden endpoint conne..."](https://github.com/langfuse/langfuse/commit/0dd2d16cba822cf096140ed9a2e8069112c6669d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32646030)</sub>

**Context used:**

- Rule used - Move imports to the top of the module instead of p... ([source](https://app.greptile.com/review/custom-context?memory=c960fc07-9928-409f-a18b-a780cbdded12))

**Learned From**

### CODE REVIEW

File: packages/shared/src/server/services/StorageService.ts
Comment (wochinge) [inline packages/shared/src/server/services/StorageService.ts:1]: I only did Azure/S3 as the others can't be provided by the user (only via env by the person doing the deployment)

File: packages/shared/src/env.ts [hunk 1]
```diff
@@ -320,6 +314,24 @@ const EnvSchema = z.object({
     .transform((s) =>
       s ? s.split(",").map((s) => s.toLowerCase().trim()) : [],
     ),
+  LANGFUSE_BLOB_STORAGE_ENDPOINT_WHITELISTED_IPS: z
```
Steffen911, line 317: Do we use whitelisted elsewhere? I think allow list is the preferred term across most OSS repos.

And can you share what values you'd set as "sane defaults" with the next major release? I wonder whether it's possible to have an exhaustive allow list, especially for Langfuse Cloud.

wochinge: I agree on the allow list term - I just went for consistency with the rest of the repository (`LANGFUSE_LLM_CONNECTION_WHITELISTED_*`, `LANGFUSE_WEBHOOK_WHITELISTED*`)

> And can you share what values you'd set as "sane defaults" with the next major release? I wonder whether it's possible to have an exhaustive allow list, especially for Langfuse Cloud.

This was an error on my side. On Cloud we usually (LLM, WebHook) allow all egress as long as it's not in the [BLOCKED_CIDRS](<https://github.com/langfuse/langfuse/blob/main/packages/shared/src/server/webhooks/ipBlocking.ts#L4>)

Steffen911: @wochinge So we would leave the defaults for cloud and not enforce anything additional?

wochinge: yes (same as for the others)

wochinge: @Steffen911 ready to go then?

Steffen911: Yep, seems good to me!

PR #13718, opened 2026-05-19
feat(ingestion): add app root detection

hassiebp:
## Summary

This PR adds an `is_app_root` marker to event ingestion and uses it in the events table root-observation filter.

## Filter behavior

The events sidebar exposes **Is Root Observation** as the visible filter option. Selecting `True` returns observations where either:

- `parent_span_id = ''`, meaning the event has no stored parent pointer, or
- `is_app_root = true`, meaning the SDK explicitly marked this span as an application root even though it may still have an OTEL parent.

The events table stores `parent_span_id` as a non-null `String`, so the app-root-aware predicate is `parent_span_id = '' OR is_app_root = true`.

## Frontend compatibility

Existing events UI URLs and saved views that contain the legacy `hasParentObservation` root-observation filter are migrated to `isRootObservation` after display-name normalization and before sidebar/saved-view validation:

- `hasParentObservation=false` becomes `isRootObservation=true`
- `hasParentObservation=true` becomes `isRootObservation=false`
- legacy boolean inequality forms are rewritten to the equivalent `isRootObservation = ...` form so the sidebar reflects the actual active filter

This keeps existing links and saved views working while making the sidebar accurately show the active **Is Root Observation** state. The shared/backend/MCP `hasParentObservation` column remains available as the literal parent-pointer filter; this migration is scoped to the events UI and saved-view path.

## Changes

- Add `is_app_root Bool DEFAULT false` to the events table dev schema and event record schema.
- Extract `langfuse.internal.is_app_root` from OTEL span attributes, accepting boolean `true` and string `"true"`.
- Force missing/null ingested `isAppRoot` values to `false` before writing events.
- Add the app-root-aware `isRootObservation` events-table filter column.
- Keep `hasParentObservation` as the literal parent-pointer filter for backward compatibility.
- Migrate legacy events UI URL and saved-view filters from `hasParentObservation` to `isRootObservation` so the sidebar reflects the active filter.

## Verification

- `pnpm --filter web exec vitest run --project server src/__tests__/server/api/otel/otelMapping.servertest.ts`
- `pnpm --filter web exec vitest run --project client src/features/filters/filter-integration.clienttest.ts`
- `pnpm --filter @langfuse/shared run typecheck`
- `pnpm --filter web run lint`
- `git diff --check`
- Browser check: a legacy `hasParentObservation=false` URL redirected to `isRootObservation=true` and activated the **Is Root Observation** sidebar facet.


Steffen911 (approved):

web/src/features/events/config/filter-config.ts
```diff
@@ -46,11 +46,10 @@ export const observationEventsFilterConfig: FilterConfig = {
...
     {
       type: "boolean" as const,
-      column: "hasParentObservation",
+      column: "isRootObservation",
       label: "Is Root Observation",
       tooltip:
-        "A root observation is the top-level observation in a trace. It has no parent observation ID. Filter to 'True' to see only root-level observations.",
-      invertValue: true, // "True" = hasParentObservation=false (is root)
+        "A root observation is top-level in a trace or marked as an app root by the SDK. Filter to 'True' to see root-level observations.",
     },
```
nimarb, line 107: Would you remove the `hasParentObservation` filter entirely? We could also rename the label to `Has OTel Parent` or similar and move the filter way down.

hassiebp: I keep `hasParentObservation` as a separate filter `column` but I would not expose it in the UI. I think the user intent is to show roots, either OTEL roots or app roots (they are not overlapping for most cases). Thus adding another facet for `Is OTEL root` is unnecessarily confusing. Users should be fine with the `Is root observation` that gives them the OTEL root, or if they used Langfuse SDKs with span filters they will see the app roots.

nimarb (approved):

PR #13720, opened 2026-05-19
feat(agent): persist and restore agent sessions

nimarb:

This PR introduces server-side persistence for in-app agent sessions: five new database tables (conversations, messages, runs, confirmation requests, credential leases), a tRPC router for conversation management, and a persistence layer that saves user/assistant messages and tracks run lifecycle.

- **DB schema & migration**: Five tables with proper cascade deletes, indexes for expected query patterns, and correct Prisma model definitions. The `providerSessionId` (Claude session identifier) is kept server-side only and never returned to clients.
- **Streaming handler**: Auth, feature-flag, and project-membership checks are all in place. Before streaming, the user message is persisted and a run record is created; run status is updated via fire-and-forget callbacks after the stream ends.
- **Frontend provider**: Manages the agent lifecycle with retry-on-invalid-session logic; after each run it calls `syncMessages` to persist the full conversation on the server.
</details>

Safe to merge if the silent-error swallowing in `updateRunStatus` is addressed; the bug causes run statuses to silently get stuck in RUNNING when a DB update fails.

The `updateRunStatus` function discards all database errors without logging, so any transient failure during run completion leaves the run permanently in RUNNING state with no operator visibility. The conversations list also has a NULL-ordering issue that pushes empty conversations to the top, and assistant messages are never associated with their run ID due to `syncMessages` not forwarding `runId`.

`persistence.ts` needs the most attention — it contains the silent error swallowing and the non-atomic create pattern. `InAppAiAgentProvider.tsx` and `router.ts` have the `runId` / NULL-ordering issues.
</details>

`````markdown
Fix the following 5 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 5
web/src/features/in-app-agent/server/persistence.ts:191-202
**`updateRunStatus` silently discards all errors with no logging**

The `.catch(() => undefined)` inside the function body means every database failure — network partition, deadlock, row lock timeout — is silently swallowed. Callers use `void updateRunStatus(...)`, so they cannot detect the failure either. A run that fails to update will remain stuck in `RUNNING` status permanently, with no record of the outcome and no signal to operators.

```suggestion
  await params.prisma.inAppAgentRun
    .update({
      where: { id: params.runId, projectId: params.projectId },
      data: {
        status: params.status,
        finishedAt:
          params.status === InAppAgentRunStatus.RUNNING ? null : new Date(),
        errorCode: params.errorCode ?? null,
        errorMessage: params.errorMessage ?? null,
      },
    })
    .catch((error) =>
      console.error("Failed to update agent run status", error),
    );
```

### Issue 2 of 5
web/src/features/in-app-agent/server/persistence.ts:96-130
**Non-atomic check-then-create in `ensureOwnedConversation`**

The `findFirst` and subsequent `create` are not wrapped in a transaction or `upsert`. If two concurrent requests arrive with the same `conversationId`, both will pass the `findFirst` returning null and both will attempt `create` with the same primary key, causing one to fail with a unique-constraint violation surfaced as a 500.

### Issue 3 of 5
web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx:223-244
**`syncMessages` never forwards `runId` to the server**

The router's `syncMessages` endpoint accepts an optional `runId`, but `syncMessagesMutation.mutate(...)` here never includes it. All assistant-response messages created by client sync will have `run_id = NULL` in the database, making per-run message queries unreliable.

### Issue 4 of 5
web/src/features/in-app-agent/server/router.ts:34-44
**NULL `lastMessageAt` sorts ahead of real timestamps in `list` results**

PostgreSQL treats `NULL` as larger than any non-null value in `ORDER BY ... DESC`, so conversations with no messages float to the top above conversations with actual activity. Use `{ lastMessageAt: { sort: "desc", nulls: "last" } }` in Prisma.

### Issue 5 of 5
web/src/features/in-app-agent/server/persistence.ts:20-27
**Dead entries in `AG_UI_ROLE_TO_DB_ROLE`**

`SYSTEM`, `TOOL`, `ACTIVITY`, and `REASONING` are mapped here but `toPersistableMessage` returns `null` for all non-user/assistant roles before the map is ever consulted. The four extra entries are unreachable dead code.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(agent): persist and restore agent s..."](https://github.com/langfuse/langfuse/commit/a099bd902221db57f723c2e0150b57ec8c21383e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32826089)</sub>

### CODE REVIEW

File: packages/shared/prisma/migrations/20260519000000_add_in_app_agent_persistence/migration.sql [hunk 1]
```diff
@@ -0,0 +1,155 @@
...
+  "id" TEXT NOT NULL,
+  "project_id" TEXT NOT NULL,
+  "conversation_id" TEXT NOT NULL,
+  "run_id" TEXT,
+  "created_by_user_id" TEXT,
+  "api_key_id" TEXT NOT NULL,
+  "expires_at" TIMESTAMP(3) NOT NULL,
+  "revoked_at" TIMESTAMP(3),
+  "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
```
nimarb: remove, api key expiry is on key

packages/shared/prisma/migrations/20260601190000_add_in_app_agent_persistence/migration.sql
```diff
@@ -0,0 +1,155 @@
...
+  "last_message_at" TIMESTAMP(3),
+  "archived_at" TIMESTAMP(3),
+  "deleted_at" TIMESTAMP(3),
+  "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+  CONSTRAINT "in_app_agent_conversations_pkey" PRIMARY KEY ("id")
+);
+
+CREATE TABLE "in_app_agent_runs" (
```
nimarb, line 17: keep track of used api key id in this table

packages/shared/prisma/migrations/20260519000000_add_in_app_agent_persistence/migration.sql
```diff
@@ -0,0 +1,155 @@
...
+  "status" "InAppAgentMessageStatus" NOT NULL DEFAULT 'COMPLETED',
+  "author_user_id" TEXT,
+  "metadata" JSONB,
+  "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+  CONSTRAINT "in_app_agent_messages_pkey" PRIMARY KEY ("id")
+);
+
+CREATE TABLE "in_app_agent_confirmation_requests" (
```
nimarb: leave out for now

nimarb:
@claude review

martingarramon (commented):
Two questions on the persistence layer:

**`ensureOwnedConversation` concurrent first-create** (`server/persistence.ts` ~L104–130): `ensureOwnedConversation` does a `findFirst` and then conditionally calls `create` outside a transaction. If two first requests for the same `conversationId` run concurrently, they could both observe no existing row and both attempt the create; depending on the table constraints, one may then fail on the duplicate insert. `appendConversationMessage` uses a `$transaction` and a `FOR UPDATE` lock once the conversation row exists. Should creation use an `upsert` or another concurrency-safe pattern so the first-create path is consistent?

**`AG_UI_ROLE_TO_DB_ROLE` unreachable entries** (`server/persistence.ts` ~L17–24): the map covers all 6 AG-UI roles, but `toPersistableMessage` returns null for anything other than `user`/`assistant` — `tool`, `system`, `activity`, and `reasoning` entries never reach the lookup. Is that a placeholder for persisting those roles later, or can the map be trimmed to match the filter? A reader could assume all 6 roles are persisted when they aren't.

nimarb:
@claude review

web/src/features/in-app-agent/schema.ts
```diff
@@ -156,6 +156,11 @@ export const InAppAgentRuntimeStateSchema = z.discriminatedUnion("type", [
     type: z.literal("newSession"),
     projectId: z.string(),
   }),
+  z.object({
+    type: z.literal("existingConversation"),
+    projectId: z.string(),
+    conversationId: z.string(),
+  }),
   z.object({
     type: z.literal("existingSession"),
```
bezbac: Should we remove the `type: z.literal("existingSession"),` here? We probably should store the claude session id in the DB now.

web/src/features/in-app-agent/schema.ts
```diff
@@ -156,6 +156,11 @@ export const InAppAgentRuntimeStateSchema = z.discriminatedUnion("type", [
     type: z.literal("newSession"),
```
bezbac: Might make sense to change this name to `newConversation` too.

web/src/features/in-app-agent/server/handler.ts
```diff
@@ -90,18 +97,46 @@ export default async function handler(request: Request) {
...
+        };
+      }
+
+      const token = verifyInAppAgentSessionToken(
+        parsedState.data.claudeSessionToken,
+        {
+          userId: auth.userId,
+          threadId: input.threadId,
+        },
+      );
```
bezbac: Since the session token in now stored in the DB, I think we can remove the `verifyInAppAgentSessionToken` and `signInAppAgentSessionToken` functions completely.

web/src/features/in-app-agent/server/persistence.ts
```diff
@@ -0,0 +1,408 @@
...
+  const existing = await params.prisma.inAppAgentConversation.findFirst({
+    where: {
+      id: params.conversationId,
+      projectId: params.projectId,
+      deletedAt: null,
+    },
+  });
+
+  if (existing) {
+    if (existing.createdByUserId !== params.userId) {
```
bezbac: Could this condition be part of the where clause? I think then the where clauses would be identical between `ensureOwnedConversation` and `getOwnedConversationOrThrow`, which might allow further deduplication.

packages/shared/prisma/migrations/20260601190000_add_in_app_agent_persistence/migration.sql
```diff
@@ -0,0 +1,75 @@
...
+  "mcp_api_key_id" TEXT,
+  "error_code" TEXT,
+  "error_message" TEXT,
+  "started_at" TIMESTAMP(3),
+  "finished_at" TIMESTAMP(3),
+  "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+  CONSTRAINT "in_app_agent_runs_pkey" PRIMARY KEY ("id")
+);
```
bezbac, line 31: Should add a field for the langfuse trace id or do we just use the ID of the run itself?

web/src/features/in-app-agent/server/persistence.ts
```diff
@@ -0,0 +1,408 @@
...
+    externalId: message.id,
+    sequenceNumber,
+    role: AG_UI_ROLE_TO_DB_ROLE[message.role],
+    content: toJsonValue(message),
+  };
+}
+
+function toJsonValue(value: unknown): Prisma.InputJsonValue {
+  return JSON.parse(JSON.stringify(value)) as Prisma.InputJsonValue;
+}
```
bezbac: What is the point of this, for most values this would be a no-op, right?

web/src/features/in-app-agent/server/router.ts
```diff
@@ -0,0 +1,200 @@
...
+  sequenceNumber: z.number().int().min(0),
+});
+
+const ConversationIdInput = z.object({
+  projectId: z.string(),
+  conversationId: z.string(),
+});
+
+export const inAppAgentRouter = createTRPCRouter({
+  list: protectedProjectProcedure
```
bezbac: Should we call this `listConversations` instead of just `list`? Same with `create` and `get`

web/src/features/in-app-agent/server/router.ts
```diff
@@ -0,0 +1,200 @@
...
+
+      const conversation = await ctx.prisma.inAppAgentConversation.create({
+        data: {
+          projectId: input.projectId,
+          createdByUserId: ctx.session.user.id,
+        },
+      });
+
+      return serializeConversation(conversation);
+    }),
```
bezbac: Do we even need this if `InAppAgentRuntimeStateSchema` has a `newSession` type?

web/src/features/in-app-agent/server/agent.ts
```diff
@@ -152,6 +185,7 @@ export function createAgUiStream(params: {
 function normalizeAdapterEvent(
   event: AgUiEvent,
   createResumeStateForSessionId: (sessionId: string) => unknown,
+  onResumeSessionId?: (sessionId: string) => void,
```
bezbac: I think `createResumeStateForSessionId` and `onResumeSessionId` could be combined

web/src/features/in-app-agent/server/handler.ts
```diff
@@ -138,26 +173,103 @@ export default async function handler(request: Request) {
...
+        projectId,
+        conversationId: conversation.id,
+        userId: auth.userId,
+        message: sanitizedInput.messages[0]!,
+        runId: sanitizedInput.runId,
+      });
+
+      const resumeSessionId =
+        tokenClaudeSessionId ?? conversation.providerSessionId ?? undefined;
+
```
bezbac: I think we can remove the `tokenClaudeSessionId` here completely

web/src/features/in-app-agent/components/InAppAgentDrawer.tsx
```diff
@@ -206,3 +274,7 @@ export function InAppAgentDrawer(props: InAppAgentDrawerProps) {
     </section>
   );
 }
+
+function getConversationTitle(conversation: InAppAgentDrawerConversation) {
+  return conversation.title?.trim() || "Untitled conversation";
+}
```
bezbac: Nit: `getConversationTitle` is only used once and very simple. I would just inline it.

web/src/features/in-app-agent/components/InAppAgentDrawer.tsx
```diff
@@ -30,13 +45,26 @@ type InAppAgentDrawerCloseButtonProps =
 
 export type InAppAgentDrawerProps = {
   error: string | null;
-  isRunning: boolean;
+  isInputDisabled: boolean;
   messages: InAppAgentDrawerMessage[];
+  conversations: InAppAgentDrawerConversation[];
```
bezbac, line 51: I guess `conversations` should be paginated

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -12,51 +12,73 @@ import {
...
-  messages: [],
+  conversationId,
 });
 
 const NOOP_CONTEXT: InAppAiAgentContextType = {
   isAvailable: false,
   open: false,
   setOpen: () => undefined,
   isRunning: false,
+  isInputDisabled: false,
```
bezbac: Nit: I think it's a little ugly to have both `isRunning` and `isInputDisabled` which uses `isRunning`. I would probably opt to return the individual parts of `isInputDisabled` instead

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -354,6 +611,21 @@ function isInvalidSessionTokenError(error: unknown): boolean {
   );
 }
 
+function isNotFoundTrpcError(error: unknown): boolean {
```
bezbac: `isNotFoundTrpcError` is only called once and with an object that is much more descriptive than `unknown`. I think `error` could be typed more strictly here which might allow us to remove some unneccessary checks.

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -114,123 +135,283 @@ function InAppAiAgentProjectProvider({
...
+    { projectId },
+    { enabled: open },
+  );
+  const conversationQuery = api.inAppAgent.get.useQuery(
+    {
+      projectId,
+      conversationId: selectedConversationId ?? "",
+    },
+    { enabled: open && Boolean(selectedConversationId) },
+  );
```
bezbac: Nit: The naming here is deceptively similar with `conversationsQuery` only differing in one single character to `conversationQuery`. I think renaming `conversationsQuery` to `conversationListQuery` might be nice.

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -114,123 +135,283 @@ function InAppAiAgentProjectProvider({
...
+    resetAgent,
+    selectedConversationId,
+    setSelectedConversationId,
+  ]);
+
+  useEffect(() => {
+    return () => {
+      resetAgent();
+    };
+  }, [resetAgent]);
```
bezbac, line 301: Personally, I think comments for the use effect would help. It took me a while of reading the code to understand that the 2nd useEffect basically just resets the local state if the remove state was deleted and the first useEffect just loads the data from the local storage.

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -114,123 +135,283 @@ function InAppAiAgentProjectProvider({
...
+        } catch (error) {
+          const errorMessage = getAgentErrorMessage(error);
+          showErrorToast("Failed to save conversation", errorMessage);
+          console.error("Failed to sync in-app agent messages", error);
+          void utils.inAppAgent.get.invalidate({
+            projectId,
+            conversationId,
+          });
+        }
+      })();
```
bezbac: This was codex, right? I Find the `void` statements incredibly ugly but we don't need to fix this here, I think we can just add an eslint rule for this.

web/src/features/in-app-agent/server/persistence.ts
```diff
@@ -0,0 +1,560 @@
...
+      },
+    });
+
+    await tx.inAppAgentConversation.update({
+      where: {
+        id: params.conversationId,
+        projectId: params.projectId,
+      },
+      data: { lastMessageAt: new Date() },
+    });
```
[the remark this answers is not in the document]
nimarb: ignore for now

web/src/features/in-app-agent/server/router.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+function getDefaultConversationTitle(date: Date) {
+  const weekday = WEEKDAY_NAMES[date.getDay()] ?? "Unknown";
+  const hours = String(date.getHours()).padStart(2, "0");
+  const minutes = String(date.getMinutes()).padStart(2, "0");
+
+  return `Chat on ${weekday} at ${hours}:${minutes}`;
+}
```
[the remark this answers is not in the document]
nimarb: not important

nimarb:
@claude review

packages/shared/prisma/schema.prisma
```diff
@@ -206,6 +211,77 @@ model ApiKey {
...
+}
+
+model InAppAgentConversation {
+  id                String                                @id @default(cuid())
+  projectId         String                                @map("project_id")
+  project           Project                               @relation(fields: [projectId], references: [id], onDelete: Cascade)
+  createdByUserId   String?                               @map("created_by_user_id")
+  createdByUser     User?                                 @relation("InAppAgentConversationCreatedByUser", fields: [createdByUserId], references: [id], onDelete: SetNull)
+  title             String?
+  visibilityScope   InAppAgentConversationVisibilityScope @default(PERSONAL) @map("visibility_scope")
```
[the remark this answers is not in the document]
nimarb: leave in for now

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -114,142 +136,214 @@ function InAppAiAgentProjectProvider({
...
+    },
+  );
+  const conversationQuery = api.inAppAgent.get.useQuery(
+    {
+      projectId,
+      conversationId: selectedConversationId ?? "",
+    },
+    { enabled: open && Boolean(selectedConversationId) },
+  );
+  const createConversationMutation = api.inAppAgent.create.useMutation();
```
bezbac: I think we should either have `api.inAppAgentConversations.list` etc or `api.inAppAgent.listConversations`. `api.inAppAgent.list` might be confusing long term if we add additional actions into the `inAppAgent` router.

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -259,64 +353,122 @@ function InAppAiAgentProviderInner({
...
+    },
+    [resetAgent, setSelectedConversationId],
+  );
+
+  const startNewConversation = useCallback(() => {
+    setError(null);
+    resetAgent();
+    setMessages([]);
+    setSelectedConversationId(null);
+  }, [resetAgent, setSelectedConversationId]);
```
bezbac: `selectConversation` and `startNewConversation` look largely the same with the only difference being the selected conversation id.

web/src/features/in-app-agent/components/InAppAiAgentProvider.tsx
```diff
@@ -259,64 +353,122 @@ function InAppAiAgentProviderInner({
...
+
+        agent.addMessage(userMessage);
+        setMessages(agent.messages.filter(isAgentConversationMessage));
+        startedRun = true;
+        runAgent(agent, conversationId);
+        return true;
+      } catch (error) {
+        const errorMessage = getAgentErrorMessage(error);
+        setError(errorMessage);
+        showErrorToast("Assistant failed", errorMessage);
```
bezbac: I noticed this in multiple places, do we always need both, the `setError` and the `showErrorToast`? I think `setError` should be enough as it's rendered in the tread and the user would be looking at it anyways.

web/src/features/in-app-agent/server/router.ts
```diff
@@ -0,0 +1,184 @@
...
+const CONVERSATION_LIST_LIMIT = 50;
+const WEEKDAY_NAMES = [
+  "Sunday",
+  "Monday",
+  "Tuesday",
+  "Wednesday",
+  "Thursday",
+  "Friday",
+  "Saturday",
+];
```
bezbac: Why hard-code these over using `date.toLocaleDateString(locale, { weekday: "long" })`?

web/src/features/in-app-agent/server/persistence.ts
```diff
@@ -0,0 +1,668 @@
...
+  for (const toolCall of next) {
+    byId.set(toolCall.id, toolCall);
+  }
+
+  return Array.from(byId.values());
+}
+
+function isAgUiEvent(event: unknown) {
+  return isRecord(event) && typeof event.type === "string";
+}
```
bezbac: This is not a very robust check 😅

bezbac (commented):
Looks very nice, I have some more minor comments and a few of the previous comments were not addressed yet.

web/src/features/in-app-agent/server/router.ts
nimarb: we dont have a locale on the server, i would just keep this for now provisionally. for actual names i want to name then via AI liek chatgpt does it

bezbac: We're using english everywhere, so "en" should be enough in terms of the locale right? But yeah this will likely get deleted anyways.

packages/shared/prisma/migrations/20260601190000_add_in_app_agent_persistence/migration.sql
nimarb: we use the run.id

web/src/features/in-app-agent/server/handler.ts
```diff
@@ -134,59 +140,134 @@ export default async function handler(request: Request) {
...
+            conversationId: conversation.id,
+            userId: auth.userId,
+            model: "haiku",
+            mcpApiKeyId: mcpApiKey.id,
+          });
+          runCreated = true;
+
+          const finishCurrentRun = (error?: {
+            errorCode: string;
+            errorMessage: string;
```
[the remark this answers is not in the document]
nimarb: not taht bad

web/src/features/in-app-agent/server/handler.ts
```diff
@@ -134,59 +144,161 @@ export default async function handler(request: Request) {
...
+          await createRun({
+            prisma,
+            runId: sanitizedInput.runId,
+            projectId,
+            conversationId: conversation.id,
+            triggeredByUserId: auth.userId,
+            // TODO: Store the exact provider model id, not the Claude SDK alias.
+            model: "haiku",
+            mcpApiKeyId: mcpApiKey.id,
+          });
```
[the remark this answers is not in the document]
nimarb: ok for now

nimarb:
@chatgpt-codex-connector review

PR #13728, opened 2026-05-19
feat(monitors): render handlebars message templates with LRU-cached compiles

marksalpeter:
## Summary

The Monitor execution loop renders two Handlebars templates per surviving monitor inside the `MonitorQueueProcessor` (name → alert title, message → alert body), and `Handlebars.compile` is the expensive step — running it per render would burn CPU on every batch.

To achieve this we:
- Added a `MonitorTemplateRenderer` class that compiles templates lazily and caches the compiled functions in an LRU keyed by template source.
- Realigned `MonitorMessageContext` to mirror [Datadog's monitor variables](https://docs.datadoghq.com/monitors/notify/variables/) by name (`comparator`, `warn_threshold`, full `is_*`/`was_*` severity pairs, plus convenience flags `is_recovery`, `is_escalation`, `is_renotify`).
- Derived every flag from a `(from, to)` Monitor pair so transitions and renotifies fall out of the data without extra params.
- Opened up `{{#and}}` and `{{#or}}` block helpers on a sandboxed `Handlebars.create()` instance — the validator allows them recursively, everything else (`if`/`unless`/`each`/`with`/`lookup`) stays rejected.
- Switched the value/threshold formatter on the metric's measure unit string (`millisecond`/`USD`/`tokens`/count types) so worker callers don't pre-format numbers.

**Note**: integration into the `MonitorQueueProcessor` is out of scope — this ticket only ships the renderer + tests. Wiring tracked separately in the monitor execution tickets.

Fixes [LFE-9811](https://linear.app/langfuse/issue/lfe-9811-add-handlebars-template-render)

## Verification

- [x] `pnpm --filter @langfuse/shared run typecheck`
- [x] `pnpm --filter @langfuse/shared exec vitest run src/features/monitor` (154 tests passing, 7 snapshots)

## Test plan

- [ ] Confirm the snapshot fixtures in `__snapshots__/template.test.ts.snap` cover every severity flag (`is_*`/`was_*`), `is_recovery`, `is_escalation`, and `is_renotify`.
- [ ] Compile a representative monitor template using `{{#and is_alert was_warning}}…{{/and}}` and verify the validator accepts it; verify `{{#if is_alert}}…{{/if}}` and `{{#each tags}}…{{/each}}` are still rejected.
- [ ] Spot-check the value formatter against each measure unit (`millisecond`, `USD`, `tokens`, count types) by rendering `{{value}} / {{threshold}}` with a 5,432 / 1,000 fixture.

### DIFF
--- a/packages/shared/src/features/monitor/template.ts
+++ b/packages/shared/src/features/monitor/template.ts
@@ -1,57 +1,105 @@
-/** template.ts contains Monitor message template validation */
+/** template.ts contains Monitor message template validation and rendering. */
+import type { MonitorSeverity, MonitorThresholdOperator } from "@prisma/client";
 import Handlebars from "handlebars";
+import { LRUCache } from "lru-cache";
+import { DAY, HOUR, MINUTE, WEEK } from "./internal";
+import type { Monitor } from "./types";
 
 /**
  * MonitorMessageContext is the allowlisted context passed to a Monitor
  * message template at render time. Templates may only reference these
- * top-level keys.
+ * top-level keys. Names mirror Datadog's monitor variables
+ * (https://docs.datadoghq.com/monitors/notify/variables/).
  */
 export type MonitorMessageContext = {
   value: string;
   threshold: string;
-  warningThreshold: string | null;
-  operator: ">" | ">=" | "<" | "<=" | "==" | "!=";
+  warn_threshold: string | null;
+  comparator: ">" | ">=" | "<" | "<=" | "==" | "!=";
   window: string;
   permalink: string;
-  tags: string[];
 
   is_ok: boolean;
-  is_warning: boolean;
   is_alert: boolean;
+  is_warning: boolean;
   is_no_data: boolean;
+  is_unknown: boolean;
 
-  is_fired: boolean;
-  is_resolved: boolean;
-  is_crossed: boolean;
+  was_ok: boolean;
+  was_alert: boolean;
+  was_warning: boolean;
+  was_no_data: boolean;
+  was_unknown: boolean;
+
+  is_recovery: boolean;
+  is_escalation: boolean;
+  is_renotify: boolean;
+
+  last_triggered_at: string;
+  last_triggered_at_epoch: string;
 };
 
 const monitorMessageContextKeys = new Set<string>([
   "value",
   "threshold",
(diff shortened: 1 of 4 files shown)


packages/shared/src/features/monitor/template.ts
```diff
@@ -87,3 +149,182 @@ const isAllowedPath = (
...
+export const DEFAULT_MONITOR_TEMPLATE_CACHE_SIZE = 1000;
+
+/**
+ * MonitorTemplateRenderer compiles Monitor message templates and caches
+ * the compiled functions in an LRU keyed by template source.
+ *
+ * `baseUrl` is the deployment URL used to build `{{permalink}}`. On the
+ * worker, pass `env.NEXTAUTH_URL`.
+ */
+export class MonitorTemplateRenderer {
```
maxdeichmann, line 182: @marksalpeter do we really need this for now? I would expect this to be fairly cheap as we dont throw thousands of these messages out for now.

packages/shared/src/features/monitor/template.ts
```diff
@@ -87,3 +149,182 @@ const isAllowedPath = (
...
+    }
+    return compiled(toMonitorMessageContext(params, this.baseUrl));
+  }
+
+  get size(): number {
+    return this.cache.size;
+  }
+}
+
+const toMonitorMessageContext = (
```
maxdeichmann, line 214: @marksalpeter this is pretty cool but cant this wait for v2?

marksalpeter: So either we should pull this and work on a feature that derives a plain english description of the alert from the query config. Or we should ship with the requirement that people write an alert message with enough context in it to be accurate. Personally I think the former is more interesting, but possibly harder than rendering a user defined message. And even if we ship the first version with that, people will requests the ability to customize the message themselves.

marksalpeter: as mentioned https://github.com/langfuse/langfuse/pull/13725#discussion_r3275950040
I will pull from scope for now.

maxdeichmann: I would have allowed people to insert any comment as a string for now.

marksalpeter:
pulling from the mvp

PR #13731, opened 2026-05-19
feat(trace): send web callouts from a trace & session

nimarb:


martingarramon (commented):
Security model reads clearly to me: host/IP validation remains the SSRF boundary, so `allowedPorts: "any"` seems like a reasonable tradeoff for browser callback URLs on non-standard dev ports. The composite index on `(project_id, enabled)` also covers the `.enabled` query well.

One small issue: `packages/shared/src/server/webhooks/validation.ts:49` still hardcodes `"Only ports 80 and 443 are allowed"`. That's accurate for the default case, but misleading when `allowedPorts` is a custom list — a caller passing `["8080"]` would see the wrong message.

LGTM otherwise.

nimarb:
@chatgpt-codex-connector review

nimarb:
@claude review

PR #13732, opened 2026-05-19
feat(monitors): add MonitorScheduler with slot-based batching

marksalpeter:
## Summary

Monitor evaluation needs a scheduler that ticks every second, claims due
monitor rows for its slot, and emits one `MonitorQueueEvent` per
`schedulerBatchId` group onto a sharded `MonitorQueue` for the downstream
processor.

To achieve this we:
- Added the sharded `MonitorQueue` Redis class + `QueueName.MonitorQueue` / `QueueJobs.MonitorJob` registration so the scheduler has a publish target (deferred from [LFE-9808](https://linear.app/langfuse/issue/LFE-9808/add-monitor-shared-zod-and-queue-schemas)).
- Added `MonitorScheduler` in `@langfuse/shared` with dependency-inverted Prisma + publish hooks; its `tick(scheduledAt)` runs a `SELECT FOR UPDATE SKIP LOCKED` claim, a single combined `UPDATE` that advances `next_run_at` for every claimed row and gate-publishes ACTIVE rows on first-run / prior-run-done / 5-minute-TTL, and a `GROUP BY scheduler_batch_id` final select.
- Added `MonitorSchedulerRunner` in `worker` that extends `PeriodicExclusiveRunner` for per-slot leader election (lock key `langfuse:monitor-scheduler:<slot>`, `onUnavailable: "fail"`), uses a 1s cadence, and routes each event to the correct `MonitorQueue` shard via `${projectId}-${schedulerBatchId}`.
- Wired `LANGFUSE_MONITOR_SCHEDULERS` (default `0` — opt-in) and instantiated one runner per slot in `worker/src/app.ts`.

**Note**: The RFC's three-stage CTE collapses into one combined `UPDATE` because Postgres forbids updating the same row twice in a single statement. Semantics are unchanged. The RFC's `jobId = "${schedulerBatchId}:${scheduledAt}"` becomes `${schedulerBatchId}-${scheduledAt.getTime()}` because BullMQ custom job IDs cannot contain `:`; dedupe semantics are identical.

**Note**: The downstream `MonitorQueueProcessor` (consumer side) is tracked separately.

Fixes [LFE-9812](https://linear.app/langfuse/issue/LFE-9812/add-monitorscheduler-with-slot-based-batching)
Related to [LFE-9808](https://linear.app/langfuse/issue/LFE-9808/add-monitor-shared-zod-and-queue-schemas)

## Verification

- [x] `pnpm --filter @langfuse/shared run test` (101 tests pass)
- [x] `pnpm --filter worker run lint`
- [x] `pnpm --filter worker run test monitorScheduler.test` (8 tests pass)
- [x] `pnpm --filter web run lint`
- [x] `pnpm exec turbo run typecheck --filter=web --filter=worker --filter=@langfuse/shared --force`

## Test plan

- [ ] Set `LANGFUSE_MONITOR_SCHEDULERS=2` on a worker, seed a few monitor rows across slots, confirm both slots tick independently and publish jobs land on the expected shards
- [ ] Run two worker pods with the same `LANGFUSE_MONITOR_SCHEDULERS=1`, verify only one publishes per tick (other reports lock miss)
- [ ] Pause a monitor and confirm `next_run_at` still advances but `last_published_run_at` never moves
- [ ] Stall a monitor's `last_completed_run_at` for >5 min and confirm the scheduler re-publishes

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -222,6 +222,7 @@ jobs:
         run: pnpm --filter @repo/eslint-plugin run test
 
   tests-shared:
+    timeout-minutes: 15
     runs-on: blacksmith-4vcpu-ubuntu-2404
     needs:
       - pre-job
@@ -230,9 +231,9 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
-      - uses: pnpm/action-setup@8912a9102ac27614460f54aedde9e1e7f9aec20d # v6.0.5
+      - uses: pnpm/action-setup@739bfe42ca9233c5e6aca07c1a25a9d34aca49b0 # v6.0.7
         with:
-          version: 10.33.0
+          version: 11.1.3
       - name: Use Node.js ${{ env.NODE_VERSION }} without cache
         if: env.CI_CACHE_ALLOWED != 'true'
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -221,6 +221,42 @@ jobs:
       - name: run eslint-plugin tests
         run: pnpm --filter @repo/eslint-plugin run test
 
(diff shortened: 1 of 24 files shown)


marksalpeter:
closed in favor of #13874

PR #13747, opened 2026-05-20
feat(agent): Connect in-app agent to langfuse MCP

bezbac:

This PR connects the in-app agent to the Langfuse MCP server by creating an ephemeral project API key per agent session, wiring it into `ClaudeAgentAdapter` as an HTTP MCP server, and deleting the key when the stream finishes. It also extracts the shared `deleteApiKeyFromDb` helper from `ApiAuthService` into `packages/shared`.

- **Ephemeral MCP key lifecycle**: `withInAppAgentMcpApiKeyCleanup` creates an in-app-agent-scoped key before starting the stream, passes it and a cleanup callback into `createAgUiStream`, and the `onFinish` hook deletes the key when the stream closes or is cancelled. The existing `isInAppAgentKey` field scopes the key to MCP-only endpoints and individual tools that opt-in via `allowInAppAgentKey: true`.
- **Stream lifecycle improvements**: `closed`, `subscription`, and `abortHandler` are lifted to the outer scope so the new `cancel()` handler can clean up the abort listener and call `finish()` symmetrically with the abort/complete paths — directly addressing the previous gap where `cancel()` left the listener dangling.
- **Refactor**: `deleteApiKeyFromDb` is extracted as a standalone function in `packages/shared`, with `ApiAuthService.deleteApiKey` and the new handler both delegating to it.
</details>

The changes are safe to merge. The ephemeral key lifecycle is correctly guarded against synchronous errors, the abort-listener leak is fixed, and double-deletion is prevented by the `cleanupPromise` dedup guard.

No correctness defects found in the changed paths. The key creation, MCP wiring, and cleanup flow all handle the expected error conditions. The previously noted gaps (try/finally for sync throws, abort listener cleanup) are both addressed in this PR.

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Handler
    participant DB
    participant Agent as ClaudeAgentAdapter
    participant MCP

    Client->>Handler: POST agent request
    Handler->>DB: create ephemeral API key (isInAppAgentKey flag)
    DB-->>Handler: mcpApiKey record
    Handler->>Agent: createAgUiStream with MCP config
    Handler-->>Client: streaming Response

    loop Agent execution
        Agent->>MCP: tool call with Basic auth
        MCP->>DB: verify auth header
        DB-->>MCP: key valid
        MCP-->>Agent: tool result
        Agent-->>Client: SSE events
    end

    alt stream finishes or is cancelled
        Agent->>Handler: onFinish fires
        Handler->>DB: delete ephemeral key
    end

    alt synchronous error during stream setup
        Handler->>DB: delete ephemeral key via catch block
    end
```
</details>

<sub>Reviews (2): Last reviewed commit: ["feat(agent): Connect in-app agent to lan..."](https://github.com/langfuse/langfuse/commit/b162971e16bc2737929d0a488bd25f4610d58af7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33408058)</sub>


bezbac:
@greptileai

martingarramon (commented):
Two questions on the ephemeral key lifecycle:

**Cleanup failure can leave the key alive**

`finish()` in `createAgUiStream` calls `onFinish` via `void Promise.resolve(...).catch(console.error)`. If the key deletion hits a transient error, the exception is logged and dropped. The `cleanupPromise = undefined` reset in `cleanupMcpApiKey`'s catch block looks like it enables a retry, but `finished = true` in `finish()` means `finish()` never fires `onFinish` again — the reset is dead code in this call path. If cleanup fails in this path, the MCP key appears to remain live unless another cleanup path handles it. Is a retry or fallback intended here?

**Process crash between key creation and stream completion**

Keys are created with `isInAppAgentKey: true`. Is there a background job that sweeps orphaned in-app-agent keys (e.g. keys older than a session TTL with `isInAppAgentKey: true`)? Not seeing one in this PR — noting it in case it lives elsewhere.

web/src/features/in-app-agent/server/agent.ts
```diff
@@ -37,11 +43,24 @@ export function createAgUiStream(params: {
...
     title: ASSISTANT_TITLE,
     systemPrompt: ASSISTANT_SYSTEM_PROMPT,
-    allowedTools: [],
+    allowedTools: ["mcp__langfuse__*"],
+    mcpServers: {
+      langfuse: {
+        type: "http",
+        url: params.options.langfuseMcp.url,
+        headers: {
+          Authorization: langfuseMcpAuthHeader,
```
nimarb, line 60: on eg connection error, this will probably leak into logs

bezbac: I just tried it, MCP failures would not trigger the `error` branch of the `.subscribe`. One can see that the MCP failed by looking at the "system:init" event in the normal `next` branch (Line 155). We're not currently logging all of the events. Even if we were logging them, the "mcp_servers" part of the event includes an object that just says "failed", not logging the header: `{ "name": "langfuse", "status": "failed" }`

nimarb (approved):

PR #13748, opened 2026-05-20
feat(agent): Trace in-app agent in langfuse

bezbac:

This PR adds Langfuse tracing to the in-app agent, wiring a new `InAppAgentInstrumentation` class into the existing `createAgUiStream` pipeline so that each agent run creates a trace with a parent `agent-run` span and individual child tool observations.

- `instrumentation.ts` handles the full event lifecycle (text output, reasoning, tool call start/args/result/end, run finish/error) with an `ended` guard that prevents double-recording across the multiple terminal paths in `agent.ts`.
- `eventCompaction.ts` is extracted from `persistence.ts` so it can be shared by both the persistence layer and the new instrumentation.
- `handler.ts` gates trace creation on `aiTelemetryEnabled` and a configured `targetProjectId`.
</details>

Safe to merge. The tracing path is entirely opt-in (gated by aiTelemetryEnabled and a configured project ID), errors in initialization or flushing are caught and logged, and the ended guard prevents double-recording across the multiple terminal paths in the stream.

The core logic is correct — all stream terminal paths (error, abort, cancel, RUN_ERROR event, normal completion) reach end/endWithError exactly once and trigger flush. The instrumentation is non-blocking and failure-safe. The two concerns flagged (unsafe internal SDK cast and hardcoded environment string) are maintenance risks rather than current defects.

instrumentation.ts warrants a second look if the Langfuse SDK is upgraded, since tool observation creation relies on an undocumented internal enqueue method.
</details>

```mermaid
sequenceDiagram
    participant Handler as handler.ts
    participant Agent as createAgUiStream
    participant Instr as InAppAgentInstrumentation
    participant SDK as Langfuse SDK (internal)
    participant Ingestion as processEventBatch

    Handler->>Agent: "createAgUiStream({ langfuseTracing })"
    Agent->>Instr: "createInAppAgentInstrumentation({ input, tracing })"
    Instr->>SDK: "langfuse.trace({ id: traceId, name: in-app-agent })"
    Instr->>SDK: "trace.span({ name: agent-run, input: lastUserMsg })"

    loop AG-UI event stream
        Agent->>Instr: recordEvents(agUiEvents)
        alt TEXT_MESSAGE_CONTENT / CHUNK
            Instr->>Instr: "output += delta"
        else TOOL_CALL_START
            Instr->>Instr: toolSpans.set(toolCallId)
        else TOOL_CALL_ARGS
            Instr->>Instr: "tool.args += delta"
        else TOOL_CALL_END + TOOL_CALL_RESULT
            Instr->>SDK: langfuse.enqueue(tool-create, body)
        else RUN_FINISHED
            Instr->>SDK: "span.update({ output }) then span.end()"
        else RUN_ERROR
            Instr->>SDK: "span.update({ level: ERROR }) then span.end()"
        end
    end

    Agent->>Instr: flush()
    Instr->>SDK: langfuse._exportLocalEvents()
    SDK-->>Instr: queued events
    Instr->>Ingestion: processEventBatch(events)
    Ingestion-->>Instr: traces persisted
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/features/in-app-agent/server/instrumentation.ts:352-356
Accessing the internal `enqueue` method via a double type-cast (`as unknown as`) is fragile. The Langfuse SDK does not expose `enqueue` as a typed public API — if the SDK renames or removes this method, tool observations will silently stop being recorded with no TypeScript compile error. Since `getInternalTracingHandler` already uses the internal `_exportLocalEvents` hook, it would be safer to expose `enqueue` via a typed wrapper in `getInternalTracingHandler`'s return value so the dependency on the private API is contained in one place and gated by the SDK boundary.

```suggestion
    (
      this.langfuse as unknown as {
        enqueue: (type: string, body: ToolObservationBody) => void;
      }
    ).enqueue("tool-create", body); // NOTE: enqueue is an internal Langfuse SDK method — update if SDK is upgraded
```

### Issue 2 of 3
web/src/features/in-app-agent/server/handler.ts:290
The `environment` value is hardcoded as the string literal `"langfuse-in-app-agent"` rather than using `LangfuseInternalTraceEnvironment.InAppAgent` (the enum just added in `types.ts`). If the enum value is ever changed, this string would silently diverge.

```suggestion
                      environment: LangfuseInternalTraceEnvironment.InAppAgent,
```

### Issue 3 of 3
web/src/features/in-app-agent/server/instrumentation.ts:340
**`completionStartTime` always equals `startTime` for tool observations.** `completionStartTime` is the Langfuse field for time-to-first-token and is used to compute per-call latency in the UI. Setting it to `tool.startTime` instead of a real first-output timestamp means the TTFT metric for every tool call will always read as zero. While tool calls don't have a literal first token, setting it to `endTime` (or just omitting it) would better reflect actual latency than an always-zero value.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(agent): Trace in-app agent in langf..."](https://github.com/langfuse/langfuse/commit/42293b6bd486dad55af847a1d9240c306537a2a9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34937993)</sub>

### DIFF
--- a/web/src/features/in-app-agent/server/instrumentation.ts
+++ b/web/src/features/in-app-agent/server/instrumentation.ts
@@ -1,0 +1,399 @@
+import { EventType } from "@ag-ui/core";
+import { getInternalTracingHandler, logger } from "@langfuse/shared/src/server";
+
+import type {
+  AgUiEvent,
+  AgUiRunAgentInput,
+} from "@/src/features/in-app-agent/schema";
+import { compactTextMessageChunks } from "@/src/features/in-app-agent/server/eventCompaction";
+
+export type InAppAgentTracingConfig = {
+  environment: string;
+  metadata: Record<string, unknown>;
+  userId: string;
+  traceId: string;
+  targetProjectId: string;
+};
+
+export type InAppAgentInstrumentationParams = {
+  input: AgUiRunAgentInput;
+  tracing?: InAppAgentTracingConfig;
+};
+
+const IN_APP_AGENT_TRACE_NAME = "in-app-agent";
+const IN_APP_AGENT_SPAN_NAME = "agent-run";
+type InternalTracingHandler = ReturnType<typeof getInternalTracingHandler>;
+type InAppAgentTrace = ReturnType<
+  InternalTracingHandler["handler"]["langfuse"]["trace"]
+>;
+type InAppAgentSpan = ReturnType<InAppAgentTrace["span"]>;
+type InAppAgentLangfuse = InternalTracingHandler["handler"]["langfuse"];
+type ToolObservationBody = {
+  id: string;
+  traceId: string;
+  parentObservationId: string | null;
+  name: string;
+  startTime: Date;
+  endTime: Date;
+  completionStartTime: Date;
+  input?: unknown;
+  output?: unknown;
+  level?: "ERROR";
+  statusMessage?: string;
+  metadata?: Record<string, unknown>;
+};
+
+export function createInAppAgentInstrumentation({
+  input,
+  tracing,
+}: InAppAgentInstrumentationParams) {
+  if (!tracing?.targetProjectId) {
+    return undefined;
+  }
+
+  try {
+    return new InAppAgentInstrumentation({
+      input,
+      metadata: tracing.metadata,
+      userId: tracing.userId,
+      traceId: tracing.traceId,
+      targetProjectId: tracing.targetProjectId,
+      environment: tracing.environment,
+    });
+  } catch (error) {
+    logger.warn("Failed to initialize in-app agent Langfuse tracing", error);
+    return undefined;
+  }
+}
+
+export class InAppAgentInstrumentation {
+  private readonly processTracedEvents: () => Promise<void>;
+  private readonly langfuse: InAppAgentLangfuse;
+  private readonly trace: InAppAgentTrace;
+  private readonly span: InAppAgentSpan;
+  private readonly toolSpans = new Map<
+    string,
+    {
+      name: string;
+      startTime: Date;
+      args: string;
+      argsComplete: boolean;
+      output?: unknown;
+    }
+  >();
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by martingarramon — COMMENTED** [review]
The overall design is solid — event-type filtering, graceful abort/error paths, tool span lifecycle gated on both `argsComplete` and `output` present.

One lifecycle edge case: `recordEvents` does not appear to guard on `this.ended`. If a `TOOL_CALL_START` arrives after the root span has closed, `startToolSpan` could add a new entry to `this.toolSpans` after the final `endOpenToolSpans()` pass. Assuming no later abort/error/duplicate-finish path calls `endOpenToolSpans()` again, that span would remain open. This may require rare out-of-order AG-UI stream events, but adding `if (this.ended) return;` at the top of `recordEvents` would make the lifecycle explicit.

**Comment by bezbac** [comment]
@greptileai

File: web/src/features/in-app-agent/server/instrumentation.ts [hunk 1]
```diff
@@ -0,0 +1,399 @@
...
+
+    this.trace = this.langfuse.trace({
+      id: params.traceId,
+      name: IN_APP_AGENT_TRACE_NAME,
+      userId: params.userId,
+      sessionId: params.input.threadId,
+      metadata: params.metadata,
+      tags: ["in-app-agent"],
+    });
+    this.span = this.trace.span({
```
nimarb, line 118: unfortunately, we will always have span type `span` here -> OK for now but pre final release we should have proper observation types set

nimarb (approved):

web/src/features/in-app-agent/server/instrumentation.ts
bezbac: https://linear.app/langfuse/issue/LFE-10168/improve-instrumentation

PR #13776, opened 2026-05-21
fix(shared): bound S3 key segment length

### LINKED ISSUE #13476: bug: S3/MinIO event storage fails silently when observation IDs exceed ext4 NAME_MAX (255 bytes)
### Describe the bug

When using self-hosted Langfuse with MinIO as S3-compatible event storage, ingestion fails silently when an observation/generation ID exceeds the ext4 `NAME_MAX` limit of 255 bytes. The Langfuse server uses these IDs as part of S3 object key path segments, and MinIO (running on ext4) rejects keys with segments longer than 255 bytes.

This issue was triggered after switching from Anthropic Claude to OpenAI GPT-5.4 via litellm proxy. With Anthropic models, observation IDs were short enough to stay within limits. With OpenAI's Responses API, litellm generates much longer composite IDs that exceed the limit.

**Expected behavior:** Langfuse should handle arbitrarily long observation IDs when constructing S3 object keys — either by truncating, hashing, or otherwise sanitizing path segments to stay within filesystem limits.

**Actual behavior:** MinIO returns a 500 error with `XMinioInvalidObjectName: Object name contains unsupported characters`. The error causes `Failed to upload events to blob storage, aborting event processing`. Traces are silently lost — they never appear in the Langfuse UI.

### Steps to reproduce

1. Self-host Langfuse (v3.172.1) with MinIO as the S3 event storage backend
2. Use litellm (v1.83.14) as a proxy with Langfuse callbacks enabled
3. Configure litellm to route requests to OpenAI GPT-5.4
4. Send a chat completion request through litellm
5. litellm internally uses the OpenAI Responses API, which returns an ID like `resp_085c6f58432c85610069fa196832e08195b5f18086d8f5ce7e` (~54 chars)
6. litellm wraps this into a composite ID encoding provider and model metadata in base64: `resp_bGl0ZWxsbTpjdXN0b21f...` (~241 chars)
7. litellm's `get_logging_id()` prepends a time prefix, producing a generation ID of ~262 chars
8. The Langfuse SDK sends this as the observation ID to the Langfuse server
9. The Langfuse server uses this ID as a path segment in the S3 object key:
   ```
   <project_id>/observation/<262-char-observation-id>/<uuid>.json
   ```
10. MinIO rejects the request because the 262-char path segment exceeds ext4's `NAME_MAX` (255 bytes)

Self-hosted

v3.172.1

- Langfuse Python SDK: 2.60.10
- litellm: 1.83.14 (using Langfuse as a logging callback)
- Model: OpenAI GPT-5.4 via litellm proxy

## Why this worked with Anthropic but fails with OpenAI

When using **Anthropic Claude** through litellm, the observation IDs are short UUID-style strings. The resulting S3 path segment is well within limits:

```
Anthropic response ID (from litellm):  chatcmpl-fed4f3f0-f9df-4189-ab77-77371fa6bdb8  (45 chars)
get_logging_id() output:               time-16-23-03-560977_chatcmpl-fed4f3f0-f9df-4189-ab77-77371fa6bdb8  (66 chars)
→ Well under 255-byte limit ✅
```

When using **OpenAI GPT-5.4** through litellm, litellm wraps the original Responses API ID with base64-encoded metadata (provider name, model hash, original response ID):

```
Original OpenAI Responses API ID:      resp_085c6f58432c85610069fa196832e08195b5f18086d8f5ce7e  (54 chars)
litellm composite ID:                  resp_bGl0ZWxsbTpjdXN0b21fbGxtX3Byb3ZpZGVyOm9wZW5haTttb2RlbF9pZDo2YzNiMTUzNDViNmYzMzM0MjIzYWNkMmJmMGFhYWE2MDdhOWQxMWE2ODBhYjgwYmVhNTM5NzQ0YWUwN2FkMTQyO3Jlc3BvbnNlX2lkOnJlc3BfMDg1YzZmNTg0MzJjODU2MTAwNjlmYTE5NjgzMmUwODE5NWI1ZjE4MDg2ZDhmNWNlN2U_  (241 chars)
get_logging_id() output:               time-16-23-03-560977_resp_bGl0ZWxsbT...<truncated>...N2U_  (262 chars)
→ Exceeds 255-byte limit ❌
```

The base64 payload decodes to:
```
litellm:custom_llm_provider:openai;model_id:6c3b15345b6f3334223acd2bf0aaaa607a9d11a680ab80bea539744ae07ad142;response_id:resp_085c6f58432c85610069fa196832e08195b5f18086d8f5ce7e
```

### Actual error from Langfuse server logs

```
Failed to upload JSON to S3 cmosdjng70006uf07fl6jge60/observation/time-16-23-03-560977_resp_bGl0ZWxsbTpjdXN0b21fbGxtX3Byb3ZpZGVyOm9wZW5haTttb2RlbF9pZDo2YzNiMTUzNDViNmYzMzM0MjIzYWNkMmJmMGFhYWE2MDdhOWQxMWE2ODBhYjgwYmVhNTM5NzQ0YWUwN2FkMTQyO3Jlc3BvbnNlX2lkOnJlc3BfMDg1YzZmNTg0MzJjODU2MTAwNjlmYTE5NjgzMmUwODE5NWI1ZjE4MDg2ZDhmNWNlN2U_/32f9b50c-18e8-478d-b946-aa2723febb43.json Object name contains unsupported characters.
XMinioInvalidObjectName: Object name contains unsupported characters.
```

```
Failed to upload event to S3
Error: Failed to upload events to blob storage, aborting event processing
```

### Current workaround

Patched litellm's `get_logging_id()` to truncate and sanitize IDs before they reach Langfuse:

```python
def get_logging_id(start_time, response_obj):
    raw_id = response_obj.get("id") or ""
    import re
    safe_id = re.sub(r'[;=+/]', '_', raw_id)
    time_prefix = "time-" + start_time.strftime("%H-%M-%S-%f") + "_"
    max_id_len = 240 - len(time_prefix)
    if len(safe_id) > max_id_len:
        safe_id = safe_id[:max_id_len]
    return time_prefix + safe_id
```

### Suggested fix

The Langfuse server should validate and sanitize S3 object key path segments before writing to storage. Regardless of how long an observation ID is, S3 keys should never exceed filesystem limits. For example:

```python
import hashlib

def safe_s3_key_segment(segment: str, max_len: int = 200) -> str:
    if len(segment) <= max_len:
        return segment
    hash_suffix = hashlib.sha256(segment.encode()).hexdigest()[:16]
    return segment[:max_len - 17] + "_" + hash_suffix
```

No

Epochex:
Fixes #13476.

MinIO running on ext4 rejects object keys with path segments >255 bytes. We currently embed `event.body.id` into both the directory segment and the file name for S3 event uploads, which can exceed that limit for long provider-generated IDs.

This change bounds each S3 key segment to <=255 bytes by truncating and appending a stable hash.

Tests:
- pnpm --filter @langfuse/shared run lint
- pnpm --filter @langfuse/shared run db:generate
- pnpm --filter @langfuse/shared run test
- pnpm --filter web run lint
- pnpm --filter worker run lint

This PR bounds S3 object-key path segments to ≤255 bytes by truncating long `eventBodyId` values (such as provider-generated IDs like `resp_aaa…`) and appending a 16-char SHA-256 hash before the upload, fixing MinIO/ext4 rejections.

- **Write side fixed, read side not updated**: The upload path in `processEventBatch.ts` correctly uses the truncated `safeEventBodyId`, but the raw `eventBodyId` is still enqueued in the Redis job payload. The worker (`worker/src/queues/ingestionQueue.ts`, line 161) rebuilds the S3 prefix directly from `job.data.payload.data.eventBodyId` without any truncation, so when an ID exceeds 255 bytes the worker constructs a path that was never written — causing a silent S3 list/download miss and dropping all events for that entity.
- The `BlobStorageFileLog` entry (line 75 of the same worker file) records the same raw `eventBodyId` in `bucket_path`, so retention and deletion operations would target a non-existent path.
- The accompanying test validates the upload path correctly but does not cover the worker's reconstruction logic.
</details>

The upload side is fixed, but the worker still reads using the untransformed eventBodyId, so any event whose body ID exceeds 255 bytes will be silently lost after this change ships.

The truncation logic itself is sound and the test for the write path is well-structured. The gap is that worker/src/queues/ingestionQueue.ts reconstructs the S3 prefix from the raw queue payload without applying the same transformation, turning a previously-rejected upload into a silently-dropped event. This is not a theoretical risk — it is the normal code path for any ingestion event whose provider-generated ID is longer than 255 characters.

worker/src/queues/ingestionQueue.ts (lines 75 and 161) needs to apply the same segment-length transformation when reconstructing S3 paths from the queue payload.
</details>

```mermaid
sequenceDiagram
    participant API as API (processEventBatch.ts)
    participant S3 as S3 / MinIO
    participant Queue as Redis IngestionQueue
    participant Worker as Worker (ingestionQueue.ts)

    API->>API: "safeEventBodyId = safeS3ObjectKeySegment(eventBodyId)"
    Note over API: truncates to ≤255 bytes + hash suffix
    API->>S3: "upload to .../{safeEventBodyId}/{safeFileName}.json"
    API->>Queue: "enqueue { eventBodyId: rawEventBodyId, fileKey: rawKey }"
    Note over Queue: raw (un-truncated) eventBodyId stored

    Worker->>Queue: dequeue job
    Worker->>Worker: build prefix using raw eventBodyId
    Note over Worker: .../{rawEventBodyId}/ DOES NOT MATCH S3 path
    Worker->>S3: "listFiles(.../{rawEventBodyId}/)"
    S3-->>Worker: [] (no files found — wrong directory)
    Worker->>Worker: log warning No events found, return
    Note over Worker: Events silently dropped
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/ingestion/processEventBatch.ts:359-385
**Worker–producer path mismatch on long `eventBodyId`**

The queue payload carries the raw `eventBodyId: eventData.eventBodyId` (potentially >255 bytes), but `worker/src/queues/ingestionQueue.ts` (line 161) reconstructs the S3 prefix with that same raw value — `…/${job.data.payload.data.eventBodyId}/` — without applying `safeS3ObjectKeySegment`. When `eventBodyId` exceeds 255 bytes, the file is uploaded to `${safeEventBodyId}/…` but the worker reads from `${rawEventBodyId}/…`, returning zero files from S3 and silently dropping all events for that entity. The same raw value is also logged in `bucket_path` of the `BlobStorageFileLog` entry (line 75 of `ingestionQueue.ts`), so retention/deletion records will point to a path that does not exist.

The fix requires either exporting `safeS3ObjectKeySegment` from the shared package and calling it in the worker's prefix construction, or including the truncated form as a separate field in the queue payload.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(shared): bound S3 key segment length"](https://github.com/langfuse/langfuse/commit/5f8335e8b6fe7edca1e03cbce93c3013b79ef23c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33239946)</sub>

### DIFF
--- a/packages/shared/src/server/ingestion/processEventBatch.s3KeySegments.test.ts
+++ b/packages/shared/src/server/ingestion/processEventBatch.s3KeySegments.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it, vi } from "vitest";
+
+const uploadedPaths: string[] = [];
+
+vi.mock("../../env", () => ({
+  env: {
+    LANGFUSE_S3_EVENT_UPLOAD_PREFIX: "events/",
+    LANGFUSE_S3_EVENT_UPLOAD_BUCKET: "langfuse",
+    LANGFUSE_S3_EVENT_UPLOAD_ACCESS_KEY_ID: "minio",
+    LANGFUSE_S3_EVENT_UPLOAD_SECRET_ACCESS_KEY: "miniosecret",
+    LANGFUSE_S3_EVENT_UPLOAD_ENDPOINT: "http://localhost:9000",
+    LANGFUSE_S3_EVENT_UPLOAD_REGION: "auto",
+    LANGFUSE_S3_EVENT_UPLOAD_FORCE_PATH_STYLE: "true",
+    LANGFUSE_S3_EVENT_UPLOAD_SSE: null,
+    LANGFUSE_S3_EVENT_UPLOAD_SSE_KMS_KEY_ID: null,
+    LANGFUSE_INGESTION_QUEUE_DELAY_MS: 0,
+    LANGFUSE_SKIP_S3_LIST_FOR_OBSERVATIONS_PROJECT_IDS: undefined,
+  },
+}));
+
+vi.mock("../instrumentation", () => ({
+  getCurrentSpan: () => ({
+    setAttribute: vi.fn(),
+  }),
+  instrumentAsync: (_opts: unknown, fn: () => Promise<unknown>) => fn(),
+  recordDistribution: vi.fn(),
+  recordIncrement: vi.fn(),
+  traceException: vi.fn(),
+}));
+
+vi.mock("../logger", () => ({
+  logger: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+vi.mock("../redis/s3SlowdownTracking", () => ({
+  isS3SlowDownError: () => false,
+  markProjectS3Slowdown: vi.fn(),
+}));
+
+vi.mock("../services/StorageService", () => ({
+  StorageService: class StorageService {},
+  StorageServiceFactory: {
+    getInstance: () => ({
+      uploadJson: async (path: string) => {
+        uploadedPaths.push(path);
+        throw new Error("boom");
+      },
+    }),
+  },
+}));
+
+vi.mock("./types", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("./types")>();
+  const { z } = await import("zod");
+
+  const minimalSchema = z
+    .object({
+      id: z.string(),
+      type: z.string(),
+      timestamp: z.string(),
+      body: z.object({ id: z.string() }).loose(),
+    })
+    .loose();
+
+  return {
+    ...actual,
+    createIngestionEventSchema: () => minimalSchema,
+  };
+});
+
+describe("processEventBatch S3 key segmentation", () => {
+  it("keeps every S3 key path segment <= 255 bytes", async () => {
+    const { processEventBatch } = await import("./processEventBatch");
+    const { eventTypes } = await import("./types");
+
+    const longId = `resp_${"a".repeat(300)}`;
+    const now = new Date().toISOString();
+
+    await expect(
+      processEventBatch(
+        [
+          {
+            id: "evt_1",
+            type: eventTypes.TRACE_CREATE,
+            timestamp: now,
+            body: { id: longId },
+          },
+        ],
+        {
+          validKey: true,
+          scope: {
+            projectId: "proj_123",
+            accessLevel: "project",
+          },
+        } as any,
+      ),
+    ).rejects.toThrow("Failed to upload events to blob storage");
+
+    expect(uploadedPaths).toHaveLength(1);
+
+    const segments = uploadedPaths[0].split("/").filter(Boolean);
+    for (const segment of segments) {
+      expect(Buffer.byteLength(segment, "utf8")).toBeLessThanOrEqual(255);
+    }
+  });
+});
--- a/packages/shared/src/server/ingestion/processEventBatch.ts
+++ b/packages/shared/src/server/ingestion/processEventBatch.ts
@@ -1,4 +1,4 @@
-import { randomUUID } from "crypto";
+import { createHash, randomUUID } from "crypto";
 import { z } from "zod";
 
 import { env } from "../../env";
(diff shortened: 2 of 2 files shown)


sumerman:
This fix is incomplete and may cause problems on existing deployments. See Greptile review. I am working on a proper fix but it may take some time before it lands.

Epochex:
Thanks for calling this out. I agree this shouldn’t merge as-is.

After looking through the worker path too, the write-side truncation leaves compatibility gaps around worker reads, blob file logs/deletion, and replay/rolling deploy behavior. That needs a more complete fix than this PR, so I’m closing it to avoid getting in the way of the proper implementation.

PR #13791, opened 2026-05-22
fix(tracing): sanitize external urls wholistically

nimarb:

This PR replaces the DOMPurify-based URL sanitisation in `MarkdownViewer` with a new `safe-url.ts` module (`getSafeLinkUrl` / `getSafeImageUrl`) that uses the WHATWG `new URL()` parser for protocol validation. `ResizableImage` is updated to call the new helpers so every link and image source is validated at the point of use before it reaches the DOM.

- **`safe-url.ts`**: New lightweight URL allow-listing module using `URL` parsing; supports safe protocols (`http:`, `https:`, `mailto:`, `tel:`), fragment-only, and same-origin absolute paths.
- **`MarkdownViewer.tsx` / `resizable-image.tsx`**: All link `href` and image `src` values are now passed through the new helpers before rendering; unsafe URLs are silently dropped and shown as plain text instead.
- **`safe-url.clienttest.ts`**: Client-side unit tests covering XSS vectors (`javascript:`, `vbscript:`, `data:`, `//` protocol-relative, relative paths) and allowed cases.
</details>

The change moves URL sanitisation in the right direction and is safe to merge; the issues found are code-quality concerns rather than exploitable vulnerabilities.

The core allow-list logic is sound and the major XSS vectors are correctly blocked. The main concern is that getSafeUrl returns the raw trimmed input rather than the WHATWG-normalised parsed.href, creating a subtle mismatch between what is validated and what is returned. Additionally, blocking data:image/… URIs is an undocumented behaviour change from the previous DOMPurify approach with no test coverage.

safe-url.ts (return value normalisation) and safe-url.clienttest.ts (missing data URI and protocol-relative coverage for image URLs)
</details>

```mermaid
flowchart TD
    A["Raw URL string\n(from Markdown / OpenAI content)"] --> B["getSafeLinkUrl / getSafeImageUrl"]
    B --> C["trim whitespace"]
    C --> D["new URL(trimmed)"]
    D -->|"throws (relative / fragment)"| E{"allowHash or\nallowAbsolutePath?"}
    D -->|"succeeds"| F{"parsed.protocol\nin allowedProtocols?"}
    F -->|"yes"| G["return trimmed ✓"]
    F -->|"no (javascript:, data:, ftp:, …)"| H["return null ✗"]
    E -->|"yes + starts with # or /"| I["return trimmed ✓"]
    E -->|"no"| J["return null ✗"]
    G --> K["Link href / Image src / ResizableImage"]
    I --> K
    H --> L["Render as plain text or null"]
    J --> L
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/components/ui/safe-url.ts:22-23
The function validates `parsed.protocol` but returns the original `trimmed` string rather than `parsed.href`. The WHATWG URL parser silently strips `	`, `
`, and `
` from URLs before processing, so a value like `"https://example.com/foo	bar"` would pass the protocol check and be returned with the tab still embedded. Using `parsed.href` returns the fully-normalised form and guarantees that what is checked equals what is returned downstream.

```suggestion
    const parsed = new URL(trimmed);
    return allowedProtocols.has(parsed.protocol) ? parsed.href : null;
```

### Issue 2 of 3
web/src/components/ui/resizable-image.tsx:123
`safeSrc` is already guaranteed to be truthy at this point in the render tree (the surrounding condition is `isImageVisible && safeSrc && isValidImage.data?.isValid`), so the `?? src` fallback is dead code and never evaluates.

```suggestion
                title={safeSrc}
```

### Issue 3 of 3
web/src/components/ui/safe-url.clienttest.ts:35-41
**Missing test for `data:image/…` URLs in `getSafeImageUrl`**

The previous DOMPurify implementation permitted `data:image/png;base64,…` URIs for `<img>` tags (DOMPurify allows `data:image` on image elements by default). The new `SAFE_IMAGE_PROTOCOLS` set contains only `"http:"` and `"https:"`, so base64-encoded images embedded directly in AI-generated content will silently stop rendering. Adding an explicit test case — and a comment in `safe-url.ts` — would document this intentional behaviour change and prevent future regressions.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(tracing): sanitize external urls who..."](https://github.com/langfuse/langfuse/commit/ca2fa104247433defbbf665cd602ca2865ceb1f6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33424024)</sub>

### DIFF
--- a/web/src/components/ui/MarkdownViewer.tsx
+++ b/web/src/components/ui/MarkdownViewer.tsx
@@ -344,7 +344,8 @@ function MarkdownRenderer({
               );
             },
             img({ src, alt }) {
-              const safeSrc = getSafeImageUrl(src);
+              const safeSrc =
+                typeof src === "string" ? getSafeImageUrl(src) : null;
               return safeSrc ? (
                 <ResizableImage src={safeSrc} alt={alt} />
               ) : null;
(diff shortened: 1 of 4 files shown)


nimarb:
@claude review

PR #13795, opened 2026-05-22
feat(mcp): Handle numeric colums in `getObservationFilterValues`

bezbac:

This PR extends the MCP `getObservationFilterValues` tool to handle numeric columns (`totalCost`, `totalTokens`, `latency`, `timeToFirstToken`, `promptVersion`) by returning a `{ min, max, avg, count }` range instead of paginated string values. It also adds `promptVersion` as a newly queryable filter column.

- **New ClickHouse query** (`getEventsNumericStatsByFilterColumn`): correctly uses `isNotNull(expression)` to filter rows, aggregates into a single bucket with `GROUP BY tuple()`, and maps a ClickHouse empty-result `NULL` aggregate to `null` in TypeScript before returning to the caller.
- **Dispatch logic**: `isNumericEventsTableColumnId` type-guards `eventsColumn` in the handler, routing numeric columns to the RANGE path and narrowing the type so non-numeric columns still satisfy `getEventFilterValuePage`'s explicit column union.
- **`inputTokens`, `outputTokens`, `inputCost`, `outputCost` correctly excluded**: their `arraySum`-based expressions always return `0` rather than `NULL`, making the `isNotNull` filter a no-op; the type-level exclusion prevents them from reaching the new query.
</details>

Safe to merge; the new numeric range query path is well-guarded and the previously concerning null-aggregate edge case is handled correctly.

The dispatch logic, TypeScript narrowing, and null-handling for empty ClickHouse results are all correct. The four arraySum columns that would silently return wrong results are excluded at the type level. The only gap is that commentCount and tokensPerSecond remain reachable via the function type signature, but no current code passes them.

No files require special attention for merging.
</details>

```mermaid
sequenceDiagram
    participant Client as MCP Client
    participant Tool as getObservationFilterValues
    participant Service as eventsService
    participant Repo as events.ts (ClickHouse)

    Client->>Tool: "{ column: "totalCost", ... }"
    Tool->>Tool: isNumericEventsTableColumnId("totalCost") true
    Tool->>Service: "getEventFilterNumericRange({ column: "totalCost", ... })"
    Service->>Repo: getEventsNumericStatsByFilterColumn(projectId, filter, "totalCost")
    Repo->>Repo: SELECT min(if(...)), max(...), avg(...), count(...) WHERE isNotNull(...)
    Repo-->>Service: "[{min, max, avg, count}] or []"
    Service-->>Tool: "{min, max, avg, count} or null"
    Tool-->>Client: "{type: RANGE, column: totalCost, range: {...} or null}"

    Client->>Tool: "{ column: "sessionId", ... }"
    Tool->>Tool: isNumericEventsTableColumnId("sessionId") false
    Tool->>Service: "getEventFilterValuePage({ column: "sessionId", limit, offset })"
    Service->>Repo: "getEventsGroupedBySessionId(projectId, filter, {limit, offset})"
    Repo-->>Service: "[{sessionId, count}, ...]"
    Service-->>Tool: "{values: [...], nextOffset?}"
    Tool-->>Client: "{type: VALUES, column: sessionId, values: [...], meta: {cursor?}}"
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/repositories/events.ts:1989-1996
`commentCount` has `internal: ""` (empty string) and `tokensPerSecond` references `start_time`/`end_time` without the `e.` table alias. Both are `NumericEventsTableColumnId` but are not excluded from `getEventsNumericStatsByFilterColumn`'s type signature the way `inputTokens` etc. are. If either column ID is ever added to a caller's column list, the generated SQL will be malformed (`min() as min` or an unresolved column reference). Extending the exclusion union to cover these prevents a silent footgun.

```suggestion
export const getEventsNumericStatsByFilterColumn = async (
  projectId: string,
  filter: FilterState,
  columnId: Exclude<
    NumericEventsTableColumnId,
    | "inputTokens"
    | "outputTokens"
    | "inputCost"
    | "outputCost"
    | "commentCount"
    | "tokensPerSecond"
  >,
) => {
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(mcp): Handle numeric colums in \`get..."](https://github.com/langfuse/langfuse/commit/4f8ed95bee97d2ef327d9b6b88194cb1650dcd65) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33439642)</sub>

### DIFF
--- a/packages/shared/src/eventsTable.ts
+++ b/packages/shared/src/eventsTable.ts
@@ -357,6 +357,18 @@ export const eventsTableCols =
 
 type EventsTableColumnId = (typeof eventsTableColsDefinition)[number]["id"];
 
+export type NumericEventsTableColumnId = Extract<
+  (typeof eventsTableColsDefinition)[number],
+  { type: "number" }
+>["id"];
+
+export const isNumericEventsTableColumnId = (
+  column: EventsTableColumnId,
+): column is NumericEventsTableColumnId =>
+  eventsTableColsDefinition.some(
+    (col) => col.id === column && col.type === "number",
+  );
+
 // Subset of columns that are allowed to be used as filters in the MCP observations API
 const OBSERVATION_MCP_ALLOWED_EVENTS_TABLE_FILTER_COLUMN_IDS = [
   "id",
--- a/packages/shared/src/server/repositories/events.ts
+++ b/packages/shared/src/server/repositories/events.ts
@@ -100,6 +100,7 @@ import {
   eventsTableCols,
   eventsTableHasParentObservationSql,
   eventsTableIsRootObservationSql,
+  type NumericEventsTableColumnId,
 } from "../../eventsTable";
 import {
   findUiColumnMapping,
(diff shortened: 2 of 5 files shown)


bezbac:
@greptileai

martingarramon (commented):
Two things from an agent-reliability angle:

**Cursor ignored silently on numeric columns**

When `isNumericEventsTableColumnId(eventsColumn)` is true, the handler returns early before `decodeObservationFilterValueCursor` runs. An agent that mistakenly passes a cursor with a numeric column will get a full RANGE response with no indication the cursor was dropped. Either document the behavior ("cursor is not applicable to numeric columns and will be ignored") or throw `InvalidRequestError` so the caller can detect the mismatch. Which path is intended here?

**`range: null` case not surfaced in the description**

The ClickHouse stats function returns `null` when no observations match the filter. The current description says "returns a range with min, max, avg, and count" — but an agent could receive `range: null`. Noting this case ("returns null if no matching observations exist") would let agents handle it explicitly rather than guarding defensively.

Minor: `avg(promptVersion)` is numerically valid but semantically odd — `min`/`max`/`count` are more useful here. Not a blocker, just noting it in case there's a simpler path (e.g. returning just `{min, max, count}` for version-like columns).

nimarb (approved):

PR #13810, opened 2026-05-23
fix(public-api): count v2 score IDs consistently

### LINKED ISSUE #13559: bug: `/api/public/v2/scores` — `meta.totalItems` still overcounts (deleted-row residue; follow-up to #6396)
### Describe the bug

`GET /api/public/v2/scores` returns a `meta.totalItems` that overcounts the actual `data[]` length. The maintainer comment on [#6396](<https://github.com/langfuse/langfuse/issues/6396>) stated the new `/v2/scores` route "should not be affected by this" — confirming here that it is still affected, but by a different underlying mechanism than the original report: **deleted score rows remain counted in** `meta.totalItems` while being correctly excluded from `data[]`.

This is essentially a regression-follow-up on [#6396](<https://github.com/langfuse/langfuse/issues/6396>). The session/dataset-score type-mismatch path described in that thread does not apply here (project has no session-attached or dataset-attached scores).

### To reproduce

Langfuse Cloud (US), project `cml6xji63001pad08i3odtvh3`, observed 2026-05-11.

```bash
curl -u "$PK:$SK" "$BASE/api/public/v2/scores?limit=100" \
  | jq '{data_len:(.data|length), total:.meta.totalItems}'
```

```json
{ "data_len": 68, "total": 82 }
```

The gap narrows as `fromTimestamp` is moved forward and collapses once the window excludes a known DELETE event:

| `fromTimestamp` | `data.length` | `meta.totalItems` | gap |
| -- | -- | -- | -- |
| (all-time) | 68 | 82 | **14** |
| 2026-04-01 | 67 | 81 | **14** |
| 2026-05-01 | 39 | 52 | **13** |
| 2026-05-08 | 14 | 14 | **0** |

Per-name decomposition isolates the gap to five score names:

| `name` filter | `data.length` | `meta.totalItems` | gap |
| -- | -- | -- | -- |
| `1a-i-cfu-error-floor` | 13 | 23 | **10** |
| `1b-misconception-surfacing` | 3 | 4 | 1 |
| `2a-i-bucket-detection` | 2 | 3 | 1 |
| `2c-bucket-non-example-callout` | 1 | 2 | 1 |
| `2d-bucket-absence-framing` | 1 | 2 | 1 |
| all others (11 names) | 49 | 49 | 0 |
| **total** | **68** | **82** | **14** |

The 10-row gap on `1a-i-cfu-error-floor` matches a known event: on 2026-05-05 we renamed legacy `1a-i-per-turn-error-floor` scores by `POST /api/public/scores` with a new name and `DELETE /api/public/scores/{old_id}` for the legacy id. The 12 visible rows on the new name carry `metadata.renamed_from = "1a-i-per-turn-error-floor"` and `metadata.renamed_at = "2026-05-05"`. **The 10 DELETEd legacy rows are still counted in** `totalItems`. The other four names show the same pattern at +1 each from one-off DELETEs during scorer iteration.

Control: a different score name (`4a-iv-engagement-signal--RETIRED-2026-05-11`) was renamed today via an in-place mechanism (no DELETE+POST). That name returns `data=12 / totalItems=12` — no phantom residue. So the residue is specifically tied to the DELETE path, not to renames generally.

Per-trace evidence (rules out trace-orphan / type-mismatch explanations): `?traceId=a236f702-e26c-443e-b923-c8c15f3306cd` returns `data=4 / totalItems=5`. The trace itself still exists; the phantom is at the score level.

Pagination is internally consistent with `data.length` (page 1 with `limit=50` returns 50, page 2 returns 18 → 68 total), so the bug is solely in the counter source, not in the page iterator.

Cloud (US). No SDK involved — reproduced directly with `curl`.

### Additional information

Practical impact: `meta.totalItems` is the natural fanout-completion signal for scoring pipelines ("did all my .POSTs land?"). Switching to `data.length` works but is surprising given the field name. We've added a note in our internal docs warning operators not to trust `totalItems` for assertions.

Suggested fix: compute `meta.totalItems` against the same view used to populate `data[]`, so soft-deleted/tombstoned rows are excluded from both. If the underlying storage retains tombstones for audit, the counter should filter them out at projection time the same way the data path does.

No (no access to the relevant service layer).

Epochex:
Fixes langfuse/langfuse#13559.

`GET /api/public/v2/scores` returns one row per score ID (via `LIMIT 1 BY s.id`), but `meta.totalItems` was counting raw rows and could overcount when multiple score versions exist.

This updates `_handleGetScoresCountForPublicApi` to count the same `LIMIT 1 BY s.id` result set, and adds a regression test covering duplicate score rows.

Verification:
- pnpm.cmd --filter web run lint

Local server tests couldn’t run here because the Postgres test DB credentials in `.env.test` aren’t configured on this machine.

This PR fixes a count discrepancy in `GET /api/public/v2/scores` where `meta.totalItems` overcounted when multiple versions of the same score ID existed in ClickHouse. The fix restructures `_handleGetScoresCountForPublicApi` to first collect deduplicated score IDs (using the same `LIMIT 1 BY s.id, s.project_id` logic as the data query) and then wraps that result in an outer `count()`.

- **Count query refactor**: replaces the flat `SELECT count()` with a two-level query — an inner `scoreIdsQuery` that mirrors the deduplication logic of the data query, wrapped by an outer `SELECT count() FROM (...)`.
- **Consistency fix for `traces_only` scope**: the count query's inner subquery was only guarding `s.session_id IS NULL`, omitting `s.dataset_run_id IS NULL`; this PR adds the missing clause to match the data query.
- **Regression test**: adds a server test that inserts two rows sharing the same score ID (simulating a versioned score) and asserts that `totalItems` is 1 and the returned `value` matches the newer row.
</details>

Safe to merge — the change fixes a real count inflation bug and the new query logic precisely mirrors the deduplication already applied to the data fetch path.

Both the count and data queries now apply the same LIMIT 1 BY s.id, s.project_id deduplication, eliminating the mismatch. The additional dataset_run_id IS NULL guard in the traces_only scope aligns the count path with the existing data path. A focused regression test covers the core scenario. No functional regressions or logic gaps were identified.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(public-api): count v2 score IDs cons..."](https://github.com/langfuse/langfuse/commit/6f987cceef68a7a02750061b7754cf63d2f1ba6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33609460)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/scores.ts
+++ b/packages/shared/src/server/repositories/scores.ts
@@ -2560,9 +2560,9 @@ export const _handleGetScoresCountForPublicApi = async ({
   const appliedScoresFilter = scoresFilter.apply();
   const appliedTracesFilter = tracesFilter.apply();
 
-  const query = `
+  const scoreIdsQuery = `
       SELECT
-        count() as count
+        s.id as id
       FROM
         scores s
           ${needsTraceJoin ? "LEFT JOIN __TRACE_TABLE__ t ON s.trace_id = t.id AND s.project_id = t.project_id" : ""}
@@ -2578,15 +2578,27 @@ export const _handleGetScoresCountForPublicApi = async ({
           WHERE
             s.project_id = {projectId: String}
             ${appliedScoresFilter.query ? `AND ${appliedScoresFilter.query}` : ""}
-            ${scoreScope === "traces_only" ? "AND s.session_id IS NULL" : ""}
+            ${scoreScope === "traces_only" ? "AND s.session_id IS NULL AND s.dataset_run_id IS NULL" : ""}
           ORDER BY
             s.timestamp desc
           LIMIT
             1 BY s.id, s.project_id
         ))
       )
+      ${scoreScope === "traces_only" ? "AND s.session_id IS NULL AND s.dataset_run_id IS NULL" : ""}
       ${appliedScoresFilter.query ? `AND ${appliedScoresFilter.query}` : ""}
       ${tracesFilter.length() > 0 ? `AND ${appliedTracesFilter.query}` : ""}
+      ORDER BY
+        s.timestamp desc, s.event_ts desc
+      LIMIT
+        1 BY s.id, s.project_id
+      `;
+
+  const query = `
+      SELECT
+        count() as count
+      FROM
+        (${scoreIdsQuery})
       `;
 
   return measureAndReturn({
--- a/web/src/__tests__/server/scores-api-v2.servertest.ts
+++ b/web/src/__tests__/server/scores-api-v2.servertest.ts
@@ -510,6 +510,64 @@ describe("/api/public/v2/scores API Endpoint", () => {
         );
       });
 
+      it("counts unique score IDs when multiple score versions exist", async () => {
+        const { projectId, auth } = await createOrgProjectAndApiKey();
+        const traceId = v4();
+        const scoreId = v4();
+
+        await createTracesCh([
+          createTrace({ id: traceId, project_id: projectId }),
+        ]);
+
+        const oldScore = createTraceScore({
+          id: scoreId,
+          project_id: projectId,
+          trace_id: traceId,
+          name: "versioned-score",
+          value: 1,
+          data_type: "NUMERIC",
+          timestamp: 1,
+          event_ts: 1,
+          created_at: 1,
+          updated_at: 1,
+        });
+        const newScore = createTraceScore({
+          id: scoreId,
+          project_id: projectId,
+          trace_id: traceId,
+          name: "versioned-score",
+          value: 2,
+          data_type: "NUMERIC",
+          timestamp: 2,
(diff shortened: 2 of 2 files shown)


martingarramon (commented):
LGTM. The count query now mirrors the data fetch — inner deduplication by ID, outer count — so `meta.totalItems` and actual result length are consistent.

Minor: the count's inner `LIMIT 1 BY s.id, s.project_id` differs from the data fetch's `LIMIT 1 BY s.id`. Equivalent now given the outer `project_id` WHERE, but worth aligning if either query is refactored independently.

Epochex:
Thanks for checking. I double-checked the current PR head: the data query and count query both use LIMIT 1 BY s.id, s.project_id now, so they are aligned. I’ll leave this as-is to avoid a churn-only change.

PR #13813, opened 2026-05-23
fix(comments): add stable orderBy to paginated listComments queries

### LINKED ISSUE #13812: bug: listComments MCP tool and GET /api/public/comments return non-deterministic pages — findMany missing orderBy
## Describe the bug

`listComments` MCP tool and the REST `GET /api/public/comments` endpoint both use `prisma.comment.findMany` without an `orderBy` clause. PostgreSQL does not guarantee a stable row order without `ORDER BY`, so paginated calls (using `take`/`skip`) can return duplicate rows or miss rows when the planner changes query plans or when rows are inserted concurrently between page fetches.

## Steps to reproduce

**Static verification (no setup required):**
Inspect the two `findMany` calls cited below. Both combine `take`/`skip` with no `orderBy`. In PostgreSQL, `OFFSET` pagination without `ORDER BY` carries no ordering guarantee.

**Runtime reproduction:**
1. Create >50 comments on a project.
2. Call `listComments` with `page=1, limit=50`; record the returned `id` list.
3. Call `listComments` with `page=2, limit=50`.
4. Repeat both calls under concurrent comment creation.
5. Results on page 1 and page 2 may overlap or have gaps.

**REST equivalent:**
```bash
curl -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" \
  "https://cloud.langfuse.com/api/public/comments?page=1&limit=50"

curl -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" \
  "https://cloud.langfuse.com/api/public/comments?page=2&limit=50"
# Order is not guaranteed to be stable between calls
```

## Root cause

**MCP:** `web/src/features/mcp/features/comments/tools.ts`, `listComments` handler, lines 124–128:
```typescript
prisma.comment.findMany({
  where,
  take: input.limit,
  skip: (input.page - 1) * input.limit,
})
```

**REST:** `web/src/pages/api/public/comments/index.ts`, lines 65–74 — same `findMany` with `take`/`skip`, no `orderBy`.

Compare: `listDatasets` (MCP, `web/src/features/mcp/features/datasets/tools.ts:165`) includes an `orderBy` clause before `take`/`skip`. The comments handler has none.

The only paginated public API endpoint for listing comments is `GET /api/public/comments` and its MCP mirror `listComments`. Other comment-listing code paths in the codebase either include `ORDER BY` or are not paginated.

## Suggested fix

Add a stable sort to both locations:
```typescript
orderBy: [{ createdAt: "desc" }, { id: "asc" }],
```

`id` as secondary sort provides a unique tiebreaker when two comments share a `createdAt` timestamp, making pagination fully deterministic.

## Langfuse Cloud or self-hosted?

Both — the bug is in application code, not environment-specific.

## Additional information

The MCP tool was added in PR #13783 (2026-05-22) mirroring the REST handler. Both the MCP handler and REST endpoint should use the same stable ordering so their paginated responses have consistent semantics.

martingarramon:
## What does this PR do?

Adds `orderBy: [{ createdAt: "desc" }, { id: "asc" }]` to the two `prisma.comment.findMany` calls that power paginated comment listing — one in the MCP `listComments` tool and one in `GET /api/public/comments`.

## Why

An agent may page through comments across turns or after a checkpoint/resume. Without `ORDER BY`, PostgreSQL returns rows in undefined order — causing an agent to skip a comment it should act on, or process the same one twice. Paginated `OFFSET`/`LIMIT` queries against an unordered result set are also vulnerable to duplicates and gaps when the query planner changes plans or rows are written concurrently between page fetches. Fixes #13812.

## Type of change

## Mandatory Tasks

## Checklist

- [x] Added a pagination-stability test: inserts 6 comments with identical `createdAt` timestamps, then asserts that fetching pages 1 and 2 produces no overlap, no gaps, and identical page 1 results on repeated calls — this covers exactly where timestamp-only ordering stays flaky.

### DIFF
--- a/web/src/__tests__/server/mcp-tools-read.servertest.ts
+++ b/web/src/__tests__/server/mcp-tools-read.servertest.ts
@@ -127,6 +127,10 @@ import {
   getMediaTool,
   handleGetMedia,
 } from "@/src/features/mcp/features/media/tools/getMedia";
+import {
+  listCommentsTool,
+  handleListComments,
+} from "@/src/features/mcp/features/comments/tools";
 
 const maybeEventsTable =
   env.LANGFUSE_ENABLE_EVENTS_TABLE_V2_APIS === "true"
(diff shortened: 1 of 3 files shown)


nimarb:
@claude review

nimarb:
hi @martingarramon thank you for the fix, this has already been added.

PR #13815, opened 2026-05-24
fix(public-api): honor events table env fallback

### LINKED ISSUE #13814: Public API `useEventsTable` fallback makes `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS` ineffective when the query param is omitted
## Summary

## Affected files

## Current behavior

```ts
  .optional()
```

The route code then does:

```ts
const useEventsTable =
```

## Why this is a bug

- explicit query override

```ts
```

## Repro

2. Call:

## History

## Suggested fix

```ts
  .optional()
  .transform((val) =>
  );
```

Then simplify route usage to:

```ts
const useEventsTable =
  query.useEventsTable ??
```

for all three routes.

## Test gap

fallintoplace:
## Summary

- preserve `undefined` in `useEventsTableSchema` when the query param is omitted
- use the env fallback consistently in the public traces and observations routes
- add regression coverage for omitted-param fallback and explicit `useEventsTable=false` overrides

Fixes #13814.

## Verification

- `pnpm --filter @langfuse/shared exec vitest run src/features/query/types.test.ts`
- `pnpm --filter @langfuse/shared run lint -- src/features/query/types.ts src/features/query/types.test.ts`
- `pnpm --filter web run lint -- src/pages/api/public/traces/index.ts src/pages/api/public/observations/index.ts 'src/pages/api/public/observations/[observationId].ts' src/__tests__/server/observations-api.servertest.ts src/__tests__/server/observation-api.servertest.ts src/__tests__/server/traces-api.servertest.ts`

Targeted web server tests were attempted, but local infra was unavailable in this environment:

- `pnpm --filter web exec dotenv -e ../.env.test -e ../.env -- vitest run --project server src/__tests__/server/observations-api.servertest.ts src/__tests__/server/observation-api.servertest.ts src/__tests__/server/traces-api.servertest.ts`
- blocked because the test harness could not reach Postgres on `localhost:5432`, and the local Docker daemon/socket was not running

This PR fixes a bug where `useEventsTableSchema`'s `.transform()` was converting `undefined` (omitted query param) to `false`, making the env-variable fallback (`LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS`) unreachable. It also unifies the env comparison to `=== "true"` across all three public API routes, and adds regression tests for both the env-fallback path and the explicit `useEventsTable=false` override.

- **`useEventsTableSchema` fix**: preserves `undefined` when the param is omitted so the `??` operator in the route handlers can correctly fall back to the env flag.
- **Route handler consistency**: `[observationId].ts` and `observations/index.ts` previously returned the raw env string (e.g. `"false"` is truthy) as `useEventsTable`; they now compare with `=== "true"`, matching `traces/index.ts`.
- **Test coverage**: new `describe("useEventsTable env fallback")` blocks in all three servertest files verify both the env-driven path and the explicit-override path, with `afterEach` teardown restoring the env var.
</details>

The change is a targeted, well-scoped bug fix with no new surface area and no risk of regressions beyond the paths it explicitly corrects.

The transform fix and the ?? operator usage are logically sound; the env-string comparison is now consistently === 'true' across all three handlers; env mutation in tests is properly isolated via afterEach teardown; and the tests cover both the positive env-fallback path and the explicit-override path. No side-effects on unrelated routes were introduced.

No files require special attention — all changed paths are straightforward and consistent with each other.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(public-api): honor events table env ..."](https://github.com/langfuse/langfuse/commit/a2b188703ea80d3958fa06eb00dccc3500555c33) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33622552)</sub>

### DIFF
--- a/packages/shared/src/features/query/types.test.ts
+++ b/packages/shared/src/features/query/types.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from "vitest";
+import { useEventsTableSchema } from "./types";
+
+describe("useEventsTableSchema", () => {
+  it("should preserve undefined if query param is omitted", () => {
+    expect(useEventsTableSchema.parse(undefined)).toBeUndefined();
+  });
+
+  it("should parse true and false string values if query param is provided", () => {
+    expect(useEventsTableSchema.parse("true")).toBe(true);
+    expect(useEventsTableSchema.parse("false")).toBe(false);
+  });
+
+  it("should parse boolean values if query param is provided", () => {
+    expect(useEventsTableSchema.parse(true)).toBe(true);
+    expect(useEventsTableSchema.parse(false)).toBe(false);
+  });
+});
--- a/packages/shared/src/features/query/types.ts
+++ b/packages/shared/src/features/query/types.ts
@@ -195,4 +195,6 @@ export const query = z
 export const useEventsTableSchema = z
   .union([z.literal("true"), z.literal("false"), z.boolean()])
   .optional()
-  .transform((val) => val === "true" || val === true);
+  .transform((val) =>
+    val === undefined ? undefined : val === "true" || val === true,
+  );
--- a/web/src/__tests__/server/observation-api.servertest.ts
+++ b/web/src/__tests__/server/observation-api.servertest.ts
@@ -78,6 +78,13 @@ const insertObservations = async (
 };
 
 describe("/api/public/observations API Endpoint", () => {
+  const originalUseEventsTableEnv =
+    env.LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS;
+
+  afterEach(() => {
+    env.LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS = originalUseEventsTableEnv;
+  });
+
   // Test suite factory to run tests against both implementations
   const runTestSuite = (useEventsTable: boolean) => {
     const suiteName = useEventsTable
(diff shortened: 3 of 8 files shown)


martingarramon (commented):
LGTM. Preserving `undefined` lets the downstream env var fallback run instead of coercing an omitted query param to `false`. Test coverage looks complete.

sumerman:
@martingarramon thank you for you contribution. To be honest the idea of switching via ENV was ultimately abandoned, but it nevertheless good to have it in a functional state. Please use the established 
```
const maybe =
  env.LANGFUSE_ENABLE_EVENTS_TABLE_V2_APIS === "true"
    ? describe
    : describe.skip;
``` 
pattern to make sure the env tests are skipped in CI jobs that run on non-events setup.

fallintoplace:
@sumerman @martingarramon Let me check again later if any change is needed.

sumerman:
oh, sorry @fallintoplace 😅 thank you for your contribution as well! yes, a change is necessary for this PR to pass our CI

fallintoplace:
@sumerman GitHub Actions is down. I will catch up later.

sumerman:
@fallintoplace thank you for the changes, unfortunately the tests are still failing

fallintoplace:
@sumerman Let me check later today.

sumerman:
@fallintoplace we decided to just kill this env variable altogether. Thank you for your efforts.

PR #13817, opened 2026-05-24
fix(datasets): add stable orderBy id tiebreaker to listDatasets and listDatasetRuns

martingarramon:
## What does this PR do?

Adds `{ id: "asc" }` as a secondary sort key to four paginated `findMany` calls that were ordering by `createdAt: "desc"` alone — two in the MCP dataset tools and two in the public REST API.

## Why

When multiple rows share the same `createdAt` timestamp, PostgreSQL does not guarantee a stable order within that group. Paginated `OFFSET`/`LIMIT` queries against an unstable sort can return overlapping or skipped records across page fetches. Agents or scripts that iterate through dataset runs to compare experiments may miss a run or process the same one twice — especially when runs are created concurrently during parallel eval execution, which is exactly when same-millisecond timestamps occur.

Same class of issue as open PR #13813 (`listComments`). The annotation queue MCP tools added in the same PR (#13783) already use compound `createdAt, id` ordering; this brings datasets and dataset-run endpoints in line.

Fixes #13816.

**Affected files:**

- `web/src/features/mcp/features/datasets/tools.ts` — `listDatasets` and `listDatasetRuns` handlers
- `web/src/pages/api/public/datasets/index.ts` — `GET /api/public/datasets`
- `web/src/pages/api/public/datasets/[name]/runs/index.ts` — `GET /api/public/datasets/[name]/runs`

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code.

## Checklist

- [x] Added pagination-stability tests for `listDatasets` and `listDatasetRuns`: inserts 6 records with identical `createdAt` timestamps, then asserts that pages 1 and 2 produce no overlap, no gaps, and identical page 1 results on repeated calls.
- I haven't checked if new and existing unit tests pass locally with my changes

<details open><summary><h3>Greptile Summary</h3></summary>

This PR fixes a pagination-stability bug across four `findMany` calls in the dataset and dataset-run listing endpoints by adding `{ id: "asc" }` as a secondary sort key alongside the existing `createdAt: "desc"` ordering. Without this tiebreaker, rows sharing the same millisecond timestamp could be returned in an arbitrary order, causing OFFSET-based pagination to produce overlapping or missing records.

- **MCP tools** (`tools.ts`): `listDatasets` and `listDatasetRuns` handlers now use compound `[{ createdAt: "desc" }, { id: "asc" }]` ordering, consistent with other MCP tools added in the same feature area.
[...]

PR #13818, opened 2026-05-25
fix(datasets): stabilize pagination when createdAt ties

### LINKED ISSUE #13816: bug: listDatasets and listDatasetRuns (MCP + public API) return non-deterministic pages — missing id tiebreaker in orderBy
## Bug

Paginated dataset listing endpoints sort only by `createdAt: "desc"` in their Prisma `findMany` calls. When multiple rows share the same `createdAt`, PostgreSQL does not guarantee a stable order within that timestamp group, so `OFFSET`/`LIMIT` pagination can return overlapping or skipped records across page fetches.

Affected call sites on current `main`:

- `web/src/features/mcp/features/datasets/tools.ts` — `listDatasets`
- `web/src/features/mcp/features/datasets/tools.ts` — `listDatasetRuns`
- `web/src/pages/api/public/datasets/index.ts` — `GET /api/public/datasets`
- `web/src/pages/api/public/datasets/[name]/runs/index.ts` — `GET /api/public/datasets/[name]/runs`

Same class of pagination-stability issue addressed in open PR #13813 for `listComments`. The annotation queue MCP tools already use compound `createdAt, id` ordering; they use `id: "desc"`, while #13813 uses `id: "asc"`. The direction is not material for stability as long as the secondary field is unique.

## Impact

Agents or scripts that page through datasets or dataset runs can miss a record or process the same record twice when multiple rows have identical `createdAt` values. This is especially plausible for dataset runs created close together during parallel eval execution.

## Fix

Add a unique secondary sort key to the affected paginated queries, matching #13813:

```ts
// before
orderBy: { createdAt: "desc" }

// after
orderBy: [{ createdAt: "desc" }, { id: "asc" }]
```

peter941221:
Fix dataset and dataset run pagination so tied createdAt rows have a deterministic secondary order.

This updates all affected public dataset listing paths to sort by createdAt DESC and id ASC, including the v2 datasets endpoint.

The failure mode was offset pagination on createdAt alone. When multiple rows shared the same timestamp, the database could return tied rows in different orders across requests, which made repeated page fetches unstable and could cause overlap or skips between adjacent pages.

This change does not alter the pagination contract. It only makes the existing ordering total and deterministic.

Validation:
- targeted REST dataset pagination regression tests
- targeted REST dataset run pagination regression tests
- targeted v2 dataset pagination regression test
- targeted MCP dataset and dataset run pagination regression test

Closes #13816.

### DIFF
--- a/web/src/__tests__/server/datasets-api.servertest.ts
+++ b/web/src/__tests__/server/datasets-api.servertest.ts
@@ -1413,6 +1413,78 @@ describe("/api/public/datasets and /api/public/dataset-items API Endpoints", () 
     expect(mergedIds).toEqual(expect.arrayContaining(datasetIds));
   });
 
+  it("should paginate v2 datasets deterministically when createdAt timestamps tie", async () => {
+    const authAndProject = await createOrgProjectAndApiKey();
+    const localAuth = authAndProject.auth;
+    const localProjectId = authAndProject.projectId;
+    const sharedCreatedAt = new Date("2025-01-01T00:00:00.000Z");
+    const datasetIds = [v4(), v4(), v4()];
+
+    await prisma.dataset.createMany({
+      data: [
+        {
+          id: datasetIds[0],
+          name: `dataset-v2-tie-a-${v4()}`,
+          projectId: localProjectId,
+          createdAt: sharedCreatedAt,
+          updatedAt: sharedCreatedAt,
+        },
+        {
+          id: datasetIds[1],
+          name: `dataset-v2-tie-b-${v4()}`,
+          projectId: localProjectId,
+          createdAt: sharedCreatedAt,
+          updatedAt: sharedCreatedAt,
+        },
+        {
+          id: datasetIds[2],
+          name: `dataset-v2-tie-c-${v4()}`,
+          projectId: localProjectId,
+          createdAt: sharedCreatedAt,
+          updatedAt: sharedCreatedAt,
+        },
+      ],
+    });
+
+    const page1 = await makeZodVerifiedAPICall(
+      GetDatasetsV2Response,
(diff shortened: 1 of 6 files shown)


martingarramon (commented):
The fix is correct for the v1 REST and MCP paths, and the failure mechanism write-up is clear.

Two things worth flagging:

The v2 datasets endpoint (`web/src/pages/api/public/v2/datasets/index.ts`) appears to have the same `orderBy: { createdAt: "desc" }` without a tiebreaker. Greptile called this out; I verified it directly. Agents on the v2 path still get unstable pages under tied timestamps.

The new test IDs (`"dataset-tie-a"`, `"run-tie-a"`) are primary keys without `ON CONFLICT` handling. On a non-reset test DB a second run hits a unique-constraint violation. Using `v4()` for IDs (as the rest of the fixture does for names) prevents this — Greptile flagged this too.

The production fix is solid. These do not undermine it, but they seem worth addressing before merge.

I'm closing #13817 in favor of this PR.

nimarb:
@claude review

peter941221:
Addressed in 0bd16a4.

Added the same deterministic createdAt DESC, id ASC ordering to the v2 datasets endpoint and covered it with a matching regression test.

bezbac:
Hi @peter941221, thanks for taking the time to contribute. We merged some larger changes earlier in #13850, which caused this PR to run into conflicts. Since the fix itself was fairly small, I went ahead and addressed the underlying issue in #13856.

Thanks again for the effort, and hope you have a great week 🙌

peter941221:
@bezbac hi, no worry. I'll continue to contribute on other issues. Have a great week!

PR #13855, opened 2026-05-26
fix(api): Do not accept limit below 1 in pagination

### LINKED ISSUE #13846: bug: six MCP list tools accept limit=0 and negative limits — publicApiPaginationZod has no lower bound, producing null totalPages or reversed results
## Bug

Six MCP list tools accept `limit=0` and negative `limit` values because they inherit `publicApiPaginationZod`, which enforces an upper bound (`lte(100)`) but has no lower bound. This produces a malformed-pagination response on `limit=0` and passes invalid input through on negative values.

## Affected tools

| Tool | Inherited schema |
|---|---|
| `listAnnotationQueues` | `GetAnnotationQueuesQuery` |
| `listComments` | `GetCommentsV1Query` |
| `listDatasets` | `GetDatasetsV2Query` |
| `listDatasetRuns` | `GetDatasetsV2Query` |
| `listModels` | `GetModelsV1Query` |
| `listScores` | `ListScoresInputSchema` |

`listObservations` is not affected — it uses `ObservationLimitSchema` (`observations/schema.ts`), which correctly enforces `.int().min(1).max(100)`.

## Root cause

`packages/shared/src/utils/zod.ts`:
```ts
export const publicApiPaginationZod = {
  limit: z.preprocess(
    (x) => (x === "" ? undefined : x),
    z.coerce.number().lte(100).default(50), // no lower bound
  ),
};
```

`web/src/features/mcp/features/publicApi.ts` — `paginationMeta` has no guard:
```ts
totalPages: Math.ceil(totalItems / limit), // Infinity when limit=0
```

## Failure modes

**`limit=0`:** Zod accepts it. `take: 0` returns an empty result set. `JSON.stringify({ totalPages: Infinity })` serializes `Infinity` as JSON `null`, so the response includes `"totalPages": null` — a malformed meta object.

**`limit < 0`:** Zod accepts it. Prisma treats negative `take` as reverse pagination, so these forward-list tools silently return results in reversed order instead of rejecting the input.

## Fix options

**Option A — tighten `publicApiPaginationZod`** (`packages/shared/src/utils/zod.ts`):  
Add `.gte(1)` to `limit`. Fixes all six affected tools at once and also tightens shared REST API validation — maintainers should confirm this is an acceptable behavior change there.

**Option B — guard in `paginationMeta`** (`web/src/features/mcp/features/publicApi.ts`):  
Use `Math.max(limit, 1)` in the denominator. Prevents the `null totalPages` case but does not reject negative values; a separate guard on `limit` would still be needed.

Option A seems more complete given it closes both cases at the schema layer.

bezbac:
Resolves #13846

Adds a minimum bound of `1` to the `limit` field in `publicApiPaginationZod`, preventing API callers from passing `limit=0` or a negative value. The single-character diff (`.gte(1)`) is well-targeted and directly resolves the reported issue.

- `publicApiPaginationZod.limit` now enforces `1 ≤ limit ≤ 100` (was `limit ≤ 100` only).
- The companion `paginationZod.limit` (used for internal/UI pagination) still uses `.nonnegative()`, which permits `0`; aligning it to `.gte(1)` would be consistent.
</details>

Safe to merge; the change is a single-line guard on the public API pagination limit with no side effects on existing valid requests.

The fix correctly blocks limit=0 and negative values from the public API. The only outstanding question is whether the internal paginationZod should receive the same treatment, but that is a separate concern and does not affect correctness of this change.

packages/shared/src/utils/zod.ts — specifically whether paginationZod.limit should also be updated to .gte(1) for consistency.
</details>

```mermaid
flowchart TD
    A[API Request with limit param] --> B{limit === ''}
    B -- yes --> C[coerce undefined → default 50]
    B -- no --> D[z.coerce.number]
    D --> E{gte 1?}
    E -- no --> F[Validation Error returned to caller]
    E -- yes --> G{lte 100?}
    G -- no --> F
    G -- yes --> H[Valid limit passed to query]
    C --> H
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/utils/zod.ts:47-51
The non-public `paginationZod.limit` uses `.nonnegative()`, which still allows `limit=0`. If a `limit` of `0` reaches a database query it can produce empty or unexpected result sets, and the behaviour is inconsistent with the fix applied to `publicApiPaginationZod`.

```suggestion
  limit: z.preprocess(
    (x) => (x === "" ? undefined : x),
    z.coerce.number().gte(1).lte(100).default(50),
  ),
};
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(api): Do not accept limit below 1 in..."](https://github.com/langfuse/langfuse/commit/ab69ed53a5c30cc455b82959cb21910816cfac6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33931694)</sub>

### DIFF
--- a/packages/shared/src/utils/zod.ts
+++ b/packages/shared/src/utils/zod.ts
@@ -44,6 +44,7 @@ export const paginationZod = {
     (x) => (x === "" ? undefined : x),
     z.coerce.number().nonnegative().default(1),
   ),
+  // For internal use, we allow a limit of 0 until LFE-9976 is merged
   limit: z.preprocess(
     (x) => (x === "" ? undefined : x),
     z.coerce.number().nonnegative().lte(100).default(50),
@@ -57,7 +58,7 @@ export const publicApiPaginationZod = {
   ),
   limit: z.preprocess(
     (x) => (x === "" ? undefined : x),
-    z.coerce.number().lte(100).default(50),
+    z.coerce.number().gte(1).lte(100).default(50),
   ),
 };
 


martingarramon (approved):
LGTM — `.gte(1)` is the right fix, and applying it to both `paginationZod` and `publicApiPaginationZod` keeps them consistent; `paginationZod` had `.nonnegative()` which also permitted `0`. Minor: a regression test for `limit=0` returning 4xx would be nice-to-have, but not blocking given the simplicity of the constraint.

sumerman (approved):

PR #13863, opened 2026-05-26
feat: matches string filters that always uses TEXT index

sumerman:
## Summary
- Adds `matches` as a public v2 observations operator for `events_*` input, output, and metadata filters.
- Centralizes FTS validation and SQL emission in shared ClickHouse helpers.
- Keeps v1 behavior unchanged while enforcing accelerated IO filters on the v2 events path.
- Updates Fern and generated OpenAPI so the contract matches the implementation.

## Testing
- Targeted server tests for observations v2, FTS rewrites, and filter validation passed.
- Shared typecheck, targeted ESLint, and diff whitespace checks passed.
- Added coverage for v1 rejection, IO acceleration rules, and metadata token-matching edge cases.

This PR adds a `matches` operator for `input`, `output`, and `metadata` string filters on the v2 observations API, routing them through ClickHouse text indexes (`hasAllTokens`). It centralises FTS SQL emission into `FTS_OPERATOR_DESCRIPTORS`, adds `assertValidFtsMatchFilter` to reject invalid targets/tokenless values, and enforces that any v2 IO filter request must include at least one indexed (`=` or `matches`) IO predicate.

- `eventsTableSingleFilter` / `eventsTableStringFilter` / `eventsTableStringObjectFilter` extend the existing filter schemas with the new `matches` operator, while keeping v1 using the unmodified `singleFilter`.
- `validateIndexedInputOutputFilters` is introduced on the v2 code path to reject requests that would trigger full-table IO scans without an accompanying indexed predicate.
- Generated OpenAPI/Fern definitions are updated to document the new operator and its semantics (case-insensitive for IO, case-sensitive for metadata).
</details>

The v2 IO-filter gate incorrectly blocks null filters on input/output columns; v1 behavior is unaffected.

The `isInputOutputFilter` helper catches any filter on an input/output field including null filters, so a standalone `input is null` v2 request passes Zod but is rejected with a misleading error. No test covers this scenario.

packages/shared/src/server/repositories/events.ts — the `isInputOutputFilter` helper and `validateIndexedInputOutputFilters` function around lines 927–943.
</details>

```mermaid
sequenceDiagram
    participant Client
    participant ZodValidation as Zod (eventsTableSingleFilter)
    participant Factory as createFilterFromFilterState
    participant FtsValidator as assertValidFtsMatchFilter
    participant IoValidator as validateIndexedInputOutputFilters
    participant ClickHouse

    Client->>ZodValidation: "POST /v2/observations?filter=[{operator:matches,...}]"
    ZodValidation-->>Client: 400 if operator not in eventsTableSingleFilter
    ZodValidation->>Factory: validated EventsTableFilterState
    Factory->>FtsValidator: validateEventsTableMatchesFilter (token/target check)
    FtsValidator-->>Client: 400 InvalidRequestError if invalid target or tokenless value
    Factory-->>Factory: build StringFilter / StringObjectFilter
    Factory->>IoValidator: validateIndexedInputOutputFilters(FilterList)
    IoValidator-->>Client: "400 if IO filter present with no = or matches sibling"
    IoValidator->>ClickHouse: hasAllTokens(lower(field), lower(param)) SQL
    ClickHouse-->>Client: 200 matching observations
```
</details>

`````markdown

---

### Issue 1 of 2
packages/shared/src/server/repositories/events.ts:927-943
**Null filters on IO columns incorrectly rejected in v2**

`isInputOutputFilter` returns `true` for any filter whose field resolves to `input` or `output`, including `NullFilter` (operator `"is null"` / `"is not null"`). Because `isFtsAcceleratedIoOperator("is null")` returns `false`, a request like `{ type: "null", column: "input", operator: "is null" }` will pass Zod validation against `eventsTableSingleFilter` (which includes `nullFilter` unchanged), be converted to a `NullFilter`, and then be rejected with `"Input/output filters require at least one \`matches\` or \`=\` operator."` — a misleading 400 that blocks a semantically valid query.

The fix is to narrow `isInputOutputFilter` to only string-operator filters, e.g.:

`const isInputOutputFilter = (filter: Filter): boolean => isFtsTextField(filter.field) && isFtsAcceleratedIoOperator(filter.operator);`

With this form the function returns `true` only for filters that are themselves accelerated IO filters, which can be used directly as the guard without a second `.some` pass.

### Issue 2 of 2
packages/shared/src/server/queries/clickhouse-sql/clickhouse-filter.ts:164-178
**`assertValidFtsMatchFilter` called twice for `matches` filters on the factory path**

`createFilterFromFilterState` in `factory.ts` already calls `validateEventsTableMatchesFilter` → `assertValidFtsMatchFilter` before constructing the `StringFilter`. When the caller then invokes `apply()`, `assertValidFtsMatchFilter` runs again with identical arguments. This is harmless on the happy path, but any future change to validation semantics must be kept in sync across both sites. Consider removing the call from `apply()` (relying on the factory-level guard) or removing it from the factory (relying on `apply()`).

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: refactor and consolidate a bit"](https://github.com/langfuse/langfuse/commit/8c87a26d44c1532e8eb7c35fbfa0396206e49396) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34023525)</sub>

### DIFF
--- a/web/src/features/mcp/features/observations/tools/listObservations.ts
+++ b/web/src/features/mcp/features/observations/tools/listObservations.ts
@@ -325,30 +325,35 @@ export const [listObservationsTool, handleListObservations] = defineTool({
           "mcp.field_groups": fieldGroups.join(","),
         });
 
-        const items = await getObservationsV2FromEventsTableForPublicApi({
-          projectId: context.projectId,
-          page: 0,
-          limit: input.limit,
-          traceId: input.traceId,
-          userId: input.userId,
-          level: input.level,
-          name: input.name,
-          type: input.type,
-          environment: input.environment,
-          parentObservationId: input.parentObservationId,
-          fromStartTime: input.fromStartTime,
-          toStartTime: input.toStartTime,
-          version: input.version,
-          advancedFilters: input.filter,
-          cursor: input.cursor
-            ? EncodedObservationsCursorV2.parse(input.cursor)
-            : undefined,
-          fields: fieldGroups,
-          expandMetadataKeys: getMetadataExpansionForProjection(
-            projectionFields,
-            input.expandMetadataKeys,
-          ),
-        });
+        const items = await getObservationsV2FromEventsTableForPublicApi(
+          {
+            projectId: context.projectId,
+            page: 0,
+            limit: input.limit,
+            traceId: input.traceId,
+            userId: input.userId,
+            level: input.level,
+            name: input.name,
+            type: input.type,
+            environment: input.environment,
+            parentObservationId: input.parentObservationId,
+            fromStartTime: input.fromStartTime,
+            toStartTime: input.toStartTime,
+            version: input.version,
+            advancedFilters: input.filter,
+            cursor: input.cursor
+              ? EncodedObservationsCursorV2.parse(input.cursor)
+              : undefined,
+            fields: fieldGroups,
+            expandMetadataKeys: getMetadataExpansionForProjection(
+              projectionFields,
+              input.expandMetadataKeys,
+            ),
+          },
+          // MCP keeps the legacy observation filter contract. Its separate
+          // selective-scope guard above limits expensive input/output filters.
+          { allowUnindexedIoFilters: true },
+        );
 
         const hasMore = items.length > input.limit;
         const dataToReturn = hasMore ? items.slice(0, input.limit) : items;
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: web/src/features/mcp/features/observations/tools/listObservations.ts [hunk 1]
```diff
@@ -325,30 +325,35 @@ export const [listObservationsTool, handleListObservations] = defineTool({
...
+              : undefined,
+            fields: fieldGroups,
+            expandMetadataKeys: getMetadataExpansionForProjection(
+              projectionFields,
+              input.expandMetadataKeys,
+            ),
+          },
+          // MCP keeps the legacy observation filter contract. Its separate
+          // selective-scope guard above limits expensive input/output filters.
+          { allowUnindexedIoFilters: true },
```
sumerman, line 355: @bezbac I am allowing an exception here. But would appreciate if you adopt `matches` once this PR lands.

PR #13866, opened 2026-05-27
fix(api): return 503 for public auth db failures

### LINKED ISSUE #13482: bug: API key validation returns 401 (not 5xx) when database is unreachable

peter941221:
Fixes the auth error classification in createAuthedProjectAPIRoute.

Failure mechanism:
When ApiAuthService rethrows a Prisma exception during API-key lookup, the route-level auth catch previously defaulted missing error.status to 401. That turns a transient database outage into an authentication failure.

Semantic change:
Prisma-backed auth failures now return 503 Service Unavailable from createAuthedProjectAPIRoute, while real invalid-credential failures still return 401.

Preserved invariant:
Only the auth error classification changes. Successful auth flow, access-level checks, and non-Prisma auth failures keep their existing behavior.

Testing:
- pnpm --filter web exec vitest run --project server-unit src/__tests__/server/unit/create-authed-project-api-route-auth-errors.servertest.ts
- pnpm --filter web exec eslint src/__tests__/server/unit/create-authed-project-api-route-auth-errors.servertest.ts
- pnpm --filter web exec eslint src/features/public-api/server/createAuthedProjectAPIRoute.ts

Issue:
- closes #13482

This PR fixes a bug where Prisma exceptions bubbling up from the API-key lookup during authentication were mistakenly surfaced as `401 Unauthorized` — because the old code fell back to `error.status || 401` on any thrown error, and Prisma errors carry no `.status` property. The fix adds an `isPrismaException` guard that maps those errors to `503 Service Unavailable` instead.

- **`createAuthedProjectAPIRoute.ts`**: Two new lines replace `error.status || 401` / `error.message || "..."` with an `isPrismaException` branch that returns 503 for Prisma errors and preserves the original `error.status ?? 401` / `error.message ?? "..."` fallback for all other thrown errors.
- **New test file**: Adds two unit tests covering the `401` (invalid credentials) and `503` (Prisma exception) paths, using `vi.hoisted` mocks for `isPrismaException` and `verifyAuthHeaderAndReturnScope`.
</details>

The change is narrowly scoped to the auth error catch block and is safe to merge; no successful auth path is modified.

The fix is correct and well-targeted. The double `isPrismaException` call is a minor readability concern, and `PrismaClientValidationError` being lumped in with connectivity errors is a semantic mismatch worth watching, but neither represents a defect in the happy path or in the primary 503 fix. The `unstablePublicEvalsErrorContract` branch is untested but was also untested before this PR.

`web/src/utils/exceptions.ts` — the inclusion of `PrismaClientValidationError` in the 503-triggering set is worth a second look since it covers query-construction bugs, not just connectivity failures.
</details>

```mermaid
flowchart TD
    A[Incoming request] --> B[verifyAuth]
    B --> C{Exception thrown?}
    C -->|No| D[Rate limit check]
    D --> E[Run handler]
    E --> F[200 response]
    C -->|Yes| G{isPrismaException?}
    G -->|Yes| H[503 Service Unavailable]
    G -->|No| I[error.status or 401]
    I --> J[401 / 403 / 404 response]
```
</details>

`````markdown

---

### Issue 1 of 1
web/src/__tests__/server/unit/create-authed-project-api-route-auth-errors.servertest.ts:110-128
The test suite does not cover the `unstablePublicEvalsErrorContract` code path. When a Prisma exception occurs and `routeConfig.errorContract === unstablePublicEvalsErrorContract`, the fix routes through `createUnstablePublicApiAuthError` / `sendUnstablePublicApiErrorResponse` with the new 503 status. Without a test for this branch, a regression there (e.g., the error contract serializing the status code differently) would go undetected.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(api): return 503 for public auth db ..."](https://github.com/langfuse/langfuse/commit/5a4627e129dfcb966fc878692ba78a2c96c31487) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34082214)</sub>


peter941221:
Ready for review now.

The scope here is just the auth error classification: invalid credentials stay 401, while Prisma-backed auth failures surface as retryable 503 responses. The focused regression test is included and current checks are green.

If you want broader handling for other infrastructure failure classes, I can follow up separately to keep this change narrow.

Steffen911:
@peter941221 Thank you for the contribution and for the addition of test-cases. I like the explicit if/else approach from https://github.com/langfuse/langfuse/pull/13494 slightly better, but we're missing exception tracing there. 
Could you patch this in your PR? I'd be happy to merge this once it's a bit cleaner and traces the exception on a span level correctly.

peter941221:
@Steffen911 Sure, I will do that now.

peter941221:
Patched.

I rewrote the auth error handling into an explicit Prisma-first branch and now trace Prisma-backed auth failures before returning the 503 response.

I also extended the unit test so it covers both response shapes:
- default JSON auth error response
- unstable public evals error contract

Verified with:
pnpm --filter web exec vitest run --project server-unit src/__tests__/server/unit/create-authed-project-api-route-auth-errors.servertest.ts

That test file passes with 3/3 tests on a clean WSL clone.

PR #13887, opened 2026-05-27
fix(evals): surface invalid code eval results

wochinge:
## Summary

- Preserve small invalid code evaluator return values as structured `returnedResult` data so users can debug malformed evaluator outputs from test runs.
- Propagate the returned result through shared dispatcher errors, user-visible eval errors, and internal trace output while keeping status messages and metadata focused on the public error message.
- Derive the test-run error response from the shared code eval error type and strip `retryable` at the web boundary, so future error fields do not need to be copied manually.
- Add regression coverage across local dispatch, Lambda dispatch, worker execution traces, and the web test-run endpoint.

## Major Decisions

- Keep `returnedResult` as structured data instead of appending it to the error message. This makes the value easier to render in the UI and avoids bloating trace status messages or metadata fields.
- Keep `retryable` internal to execution handling for test runs. The web response only exposes the debugging fields users need for fixing evaluator code.

## Review Focus

- Confirm the returned evaluator payload is only exposed for user-visible invalid-result paths and does not leak into internal infrastructure errors.
- Check the trace output shape for invalid evaluator results, especially that metadata and status messages stay concise.
- Review the test-run API response shape because it now includes optional structured `returnedResult` data on failed test runs.

This PR surfaces the raw evaluator return value when a code eval produces an invalid shape, giving users actionable debugging context in the test-run API instead of a generic "validation failed" message. It also wires `returnedResult` through the internal trace `output` field while deliberately keeping it out of string fields (`statusMessage`, metadata), and strips `retryable` at the API boundary.

- **`CodeEvalDispatcherError`** gains `returnedResult?: unknown`, populated by `parseDispatchResult` when Zod rejects the evaluator output. Lambda path is safe: `assertDispatchResultWithinByteLimit` runs before parsing, bounding the captured value to ≤256 KiB. The local dispatcher skips that guard, so a VM result that fails schema validation but is non-JSON-serializable (e.g., `BigInt`, circular reference) would propagate to the tRPC response without a serialization guard.
- **`codeEvalTestRun.ts`** derives `CodeEvalTestRunError` from `Omit<CodeEvalUserVisibleError, \"retryable\">` and adds `toCodeEvalTestRunError` to cleanly drop `retryable` at the boundary while preserving `returnedResult`.
- **Test coverage** spans the local dispatcher, Lambda dispatcher (valid-JSON invalid-schema and non-JSON cases), the worker-level trace assertion, and a tRPC integration test for the end-to-end error response shape.
</details>

Safe to merge; changes are scoped to the error-propagation path for invalid eval results, with no changes to the happy path or score persistence logic.

The chaining of returnedResult through the error → trace → API path is correct and well-tested. The one gap is that the local (development-only) dispatcher stores the raw VM result without a serializability guard, so a non-JSON-safe evaluator return (BigInt, circular ref) would cause tRPC serialization to throw rather than returning a structured error response. This only affects the insecure-local dev dispatcher, not production Lambda flows.

packages/shared/src/server/evals/codeEvalDispatcherTypes.ts — specifically the parseDispatchResult function where returnedResult is captured without a serializability check for the local dispatcher path.
</details>

```mermaid
sequenceDiagram
    participant Evaluator as User Evaluator Code
    participant Dispatcher as CodeEvalDispatcher(Local/Lambda)
    participant Parse as parseDispatchResult
    participant Exec as runCodeBasedEvaluationDispatch
    participant Trace as InternalTraceWriter
    participant API as codeEvalTestRun API

    Evaluator->>Dispatcher: "return { score: 1 }"
    Note over Dispatcher: Lambda only: assertDispatchResultWithinByteLimit()
    Dispatcher->>Parse: "parseDispatchResult({ score: 1 })"
    Parse-->>Dispatcher: "throw CodeEvalDispatcherError { code: INVALID_RESULT, returnedResult: { score: 1 } }"
    Dispatcher-->>Exec: throw (re-thrown)
    Exec->>Exec: getCodeEvalUserVisibleError(error) → visibleError includes returnedResult
    Exec->>Exec: "build traceError { code, message, retryable, returnedResult }"
    Exec->>Trace: "writeCodeEvalTrace(output: { error: traceError })"
    Note over Exec,Trace: returnedResult in JSON output only, not in statusMessage or metadata strings
    Exec-->>API: "{ success: false, error: { code, message, returnedResult } }"
    Note over API: toCodeEvalTestRunError strips retryable before returning to caller
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/evals/codeEvalDispatcherTypes.ts:205-219
**Unguarded non-serializable `returnedResult` on local dispatcher path**

`parseDispatchResult` stores `result` verbatim as `returnedResult`. For the Lambda dispatcher `assertDispatchResultWithinByteLimit` runs before this call, so by the time parsing is attempted the value is bounded JSON. The local dispatcher, however, calls `parseDispatchResult` directly with the raw VM result, applying no size or serializability guard. If a user's evaluator returns a non-JSON-serializable value — `BigInt`, a VM-realm `Date`, or a circular reference — `returnedResult` carries that value all the way to the tRPC test-run response, where plain `JSON.stringify` will throw and convert a validation failure into an unexpected 500. The trace write is safe (`writeCodeEvalTraceSafely` swallows errors), but the API response is not.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(evals): surface invalid code eval re..."](https://github.com/langfuse/langfuse/commit/d1ae09ff329e67f674d4915f73cf2c66db4a3ed3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34188098)</sub>

### DIFF
--- a/packages/shared/src/server/evals/codeEvalDispatcherTypes.ts
+++ b/packages/shared/src/server/evals/codeEvalDispatcherTypes.ts
@@ -143,6 +143,7 @@ export const CodeEvalDispatcherErrorCodes = CodeEvalDispatcherErrorCode.enum;
 export class CodeEvalDispatcherError extends Error {
   public readonly code: CodeEvalDispatcherErrorCode;
   public readonly retryable: boolean;
+  public readonly returnedResult?: unknown;
 
   constructor(
     message: string,
@@ -150,12 +151,14 @@ export class CodeEvalDispatcherError extends Error {
       code: CodeEvalDispatcherErrorCode;
       retryable?: boolean;
       cause?: unknown;
+      returnedResult?: unknown;
     },
   ) {
     super(message, { cause: options.cause });
     this.name = "CodeEvalDispatcherError";
     this.code = options.code;
     this.retryable = options.retryable ?? false;
+    this.returnedResult = options.returnedResult;
   }
 }
 
@@ -205,7 +208,10 @@ export function parseDispatchResult(result: unknown): DispatchResult {
   if (!parsed.success) {
     throw new CodeEvalDispatcherError(
       `Invalid code eval result: ${parsed.error.message}`,
-      { code: CodeEvalDispatcherErrorCodes.INVALID_RESULT },
+      {
+        code: CodeEvalDispatcherErrorCodes.INVALID_RESULT,
+        returnedResult: result,
+      },
     );
   }
 
--- a/packages/shared/src/server/evals/codeEvalExecution.ts
+++ b/packages/shared/src/server/evals/codeEvalExecution.ts
@@ -72,6 +72,7 @@ export type CodeEvalUserVisibleError = {
   code: CodeEvalUserVisibleErrorCode;
   message: string;
   retryable: boolean;
+  returnedResult?: unknown;
 };
 
 export class CodeEvalExecutionError extends Error {
@@ -129,6 +130,7 @@ function buildCodeEvalPayload(params: {
 function getCodeEvalErrorDetails(error: unknown): {
   name: string;
   message: string;
+  returnedResult?: unknown;
   code: CodeEvalUserVisibleErrorCode;
   retryable: boolean;
 } {
@@ -136,6 +138,7 @@ function getCodeEvalErrorDetails(error: unknown): {
     return {
       name: error.name,
       message: error.message,
+      returnedResult: error.returnedResult,
       code: error.code,
       retryable: error.retryable,
     };
@@ -175,6 +178,9 @@ export function getCodeEvalUserVisibleError(
     code: details.code,
     message: message ?? details.message,
     retryable: details.retryable,
+    ...(details.returnedResult !== undefined
+      ? { returnedResult: details.returnedResult }
+      : {}),
   };
 }
 
@@ -240,6 +246,9 @@ export async function runCodeBasedEvaluationDispatch(params: {
       code: visibleError.code,
       message: visibleError.message,
       retryable: errorDetails.retryable,
+      ...(visibleError.returnedResult !== undefined
+        ? { returnedResult: visibleError.returnedResult }
+        : {}),
     };
     const errorCodeForTrace = errorDetails.code;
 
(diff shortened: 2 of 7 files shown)


Quantum0uasar:
The approach here is exactly right — preserving malformed evaluator outputs as structured `returnedResult` data rather than silently swallowing them is critical for debuggability. A few notes:

## Propagation through dispatcher chain

Propagating `returnedResult` through the shared dispatcher errors is the correct design: the evaluator pipeline is complex (local dispatch, Lambda, worker nodes), and losing the raw returned value at any intermediate layer makes it impossible to diagnose whether the malfunction is in the evaluator logic, the marshalling layer, or the payload schema.

## Stripping `retryable` at the web boundary

Good call removing `retryable` from the web-facing response. It's an internal scheduler concern, not something end users should see or act on. The pattern of deriving error type at the boundary and stripping internal fields is clean.

## Observability consideration

Beyond surfacing the result in the UI, it may be worth logging the `returnedResult` at `WARN` or `ERROR` level on the server side as well (with trace/span context). Some users monitor logs rather than the UI, and a structured log line with the eval job ID + bad payload would make these failures searchable in tools like Datadog, Grafana Loki, etc.

## Edge case: very large `returnedResult`

If an evaluator returns a large object (e.g., accidentally returns the full LLM response instead of a score), storing it as `returnedResult` could be expensive. It might be worth truncating at a reasonable limit (e.g., 10KB) with a `[truncated]` marker before persisting, to prevent unbounded DB writes.

## Regression test coverage

Glad to see regression coverage across local dispatch, Lambda, and worker nodes. The key cases to ensure are covered:
1. Evaluator returns a non-numeric value where a score is expected
2. Evaluator returns valid JSON but wrong schema
3. Evaluator raises an exception (vs. returning a bad value)
4. `returnedResult` is `null` vs. missing entirely (both should be handled gracefully in the UI)

Solid fix for a class of issues that's notoriously hard to debug without this kind of structured error surfacing.

PR #13921, opened 2026-05-28
fix(api): honor useEventsTable env fallback

### LINKED ISSUE #13814: Public API `useEventsTable` fallback makes `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS` ineffective when the query param is omitted

Epochex:
Fixes #13814.

The shared `useEventsTableSchema` now preserves the omitted state (`undefined`) instead of coercing it to `false`. Public traces/observations routes use nullish coalescing to fall back to `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS === "true"` when the query param is not provided.

This also fixes the observations routes which previously fell back to the raw env string, which is truthy even when set to `"false"`.

Tests:
- `node node_modules\.pnpm\vitest@4.1.4_@opentelemetry_30e46a091a16bf4f41d45bc501604489\node_modules\vitest\vitest.mjs run packages/shared/src/features/query/types.test.ts`
- `pnpm --filter web exec vitest run --project in-source src/features/public-api/types/useEventsTableQueryParsing.test.ts`

### DIFF
--- a/packages/shared/src/features/query/types.test.ts
+++ b/packages/shared/src/features/query/types.test.ts
@@ -1,0 +1,19 @@
+import { describe, it, expect } from "vitest";
+import { z } from "zod";
+
+import { useEventsTableSchema } from "./types";
+
+describe("useEventsTableSchema", () => {
+  it("preserves undefined when query param is omitted", () => {
+    const schema = z.object({ useEventsTable: useEventsTableSchema });
+    const parsed = schema.parse({});
+    expect(parsed.useEventsTable).toBeUndefined();
+  });
+
+  it("parses boolean-like values", () => {
+    expect(useEventsTableSchema.parse("true")).toBe(true);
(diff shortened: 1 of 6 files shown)


Steffen911:
This is an internal setting that is not meant to be consumed by self-hosters yet. Therefore, closing this. We'll update https://github.com/orgs/langfuse/discussions/12518 once users can opt-in to the v4 behaviours for self-hosted instances.

PR #13923, opened 2026-05-28
fix(mcp): compat for allOf/anyOf in scoreConfig tools

nimarb:

This PR replaces a simple `type: \"object\"` injection with a proper schema normalization pass that flattens root-level `allOf`/`anyOf`/`oneOf` (produced by Zod's `.and()` intersections on score-config tools) into a plain object schema that MCP model validators accept.

- `input-schema-compat.ts` is introduced to recursively collect and merge `properties`/`required` from composition-keyword schemas, preserving `definitions` and other metadata through `copyRootMetadata`.
- A regression test is added to `mcp-public-api-tools.servertest.ts` that asserts no registered tool exposes root-level composition keywords or unresolved `$ref` pointers after normalization.
</details>

Safe to merge — the normalization logic correctly handles Zod's allOf/anyOf output, definitions are preserved, and a targeted regression test guards against regressions.

The core path (Zod `.and()` → `allOf` → normalized flat object schema) is sound and well-tested. Three edge cases are worth watching: the union `required` intersection collapses to empty if any `anyOf` variant has no required fields; `additionalProperties: false` is dropped from normalized schemas; and `anyOf` is silently ignored when `oneOf` is also present. None are reachable with the current score-config Zod schemas.

web/src/features/mcp/core/input-schema-compat.ts — the mergeShapes union path and DROPPED_ROOT_KEYS set are the areas most likely to need revisiting if new tools introduce anyOf/oneOf or rely on additionalProperties constraints.
</details>

```mermaid
flowchart TD
    A[Zod schema with .and / intersection] -->|z.toJSONSchema| B{Root schema shape?}
    B -->|type:object, no composition keywords| C[Return as-is fast path]
    B -->|allOf at root| D[collectObjectShape allOf
mergeShapes intersection]
    B -->|anyOf or oneOf at root| E[collectObjectShape anyOf/oneOf
mergeShapes union]
    D --> F[Union all required fields
Merge all properties]
    E --> G[Intersect required fields
Merge all properties]
    F --> H[copyRootMetadata
preserves definitions etc]
    G --> H
    H --> I[Output: type:object
properties, required, metadata]
    I --> J[MCP ToolDefinition
inputSchema]
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/features/mcp/core/input-schema-compat.ts:56-62
**Union `required` intersection depends on first shape ordering**

The `union` branch seeds the intersection with `validShapes[0]?.required` and then cross-filters against every other shape. Mathematically this is correct — any field in the true intersection must appear in every shape, including shape 0 — but the result silently drops all required fields if `validShapes[0]` is an empty shape (e.g. a bare `z.object({})` variant in `anyOf`). If any Zod `anyOf` variant has no required fields, the entire normalized schema will have `required: []`, making every field optional. Worth guarding against or documenting as a known trade-off.

### Issue 2 of 3
web/src/features/mcp/core/input-schema-compat.ts:9-15
**`additionalProperties` is silently stripped from composition-keyword schemas**

`DROPPED_ROOT_KEYS` discards `additionalProperties` when the schema contains `allOf`/`anyOf`/`oneOf` at root. For the subset of Zod schemas that emit `additionalProperties: false` alongside a composition keyword, the normalized output omits that constraint. Zod's runtime validation still rejects extra fields, but LLM clients reading the schema have no signal to avoid sending them, which will produce tool-call errors at runtime. If the normalized schema should remain strict, re-apply `additionalProperties: false` unconditionally in `normalizeMcpInputSchema`.

### Issue 3 of 3
web/src/features/mcp/core/input-schema-compat.ts:65-70
**`oneOf` and `anyOf` are mutually exclusive; only the first present wins**

`const oneOfOrAnyOf = asArray(schema.oneOf) ?? asArray(schema.anyOf)` means that if a schema has both `oneOf` and `anyOf` at the root, `anyOf` is silently ignored. While Zod's current JSON Schema output is unlikely to emit both at the same level, the function makes no assumption explicit. A comment here clarifying the priority would prevent surprises if the Zod serializer behaviour ever changes.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["more simp"](https://github.com/langfuse/langfuse/commit/7d68ce02c0c974025e25da9bd5d1a54de85d4480) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34476250)</sub>

### DIFF
--- a/web/src/__tests__/server/mcp-public-api-tools.servertest.ts
+++ b/web/src/__tests__/server/mcp-public-api-tools.servertest.ts
@@ -89,6 +89,20 @@ import { handleGetScoreConfig } from "@/src/features/mcp/features/scores/tools/g
 import { handleListScoreConfigs } from "@/src/features/mcp/features/scores/tools/listScoreConfigs";
 import { handleUpdateScoreConfig } from "@/src/features/mcp/features/scores/tools/updateScoreConfig";
 
+const TOP_LEVEL_SCHEMA_COMPOSITION_KEYWORDS = [
+  "oneOf",
+  "anyOf",
+  "allOf",
+] as const;
+
+const getTopLevelSchemaCompositionKeywords = (
+  schema: Record<string, unknown>,
+) =>
+  TOP_LEVEL_SCHEMA_COMPOSITION_KEYWORDS.filter((keyword) => keyword in schema);
+
+const referencesDefinitions = (schema: Record<string, unknown>) =>
+  JSON.stringify(schema).includes("#/definitions/");
+
 const createScoreConfig = async (projectId: string) =>
   prisma.scoreConfig.create({
     data: {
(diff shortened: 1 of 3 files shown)


bezbac:
Superseded by https://github.com/langfuse/langfuse/pull/13927

PR #13932, opened 2026-05-29
fix(public-api): respect events-table env flag when query param omitted

### LINKED ISSUE #13814: Public API `useEventsTable` fallback makes `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS` ineffective when the query param is omitted

ashishpatel26:
## Problem

The public traces/observations APIs are designed to fall back to the `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS` env flag when the `useEventsTable` query param is omitted:

- `?useEventsTable=true` → force events table
- `?useEventsTable=false` → force legacy path
- omitted → use the env flag

But `useEventsTableSchema` (`packages/shared/src/features/query/types.ts`) coerced an **omitted** param to `false` via its `.transform`:

```ts
.optional()
.transform((val) => val === "true" || val === true) // undefined -> false
```

Route handlers then guard with `query.useEventsTable !== undefined`, which is never true for the omitted case — so `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS=true` was effectively ignored unless the client explicitly sent `?useEventsTable=true`.

Additionally, the two observation routes compared the env value (`"true" | "false"` enum) as a raw truthy string, so even when the fallback was reached the string `"false"` evaluated truthy — always selecting the events table. The traces route already compared `=== "true"`.

## Fix

- Preserve `undefined` in the schema transform so an omitted param stays absent.
- Use `??` in the handlers to fall back to the env flag.
- Compare the env flag against `"true"` explicitly in the observation routes, matching the traces route.
- Add a unit test for the schema covering `"true"` / `"false"` / boolean / `undefined`.

## Impacted packages

- `packages/shared`
- `web`

## Verification

- `pnpm --filter @langfuse/shared run lint` + `typecheck`
- `pnpm --filter web run lint` + `typecheck`
- New unit test: `packages/shared/src/features/query/types.test.ts` (5 passed)

Fixes #13814

This PR fixes a bug where `LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS` was silently ignored on the public traces/observations API routes when `useEventsTable` was omitted from the query string, and also fixes the observation routes comparing the env value as a raw truthy string instead of `=== "true"`.

- **Schema fix (`types.ts`)**: The `useEventsTableSchema` transform now preserves `undefined` for an omitted param, so route handlers can distinguish "omitted" from "explicit false" and fall through to the env flag via `??`.
- **Handler fix (3 routes)**: All three routes now use `query.useEventsTable ?? (env.LANGFUSE_ENABLE_EVENTS_TABLE_OBSERVATIONS === "true")`, correctly using `??` (not `||`) so an explicit `false` still opts out of the events table.
- **Tests**: A new unit test in `types.test.ts` covers all four input variants plus the `undefined` regression case.
</details>

The change is safe to merge — it corrects two silent mis-routings without touching any data-write paths or authentication logic.

All three handler changes are consistent and mechanically straightforward: the schema now passes undefined through instead of coercing it to false, and the ?? operator correctly short-circuits on false (explicit opt-out) while falling back to the env flag on undefined (omitted). The targeted unit tests directly cover the regression scenario. No broader logic is affected.

No files require special attention; all changed files apply the same two-line pattern uniformly.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(public-api): respect events-table en..."](https://github.com/langfuse/langfuse/commit/582276e1a5a954318de3316b125e24dab3952904) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34552621)</sub>


ashishpatel26:
@claude review

ashishpatel26:
@claude review

Steffen911:
This is an internal setting that is not meant to be consumed by self-hosters yet. Therefore, closing this. We'll update https://github.com/orgs/langfuse/discussions/12518 once users can opt-in to the v4 behaviours for self-hosted instances.

PR #13935, opened 2026-05-29
perf(blob-export): replace observations FINAL with LIMIT 1 BY

niklassemmler:
## Summary

Fixes three production OOM incidents ([LFE-10031](https://linear.app/langfuse/issue/LFE-10031)) traced to `MEMORY_LIMIT_EXCEEDED` on `FROM observations FINAL` in the v3 blob storage export.

`FINAL` forces a k-way merge-sort across all parts before the time-range `WHERE` can be applied — measured: **75 GiB read for 1.08M rows in ~9 minutes** on a single 1-hour export window. The replacement `ORDER BY event_ts DESC / LIMIT 1 BY` subquery reads the same window in **0.5 s, reading 1.3 MB**.

### Why `is_deleted = 0` is in the outer `WHERE`

Placing it inside the inner `WHERE` would cause `LIMIT 1 BY` to pick the last *non-deleted* version of a row whose latest version is actually `is_deleted = 1`, resurrecting soft-deleted observations. The outer placement is enforced by code structure and covered by the new regression test.

### Scope

- Only `getObservationsForBlobStorageExport` (v3 path). The v4 path (`getEventsForBlobStorageExport`) already uses `LIMIT 1 BY` and is untouched.
- Other `FROM observations FINAL` usages (analytics aggregations) are out of scope.

### Impacted packages

- `@langfuse/shared` — `getObservationsForBlobStorageExport` SQL rewrite
- `worker` — soft-delete resurrection regression test

## Test plan

- [x] `pnpm --filter @langfuse/shared run lint` — passed
- [x] `pnpm --filter worker run lint` — passed
- [x] Manual hash-parity check: row count and `groupBitXor(cityHash64(tuple(*)))` verified identical between old and new query on a 1-hour window for a high-churn project (done during planning)
- [ ] `worker/src/__tests__/batchExport.test.ts > should not resurrect soft-deleted observations` — requires live DB; will run in CI

Replaces `FROM observations FINAL` with a manual `ORDER BY event_ts DESC / LIMIT 1 BY id, project_id` subquery to fix three production OOM incidents caused by `FINAL` forcing a full-table k-way merge before the time-range filter could be applied. The explicit outer `WHERE is_deleted = 0` correctly replicates the deletion-mask behaviour of the `ReplacingMergeTree(event_ts, is_deleted)` engine that `FINAL` applied implicitly.

- `getObservationsForBlobStorageExport` in `observations.ts` is rewritten from a flat `FINAL` scan to a subquery that reads only the requested time window, orders by `event_ts DESC`, deduplicates with `LIMIT 1 BY id, project_id`, then excludes soft-deleted rows in the outer `WHERE`.
- A regression test is added in `batchExport.test.ts` that inserts two versions of the same observation (alive then deleted) and asserts the export stream returns zero rows, guarding against soft-delete resurrection.
</details>

Safe to merge; the rewrite is well-reasoned, the table schema confirms the dedup semantics, and a hash-parity check was performed against production data.

The query transformation is correct for the common case: deletion events carry the same start_time as the original observation, so both versions land in the same time-window scan and LIMIT 1 BY event_ts DESC correctly picks the tombstone. The one gap is that the LIMIT 1 BY id, project_id dedup key is coarser than the table's ORDER BY (project_id, type, toDate(start_time), id) — if any observation ever had the same (id, project_id) with differing type or a cross-day start_time, the extra versions would be silently dropped. Current data appears clean (hash-parity verified), but the assumption is undocumented.

The SQL change in packages/shared/src/server/repositories/observations.ts deserves a second look to confirm the dedup-key assumption holds for all observation update patterns in production.
</details>

```mermaid
sequenceDiagram
    participant Caller
    participant ClickHouse as ClickHouse (observations)

    Note over Caller,ClickHouse: OLD path – FINAL forces full k-way merge
    Caller->>ClickHouse: SELECT cols FROM observations FINAL WHERE project_id AND start_time range
    ClickHouse-->>ClickHouse: Merge ALL parts across full table (75 GiB read, ~9 min)
    ClickHouse-->>Caller: Deduplicated rows (after WHERE)

    Note over Caller,ClickHouse: NEW path – manual dedup within window
    Caller->>ClickHouse: Inner: SELECT cols, is_deleted FROM observations WHERE project_id AND start_time range ORDER BY event_ts DESC LIMIT 1 BY id, project_id
    ClickHouse-->>ClickHouse: Scan only requested time-window partitions (1.3 MB read, ~0.5 s)
    ClickHouse-->>ClickHouse: "Outer: SELECT * EXCEPT(is_deleted) WHERE is_deleted = 0"
    ClickHouse-->>Caller: Deduplicated, non-deleted rows
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/repositories/observations.ts:1857-1858
**Dedup key narrower than table ORDER BY**

The table is defined as `ORDER BY (project_id, type, toDate(start_time), id)` — meaning two rows with the same `(id, project_id)` but different `type` or calendar-day boundary in `start_time` are distinct rows and would not be collapsed by `FINAL`. `LIMIT 1 BY id, project_id` uses a coarser key, so in that edge case only the latest-`event_ts` row survives and the other is silently dropped from the export. The hash-parity check in the PR description provides confidence this doesn't occur on current data, but it's a latent correctness difference worth documenting as an explicit assumption: each `(id, project_id)` pair always has a single canonical `(type, toDate(start_time))` across all its versions.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["perf(blob-export): replace observations ..."](https://github.com/langfuse/langfuse/commit/2e0bd8fc9ee1cb0c3f69ae133af088788c798ec2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34573652)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/observations.ts
+++ b/packages/shared/src/server/repositories/observations.ts
@@ -1843,10 +1843,12 @@ export const getObservationsForBlobStorageExport = function (
       tool_call_names,
       tool_definitions,
       usage_pricing_tier_name
-    FROM observations FINAL
+    FROM observations
     WHERE project_id = {projectId: String}
     AND start_time >= {minTimestamp: DateTime64(3)}
     AND start_time <= {maxTimestamp: DateTime64(3)}
+    ORDER BY event_ts DESC
+    LIMIT 1 BY id, project_id, type
   `;
 
   const records = queryClickhouseStream<Record<string, unknown>>({

### CODE REVIEW

**Comment by niklassemmler** [comment]
* **#13935** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/13935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/13935?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

**Review by sumerman — APPROVED** [review]
❤️

File: packages/shared/src/server/repositories/observations.ts [hunk 1]
```diff
@@ -1808,45 +1808,56 @@ export const getObservationsForBlobStorageExport = function (
...
+        usage_pricing_tier_name,
+        is_deleted
+      FROM observations
+      WHERE project_id = {projectId: String}
+      AND start_time >= {minTimestamp: DateTime64(3)}
+      AND start_time <= {maxTimestamp: DateTime64(3)}
+      ORDER BY event_ts DESC
+      LIMIT 1 BY id, project_id
+    )
+    WHERE is_deleted = 0
```
Steffen911: This is not actually used in our setup and we can drop it.

PR #13941, opened 2026-05-29
chore: adding extension points for matches operator

sumerman:


web/src/components/table/key-value-filter-builder.tsx
```diff
@@ -365,21 +384,19 @@ export function KeyValueFilterBuilder(props: KeyValueFilterBuilderProps) {
                   value={filter.operator}
                   onValueChange={(value) =>
                     handleFilterChange(index, {
-                      operator: value as "=" | "contains" | "does not contain",
+                      operator: value as StringKeyValueOperator,
```
bezbac, line 387: Might make sense to validate here instead of casting

web/src/features/filters/hooks/useSidebarFilterState.tsx
```diff
@@ -90,11 +102,40 @@ export function decodeAndNormalizeFilters(
...
+  columnDefinitions: ColumnDefinition[],
+  migrateFilterState?: FilterStateMigration<FilterState[number]>,
+): FilterState {
+  return decodeAndNormalizeFiltersWithSchema(
+    filtersQuery,
+    columnDefinitions,
+    singleFilter,
+    migrateFilterState,
+  );
 }
```
bezbac, line 118: Is this wrapper really necessary? I checked and I think there is only one "real" callsite to `decodeAndNormalizeFilters` outside of tests.

web/src/features/filters/lib/filter-config.ts
```diff
@@ -1,5 +1,54 @@
...
+  "contains",
+  "does not contain",
+] as const satisfies readonly StringKeyValueOperator[];
+
+export const FULL_TEXT_STRING_KEY_VALUE_OPERATORS = [
+  FTS_MATCH_OPERATOR,
+  "=",
+  "contains",
+  "does not contain",
+] as const satisfies readonly StringKeyValueOperator[];
```
bezbac, line 51: Are these always a superset of `DEFAULT_STRING_KEY_VALUE_OPERATORS`? Might make sense to spread `DEFAULT_STRING_KEY_VALUE_OPERATORS`?

web/src/features/filters/lib/filter-config.ts
```diff
@@ -1,5 +1,54 @@
...
+  | "starts with"
+  | "ends with";
+
+export type StringKeyValueOperator =
+  | "="
+  | "contains"
+  | "does not contain"
+  | "starts with"
+  | "ends with"
+  | typeof FTS_MATCH_OPERATOR;
```
bezbac, line 36: Same with this, might make sense to make `StringKeyValueOperator` extend `StringFilterOperator`

bezbac (commented):
I'm not very familiar with this code here. I just had a brief look and have some more nitpicks / stylistic comments. I think someone with more experience around the actual logic should definitely have a look too.

PR #13944, opened 2026-05-29
feat(mcp): Add listUsers tool

bezbac:

This PR introduces a `listUsers` MCP tool that lets MCP clients discover project members — including their IDs, names, and email addresses — to support downstream tools like `createAnnotationQueueAssignment`. The implementation follows the existing feature-module pattern and is backed by the already-tested `getUserProjectRoles` / `getUserProjectRolesCount` query helpers.

- A new `userService.ts` is added at `web/src/features/users/server/` that wraps the shared query helpers with a pagination and search-filter API; the resulting `listUsersForProject` function is called by the new MCP handler.
- `buildUserSearchFilter` is duplicated from `membersRouter.ts` rather than shared, and `filterCondition` is always hardcoded to `[]`, leaving minor cleanup opportunities.
- The response exposes user email addresses to any holder of a project API key; this mirrors what the existing TRPC route already returns to authenticated users, but widens the accessible surface to API-key-scoped callers.
</details>

Safe to merge; the new tool is read-only and follows established MCP patterns, with no breaking changes to existing tools.

The implementation is straightforward and correctly reuses existing, tested query helpers. The main things to revisit are a duplicated helper function, a constant empty filterCondition that can be simplified, and the fact that user email addresses are now enumerable by any project API key holder via MCP — a wider PII surface than the current TRPC route allows.

web/src/features/users/server/userService.ts warrants a second look for the duplicated filter helper and the unused filterCondition threading. web/src/features/mcp/features/users/tools/listUsers.ts deserves consideration of whether email belongs in the public MCP response.
</details>

```mermaid
sequenceDiagram
    participant Client as MCP Client
    participant MCP as MCP Server (bootstrap.ts)
    participant Tool as listUsers handler (listUsers.ts)
    participant Service as listUsersForProject (userService.ts)
    participant DB as PostgreSQL (via Prisma)

    Client->>MCP: "call listUsers({searchQuery?, page, limit})"
    MCP->>Tool: "handleListUsers(input, context{projectId, orgId})"
    Tool->>Service: "listUsersForProject({projectId, orgId, searchQuery, page, limit})"
    par Fetch users and count
        Service->>DB: "getUserProjectRoles(..., LIMIT N, OFFSET page*N)"
        Service->>DB: getUserProjectRolesCount(...)
    end
    DB-->>Service: "[{id, name, email, role}]"
    DB-->>Service: totalCount (bigint to Number)
    Service-->>Tool: "{users: [{id, name, email}], totalCount}"
    Tool-->>MCP: ListUsersResponse.parse(result)
    MCP-->>Client: "{users, totalCount}"
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/features/users/server/userService.ts:7-13
**Duplicated search filter function**

`buildUserSearchFilter` here is a near-identical copy of the same function in `web/src/features/rbac/server/membersRouter.ts` (lines 36–50). Both build the same `AND (u.name ILIKE ... OR u.email ILIKE ...)` fragment; the only difference is that the router version uses `Prisma.join` while this one inlines the `OR`. If the filter logic ever changes, both sites will need updates independently. This function could be extracted to a shared location (e.g., `packages/shared/src/server/auth/userProjectRoleAuth.ts` alongside the query helpers it feeds).

### Issue 2 of 3
web/src/features/users/server/userService.ts:29
**Unused `filterCondition` parameter**

`filterCondition` is hardcoded to `[]` and passed straight through to `getUserProjectRoles` / `getUserProjectRolesCount`. The shared query helper already handles an empty `FilterState` gracefully, but threading a constant empty array through the call chain makes the code appear to support dynamic column filtering when it never actually does. If extra column-level filtering is not planned, the parameter can be dropped from the call sites and the `FilterState` import removed from this file.

### Issue 3 of 3
web/src/features/mcp/features/users/tools/listUsers.ts:14-22
**User email addresses exposed to MCP callers**

The response schema includes `email` (and `name`), which are PII. The existing TRPC member-listing route is gated behind an authenticated Langfuse session, whereas any holder of a project API key (or in-app agent key) can call this MCP tool and enumerate the full set of members' email addresses. If email is only needed to help an LLM find a user and then pass their `id` to another tool, consider omitting it from the response or keeping only the `id` and a display label, reducing the PII surface exposed via API key-scoped calls.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(mcp): Add listUsers tool"](https://github.com/langfuse/langfuse/commit/7ab10b4660e7156849efc8a4a6d6911284947735) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34580893)</sub>

### CODE REVIEW

File: web/src/features/mcp/features/users/tools/listUsers.ts [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+
+const ListUsersResponse = z.object({
+  data: z.array(
+    z.object({
+      id: z.string(),
+      name: z.string().nullable(),
+    }),
+  ),
+  meta: paginationMetaResponseZod,
+});
```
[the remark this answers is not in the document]
bezbac: I think this might make sense to keep, as otherwise MCP will not be able to find specific users.

packages/shared/src/server/auth/userProjectRoleAuth.ts
```diff
@@ -73,6 +86,7 @@ function generateUserProjectRolesQuery({
         AND NOT EXISTS (
           SELECT 1 FROM project_memberships pm 
           WHERE pm.org_membership_id = om.id
+            AND pm.project_id = ${projectId}
```
nimarb, line 89: do we need this here? we just check for same org which should be fine or am i missing sth

nimarb (approved):

packages/shared/src/server/auth/userProjectRoleAuth.ts
bezbac: Hmm, I think it's fine.

As I understand it the full query is a union of the default organizational members and the project specific overrides. The default members should be all users who do not have an override for the _specific_ project we are looking for, if they have an override for another project we should ignore that, then we join it with the users from project_memberships.

So assume we have an org:
- User A (no explicit project membership overrides)
- User B (override for project X)
- User C (override for project Y)

When asking for all users in project X, User A and C should be read from the org membership and user B should be read from the project membership. Without the `AND pm.project_id = ${projectId}` the query would read B and C from the project memberships but C does not have an entry.

I hope I understand this correctly.

PR #13971, opened 2026-06-01
fix(annotation-queues): Race-safety of `createAnnotationQueueForApi`

bezbac:
Flagged by greptile here: https://github.com/langfuse/langfuse/pull/13850#discussion_r3303384666

This PR eliminates a TOCTOU race in `createAnnotationQueueForApi` by consolidating the hobby-plan count check, score-config existence validation, and the queue `create` call into a single `SERIALIZABLE` transaction. A `.catch` handler maps Prisma error codes `P2002` (unique constraint) and `P2034` (serialization failure) to user-friendly `InvalidRequestError` / `MethodNotAllowedError` responses.

- The transaction body correctly performs the hobby-plan count, score-config `findMany`, and `annotationQueue.create` in one atomic unit, with the database-level unique constraint as a final backstop for name collisions.
- The `P2034` recovery path re-checks name collisions and the hobby queue count against a fresh outer `prisma` connection (correct, since `tx` is already aborted), and re-throws the original error when neither condition is met.
- One gap: the P2034 handler does not re-validate `scoreConfigIds`, so if the serialization failure was caused by a concurrent score-config deletion the caller receives a raw P2034 Prisma error rather than the descriptive "score config IDs cannot be found" message.
</details>

Safe to merge; the core race-condition fix is correct and the error-handling paths are well thought out.

The serializable transaction correctly closes the original TOCTOU window, and the P2002/P2034 catch handlers cover the two most likely concurrent-write scenarios. The one uncovered path — a concurrent score-config deletion triggering P2034 — would surface a raw Prisma error to the caller rather than a friendly message, but it does not cause data corruption or a silent failure. The change is narrow and the logic is straightforward.

The P2034 recovery block in `publicAnnotationQueueService.ts` (lines 226–259) is worth a second look to confirm whether a score-config re-validation step should be added for completeness.
</details>

```mermaid
flowchart TD
    A[createAnnotationQueueForApi] --> B["prisma.$transaction(SERIALIZABLE)"]
    B --> C{plan === cloud:hobby?}
    C -- Yes --> D[COUNT queues]
    D --> E{count >= 1?}
    E -- Yes --> F[throw MethodNotAllowedError]
    E -- No --> G[findMany scoreConfigs]
    C -- No --> G
    G --> H{all scoreConfigIds found?}
    H -- No --> I[throw InvalidRequestError]
    H -- Yes --> J[annotationQueue.create]
    J --> K[return queue record]
    K --> L[auditLog + return API response]
    B -- error --> M{P2002 unique constraint?}
    M -- Yes --> N[throw InvalidRequestError name already exists]
    M -- No --> O{P2034 serialization failure?}
    O -- Yes --> P[findFirst queue by name]
    P --> Q{existingQueue?}
    Q -- Yes --> R[throw InvalidRequestError name already exists]
    Q -- No --> S{plan === cloud:hobby?}
    S -- Yes --> T[COUNT queues via prisma]
    T --> U{count >= 1?}
    U -- Yes --> V[throw MethodNotAllowedError]
    U -- No --> W[re-throw P2034 error]
    S -- No --> W
    O -- No --> X[re-throw original error]
```
</details>

`````markdown

---

### Issue 1 of 1
web/src/features/annotation-queues/server/publicAnnotationQueueService.ts:226-259
**P2034 handler silently swallows score-config race**

The P2034 (serialization failure) catch block only re-inspects name collisions and the hobby queue count. If the failure was actually triggered by a concurrent deletion of one of the `scoreConfigIds` supplied in the request, neither check will fire and the raw P2034 Prisma error is re-thrown directly to the caller — bypassing the friendly `"At least one of the score config IDs cannot be found"` message and instead surfacing an opaque internal error. Adding a score-config re-validation step (analogous to the existing name and plan checks) would give callers an actionable error message in this case.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: Race-safety of createAnnotationQueu..."](https://github.com/langfuse/langfuse/commit/568da4fa321446e4e1c412fd316ea25169f5a442) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34905910)</sub>

### CODE REVIEW

File: web/src/__tests__/server/mcp-tools-write.servertest.ts [hunk 1]
```diff
@@ -54,6 +66,143 @@ describe("MCP Write Tools", () => {
...
+        handleCreateAnnotationQueue(
+          {
+            name: queueName,
+            description: "Duplicate MCP queue",
+            scoreConfigIds: [scoreConfig.id],
+          },
+          context,
+        ),
+      ).rejects.toThrow("A queue with this name already exists.");
+    });
```
nimarb: do we need to test this here? this should be on the public api level

web/src/__tests__/server/mcp-tools-write.servertest.ts
```diff
@@ -54,6 +66,143 @@ describe("MCP Write Tools", () => {
...
+          "Maximum number of annotation queues reached on Hobby plan.",
+        );
+      } finally {
+        if (previousCloudRegion === undefined) {
+          delete process.env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION;
+        } else {
+          process.env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION = previousCloudRegion;
+        }
+      }
+    });
```
nimarb: is this also not handled on API (service) level? if so, we don't need an additional test here

nimarb (approved):

PR #13989, opened 2026-06-02
fix(otel): extract Google ADK cached_content_token_count into usage d…

### LINKED ISSUE #13988: bug: Anthropic cache-read tokens dropped before Langfuse persists usage (Google ADK + LiteLLM)
### Describe the bug

Hey! Pretty sure I ran into the Python / Google-ADK cousin of langfuse/langfuse#13811 (same kind of thing, but that one's about the JS langchain integration).

My setup: a Google ADK agent running Claude (Sonnet) through LiteLLM, with Anthropic prompt caching turned on for the system prompt. For tracing I just create a `Langfuse()` client at startup so it registers the global OpenTelemetry tracer provider, and ADK's spans get picked up from there — I'm not making any manual generation calls on the main path.

The cache is clearly working: LiteLLM's response comes back with `cache_read_input_tokens` filled in on the cached turns. But by the time the generation lands in Langfuse, the usage only has `input` / `output` / `total` — no `cache_read` / `cache_creation` breakdown at all. So every cached token gets billed as fresh input, and since Anthropic only charges \~10% for cache reads, the cost Langfuse reports on cache-heavy turns ends up way higher than what I'm actually paying (several times over on turns where most of the prompt is cached).

For what it's worth, the in-process `langfuse.openai` wrapper handles this fine — its `_parse_usage` keeps `prompt_tokens_details` / `input_token_details`. So this looks specific to the OTel / ADK ingestion path, not the SDK in general.

**Expected:** the stored generation keeps the cache split (something like `input_cache_read`, ideally `input_cache_write` too) and Langfuse applies the discounted cache-read price.

**Actual:** the whole prompt is priced as uncached input, so cost is overcounted.

### Steps to reproduce

Rough repro — fire the same large system prompt twice so the second turn hits Anthropic's prompt cache:

```python
from google.adk.models.lite_llm import LiteLlm
from google.adk.plugins import BasePlugin

model = LiteLlm(
    model="anthropic/claude-sonnet-4-6",
    api_base="<your anthropic-compatible gateway>",
    api_key="<key>",
    drop_params=True,
    # litellm marks the system message ephemeral -> Anthropic prompt caching
    cache_control_injection_points=[
        {"location": "message", "role": "system", "control": {"type": "ephemeral"}}
    ],
)

# tiny plugin just to dump what ADK hands back
class UsageProbe(BasePlugin):
    async def after_model_callback(self, *, callback_context, llm_response):
        print("ADK usage_metadata:", llm_response.usage_metadata)
```

Wire that model into an agent, run the same prompt twice, then open the two generations in Langfuse.

Following the token count through the layers (same idea as langfuse/langfuse#13811):

* **LiteLLM raw response** → `cache_read_input_tokens` is there (a few thousand on turn 2). ✅
* **Langfuse stored usage** → only `input` / `output` / `total`, cache numbers gone. ❌

The one bit I haven't fully nailed down is the middle layer — whether ADK's `usage_metadata.cached_content_token_count` already comes back empty (→ google-adk isn't putting it on the span) or whether it's there and Langfuse's OTel ingestion just isn't mapping it (→ the direct Python-side equivalent of langfuse/langfuse#13811). A pointer on where to look would be great — happy to dig in once I know which side owns it.

Self-hosted

*No response*

* `langfuse` (Python SDK): **3.8.1**
* `google-adk`: **1.31.1**
* `litellm`: **1.79.3**
* `opentelemetry-sdk`: **1.38.0**
* Model: `claude-sonnet-4-6` via the LiteLLM `anthropic/` provider, prompt caching enabled with `cache_control_injection_points` on the system message
* Python: **3.10**

### Additional information

Possibly related, possibly its own thing: on this ADK version I also see two token-bearing spans per LLM call (`call_llm` and `generate_content`), and both seem to get persisted as generations, which double-counts tokens independently of the cache issue. Happy to split that into a separate report if it's unrelated — just flagging it in case they share a code path.

Yes

dddlllfff:
## What does this PR do?

Google ADK emits its OpenTelemetry spans under the `gcp.vertex.agent` instrumentation scope. For LLM generations it records the prompt/completion token counts as `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` (already mapped by the generic `gen_ai` extractor), but the **cached-token count is only available inside the serialized response** on `gcp.vertex.agent.llm_response` (`usage_metadata.cached_content_token_count`) — it is never emitted as a `gen_ai.usage.*` attribute.

As a result, `OtelIngestionProcessor.extractUsageDetails` never sees the cached tokens: the whole prompt is recorded as uncached `input` and billed at the full input rate. On Anthropic/Gemini prompt-cache-heavy turns this inflates the reported cost several-fold, since cache reads are billed at ~10% of the input price.

This PR adds a Google ADK branch to `extractUsageDetails` that parses `cached_content_token_count` out of `gcp.vertex.agent.llm_response` and records it as `input_cached_tokens`, leaving `input` as the uncached remainder. `gen_ai.usage.input_tokens` already includes the cached tokens (matching how LiteLLM/Gemini report usage), so this mirrors the existing OpenAI / Anthropic / Bedrock provider-metadata cache handling a few lines above.

Tool-call spans (where ADK sets `gcp.vertex.agent.llm_response` to `{}`) are unaffected.

Fixes #13988

## Type of change

## Mandatory Tasks

## Checklist

- Added a focused test in `web/src/__tests__/server/api/otel/otelMapping.servertest.ts` that feeds a representative ADK `call_llm` span (`gen_ai.usage.input_tokens=5000`, `gcp.vertex.agent.llm_response` carrying `cached_content_token_count=2560`) and asserts `input=2440`, `output=200`, `input_cached_tokens=2560`.
- The new logic mirrors the existing provider cache branches and adds no new dependencies.

This PR fixes a billing accuracy issue for Google ADK (`gcp.vertex.agent`) spans by extracting `cached_content_token_count` from the `gcp.vertex.agent.llm_response` JSON payload and recording it as `input_cached_tokens`, with `input` adjusted to the uncached remainder.

- Adds a new branch in `extractUsageDetails` that parses the serialized LLM response from `gcp.vertex.agent.llm_response`, reads `usage_metadata.cached_content_token_count`, and performs the same cache-deduction logic already present for the Vercel AI SDK's OpenAI/Anthropic/Bedrock provider-metadata paths.
- Includes a focused integration test covering the main scenario (5000 total input, 2560 cached → 2440 uncached input, 200 output).
</details>

Safe to merge; the change is additive and only activates for spans carrying the ADK-specific `gcp.vertex.agent.llm_response` attribute.

The cache-deduction logic is correct and mirrors existing provider branches. The only concern is that the detection relies purely on attribute presence rather than combining it with `instrumentationScopeName`, which is how every other special-case branch in the same function is guarded. This is unlikely to cause a real problem today but departs from the established pattern.

`OtelIngestionProcessor.ts` around the new ADK branch deserves a second look to decide whether to add an `instrumentationScopeName` guard consistent with surrounding branches.
</details>

```mermaid
flowchart TD
    A[extractUsageDetails called] --> B{langfuse.observation.usageDetails\nattribute present?}
    B -- Yes --> Z[Return attribute value]
    B -- No --> C{instrumentationScopeName\n== 'genkit-tracer'?}
    C -- Yes --> D[Extract Genkit usage\nfrom genkit:output]
    C -- No --> E{instrumentationScopeName\n== 'ai'?}
    E -- Yes --> F[Extract AI SDK usage\nfrom gen_ai.usage.*\n+ ai.response.providerMetadata]
    E -- No --> G{instrumentationScopeName\n== 'pydantic-ai'?}
    G -- Yes --> H[extractGenericGenAiUsageDetails]
    G -- No --> I{NEW: 'gcp.vertex.agent.llm_response'\nin attributes?}
    I -- Yes --> J[extractGenericGenAiUsageDetails\nReturns input=5000, output=200]
    J --> K[parseJsonPayload\ngcp.vertex.agent.llm_response]
    K --> L{cached_content_token_count\nis number > 0?}
    L -- Yes --> M[input_cached_tokens = 2560\ninput = max 5000-2560, 0 = 2440]
    L -- No --> N{usageDetails non-empty?}
    M --> N
    N -- Yes --> O[Return usageDetails\ninput=2440, output=200,\ninput_cached_tokens=2560]
    N -- No --> P[extractGenericGenAiUsageDetails\nfallback]
    I -- No --> P
```
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/otel/OtelIngestionProcessor.ts:2400-2422
**Attribute-based detection with no `instrumentationScopeName` guard**

All other special-case branches (`genkit-tracer`, `ai`, `pydantic-ai`) guard on `instrumentationScopeName` first, then inspect attributes. This branch only checks for the presence of `gcp.vertex.agent.llm_response`. If an unrelated instrumentation library ever sets an attribute with that exact name, `extractGenericGenAiUsageDetails` would be silently called a second time inside this branch, the cached-token deduction could produce unexpected results, and the fallback at line 2424 would never be reached. Adding `instrumentationScopeName === "gcp.vertex.agent"` as a secondary guard (similar to how pydantic-ai works) would eliminate that ambiguity.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(otel): extract Google ADK cached\_con..."](https://github.com/langfuse/langfuse/commit/998bb37984976125037732ab7efc0752eaac65ca) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35081309)</sub>


nimarb:
@claude review

PR #13995, opened 2026-06-02
feat(scores): add v3 scores API — polymorphic value, flag-gated (Phase 1)

niklassemmler:
## Summary

Phase 1 of the [LFE-9539](https://linear.app/langfuse/issue/LFE-9539) Scores v3 API redesign: minimal working list endpoint behind a feature flag with a polymorphic \`value\` field.

- **\`GET /api/public/v3/scores\`** — list scores, limit-only pagination (default 50, max 100)
- **Polymorphic \`value\`**: \`NUMERIC → number\`, \`BOOLEAN → boolean\`, \`CATEGORICAL/TEXT/CORRECTION → string\` — no more \`stringValue\` / \`value\` split in the response
- **Feature-flagged** behind \`LANGFUSE_ENABLE_SCORES_V3_API=true\`; returns 404 when disabled
- **No JOIN, no COUNT** — pure ClickHouse read path; deduplication via \`LIMIT 1 BY id, project_id\`
- **\`ORDER BY timestamp DESC, id DESC, event_ts DESC\`** — \`id\` is the stable pagination tiebreaker; \`event_ts\` is last, used only for \`LIMIT 1 BY\` deduplication of multi-version rows
- **v1/v2 untouched** — v3 owns its own service module (\`scores-api-v3.ts\`) with no imports from the v1/v2 kernel

### Why a new service module?

v1/v2 couple score fetching to a Postgres trace JOIN and a COUNT query — the root cause of the 30 s timeout at scale. v3 starts clean: pure ClickHouse, no JOIN, no COUNT. Keeping v3 isolated means v1/v2 can be deleted later with a clean \`rm\`.

### Impacted packages

- \`web\` — new API route (\`/v3/scores/index.ts\`), service module, env var, integration tests
- \`@langfuse/shared\` — new Zod schemas (\`GetScoresV3\`, \`GetScoresResponseV3\`, \`APIScoreV3\`)
- \`fern/\` + \`web/public/generated/api/openapi.yml\` — new \`ScoresV3\` service block

## Test plan

- [x] \`pnpm --filter web run lint\` — passed
- [x] \`pnpm run typecheck\` — passed
- [x] \`pnpm --filter web exec vitest run scores-api-v3\` — 10 unit tests (polymorphicValue) always pass; integration tests gated on \`LANGFUSE_ENABLE_SCORES_V3_API=true\`
- [x] Fern generation — all checks passed, TypeScript + Python SDKs + OpenAPI regenerated
- [ ] Integration tests with \`LANGFUSE_ENABLE_SCORES_V3_API=true\` — require live ClickHouse


niklassemmler:
* **#14006** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14006?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14005** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14005?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14001** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14001?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13996** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/13996?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13995** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/13995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/13995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,126 @@
...
+    s.trace_id as trace_id,
+    s.observation_id as observation_id,
+    s.session_id as session_id,
+    s.dataset_run_id as dataset_run_id
+  FROM scores s
+  WHERE s.project_id = {projectId: String}
+  ORDER BY s.timestamp DESC, s.event_ts DESC, s.id DESC
+  LIMIT 1 BY s.id, s.project_id
+  LIMIT {limit: Int32}
+`;
```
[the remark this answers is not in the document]
niklassemmler: we will use it for the fields PR on top. Tbf we could have added it later

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,131 @@
...
```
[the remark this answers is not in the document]
niklassemmler: Let's address this as part of the performance review

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,131 @@
...
+  projectId: string;
+  scoreId: string;
+}): Promise<APIScoreV3 | undefined> {
+  const score = await getScoreById({
+    projectId: params.projectId,
+    scoreId: params.scoreId,
+  });
+  if (!score) return undefined;
+  return domainToV3(score);
+}
```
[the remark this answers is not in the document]
niklassemmler: addressed

web/src/features/public-api/server/scores-api-v3.ts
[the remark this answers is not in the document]
niklassemmler: removed the single score by id query so this no longer applies

fern/apis/server/_drafts/scores-v3.yml
```diff
@@ -0,0 +1,105 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/main/fern.schema.json
+imports:
```
Steffen911, line 2: Be aware that once we merge this, it is publicly documented. I.e. the other PRs need to follow quickly or be merged into this one.

fern/apis/server/_drafts/scores-v3.yml
```diff
@@ -0,0 +1,105 @@
...
+      docs: |
+        Get a list of scores with a polymorphic `value` field (v3).
+
+        The `value` field type depends on `dataType`:
+        - `NUMERIC` → number
+        - `BOOLEAN` → boolean
+        - `CATEGORICAL`, `TEXT`, `CORRECTION` → string
+
+        This endpoint is behind the `LANGFUSE_ENABLE_SCORES_V3_API` feature flag
+        and returns 404 when the flag is disabled.
```
Steffen911: External users don't have a way of seeing which flags are configured. I would rather say that this is restricted to the cloud version initially. Do we have similar wording for metrics or observations @sumerman ? How do we make it clear that they may not be available for self-hosters?

packages/shared/src/features/scores/interfaces/api/index.ts
```diff
@@ -3,3 +3,4 @@ export * from "./v1/validation";
 export * from "./v1/endpoints";
 export * from "./v2/validation";
 export * from "./v2/endpoints";
+export * from "./scores-v3";
```
Steffen911: Didn't you like the established pattern? 😄

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,125 @@
...
+      }
+      return score.stringValue;
+    case ScoreDataTypeEnum.CORRECTION:
+      if (score.longStringValue == null) {
+        throw new InternalServerError(
+          "Score with dataType CORRECTION is missing its longStringValue",
+        );
+      }
+      return score.longStringValue;
+    default:
```
Steffen911: We usually add an exhaustiveCheck like

```
   const _exhaustiveCheck: never = score.dataType
```

There should be a few examples on this across the code.

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,125 @@
...
+    }),
+    source: score.source,
+    timestamp: score.timestamp,
+    environment: score.environment,
+    createdAt: score.createdAt,
+    updatedAt: score.updatedAt,
+  } as APIScoreV3;
+}
+
+const v3ListQuery = `
```
Steffen911, line 76: We could consider moving this into the repositories/scores.ts to keep queries outside of the core logic.

web/src/features/public-api/server/scores-api-v3.ts
```diff
@@ -0,0 +1,125 @@
...
+    s.queue_id as queue_id,
+    s.execution_trace_id as execution_trace_id,
+    s.trace_id as trace_id,
+    s.observation_id as observation_id,
+    s.session_id as session_id,
+    s.dataset_run_id as dataset_run_id,
+    s.is_deleted as is_deleted
+  FROM scores s
+  WHERE s.project_id = {projectId: String}
+  ORDER BY s.timestamp DESC, s.id DESC, s.event_ts DESC
```
Steffen911, line 102: One more idea: Have you tried whether it makes a difference to `ORDER BY s.project_id, toDate(s.timestamp) DESC, s.id DESC, s.event_ts DESC` and then apply the LIMIT BY and LIMIT?
Clickhouse has a readinorder optimization if a ORDER BY statement matches the sorting key and this may allow it to return early instead of sorting the full table on a non-date-filtered scan.

Could you find a big cloud project and run both variants with explain pipeline and check whether a ReadInReverseOrder is applied?

fern/apis/server/_drafts/scores-v3.yml
niklassemmler: Good point. Moved to draft.

web/src/features/public-api/server/scores-api-v3.ts
niklassemmler: That's a cool way of handling it!

web/src/features/public-api/server/scores-api-v3.ts
niklassemmler: @Steffen911 I like moving it, but this would make it inconsistent with other servers in the same folder. How about we do a separate cleanup where we move all queries?

web/src/features/public-api/server/scores-api-v3.ts
niklassemmler: For sure, but we can do this improvement as a follow-up, right?

I want to make sure this works end-to-end first. Ofc we need to make sure we don't publish it early. Otherwise we may have issues with the cursor.

packages/shared/src/features/scores/interfaces/api/index.ts
niklassemmler: Shame on me 😬

web/src/features/public-api/server/scores-api-v3.ts
Steffen911: @niklassemmler Makes sense to me. Up to you on how to schedule and address this (if at all).

Steffen911 (approved):

PR #14017, opened 2026-06-03
feat(worker): V4 historic backfill background migration chain

Steffen911:
Stacked on #13623. Extracts the V4 self-hosted historic backfill **background migrations** (the data-movement logic) out of #13623 into this dedicated PR so they can be reviewed independently from the `events_only` read-path changes.

#13623 (base) retains the removals of the old backfill scripts and all `LANGFUSE_MIGRATION_V4_WRITE_MODE` / preview-opt-in read-path changes. This PR adds the new chain on top.

## What's here
- **Prisma migrations** registering the chain (dormant, env-gated):
  `createRootSpansFromTraces` → `rewriteObservationsToPidTidSorting` → `backfillEventsFullFromObservations` → `backfillEventsFullFromDatasetRunItems` → `dropPidTidSortingTables`
- **Background migration implementations** + shared `utils/backfillBase` (`ConcurrentQueryManager`, recovery helpers). `backfillEventsFullFromObservations` supersedes the old `backfillEventsHistoricFromParts` removed in the base PR.
- **`BackgroundMigrationManager` `envGate` support** — dormant rows are skipped in the `findFirst` predicate so they never head-of-line block later migrations; README docs the mechanism.
- **`LANGFUSE_BACKGROUND_MIGRATION_*` env gates** in the worker `EnvSchema`.

## Impacted packages
- `worker` (background migrations, env, manager)
- `packages/shared` (Prisma migrations)

## Verification
- `pnpm --filter worker run lint` — pass (pre-commit hook)
- `pnpm --filter worker run typecheck` — pass on **both** this branch and the base branch
- **Lossless split verified:** the tip tree of this branch is byte-for-byte identical to the pre-split branch HEAD (`git diff <pre-split> HEAD` is empty), i.e. base + this PR == the original combined branch.

## Merge order
Merge #13623 first; this PR will retarget to `main` automatically.

This PR introduces the V4 historic backfill migration chain for self-hosted Langfuse deployments: five sequentially-ordered background migrations that create root-span events from traces, rewrite observations into a sort-optimised scratch table, backfill `events_full` from observations and dataset-run items, and finally drop the scratch table. All five rows ship dormant via env gates so they never block unrelated migrations until an operator opts in.

- **New migrations** (`createRootSpansFromTraces` → `rewriteObservationsToPidTidSorting` → `backfillEventsFullFromObservations` → `backfillEventsFullFromDatasetRunItems` → `dropPidTidSortingTables`): implement the data-movement chain using a fire-and-poll pattern for ClickHouse queries, with per-chunk retry, part-merge detection, and crash-recovery via persisted state in `background_migrations.state`.
- **`BackgroundMigrationManager` env-gate support**: dormant rows are filtered out at `findFirst` time so they never head-of-line block other migrations; new `LANGFUSE_BACKGROUND_MIGRATION_V4_*` env vars added to `EnvSchema`.
- **`utils/backfillBase`**: shared `ConcurrentQueryManager`, `fireQuery`, `recoverInProgressTodos`, and chain-dependency guard `checkPredecessorMigrationFinalized`.
</details>

The chain is not safe to merge as-is: the DRI backfill migration can silently drop experiment enrichment writes on graceful shutdown, and resumes from a cursor that has already moved past the lost data.

The `backfillEventsFullFromDatasetRunItems` migration advances its Postgres cursor after each successfully processed batch but only flushes the ClickhouseWriter buffer at the very end of a normal run. If the worker is shut down gracefully between a cursor advance and the final flush, the buffered enriched-span writes are discarded while the cursor skips over those DRIs permanently on the next run. The rest of the chain (M1–M3, M5) uses direct server-side ClickHouse INSERTs via the fire-and-poll pattern and is not affected. The env-gate mechanism, predecessor-chain guards, merge-stop/resume logic, and part-integrity verification in M3 are all well-implemented.

worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts — specifically the abort return path and the DescendantCapExceededError rethrow path, both of which exit without calling flushAll on the ClickhouseWriter.
</details>

```mermaid
sequenceDiagram
    participant M as BackgroundMigrationManager
    participant M1 as M1: createRootSpansFromTraces
    participant M2 as M2: rewriteObservationsToPidTidSorting
    participant M3 as M3: backfillEventsFullFromObservations
    participant M4 as M4: backfillEventsFullFromDatasetRunItems
    participant M5 as M5: dropPidTidSortingTables
    participant CH as ClickHouse
    participant PG as Postgres (state)

    Note over M: Gate: V4_ENABLE_HISTORIC_BACKFILL=true
    M->>M1: validate() + run()
    M1->>CH: INSERT INTO events_full FROM traces (per partition)
    M1->>PG: finishedAt set
    M->>M2: validate() + run()
    M2->>CH: CREATE TABLE IF NOT EXISTS observations_pid_tid_sorting
    M2->>CH: INSERT INTO observations_pid_tid_sorting FROM observations (per partition)
    M2->>CH: SYSTEM STOP MERGES observations_pid_tid_sorting
    M2->>PG: finishedAt set
    M->>M3: validate() checks M2 finishedAt, run()
    M3->>CH: INSERT INTO events_full FROM observations_pid_tid_sorting LEFT JOIN traces (per part)
    M3->>PG: finishedAt set
    M->>M4: validate() checks M3 finishedAt, run()
    M4->>CH: SELECT DRI batches (cursor-paginated)
    M4->>CH: Write enriched spans via ClickhouseWriter
    M4->>PG: cursor advanced per batch, finishedAt set
    Note over M: Gate: V4_DROP_PID_TID_SORTING_TABLES=true
    M->>M5: validate() checks M4 finishedAt, run()
    M5->>CH: DROP TABLE IF EXISTS observations_pid_tid_sorting SYNC
    M5->>PG: finishedAt set
```
</details>

`````markdown

---

### Issue 1 of 3
worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts:649-653
**Unflushed writes on graceful abort**

`writeEnrichedSpans` queues data to `ClickhouseWriter`'s internal buffer — it is not immediately persisted. The cursor is advanced in Postgres *before* the sleep at line 646, so if `abort()` fires during that sleep and the loop exits here, any buffered writes from the last completed `processBatch` call are dropped silently. On the next run the cursor skips those DRIs and the experiment enrichment for them is permanently lost. The `flushAll` call at line 657 that handles the normal-exit path must also be reached on abort; returning early without it is what creates the gap.

### Issue 2 of 3
worker/src/backgroundMigrations/backgroundMigrationManager.ts:110-115
The `require()` call inside the transaction callback is a dynamic in-function import. The custom style guide asks for imports to be placed at the top of the module rather than inside functions. Because the script name is a runtime value the import genuinely must stay dynamic, but adding a suppression comment documents the intentional deviation and keeps linters quiet.

```suggestion
            BackgroundMigrationManager.activeMigration = {
              id: migration.id,
              name: migration.name,
              args: migration.args as any,
              // eslint-disable-next-line @typescript-eslint/no-var-requires
              migration: new (require(`./${migration.script}`).default)(),
            };
```

### Issue 3 of 3
worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts:446-457
**Partial write before `DescendantCapExceededError` is not flushed**

When the descendant cap is exceeded mid-batch, `writeEnrichedSpans(allEnrichedSpans)` queues the spans processed so far, then the error is thrown. `run()` re-throws immediately (line 619), exiting without calling `flushAll`. Those queued writes are lost. Unlike the abort path, the cursor has *not* advanced here (advance happens after the try block), so the affected DRIs will be re-processed on the next run. The duplicate rows ClickHouse would eventually deduplicate — but the first attempt's queued writes are silently discarded with no indication they were dropped.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(worker): V4 self-hosted historic ba..."](https://github.com/langfuse/langfuse/commit/13dcfb8df6a384970f607b471d60a000ec822c8f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35329041)</sub>

**Context used:**

**Learned From**

### CODE REVIEW

File: worker/src/backgroundMigrations/utils/backfillBase.ts [hunk 1]
```diff
@@ -0,0 +1,412 @@
...
+      try {
+        for (const [queryId, todo] of this.activeQueries) {
+          try {
+            const status = await pollQueryStatus(queryId);
+
+            if (status === "completed") {
+              this.activeQueries.delete(queryId);
+              try {
+                await onComplete(todo, true);
+                await scheduleNext();
```
Steffen911, line 63: > Clearly it is implied that scheduleNext updates activeQueries but it is not evident from the types and not mentioned in the comments.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+              }
+            } else if (status === "failed" || status === "not_found") {
+              this.activeQueries.delete(queryId);
+              const error =
+                status === "failed"
+                  ? await getQueryError(queryId)
+                  : "Query not found in query_log";
+              try {
+                await onComplete(todo, false, error);
+                await scheduleNext();
```
Steffen911, line 79: > So scheduleNext is called inside 2 branches of a non-exhaustive if statement. From the current QueryStatus definition it is evident that we don't want to scheduleNext while a query is still running. But I would prefer if that were explicit and the if/else were exhaustive.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+/**
+ * Fires a long-running ClickHouse query, confirms it's tracked in
+ * system.processes, then aborts the HTTP connection so the query continues
+ * server-side and we poll for completion via pollQueryStatus.
+ */
+export async function fireQuery({
+  query,
+  queryId,
+  params,
+  retryCount = 0,
```
Steffen911, line 277: Consider rebranding to attemptNumber.

> While reading this I thought that retryCount controls how many times this function retries the query. Instead, it looks like a retry attempt number that is used to pick from retrySettings. IMO should have a different name.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+    );
+    throw error;
+  }
+
+  logger.info(
+    `${logPrefix} Query ${queryId} confirmed running, aborting HTTP connection`,
+  );
+  abortController.abort();
+
+  queryPromise.catch((err) => {
```
Steffen911, line 336: Consider to catch before aborting.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+  retryWaitMs?: number;
+  logPrefix?: string;
+}
+
+/**
+ * Fires a long-running ClickHouse query, confirms it's tracked in
+ * system.processes, then aborts the HTTP connection so the query continues
+ * server-side and we poll for completion via pollQueryStatus.
+ */
+export async function fireQuery({
```
Steffen911, line 273: > can potentially be a general purpose helper along with pollQueryStatus

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -0,0 +1,723 @@
...
+
+      if (success) {
+        // Verify the part still exists. If a merge consolidated it mid-run
+        // the source rows are now in the merged successor — which is not in
+        // our todo list. Abort so the operator can clear `state.chunksLoaded`
+        // (or wipe state) and re-run to pick up the merged successor.
+        const partStillActive = await this.verifyPartStillActive(todo.partId);
+        if (!partStillActive) {
+          logger.error(
+            `[Backfill Events Observations] CRITICAL: Part ${todo.partId} no longer active after processing — ` +
```
Steffen911, line 514: > This is something that belongs in "Troubleshooting & FAQ" section along with the expected steps to recover.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -0,0 +1,723 @@
...
+        o.start_time AS start_time,
+        o.end_time,
+        o.name,
+        o.type,
+        o.environment,
+        coalesce(o.version, t.version) as version,
+        coalesce(t.release, '') as release,
+        t.tags as tags,
+        t.public as public,
+        t.bookmarked AND (o.parent_observation_id IS NULL OR o.parent_observation_id = '') AS bookmarked,
```
Steffen911, line 262: Add comment that bookmarked is only meant to be present on root spans.

worker/src/backgroundMigrations/utils/backfillBase.ts
```diff
@@ -0,0 +1,412 @@
...
+ *
+ * The `BackgroundMigrationManager` has no dependency model — it runs every eligible
+ * row in name order and marks each finished/failed independently, so a failed
+ * upstream step does NOT by itself stop a downstream step from running on
+ * partial data. Each downstream step calls this from its `validate()`, which
+ * routes through the manager's validation-failure path and records `failedAt`.
+ * Because the next step in turn guards on its own predecessor, a single failure
+ * transitively halts the rest of the chain.
+ */
+export async function checkPredecessorMigrationFinalized(
```
Steffen911, line 153: > From what I can tell recovering from this and some other failure modes requires manual state manipulation in PostgreSQL. I don't think we have it documented anywhere.

Document recovery for failing background migrations.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -0,0 +1,723 @@
...
+
+      if (pendingTodos.length === 0 && manager.activeCount === 0) {
+        manager.stopPolling();
+        state.phase = "completed";
+        await this.updateState(state);
+        logger.info("[Backfill Events Observations] All parts completed!");
+        return;
+      }
+
+      if (manager.activeCount >= config.concurrency!) return;
```
Steffen911, line 444: > nit: I find it somewhat strange that ConcurrentQueryManager doesn't actually control concurrency. But this has no bearing on correctness.

worker/src/backgroundMigrations/backfillEventsFullFromObservations.ts
```diff
@@ -0,0 +1,723 @@
...
+    );
+
+    for (const todo of stillRunningTodos) {
+      manager.addQuery(todo, todo.queryId!);
+      logger.info(
+        `[Backfill Events Observations] Added recovered running query ${todo.queryId} for part ${todo.partId} to manager`,
+      );
+    }
+
+    const slotsAvailable = config.concurrency! - stillRunningTodos.length;
```
Steffen911, line 572: > @[sumerman](https://github.com/sumerman)
> sumerman
> [2 days ago](https://github.com/langfuse/langfuse/pull/13623#discussion_r3343202252)
> I believe this is a data race. As both this loop and startPolling may call scheduleNext which does SELECT, …, UPDATE outside of a transaction. I think we should either move state management outside of scheduleNext or relinquish control to ConcurrentQueryManager completely and await until aborted or done. Or remove setInterval from ConcurrentQueryManager and run its iterations from the while loop below.
> 
> Member
> @[sumerman](https://github.com/sumerman)
> sumerman
> [2 days ago](https://github.com/langfuse/langfuse/pull/13623#discussion_r3343224126)
> Asked codex to check if it exists elsewhere. Here is the result:
> 
> Yes. The same race pattern exists in two other background migrations:
> createRootSpansFromTraces.ts (line 309): scheduleNext does loadState → picks first pending todo → updateState outside a transaction. It is invoked by ConcurrentQueryManager.startPolling at line 412 (line 412) and by the initial slot-fill loop at line 426 (line 426).
> 
> rewriteObservationsToPidTidSorting.ts (line 437): same shape. scheduleNext mutates JSON state at line 459 (line 459), persists it at line 463 (line 463), and is called both from polling at line 540 (line 540) and the slot-fill loop at line 554 (line 554).
> 
> The shared trigger is backfillBase.ts (line 50): the interval callback calls scheduleNext after completion/failure at line 63 (line 63) and line 79 (line 79). isPolling serializes interval ticks, but it does not serialize against external calls to the same scheduleNext.

worker/src/backgroundMigrations/backfillEventsFullFromDatasetRunItems.ts
```diff
@@ -0,0 +1,764 @@
...
+        t.public,
+        t.name AS trace_name,
+        coalesce(t.user_id, '') AS user_id,
+        coalesce(t.session_id, '') AS session_id
+      FROM traces t
+      WHERE t.project_id IN {projectIds: Array(String)}
+        AND t.id IN {traceIds: Array(String)}
+        AND t.timestamp >= {minTime: DateTime64(3)}
+        AND t.timestamp <= {maxTime: DateTime64(3)}
+      ORDER BY t.event_ts DESC
```
Steffen911, line 370: > shouldn't this ORDER BY be more aligned with traces table order or do we not expect large volumes of rows from individual executions?

Steffen911 (commented):
@sumerman Previous review comments.

Steffen911:
Claude comment:

```
**Additional findings** (outside current diff — PR may have been updated during review):

- 🔴 `worker/src/backgroundMigrations/utils/backfillBase.ts:296-325` — In `fireQuery` (worker/src/backgroundMigrations/utils/backfillBase.ts:300-336), `queryPromise = commandClickhouse(...)` is created with no error handler, then control awaits `sleep(initialWaitMs)` (5s) plus up to two `pollQueryStatus` calls separated by another `sleep(retryWaitMs)` (15s) before `queryPromise.catch(...)` is finally attached at line 336 — a 5-21 second window in which any rejection (CH parse error, auth failure, ECONNREFUSED, TLS failure, malformed param, socket reset under backpressure) lands unhandled. Each migration script's `main()` installs a `process.on('unhandledRejection', ...) -> process.exit(1)` handler, and Node 24's default `--unhandled-rejections=throw` does the same in worker-mode; either way a transient blip kills the pod mid-migration and bypasses the migration's own retry logic, turning a one-off failure into a crash loop. Fix: move `queryPromise.catch(...)` to immediately after the `commandClickhouse(...)` call on line 310, before any `await`.

  <details><summary>Extended reasoning...</summary>

  ### What the bug is

  `fireQuery` at `worker/src/backgroundMigrations/utils/backfillBase.ts:282-348` creates a long-running CH query and then immediately yields:

  ```ts
  const queryPromise = commandClickhouse({ query, ..., abortSignal });   // line 300 — no .catch
  await sleep(initialWaitMs);                                            // line 312 — 5s default
  try {
    const status = await pollQueryStatus(queryId);                       // line 315
    if (status === 'not_found') {
      await sleep(retryWaitMs);                                          // line 317 — 15s
      const retryStatus = await pollQueryStatus(queryId);                // line 318
      if (retryStatus === 'not_found') throw new Error(...);
    }
  } catch (error) { ... throw error; }
  abortController.abort();                                               // line 334
  queryPromise.catch((err) => { ... });                                  // line 336 — finally attached
  ```

  The .catch is attached 5-21 seconds after the promise is created. Any rejection that resolves in that window has no handler in the current microtask, so Node fires the `unhandledRejection` event.

  ### Rejection paths inside the window

  `commandClickhouse` (packages/shared/src/server/repositories/clickhouse.ts) wraps `clickhouseClient().command(...)` in `instrumentAsync`. It can reject early from:
  - ClickHouse parse/syntax error returned as a fast HTTP error before the query starts.
  - HTTP 401/403 from misconfigured credentials.
  - ECONNREFUSED / DNS failure / TLS handshake failure.
  - Client-side `@clickhouse/client` validation (malformed query params).
  - Socket reset under backpressure.

  All of these can reject well before the 5s initial sleep completes, and certainly before the 20s mark.

  ### Why Node terminates the worker

  **CLI mode**: each migration's `main()` installs:
  ```ts
  process.on('unhandledRejection', (reason) => { logger.error(...); process.exit(1); });
  ```
  Confirmed in createRootSpansFromTraces.ts, rewriteObservationsToPidTidSorting.ts, backfillEventsFullFromObservations.ts, backfillEventsFullFromDatasetRunItems.ts, and dropPidTidSortingTables.ts. Process exits with code 1.

  **Worker mode** (BackgroundMigrationManager): no worker-level `unhandledRejection` handler exists (verified by grep across worker/src). Node 24's default `--unhandled-rejections=throw` (since Node 15) terminates the process. `worker/package.json` engines pins `node: "24"`.

  The microtask timing is decisive: `unhandledRejection` fires at the end of the task that produced the rejection, before our code resumes from `await sleep(...)`. The .catch attached 5-21 seconds later only suppresses the followup `rejectionHandled` event — it does not unwind the already-fired termination.

  ### Impact

  Any fast-failing CH-side error during the 5-21s window crashes the worker pod mid-migration. The migration's own retry logic (`maxRetries`, `retrySettings`) is bypassed entirely because the process dies before `pollQueryStatus` can report failure and `onComplete` can increment the retry count. The manager records the row's lock-expiry, another worker picks it up, hits the same fast-failing condition, dies again — a transient blip becomes a crash loop until the operator intervenes.

  This utility is shared by M1 (`createRootSpansFromTraces`), M2 (`rewriteObservationsToPidTidSorting`), and M3 (`backfillEventsFullFromObservations`). All three are gated behind `LANGFUSE_BACKGROUND_MIGRATION_V4_ENABLE_HISTORIC_BACKFILL` and ship dormant — so this does not affect live traffic today — but it lies directly on the documented V4 historic-backfill activation path. Every self-hoster who flips that gate is exposed.

  ### Step-by-step proof

  1. Operator sets `LANGFUSE_BACKGROUND_MIGRATION_V4_ENABLE_HISTORIC_BACKFILL=true` and restarts the worker.
  2. `BackgroundMigrationManager` picks up M1. M1 enumerates partitions and calls `fireQuery({ query: ..., queryId: 'backfill-p-202604-...' })`.
  3. CH is misconfigured (or transiently unavailable). `commandClickhouse` issues the POST; the server returns an HTTP 401 within ~50ms. The CH client's promise rejects.
  4. `fireQuery` is awaiting `sleep(5000)` on line 312. The rejection lands with no handler attached.
  5. End of the microtask: Node emits `unhandledRejection`. In worker-mode, Node 24's default throws and the process exits with code 1. In CLI-mode, the main() handler logs and calls `process.exit(1)` directly.
  6. The migration row's `lockedAt` was set; `heartBeat` stops. After the lock expiry, another worker picks the row up. M1 re-runs from scratch, hits the same 401, crashes again. Crash loop.
  7. The retry-count machinery in `onComplete` and the per-todo `retrySettings` are never consulted because the process dies before `pollQueryStatus` reports failure.

  ### Fix

  Single-block move on line 310 — attach the .catch immediately after creating queryPromise, before any await:

  ```ts
  const queryPromise = commandClickhouse({ ... });
  queryPromise.catch((err) => {
    if (err?.name === 'AbortError' || err?.message?.includes('aborted')) {
      logger.debug(`${logPrefix} Query ${queryId} HTTP connection aborted as expected`);
    } else {
      logger.info(`${logPrefix} Query ${queryId} promise rejected: ${err?.message}`, err);
    }
  });

  await sleep(initialWaitMs);
  // ... rest unchanged, but remove the duplicate queryPromise.catch(...) on line 336
  ```

  The handler attached pre-await captures rejections during the pre-abort window so they are handled rather than unhandled. The AbortError suppression still works on the happy path because aborts produced after this point reject the same promise, and the registered handler catches them too. This matches the reviewer's intuition in PR comment 3341242018: 'AFAIK .catch should be set before aborting'.

  </details>
- 🔴 `worker/src/backgroundMigrations/utils/backfillBase.ts:50-79` — `scheduleNext` is invoked from two independent async contexts: the slot-fill loop at the end of each migration's `run()` (e.g. `backfillEventsFullFromObservations.ts:572-575`, `createRootSpansFromTraces.ts:537-540`, `rewriteObservationsToPidTidSorting.ts:553-556`) and the setInterval callback installed by `ConcurrentQueryManager.startPolling` at `utils/backfillBase.ts:50-100`. The `isPolling` flag on lines 51-52,98 only serializes interval ticks against each other — it does not exclude the externally invoked `scheduleNext` from the slot-fill loop, so the two callers can interleave at every `await loadState() → mutate state.todos[idx] → updateState() → fireQuery` step. The result is that the same first-pending todo gets picked twice, two queryIds are issued, the last writer wins in DB (the loser's queryId becomes an orphan untracked in `state.activeQueries`/`manager.activeQueries`), both `fireQuery` calls launch long-running CH INSERTs for the same source part, and the effective concurrency briefly exceeds the configured cap. ReplacingMergeTree dedup hides the data divergence, but operators see doubled CH compute and lose JS-side observability on the orphan; if the worker dies, recovery cannot reattach to it. Maintainer flagged this in comments 3343202252/3343224126 across all three migrations. Fix: move all scheduling into the polling loop (drop the slot-fill prefill), or hold a mutex around the `loadState→updateState` block so the two callers serialize.

  <details><summary>Extended reasoning...</summary>

  ### What the bug is

  `ConcurrentQueryManager.startPolling` at `worker/src/backgroundMigrations/utils/backfillBase.ts:38-100` installs a `setInterval` whose callback iterates `this.activeQueries`, awaits `pollQueryStatus`, and on completion/failure invokes `await onComplete(todo, ...)` followed by `await scheduleNext()` (lines 62-79). Each migration's `run()` *also* invokes `scheduleNext` directly from a slot-fill `for (let i = 0; i < slotsAvailable; i++) { await scheduleNext(); }` loop after the manager is already polling:

  - `backfillEventsFullFromObservations.ts:572-575`
  - `createRootSpansFromTraces.ts:537-540`
  - `rewriteObservationsToPidTidSorting.ts:553-556`

  The only serialization inside the manager is the `isPolling` flag (`backfillBase.ts:51-52` / `:98`): it stops re-entrant interval ticks but does nothing about the externally-invoked `scheduleNext` running in parallel from the slot-fill loop.

  ### The race

  Each `scheduleNext` does, in order:

  1. `state = await this.loadState()` — Postgres round-trip; suspends.
  2. Pick `state.todos.filter(t => t.status === "pending")[0]` synchronously.
  3. Mutate `state.todos[idx].status = "in_progress"` and assign a new `queryId` synchronously.
  4. `state.activeQueries.push(queryId)`.
  5. `await this.updateState(state)` — Postgres round-trip; suspends.
  6. `await fireQuery({ query, queryId, ... })` — sleeps 5-20s by design (`fireQuery` uses `await sleep(initialWaitMs)` at `backfillBase.ts:266`).

  The yields at (1), (5), and (6) interleave the slot-fill caller (A) with the timer-driven caller (B). Both call `loadState()` against the pre-update row, both pick `pendingTodos[0]` (same todo), both compute new queryIds, both `updateState()`. The DB row ends up with B's queryId (last writer wins), A's queryId is now an orphan in `manager.activeQueries` not referenced by any persisted state, and **both** call `fireQuery` — two long-running CH INSERTs targeting the same source part.

  ### Impact

  1. **Doubled CH compute** for the contested partition/part.
  2. **Concurrency cap breach.** `manager.addQuery` happens after `fireQuery` returns, so for the duration of the race, `manager.activeCount` transiently exceeds `config.concurrency` (operator-configurable up to 8+ via the CLI / `background_migrations.args`).
  3. **Orphan queryId.** A's queryId is in `manager.activeQueries` but is not the queryId persisted in `state.todos[idx]`. If the worker dies, recovery reads the persisted queryId only — the orphan CH query continues server-side untracked, and its completion will never be reflected in any tracking surface.
  4. **Tracking inconsistency.** `state.todos[idx].queryId` retains only the last writer's value, so the per-todo trail of "which queryId did the work" is wrong for ~half the contested cases.

  ReplacingMergeTree on `events_full` (keyed on `(span_id, start_time)` with `event_ts` as the version column) deduplicates the doubled rows, so this is **not** a data-correctness bug. The operator-facing impact is wasted CH compute, a brief over-concurrency window, and lost observability on the orphan query.

  ### Trigger conditions and severity

  - Default `concurrency=1` makes the slot-fill loop a single iteration that completes in <30s — typically before the first poll tick at `pollIntervalMs=30_000` — so the race rarely fires in the default config.
  - The race fires reliably when an operator bumps `--concurrency` past 1 (a documented and PR-exposed knob via `background_migrations.args` JSONB and CLI). With `concurrency=8`, slot-fill takes ~40s+ due to `fireQuery`'s 5s sleep per iteration, overlapping with the first poll tick.
  - It also fires whenever the recovery path adds `stillRunningTodos` to `manager.activeQueries` before the slot-fill loop runs (lines 565-571 of `backfillEventsFullFromObservations.ts`): if any recovered query completes within the first poll window, the timer-driven `scheduleNext` fires while the slot-fill loop is still iterating.

  Filed as **normal**: it's a structural correctness flaw the maintainer explicitly flagged, repeated across all three migrations, sitting on the documented V4 rollout path; the data-impact is muted by ReplacingMergeTree dedup but the operator-facing CH-load doubling and orphan-query observability gap are real.

  ### Step-by-step proof

  Take `backfillEventsFullFromObservations.ts` with `concurrency=4`:

  1. `manager.startPolling(30_000, onComplete, scheduleNext)` arms `setInterval` (line 569).
  2. Recovery adds two `stillRunningTodos` to `manager.activeQueries` (lines 565-571). Slot-fill computes `slotsAvailable = 4 − 2 = 2`.
  3. Slot-fill iteration 1 (A) enters `scheduleNext`. A awaits `loadState()` → reads `state` v1, picks `todos[0]`, mutates to `in_progress` + assigns `queryId_A`, awaits `updateState()` → writes `state` v2 to DB, awaits `fireQuery` (sleeps 5s).
  4. At T+30s, `setInterval` fires. `isPolling=false`, so the callback enters. It iterates `manager.activeQueries`. One of the two recovered queries has completed within the window — common for short partitions. `pollQueryStatus` returns `completed`. The callback awaits `onComplete(recoveredTodo, true)`, then awaits `scheduleNext` (B) at line 67.
  5. B awaits `loadState()` → reads `state` v2 (with `todos[0]` already `in_progress` and `queryId_A`). B's `pendingTodos = state.todos.filter(t => t.status === "pending")` correctly skips `todos[0]` and picks `todos[1]`. Good so far.

     But the race is not at iteration 1 — the race is at the *next* slot-fill iteration:

  6. A's `fireQuery` returns, A's `scheduleNext` returns, slot-fill loop advances to iteration 2 (A'). A' awaits `loadState()` → reads `state` v3 (which now contains B's write for `todos[1]`).
  7. Meanwhile B's `fireQuery` is still in its initial 5s sleep. Another poll tick fires at T+60s; another recovered query has completed in the interim. Callback awaits `onComplete`, then awaits `scheduleNext` (B').
  8. A' has already loaded state v3 and is between (1) and (5). A' picks `todos[2]` (first pending in v3), mutates in memory, but has not yet `updateState()`'d.
  9. B' enters `scheduleNext`, awaits `loadState()` → reads the same v3 (A' has not committed yet). B' also picks `todos[2]`, mutates in memory with a different `queryId_B'`.
  10. A' awaits `updateState()` → writes v4 with `todos[2].queryId = queryId_A'`. A' awaits `fireQuery(queryId_A')` → CH INSERT #1 starts.
  11. B' awaits `updateState()` → writes v5 with `todos[2].queryId = queryId_B'` (last writer wins, A's queryId is now lost from the DB). B' awaits `fireQuery(queryId_B')` → CH INSERT #2 starts for the **same source part**.
  12. Both INSERTs run concurrently. `manager.addQuery` is called for both, so `manager.activeQueries` now holds `queryId_A'` and `queryId_B'`. `manager.activeCount` transiently exceeds `config.concurrency=4` until one finishes.
  13. ReplacingMergeTree on `events_full` deduplicates the doubled rows on next merge. Both inserts complete; `onComplete` fires for each. Only `queryId_B'` is in `state.todos[2].queryId`, so when `onComplete(todoA', true)` runs, `state.activeQueries.filter(q => q !== queryId_A')` is a no-op for the persisted side.
  14. If the worker dies between steps (11) and (13), the persisted state has only `queryId_B'`; `queryId_A'` keeps running on CH server-side until completion, untracked anywhere.

  ### Why existing safeguards don't catch it

  - The `isPolling` flag is the only mutual-exclusion primitive in the manager and only excludes interval re-entry, not external callers.
  - `state.activeQueries.push(queryId)` happens before `updateState`, so the array briefly contains the loser's queryId in the in-memory state, but the persisted state never reflects it after the winner overwrites.
  - `manager.addQuery` happens after `fireQuery`'s 5s sleep returns — long after both racers have already issued their CH queries — so concurrency tracking can't preempt the race.
  - The post-INSERT `verifyPartStillActive` check at `backfillEventsFullFromObservations.ts:506-507` only confirms the part is still in `system.parts`; it does not detect that two INSERTs for that part ran.

  ### Fix

  The maintainer explicitly proposed two equivalent options at comment 3343202252:

  1. **Drop the slot-fill prefill** and let the polling loop fill the concurrency budget. Replace the slot-fill loop with a single `await scheduleNext()` (or remove the loop entirely and lower the poll interval), so the only caller is the setInterval callback. The `isPolling` flag then provides the necessary serialization.
  2. **Hold a mutex around the `loadState → updateState` block** so the two callers serialize at the critical section. A simple `async-mutex` (or a single `Promise` chained off `this._writeLock`) on the manager would work.

  Option (1) is structurally simpler and matches the maintainer's preferred direction; it also closes the race without adding new primitives. The same fix needs to land in all three migrations or be hoisted into `backfillBase.ts`'s helper surface.

  </details>
```

PR #14018, opened 2026-06-03
refactor(agent): Switch agent framework

bezbac:
Resolves LFE-10113

This PR replaces the `ClaudeAgentAdapter` (backed by `@ag-ui/claude-agent-sdk`) with a `MastraAgent` (backed by `@ag-ui/mastra` + `@mastra/core/agent` + `@mastra/mcp`), routing LLM calls through the AI SDK's Amazon Bedrock provider. Conversation continuity no longer relies on a pod-local Claude session ID; instead the server reconstructs full history from the DB on every request and injects it into the agent input.

- **Framework swap**: `ClaudeAgentAdapter` removed; `MastraAgent` + `MCPClient` created per-run in an async `createMastraAdapter` helper, with proper cleanup in all termination paths (complete, abort, cancel, error).
- **History replay**: `getConversationMessages` is called before the stream starts and the result is prepended to the current-turn input via `withConversationHistory`, eliminating the session-ID resume path.
- **Event persistence extended**: `MESSAGES_SNAPSHOT` and `TEXT_MESSAGE_CHUNK` events are now persisted and replayed by `createConversationMessageAccumulator`; `MESSAGES_SNAPSHOT` is added to `shouldFlushPersistedEvent` so state is written to the DB at each snapshot.
- **New required env var**: `LANGFUSE_IN_APP_AGENT_BEDROCK_MODEL_ID` must be set; the handler returns 412 if it is missing.
</details>

Safe to merge; all termination paths correctly unsubscribe, clean up the MCPClient, and finish the DB run record.

The core stream lifecycle logic is sound — async adapter creation races are guarded, `cleanupAdapter` is called in `finish()` on every exit path, and history reconstruction via `upsertMessage` is idempotent. The two findings are non-blocking: the hardcoded model string and the speculative snapshot storage growth.

web/src/features/in-app-agent/server/handler.ts and web/src/features/in-app-agent/server/persistence.ts
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/in-app-agent/server/handler.ts:184-193
Now that `bedrockModelId` is resolved from the environment and validated earlier in this same function, the TODO on this line can be resolved. Storing `"haiku"` means the `inAppAgentRun` row records a generic alias instead of the real model ID (e.g., `us.anthropic.claude-3-haiku-20240307-v1:0`), which reduces auditability and accuracy of stored run metadata.

```suggestion
          await createRun({
            prisma,
            runId: currentTurnInput.runId,
            projectId,
            conversationId: conversation.id,
            triggeredByUserId: userId,
            model: bedrockModelId,
            mcpApiKeyId: mcpApiKey.id,
          });
```

### Issue 2 of 2
web/src/features/in-app-agent/server/persistence.ts:330-343
**MESSAGES_SNAPSHOT may grow quadratically across runs**

If Mastra's `MastraAgent` emits a full-conversation snapshot (all messages up to the current turn), each run's persisted `MESSAGES_SNAPSHOT` event will contain the entire history repeated. The `compactEvents` call inside `replaceRunEvents` operates per-run — it cannot consolidate snapshots across runs — so the DB would accumulate O(n²) total message data over n turns. For long conversations this could become a significant storage issue.

Consider filtering or truncating `MESSAGES_SNAPSHOT` messages to only assistant messages produced in the current run before persisting, since prior turns are already captured in their own `RUN_STARTED` / `MESSAGES_SNAPSHOT` events and are reconstructed correctly by `upsertMessage`.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(agent): Switch agent framework"](https://github.com/langfuse/langfuse/commit/81531c1ff3873edf09476895db899038f709cfe1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35636349)</sub>

### CODE REVIEW

File: web/src/features/in-app-agent/server/persistence.ts [hunk 1]
```diff
@@ -309,6 +302,17 @@ export function getMessagesFromEvents(events: readonly AgUiEvent[]) {
...
 
+function sanitizeConversationMessagesForReplay(
+  messages: readonly AgUiMessage[],
+): readonly AgUiMessage[] {
+  const messagesWithoutOrphanToolCalls =
+    dropUnpairedAssistantToolCalls(messages);
+  const messagesWithoutEmptyAssistants = dropEmptyAssistantMessages(
+    messagesWithoutOrphanToolCalls,
+  );
+  return dropUnansweredUserMessages(messagesWithoutEmptyAssistants);
```
nimarb: why do we drop user messages?

web/src/features/in-app-agent/server/persistence.ts
```diff
@@ -641,6 +691,110 @@ function mergeMessages(existing: AgUiMessage, next: AgUiMessage): AgUiMessage {
...
+          (getString(event, "delta") ?? ""),
+      };
+      continue;
+    }
+
+    compactedEvents.push(event);
+  }
+
+  return compactedEvents;
+}
```
nimarb, line 721: why do we need to write our own additional compactor here? i think we can just use the compactEvents by AG-UI

pnpm-workspace.yaml
```diff
@@ -39,7 +42,14 @@ overrides:
...
   ip-address@10.1.0: 10.2.0
-  "@anthropic-ai/claude-agent-sdk>@anthropic-ai/sdk": 0.95.1
+  "@ai-sdk/provider-utils@<=4.0.26": 4.0.27
+  "@ai-sdk/provider-utils-v5": "npm:@ai-sdk/provider-utils@4.0.27"
+  "@copilotkit/runtime@0.0.0-mme-ag-ui-0-0-46-20260227141603>@ag-ui/langgraph": 0.0.29
+  "@copilotkit/shared@0.0.0-mme-ag-ui-0-0-46-20260227141603>uuid": 11.1.1
+  "@langchain/langgraph@1.2.7>uuid": 11.1.1
+  "@langchain/langgraph-checkpoint@1.0.1>uuid": 11.1.1
+  "@langchain/langgraph-sdk@1.8.7>uuid": 13.0.2
+  "@langchain/langgraph-sdk@1.9.10>uuid": 13.0.2
```
nimarb: why do we need uuid overrides here

web/src/features/in-app-agent/server/persistence.ts
bezbac: From my understanding `compactEvents` does not work on `TEXT_MESSAGE_CHUNK`.  I had a look at their source code right now: https://github.com/ag-ui-protocol/ag-ui/blob/9f93bba2852f8d8b7ae5f4a0d40a46199ab00e30/sdks/typescript/packages/client/src/compact/compact.ts#L27

nimarb: do we need to use the chunks?

bezbac: @ag-ui/mastra only emits `TEXT_MESSAGE_CHUNK` and not `TEXT_MESSAGE_CONTENT`.

See:
- https://github.com/search?q=repo%3Aag-ui-protocol%2Fag-ui+path%3A%2F%5Eintegrations%5C%2Fmastra%5C%2Ftypescript%5C%2F%2F+TEXT_MESSAGE_CONTENT&type=code
- https://github.com/ag-ui-protocol/ag-ui/blob/604ca57a5e9f684415086eaa2b7be3c9ae8a5390/integrations/mastra/typescript/src/__tests__/integration.test.ts#L59

nimarb (approved):

PR #14036, opened 2026-06-04
fix(ui): sessions view gets into infinite rerendering loop while translated with google translate

nkabardin:
Virtualized lists go crazy while the page is translated by google translate. We rerender the items all the time due to recreated unstable props. While it might be possible to patch it with memoization, we need a proper stable long-term solution. The approach is described in agentic documentation in #14039 

The proper, we believe, is in the deeper data flow rework. The problem: the virtualized session item rendering components are too smart, doing too much. And when external plug-in (google translate) changes things, rerendering just can't stop redoing itself. This PR implements an architectural step towards making it rerender less and behave better with google translate. It also integrates zustand as this is a decision we've taken as a team — we'll not introduce the big global storage of state, but we'll have local state managed in zustand stores for big features.

This PR introduces a Zustand-backed store and a custom stable measurement state machine to break the infinite re-render loop that occurred when Google Translate mutated the DOM while virtualized lists were actively measuring element sizes.

- A new `sessionDetailStore` (zustand vanilla store) centralises `loadedTraceIds` and `showCorrections`, replacing per-component `useState` so lazy-load decisions survive cross-component mutations.
- `useStableVirtualRowMeasurement` replaces `virtualizer.measureElement` with a ResizeObserver that feeds a debounced oscillation-clamping state machine (`stableVirtualRowMeasurementState`), preventing the virtualizer from issuing infinite resize cascades when translated text causes repeated layout thrashing.
- `LazyTraceRow` / `LazySessionTraceEventsRow` are refactored to consume state from the store, decoupling render decisions from props that Google Translate could destabilise.
</details>

Safe to merge; no functional regressions identified. The Zustand store integration and stable measurement state machine address the root cause correctly, and test coverage for both new modules is solid.

The core measurement logic is well-reasoned and thoroughly tested. The two edge cases flagged — frozen state never releasing when height stabilises at the frozen minimum, and the missing explicit cleanup in the itemKeyChanged effect branch — are latent rather than immediately harmful. The overscan increase from 1 to 10 is the most likely source of a noticeable regression on large sessions with many traces.

stableVirtualRowMeasurementState.ts (freeze-release completeness), useStableVirtualRowMeasurement.ts (itemKeyChanged cleanup), and the overscan value in index.tsx (both SessionPage and SessionEventsPage).
</details>

```mermaid
sequenceDiagram
    participant GT as Google Translate
    participant SVR as SessionVirtualizedRow
    participant USVR as useStableVirtualRowMeasurement
    participant SMS as stableVirtualRowMeasurementState
    participant VZ as TanStack Virtualizer

    GT->>SVR: mutates DOM (resize)
    SVR->>USVR: ResizeObserver fires
    USVR->>USVR: scheduleHeightCommit (rAF)
    USVR->>SMS: commitHeight(newHeight, now)
    SMS->>SMS: detect oscillation pair
    alt "oscillationCount < maxOscillationCount"
        SMS-->>USVR: return newHeight
        USVR->>VZ: resizeItem(index, newHeight)
    else "oscillationCount >= maxOscillationCount (freeze)"
        SMS-->>USVR: return null (clamped)
        USVR->>USVR: no resize, freeze active
    end

    participant SDS as SessionDetailStore (Zustand)
    participant LTR as LazyTraceRow / LazySessionTraceEventsRow
    participant IO as IntersectionObserver

    IO->>LTR: element enters viewport
    LTR->>SDS: markTraceLoaded(traceId)
    SDS-->>LTR: "loadedTraceIds[traceId] = true"
    LTR->>LTR: swap skeleton to real content
    USVR->>USVR: scheduleHeightCommit on resize
```
</details>

`````markdown

---

### Issue 1 of 4
web/src/components/session/LazySessionTraceEventsRow.tsx:133-158
**Duplicated `observe()` helper across three files**

The identical `listeners` map, `sharedObserver` singleton, and `observe()`/cleanup pattern appear verbatim in `TraceRow.tsx`, `TraceEventsRow.tsx`, and this file. Each module gets its own `IntersectionObserver` instance observing the same root, which multiplies the observer count when both lazy-load variants are on screen simultaneously (e.g., when `SessionAnnotationProcessor` is open alongside the session page). Consider extracting this into a shared utility like `sharedIntersectionObserve.ts` so all three consumers share a single observer.

### Issue 2 of 4
web/src/components/session/index.tsx:310-314
**`overscan: 10` may over-render on large sessions**

Raising overscan from 1 to 10 keeps up to 20 extra rows in the DOM at once. For projects with sessions that have hundreds of traces this means the IntersectionObserver `rootMargin: "400px"` pre-load zone fires for many more rows simultaneously, which could cause a burst of concurrent data fetches and increased memory pressure. The same change was made in `SessionEventsPage`. Consider aligning this with a more conservative value (e.g. 3–5) or documenting the rationale, since overscan=10 is unusually high relative to TanStack Virtual's default of 3.

### Issue 3 of 4
web/src/components/session/useStableVirtualRowMeasurement.ts:80-116
**`itemKeyChanged` branch schedules work without a returned cleanup**

When `itemKeyChanged` is `true`, the effect calls `scheduleHeightCommit` (which registers a `requestAnimationFrame`) and then `return`s without providing a cleanup function. If the component unmounts before that frame fires, the frame is reliably cancelled by the ref callback receiving `null` → `cancelScheduledWork()`. That path is safe today, but it relies on React's ref-callback execution happening before effects cleanup runs on unmount. Returning `cancelScheduledWork` here would make the invariant explicit and guard against any future refactoring that changes the ref callback timing.

### Issue 4 of 4
web/src/components/session/stableVirtualRowMeasurementState.ts:129-139
**Frozen state is never released when height stabilises at or above `frozenMinHeight`**

The freeze-release check requires `roundedHeight < state.frozenMinHeight`. If, after the oscillation window expires, the element settles at exactly `frozenMinHeight` (or higher), the condition is never true and `oscillationPair`, `oscillationCount`, and `frozenMinHeight` remain set indefinitely. The stale frozen state doesn't cause incorrect sizing (since `Math.max(frozenMinHeight, newHeight)` is a no-op when `newHeight >= frozenMinHeight`), but it means the oscillation detector is never truly reset once oscillation stabilises at the frozen value. A separate cleanup path that clears the frozen state when the window has elapsed regardless of direction would avoid this.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(sessions): release stale virtual row..."](https://github.com/langfuse/langfuse/commit/79bfc681acf70a463c5b3841af1c4bc8478b285c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35836967)</sub>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

web/src/components/session/actions/downloadSessionAsJson.ts
nimarb, line 1: i think this functionality is entirely broken / not available? can you confirm?

web/src/components/session/stableVirtualRowMeasurementState.clienttest.ts
nimarb, line 1: are all tests here absolutely essential? each test increases CI runtime a slight bit so we want essential tests only. i think we could maybe merge the "clamps the observation that triggers..." and "keeps a frozen shrink clamped while active" tests

nimarb (approved):

web/src/components/session/actions/downloadSessionAsJson.ts
nkabardin: it still works, in "legacy" version of session view. doesn't exist in v4 though. 
but it works, so, I'd keep it

Screenshot 2026-06-10 at 10 31 55

web/src/components/session/stableVirtualRowMeasurementState.clienttest.ts
nkabardin: actually I'd keep those tests. Each of the test was added to cover specific problem I've captured in debugging and reviewing. The measurement code is pretty hard to understand, and it's OK, it could stay a black box of browser-specific magic. Those things are never too readable. But to make sure we don't break it's intent in the future, this test is quite useful

PR #14040, opened 2026-06-04
feat(batch-export): read sessions from events table for v4 users

niklassemmler:
## Summary

- Branch the **sessions** batch export on a new `useEventsTable` flag so v4 beta users export session data from the ClickHouse events table instead of the legacy traces path, keeping exports consistent with what those users see in the UI.
- The flag is snapshotted from the user's v4 beta flag at dispatch and persisted in the job query (`BatchExportQuerySchema`), so the worker reads the dispatch-time snapshot rather than the live user record — no flag-flip race between dispatch and pickup.
- The events path mirrors the Sessions UI flow: list rows via `getSessionsTableFromEvents` (filter/order/pagination), then metrics (duration, cost, usage) via `getSessionMetricsFromEvents`, merged by session id. Both readers expose the same field names, so the row mapping is shared with the legacy path.
- Legacy `getSessionsWithMetrics` path is unchanged when the flag is false/absent (self-hosted stays on legacy, since the beta flag is Cloud-only).
- The events-table **observations** export already routes through `getEventsStream`; added a regression test that locks that route in.

## Impacted packages

- `packages/shared` — `useEventsTable?: boolean` added to `BatchExportQuerySchema`.
- `web` — `batchExport.create` dispatcher snapshots `v4BetaEnabled` into the persisted query.
- `worker` — `getDatabaseReadStream` sessions case branches to the events-table readers when the flag is set.

## Verification

- `pnpm --filter @langfuse/shared --filter worker --filter web run typecheck` — green
- `pnpm --filter @langfuse/shared --filter worker --filter web run lint` — green
- Added worker integration tests (sessions-from-events + events-export route guard) and web dispatch-snapshot tests; these need live infra and run in CI.
- Browser review of the live export flows pending.

This PR adds a `useEventsTable` flag to the batch-export sessions path so that v4-beta users read session data from the ClickHouse events table instead of the legacy traces path. The flag is snapshotted from the user's session at dispatch time, persisted in the job's query column, and passed through to the worker's paginated stream reader.

- `BatchExportQuerySchema` gains an optional `useEventsTable` boolean; the tRPC mutation stamps the live `v4BetaEnabled` value onto the query object before persisting it, preventing the worker from ever reading the stale/live user record.
- The paginated reader's `sessions` case now forks on `useEventsTable`: the events-table path calls `getSessionsTableFromEvents` for the paginated list and `getSessionMetricsFromEvents` for metrics, then joins them by `session_id`; the legacy path is unchanged.
- New server-test cases verify that the flag is correctly snapshotted for both enabled and disabled users; a new worker test verifies that the events-table stream returns the expected sessions with correct `countTraces`.
</details>

The change is safe for non-sessions exports and for users without the v4 beta flag; the sessions export via the events table is functionally new, and the two-query join does not forward the export cutoff timestamp to the metrics call.

The core dispatch-time snapshot mechanism is solid and well-tested. The concern is in the worker's new sessions branch: getSessionMetricsFromEvents is called without the cutoffCreatedAt upper bound that the list query carries through sessionsFilter. This means trace_count, duration, totalCost, and related fields in the export can reflect events ingested after the job was enqueued, while the legacy path forwards the cutoff all the way into its metrics calculation.

worker/src/features/database-read-stream/getDatabaseReadStream.ts — specifically the useEventsTable branch in the sessions case, where the metrics query is missing the cutoff filter.
</details>

```mermaid
sequenceDiagram
    participant Client
    participant tRPC as batchExportRouter (web)
    participant DB as Prisma / Postgres
    participant Queue as BatchExportQueue
    participant Worker as handleBatchExportJob
    participant CH_List as ClickHouse (getSessionsTableFromEvents)
    participant CH_Metrics as ClickHouse (getSessionMetricsFromEvents)

    Client->>tRPC: create sessions export
    tRPC->>tRPC: stamp useEventsTable from v4BetaEnabled
    tRPC->>DB: persist batchExport with snapshotted query
    tRPC->>Queue: enqueue BatchExportJob
    Queue->>Worker: dequeue job
    Worker->>DB: fetch job reads persisted useEventsTable
    alt useEventsTable true
        Worker->>CH_List: getSessionsTableFromEvents with sessionsFilter and page
        CH_List-->>Worker: list of session_id and trace_ids
        Worker->>CH_Metrics: getSessionMetricsFromEvents with sessionIds only
        CH_Metrics-->>Worker: metrics with cost and duration
        Worker->>Worker: join by session_id and drop unmatched
    else useEventsTable false or undefined
        Worker->>CH_List: getSessionsWithMetrics with sessionsFilter
        CH_List-->>Worker: sessions with metrics
    end
    Worker->>DB: traceSession findMany for bookmarked and public
    Worker-->>Client: export rows
```
</details>

`````markdown

---

### Issue 1 of 2
worker/src/features/database-read-stream/getDatabaseReadStream.ts:232-236
**Metrics query does not inherit the export cutoff upper bound**

`getSessionMetricsFromEvents` is called with only `projectId` and `sessionIds` – the `cutoffCreatedAt` upper bound applied to the list query (via `createdAtCutoffFilter` in `sessionsFilter`) is not forwarded. As a result, the aggregated metrics for a session (trace count, duration, total cost, etc.) can reflect events ingested *after* the export was queued, diverging from the snapshot semantics that the cutoff is supposed to enforce. The legacy path passes `sessionsFilter` (which carries the cutoff) all the way into `getSessionsWithMetrics`, so the two paths behave differently.

### Issue 2 of 2
worker/src/features/database-read-stream/getDatabaseReadStream.ts:238-241
**Silent session drop without observability**

Sessions that appear in the paginated list but lack a corresponding entry in `metricsById` are silently filtered out. The test comment even documents this ("sessions without metrics are dropped"), but there is no `logger.warn` or counter emitted when this happens. In production, if a ClickHouse timing gap or aggregation edge case causes the metrics call to miss a session, the export will simply produce fewer rows than the list query returned, with no trace in logs or metrics to surface the discrepancy. Adding at least a debug-level log for the dropped count would make the behavior observable.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(batch-export): read sessions from e..."](https://github.com/langfuse/langfuse/commit/87f30be6fb5ff06353dcac81ef224e3fe31bfd5d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35685445)</sub>

### DIFF
--- a/packages/shared/src/server/services/sessions-ui-table-events-service.ts
+++ b/packages/shared/src/server/services/sessions-ui-table-events-service.ts
@@ -175,7 +175,7 @@ export const getSessionsWithMetricsFromEvents = async (props: {
       limit: props.limit,
       page: props.page,
       clickhouseConfigs: props.clickhouseConfigs,
-      tags: { kind: "metrics" },
+      tags: { kind: "analytic_events" },
     },
   );
 
--- a/packages/shared/src/server/services/sessions-ui-table-events-service.ts
+++ b/packages/shared/src/server/services/sessions-ui-table-events-service.ts
@@ -9,6 +9,7 @@ import {
   FilterList,
   StringOptionsFilter,
   orderByToClickhouseSql,
+  type SessionEventsMetricsRow,
 } from "../queries";
 import { createFilterFromFilterState } from "../queries/clickhouse-sql/factory";
 import {
@@ -150,6 +151,38 @@ export const getSessionsTableFromEvents = async (props: {
   return rows.map((row) => ({
     ...row,
     trace_count: Number(row.trace_count),
+  }));
+};
+
+// Single-query equivalent of getSessionsTableFromEvents + getSessionMetricsFromEvents,
+// mirroring the legacy getSessionsWithMetrics. Used by batch export so the metrics
+// aggregation inherits the same filter (incl. the createdAt cutoff) and clickhouseConfigs
+// (extended HTTP timeouts) as the row query, in a single round-trip.
+export const getSessionsWithMetricsFromEvents = async (props: {
+  projectId: string;
+  filter: FilterState;
+  orderBy?: OrderByState;
+  limit?: number;
+  page?: number;
+  clickhouseConfigs?: ClickHouseClientConfigOptions | undefined;
+}) => {
+  const rows = await getSessionsTableFromEventsGeneric<SessionEventsMetricsRow>(
+    {
+      select: "metrics",
+      projectId: props.projectId,
+      filter: props.filter,
+      orderBy: props.orderBy,
+      limit: props.limit,
+      page: props.page,
+      clickhouseConfigs: props.clickhouseConfigs,
+      tags: { kind: "metrics" },
+    },
+  );
+
+  return rows.map((row) => ({
+    ...row,
+    trace_count: Number(row.trace_count),
+    total_observations: Number(row.total_observations),
   }));
 };
 
--- a/web/src/features/batch-exports/server/batchExport.ts
+++ b/web/src/features/batch-exports/server/batchExport.ts
@@ -32,7 +32,15 @@ export const batchExportRouter = createTRPCRouter({
           scope: "batchExports:create",
         });
 
-        const { projectId, query, format, name } = input;
+        const { projectId, format, name } = input;
+
+        // Snapshot the user's v4 beta flag into the persisted query so the
+        // worker reads events-aware data sources from the dispatch-time
+        // snapshot, never the live user record. Overrides any client-sent value.
+        const query = {
+          ...input.query,
+          useEventsTable: ctx.session.user.v4BetaEnabled ?? false,
+        };
 
         if (query.tableName === BatchExportTableName.AuditLogs) {
           throwIfNoEntitlement({
--- a/packages/shared/src/features/batchExport/types.ts
+++ b/packages/shared/src/features/batchExport/types.ts
@@ -54,6 +54,11 @@ export const BatchExportQuerySchema = z.object({
   orderBy,
   limit: z.number().optional(),
   page: z.number().optional(),
+  // Snapshotted at dispatch time from the user's v4 beta flag. When true, the
+  // sessions export reads from the ClickHouse events table instead of the
+  // legacy traces path. Persisted in the job's query column so the worker reads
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by niklassemmler** [comment]
* **#14040** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14040?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

File: web/src/features/batch-exports/server/batchExport.ts [hunk 1]
```diff
@@ -32,7 +32,15 @@ export const batchExportRouter = createTRPCRouter({
...
 
-        const { projectId, query, format, name } = input;
+        const { projectId, format, name } = input;
+
+        // Snapshot the user's v4 beta flag into the persisted query so the
+        // worker reads events-aware data sources from the dispatch-time
+        // snapshot, never the live user record. Overrides any client-sent value.
+        const query = {
+          ...input.query,
+          useEventsTable: ctx.session.user.v4BetaEnabled ?? false,
```
Steffen911, line 42: Do we need to do the same thing for the batchAction execution?

Steffen911 (commented):
Nice change and setup!

web/src/features/batch-exports/server/batchExport.ts
niklassemmler: Will be addressed in LFE-10264

Steffen911 (approved):

packages/shared/src/server/services/sessions-ui-table-events-service.ts
```diff
@@ -153,6 +154,38 @@ export const getSessionsTableFromEvents = async (props: {
...
+      limit: props.limit,
+      page: props.page,
+      clickhouseConfigs: props.clickhouseConfigs,
+      tags: { kind: "analytic_events" },
+    },
+  );
+
+  return rows.map((row) => ({
+    ...row,
+    trace_count: Number(row.trace_count),
```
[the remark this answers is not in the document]
niklassemmler: We created LFE-10268 to track this

PR #14043, opened 2026-06-04
feat(observations-v2): add subquery-IN rewrite behind env flag

niklassemmler:
## Summary

- Replaces the CTE+JOIN late-materialization split query for the v2 observations public API (`GET /api/public/v2/observations` + the two observations MCP tools) with a JOIN-free form: a single outer `SELECT` against `events_full` gated by an `(span_id, trace_id, start_time, project_id) IN (SELECT ...)` subquery that runs filter/sort/limit on the inner table. ClickHouse builds the inner hash set once and prunes the outer scan, avoiding JOIN materialization and double projection of join keys.
- Inner reads `events_core` (truncated, cheap) and escalates to `events_full` only when the existing `forceFullTable()` heuristic fires (FTS / substring filters). Outer always reads `events_full`, so input/output/metadata are returned full — matching today's behavior in this branch.
- Gated by a new env var `LANGFUSE_OBSERVATIONS_V2_SUBQUERY_REWRITE` (default `"false"`). Both code paths coexist; the flag is a fast-revert kill-switch because ClickHouse query plans can shift even when output is byte-identical. Flag retirement is a follow-up after one clean release cycle.
- Scope: the rewrite only replaces the `needsIOCTE` branch (io requested, or metadata with `expandMetadata`). The non-CTE simple path — and its truncated-metadata default from `events_core` — is untouched, so output stays byte-identical to the current path.
- The new query carries `log_comment = 'observations-v2-subquery-rewrite'` for query_log filtering during soak. No API contract change, no migration.

## Impacted packages

- `packages/shared`: new `buildEventsFullTableSubqueryQuery` and `shouldUseObservationsSubqueryRewrite()`; flag routing in `getObservationsV2FromEventsTableForPublicApi`; env var registration. Existing `buildEventsFullTableSplitQuery` left intact.
- `web`: env var registration (`env.mjs`); new servertests.

## Verification

- `pnpm db:generate` + shared build, then lint + typecheck clean for all changed files.
- Added servertests: (a) subquery and CTE paths return identical rows across field-group combos (both flag values via shared-env override); (b) canonical ORDER BY under the subquery path. Servertests need live ClickHouse and run in CI.
- Rollback is an env flip (`...=false`); no code rollback needed. Post-deploy perf is watched via ClickHouse query-time dashboards.

This PR introduces a `LANGFUSE_OBSERVATIONS_V2_SUBQUERY_REWRITE` kill-switch that, when enabled, replaces the existing CTE+JOIN split query on the `GET /api/public/v2/observations` endpoint with a single `IN (subquery)` pattern — the inner subquery filters and paginates on `events_core`, and the outer SELECT fetches full fields from `events_full` for only those rows.

- **New builder function** (`buildEventsFullTableSubqueryQuery`): emits one query where `events_full` is scanned only for the rows identified by the inner identity-tuple hash set, avoiding the JOIN overhead of the CTE split approach.
- **Flag gating** in `query-options.ts` and env schemas: defaults off, allowing instant rollback without a code deploy.
- **Test suite**: verifies byte-identical results vs. the CTE path across multiple field-group combinations, and validates canonical `ORDER BY` under the rewrite.
</details>

Safe to merge with the flag off (the default); the concern is in the new subquery path that only activates when the env flag is flipped on.

The outer ORDER BY is a hardcoded string literal that must stay byte-identical to what `EventsQueryBuilder.orderByColumns(orderByForObservationsQuery())` produces. Currently they match, but any future change to the canonical sort (adding a tiebreak, renaming a function) would silently diverge the order seen by API consumers from what the cursor pagination machinery expects, potentially causing incorrect page boundaries under the new code path. Additionally, the `selectRaw` call that builds the inner identity-tuple SELECT appends rather than replaces, relying implicitly on no prior `selectFieldSet` having been called — a fragile contract not enforced by assertion. The flag-off default means production traffic is unaffected today, but both issues would surface the moment the flag is enabled.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts — the new `buildEventsFullTableSubqueryQuery` function, specifically the hardcoded ORDER BY and the `selectRaw` accumulation behavior.
</details>

```mermaid
sequenceDiagram
    participant Client
    participant API as "GET /v2/observations"
    participant Repo as "getObservationsV2"
    participant CH as ClickHouse

    Client->>API: "GET /v2/observations?fields=io,..."
    API->>Repo: "opts (fields, filters, cursor)"
    Repo->>Repo: "buildObservationsQueryComponents"
    Repo->>Repo: "shouldUseObservationsSubqueryRewrite()?"

    alt "SUBQUERY_REWRITE=true AND needsIOCTE"
        Repo->>Repo: "buildEventsFullTableSubqueryQuery"
        Note over Repo: "Inner: SELECT identity-tuple FROM events_core LIMIT n+1"
        Note over Repo: "Outer: SELECT full fields FROM events_full WHERE tuple IN (inner)"
        Repo->>CH: "Single subquery-IN query"
        CH-->>Repo: "Full rows"
    else "needsIOCTE (CTE+JOIN)"
        Repo->>Repo: "buildEventsFullTableSplitQuery"
        Note over Repo: "base CTE on events_core + io CTE on events_full + JOIN"
        Repo->>CH: "CTE+JOIN query"
        CH-->>Repo: "Full rows"
    else "simple path"
        Repo->>CH: "Single query on events_core"
        CH-->>Repo: "Rows with truncated I/O"
    end

    Repo->>Repo: "enrichObservationsWithModelData"
    Repo-->>API: "observations array"
    API-->>Client: "JSON response"
```
</details>

`````markdown

---

### Issue 1 of 3
packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts:2041-2043
**Hardcoded ORDER BY drifts silently from the canonical sort**

The outer `outerOrderBy` string is manually written to match what `EventsQueryBuilder.orderByColumns(orderByForObservationsQuery("e"))` produces today. Nothing enforces that these two stay in sync. If `orderByForObservationsQuery` or the `orderByColumns` prefix-injection logic is ever updated (e.g., a new tiebreak column), the inner and outer sorts diverge: the inner controls which rows appear, but the outer determines the sort seen by API callers — a discrepancy would produce a response ordered differently from what the cursor relies on, silently breaking keyset pagination.

### Issue 2 of 3
packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts:2024-2026
**`selectRaw` appends rather than replaces — guard is implicit**

`selectRaw` pushes to `rawSelectExpressions` without clearing existing entries. The function relies entirely on the caller having skipped `selectFieldSet` (the `if (!useSubqueryRewrite)` branch). If a future code path calls `selectFieldSet` on the inner builder before it reaches here, the inner `SELECT` would emit both field-set expressions and the identity-tuple columns. ClickHouse's positional IN-tuple matching would then misalign if a non-identity column appears before position 4, silently corrupting which rows are matched by the outer `IN` clause. A defensive guard (e.g., asserting `selectFields.size === 0 && rawSelectExpressions.length === 0`) would make this coupling explicit.

### Issue 3 of 3
web/src/__tests__/server/observations-api-v2.servertest.ts:1787-1803
**No coverage for the `hasTraceFilter` path under the subquery rewrite**

All new test URLs filter by `traceId`, which is an `events_*` column and therefore does not trigger `hasTraceFilter`. When a filter on a trace-level column (e.g., `userId`, `tags`, `name`) is present, `buildObservationsQueryComponents` returns an external `traces` CTE and the inner builder includes `LEFT JOIN traces t`. In the subquery rewrite, that CTE lands in the outer `WITH` while the inner subquery still references `traces t` via JOIN. Whether ClickHouse resolves outer-scope CTEs inside an `IN` subquery is not exercised by any test here, leaving this corner case unverified.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(observations-v2): add subquery-IN r..."](https://github.com/langfuse/langfuse/commit/b6358008c37146cc7941758c13eb561e2d08400a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35691315)</sub>

### CODE REVIEW

**Comment by niklassemmler** [comment]
* **#14043** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14043?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

File: packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts [hunk 1]
```diff
@@ -1978,3 +1998,80 @@ export function buildEventsFullTableSplitQuery(opts: {
...
+
+  const tuple = `(${SUBQUERY_IDENTITY_TUPLE.join(", ")})`;
+  const queryParts: string[] = [];
+  queryParts.push(
+    `SELECT\n  ${outerSelect.join(",\n  ")}`,
+    "FROM events_full e",
+    "WHERE e.project_id = {projectId: String}",
+    `  AND ${tuple} IN (\n${innerQuery}\n)`,
+    outerOrderBy,
+    "SETTINGS log_comment = 'observations-v2-subquery-rewrite'",
```
sumerman: I am not sure if this is concatenated or overrides other setting and / or other `log_comment`.

packages/shared/src/server/repositories/events.ts
```diff
@@ -1500,19 +1502,38 @@ export const getObservationsV2FromEventsTableForPublicApi = async (
...
+  // The rewrite only replaces the CTE+JOIN split query (needsIOCTE branch);
+  // the non-CTE simple path is untouched. External CTEs are excluded: their
+  // resolution inside the IN subquery is unverified, and no v2 filter
+  // currently produces one, so that case keeps the CTE+JOIN path.
+  const useSubqueryRewrite =
+    needsIOCTE &&
+    externalCTEs.length === 0 &&
+    shouldUseObservationsSubqueryRewrite();
+
+  if (!useSubqueryRewrite) {
```
sumerman: nit: I don't feel good about subquery path and CTE path being torn in half. But I am not sure if there is a neater way write this.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
niklassemmler: addressed. I am now using query tags.

packages/shared/src/server/repositories/events.ts
niklassemmler: @sumerman  I have restructured it. Let me know if you like this better

sumerman (approved):
Nice, thank you!

PR #14061, opened 2026-06-05
fix: correct AM/PM period when toggling in time picker

### LINKED ISSUE #13687: bug: trace time picker swaps AM/PM in the time-frame preview
### Describe the bug

After selecting a start time in the traces time-frame picker (Fast Preview), the rendered preview shows the chosen time with AM and PM swapped (i.e. a 12-hour offset). The filter behaves accordingly:

- Picking **02:00 (2 AM)** → the preview displays **14:00**, and traces from earlier in the day are excluded.
- Picking **14:00 (2 PM)** → the preview displays **02:00**, and only traces from after 02:00 local time are shown.

So to see traces from today at around 14:00 local time, the start time has to be set to "2 PM" (which then renders as "02:00") — the opposite of what the labels indicate.

### Steps to reproduce

1. Have traces ingested today at around 14:00 local time.
2. Go to the Traces view (Fast Preview enabled).
3. Open the date/time picker and set the start time to "today 02:00 (AM)". Observe that the time-frame preview shows "14:00" and the traces table is empty.
4. Change the start time to "today 14:00 (PM)". Observe that the preview now shows "02:00" and the traces appear.

Expected: the preview should display the time exactly as it was picked, and the filter should include traces with timestamps after the picked instant.

Langfuse Cloud

N/A

N/A — UI bug.

### Additional information

The symmetric 12-hour swap (2 AM ↔ 14:00, 2 PM ↔ 02:00) in both the preview and the resulting filter suggests AM/PM is being inverted when the picker value is converted to 24-hour time.

No

chaudl113:
## What does this PR do?

Fixes #13687

When switching between AM and PM in the time period selector, the `handleValueChange` callback used the stale closure value of `period` (the previous value) instead of `value` (the newly selected value) when calling `setDateByType`. This caused the 12-to-24 hour conversion to use the wrong period, resulting in a 12-hour swap in the displayed time.

## Root Cause

In `time-period-select.tsx`, `handleValueChange(value: Period)`:
```tsx
// Before (bug): uses stale 'period' from closure
period === "AM" ? "PM" : "AM"

// After (fix): uses the newly selected 'value' directly
value
```

## Example

1. Time is 02:00 AM
2. User switches period dropdown to PM
3. **Before**: `setDateByType` receives `"AM"` (old period) → time stays 02:00
4. **After**: `setDateByType` receives `"PM"` (new value) → time correctly becomes 14:00

## How to test

1. Open any date/time picker (e.g., Traces filter)
2. Set time to 02:00 AM
3. Switch period dropdown to PM
4. **Before fix**: Time shows 02:00 (unchanged)
5. **After fix**: Time correctly shows 14:00

This PR bundles three fixes: a stale-closure bug in the AM/PM time picker (the headline fix), a localStorage-backed layout persistence for experiment results, and a credentials-guard on the sign-up page that prevents the password step from appearing when `AUTH_DISABLE_USERNAME_PASSWORD` is set.

- **`time-period-select.tsx`**: Replaces `period === \"AM\" ? \"PM\" : \"AM\"` with `value` so `setDateByType` receives the newly selected period rather than its logical inverse.
- **`useExperimentResultsState.ts`**: Drops `withDefault` from the `layout` URL param and introduces `useLocalStorage` so the preferred layout survives cross-page navigation without polluting clean URLs.
- **`sign-up.tsx`**: Guards `showPasswordStep` and the `handleContinue` fallback path with `authProviders.credentials`, so a credentials-disabled instance no longer surfaces a password form and instead shows a helpful error.
</details>

Safe to merge — all three changes address real defects or clear gaps, with no regressions introduced on the main paths.

The time-picker closure fix is minimal and clearly correct. The localStorage layout persistence is well-structured, though there is one worth-noting behavior: setLayout writes the value to the URL query param on every call but never removes it, so a URL copied after a layout switch carries ?layout=… and will silently override any localStorage preference for whoever opens it. The sign-up credentials guard is correct; the only edge case (both SSO and credentials disabled) predates this PR and was already broken.

web/src/features/experiments/hooks/useExperimentResultsState.ts deserves a second look for the URL-param-vs-localStorage precedence behavior on shared links.
</details>

```mermaid
sequenceDiagram
    participant User
    participant TimePeriodSelect
    participant setDateByType

    Note over TimePeriodSelect: Before fix
    User->>TimePeriodSelect: Select "PM"
    TimePeriodSelect->>TimePeriodSelect: handleValueChange("PM")
    TimePeriodSelect->>setDateByType: "period==="AM"?"PM":"AM" passes inverted stale value"
    Note over setDateByType: Bug: wrong period used

    Note over TimePeriodSelect: After fix
    User->>TimePeriodSelect: Select "PM"
    TimePeriodSelect->>TimePeriodSelect: handleValueChange("PM")
    TimePeriodSelect->>setDateByType: "passes value = "PM" directly"
    setDateByType-->>TimePeriodSelect: corrected Date (e.g. 14:00)
    TimePeriodSelect->>User: Display updated time
```
</details>

`````markdown

---

### Issue 1 of 1
web/src/features/experiments/hooks/useExperimentResultsState.ts:85-88
**URL param persists and overrides localStorage on shared links**

`setLayout` always writes to `setState({ layout: newLayout })`, so the `layout` query param is baked into the URL and never cleared. A user who shares their URL (e.g. `?layout=list`) will cause the recipient's localStorage preference to be silently overridden on every visit via that link — and thereafter the recipient's URL also carries `?layout=list`. Consider either clearing the URL param after syncing to localStorage (`setState({ layout: undefined })`) so clean URLs fall back to the saved preference, or accepting the current precedence and documenting it.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: correct AM/PM period when toggling ..."](https://github.com/langfuse/langfuse/commit/767a48f4cb6cd88d12c8a29c2cab599269e382fd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35852501)</sub>

### DIFF
--- a/web/src/components/ui/time-period-select.tsx
+++ b/web/src/components/ui/time-period-select.tsx
@@ -46,7 +46,7 @@ export const TimePeriodSelect = React.forwardRef<
           new Date(date),
           hours.toString(),
           "12hours",
-          period === "AM" ? "PM" : "AM",
+          value,
         ),
       );
     }
--- a/web/src/features/experiments/hooks/useExperimentResultsState.ts
+++ b/web/src/features/experiments/hooks/useExperimentResultsState.ts
@@ -4,6 +4,7 @@ import {
   ArrayParam,
   StringParam,
 } from "use-query-params";
+import useLocalStorage from "@/src/components/useLocalStorage";
 
 const MAX_COMPARISONS = 4;
 
@@ -11,7 +12,7 @@ export function useExperimentResultsState() {
   const [state, setState] = useQueryParams({
     baseline: withDefault(StringParam, undefined),
     c: withDefault(ArrayParam, []),
-    layout: withDefault(StringParam, "grid"),
+    layout: StringParam,
     itemVisibility: withDefault(StringParam, "baseline-only"),
   });
 
@@ -76,10 +77,15 @@ export function useExperimentResultsState() {
     setComparisonIds(comparisonIds.filter((existingId) => existingId !== id));
   };
 
-  // Layout management
-  const layout = (state.layout as "grid" | "list") ?? "list";
+  // Layout management - persist to localStorage so preference survives navigation
+  const [savedLayout, setSavedLayout] = useLocalStorage<"grid" | "list">(
+    "experimentResultsLayout",
+    "grid",
+  );
+  const layout = (state.layout as "grid" | "list") ?? savedLayout ?? "grid";
   const setLayout = (newLayout: "grid" | "list") => {
     setState({ layout: newLayout });
+    setSavedLayout(newLayout);
   };
 
   // Item visibility management
--- a/web/src/pages/auth/sign-up.tsx
+++ b/web/src/pages/auth/sign-up.tsx
@@ -94,8 +94,9 @@ function StandardSignupFlow({
 
   // Two-step login flow: ask for email first, detect SSO, then either redirect to SSO or reveal password field.
   // Skip this flow when no SSO is configured - show password field immediately
+  // Also skip when credentials auth is disabled (AUTH_DISABLE_USERNAME_PASSWORD)
   const [showPasswordStep, setShowPasswordStep] = useState<boolean>(
-    !authProviders.sso,
+    !authProviders.sso && authProviders.credentials,
   );
   const [continueLoading, setContinueLoading] = useState<boolean>(false);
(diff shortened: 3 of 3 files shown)


nimarb (changes requested):
same, please split into separate PRs per change

PR #14072, opened 2026-06-07
feat(rbac): let cloud admins view org & project settings pages

Steffen911:
Add an admin-only fallback: when the session lookup misses and the user is an admin, resolve the org/project from new admin-aware tRPC reads (organizations.byId / projects.byId). Both are built on the existing protectedOrganizationProcedure / protectedProjectProcedure, so:

- access is enforced server-side exactly as before (non-admin non-members are rejected, and the query is disabled client-side for them, so their behavior is unchanged),
- the orgId is taken from the server-resolved ctx.session.orgId, never from client input, so this cannot be used to read across orgs,
- admin access stays audited via the existing sendAdminAccessWebhook.

A shared buildAdminOrgContext helper returns the exact session organization shape (role OWNER, plan, cloudConfig, projects), so all existing hook consumers keep their types and behavior.

This PR fixes a blank-page bug for Langfuse cloud admins viewing customer org and project settings: the existing hooks (`useOrganization`, `useProject`) only read from the session, and since admins aren't session-members of customer orgs, the lookup missed and the page showed nothing. The fix adds two new admin-only tRPC procedures (`organizations.byId`, `projects.byId`) that return the same `SessionOrganization` shape the hooks already expect, enabling a fallback read when the session lookup misses.

- Two new tRPC query procedures are added behind the existing `protectedOrganizationProcedure` / `protectedProjectProcedure` guards, which already handle admin bypass and audit webhooks.
- A shared `buildAdminOrgContext` helper constructs the session-compatible org/project shape from a raw DB lookup; it includes all non-deleted projects in the org, returned from `projects.byId` to any authenticated project member since the server endpoint lacks a matching admin guard.
- The client-side hooks correctly set `enabled: isAdmin && !fromSession` so non-admins never trigger the new queries.
</details>

Safe to merge for the admin-settings use case, but projects.byId may expose the full list of an org's projects to any project member who crafts a direct tRPC call, which could be broader than intended.

The org hook fallback and organizations.byId are well-scoped. The projects.byId endpoint's server handler returns buildAdminOrgContext whose output includes all non-deleted projects of the org — not just the one being queried. The client hook enforces isAdmin before calling it, but the server procedure (protectedProjectProcedure) allows any project member to invoke it directly. Whether this is an issue depends on Langfuse's project-visibility model, but the discrepancy between client-side and server-side guards warrants review before shipping to production.

web/src/features/projects/server/projectsRouter.ts — the new byId handler and whether a server-side admin guard belongs there.
</details>

```mermaid
sequenceDiagram
    participant Admin as Cloud Admin
    participant Hook as useOrganization / useProject
    participant TRPC as tRPC Middleware
    participant DB as Database

    Note over Admin,DB: Admin visits customer org/project settings page

    Admin->>Hook: page load (orgId / projectId from router)
    Hook->>Hook: session.organizations lookup miss (admin not a member)
    Hook->>TRPC: organizations.byId / projects.byId

    Note over TRPC: protectedOrganizationProcedure / protectedProjectProcedure
    TRPC->>TRPC: enforceIsAuthedAndOrgMember / enforceUserIsAuthedAndProjectMember
    TRPC->>DB: validate orgId/projectId, fetch orgId for project
    DB-->>TRPC: orgId resolved
    TRPC->>TRPC: sendAdminAccessWebhook (audit)
    TRPC->>TRPC: set ctx.session.orgId (server-resolved)

    TRPC->>DB: buildAdminOrgContext(orgId) org + all non-deleted projects
    DB-->>TRPC: organization record + projects[]
    TRPC-->>Hook: "SessionOrganization shape (role=OWNER)"

    Hook-->>Admin: org/project context available, settings page renders
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/features/projects/server/projectsRouter.ts:307-324
**Endpoint accessible to non-admin project members, leaking full org project list**

`protectedProjectProcedure` allows any authenticated user who is a member of the supplied `projectId` to call this endpoint. When a non-admin org member calls it directly, `buildAdminOrgContext` returns `organization.projects` containing **all** non-deleted projects in the org (with their names, metadata, retentionDays, etc.) — not just the one they have access to. The client-side `useProject` hook correctly gates the call behind `isAdmin`, but there is no corresponding server-side guard.

If project membership is meant to be siloed (users should only see projects they belong to), consider adding an admin check in the handler, or returning only the requested project's data and a minimal org shape instead of the full `buildAdminOrgContext` result.

### Issue 2 of 3
web/src/features/projects/server/projectsRouter.ts:307-324
**Redundant full-org fetch when only one project is needed**

`buildAdminOrgContext` fetches the organization record **plus all non-deleted projects** via a joined query. For orgs with many projects this is wasteful — the handler only needs a single project entry. The middleware (`enforceUserIsAuthedAndProjectMember`) already performs a focused DB read to resolve `orgId`; the handler then issues a second, broader query to fetch all of that org's projects just to filter to one.

Consider a targeted fetch using `ctx.session.projectId` (already validated by the middleware) to avoid the full project list scan.

### Issue 3 of 3
web/src/features/projects/server/projectsRouter.ts:314-316
The project lookup uses `input.projectId` (raw client value) rather than `ctx.session.projectId`, which the `enforceUserIsAuthedAndProjectMember` middleware sets from the same validated input. Using the session-resolved value is more consistent with how `ctx.session.orgId` is preferred over `input.orgId` throughout the rest of the tRPC layer.

```suggestion
      const project = organization?.projects.find(
        (p) => p.id === ctx.session.projectId,
      );
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(rbac): let cloud admins view org &amp; ..."](https://github.com/langfuse/langfuse/commit/894ead747f8f5b0270fae6f8cd516230447b76d9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36075252)</sub>

### CODE REVIEW

File: web/src/features/projects/hooks.ts [hunk 1]
```diff
@@ -22,15 +24,41 @@ export const useProject = (projectId: string | null) => {
...
+      meta: { silentHttpCodes: [404] },
+    },
+  );
+
+  if (fromSession) {
+    return {
+      project: fromSession.project,
+      organization: fromSession.organization,
+    };
+  }
```
Steffen911, line 49: Couldn't we do that ahead of the projects.byId query? That way we may skip a backend call? Not sure if relevant though or if nextjs forbids it for reasons.

web/src/features/organizations/server/adminOrgContext.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * This exists so Langfuse admins — who are not members of customer orgs and
+ * therefore have no entry in their session — can still resolve an org/project
+ * for client hooks (useOrganization/useProject) when viewing customer settings.
+ *
+ * SECURITY: this helper performs no access control. Callers must only invoke it
+ * after access has been enforced (e.g. behind protectedOrganizationProcedure /
+ * protectedProjectProcedure), and must pass the server-resolved orgId
+ * (ctx.session.orgId), never a raw client-supplied id. Admins are represented
+ * as Role.OWNER, mirroring how the tRPC middleware elevates them server-side.
```
Steffen911: Could we pass an existing context and then re-confirm that the invoking user is an admin?

web/src/features/projects/hooks.ts
niklassemmler: This was exactly my first thought! But claude says the adminFallback query isn't executed before the value is read (lazy evaluation). And there was a different UI/NextJS reason that I forgot.

web/src/features/organizations/server/adminOrgContext.ts
niklassemmler: great idea, this would simplify this logic

web/src/features/projects/hooks.ts
niklassemmler: > The restructure he's gesturing at is forbidden — but by React, not Next.js. Moving the useQuery call below an early return would make it a conditionally-executed hook, which violates React's rules of hooks (every hook must run in the same order on every render; otherwise React's internal hook slots misalign and the component crashes when fromSession flips between renders).

Sounds like frontend magic to me 🙄

PR #14098, opened 2026-06-08
feat(blob-exporter): gate isEnrichedBlobExportAvailable on V4 env vars for self-hosted

niklassemmler:
## What does this PR do?

Self-hosted instances with `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true` have `events_full` populated via the V4 migration path and should be able to use the enriched blob export source. Previously `isEnrichedBlobExportAvailable` returned `isCloud` only, so self-hosted operators with the preview flag enabled were incorrectly locked out of the EVENTS export source in the UI.

**Changes:**
- Extends `isEnrichedBlobExportAvailable` in `packages/shared` to accept an optional `isV4PreviewEnabled` boolean, returning `isCloud || isV4PreviewEnabled`
- Moves the availability check server-side: the `blobStorageIntegration.get` tRPC procedure now computes `isEnrichedExportAvailable` from `env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` and `env.LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN`, returning it alongside the persisted config
- Updates the blobstorage settings UI to read `eventsExportAvailable` from the tRPC response instead of calling the client-side predicate (which could never read the server-only env var)
- Adds 3 server tests covering Cloud, self-hosted-off, and self-hosted-on cases for the new flag

## Type of change

## Mandatory Tasks

## Checklist

- [x] I have self-reviewed the code
- [x] My code follows the style guidelines of this project (`pnpm run format`)
- [x] My changes generate no new warnings (`pnpm run lint`)
- [x] New and existing relevant tests pass locally with my changes

This PR gates the enriched blob export source (`EVENTS`) for self-hosted instances that have enabled `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN`, fixing a case where those operators were incorrectly locked out of the export source selector.

- Extends `isEnrichedBlobExportAvailable` to accept an optional `isV4PreviewEnabled` flag and moves the availability check server-side in the `blobStorageIntegration.get` tRPC procedure, which now returns `{ config, isEnrichedExportAvailable }` instead of just the config record.
- Updates the blobstorage settings UI to read `eventsExportAvailable` from the tRPC response (covering the server-only env var) rather than the client-side predicate.
- Adds three server-side tests covering the Cloud, self-hosted-off, and self-hosted-on scenarios for the new flag.
</details>

The main change is safe for existing users, but there is a form initialization bug that affects the exact user group this PR targets.

The `useEffect` in `BlobStorageIntegrationSettingsForm` only re-runs when the `state` prop (the persisted config) changes. For users who have not yet created a blob-storage config — the primary audience for this feature addition — `state` stays `undefined` both before and after the tRPC query resolves. `isEnrichedExportAvailable` changes from `false` to `true` once the query returns, but the effect never fires to reset the form. The Export Source dropdown appears for these users with `TRACES_OBSERVATIONS` as the selected value instead of the intended `EVENTS` default. Before this PR, `eventsExportAvailable` was derived from the immediately-available `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` client env var, so this stale-closure path was never reached.

The `useEffect` dependency array in `web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx` needs `eventsExportAvailable` (or `isEnrichedExportAvailable`) added to prevent the stale-closure scenario for users without an existing config.
</details>

`````markdown

---

### Issue 1 of 1
web/src/pages/project/[projectId]/settings/integrations/blobstorage.tsx:327-328
The `useEffect` that resets the form tracks only `state` (the persisted config) in its dependency array, but now also closes over `eventsExportAvailable`, which is derived from the `isEnrichedExportAvailable` prop. When a user has no existing blob-storage config, `state` is `undefined` before and after the tRPC query resolves, so the effect never re-runs. Meanwhile `isEnrichedExportAvailable` changes from `false` to `true` once the query returns, causing `eventsExportAvailable` to become `true` and `showExportSourceField` to flip to `true` — but the form is never reset, so the Export Source dropdown appears with the wrong default of `TRACES_OBSERVATIONS` instead of `EVENTS`. This is a regression specifically for the new target users (self-hosted V4-preview instances and pre-cutoff Cloud users) who have not yet configured blob storage.

```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [state, eventsExportAvailable]);
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(blob-exporter): gate isEnrichedBlob..."](https://github.com/langfuse/langfuse/commit/22f780ca8fe94d27e3426834b266c352d3dcf487) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36249692)</sub>

### DIFF
--- a/packages/shared/src/features/analytics-integrations/blob-export-gate.ts
+++ b/packages/shared/src/features/analytics-integrations/blob-export-gate.ts
@@ -39,8 +39,13 @@ export function isLegacyBlobExportAllowed(
 
 /**
  * Whether this deployment has the enriched events export path.
- * Today equivalent to `isLangfuseCloud`; flip when self-hosted is provisioned with the migration.
+ * True for Cloud, or for self-hosted instances that have opted into the V4 preview
+ * via LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN (server-side only — pass the flag
+ * from a server-computed context rather than reading it client-side).
  */
-export function isEnrichedBlobExportAvailable(isCloud: boolean): boolean {
-  return isCloud;
+export function isEnrichedBlobExportAvailable(
+  isCloud: boolean,
+  isV4PreviewEnabled?: boolean,
+): boolean {
+  return isCloud || isV4PreviewEnabled === true;
 }
--- a/web/src/__tests__/server/blob-storage-integration-trpc.servertest.ts
+++ b/web/src/__tests__/server/blob-storage-integration-trpc.servertest.ts
@@ -661,4 +661,71 @@ describe("Blob Storage Integration tRPC Router", () => {
       expect(result.isEnrichedExportAvailable).toBe(true);
     });
   });
+
+  describe("update: enriched export source guard", () => {
+    const originalRegion = env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION;
+    const originalV4Preview = env.LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN;
+
+    afterEach(() => {
+      (env as any).NEXT_PUBLIC_LANGFUSE_CLOUD_REGION = originalRegion;
+      (env as any).LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN =
+        originalV4Preview;
+    });
+
+    it("rejects EVENTS on self-hosted without V4 preview opt-in", async () => {
+      (env as any).NEXT_PUBLIC_LANGFUSE_CLOUD_REGION = undefined;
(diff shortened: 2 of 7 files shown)


niklassemmler:
* **#14098** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #14128, opened 2026-06-09
fix: use upsert for dataset run creation to prevent race condition (#13829)

### LINKED ISSUE #13829: bug: dataset_run_items.create race condition causes unique constraint violation on datasetRuns(dataset_id, project_id, name)
## Describe the bug

When calling `dataset.run_experiment()` with `max_concurrency > 1` on a fresh `run_name`, multiple concurrent items each try to create the same dataset run via `api.dataset_run_items.create()`. The backend Prisma layer appears to call `datasetRuns.create()` directly (not upsert), so all items except one fail with:

```
prisma:error
Invalid `prisma.datasetRuns.create()` invocation:
Unique constraint failed on the fields: (`dataset_id`,`project_id`,`name`)
```

In the Python SDK these surface as repeated log lines:

```
ERROR [langfuse] Failed to create dataset run item: Unique constraint failed on the fields: (`dataset_id`,`project_id`,`name`)
```

Net effect: the dataset run gets created by the first winner, but the remaining N-1 items lose their link to the run (no `dataset_run_item` row), so the run appears incomplete in the UI even though all traces are emitted correctly.

## To reproduce

```python
from langfuse import Langfuse, Evaluation

client = Langfuse(host="...", public_key="...", secret_key="...")
dataset = client.get_dataset("orchestrator-smoke")  # 28 items

def task(*, item, **kwargs):
    return {"echo": (item.input or {}).get("q", "")[:40]}

def scorer(*, input, output, expected_output, metadata, **kwargs):
    return [Evaluation(name="dummy", value=1.0)]

# Fails with multiple unique-constraint errors:
dataset.run_experiment(
    name="repro",
    run_name="repro-run-1",   # fresh run_name
    task=task,
    evaluators=[scorer],
    max_concurrency=5,         # >1 triggers the race
)
```

Setting `max_concurrency=1` makes the issue disappear because only one item attempts to create the run.

## Expected behavior

`api.dataset_run_items.create` should upsert the parent `dataset_run` when `run_name` is new, so N parallel callers all converge to the same `dataset_run_id` instead of racing on `create`.

Alternatively, the SDK could pre-create the run once before fanning out items.

## Environment

- Self-hosted `langfuse/langfuse:3.175.0` (latest v3 as of 2026-05-21)
- Worker `langfuse/langfuse-worker:3`
- ClickHouse + Postgres + Redis + MinIO (standard v3 stack)
- Python SDK: `langfuse==4.6.1`
- Python: 3.11

## Workaround

`max_concurrency=1` serializes all `dataset_run_items.create` calls so the first creates the run and the rest reuse it. Cost: experiments run serially instead of in parallel.

michaelxer:
## Problem

When calling `dataset.run_experiment()` with `max_concurrency > 1` on a fresh `run_name`, multiple concurrent `POST /api/public/dataset-run-items` calls race to create the same dataset run. The `createOrFetchDatasetRun` function catches the unique-constraint error and re-fetches the existing run, which is functionally correct. However, Prisma still logs the constraint violation at the driver level on every concurrent call:

```
prisma:error
Invalid `prisma.datasetRuns.create()` invocation:
Unique constraint failed on the fields: (`dataset_id`,`project_id`,`name`)
```

With `max_concurrency=5` on a 28-item dataset, this produces 27 error log lines per experiment run.

## Fix

Replace the create-then-catch pattern with a single `prisma.datasetRuns.upsert()` on the `(datasetId, projectId, name)` unique constraint. Upsert is atomic: concurrent callers all converge to the same run without triggering error logs.

The `isUniqueConstraintError` helper is no longer needed and is removed.

## Changes

- `packages/shared/src/server/repositories/dataset-runs.ts`: Replace `findUnique` + `create` + `catch` with `upsert`. 1 file changed, -41 net lines.

## Verification

- The existing `POST /api/public/dataset-run-items` integration tests in `datasets-api.servertest.ts` exercise this code path.
- The Prisma schema confirms `@@unique([datasetId, projectId, name])` on `DatasetRuns`, and `prisma.datasetRuns.upsert` with `datasetId_projectId_name` is already used in the seed script (`packages/shared/scripts/seeder/seed-postgres.ts`).
- `pnpm run lint` passes on the changed file.

## Limitations

- Could not run the full test suite locally (requires Postgres, ClickHouse, Redis, MinIO). CI will verify.
- The experiments TRPC router (`web/src/features/experiments/server/router.ts:250`) also uses a bare `datasetRuns.create()`, but that path creates the run once before queuing items and is not affected by the concurrent race. A follow-up could unify it to use `createOrFetchDatasetRun` for consistency.

Fixes #13829

This PR replaces a `findUnique` + `create` + `catch(UniqueConstraintError)` pattern in `createOrFetchDatasetRun` with a single `prisma.datasetRuns.upsert()` call, eliminating spurious Prisma constraint-violation error logs when concurrent requests race to create the same dataset run.

- **`dataset-runs.ts`**: The helper `isUniqueConstraintError` is removed along with its three-step try/catch logic. A single `upsert` on the `(datasetId, projectId, name)` composite unique index replaces it; the `update: {}` empty block means concurrent callers that find an existing run return it unchanged without modifying any fields — preserving the original semantics of the function.
- The sole caller in `publicDatasetService.ts` is unaffected; the function signature and return type are identical.
</details>

Safe to merge — the change is a focused, mechanical replacement of a three-step optimistic-concurrency pattern with a single upsert; the function signature, return type, and fetch-without-modifying-if-exists semantics are all preserved.

The upsert correctly targets the same composite unique index (datasetId, projectId, name) that the old code relied on. The update: {} empty block is valid in Prisma and reproduces the original return existing row unchanged behaviour. The sole caller in publicDatasetService.ts is unaffected. The deleted isUniqueConstraintError helper has no other references in the codebase.

</details>

```mermaid
sequenceDiagram
    participant C1 as Caller 1
    participant C2 as Caller 2 (concurrent)
    participant PG as PostgreSQL

    Note over C1,PG: Before (create-then-catch)
    C1->>PG: findUnique (datasetId, projectId, name)
    C2->>PG: findUnique (datasetId, projectId, name)
    PG-->>C1: null
    PG-->>C2: null
    C1->>PG: INSERT INTO DatasetRuns ...
    C2->>PG: INSERT INTO DatasetRuns ...
    PG-->>C1: row created
    PG--xC2: UNIQUE CONSTRAINT VIOLATION (logged!)
    C2->>PG: findUnique (retry fetch)
    PG-->>C2: existing row

    Note over C1,PG: After (upsert)
    C1->>PG: INSERT ... ON CONFLICT (datasetId,projectId,name) DO UPDATE SET ... (upsert)
    C2->>PG: INSERT ... ON CONFLICT (datasetId,projectId,name) DO UPDATE SET ... (upsert)
    PG-->>C1: row (created or existing)
    PG-->>C2: row (existing, no error logged)
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: use upsert for dataset run creation..."](https://github.com/langfuse/langfuse/commit/d9f4647d6186717a91e191e525bb690c4e851a1a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36470467)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/dataset-runs.ts
+++ b/packages/shared/src/server/repositories/dataset-runs.ts
@@ -5,27 +5,18 @@ import type z from "zod";
 
 type Json = z.infer<typeof jsonSchema>;
 
-const isUniqueConstraintError = (error: any): boolean => {
-  return (
-    error.code === "P2002" || // Prisma unique constraint
-    error.message?.includes("duplicate key") ||
-    error.message?.toLowerCase().includes("unique constraint") ||
-    error.message?.includes("violates unique constraint")
-  );
-};
-
 /**
- * Create or fetch a dataset run with optimistic concurrency handling.
+ * Create or fetch a dataset run using an atomic upsert.
  *
  * Behavior:
- * - First tries to find an existing run by (projectId, datasetId, name).
- * - If not found, attempts to create it.
- * - If creation fails due to a unique constraint (likely created concurrently),
- *   fetches and returns the existing run.
- * - If all steps fail, throws an error.
+ * - Uses Prisma upsert on the (datasetId, projectId, name) unique constraint
+ *   so concurrent callers all converge to the same run without errors.
+ * - If the run already exists, returns it unchanged.
  *
- * Rationale: The public API can receive many POST requests almost simultaneously,
- * which is not concurrency-safe without this guard.
+ * Rationale: The public API can receive many POST requests almost simultaneously
+ * (e.g. dataset.run_experiment with max_concurrency > 1). The previous
+ * create-then-catch approach worked but logged Prisma unique-constraint errors
+ * on every concurrent call. Upsert is atomic and silent.
  */
 export const createOrFetchDatasetRun = async ({
   projectId,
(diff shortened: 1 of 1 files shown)


michaelxer:
Closing this PR as it's been open for over a week without review. The upsert approach is still valid — happy to reopen if the team decides to revisit dataset run creation race conditions.

PR #14137, opened 2026-06-09
fix(media): harden media database writes

hassiebp:
## What

Fixes the P0 media API DB failure mode from concurrent upload URL requests while keeping the change intentionally narrow.

The media table has two uniqueness constraints:

- `(project_id, id)` for the public `mediaId` and blob path identity
- `(project_id, sha_256_hash)` for content dedupe

Under high concurrency for the same media, Postgres can surface a conflict through either unique index. The previous upsert named only `(project_id, sha_256_hash)`, so concurrent inserts could still fail on the separate `(project_id, id)` uniqueness.

This PR changes the write to:

1. Keep the raw `INSERT ... ON CONFLICT DO NOTHING` with no conflict target, so conflicts from either media unique constraint are absorbed.
2. Always run Prisma `media.updateMany` guarded by `projectId`, derived `id`, and full `sha256Hash` to refresh metadata and prove the existing row is the same media.
3. Throw a clear invariant error if no row matches both the derived media ID and the full hash.

## Why

This avoids selecting only one of the two media unique indexes as the upsert arbiter. It keeps both uniqueness constraints, because they encode useful invariants: one for the public/storage identity and one for exact content dedupe.

The change intentionally favors a simple two-step write over a more complex CTE or conditional read/write flow. The second step is an unconditional guarded update because media upload URL requests may need to refresh metadata after failed upload attempts.

The change does not add broad retry handling or a new test suite. Transient DB retry/rate-limit follow-ups should be handled separately.

## Verification

- `pnpm exec prettier --check web/src/features/media/server/mediaService.ts`
- `pnpm --filter web exec eslint src/features/media/server/mediaService.ts --max-warnings 0`
- pre-commit hook ran repo Prettier and `turbo run lint`

Attempted existing media server test:

- `DATABASE_URL=postgresql://postgres:postgres@localhost:5432/langfuse NEXTAUTH_URL=http://localhost:3000 SALT=abcdefghijklmnopqrstuvwxyz123456 CLICKHOUSE_URL=http://localhost:8123 CLICKHOUSE_USER=default CLICKHOUSE_PASSWORD=password LANGFUSE_S3_EVENT_UPLOAD_BUCKET=test-bucket pnpm --filter web exec vitest run --project server src/__tests__/server/media.servertest.ts`
- blocked locally because database `langfuse` does not exist


hassiebp:
@claude review

web/src/features/media/server/mediaService.ts
```diff
@@ -218,12 +218,14 @@ async function upsertMediaRecord(params: {
...
-  for (let retryCount = 0; retryCount < maxRetries; retryCount += 1) {
-    try {
-      await prisma.$queryRaw`
+  // Media has two unique constraints: (project_id, id) for the public mediaId
+  // and (project_id, sha_256_hash) for content dedupe. Under concurrent uploads
+  // of the same file, either unique index can observe the speculative insert
+  // first. Omitting the conflict target makes DO NOTHING absorb conflicts from
+  // both constraints; the guarded Prisma update below then proves that the
+  // existing row has both the expected mediaId and full hash before reusing it.
+  await prisma.$executeRaw`
```
hassiebp, line 228: put into tx

PR #14168, opened 2026-06-10
feat(eval): add variable diagnostics to judge traces

RitwijParmar:
## What

Adds evaluator variable diagnostics to LLM-as-judge execution traces.

When a judge prompt is built, Langfuse now records whether each template variable was resolved, empty, or missing before the model call. The diagnostics are attached to both the worker span attributes and the internal evaluator trace metadata, so failed or suspicious eval runs are easier to debug without exposing the actual variable values.

## Why

A common failure mode for trace and dataset evaluators is a mapping that does not find the expected trace, observation, dataset field, or JSON selector output. Today that can turn into an empty prompt variable or a later confusing failure. This gives the eval execution trace enough structure to answer: which variables were present, which were empty, and which were never extracted.

## Details

- Adds reusable variable diagnostics helpers in evalRuntime
- Tracks resolved, empty, missing, and duplicate evaluator variables
- Adds count and variable-name metadata to the LLM judge trace
- Adds span attributes for the same diagnostics
- Avoids logging or storing variable values in diagnostics
- Adds helper tests and LLM execution metadata coverage

## Validation

- Passed: ./node_modules/.bin/tsc --noEmit --skipLibCheck from worker
- Passed: worker Prettier check on changed files
- Passed: git diff --check
- Could not run Vitest locally because Rollup fails at startup on this machine with a macOS native binary code-signature error for @rollup/rollup-darwin-arm64 before tests load.

This PR adds variable diagnostics to LLM-as-judge evaluation traces, recording whether each template variable was resolved, empty (extracted but blank), or missing (never found) prior to the model call. Diagnostics are attached as both span attributes and internal trace metadata, enabling easier debugging of evaluators without leaking actual variable values.

- New helpers in `evalRuntime.ts` (`buildEvalVariableDiagnostics`, `buildEvalVariableDiagnosticsMetadata`, `buildEvalVariableDiagnosticsSpanAttributes`) compute and format the diagnostics; `evalService.ts` calls them just before prompt compilation and merges the result into the existing execution metadata.
- Tests cover the main categorisation logic (resolved, empty, missing, duplicate), metadata formatting, span attribute formatting, and an integration-level check confirming that actual variable values are not leaked into trace metadata.
</details>

Safe to merge; the change is additive and does not alter existing eval execution logic or scoring paths.

The diagnostics helpers are well-isolated and the wiring in evalService correctly slots the new metadata in alongside the existing execution metadata. The one edge case worth attention is that empty arrays and empty objects stringify to non-empty strings (e.g. `"[]"`, `"{}"`), so those values are classified as "resolved" rather than "empty" — the diagnostics would show full coverage for a variable whose extracted value is `[]`, even though the LLM prompt receives a structurally empty placeholder.

worker/src/features/evaluation/evalRuntime.ts — the empty-container classification in buildEvalVariableDiagnostics
</details>

```mermaid
sequenceDiagram
    participant ES as evalService (runLLMAsJudgeEvaluation)
    participant ER as evalRuntime
    participant Span as Worker Span
    participant LLM as callLLM (traceSinkParams)

    ES->>ER: buildEvalVariableDiagnostics(templateVariables, extractedVariables)
    ER-->>ES: "EvalVariableDiagnostics {resolved, empty, missing, duplicate, coverageRatio}"

    ES->>Span: setAttributes(buildEvalVariableDiagnosticsSpanAttributes(diagnostics))

    ES->>ER: buildEvalVariableDiagnosticsMetadata(diagnostics)
    ER-->>ES: "Record<string, string> (counts + names)"

    ES->>ES: "evalExecutionMetadata = {executionMetadata, ...diagnosticsMetadata}"

    alt "missingVariables.length > 0 || emptyVariables.length > 0"
        ES->>ES: logger.warn(Evaluation variables resolved with gaps)
    end

    ES->>ER: compileEvalPrompt(templatePrompt, extractedVariables)
    ER-->>ES: compiled prompt string

    ES->>LLM: "callLLM({ traceSinkParams: { metadata: evalExecutionMetadata } })"
    LLM-->>ES: LLM response + executionTraceId

    ES-->>ES: "return { scores, executionTraceId, metadata: evalExecutionMetadata }"
```
</details>

`````markdown

---

### Issue 1 of 2
worker/src/features/evaluation/evalRuntime.ts:69-76
An empty array `[]` or empty object `{}` stringifies to `"[]"` / `"{}"` via `JSON.stringify`, so `.trim() === ""` is false and the variable is classified as `"resolved"` even though it carries no meaningful content. A user whose variable mapping returns an empty list (e.g., no observations found) will see `"resolved"` in the diagnostics while the prompt actually receives `[]`, which can be just as unhelpful as a blank string. Adding an explicit check for these structural empties would keep the diagnostics consistent with the user's intent.

```suggestion
    const stringifiedValue = parseUnknownToString(extractedVariable.value);
    const isStructurallyEmpty =
      (Array.isArray(extractedVariable.value) &&
        extractedVariable.value.length === 0) ||
      (typeof extractedVariable.value === "object" &&
        extractedVariable.value !== null &&
        !Array.isArray(extractedVariable.value) &&
        Object.keys(extractedVariable.value).length === 0);
    if (stringifiedValue.trim() === "" || isStructurallyEmpty) {
      return {
        variable,
        status: "empty" as const,
        valueType: getVariableValueType(extractedVariable.value),
      };
    }
```

### Issue 2 of 2
worker/src/features/evaluation/evalRuntime.ts:95-105
**Coverage ratio only counts strictly-resolved variables**

`coverageRatio` is computed as `resolvedVariables.length / templateVariables.length`, so empty-but-extracted variables reduce the ratio in the same way as truly missing ones. An evaluator that has every variable extracted but one of them blank would report the same ratio as one that never found that variable at all. Depending on how the ratio is surfaced in dashboards or alerting, this distinction could matter — consider also exposing an `extractedRatio` (`(resolved + empty) / total`) alongside the existing ratio, or renaming the current field to `resolvedRatio` to prevent misinterpretation.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(eval): add variable diagnostics to ..."](https://github.com/langfuse/langfuse/commit/efedfb149044617817b5a9459da38349277cca27) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36742978)</sub>

### DIFF
--- a/worker/src/features/evaluation/evalRuntime.ts
+++ b/worker/src/features/evaluation/evalRuntime.ts
@@ -26,6 +26,7 @@ export type EvalVariableDiagnostics = {
   missingVariables: string[];
   duplicateVariables: string[];
   coverageRatio: number;
+  extractedRatio: number;
 };
 
 export function compileEvalPrompt(params: CompileEvalPromptParams): string {
@@ -66,8 +67,7 @@ export function buildEvalVariableDiagnostics(params: {
       return { variable, status: "missing" as const };
     }
 
-    const stringifiedValue = parseUnknownToString(extractedVariable.value);
-    if (stringifiedValue.trim() === "") {
+    if (isEmptyVariableValue(extractedVariable.value)) {
       return {
         variable,
         status: "empty" as const,
@@ -102,6 +102,11 @@ export function buildEvalVariableDiagnostics(params: {
       params.templateVariables.length === 0
         ? 1
         : resolvedVariables.length / params.templateVariables.length,
+    extractedRatio:
+      params.templateVariables.length === 0
+        ? 1
+        : (resolvedVariables.length + emptyVariables.length) /
+          params.templateVariables.length,
   };
 }
 
@@ -116,6 +121,7 @@ export function buildEvalVariableDiagnosticsMetadata(
     eval_variable_duplicate_count:
       diagnostics.duplicateVariables.length.toString(),
     eval_variable_coverage_ratio: diagnostics.coverageRatio.toFixed(3),
+    eval_variable_extracted_ratio: diagnostics.extractedRatio.toFixed(3),
     ...(diagnostics.emptyVariables.length > 0
       ? { eval_variable_empty_names: diagnostics.emptyVariables.join(",") }
       : {}),
@@ -140,6 +146,7 @@ export function buildEvalVariableDiagnosticsSpanAttributes(
     "eval.variable.missing_count": diagnostics.missingVariables.length,
     "eval.variable.duplicate_count": diagnostics.duplicateVariables.length,
     "eval.variable.coverage_ratio": diagnostics.coverageRatio,
+    "eval.variable.extracted_ratio": diagnostics.extractedRatio,
     ...(diagnostics.emptyVariables.length > 0
       ? { "eval.variable.empty_names": diagnostics.emptyVariables }
       : {}),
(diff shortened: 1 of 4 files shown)


RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #14170, opened 2026-06-10
fix(media): drop media link foreign keys

hassiebp:
## What changed

- Drops the remaining `trace_media -> media` and `observation_media -> media` composite foreign keys via Prisma migration.
- Removes the corresponding Prisma relation fields so future migrations do not recreate those FKs.
- Makes media deletion explicitly remove `trace_media` and `observation_media` rows for deleted media IDs before deleting `media` rows.
- Replaces the trace-deletion orphan-media relation filter with `NOT EXISTS` checks over the link tables.
- Updates the trace deletion test to assert media link rows are explicitly removed.

## Why

During the 2026-06-10 prod-us media API incident, Aurora Performance Insights showed the hot database work concentrated on:

```sql
INSERT INTO "observation_media" ("id", "project_id", "trace_id", "observation_id", "media_id", "field")
VALUES ($1, $2, $3, $4, $5, $6)
ON CONFLICT DO NOTHING
```

The dominant waits were `LWLock:MultiXactMemberSLRU`, `LWLock:MultiXactMemberBuffer`, `LWLock:MultiXactOffsetBuffer`, and `Lock:transactionid`. Datadog child spans showed fanout across multiple hot media IDs: many distinct observations concurrently referenced the same reusable media assets.

The synchronous media-link insert path therefore turns hot media rows into parent-row FK lock hotspots. `ON CONFLICT DO NOTHING` keeps idempotency, but it does not avoid FK checks, parent-row locking, unique-index probes, or MultiXact overhead.

## Impact

This keeps media-link idempotency and lookup indexes, but removes the parent-row FK check from the high-write upload URL path. Link cleanup is now explicit in media deletion code rather than relying on FK cascade from `media`.

## Validation

Passed:

- `pnpm --filter worker run build`
- `pnpm --filter @repo/eslint-plugin run build`
- commit hook: `prettier --check "**/*.{js,jsx,ts,tsx,css}" --experimental-cli` and `turbo run lint`

Attempted but blocked by local environment credentials:

- `pnpm --filter worker exec vitest run src/__tests__/traceDeletion.test.ts src/__tests__/projectDeletionProcessing.test.ts src/__tests__/dataRetentionProcessing.test.ts`
- Initial run required shared build; after building, the suite was blocked by missing env vars, then by local ClickHouse auth, expired AWS SSO for S3, and missing `SALT`.

## Linear

I prepared the Linear issue text, but the writable Linear app connector currently requires reauthentication, so issue creation is blocked from Codex until the app is reauthorized.

This PR removes the composite foreign keys from `trace_media` and `observation_media` to `media`, replacing the implicit FK-cascade cleanup with explicit link-row deletion in `deleteMediaFiles` and a new `deleteMediaLinkRowsByProjectId` helper. The change directly addresses MultiXact lock contention on hot media rows during high-concurrency upload URL requests.

- **Migration + schema**: drops `trace_media_media_id_project_id_fkey` and `observation_media_media_id_project_id_fkey` via `ALTER TABLE … DROP CONSTRAINT IF EXISTS`; removes the Prisma relation fields so future `db:generate` runs do not recreate them.
- **`deleteMediaFiles`**: now explicitly deletes link rows inside the per-batch PG transaction before deleting `media` rows, keeping deletion atomic per batch without relying on cascade.
- **Orphan detection in trace deletion**: since the Prisma relation fields are gone, the `TraceMedia: { none: {} }` / `ObservationMedia: { none: {} }` filter is replaced with equivalent `NOT EXISTS` sub-queries in a `$queryRaw`.
</details>

The core FK drop and batch-deletion logic are sound, but two spots leave media rows or S3 objects stranded on worker restart.

The new `deleteMediaLinkRowsByProjectId` uses a single, unbounded `$transaction` over potentially millions of rows per project; a timeout leaves project deletion partially applied with no way to resume, unlike the batched approach used by `deleteMediaFiles`. More critically, in `deleteMediaItemsForTraces` the link-row delete and the orphaned-media probe are two separate round-trips without a surrounding transaction: a worker crash between them removes the link rows but leaves `media` rows and S3 objects permanently unreachable because retries see an empty `allMediaIds` and return early.

`packages/shared/src/server/media-deletion.ts` (unbounded transaction in `deleteMediaLinkRowsByProjectId`) and `worker/src/features/traces/processClickhouseTraceDelete.ts` (non-atomic two-phase link-delete to orphan-probe).
</details>

```mermaid
sequenceDiagram
    participant PD as projectDeleteProcessor
    participant DMF as deleteMediaFiles
    participant DMLR as deleteMediaLinkRowsByProjectId
    participant S3
    participant PG as PostgreSQL

    PD->>PD: findAllMediaByProjectId
    alt S3 configured
        loop per batch (10k)
            PD->>DMF: deleteMediaFiles(batch)
            DMF->>S3: deleteFiles(bucketPaths)
            DMF->>PG: transaction traceMedia.deleteMany + observationMedia.deleteMany + media.deleteMany
        end
    end
    PD->>DMLR: deleteMediaLinkRowsByProjectId
    DMLR->>PG: transaction traceMedia.deleteMany(projectId) + observationMedia.deleteMany(projectId)
    PD->>PG: project.delete() cascades to media rows

    Note over PD,PG: Trace deletion path
    participant TDD as deleteMediaItemsForTraces
    TDD->>PG: findMany traceMedia + observationMedia by traceId
    TDD->>PG: Promise.all deleteMany by traceId
    TDD->>PG: queryRaw NOT EXISTS orphanedMedia
    TDD->>S3: deleteFiles orphanedMedia.bucketPaths
    TDD->>PG: media.deleteMany orphanedMedia.ids
```
</details>

`````markdown

---

### Issue 1 of 2
packages/shared/src/server/media-deletion.ts:54-65
**Unbounded transaction on `deleteMediaLinkRowsByProjectId`**

Both `deleteMany` calls are wrapped in a single `prisma.$transaction`. For a project with millions of `trace_media` or `observation_media` rows, this generates a single `DELETE WHERE project_id = ?` per table, and the combined transaction holds row-locks and accumulates WAL for the entire lifetime of both deletes. If either delete times out, the whole project-deletion retry re-enters the same path and faces the same oversized transaction again.

`deleteMediaFiles` uses `BATCH_SIZE = 10_000` for exactly this reason; applying the same chunked approach here (e.g. looping `deleteMany` with `take: BATCH_SIZE` or equivalent cursor pages) would bound the lock window and make partial failures easier to recover from.

### Issue 2 of 2
worker/src/features/traces/processClickhouseTraceDelete.ts:55-72
**Non-atomic link deletion and orphan detection can permanently strand media**

The `traceMedia` and `observationMedia` link rows are deleted here, and the orphaned-media probe runs later at lines 83–105. These two steps are not in the same transaction. If the process crashes (or the worker is killed) after the `Promise.all` deletes succeed but before `prisma.media.deleteMany` at line 114 completes, on the next retry `findMany` for those trace IDs returns empty sets (the link rows are gone), so `allMediaIds` is empty and the function returns early — leaving the formerly-referenced `media` rows and their S3 objects as permanent orphans.

This two-phase structure was present before this PR, but with FKs in place the Prisma relation queries depended on the link rows existing to identify orphans; now that the FK guard is removed, a crash in this window is the only remaining integrity backstop. Wrapping the link deletions and the subsequent orphan fetch+delete into a single serializable transaction would eliminate the window.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(media): drop media link foreign keys"](https://github.com/langfuse/langfuse/commit/75e2459ad71948314d99b073d88defc94d8b31ba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36758043)</sub>


worker/src/features/traces/processClickhouseTraceDelete.ts
sumerman, line 55: since we are touching this might make to bundle these two into a single tx

sumerman (approved):
Thank you!

hassiebp:
@claude review

PR #14180, opened 2026-06-10
feat(agent): Add redirect tool

bezbac:

This PR adds a `langfuse_proposeRedirect` tool to the in-app agent that lets the AI propose in-app navigation actions displayed as clickable buttons in the chat UI. The schema uses a flat-params-plus-superRefine pattern to produce an LLM-friendly flat JSON Schema while still enforcing destination-specific parameter constraints via an internal discriminated union.

- **Schema (`schema.ts`)**: Defines a broad flat input schema validated at runtime against a strict discriminated union via `superRefine`, keeping the LLM-visible JSON Schema as a flat object with no `oneOf`/`anyOf`.
- **URL construction (`redirectAction.ts`)**: Converts validated tool-call arguments to typed internal paths; handles V4 vs. non-V4 column/operator differences for `traceId`, `version`, and the timestamp column, though `release` is silently dropped under V4.
- **Frontend integration**: The `ControlledInAppAgentWindow` extracts redirect tool calls from agent messages, builds the `href` client-side (never from raw LLM text), and renders each as a `RedirectAction` button using `router.push`.
</details>

Safe to merge; all navigation href construction is done from schema-validated inputs and never from raw LLM text, so there is no open-redirect or injection risk.

The core redirect flow is well-designed: URLs are always built from Zod-validated tool-call arguments client-side, not from the tool result, so a faulty execute response cannot affect navigation. The main concerns are the release filter being silently dropped for V4 users (potentially misleading redirect button labels) and a brief flash-in for traces-with-params buttons while the V4 flag initializes. Neither affects correctness of other features or data integrity.

Pay closest attention to redirectAction.ts — specifically the release filter branch — and ControlledInAppAgentWindow.tsx for the isV4TracesInitializing guard.
</details>

```mermaid
sequenceDiagram
    participant User
    participant InAppAgentWindow
    participant AgentServer as Agent (server)
    participant LLM
    participant RedirectTool as langfuse_proposeRedirect

    User->>InAppAgentWindow: sends message
    InAppAgentWindow->>AgentServer: submit(message)
    AgentServer->>LLM: stream with tools + system prompt
    LLM->>AgentServer: assistant message + tool call (proposeRedirect args)
    AgentServer->>RedirectTool: execute(validated input)
    RedirectTool-->>AgentServer: "{status: proposal_created, destination}"
    AgentServer-->>InAppAgentWindow: stream events (tool call args in message)
    Note over InAppAgentWindow: getRedirectActionMessage() builds href from tool call args (NOT from tool result)
    InAppAgentWindow->>User: renders RedirectAction button
    User->>InAppAgentWindow: clicks button
    InAppAgentWindow->>User: router.push(href) — client-side navigation
```
</details>

`````markdown

---

### Issue 1 of 3
web/src/ee/features/in-app-agent/components/utils/redirectAction.ts:233-246
**`release` filter silently dropped for V4 users**

When `isV4TracesEnabled` is `true`, the `release` filter is quietly omitted from the generated URL. The agent has no way to know whether a user is on V4, so it may produce a redirect with `filters.release` set and a label implying that filter is active — but clicking the button navigates to an unfiltered traces list. Meanwhile, `version` is handled for both V4 and non-V4 using the right column/operator per variant. Consider either removing `release` from `InAppAgentTracingFiltersSchema` entirely, or applying an equivalent V4 substitution the same way `version` and `timestamp`/`startTime` are handled.

### Issue 2 of 3
web/src/ee/features/in-app-agent/server/agent.ts:548-556
**Redundant second parse of already-validated input**

`createTool` validates the input against `inputSchema` before calling `execute`, so `input` is already a typed `InAppAgentRedirectToolInput` value. Calling `InAppAgentRedirectToolInputSchema.parse(input)` inside `execute` is a no-op under normal conditions but adds a parsing cost on every invocation. Destructuring `destination` directly from the typed parameter is cleaner.

```suggestion
        execute: async ({ destination }) => {
          return {
            status: "proposal_created" as const,
            destination,
          };
        },
```

### Issue 3 of 3
web/src/ee/features/in-app-agent/components/ControlledInAppAgentWindow.tsx:152-154
**Redirect button for traces with params absent while V4 initializes**

`getTracesRedirectHref` returns `null` whenever `isV4TracesInitializing` is `true`, so if the agent response arrives before the V4 feature flag finishes loading, no redirect button is rendered. Once `isV4TracesInitializing` flips to `false`, the `useMemo` recomputes and the button flashes in. For destinations without params (e.g., bare `traces`, `sessions`) the button renders immediately, creating an inconsistency in the same conversation. A loading skeleton placeholder for the button, or a brief debounce on initial render, would avoid the visual jump.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(agent): Add redirect tool"](https://github.com/langfuse/langfuse/commit/73e830f1d97866682538790af861fc3d2f8d7319) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36908126)</sub>

### CODE REVIEW

File: web/src/ee/features/in-app-agent/schema.ts [hunk 1]
```diff
@@ -186,3 +198,210 @@ export const InAppAgentRuntimeStateSchema = z.discriminatedUnion("type", [
...
+  .superRefine((value, ctx) => {
+    if (!InAppAgentTableTimeRangeStrictSchema.safeParse(value).success) {
+      ctx.addIssue({
+        code: "custom",
+        message: "Provide either a preset or both from and to.",
+      });
+    }
+  }) as z.ZodType<z.infer<typeof InAppAgentTableTimeRangeStrictSchema>>;
+
+const InAppAgentTracingFiltersSchema = z.object({
```
nimarb: you probably tried, but there is no way to reuse existing schemas for filtering right? so that when we add something it's immediately reflected and we don't need to care about this here

bezbac: Yeah, I would LOVE to be able to do this. Unfortunately, `web/src/ee/features/in-app-agent/schema.ts` is expected to be imported from the clients and the existing schemas are close to the clickhouse table definitions and load server-only things so when I imported them it blew up.

The file actually has a comment that I introduced in another PR on this: https://github.com/langfuse/langfuse/blob/994e5530e2f0ecfedf3468333675e1290b364761/web/src/ee/features/in-app-agent/schema.ts#L1-L5

bezbac: We can probably have a look at refactoring this to make it possible but I'd do it outside of this PR.

nimarb (approved):

PR #14192, opened 2026-06-11
chore(observability): drop redundant aws-ecs and process resource detectors

Steffen911:

This PR removes the `awsEcsDetector` and `processDetector` resource detectors from the OpenTelemetry SDK configuration in both the `web` and `worker` packages, and drops the `@opentelemetry/resource-detector-aws` dependency entirely.

- The motivation is bandwidth savings: Datadog's OTLP intake flattens resource attributes onto every span, so the removed detectors were adding ~1KB of duplicated infra metadata per span that the Datadog agent already supplies independently.
- `containerDetector` is intentionally preserved because the Datadog agent relies on the `container.id` resource attribute to correlate container-level metadata, and `envDetector` is kept for OTEL environment variable support. The change is applied symmetrically in both packages and includes the matching `pnpm-lock.yaml` update.
</details>

Safe to merge — the change removes two resource detectors whose output was already being supplied by the Datadog agent, reducing per-span ingest overhead without dropping any attributes the tracing pipeline actually depends on.

The removal is symmetric across both packages, the lockfile is updated to match, and the one detector that Datadog requires for container-level correlation (containerDetector) is explicitly kept. The inline comments document the rationale clearly, and no logic outside of resource detection is touched.

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into steffen/drop-re..."](https://github.com/langfuse/langfuse/commit/c958ef85f763d0533c42f0613322c103f41a08cc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36961070)</sub>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -479,9 +479,6 @@ importers:
       '@opentelemetry/instrumentation-winston':
         specifier: ^0.57.0
         version: 0.57.0(@opentelemetry/api@1.9.0)
-      '@opentelemetry/resource-detector-aws':
-        specifier: ^2.13.0
-        version: 2.14.0(@opentelemetry/api@1.9.0)
       '@opentelemetry/resource-detector-container':
         specifier: ^0.8.4
         version: 0.8.5(@opentelemetry/api@1.9.0)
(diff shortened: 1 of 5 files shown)


Steffen911:
This removes infra attribution, so rather intrusive. Will keep things around.

PR #14201, opened 2026-06-11
fix(batch-actions): snapshot v4 events-table flag for session batch actions

niklassemmler:
## What does this PR do?

Session batch actions (select-all → add to annotation queue) follow the same persist-query-then-execute-in-worker pattern as batch exports, but never snapshotted the user's v4 beta flag. The worker therefore always resolved the session set via the legacy traces-based query — even when the user's sessions UI table is backed by the events service. With identical filters the two reads can disagree, so sessions visible in the UI could be skipped (or extra ones included) when the bulk action executes.

This PR mirrors the batch-export fix from #14040:

- Snapshot `ctx.session.user.v4BetaEnabled` as `query.useEventsTable` in `createBatchActionJob` at dispatch time (covers all batch-action call sites; overrides any client-sent value).
- Persist it via `BatchActionQuerySchema` so it travels inside the `BatchActionProcessingEvent` payload.
- Pass it through to `getDatabaseReadStreamPaginated` in `handleBatchActionJob`, whose sessions branch (added in #14040) routes to the events-table reader when the flag is true.
- Regression test: seeds sessions only in the events table and runs `session-add-to-annotation-queue` with `useEventsTable: true`; without the pass-through the legacy reader finds nothing and the test fails.

Deliberately out of scope: observation batch actions switch table names via the `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN` env flag rather than the user flag. Aligning that mechanism to the user flag would disable the events table for self-hosted preview users (the user flag is Cloud-only), so it is left unchanged.

Fixes [LFE-10264](https://linear.app/langfuse/issue/LFE-10264)

## Type of change

## Mandatory Tasks

This PR fixes a data-consistency bug in session batch actions (e.g. "select-all → add to annotation queue") where the worker always used the legacy ClickHouse traces reader instead of the events-table reader, causing the worker's session set to diverge from what the UI displayed for v4-beta users.

- **Snapshot at dispatch**: `createBatchActionJob` now snapshots `session.user.v4BetaEnabled` as `query.useEventsTable`, overriding any client-sent value, so the worker uses the same data source the UI used when the job was created.
- **Schema and handler update**: `useEventsTable` is added to `BatchActionQuerySchema` (optional, backward-compatible with in-flight jobs) and threaded through to `getDatabaseReadStreamPaginated`, which already contains the branch added in #14040.
- **Regression test**: Seeds session data only in the ClickHouse events table and asserts that `session-add-to-annotation-queue` with `useEventsTable: true` finds and enqueues both sessions; without the pass-through the legacy reader returns nothing.
</details>

Safe to merge — the change is a minimal, targeted fix that mirrors an already-merged pattern from #14040, the schema addition is backward-compatible with in-flight jobs, and a regression test validates the corrected code path end-to-end.

The fix is a one-line pass-through in the handler and a straightforward server-side snapshot in the dispatcher. The schema field is optional so old queued jobs with no useEventsTable will silently fall back to the legacy reader. The ?? false default correctly gates the new path to explicit opt-in users. The regression test is well-constructed and would have caught the original bug. No functional correctness issues found.

No files require special attention. The only observation is a minor comment imprecision in the test file.
</details>

```mermaid
sequenceDiagram
    participant UI as Browser / tRPC
    participant Server as createBatchActionJob
    participant Queue as BullMQ Queue
    participant Worker as handleBatchActionJob
    participant CH as ClickHouse (events table)
    participant CHLeg as ClickHouse (legacy traces)
    participant PG as Postgres

    UI->>Server: "createBatchActionJob({ query, session })"
    Note over Server: Snapshot v4BetaEnabled → useEventsTable<br/>(overrides any client value)
    Server->>Queue: "enqueue BatchActionProcessingJob<br/>{ query: { ...query, useEventsTable } }"

    Queue->>Worker: dequeue job

    Worker->>Worker: "route: session-add-to-annotation-queue<br/>tableName = Sessions → getDatabaseReadStreamPaginated"

    alt "useEventsTable = true (v4 beta user)"
        Worker->>CH: getSessionsWithMetricsFromEvents(filter)
        CH-->>Worker: session rows
    else "useEventsTable = false / undefined (legacy)"
        Worker->>CHLeg: getSessionsWithMetrics(filter)
        CHLeg-->>Worker: session rows
    end

    Worker->>PG: traceSession.findMany (bookmarked/public)
    PG-->>Worker: session metadata

    Worker->>PG: processAddSessionsToQueue(sessionIds)
    PG-->>Worker: annotationQueueItem records created
```
</details>

`````markdown

---

### Issue 1 of 1
worker/src/__tests__/batchAction.test.ts:884-885
The comment says sessions exist "only in the events table," but `traceSession` records are also created in Postgres above (which the read stream needs for `bookmarked`/`public` metadata and potential FK constraints). What's exclusively in the events table is the trace/generation data that makes sessions discoverable — `getSessionsWithMetrics` (legacy reader) won't find them because no rows exist in the legacy ClickHouse traces table.

```suggestion
    // No trace data is written to the legacy ClickHouse traces table — only to
    // the events table — so the legacy reader (useEventsTable: false) would
    // find no sessions. This verifies the flag is passed through correctly.
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: trim comments"](https://github.com/langfuse/langfuse/commit/8d063094d024ad6635fb9237749d86f2287992c4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36994657)</sub>

### DIFF
--- a/packages/shared/src/features/batchAction/types.ts
+++ b/packages/shared/src/features/batchAction/types.ts
@@ -35,6 +35,11 @@ export const BatchActionQuerySchema = z.object({
   orderBy,
   searchQuery: z.string().optional(),
   searchType: z.array(TracingSearchType).optional(),
+  // Snapshotted at dispatch time from the user's v4 beta flag. When true, the
+  // sessions read stream resolves the selection from the ClickHouse events
+  // table instead of the legacy traces path. Persisted in the job payload so
+  // the worker reads the snapshot, never the live user record.
+  useEventsTable: z.boolean().optional(),
 });
 
 export type BatchActionQuery = z.infer<typeof BatchActionQuerySchema>;
--- a/packages/shared/src/features/batchAction/types.ts
+++ b/packages/shared/src/features/batchAction/types.ts
@@ -35,10 +35,8 @@ export const BatchActionQuerySchema = z.object({
   orderBy,
   searchQuery: z.string().optional(),
   searchType: z.array(TracingSearchType).optional(),
-  // Snapshotted at dispatch time from the user's v4 beta flag. When true, the
-  // sessions read stream resolves the selection from the ClickHouse events
-  // table instead of the legacy traces path. Persisted in the job payload so
-  // the worker reads the snapshot, never the live user record.
+  // Dispatch-time snapshot of the user's v4 beta flag; routes the sessions
+  // read stream to the events table instead of the legacy traces path.
   useEventsTable: z.boolean().optional(),
 });
 
--- a/web/src/features/table/server/createBatchActionJob.ts
+++ b/web/src/features/table/server/createBatchActionJob.ts
@@ -57,9 +57,8 @@ export const createBatchActionJob = async ({
     tableName,
   });
 
-  // Snapshot the user's v4 beta flag into the persisted query so the worker
-  // resolves the selection from the same data source as the UI table the user
-  // selected from, never the live user record. Overrides any client-sent value.
+  // Snapshot the user's v4 beta flag so the worker reads from the same data
+  // source as the UI table; overrides any client-sent value.
   const queryWithSnapshot: BatchActionQuery = {
     ...query,
     useEventsTable: session.user.v4BetaEnabled ?? false,
--- a/web/src/features/table/server/createBatchActionJob.ts
+++ b/web/src/features/table/server/createBatchActionJob.ts
@@ -28,6 +28,7 @@ type CreateBatchActionJob = {
   session: {
     user: {
       id: string;
+      v4BetaEnabled?: boolean | null;
     };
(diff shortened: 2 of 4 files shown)


niklassemmler:
* **#14201** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #14228, opened 2026-06-12
feat(evals): allow deleting evaluators (UI, API, MCP)

wochinge:
## Summary

- Evaluators (eval templates) can now be deleted from the UI, the unstable public API (`DELETE /unstable/evaluators/{id}`), and MCP — deletion removes all versions of the evaluator and is blocked with `409` while running evaluators / evaluation rules still reference it (surfaced upfront in the UI delete popover). Langfuse-managed evaluators cannot be deleted.
- Data access and deletion rules live in a new `evaluatorRepository` + `evalTemplateDeletion` service shared by all three surfaces, with per-version audit logging.
- New migration adds an index on `job_executions(job_template_id)` so the FK's `ON DELETE SET NULL` does not seq-scan the table on template deletion.

## Linear

- [LFE-10292](https://linear.app/langfuse/issue/LFE-10292/allow-deleting-evaluators)

## Review Focus

- Migration: `CREATE INDEX CONCURRENTLY IF NOT EXISTS` on the large `job_executions` table.
- Public API contract: `fern/apis/server/definition/unstable/evaluators.yml` + regenerated `openapi.yml` (family-deletion semantics, `409`/`403` behavior).


web/src/features/evals/server/evalTemplateDeletion.ts
```diff
@@ -0,0 +1,180 @@
...
+  // reference check and the delete see a consistent snapshot.
+  const deletedVersions = await prisma.$transaction(async (tx) => {
+    // lock first: blocks concurrent rule creation (FK takes FOR KEY SHARE on
+    // the template row) from slipping in between the check and the delete
+    await lockEvalTemplateFamilyVersions({
+      tx,
+      projectId,
+      name: template.name,
+      type: template.type,
+    });
```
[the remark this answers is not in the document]
wochinge: It's fine at this point.

web/src/features/evals/server/unstable-public-api/evaluator-service.ts
```diff
@@ -345,3 +346,19 @@ export async function createPublicEvaluator(params: {
...
+  // an evaluator in the public contract is the whole family; deleting it
+  // removes all stored versions
+  await deleteEvalTemplateFamily({
+    prisma,
+    projectId: params.projectId,
+    evalTemplateId: params.evaluatorId,
+    auditScope: params.auditScope,
+    referencingEntityName: "evaluation rule",
+  });
+}
```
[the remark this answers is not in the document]
wochinge: I'd accept this for the unstable APi - what do you think @hassiebp ?

packages/shared/prisma/migrations/20260612000000_add_job_executions_job_template_id_idx/migration.sql
```diff
@@ -0,0 +1,4 @@
+-- CreateIndex
+-- Required so the FK's ON DELETE SET NULL on job_executions.job_template_id
+-- does not seq-scan the table when an eval template is deleted.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "job_executions_job_template_id_idx" ON "job_executions"("job_template_id");
```
wochinge: this is our spicy migration that I was talking about @hassiebp . Not worried though - same as we ran end of march https://github.com/langfuse/langfuse/pull/12760

hassiebp: Is this strictly necessary? Instead of building and maintaining another massive index on `job_executions` we could also simply change the `on delete` on the FK to do nothing. It's true that we might have orphaned records in job executions pointing to delated templates, but this may be acceptable. Could you please explore and make a recommendation?

wochinge: > change the on delete on the FK to do nothing.

That would only work if drop the foreign key relation completely.

`onDelete: NoAction` would just result in blocking the deletion of templates if referenced by a jobExecution which would effectively block all deletions.

> The default ON DELETE action is ON DELETE NO ACTION; this does not need to be specified. **This means that the deletion in the referenced table is allowed to proceed. But the foreign-key constraint is still required to be satisfied, so this operation will usually result in an error.** But checking of foreign-key constraints can also be deferred to later in the transaction (not covered in this chapter). In that case, the NO ACTION setting would allow other commands to “fix” the situation before the constraint is checked, for example by inserting another suitable row into the referenced table or by deleting the now-dangling rows from the referencing table.

wochinge: I'd bite the bullet and simply add the index for now (given that we plan to migrate to CH tables anywhere soon enough for `JobExecution`

hassiebp: What would be the impact of dropping the FK altogether?

hassiebp: Not yet a fan of adding the index

wochinge: Dropping the foreign key would mean
- we need to also drop the prisma relations (both sided) as prisma relies on the foreign key for the relation
- I checked and no queries would be impacted that (they don't use the relationship attribute)

Given that we fully control the job execution side (not publicly exposed) it would be fine to drop the extra validation that that the foreign key provides us. 

Let me change the PR to that.

wochinge: done

web/src/components/deleteButton.tsx
```diff
@@ -462,47 +478,130 @@ export function DeleteEvaluationModelButton(
...
-//         })
-//       }
-//       entityToDeleteName="template"
-//       executeDeleteMutation={executeDeleteMutation}
-//       isDeleteMutationLoading={templateMutation.isLoading}
-//       enabled={hasModelBasedEvaluationEntitlement}
-//     />
-//   );
-// }
+export function DeleteEvalTemplateButton(
```
hassiebp: nit: This component should live in a separate file

web/src/features/evals/server/unstable-public-api/evaluator-service.ts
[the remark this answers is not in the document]
hassiebp: agree

web/src/features/evals/server/router.ts
```diff
@@ -1628,52 +1632,49 @@ export const evalRouter = createTRPCRouter({
...
+      });
+
+      return findEvalTemplateFamilyUsage({
+        prisma: ctx.prisma,
+        projectId,
+        evalTemplateId,
+      });
+    }),
+
+  deleteEvalTemplate: protectedProjectProcedure
```
hassiebp, line 1651: Have you made sure that only user-created template can be deleted? Langfuse-managed templates should not be allowed to be deleted

wochinge: yes - see line 91 of `web/src/features/evals/server/evalTemplateDeletion.ts`

hassiebp (approved):

web/src/components/deleteButton.tsx
wochinge: done

PR #14257, opened 2026-06-15
feat(datasets): media attachments for dataset items

wochinge:
## What does this PR do?

Adds end-to-end media attachment support for dataset items. A new `dataset_item_media` table links uploaded media to a specific item version; item writes parse and persist media references, and read paths (tRPC + public API) resolve them to signed download URLs. The create/edit dialogs gain an upload button, editor drop/paste, and a preview section, the view and historical-version pages render each version's saved attachments, and media is cleaned up when its owning trace/dataset is deleted or ages out via retention. Separately, it isolates CodeMirror re-renders so typing in one dataset item field no longer re-renders or reconfigures the other editors.

Fixes https://linear.app/langfuse/issue/LFE-10291/v1-ui-changes & https://linear.app/langfuse/issue/LFE-10290/v1-backend-changes

## Recording / Walkthrough
https://www.loom.com/share/ba6cec8eb53a4c32afbb9696ffd08d47

## Type of change

## Mandatory Tasks


web/src/features/datasets/server/dataset-router.ts
```diff
@@ -717,6 +726,104 @@ export const datasetRouter = createTRPCRouter({
...
     }),
+  // Issue a presigned upload URL for media attached to a dataset item from the
+  // UI (no trace context). The dataset item association is created when the
+  // item is saved, which parses the media reference out of the item JSON.
+  getItemMediaUploadUrl: protectedProjectProcedure
+    .input(
+      z.object({
+        projectId: z.string(),
+        contentType: z.enum(MediaContentType),
+        contentLength: z.number().positive().int(),
```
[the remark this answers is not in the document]
wochinge: Good catch, agreed. Fixed in 807590bfe: `getItemMediaUploadUrl` now rejects `contentLength > LANGFUSE_S3_MEDIA_MAX_CONTENT_LENGTH` before signing the URL, mirroring the public media route, with a server test covering the oversized case.

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -474,6 +510,20 @@ export async function deleteDatasetItem(props: {
...
+      // kept. STATEFUL items have a single version, so this covers all rows.
+      await syncDatasetItemMedia({
+        projectId: props.projectId,
+        items: [
+          {
+            datasetId: item.datasetId,
+            datasetItemId: item.id,
+            datasetItemValidFrom: item.validFrom,
+          },
+        ],
```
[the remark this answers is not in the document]
wochinge: This was an existing issue. Wouldn't tackle here as it could significantly slow down retention queries while being unlikely to actually be hit in real life

packages/shared/src/server/media-deletion.ts
```diff
@@ -103,6 +111,12 @@ export async function deleteMediaFiles(params: {
...
       }),
+      prisma.datasetItemMedia.deleteMany({
+        where: {
+          projectId,
+          mediaId: { in: mediaIds },
+        },
+      }),
       prisma.media.deleteMany({
         where: {
           id: { in: mediaIds },
```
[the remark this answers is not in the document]
wochinge: pre-existing issue.

Very unlikely to happen + would be very costly to fix.

wochinge:
@claude review

packages/shared/prisma/migrations/20260612000000_add_dataset_item_media/migration.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE "media" ADD COLUMN "retained_by_dataset_at" TIMESTAMP(3);
+
+-- CreateTable
+CREATE TABLE "dataset_item_media" (
+    "id" TEXT NOT NULL,
+    "project_id" TEXT NOT NULL,
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "media_id" TEXT NOT NULL,
+    "dataset_id" TEXT NOT NULL,
```
wochinge, line 8: strictly not needed - makes it just much simpler to delete all media for a given dataset

packages/shared/src/server/repositories/dataset-items.ts
```diff
@@ -350,35 +356,61 @@ export async function upsertDatasetItem(
...
-        item = res;
-      } else {
-        const res = await prisma.datasetItem.create({
-          data: {
-            ...itemData,
-            datasetId: dataset.id,
-            projectId: props.projectId,
-          },
-        });
+      await prisma.$transaction(async (tx) => {
```
wochinge, line 392: most of the diff here due to putting everything in one transaction so there is no issues with race conditions

web/src/features/datasets/components/EditDatasetItemDialog.tsx
```diff
@@ -99,21 +100,9 @@ export const EditDatasetItemDialog = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [datasetItem?.id, open]);
 
-  const inputValue = form.watch("input");
```
wochinge, line 102: refactored this a bit as everything as re-rendering if one of the item fields changed it's value (not great when typing 😆 )

web/src/components/ui/CodeJsonViewer.tsx
```diff
@@ -150,7 +150,7 @@ export function JSONView(props: {
           <div className="text-muted-foreground my-1 px-0 py-1 text-xs">
             Media
           </div>
-          <div className="flex flex-wrap gap-2 p-4 pt-1">
+          <div className="flex flex-wrap gap-2 pt-1 pb-4">
```
wochinge, line 153: addressing some alignment issues here:

Image

web/src/components/ui/resizable-image.tsx
wochinge, line 1: changes here are to address the totally misaligned "open in new tab" icon

Image

web/src/features/datasets/hooks/useDatasetItemMediaUpload.ts
```diff
@@ -0,0 +1,126 @@
...
+        const sha256Hash = await sha256Base64(buffer);
+
+        const { mediaId, uploadUrl } = await getUploadUrl.mutateAsync({
+          projectId,
+          contentType: file.type as MediaContentType,
+          contentLength: file.size,
+          sha256Hash,
+        });
+
+        // uploadUrl is null when the content already exists (dedupe by hash)
```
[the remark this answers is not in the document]
wochinge: In my opinion fine - I wouldn't want to block the user to quit things and we just lose a bit of storage worst case

worker/src/features/traces/processClickhouseTraceDelete.ts
```diff
@@ -95,32 +75,78 @@ const deleteMediaItemsForTraces = async (
...
-        where: {
-          projectId,
-          id: {
-            in: orphanedMedia.map((f) => f.id),
-          },
-        },
-      });
+      s3DeletedMediaIds.push(...orphanedMedia.map((m) => m.id));
     }
   }
```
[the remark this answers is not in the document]
wochinge: edge case - not gonna make this query even heavier

fern/apis/server/definition/commons.yml
```diff
@@ -638,6 +638,49 @@ types:
       datasetName: string
       createdAt: datetime
       updatedAt: datetime
+      mediaReferences:
+        type: optional<list<DatasetItemMediaReference>>
```
hassiebp: nit: why optional? could be simply empty list

fern/apis/server/definition/commons.yml
```diff
@@ -638,6 +638,49 @@ types:
...
+        docs: The dataset item field containing the reference
+      referenceString:
+        type: string
+        docs: The Langfuse media reference string, e.g. `@@@langfuseMedia:type=image/png|id=...|source=bytes@@@`
+      jsonPath:
+        type: string
+        docs: JSONPath of the string holding the reference within the field, e.g. `$['image']`
+      media:
+        type: nullable<DatasetItemMediaReferenceMedia>
+        docs: The resolved media record. Null if the referenced media does not exist or has not been uploaded successfully.
```
hassiebp: Why are we not blocking item creation if media upload failed?

fern/apis/server/definition/dataset-items.yml
```diff
@@ -19,6 +19,12 @@ service:
       path-parameters:
         id:
           type: string
+      request:
+        name: GetDatasetItemRequest
+        query-parameters:
+          includeMediaReferences:
```
hassiebp: why do we need this option? Why not always resolve?

fern/apis/server/definition/dataset-items.yml
```diff
@@ -38,6 +44,9 @@ service:
               ISO 8601 timestamp (RFC 3339, Section 5.6) in UTC (e.g., "2026-01-21T14:35:42Z").
               If provided, returns state of dataset at this timestamp.
               If not provided, returns the latest version. Requires datasetName to be specified.
+          includeMediaReferences:
```
hassiebp: again, why not resolving this by default, would love to understand reasoning

fern/apis/server/definition/media.yml
```diff
@@ -86,8 +86,8 @@ types:
         type: string
         docs: The SHA-256 hash of the media record
       field:
-        type: string
-        docs: The trace / observation field the media record is associated with. This can be one of `input`, `output`, `metadata`
+        type: optional<string>
+        docs: The trace / observation field the media record is associated with. This can be one of `input`, `output`, `metadata`. Required if traceId is provided, ignored otherwise.
```
hassiebp: Field should be also required for dataset item requests, no?

packages/shared/prisma/migrations/20260612000000_add_dataset_item_media/migration.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "dataset_item_media_item_version_field_path_key" ON "dataset_item_media"("project_id", "dataset_item_id", "dataset_item_valid_from", "field", "json_path");
+
+-- CreateIndex
+CREATE INDEX "dataset_item_media_project_id_media_id_idx" ON "dataset_item_media"("project_id", "media_id");
+
+-- CreateIndex
+CREATE INDEX "dataset_item_media_project_id_dataset_id_idx" ON "dataset_item_media"("project_id", "dataset_id");
```
hassiebp, line 32: does it make sense from our query patterns to add dataset_item_id to composite index?

packages/shared/prisma/migrations/20260612000000_add_dataset_item_media/migration.sql
```diff
@@ -0,0 +1,28 @@
+-- AlterTable
+ALTER TABLE "media" ADD COLUMN "retained_by_dataset_at" TIMESTAMP(3);
```
hassiebp: Do we really need this column? Could'nt we get away with a join with dataset_item_media?

packages/shared/src/server/media-deletion.ts
```diff
@@ -103,6 +111,12 @@ export async function deleteMediaFiles(params: {
           mediaId: { in: mediaIds },
         },
       }),
+      prisma.datasetItemMedia.deleteMany({
```
hassiebp, line 142: We should not delete media if referenced by a dataset item, and we should also not remove the link here. We should only delete media if dataset is deleted.

packages/shared/prisma/migrations/20260612000000_add_dataset_item_media/migration.sql
wochinge: Yes - otherwise the media retention queries are super slow (did some checks on this last week because I had the conversation with Max as well) https://linear.app/langfuse/document/datasets-multimodal-support-rfc-77134903b483#comment-9fe275ea

fern/apis/server/definition/dataset-items.yml
wochinge: If included, we need to fetch all related media per item _and_ generate signed S3 links for all them. The signing doesn't require network but still overhead that I wanted to avoid initially. No strong opinion though.

For single gets we can make `true` the default but that be slightly inconsistent in my opinion.

fern/apis/server/definition/media.yml
wochinge: we derive that when we actually see the payload in the dataset item create/update endpoint. There we can actually check the payload and and don't need to trust the client on it.

packages/shared/prisma/migrations/20260612000000_add_dataset_item_media/migration.sql
wochinge: this one is only used for clean ups of entire datasets. So should be fine (in any case cheap to change as I don't expect this table to get huge)

fern/apis/server/definition/commons.yml
wochinge: that's because it depends on the query param `includeMediaReferences`

packages/shared/src/server/media-deletion.ts
wochinge: Yes. The function you're looking at is handling 
- project deletes though -> We should do a clean cleanup here
- retention deletes -> in this case we filter out dataset referenced media beforehand (see `findExpiredMediaByProjectId`)

fern/apis/server/definition/commons.yml
wochinge: we're doing that. The comment was overly defensive for some edge case (there is some race conditions that could in theory happen if a user uploads media (but not yet linked to a dataset) that is just flagged for retention deletion and then linked to a dataset before the actual deletion query runs). Fixed the comment and enforcing non nullability

fern/apis/server/definition/dataset-items.yml
hassiebp: I think we should drop this option, keeps interface clean and the overhead we have is manageable:
- we have an index on dataset_item_media to quickly find media for item
- presigned URL generation is cheap

hassiebp: What's expensive is if actual media bytes are involved, and they are not here. Would you follow my reasoning?

wochinge: Yes, fine. In the proposal I overestimated the performance implications of the signing somewhat. Will adapt.

fern/apis/server/definition/dataset-items.yml
```diff
@@ -19,6 +19,12 @@ service:
       path-parameters:
         id:
           type: string
+      request:
+        name: GetDatasetItemRequest
+        query-parameters:
+          includeMediaReferences:
+            type: optional<boolean>
+            docs: If true, resolve Langfuse media references in input, expectedOutput, and metadata to signed download URLs and include them as mediaReferences. Defaults to false.
```
hassiebp: aligned: to be removed

wochinge: done

fern/apis/server/definition/media.yml
wochinge: now required

worker/src/queues/projectDelete.ts
```diff
@@ -39,6 +39,8 @@ export const projectDeleteProcessor: Processor = async (
 
   logger.info(`Deleting ${projectId} in org ${orgId}`);
 
+  await deleteMediaLinkRowsByProjectId({ projectId });
```
hassiebp, line 42: why did we move this to prior to s3 / pg delete? it will now try to delete a lot more columns than before, before it was just a safeguard for dangling observationMedia / traceMedia rows that theoretically existed due to the dropped FK constraint

worker/src/features/media-retention-cleaner/index.ts
```diff
@@ -133,9 +133,17 @@ export class MediaRetentionCleaner extends PeriodicExclusiveRunner {
...
+      -- (null validFrom) are sweepable, matching deleteMediaFiles. Counting them
+      -- here would hide projects whose only expired media is abandoned uploads.
+      LEFT JOIN dataset_item_media dim
+        ON dim.project_id = m.project_id
+        AND dim.media_id = m.id
+        AND dim.dataset_item_valid_from IS NOT NULL
       WHERE p.retention_days > 0
         AND p.deleted_at IS NULL
         AND m.created_at <= NOW() - (p.retention_days || ' days')::interval
+        AND dim.media_id IS NULL
```
hassiebp, line 146: I don't understand why we need this condition, could you explain?

packages/shared/src/server/services/blobStorageEndpointValidation.ts
```diff
@@ -15,8 +15,15 @@ const STRICT_BLOB_STORAGE_ENDPOINT_WHITELIST: OutboundUrlValidationWhitelist = {
   ip_ranges: [],
 };
```
hassiebp, line 17: why are we touching this file?

packages/shared/src/server/media-deletion.ts
```diff
@@ -86,33 +92,82 @@ export async function deleteMediaFiles(params: {
...
+    const deletableMediaIds = deletableBatch.map((f) => f.id);
 
-    await storageClient.deleteFiles(batch.map((f) => f.bucketPath));
+    await storageClient.deleteFiles(deletableBatch.map((f) => f.bucketPath));
     await prisma.$transaction([
       prisma.traceMedia.deleteMany({
         where: {
           projectId,
-          mediaId: { in: mediaIds },
+          mediaId: { in: deletableMediaIds },
```
hassiebp: I think we can keep this to be all `mediaIds` and not just `deletableMediaIds` as we _do_ want to delete the trace / observation media links together with the traces/observations, but only keep the underlying media and datasetItemMedia association row if it was used in a dataset item

packages/shared/src/server/services/blobStorageEndpointValidation.ts
wochinge: to ease testing locally (otherwise it requires that `https` is used)

worker/src/features/media-retention-cleaner/index.ts
wochinge: We can only clean up media that has no attached dataset item.
Above we're doing a left join that implies "every row that has no `dataset_item_media.media.id` attached (and fulfills the other conditions) is deletable

worker/src/queues/projectDelete.ts
wochinge: `deleteMediaFiles` deliberately skips media that’s still associated with a dataset item (the retention guard). So on project delete we have to clear the associations first, otherwise that media is skipped and its S3 files are orphaned (the project-delete cascade will eventually drop the PG media rows but not the S3 objects). 

The media cleanup depends on the association cleanup running first - hence we needed to move it up.

packages/shared/src/server/media-deletion.ts
wochinge: agree - good catch!

wochinge: done

hassiebp (approved):

worker/src/features/datasets/processClickhouseDatasetDelete.ts
```diff
@@ -20,6 +21,11 @@ export const processClickhouseDatasetDelete = async (
   try {
     switch (deletionType) {
       case "dataset":
+        // Always drop the dataset_item_media link rows; no FK cascades them.
+        await deleteDatasetMediaLinksByDatasetId({
+          projectId,
+          datasetId,
+        });
         await deleteDatasetRunItemsByDatasetId({ projectId, datasetId });
         break;
```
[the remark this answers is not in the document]
wochinge: accepted - that's the case for any retention 0 project atm

PR #14264, opened 2026-06-15
feat(evals): reapply evaluator deletion

wochinge:
## Summary

- Reapplies evaluator deletion after the revert of PR #14228, restoring UI, unstable public API, and MCP deletion support for user-maintained evaluators.
- Keeps deletion blocked while evaluation rules still reference the evaluator family and preserves the Langfuse-managed evaluator guard.
- Drops the `job_executions.job_template_id` foreign key instead of adding a large index, with a `30s` scoped `lock_timeout` so the migration fails fast rather than waiting indefinitely for an exclusive table lock.

## Linear

- [LFE-10292](https://linear.app/langfuse/issue/LFE-10292/allow-deleting-evaluators)

## Major Decisions

- `job_template_id` remains historical metadata on `job_executions`; the Prisma relation and FK are removed because job executions are created internally and current queries do not depend on that relation.
- The FK-drop migration uses `SET LOCAL lock_timeout = '30s'` inside an explicit transaction to bound lock acquisition while allowing short write bursts to clear.

## Review Focus

- Migration behavior on `job_executions`: metadata-only FK drop, but still requires a brief exclusive lock with a 30s acquisition timeout.
- Reapplied evaluator deletion surfaces: UI, unstable public API, MCP, and repository/deletion service tests.

This PR reapplies evaluator (eval template family) deletion across the UI, unstable public API, and MCP surfaces, originally reverted in #14228. It also drops the now-unnecessary `job_executions.job_template_id` foreign key via a migration that scopes a 30-second lock timeout.

- **New deletion service** (`evalTemplateDeletion.ts`): resolves all versions of a template family by `(projectId, name, type)`, acquires a `FOR UPDATE` row lock to serialise against concurrent rule creation, re-checks that no `JobConfiguration` still references any version inside the transaction, then deletes all versions atomically.
- **Three entry points**: tRPC `deleteEvalTemplate` mutation (session audit log), public API `DELETE /unstable/evaluators/{evaluatorId}` (API-key audit log), and a new MCP `deleteEvaluator` tool — all sharing a single `deletePublicEvaluator` service and consistent error mapping (403 for managed evaluators, 409 while rules still reference the family).
- **Migration**: drops the FK with `SET LOCAL lock_timeout = '30s'` to bound lock-acquisition wait time; the `IF EXISTS` guard makes it re-entrant.
</details>

Safe to merge. The core deletion logic is transactionally sound, authorization checks are in place on all three surfaces, and the migration's lock-timeout guard is appropriate for a production table.

The `FOR UPDATE` lock strategy correctly handles the concurrent rule-creation race, the FK drop migration works as intended, and error mapping is consistent across tRPC/API/MCP. Two non-blocking observations: the migration's explicit `BEGIN`/`COMMIT` interacts with Prisma's default transaction wrapping in a way that emits a PostgreSQL warning on every deploy run, and INACTIVE job configs block deletion while the UI and error messages label them as "running evaluators", which could confuse users who have already deactivated their evaluation rules.

migration.sql (double-transaction interaction with Prisma), evalTemplateDeletion.ts (INACTIVE config wording in blocker message)
</details>

```mermaid
sequenceDiagram
    participant Client as Client (UI / API / MCP)
    participant Router as tRPC/API/MCP handler
    participant Svc as deleteEvalTemplateFamily
    participant DB as PostgreSQL (transaction)

    Client->>Router: DELETE evaluator (projectId, evalTemplateId)
    Router->>Router: Auth / RBAC check
    Router->>Svc: deleteEvalTemplateFamily(prisma, projectId, evalTemplateId)
    Svc->>DB: findEvalTemplateById(id) — no tx
    DB-->>Svc: template (or null)
    Svc->>Svc: guard: not found / wrong project / managed → throw
    Svc->>DB: BEGIN transaction
    DB->>DB: SELECT ... FOR UPDATE (lock all family versions)
    DB->>DB: findEvalTemplateFamilyVersions
    DB->>DB: findJobConfigurationsReferencingEvalTemplates
    alt referencing configs exist
        DB-->>Svc: referencingConfigs (ACTIVE or INACTIVE)
        Svc-->>Router: throw LangfuseConflictError (409)
        Router-->>Client: 409 Conflict
    else no references
        DB->>DB: deleteMany(evalTemplateIds)
        DB-->>Svc: COMMIT — returns deletedVersions
        Svc->>Svc: auditLog per deleted version (if auditScope)
        Svc-->>Router: deletedVersions
        Router->>Router: auditLog (session) — tRPC only
        Router-->>Client: 200 OK / success toast
    end
```
</details>

`````markdown

---

### Issue 1 of 2
packages/shared/prisma/migrations/20260612000000_drop_job_executions_job_template_id_fkey/migration.sql:1-5
**Migration transaction nesting with Prisma's default wrapper**

Prisma wraps each migration in its own `BEGIN`/`COMMIT` by default. Because the migration file already contains its own `BEGIN`/`COMMIT`, the inner `BEGIN` becomes a PostgreSQL no-op (warning only), `SET LOCAL lock_timeout` applies to Prisma's outer transaction as intended, the inner `COMMIT` commits Prisma's outer transaction, and Prisma's subsequent `COMMIT` call lands outside a transaction — PostgreSQL emits `WARNING: there is no transaction in progress` and treats it as a no-op. The migration works, but the no-op `COMMIT` at the end may surface confusing warnings in migration logs.

If the intent is to own the full transaction boundary, adding `-- This migration does not use a transaction` at the top would make Prisma skip its wrapping and let the explicit `BEGIN`/`COMMIT` here be the sole transaction. Without that pragma, the inner `COMMIT` terminates Prisma's transaction earlier than Prisma expects.

### Issue 2 of 2
web/src/features/evals/server/evalTemplateDeletion.ts:115-130
**INACTIVE job configs block deletion but are labelled "running"**

`findJobConfigurationsReferencingEvalTemplates` does not filter by `status`, so an INACTIVE `JobConfiguration` also prevents deletion. This is the correct data-integrity choice (a nulled `evalTemplateId` on re-activation would silently break the config), but the UI surfaced by `delete-eval-template-button.tsx` and the error message built in `buildInUseErrorMessage` both use the label `running evaluator(s)`. A user who has deactivated all their evaluation rules will see "Cannot delete — used by N running evaluator(s)" even though none are actually active, making the blocker hard to diagnose.

Consider either labelling the referencing configs without the "running" qualifier, or filtering to ACTIVE-only for both the advisory query and the transactional check (noting that the latter would leave INACTIVE configs with a dangling `evalTemplateId` if the user deletes the template and then tries to re-activate the config).

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(evals): reapply evaluator deletion"](https://github.com/langfuse/langfuse/commit/75f1cf79bb3fd534b048f28dfcabbe83a5f36740) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37692179)</sub>

### CODE REVIEW

File: packages/shared/prisma/migrations/20260612000000_drop_job_executions_job_template_id_fkey/migration.sql [hunk 1]
```diff
@@ -0,0 +1,5 @@
+-- DropForeignKey
+BEGIN;
+SET LOCAL lock_timeout = '30s';
```
wochinge: rather fail early instead of blocking forever

PR #14300, opened 2026-06-16
feat(media): external endpoint for media download presigned URLs

### LINKED ISSUE #14299: bug: media download URLs not browser accessible on self hosted MinIO (no external endpoint option)
### Describe the bug

In a self hosted Langfuse instance that uses MinIO as the in cluster S3 backend, media uploads and media downloads share a single endpoint set by `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT`. That one value is used both for server side S3 operations and for presigning upload (PUT) URLs that the SDK or agent consumes, and for presigning download (GET) URLs that are handed to the browser (the Langfuse UI media player) and to API consumers.

Because it is one value, self hosted operators are forced into a no win choice:

* Set it to the internal cluster URL (for example `http://minio.namespace.svc.cluster.local:9000`). The in cluster agent uploads fine, but the browser cannot resolve that DNS name, and even with a hosts file entry plus port forwarding the UI player is blocked by mixed content (the UI is HTTPS, the media URL is HTTP).
* Set it to an external HTTPS URL. The browser works, but routing in cluster server and SDK traffic out and back is awkward and can break the upload path.

SigV4 presigned URLs bind the signature to the Host header only, so the download host can safely differ from the internal one as long as it is the host the URL was signed against. That is exactly what a separate external endpoint enables.

This is the same self hosting gap that several earlier reports hit and that were closed as configuration or documentation questions (#4369, #4717, #6066, #6188). There is currently no way to keep uploads on the internal endpoint while serving browser facing download URLs from a public host.

### Precedent in the codebase

Langfuse batch export already solves this exact split. It supports a separate external endpoint via `LANGFUSE_S3_BATCH_EXPORT_EXTERNAL_ENDPOINT`, so the presigned download URL handed to the user points at a public host while internal operations use the internal endpoint. Media has no equivalent.

### Proposed fix

Add an external endpoint option for media that mirrors batch export:

* New environment variable `LANGFUSE_S3_MEDIA_UPLOAD_EXTERNAL_ENDPOINT` (naming matches the batch export convention).
* Download (GET) presigned URLs returned to clients are signed against the external endpoint when it is set.
* Upload (PUT) presigning and all server side S3 operations continue to use the internal `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT`.
* If the external variable is unset, behaviour is identical to today (fully backward compatible).

### Steps to reproduce

1. Self host Langfuse with MinIO as the media backend and set `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT` to the internal cluster URL.
2. Send a trace with media (for example a multimodal image) so an object lands in MinIO.
3. Open the trace in the web UI and try to view the media in the player.
4. Observe that the media URL points at the internal host, which the browser cannot resolve, so the player fails to load (or is blocked as mixed content when the UI is HTTPS).

### Langfuse Cloud or self hosted?

Self hosted.

### If self hosted, what version are you running?

Observed on v3.179. The single endpoint behaviour is still present on the current main branch.

Not SDK specific. It affects any media viewed in the web UI or fetched through `GET /api/public/media/:mediaId` on a self hosted instance whose S3 endpoint is reachable only from inside the cluster.

### Additional information

A working implementation is attached as a draft pull request. It adds `LANGFUSE_S3_MEDIA_UPLOAD_EXTERNAL_ENDPOINT`, points the media download presign call sites at a client built against the external endpoint, keeps upload presigning and server operations on the internal endpoint, and was verified against a real MinIO: a GET URL signed against a different external host returns the object with HTTP 200 when that host fronts the store.

Yes, a draft pull request is attached.

Oscmage:
## What does this PR do?

Adds an external endpoint option for media download presigned URLs, mirroring the existing batch export split, so self hosted deployments (e.g. MinIO in a cluster) can serve browser facing media downloads from a public host while keeping uploads and server side S3 operations on the internal endpoint.

New optional env var `LANGFUSE_S3_MEDIA_UPLOAD_EXTERNAL_ENDPOINT` (named to match `LANGFUSE_S3_BATCH_EXPORT_EXTERNAL_ENDPOINT`):

- Download (GET) presigned URLs are signed against the external endpoint when set.
- Upload (PUT) presigning and all server side S3 operations keep using the internal `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT`.
- Unset leaves behaviour identical to today (backward compatible).

Implementation notes:

- `getMediaDownloadStorageServiceClient` builds a client with `externalEndpoint` and falls back to the internal client when the var is unset. The three media download presign call sites (`mediaService.getMedia` and the two `media` tRPC router reads) use it; `getSignedUploadUrl` stays on the internal client.
- A dedicated download client is needed because `S3StorageService` signs both GET and PUT through its `signedUrlClient`, so a single shared client with an external endpoint would push uploads onto the external host too.
- The shared `getS3MediaStorageClient` (worker side cleanup/retention) is untouched; it never hands download URLs to clients, matching how the batch export external var lives only where presigning happens.

Testing:

- Server unit test runs the real S3 presigner and asserts the resulting URL host (download external, upload internal, fallback internal).
- Verified against a real MinIO: a GET URL signed against a different external host returns the object with HTTP 200 when that host fronts the store.

Fixes #14299

## Type of change

## Mandatory Tasks


sumerman:
@claude review

PR #14325, opened 2026-06-17
feat(credentials): generic short-lived credential interface and refresh manager

Oscmage:
## What does this PR do?

Introduces the generic, provider agnostic core for short lived (rotating) infrastructure credentials. This is the foundation for the approach we discussed in #14278 (cc @Steffen911 @jannikmaierhoefer). Redis and Postgres support follow as separate PRs on top of this, as agreed in that thread.

- `ManagedCredentialProvider`: a small interface a backend implements to mint a fresh token (`fetchToken()` returning `{ token, expiresOnTimestamp }`), plus an optional `username` principal.
- `RefreshingTokenManager`: fetches the token, refreshes it ahead of expiry (configurable ratio of the token lifetime), and notifies subscribers via `onRefresh` so a live connection can re-authenticate when the token rotates.

Foundation only: no new dependencies, no providers, and nothing wired into Redis or Postgres yet. It is inert until a consumer uses it, and the surface is kept intentionally minimal (only what the first consumer, Redis, needs).

## Follow ups (separate PRs)

- Redis support: an `azure-managed-identity` provider plus the ioredis binding and env vars (ready to open once this lands).
- Postgres support: a `pg` driver adapter password callback that reuses the same manager.

Refs #14278

## Type of change

## Mandatory Tasks

This PR introduces the foundational infrastructure for short-lived, rotating credentials in the shared package — a `ManagedCredentialProvider` interface for minting tokens and a `RefreshingTokenManager` that pre-emptively refreshes tokens ahead of expiry and fans out to subscribers. No provider implementations or Redis/Postgres wiring are included; the new code is inert until consumed.

- **`RefreshingTokenManager`**: fetches an initial token via `start()`, schedules a proactive refresh at `expirationRefreshRatio × remainingTTL`, retries with a 5 s back-off on failure, and notifies all `onRefresh` listeners after each successful rotation.
- **`types.ts` / `index.ts`**: minimal, well-scoped interfaces (`ManagedAccessToken`, `ManagedCredentialProvider`) exported through the shared server barrel.
- **Tests**: cover the happy path (initial fetch, timer-driven refresh, subscriber notification, `stop()` disarming), but the retry-on-failure branch is untested.
</details>

Safe to merge as inert foundation code, but the stopped-flag bug means any consumer that restarts the manager after stopping it would silently lose all future token refreshes.

The start() method never resets this.stopped, so a stop() then start() sequence returns a valid initial token but arms no refresh timer — the token quietly expires without rotation or subscriber notification.

RefreshingTokenManager.ts — the start() method and constructor options handling both need attention before the first consumer lands.
</details>

`````markdown

---

### Issue 1 of 3
packages/shared/src/server/auth/credentials/RefreshingTokenManager.ts:32-37
`stopped` flag is never reset in `start()`, so calling `stop()` followed by `start()` on the same instance silently disarms refresh scheduling. The initial token is returned successfully, but `scheduleRefresh` returns immediately due to `if (this.stopped) return`, leaving the token to expire without ever being refreshed or subscribers being notified.

```suggestion
  // Fetch the first token and arm the refresh-ahead timer.
  public async start(): Promise<ManagedAccessToken> {
    this.stopped = false;
    const token = await this.provider.fetchToken();
    this.scheduleRefresh(token);
    return token;
  }
```

### Issue 2 of 3
packages/shared/src/server/auth/credentials/RefreshingTokenManager.ts:28-29
`expirationRefreshRatio` is not validated. A value greater than `1.0` would schedule the refresh after the token has already expired (e.g. `ratio = 1.5` on a 1-hour token fires the timer 90 minutes from now). A value of `0` collapses to `MIN_REFRESH_DELAY_MS` causing a 1-second tight loop. Both are silent misconfigurations.

```suggestion
    const ratio =
      options.expirationRefreshRatio ?? DEFAULT_EXPIRATION_REFRESH_RATIO;
    if (ratio <= 0 || ratio >= 1) {
      throw new RangeError(
        `expirationRefreshRatio must be in the range (0, 1), got ${ratio}`,
      );
    }
    this.expirationRefreshRatio = ratio;
```

### Issue 3 of 3
packages/shared/src/server/auth/credentials/RefreshingTokenManager.test.ts:63-71
**Missing tests for failure-and-retry path** — There is no test covering the `catch` branch in `refreshFromTimer`. Concretely: (1) a failing `fetchToken` should log a warning and schedule a `RETRY_DELAY_MS` retry rather than re-arming the normal expiration timer, and (2) a subsequent successful fetch after a retry should still notify subscribers and reschedule correctly. Without these, a regression in the retry branch would be invisible.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/credential..."](https://github.com/langfuse/langfuse/commit/a8d99255c06ab30bba512a8fe62aacd2ebc562d4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38133642)</sub>


Oscmage:
@Steffen911 Let me know if you need something more before you can take a look at this PR!

Steffen911 (approved):
LGTM, thank you @Oscmage

PR #14326, opened 2026-06-17
feat(web): add sentinel redirect for /project/~/[...path]

marksalpeter:
## Summary

There is no way to link directly to "whatever project the user can access first" — every deep link requires a hard-coded project ID.

To achieve this we:
- Added a sentinel route at `/project/~/[[...path]]` that redirects server-side to the first project the user has access to, preserving the full sub-path
- Redirected unauthenticated users to sign-in with a `callbackUrl` so the redirect completes after login

## Verification

- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [ ] Browser review

## Test plan

- [ ] Navigate to `/project/~/` — confirm redirect to first accessible project dashboard
- [ ] Navigate to `/project/~/traces` — confirm redirect preserves sub-path
- [ ] Navigate while unauthenticated — confirm redirect to sign-in with callbackUrl
- [ ] Navigate with a real project UUID — confirm no interference with normal routing

This PR adds a single new Next.js page at `web/src/pages/project/~/[[...path]].tsx` that acts as a server-side sentinel redirect, allowing callers to link to \"the first project the user can access\" without knowing its ID in advance. An optional numeric prefix in the path (e.g. `/project/~/2/traces`) selects the Nth accessible project (0-based), and unauthenticated users are bounced to sign-in with a `callbackUrl` so the redirect completes after login.

- Adds `getServerSideProps` that reads the session's `organizations[].projects` list, optionally extracts a numeric index from the first path segment, and issues a non-permanent redirect to `/project/{id}/[...subpath]`.
- Contains a logic error where an explicit index of `0` is not stripped from the forwarded path, and soft-deleted projects are not excluded from the candidate pool.
</details>

The redirect logic works for the common case but has a defect when an explicit index of 0 is supplied — the digit segment leaks into the forwarded path rather than being consumed as the index.

The explicit-0 path bug means /project/~/0/traces redirects to /project/{id}/0/traces instead of /project/{id}/traces, which is a silent misbehavior that would be hard to debug. A second, smaller concern is that soft-deleted projects aren't excluded from the candidate list. The fix for the primary issue is a one-liner, but it should be verified before merging.

web/src/pages/project/~/[[...path]].tsx — the index extraction and path-slice logic needs attention.
</details>

```mermaid
sequenceDiagram
    participant Browser
    participant NextJS as Next.js SSR /project/~/[[...path]]
    participant Auth as getServerAuthSession
    participant Session as Session (orgs/projects)

    Browser->>NextJS: GET /project/~/[optional-index]/[...subpath]
    NextJS->>Auth: getServerAuthSession(req, res)
    Auth-->>NextJS: session (or null)

    alt No session
        NextJS-->>Browser: "302 to /auth/sign-in?callbackUrl=..."
        Browser->>NextJS: (after sign-in) GET /project/~/[...path]
    end

    NextJS->>Session: flatMap orgs to projects
    Session-->>NextJS: allProjects[]

    alt rawPath[0] matches digit-only
        NextJS->>NextJS: "index = parseInt(rawPath[0]), path = rawPath.slice(1)"
    else no numeric prefix
        NextJS->>NextJS: "index = 0, path = rawPath"
    end

    alt project at index exists
        NextJS-->>Browser: "302 to /project/{id}/[...path]"
    else no project at index
        NextJS-->>Browser: 302 to /
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Browser
    participant NextJS as Next.js SSR /project/~/[[...path]]
    participant Auth as getServerAuthSession
    participant Session as Session (orgs/projects)

    Browser->>NextJS: GET /project/~/[optional-index]/[...subpath]
    NextJS->>Auth: getServerAuthSession(req, res)
    Auth-->>NextJS: session (or null)

    alt No session
        NextJS-->>Browser: "302 to /auth/sign-in?callbackUrl=..."
        Browser->>NextJS: (after sign-in) GET /project/~/[...path]
    end

    NextJS->>Session: flatMap orgs to projects
    Session-->>NextJS: allProjects[]

    alt rawPath[0] matches digit-only
        NextJS->>NextJS: "index = parseInt(rawPath[0]), path = rawPath.slice(1)"
    else no numeric prefix
        NextJS->>NextJS: "index = 0, path = rawPath"
    end

    alt project at index exists
        NextJS-->>Browser: "302 to /project/{id}/[...path]"
    else no project at index
        NextJS-->>Browser: 302 to /
    end
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/pages/project/~/[[...path]].tsx:27-32
The `index === 0` check conflates two distinct cases: (a) no numeric prefix was provided, and (b) the user explicitly passes `0` as the index. When `rawPath` is `['0', 'traces']`, `index` is `0` and `path` stays as `['0', 'traces']`, so the redirect lands at `/project/{id}/0/traces` instead of `/project/{id}/traces`. Only case (b) should strip the first segment.

```suggestion
  const rawPath = ctx.params?.path ?? [];
  const hasNumericPrefix =
    rawPath.length > 0 && /^\d+$/.test(rawPath[0]!);
  const index = hasNumericPrefix ? parseInt(rawPath[0]!, 10) : 0;
  const path = hasNumericPrefix ? rawPath.slice(1) : rawPath;
```

### Issue 2 of 2
web/src/pages/project/~/[[...path]].tsx:25
**Soft-deleted projects included in redirect target pool**

`flatMap` collects all projects from the session, including those with a non-null `deletedAt`. If the first (or Nth) accessible project happens to be soft-deleted, the user gets redirected to that project's dashboard, which will likely render an error page rather than the intended content. Consider filtering with `.filter((p) => !p.deletedAt)` before indexing.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(web): drop index support from senti..."](https://github.com/langfuse/langfuse/commit/d0aba5787275c0c01305eb3ecb5faf80c13f1923) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38137163)</sub>

### DIFF
--- a/web/src/pages/project/~/[[...path]].tsx
+++ b/web/src/pages/project/~/[[...path]].tsx
@@ -65,7 +65,7 @@ const signInRedirect = async ({ getSession, resolvedUrl }: SentinelContext) => {
     );
 };
 
-/** projectRedirect resolves to the last visited project, the first accessible project, or home if no projects are availalbe. */
+/** projectRedirect resolves to the last visited project, the first accessible project, or home if no projects are available. */
 const projectRedirect = async ({
   cookie,
   origin,
--- a/web/src/pages/project/~/[[...path]].tsx
+++ b/web/src/pages/project/~/[[...path]].tsx
@@ -6,7 +6,7 @@ export default function RedirectToFirstProject() {
   return null;
 }
 
-/** getServerSideProps redirects to the Nth project the user has access to, preserving the sub-path. */
+/** getServerSideProps redirects to the first project the user has access to, preserving the sub-path. */
 export const getServerSideProps: GetServerSideProps = async (ctx) => {
   const session = await getServerAuthSession({
     req: ctx.req,
@@ -22,21 +22,16 @@ export const getServerSideProps: GetServerSideProps = async (ctx) => {
     };
   }
 
-  const allProjects = session.user.organizations.flatMap((org) => org.projects);
+  const firstProject = session.user.organizations
+    .flatMap((org) => org.projects)
+    .at(0);
 
-  const rawPath = ctx.params?.path ?? [];
-  const index =
-    rawPath.length > 0 && /^\d+$/.test(rawPath[0]!)
-      ? parseInt(rawPath[0]!, 10)
-      : 0;
-  const path = index === 0 ? rawPath : rawPath.slice(1);
-
-  const project = allProjects.at(index);
-  if (!project) {
+  if (!firstProject) {
     return { redirect: { destination: "/", permanent: false } };
   }
 
-  const destination = ["/project", project.id, ...path].join("/");
+  const path = (ctx.params?.path as string[] | undefined) ?? [];
+  const destination = ["/project", firstProject.id, ...path].join("/");
 
   return { redirect: { destination, permanent: false } };
 };
--- a/web/src/pages/project/~/[[...path]].tsx
+++ b/web/src/pages/project/~/[[...path]].tsx
@@ -1,0 +1,42 @@
+import { type GetServerSideProps } from "next";
+
+import { getServerAuthSession } from "@/src/server/auth";
+
+export default function RedirectToFirstProject() {
+  return null;
+}
+
+/** getServerSideProps redirects to the Nth project the user has access to, preserving the sub-path. */
+export const getServerSideProps: GetServerSideProps = async (ctx) => {
+  const session = await getServerAuthSession({
+    req: ctx.req,
+    res: ctx.res,
+  });
+
+  if (!session?.user) {
+    return {
+      redirect: {
+        destination: `/auth/sign-in?callbackUrl=${encodeURIComponent(ctx.resolvedUrl)}`,
+        permanent: false,
+      },
+    };
+  }
+
+  const allProjects = session.user.organizations.flatMap((org) => org.projects);
+
+  const rawPath = ctx.params?.path ?? [];
+  const index =
+    rawPath.length > 0 && /^\d+$/.test(rawPath[0]!)
+      ? parseInt(rawPath[0]!, 10)
+      : 0;
+  const path = index === 0 ? rawPath : rawPath.slice(1);
+
+  const project = allProjects.at(index);
+  if (!project) {
+    return { redirect: { destination: "/", permanent: false } };
+  }
+
+  const destination = ["/project", project.id, ...path].join("/");
+
+  return { redirect: { destination, permanent: false } };
+};
--- a/web/src/pages/project/~/[[...path]].tsx
+++ b/web/src/pages/project/~/[[...path]].tsx
@@ -31,7 +31,11 @@ export const getServerSideProps: GetServerSideProps = async (ctx) => {
   }
 
   const path = (ctx.params?.path as string[] | undefined) ?? [];
-  const destination = ["/project", firstProject.id, ...path].join("/");
+  const destination = [
+    "/project",
+    firstProject.id,
+    ...path.map(encodeURIComponent),
+  ].join("/");
 
   return { redirect: { destination, permanent: false } };
 };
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: web/src/pages/project/~/[[...path]].tsx [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const firstProject = session.user.organizations
+    .flatMap((org) => org.projects)
+    .at(0);
+
+  if (!firstProject) {
+    return { redirect: { destination: "/", permanent: false } };
+  }
+
```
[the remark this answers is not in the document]
Comment (marksalpeter) [reply web/src/pages/project/~/[[...path]].tsx]: this is intended behavoir

web/src/pages/project/~/[[...path]].tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (!session?.user) {
+    return {
+      redirect: {
+        destination: `/auth/sign-in?callbackUrl=${encodeURIComponent(ctx.resolvedUrl)}`,
+        permanent: false,
+      },
+    };
+  }
+
+  const firstProject = session.user.organizations
```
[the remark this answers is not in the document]
Comment (marksalpeter) [reply web/src/pages/project/~/[[...path]].tsx]: Not an issue — the session query in `auth.ts` already filters `deletedAt: { equals: null }` on `organization.projects` before the session is built (see [auth.ts:768-773](https://github.com/langfuse/langfuse/blob/main/web/src/server/auth.ts#L768-L773)), so `flatMap` here can never surface a soft-deleted project.

web/src/pages/project/~/[[...path]].tsx
```diff
@@ -0,0 +1,37 @@
...
+    .at(0);
+
+  if (!firstProject) {
+    return { redirect: { destination: "/", permanent: false } };
+  }
+
+  const path = (ctx.params?.path as string[] | undefined) ?? [];
+  const destination = ["/project", firstProject.id, ...path].join("/");
+
+  return { redirect: { destination, permanent: false } };
```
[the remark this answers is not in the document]
Comment (marksalpeter) [reply web/src/pages/project/~/[[...path]].tsx]: Fixed in 0c59be7: `path.map(encodeURIComponent)` re-encodes each decoded catch-all segment before joining.

web/src/pages/project/~/[[...path]].tsx
```diff
@@ -0,0 +1,41 @@
...
+        destination: `/auth/sign-in?callbackUrl=${encodeURIComponent(ctx.resolvedUrl)}`,
+        permanent: false,
+      },
+    };
+  }
+
+  const firstProject = session.user.organizations
+    .flatMap((org) => org.projects)
+    .at(0);
+
```
Comment (maxdeichmann) [inline web/src/pages/project/~/[[...path]].tsx:79]: - First project could be the demo project. We should remove this.
- How would we deal with RBAC here? There could be a case we forward to a page where the user does not have the permissions to do something. technically monitor-create could be forbidden but we still forward the user. Thus we would need an error state. No blocker from my side.

Comment (maxdeichmann) [reply web/src/pages/project/~/[[...path]].tsx:79]: Actually we could also make sure to have the demo project last in the list.

Comment (marksalpeter) [reply web/src/pages/project/~/[[...path]].tsx:79]: I like having the demo project last in the list. I also think it would be better to set a cookie with the last used project in it. If someone is hopping back and fourth between docs and the app, we don't switch them to a different project.

Comment (maxdeichmann) [reply web/src/pages/project/~/[[...path]].tsx:79]: even better!

Comment (nimarb) [reply web/src/pages/project/~/[[...path]].tsx:79]: 2026-06-17_17 51 09-Google Chrome@2x
longer term a google style enumeration could also work well if the order is stable , i like it!

Comment (marksalpeter) [reply web/src/pages/project/~/[[...path]].tsx:79]: @nimarb I actually had this in the original commit, but I rolled it back because I was always fighting against this in the gcp console (the user order is unstable, so deep links change depending on which accounts you've logged into and in which order). 

I think `project/[project_id/*` is much less ambiguous and much more stable than `project/~/0/...` imho.

PR #14346, opened 2026-06-18
fix(search-bar): always land on a trailing space when commiting the value

nkabardin:

This PR refactors the search composer's "append intent" mechanism — previously a deferred `appendIntent` state flag — into an immediate trailing-space insertion model. All paths that previously set `appendIntent=true` (value pick at end, ArrowRight-at-end, click past text) now directly append a space and move the caret past the last pill; the new `commit(advanceToTrailingSpace)` parameter handles the Enter-at-end case similarly.

- Removes the `appendIntent` React state and all its guard clauses (re-pick deduplication, prefix insertion, selectionchange reset), replacing them with direct trailing-space writes in each gesture handler.
- Updates `commit()` to accept an `advanceToTrailingSpace` flag; when `true` and the caret was at the end of a non-whitespace-terminated committed query, it appends a space and reopens autocomplete instead of closing it.
- Recents are now always surfaced when autocomplete is open (the `!appendIntent` suppression is removed), so a trailing-space landing now shows both field suggestions and recent queries.
</details>

Safe to merge; the trailing-space refactor simplifies state management and all the critical gesture paths look correct. The two observations are design trade-offs rather than data-corruption or crash risks.

The refactor removes a well-tested deferred state mechanism and replaces it with immediate writes. The main open question is whether showing recents at the trailing-space landing (where they were previously suppressed) is the right UX — picking one silently replaces all current filters. The commit guard also has an implicit dependency on async draft normalisation that could silently invert if the commit implementation ever becomes synchronous. Neither issue breaks correctness today, but both are worth a second pair of eyes before landing.

SearchComposer.tsx — specifically the recents memoisation change around line 329 and the committed guard in commit around line 688.
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/search-bar/components/SearchComposer.tsx:329-343
**Recents now replace the full draft from the trailing-space landing**

With `appendIntent` removed, `recents` is populated whenever `autocompleteOpen` is true — including the trailing-space state the user lands in after pressing Enter at the end of a valid query. Picking a recent via `pickOption` (line 718) calls `setDraftWithSelection(option.query, ...)`, which replaces the *entire* current draft with the recent query. Previously the `!appendIntent` guard was explicitly there to prevent this:

> "a recent is a complete query, not a token to append, so showing them there would let a pick silently replace the in-progress draft."

A user who builds `level:error environment:production`, presses Enter, then accidentally clicks a recent from the dropdown would silently lose both filters with no undo prompt. The new justification ("consistent with picking a recent at the start of an empty bar") is reasonable, but the old comment directly anticipated and blocked this scenario. If showing recents here is intentional, it may be worth confirming the decision and updating the UX to provide a more visible signal that a recent pick replaces (rather than appends to) the current query.

### Issue 2 of 2
web/src/features/search-bar/components/SearchComposer.tsx:686-698
**`committed` may still carry a trailing space when the guard fires**

`storeApi.getState().draft` is read synchronously right after `commitToFilterState()`, but the comment elsewhere says normalization happens via a React `useEffect` (`resetTo` effect) — meaning the store draft may still hold `"level:error "` (with its trailing space) at this point. If `committed` ends in whitespace, `!/\s$/.test(committed)` is `false`, so the branch falls into `setAutocompleteOpen(false)` rather than appending a space and reopening.

This matters if `commitToFilterState()` happens to normalise the draft synchronously back to `"level:error"`: the guard would then never fire on the second Enter (it would always pass, perpetually keeping autocomplete open). Conversely, if normalization is always async, the guard fires correctly. The behaviour is correct today under the async assumption, but the correctness of the `advanceToTrailingSpace` branch is silently coupled to whether `commitToFilterState` touches the store draft synchronously. A brief comment explaining why reading the pre-normalisation draft here is intentional would prevent future regressions.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(search-bar): always land on a traili..."](https://github.com/langfuse/langfuse/commit/004f4d4bc89a1022e7ef16b7a58f680936a335a5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38340041)</sub>


nkabardin:
@claude review

PR #14350, opened 2026-06-18
feat(slack): add Slack marketplace install flow

wochinge:
## Summary

- Direct install is a [Slack marketplace requirement](https://docs.slack.dev/slack-marketplace/slack-marketplace-app-guidelines-and-requirements/#direct_install)
- Adds a Slack **Marketplace "install-then-link"** flow: a workspace can be installed from the Marketplace Direct Install URL *before* a Langfuse project is chosen, then linked to a project afterward.
- `slack_integrations.project_id` becomes **nullable** to hold pending (unlinked) installs. Pending rows carry an `expires_at` TTL, the installing Slack user, and a hashed one-time **claim token** so only the person who completed the OAuth (and received the token via the onboarding redirect) can view/link that install. The claim is enforced in the DB query and matched on `teamId` + hash.
- New unauthenticated **Direct Install route** (`/api/public/slack/marketplace-install`) that 302s straight to Slack OAuth via `directInstall` (which also sets the state cookie). The OAuth callback stores a pending row, issues a claim token, and redirects to **`/slack/direct-setup`** — an onboarding page (matching the organizations overview layout) where the user links the install to a project. Backed by `getConnectableProjects` (automations:CUD-scoped, with connection status) and `linkPendingInstallation`.
- The public marketplace URL is also protected by a rate limiter (based on IP). The route just redirects to Slack (no database stuff involved), but better be on the safe side
- Manifest: disable the App Home / Messages tabs (this is an outbound-only app).

https://www.loom.com/share/8fe9c48763c8465d9fd2bcea52ba9d7d

## Linear

- [LFE-10329](https://linear.app/langfuse/issue/LFE-10329/investigate-slack-marketplace-integration)

## Review Focus

- **Security of the public route + claim model** — `marketplace-install` is intentionally unauthenticated (DB-free 302); the claim token gates who can link a pending install.
- **Migration**: `project_id` nullable + new columns (`expires_at`, `installing_slack_user_id`, `claim_token_hash`) and the now-optional `project` relation.


packages/shared/prisma/migrations/20260617000000_slack_integrations_pending_columns/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+-- distinct, so multiple unlinked (pending) rows can coexist.
+ALTER TABLE "slack_integrations" ALTER COLUMN "project_id" DROP NOT NULL;
+
+-- Pending-only columns (NULL once the install is linked to a project).
+ALTER TABLE "slack_integrations" ADD COLUMN "installing_slack_user_id" TEXT;
+ALTER TABLE "slack_integrations" ADD COLUMN "expires_at" TIMESTAMP(3);
+
+-- One-time claim token hash: the raw token is carried only by the onboarding
+-- URL and is required before a pending install can be linked to a project.
+ALTER TABLE "slack_integrations" ADD COLUMN "claim_token_hash" TEXT;
```
wochinge, line 12: Technically we don't need this but avoids issues if 2 users want to connect Slack to their workspaces at the same time and then steal each others pending integrations). Just makes it a bit cleaner

web/src/features/slack/app_manifest.json
```diff
@@ -6,6 +6,11 @@
     "long_description": "The Langfuse Slack App keeps your team in sync by delivering real-time observability and alerts directly to your Slack channels. Whether you’re monitoring AI traces, catching runtime errors, or tracking latency and performance metrics, Langfuse ensures that you’re always in the loop.\r\n\r\nWith flexible configuration per project, you can:\r\n\t•\tGet notified about critical errors and anomalies\r\n\t•\tReceive alerts on trace completions, slow response times, or quality degradations\r\n\t•\tQuickly jump from a Slack message to the exact trace in Langfuse\r\n\t•\tKeep dev, ops, and ML teams aligned without context switching\r\n\r\nStay on top of your LLM applications — without leaving Slack."
   },
   "features": {
+    "app_home": {
```
wochinge, line 9: no needed since we're an app with only outgoing connections. Disabled it also as that's perceived as wrong usage by Slack: https://docs.slack.dev/slack-marketplace/slack-marketplace-app-guidelines-and-requirements/#home

web/src/features/slack/server/oauth-handlers.ts
```diff
@@ -64,23 +55,56 @@ export async function handleCallback(
...
+              res.redirect("/slack/direct-setup");
+              return;
+            }
+            const onboardingUrl = `/slack/direct-setup?team_id=${encodeURIComponent(
+              teamId,
+            )}&team_name=${encodeURIComponent(
+              teamName ?? "",
+            )}&claim=${encodeURIComponent(claimToken)}`;
+            res.redirect(onboardingUrl);
+            return;
```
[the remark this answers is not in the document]
wochinge: addressed

wochinge:
@claude review

ee/package.json
```diff
@@ -34,6 +34,7 @@
     "zod": "^4.3.6"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.6",
```
wochinge: some issues from `main` that bubbled up here

packages/shared/package.json
```diff
@@ -146,6 +146,7 @@
     "zod": "^4.3.6"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.6",
```
maxdeichmann: why do we need all of these?

wochinge: had some unrelated lint issues bubbling up from `main` - this addresses these (see https://github.com/langfuse/langfuse/actions/runs/27767472712/job/82158043268?pr=14350)

packages/shared/prisma/schema.prisma
```diff
@@ -1762,20 +1762,32 @@ model Monitor {
...
+  project   Project? @relation(fields: [projectId], references: [id], onDelete: Cascade)
 
   // Installation details (encrypted using shared encryption utilities)
   teamId    String @map("team_id") // Slack workspace ID
   teamName  String @map("team_name") // Human-readable workspace name
   botToken  String @map("bot_token") // Encrypted bot token for API calls
   botUserId String @map("bot_user_id") // Bot user ID for workspace
 
+  // Pending-only fields (NULL once linked): when an unlinked pending install is
+  // purged, and the hashed one-time claim token required to link it.
```
maxdeichmann, line 1782: nice comments, make this easy to understand what is going on here.

packages/shared/prisma/schema.prisma
```diff
@@ -1762,20 +1762,32 @@ model Monitor {
...
+  expiresAt      DateTime? @map("expires_at")
+  claimTokenHash String?   @map("claim_token_hash")
```
maxdeichmann, line 1784: in the linking process to projects, is there any security issue, where i could just link to any project?

packages/shared/src/server/services/SlackService.ts
```diff
@@ -128,52 +180,108 @@ export function parseSlackInstallationMetadata(
...
+                    teamId,
+                    teamName,
+                    botToken: encrypt(botToken),
+                    botUserId,
+                    expiresAt: new Date(
+                      Date.now() + SLACK_PENDING_INSTALL_TTL_MS,
+                    ),
+                  },
+                }),
+              ]);
```
maxdeichmann: i would serialize these as well to basically lock the two steps.

packages/shared/prisma/schema.prisma
wochinge: That's what came up here: https://github.com/langfuse/langfuse/pull/14350#discussion_r3436410063

Once the "pending" Slack installation is in, the URL could be intercepted in between and then this attacker could link the Slack integration to their project.

That's why we added the claim to the cookie to ensure that only the user who added the pending installation can also link it to a project later.

packages/shared/package.json
maxdeichmann: for this, i just had to merge latest main. maybe i am wrong here. keep inside if needed.

web/src/features/slack/server/oauth-handlers.ts
```diff
@@ -64,23 +56,62 @@ export async function handleCallback(
...
+              });
+              res.redirect("/slack/direct-setup");
+              return;
+            }
+            // Deliver the claim as an httpOnly cookie bound to this browser
+            // rather than a URL parameter — it is a bearer credential for a
+            // live bot token and must not land in history/Referer. The link
+            // procedures read it server-side. team_id/team_name stay in the URL
+            // (not secret) for display.
+            setPendingInstallClaimCookie(res, teamId, claimToken);
```
maxdeichmann, line 101: For my understanding:

we need the claim in the FE so that we know which of the unmatched installations should be matched when the user clicks on a project

web/src/features/slack/server/oauth-handlers.ts
```diff
@@ -64,23 +56,62 @@ export async function handleCallback(
...
+            return;
+          }
+
+          // Marketplace flow: no project chosen yet (the Direct Install URL
+          // redirects straight to OAuth). storeInstallation already saved a
+          // pending row; send the user to onboarding to link it to a project.
+          if (!projectId) {
+            const claimToken =
+              await SlackService.getInstance().issuePendingInstallationClaim(
+                teamId,
```
maxdeichmann, line 87: this is not considered secret, right? Thus if i can guess the teamId i could link a different orgs slack account to my langfuse project?

web/src/features/slack/server/oauth-handlers.ts
wochinge: almost:

> when the user clicks on a project

When the user is redirected to Langfuse, and gets presented with the view to connect their projects then we check if there is a pending installation matching team ID + claim (if there is no pending installation with their claim -> they don't even get the option to connect)

packages/shared/src/server/services/SlackService.ts
wochinge: done

packages/shared/prisma/schema.prisma
maxdeichmann: got it

web/src/features/slack/server/oauth-handlers.ts
maxdeichmann: got it. makes sense.

web/src/features/slack/server/oauth-handlers.ts
```diff
@@ -64,23 +56,62 @@ export async function handleCallback(
...
+            const onboardingUrl = `/slack/direct-setup?team_id=${encodeURIComponent(
+              teamId,
+            )}&team_name=${encodeURIComponent(teamName)}`;
+            res.redirect(onboardingUrl);
+            return;
+          }
+
+          // In-app "Connect" flow: the install is already linked to the project.
+          // Create an audit log; the session is still valid from when the user
+          // initiated the install.
```
maxdeichmann, line 111: this here is the path for
- user connects to slack out of languse AND
- user selected a project that he wants to connect the existing, non linked, project to

wochinge: yes (old path)

packages/shared/package.json
wochinge: nice - dropped it

web/src/features/slack/server/pendingInstallClaimCookie.ts
```diff
@@ -0,0 +1,87 @@
...
+  const value = encodeURIComponent(
+    JSON.stringify({ teamId, claim } satisfies ClaimCookiePayload),
+  );
+  const secure = env.NEXTAUTH_URL?.startsWith("https://") ?? false;
+  const cookie = [
+    `${PENDING_INSTALL_CLAIM_COOKIE}=${value}`,
+    "Path=/",
+    "HttpOnly",
+    "SameSite=Lax",
+    `Max-Age=${MAX_AGE_SECONDS}`,
```
maxdeichmann, line 38: IMO we have some cookie defaults already in the auth path. maybe there is a chance to unify this behavior. but likely over engineered

web/src/features/slack/server/router.ts
```diff
@@ -374,4 +381,145 @@ export const slackRouter = createTRPCRouter({
...
+   * Get the pending (unlinked) Marketplace installation for a Slack workspace.
+   * Used by the /slack/direct-setup onboarding page to show which workspace is being
+   * linked. Authenticated but not project-scoped: the install isn't owned by a
+   * project yet. Returns null if there is no pending install or it has expired.
+   */
+  // The claim is a bearer token for a live bot token: it is delivered as an
+  // httpOnly cookie (never a URL param or tRPC input) and read server-side
+  // here. Non-traced so the cookie-derived claim never reaches input telemetry.
+  getPendingInstallation: protectedProcedureWithoutTracing
+    .input(z.object({ teamId: z.string() }))
```
maxdeichmann, line 395: same comment for teamid

maxdeichmann: Yet, user also needs to have the claim here.

web/src/features/slack/server/router.ts
```diff
@@ -374,4 +381,145 @@ export const slackRouter = createTRPCRouter({
...
+  /**
+   * Get the pending (unlinked) Marketplace installation for a Slack workspace.
+   * Used by the /slack/direct-setup onboarding page to show which workspace is being
+   * linked. Authenticated but not project-scoped: the install isn't owned by a
+   * project yet. Returns null if there is no pending install or it has expired.
+   */
+  // The claim is a bearer token for a live bot token: it is delivered as an
+  // httpOnly cookie (never a URL param or tRPC input) and read server-side
+  // here. Non-traced so the cookie-derived claim never reaches input telemetry.
+  getPendingInstallation: protectedProcedureWithoutTracing
```
maxdeichmann, line 394: do we need rbac here?

web/src/features/slack/server/oauth-handlers.ts
wochinge: Not in this path and not later:

The oauth handler callback is called when Slack redirects back to us (once the user approved the Slack app permissions etc). In this case the `InstallProvider` verifies the callback validity (uses its own state cookie).

Then the second path would be the `getPendingInstallations` -> This is secured by our own cookie to ensure it's the same browser

web/src/features/slack/server/router.ts
wochinge: replied here: https://github.com/langfuse/langfuse/pull/14350#discussion_r3441561578

web/src/features/slack/server/router.ts
wochinge: We check that the user is auth-ed. Issue is that the pending installation doesn't belong to an org/project yet -> can't do RBAC in this case.
But all the other routes are secured against `automations:CUD`

web/src/features/slack/server/pendingInstallClaimCookie.ts
wochinge: re-used them! thanks for the pointer!

web/src/features/slack/server/oauth-handlers.ts
```diff
@@ -64,23 +56,62 @@ export async function handleCallback(
...
+            // rather than a URL parameter — it is a bearer credential for a
+            // live bot token and must not land in history/Referer. The link
+            // procedures read it server-side. team_id/team_name stay in the URL
+            // (not secret) for display.
+            setPendingInstallClaimCookie(res, teamId, claimToken);
+            const onboardingUrl = `/slack/direct-setup?team_id=${encodeURIComponent(
+              teamId,
+            )}&team_name=${encodeURIComponent(teamName)}`;
+            res.redirect(onboardingUrl);
+            return;
```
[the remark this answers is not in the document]
wochinge: addressing this in https://github.com/langfuse/langfuse/pull/14397

PR #13904, opened 2026-05-28
fix(web): show session comments in annotation queues

xstraven:
## What does this PR do?

Adds the missing session-level comments control to the session annotation queue item view.

Trace annotation queue items already expose comments through the shared trace header, but session annotation queue items did not show the session comments drawer. This PR wires the existing `CommentDrawerButton` into `SessionAnnotationProcessor` and fetches the session comment count with `comments.getCountByObjectId` for `objectType: "SESSION"`.

This reuses the existing comments drawer, permissions, count badge, comment creation, mentions, and deep-link behavior.

Fixes: https://github.com/langfuse/langfuse/issues/13903

Impacted package:
- `web`

Dependencies:
- None

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- [x] I have read the [contributing guide](https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md)
- [x] My code follows the style guidelines of this project
- [x] I have checked if my PR needs changes to the documentation
- [x] I have checked if my changes generate no new warnings
- [x] I have considered tests for this change

## Verification

- `pnpm --filter web run lint`
- Browser-checked a session annotation queue item:
  - confirmed the session header shows the comments button
  - opened the comments drawer
  - added a session comment
  - confirmed the comment appears and the count updates
- also checked a trace annotation queue item still shows its existing comments button

No new automated component test was added because this is a small wiring/parity fix using the existing comments surface. `TraceAnnotationProcessor` does not have dedicated coverage for comment visibility either, so I hope this is acceptable.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR wires the existing `CommentDrawerButton` into `SessionAnnotationProcessor` so that annotation queue items for sessions now expose the comments drawer — matching the behavior already present for trace annotation queue items.

- Adds a `sessionCommentCounts` tRPC query (`comments.getCountByObjectId` with `objectType: \"SESSION\"`) and passes its result via `getNumberFromMap` to `CommentDrawerButton`, following the identical pattern already used in `web/src/components/session/index.tsx`.
- Imports for `CommentDrawerButton` and `getNumberFromMap` are added at the top of the module, consistent with the project's import-placement rule.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

Safe to merge — the change is a minimal, focused wiring fix with no novel logic.

[...]

PR #14430, opened 2026-06-21
feat(annotation): link session ID from queue processor

xstraven:
## What does this PR do?

Makes the session ID in the session annotation processor link directly to the corresponding session page.

The session opens in a new tab so annotators can inspect its full details without losing their place in the queue. We find this especially cumbersome right now, since we sometimes rely on full session details like comments (also see https://github.com/langfuse/langfuse/pull/13904), or scores present that are not part of the annotation queue, to make the final judgement on assigning a score.

## Impacted package

- web

## Verification

- Added a session to an annotation queue locally
- Opened the session annotation processor
- Verified the session ID links to the correct session
- Verified the session page opens successfully in a new tab
- Verified there are no browser console errors
- Pre-commit checks: Prettier and all repository lint tasks

No automated tests were added because this is a small, directly verifiable navigation change.


xstraven:
I do have some more ideas for the Session Annotation Processor that circle around including more of the details from the Session page, but this would be already be helpful and does not require deeper ui/ux considerations for the annotation workflows.

nimarb:
@claude review

PR #14487, opened 2026-06-23
fix(ui): expand sidebar filter sections with active URL filters on load

nkabardin:
## What does this PR do?

When a table is opened via a URL with applied filters (bookmark / deep link), the sidebar filter sections rendered **collapsed** even though those filters were active. This seeds the initial expanded state so sections with an active URL-derived filter open on load; sections without one stay collapsed, and manual expand/collapse afterward is unaffected.

Implemented in the shared `useSidebarFilterState` hook (a one-time, mount-only, additive seed via a ref guard — never overrides later user toggles), so it applies to all sidebar-filter tables. Filter client tests pass (101/101); verified manually on the observations table.

LFE-10164

## Type of change

## Mandatory Tasks
- [x] Self-reviewed the code.

This PR fixes a UX bug where sidebar filter sections rendered collapsed on load even when the user arrived via a deep link or bookmark with pre-applied URL filters. The fix is isolated to `useSidebarFilterState`, applying to all sidebar-filter tables.

- Adds a mount-only `useEffect` guarded by a `useRef` that reads `explicitFilterState` at first render, finds which filter columns correspond to known facets, and calls `setExpandedString` with a functional update to additively expand those sections without clobbering existing expanded state.
- Uses `explicitFilterState` (not `effectiveFilterState`) so the implicit managed-environment default is excluded — the environment section only auto-expands when the user explicitly authored an environment filter.
</details>

Safe to merge; the change is additive, isolated to a single hook, and touches no data-mutation paths.

The seeding logic is well-structured — functional update, deduplication via Set, and a ref guard that prevents repeated seeding. Two non-blocking edge cases exist: the mount-time closure over explicitFilterState could read an empty value on Pages Router, and the instance-scoped ref resets on every remount causing re-expansion after a manual collapse. Neither path causes data loss or breakage.

web/src/features/filters/hooks/useSidebarFilterState.tsx — specifically the new useEffect block around lines 565-592.
</details>

```mermaid
flowchart TD
    A[Component Mounts] --> B{hasSeededExpansion .current?}
    B -- true --> C[Return early]
    B -- false --> D[Set ref = true]
    D --> E{explicitFilterState length === 0?}
    E -- yes --> F[Return early]
    E -- no --> G[Build facetColumns Set from config.facets]
    G --> H[Filter active columns to known facets]
    H --> I{activeFacetColumns length === 0?}
    I -- yes --> J[Return early]
    I -- no --> K[setExpandedString functional update]
    K --> L{Column already in expandedSet?}
    L -- yes --> M[Skip]
    L -- no --> N[Add to next and expandedSet]
    N --> L
    M --> O{Any new columns added?}
    N --> O
    O -- no --> P[Return same string reference]
    O -- yes --> Q[Return next.join comma]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[Component Mounts] --> B{hasSeededExpansion .current?}
    B -- true --> C[Return early]
    B -- false --> D[Set ref = true]
    D --> E{explicitFilterState length === 0?}
    E -- yes --> F[Return early]
    E -- no --> G[Build facetColumns Set from config.facets]
    G --> H[Filter active columns to known facets]
    H --> I{activeFacetColumns length === 0?}
    I -- yes --> J[Return early]
    I -- no --> K[setExpandedString functional update]
    K --> L{Column already in expandedSet?}
    L -- yes --> M[Skip]
    L -- no --> N[Add to next and expandedSet]
    N --> L
    M --> O{Any new columns added?}
    N --> O
    O -- no --> P[Return same string reference]
    O -- yes --> Q[Return next.join comma]
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/filters/hooks/useSidebarFilterState.tsx:565-592
**Stale-closure race on first render with Pages Router**

The effect captures `explicitFilterState` in a `[]`-dependency closure and fires synchronously after the very first render. In Next.js Pages Router, `useQueryParam` reads from `router.query`, which is **empty on the first render** for direct-navigation pages and is only populated on the second (client-side hydration) render. On that first render `explicitFilterState` will be `[]`, the early-return fires (`explicitFilterState.length === 0`), and then the ref guard is already `true` — so when the URL params arrive on the second render the effect never re-fires and the expansion seeding is permanently skipped for that page load.

For first-time deep-link visitors (no prior session storage state to fall back to) on Pages Router tables, the sidebar sections will remain collapsed even though filters are active. If this codebase has migrated fully to App Router the issue does not apply, since `useSearchParams()` is synchronous there — but that assumption is worth confirming.

### Issue 2 of 2
web/src/features/filters/hooks/useSidebarFilterState.tsx:564
**Re-expansion fires on every remount, overriding a manual collapse**

`hasSeededExpansionFromActiveFilters` lives on the component instance, so it resets to `false` whenever the component unmounts and remounts (e.g., navigating away and back to the same deep link). Each new mount re-seeds the expansion for any active-filter columns, regardless of whether the user had previously collapsed those sections. A user who deliberately folds a section, navigates away, and returns to the same filtered URL will have the section re-expanded.

Whether this is the intended behaviour depends on product requirements. If manual collapse should survive cross-navigation in the same session (session storage is already used for this), you'd want to guard on a session-storage flag rather than an instance ref — or simply accept the current behavior if "always show active sections on arrival" is the desired UX.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ui): expand sidebar filter sections ..."](https://github.com/langfuse/langfuse/commit/cbb6bce9d3625c5616374a11d09d1fa0334ee06f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39325097)</sub>

### DIFF
--- a/web/src/features/filters/hooks/useSidebarFilterState.tsx
+++ b/web/src/features/filters/hooks/useSidebarFilterState.tsx
@@ -1,5 +1,5 @@
 import type React from "react";
-import { useCallback, useMemo, useEffect, useState } from "react";
+import { useCallback, useMemo, useEffect, useRef, useState } from "react";
 import { StringParam, useQueryParam } from "use-query-params";
 import {
   type FilterState,
@@ -548,6 +548,48 @@ export function useSidebarFilterState(
       : stateLocationType === "memory"
         ? memoryFilterState
         : urlFilterState;
+
+  // LFE-10164: When arriving via a URL/deep link that already carries applied
+  // filters, expand the sidebar sections that have an active filter. Sidebar
+  // sections are collapsed by default; without this, a bookmarked/shared link
+  // would render its active facets collapsed. This runs once on initial load
+  // (guarded by a ref) so it only seeds the default state and never overrides
+  // subsequent manual expand/collapse. Sections without an active filter keep
+  // their default state.
+  //
+  // We key off `explicitFilterState` (the URL/memory/peek-authored filters),
+  // which already excludes the implicit hidden-environment default — so the
+  // managed-environment section is only auto-expanded when the user actually
+  // authored an environment filter.
+  const hasSeededExpansionFromActiveFilters = useRef(false);
+  useEffect(() => {
+    if (hasSeededExpansionFromActiveFilters.current) return;
+    hasSeededExpansionFromActiveFilters.current = true;
+
+    if (explicitFilterState.length === 0) return;
+
+    const facetColumns = new Set(config.facets.map((facet) => facet.column));
+    const activeFacetColumns = explicitFilterState
+      .map((filter) => filter.column)
+      .filter((column) => facetColumns.has(column));
+    if (activeFacetColumns.length === 0) return;
+
+    setExpandedString((current) => {
+      const expanded = current.split(",").filter(Boolean);
+      const expandedSet = new Set(expanded);
+      const next = [...expanded];
+      for (const column of activeFacetColumns) {
+        if (!expandedSet.has(column)) {
+          expandedSet.add(column);
+          next.push(column);
+        }
+      }
+      return next.length === expanded.length ? current : next.join(",");
+    });
+    // Intentionally runs only on initial mount; the ref guard prevents re-seeding
+    // after later filter changes so manual collapse is respected.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   const managedEnvironmentPolicyConfig = useMemo(
     () => buildManagedEnvironmentPolicyConfig(implicitDefaultConfig),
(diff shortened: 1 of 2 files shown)


nkabardin:
Refactored to remove the mount `useEffect` (commit `360348a4c`). Sidebar section expansion is now seeded via **during-render reconciliation** against `explicitFilterState`, with the "seeded" marker persisted in session storage alongside the expanded state. This addresses the review feedback:

- **Pages-Router late-arriving params:** no one-shot guard — when the URL filter arrives on a later render it's detected as newly-active and the section expands (deep-link now opens on load; added a test reproducing the race).
- **Remount re-seed vs. manual collapse:** the seeded marker lives in session storage (not a per-mount ref), so a section the user deliberately collapsed is not re-expanded on remount; a genuinely new active filter still auto-expands.

105/105 filter client tests (+4 new), typecheck/lint clean, Playwright-verified on a deep-link URL. Resolving the related threads.

nkabardin:
@claude review

PR #14491, opened 2026-06-23
refactor(web): move all Radix/Vaul overlays onto the layer system (LFE-10471)

nkabardin:
## TL;DR

Overlays — dialogs, dropdowns, tooltips, toasts, the in-app assistant — used to portal to `<body>` and fight over stacking with hand-tuned **z-index numbers**. This finishes moving them onto a **layer system** where they stack by **structure (DOM order)**, not numbers. No overlay carries a z-index anymore, and a lint rule stops anyone re-introducing one.

Fixes `LFE-10471`. Fixes the toasts-behind-peek bug (`LFE-10408`) as a side effect.

## Why we needed this

Every overlay rendered at `<body>` and competed for "who's on top" with an escalating magic z-index — `z-50`, then `z-51`, `z-60`, `z-[9999]`. Whoever picked the bigger number won, until the next overlay reset the race. That produced real, recurring bugs: **toasts hidden behind the trace peek**, the **search-bar error tooltip clipped**, the **nav dropdown colliding with the assistant window**. Each fix was another number; none were stable.

## The approach

We started the layer system in **[#14342](https://github.com/langfuse/langfuse/pull/14342)**: overlay containers declared once as `<body>` siblings that paint in DOM order, with the whole app isolated inside `#__next` so nothing in-app can escape to compete. That PR added the first layer and fixed the clipped tooltip.

**This PR finishes the job** — it deeply integrates **Radix & Vaul** so that when their primitives portal (which they'd otherwise do to `<body>` with a z-index), they portal **into a layer** instead. Ordering becomes the layer's job, not a number's.

## The layers (low → high)

| Layer | Holds | Why it sits here |
| --- | --- | --- |
| **agent** | the in-app assistant window | persistent/draggable; above the page, below every transient overlay |
| **modal** | Dialog, AlertDialog, Sheet (incl. the table peek), Drawer | blocking surfaces, above the page + assistant |
| **popover** | Popover, DropdownMenu, Select, HoverCard | **above modal**, so a Select/Popover opened *inside* a Dialog still renders on top |
| **tooltip** | Tooltip + bespoke anchored tooltips | hints stay above their trigger, even inside a modal |
| **toast** | Sonner toasts | last, so they always sit above everything — by order alone |

## Result

- **Zero z-index on any overlay.** The escape-hatch numbers (`z-51` / `z-60` / `z-[9999]`) are gone.
- Remaining z-index is only **legit in-app chrome** (sticky headers, top banners, the bulk-action bar) that lives inside `#__next` and can't compete with overlays.
- A new lint rule (`@repo/no-overlay-zindex`) makes re-introducing an overlay z-index a **build failure**, so the bug class can't return.
- **Storybook** now documents the layer system as a design-system reference (*Design System → Overlay Layers*).

<details>
<summary><b>Engineering detail</b> (mechanism, a11y, verification)</summary>

### Mechanism
- `layer.tsx`: `LAYER_ORDER = ["agent","modal","popover","tooltip","toast"]` (later = on top). `_document.tsx` maps this list to per-layer `<body>`-sibling containers; `globals.css` isolates `#__next` (`isolation: isolate`) so nothing inside can paint over them.
- `useLayerContainer(name)` hands a layer's DOM element to a Radix/Vaul `*.Portal`'s `container` prop. **SSR-safe**: `container` is `null` until mounted → falls back to `document.body`, identical to today on first paint.
- Each `ui/*` overlay wrapper routes its portal into its layer and drops its `z-50`/`z-9999`.

### Pointer events
Layer containers are `pointer-events: none` so the empty space around a non-modal overlay (e.g. the table peek) stays click-through. Every portaled overlay opts back in via one global rule — `[data-overlay-root] > [data-layer] > * { pointer-events: auto }` — so all overlays are interactive by construction, modal or not.

### `popover` above `modal`
The real regression risk is a Select/Popover/Dropdown opened **inside a Dialog** (common in forms). Ordering `popover` above `modal` preserves the old "newest-opened wins" behaviour: the in-dialog Select renders above the dialog.

### a11y preserved
`container` only re-parents the portal DOM. Focus trap, Escape, aria wiring and `data-state` animations live on the Radix Root/Content, not the portal parent. The peek's `modal={false}` click-through is unchanged. Intra-content z-index (a dialog's sticky header) is left intact — only overlay-level z-index was removed.

### Guardrail
`@repo/no-overlay-zindex` (in `@repo/eslint-plugin`, scoped in `web/eslint.config.mjs`): `wrapper` mode bans high/arbitrary z-index on `ui/*` overlay wrappers; `overlay-content` mode bans it app-wide on overlay content. Shipped with full unit tests.

### Verification
- **Static:** typecheck + eslint clean; `build-storybook` compiles (incl. the new docs page); no `z-50`/`z-9999` left in overlay wrappers; lint rule confirmed firing on an injected `z-9999`.
- **Playwright (headless, demo@ on the seeded project):** DropdownMenu / standalone Select / Dialog / Sheet(peek) each asserted in the correct layer with computed z-index `auto`; toast fired over an open peek paints above it; Select-inside-Dialog renders above the dialog; layer container DOM order asserted. The peek **close-button click-through** regression was reproduced via `document.elementFromPoint`, fixed (the pointer-events rule above), and re-verified: 9/9 overlay types capture clicks + the real peek closes.

### Needs a human / cloud-review sweep
Touches shared wrappers across many call sites; representative high-risk cases verified, but a reviewer should eyeball Drawer drag-to-dismiss, nested DropdownMenu `SubContent`, HoverCard, Tooltip under the search bar's `tooltip` layer, AlertDialog, and the assistant-window items (its entitlement is cloud-only, so enable locally to test).

</details>


nkabardin:
Addressed both nits (`a5c93c121`, `0c6d4935f`):
- **eslint rule**: now strips variants *before* normalizing, so `md:!z-50` / `hover:!z-9999` / `dark:!z-[9999]` are caught (tests added; plugin suite 265 green).
- **DropdownMenuPortal**: now routes through the `popover` layer like the other `*Portal` wrappers (preserves an explicit `container` override).

Resolving.

nkabardin:
@claude review

nkabardin:
@claude review

PR #14506, opened 2026-06-24
feat(shared): add parquet flag to blob export tuning (LFE-10463)

niklassemmler:
## What does this PR do?

Phase 1 of the Parquet blob-export format (LFE-10463): the shared-types groundwork.

Adds a `parquet` boolean to the per-project blob-export tuning contract:

- `parquet: z.boolean().optional()` on `BlobExportTuningSchema` (the write-validation schema).
- `parquet: boolean` on `ResolvedBlobExportTuning`, defaulting to `false`.
- `resolveBlobExportTuning()` resolves it like the other booleans, and enforces precedence: when both `parquet` and `rawPassthrough` are enabled it keeps `parquet`, forces `rawPassthrough` to `false`, and emits a warning — they are mutually exclusive ClickHouse execution paths.

Schema + resolver only. No export path consumes the flag yet; Phase 2 wires the `exec()`-based Parquet streaming util, the per-table repository functions, and the job-handler dispatch.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- [x] Added unit tests covering the new `parquet` flag, the wrong-typed/absent cases, and the parquet-over-rawPassthrough precedence.
- [x] Existing and new unit tests pass locally (`vitest` — 35 passing); `pnpm run lint` passes repo-wide.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR adds a `parquet` boolean flag to the `BlobExportTuningSchema` and resolver, enabling a ClickHouse-native `FORMAT Parquet` export path as an alternative to the existing JSONEachRow raw-passthrough. When both `parquet` and `rawPassthrough` are set, `parquet` wins and `rawPassthrough` is forced to false with a warning.

- Adds `parquet: z.boolean().optional()` to the Zod write schema and `parquet: boolean` to the `ResolvedBlobExportTuning` type, with appropriate fallback in the resolver.
- Implements conflict resolution so that `parquet: true` overrides `rawPassthrough: true`, matching the comment that these are mutually exclusive execution paths.
- Ships comprehensive tests covering the new flag in isolation, the conflict case, wrong-type fall-back, and the interaction with `rawPassthrough` only (no parquet).
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

[...]

PR #14509, opened 2026-06-24
feat(worker): stream ClickHouse-native Parquet blob exports (LFE-10463)

niklassemmler:
## What does this PR do?

Phase 2 of the Parquet blob-export format ([LFE-10463](https://linear.app/langfuse/issue/LFE-10463); follows #14506, the shared-types phase). When a project's `exportTuning.parquet` flag is set, every export table (traces, observations, scores, events) is written via ClickHouse's native `FORMAT Parquet`, streamed straight to upload — no worker gzip, no per-row JSON parse/enrich/serialize. Columnar encoding and compression run on ClickHouse query threads, eliminating the single-threaded gzip bottleneck that dominates large exports.

**How it works**

- `queryClickhouseExecRaw()` (new, in `repositories/clickhouse.ts`) runs the query via `clickhouseClient().exec()`, which returns the raw HTTP body as a `Readable`. It appends `FORMAT <value>` to the SQL (exec has no `format` param) and wraps the body in an exception-tag-scanning `Transform`.
- `exec()` skips the client's row-parsing machinery, so it does no mid-stream exception detection. The new `ClickhouseExecExceptionTagTransform` restores that for binary formats: it scans for ClickHouse's end-of-stream `\r\n__exception__\r\n<tag>` marker (CH ≥ 25.11) and errors the stream on a failed query, so the pipeline aborts **before** any S3 commit instead of uploading a truncated object. Same version caveat as raw passthrough.
- Per-table Parquet export fns reuse the existing export SQL; `traces`/`scores` had their inline SQL extracted into shared query builders so the standard and Parquet paths can't drift.
- Handler: a `parquetEligible` dispatch guard (overrides `fileType`/`compressed`, takes precedence over `rawPassthrough`), `.parquet` extension + `application/vnd.apache.parquet` content type, metrics tagged `format=parquet`, and row count `0` (a binary stream has no row boundaries).
- Adds `@clickhouse/client-common` as a direct dependency for the exception-tag helpers (already present transitively at the same version).

**Behavior is gated entirely behind `exportTuning.parquet`** (set via DB, no UI/API surface yet — see [LFE-10463](https://linear.app/langfuse/issue/LFE-10463)). Existing integrations are untouched.

Fixes # (no GitHub issue — tracked in Linear)

## Type of change

## Mandatory Tasks

## Checklist

- Tests added: exception-tag `Transform` unit tests (synthetic trailers, chunk-boundary splits, passthrough when detection is disabled), live-ClickHouse integration (valid Parquet `PAR1` magic; a failing query surfaces an error rather than truncating), and parquet dispatch wiring (`.parquet` filenames per table, `compressed` ignored, precedence over `rawPassthrough`).
- Verified locally: `pnpm turbo lint`/`typecheck` for shared + worker, all touched test suites green against live ClickHouse 25.12.

This PR adds a ClickHouse-native `FORMAT Parquet` export path for the blob-storage integration (LFE-10463), streaming binary Parquet bytes directly from ClickHouse to object storage — bypassing the per-row JS parse/enrich/serialize pipeline and worker-side gzip. A per-project `exportTuning.parquet` flag activates it; when set it overrides `fileType`, `compressed`, and takes precedence over `rawPassthrough`.

- **New `queryClickhouseExecRaw`** in `clickhouse.ts` calls `clickhouseClient().exec()`, appends `FORMAT <format>` to the SQL, and wraps the raw HTTP body in a new `ClickhouseExecExceptionTagTransform` that scans for the ClickHouse ≥ 25.11 end-of-stream exception trailer, erroring the stream on mid-stream failures before any S3 commit.
- **`ClickhouseExecExceptionTagTransform`** is a bounded-memory `Transform` (16-byte carry between chunks) with solid unit test coverage across marker split, passthrough-disabled, and partial-trailer cases.
- **Four new `getXxxForBlobStorageExportParquet` functions** in `traces.ts`, `scores.ts`, `observations.ts`, and `events.ts` share the existing query-builder SQL; the existing `getXxxForBlobStorageExport` stream functions are refactored onto the same shared builders with no behaviour change.
</details>

The core binary streaming pipeline is well-designed and tested, but the exception-tag Transform's _flush path has a gap where a null return from extractErrorAtTheEndOfChunk would silently complete the stream instead of erroring it, risking a corrupt Parquet object being committed to storage.

The new ClickhouseExecExceptionTagTransform._flush calls extractErrorAtTheEndOfChunk and passes the result directly to callback() without a null guard. If the library returns null for a malformed trailer, callback(null) ends the stream cleanly, bypassing the upload-abort guarantee that is the primary safety property of this feature. The unit tests do not exercise the null-return edge case. The span lifecycle on stream destroy() without error is also unguarded.

clickhouseExecExceptionTag.ts — the _flush null-guard gap; clickhouse.ts — span close-event handling in queryClickhouseExecRaw.
</details>

```mermaid
sequenceDiagram
    participant Worker as BlobStorage Worker
    participant CH as ClickHouse
    participant T as ExceptionTagTransform
    participant TC as TimedByteCounter
    participant S3 as Object Storage

    Worker->>CH: exec("SELECT ... FORMAT Parquet")
    CH-->>Worker: 200 OK + x-clickhouse-exception-tag header
    Note over Worker: queryClickhouseExecRaw wraps body in ExceptionTagTransform

    loop Stream chunks
        CH-->>T: binary Parquet chunk
        T->>T: scan for marker
        alt No marker found
            T-->>TC: pass chunk through
            TC-->>S3: upload chunk (multipart)
        else Marker detected
            T->>T: accumulate error trailer
            T--xTC: error(parsed ClickHouse exception)
            Note over S3: upload aborted before commit
        end
    end

    alt Stream ends cleanly
        T->>T: _flush - emit withheld carry bytes
        TC-->>S3: finalize multipart upload
    else Stream ends with trailer
        T->>T: _flush - extractErrorAtTheEndOfChunk
        T--xS3: stream error propagated via pipeline
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Worker as BlobStorage Worker
    participant CH as ClickHouse
    participant T as ExceptionTagTransform
    participant TC as TimedByteCounter
    participant S3 as Object Storage

    Worker->>CH: exec("SELECT ... FORMAT Parquet")
    CH-->>Worker: 200 OK + x-clickhouse-exception-tag header
    Note over Worker: queryClickhouseExecRaw wraps body in ExceptionTagTransform

    loop Stream chunks
        CH-->>T: binary Parquet chunk
        T->>T: scan for marker
        alt No marker found
            T-->>TC: pass chunk through
            TC-->>S3: upload chunk (multipart)
        else Marker detected
            T->>T: accumulate error trailer
            T--xTC: error(parsed ClickHouse exception)
            Note over S3: upload aborted before commit
        end
    end

    alt Stream ends cleanly
        T->>T: _flush - emit withheld carry bytes
        TC-->>S3: finalize multipart upload
    else Stream ends with trailer
        T->>T: _flush - extractErrorAtTheEndOfChunk
        T--xS3: stream error propagated via pipeline
    end
```

</a>
</details>

`````markdown

---

### Issue 1 of 3
packages/shared/src/server/repositories/clickhouseExecExceptionTag.ts:109-117
**Potential silent stream completion when `extractErrorAtTheEndOfChunk` returns null**

`extractErrorAtTheEndOfChunk` is documented (and can behave) as returning `Error | null` — null when the trailer cannot be parsed. When null is returned, `this.wrapError(null)` yields null (default wrapError is identity), and `callback(null)` ends the stream cleanly instead of erroring it. The net effect is that a ClickHouse failure after the marker is detected would silently complete the stream, allowing a truncated or corrupt Parquet file to be committed to object storage. A null-guard and a fallback `new Error("ClickHouse mid-stream failure (unparseable trailer)")` would close this gap.

### Issue 2 of 3
packages/shared/src/server/repositories/clickhouse.ts:135-141
**Span leaks when consumer destroys the stream without an error**

The span is closed only on `end` and `error` events. When `pipeline()` in the caller aborts an upload by calling `stream.destroy()` with no argument — or when the BullMQ job times out and the Node.js `pipeline` tears down the chain — `wrapped` emits `close` but neither `end` nor `error`. The span then stays open until the process exits. Adding `wrapped.once("close", () => span.end())` alongside the existing `error` listener would cover all teardown paths.

### Issue 3 of 3
packages/shared/src/server/repositories/clickhouse.ts:105-110
**`queryId` is always `undefined` when the pre-200 error fires**

`queryId` is declared before the `try`, assigned at line 112 (`queryId = res.query_id`), but the `.catch()` callback that calls `enrichWithQueryId(error as Error, queryId)` runs synchronously before the assignment — `queryId` is still `undefined` at that point. The resulting error message will never include the query ID for exec-level (pre-200) failures. The pattern is copied from `queryClickhouseStream`, so this is pre-existing, but now that there is a new public function it may be worth correcting for debuggability.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(worker): stream ClickHouse-native P..."](https://github.com/langfuse/langfuse/commit/96012ffd4adb6af19e158f7a68c7d02315b8cee3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39518635)</sub>

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -104,6 +104,7 @@
     "@aws-sdk/s3-request-presigner": "^3.1013.0",
     "@azure/storage-blob": "^12.26.0",
     "@clickhouse/client": "^1.18.5",
+    "@clickhouse/client-common": "^1.18.5",
     "@google-cloud/storage": "^7.19.0",
     "@langchain/anthropic": "^1.3.26",
     "@langchain/aws": "^1.3.4",
--- a/packages/shared/src/server/repositories/clickhouse.ts
+++ b/packages/shared/src/server/repositories/clickhouse.ts
@@ -1,9 +1,12 @@
+import { type Readable } from "stream";
 import { env } from "../../env";
 import {
   clickhouseClient,
   convertDateToClickhouseDateTime,
   PreferredClickhouseService,
 } from "../clickhouse/client";
+import { EXCEPTION_TAG_HEADER_NAME } from "@clickhouse/client-common";
+import { ClickhouseExecExceptionTagTransform } from "./clickhouseExecExceptionTag";
 import { logger } from "../logger";
 import { getTracer, instrumentAsync } from "../instrumentation";
 import { randomUUID } from "crypto";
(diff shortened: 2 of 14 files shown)


niklassemmler:
* **#14509** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14509?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
@claude review

PR #14514, opened 2026-06-24
feat(blobstorage): circuit-breaker disables export after repeated failures (LFE-10279)

niklassemmler:
## What does this PR do?

Adds a **circuit-breaker** to the blob storage export integration so a repeatedly-failing chunk stops retrying forever.

A failing export chunk was requeued indefinitely: on error `lastSyncAt` never advances, so the scheduler re-enqueues the same `(projectId, lastSyncAt)` chunk every cycle. BullMQ also retries each job (`attempts: 5`, exponential backoff) within a cycle, so the same failure recurs at both layers. A single wedged or misconfigured project could emit thousands of errors and keep the failure monitor firing with no path to resolution.

Now the integration tracks consecutive failed **sync runs**. A sync run is a fully retry-exhausted BullMQ job: BullMQ already absorbs transient errors within a run, so only the final retry counts toward the breaker — a blip it recovers from never advances the counter or emails owners. Once consecutive failed runs reach `LANGFUSE_BLOB_STORAGE_MAX_CONSECUTIVE_FAILURES` (default 3, kept below the per-job retry count) the integration is auto-disabled (`enabled = false`) and project owners are emailed that it was paused. The counter resets on any successful run, and on manual re-enable. Once disabled, the scheduler no longer selects the project (it filters `enabled: true`), so both the retry loop and the monitor noise stop until an owner re-enables it.

The pause email reuses the existing `BlobStorageExportFailedEmail` path with a new "paused" variant and bypasses the failure-notification cooldown, so owners always learn the integration was disabled. The auto-disable runs as an atomic `enabled true→false` claim that also seeds the notification timestamp, so concurrent workers can't send duplicate or contradictory pause/failed emails for the same trip.

Fixes LFE-10279.

Scope note: the ticket also proposed splitting the failure metric by error class (customer-config vs. infra). That half was deliberately deferred to a follow-up; this PR is the circuit-breaker only.

## Type of change

## Mandatory Tasks

## Checklist

- [x] Added tests that prove the feature works (disable-after-N-failures, reset-on-success, and final-retry-only gating in `blobStorageIntegrationProcessing.test.ts`).
- [x] New and existing unit tests pass locally (targeted worker tests + repo lint + typecheck).
- [x] DB migration included (`consecutive_failures` column).

This PR adds a circuit-breaker to the blob-storage export pipeline: after `LANGFUSE_BLOB_STORAGE_MAX_CONSECUTIVE_FAILURES` consecutive failures the integration is auto-disabled, project admins are emailed a "paused" notification, and the scheduler's `enabled: true` filter stops the retry loop. Re-enabling via the settings UI grants a fresh failure budget by resetting the counter.

- A new `consecutive_failures` column is added to `blob_storage_integrations` via migration; it increments atomically in Prisma on each error and resets to 0 on success.
- The disable claim uses an atomic `updateMany` (`enabled: true` filter) to ensure only one worker among concurrent processors sends the "paused" email; the existing cooldown timestamp is seeded in the same write to prevent a racing "failed" email from slipping through.
- Re-enabling a disabled integration in `upsertBlobStorageIntegration` resets `consecutiveFailures`, `lastError`, and `lastErrorAt` — but does not clear `lastFailureNotificationSentAt`, so the first post-re-enable failure can be silently swallowed by the notification cooldown.
</details>

Safe to merge with the notification-cooldown fix applied — the circuit breaker logic and atomic disable claim are correct, but post-re-enable failures can be silently swallowed until the cooldown expires.

The reactivation path in upsertBlobStorageIntegration resets consecutiveFailures, lastError, and lastErrorAt but leaves lastFailureNotificationSentAt untouched. Because the circuit-breaker updateMany seeds that timestamp when it disables the integration, a user who re-enables and hits another failure within the cooldown window gets no email notification. The test for the re-enable path also never sets lastFailureNotificationSentAt in setup, so it cannot catch this gap. Everything else looks correct.

web/src/features/blobstorage-integration/service.ts — the reactivated spread needs lastFailureNotificationSentAt: null; the companion test should be updated to seed and assert that field.
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(blobstorage): drop redundant boolean..."](https://github.com/langfuse/langfuse/commit/1e1d4d48b834d65e3c6d2e1de5f292bccc65c76a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39537117)</sub>

### DIFF
--- a/packages/shared/prisma/migrations/20260624000000_add_blob_storage_consecutive_failures/migration.sql
+++ b/packages/shared/prisma/migrations/20260624000000_add_blob_storage_consecutive_failures/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "blob_storage_integrations" ADD COLUMN "consecutive_failures" INTEGER NOT NULL DEFAULT 0;
--- a/packages/shared/prisma/schema.prisma
+++ b/packages/shared/prisma/schema.prisma
@@ -1244,6 +1244,11 @@ model BlobStorageIntegration {
   lastError   String?   @map("last_error")
   lastErrorAt DateTime? @map("last_error_at")
 
+  // Circuit breaker: consecutive failed sync runs. Reset to 0 on success; once
+  // it reaches the configured threshold the integration is auto-disabled
+  // (enabled = false) so a wedged chunk stops retrying forever (LFE-10279).
+  consecutiveFailures Int @default(0) @map("consecutive_failures")
+
   // Notification tracking
   lastFailureNotificationSentAt DateTime? @map("last_failure_notification_sent_at")
 
--- a/packages/shared/src/server/services/email/blobStorageExportFailed/BlobStorageExportFailedEmailTemplate.tsx
+++ b/packages/shared/src/server/services/email/blobStorageExportFailed/BlobStorageExportFailedEmailTemplate.tsx
@@ -17,17 +17,23 @@ import {
 type BlobStorageExportFailedEmailTemplateProps = {
   projectName: string;
   settingsUrl: string;
+  // When true the export was auto-disabled after repeated failures and needs
+  // to be re-enabled by the customer once the underlying issue is resolved.
+  paused?: boolean;
 };
 
 export const BlobStorageExportFailedEmailTemplate = ({
   projectName,
   settingsUrl,
+  paused = false,
 }: BlobStorageExportFailedEmailTemplateProps) => {
   return (
     <Html>
       <Head />
       <Preview>
-        Blob storage export failed for project &quot;{projectName}&quot;
+        {paused
+          ? `Blob storage export paused for project "${projectName}"`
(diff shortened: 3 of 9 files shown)


niklassemmler:
* **#14514** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
For now we won't merge this. We don't have a proper communication path to ensure blob exporters disabled due to retries are not forgotten.

PR #14283, opened 2026-06-16
feat(observations-v2): add shadow query validation for subquery-IN rewrite

niklassemmler:
## What does this PR do?

Adds a **shadow-query validation mode** for the observations v2 subquery-IN rewrite. When `LANGFUSE_OBSERVATIONS_V2_SHADOW_QUERY=true`, every V2 observations query that takes the `cte-join` path also fires the `subquery-rewrite` path fire-and-forget, compares result IDs/ordering, and emits Datadog metrics.

The shadow path **never affects the API response** — it runs asynchronously after the primary query returns.

### Metrics emitted (`langfuse.observations_v2.shadow_query.*`)

| Metric | Meaning |
|---|---|
| `.match` | Both paths returned identical IDs in identical order |
| `.row_count_mismatch` | Different number of rows returned |
| `.order_mismatch` | Same rows, different sort order |
| `.error` | Shadow path threw an exception |
| `.duration_ms` | Absolute shadow path latency |
| `.duration_delta_ms` | Shadow minus primary (negative = shadow is faster) |

### Rollout plan

1. Enable `LANGFUSE_OBSERVATIONS_V2_SHADOW_QUERY=true` in one region
2. Monitor metrics for 48h — expect 0 mismatches
3. If clean: enable remaining regions for 24h
4. Cutover: flip `LANGFUSE_OBSERVATIONS_V2_SUBQUERY_REWRITE=true`, disable shadow
5. Remove shadow code after 1 week bake

### Isolation for removal

All shadow logic is in a single `runSubqueryRewriteShadowQuery()` function with boundary comments. Removal = delete the function, its one call site, and the env var.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring

## Checklist

- [x] Tested with `pnpm run lint`, `pnpm run typecheck`, and unit tests
- [x] Shadow path is fire-and-forget — never blocks or changes the API response
- [x] New env var defaults to `false` — no behavioral change without opt-in

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary><h3>Greptile Summary</h3></summary>

This PR introduces a fire-and-forget shadow validation layer for the observations V2 subquery-IN rewrite. When `LANGFUSE_OBSERVATIONS_V2_SHADOW_QUERY=true`, every query that takes the `cte-join` path also runs the `subquery-rewrite` path asynchronously and emits Datadog comparison metrics — the shadow never blocks or modifies the API response.

[...]

PR #14515, opened 2026-06-24
feat(filters): nested AND/OR filter expressions for events — backend + search bar (LFE-10421)

nkabardin:
## TL;DR

`(level:ERROR OR level:WARNING) AND -environment:prod AND name:*checkout*` now
works **end-to-end** — typed in the search bar, carried through the URL, run on
ClickHouse. The backend filter contract understands a nested boolean tree
(cross-field `OR`, bracket grouping) instead of only a flat implicit-`AND`
`FilterCondition[]`, and the GA search bar produces and consumes it.

Verified live (browser → tRPC → real ClickHouse): typing `level:ERROR OR type:SPAN`
writes an OR tree to the URL and returns the correct union (`55,275` =
`max(1316, 54592) ≤ … ≤ sum`); a tree URL re-derives the bar text; a depth-9 tree
is rejected `400`. Flat filters are byte-for-byte unchanged.

## Why

The grammar search bar (GA on the v4 events tables) already parses a full boolean
AST, but the backend contract was flat (`FilterState`, implicit AND) and the bar's
adapter *rejected* cross-field OR / groups at lowering. So those queries — and
bracket grouping — were commit-blocking diagnostics. v2 makes the whole stack
understand a **tree**.

## What

Decision 1C (hybrid) + Decision 2 Option A (extend the DTOs into a tree, recursive
parenthesizing emitter over the existing `Filter` classes), per the RFC. Ports the
already-prototyped backend (#12814) onto current `main`, adds the bounds it lacked,
and wires the FE through the **shared** filter stack.

**Backend contract**
- Recursive `filterExpression` / `filterGroup` (`z.lazy`); `filterInput =
  union(flatArray, expression)`; `normalizeFilterExpressionInput` (flat → top-level
  AND). Leaf `Filter` classes (randomized typed params, value never interpolated)
  are **unchanged** — only the combiner (`applyCompiledFilterNode` / `FilterTree`)
  became recursive.
- **Tenant isolation stays structural**: projectId/env/time wrap the user tree as
  an outer AND; `findMandatory` (AND-only leaves) gates optimizations (trace-id
  hash prune, start-time bound) so they never fire on a leaf under an OR.
- **Bounds**: depth ≤ 8, ≤ 64 leaf conditions, rejected at the tRPC boundary.
- events list/count + router accept `filterInput`; comment filters resolve
  recursively; positionInTrace must be AND-reachable.

**Frontend (full integration)**
- `astToFilterInput` lowers the AST to a flat array (pure AND — sidebar-owned) or a
  nested tree (cross-field OR / group). Per-leaf lowering is **shared** with the
  flat `astToFilterState` (no parity drift); `filterInputToQueryText` renders a tree
  back via the precedence-aware serializer.
- The URL `filter` param codec carries a tree as JSON (leading `{`); flat stays the
  legacy delimited format, so **every existing URL + saved view is unchanged**.
- `useSidebarFilterState` gains `explicitFilterExpression` / `effectiveFilterExpression`
  / `isAdvancedFilter` / `setFilterExpression`; the flat fields stay **byte-identical**,
  so the ~11 other tables that reuse the hook are untouched.
- Facet sidebar shows a read-only "advanced filter — edit in the search bar" state
  when a tree is active (the bar is the editor; "Clear all" returns to facets).

## Result

- Unit + integration tests: contract/emitter, depth/condition bounds, tenant-guard
  (project filter stays outside a user OR; trace-hash prune only on AND-reachable
  traceId), flat↔tree lowering parity, reverse round-trip, ClickHouse nested AND/OR
  list+count. Existing 179 search-bar client tests updated + green.
- Browser-verified the full round-trip + correctness + bounds + sidebar state.
- Addressed two review rounds: bounds off-by-one, singleton-OR mandatory-leaf
  sync, positionInTrace message, comment-filter cleanup, plus a hidden-env leak
  (env filter under an OR no longer suppresses the managed-env default) and the
  bulk-action gating below.
- Rebased onto current `main` (post #14521 table-views URL-as-source-of-truth);
  reconciled the new search-bar AI apply path with the tree-aware commit.

## ⚠️ Scoped follow-ups (flat path has no regression)

1. **Saved views of a tree.** Storing/restoring a nested filter needs widening the
   **9-table-shared** `useTableViewManager` (its `setFilters`/`currentFilterState`
   are flat-typed; widening breaks the other 8 under strict function variance) +
   the `TableViewPreset` schema + create/update tRPC. Saving a view *while a tree is
   active* currently captures the flat state only — a clean next PR.
2. **Bulk actions on a tree** (export / run-eval / add-to-dataset / annotation
   queue) consume a flat `FilterState`. When a tree filter is active these
   "select-all-matching" actions are **disabled** (with a reason) rather than
   silently widening to every row; explicit row selection stays enabled. Widening
   their backends to accept a tree is a follow-up.
3. **ClickHouse execution caps** (§3.5 `max_execution_time` / `max_rows_to_read`
   for OR queries) intentionally **not set blindly**: a per-query value could relax
   a stricter server-side limit, so it needs a prod-informed value (Nikita/Mark).
   The depth (≤8) / condition (≤64) bounds are the deterministic structural blast
   protection that ships here.

NOT-of-a-group stays the prototype's De-Morgan-to-leaves (the backend has no general
NOT); an explicit NOT node is a noted follow-up.

<details>
<summary>Files & mechanism</summary>

Backend: `interfaces/filters.ts`, `clickhouse-filter.ts` (`FilterTree`,
`CompiledFilterCollection`), `factory.ts` (`createFilterTreeFromFilter{Expression,Input}`),
`event-query-builder.ts` (`findMandatory`), `repositories/events.ts`,
`services/commentFilterService.ts`, events `{types,eventsRouter,eventsService}.ts`.

Frontend: `search-bar/lib/{adapter,filter-state-to-query,validate,commit}.ts`,
`search-bar/hooks/useEventsSearchBar.ts`, `filters/lib/filter-query-encoding.ts`
(`encode/decodeFilterInput`), `filters/hooks/useSidebarFilterState.tsx`,
`events/components/EventsTable.tsx`, `events/hooks/useEventsTableData.ts`,
`table/data-table-controls.tsx` (read-only state). README updated.

Seed/related: #12814 (the Phase-1 backend ported), #12845 (lucene FE prototype),
#14283 (v4 subquery-IN rewrite). Linear: LFE-10421.

</details>

### DIFF
--- a/.agents/skills/seed-test-data/SKILL.md
+++ b/.agents/skills/seed-test-data/SKILL.md
@@ -29,19 +29,18 @@ this.
 
 ## Need → command
 
-| I need...                                                                               | Command                                                                                                                                                                                                                              |
-| --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| A very complex observation tree (v3)                                                    | `pnpm run seed -- trace-tree --observations 5000 --depth 12 --breadth 500`                                                                                                                                                           |
-| The same tree readable in the v4 events UI                                              | add `--v4` (writes `events_full`; `events_core` fills via MV)                                                                                                                                                                        |
-| Async parents whose subtree outlives their own span (subtree wall-clock duration badge) | add `--async-parents` to `trace-tree` (root + hub end immediately while children keep running)                                                                                                                                       |
-| A super tough session (v3 legacy session view)                                          | `pnpm run seed -- long-session --traces 300 --observations-per-trace 8`                                                                                                                                                              |
-| Many traces for list/filter performance                                                 | `pnpm run seed -- many-traces --count 100000 --days 14`                                                                                                                                                                              |
-| Scores with spaces in the name (filter/grammar testing)                                 | `pnpm run seed -- scored-traces --traces 24 --v4`                                                                                                                                                                                    |
-| Varied human-annotation queues (annotate UI / keyboard testing)                         | `pnpm run seed -- annotation-queue --core-items 12` (creates a "core types" queue covering every score-field render path + an "edge cases" queue with archived/stale/partial scores and observation/session/deleted/completed items) |
-| Huge/malformed/unicode payloads                                                         | `pnpm run seed -- trace-tree --payload-bytes 1000000 --payload-style malformed` (styles: json, text, malformed, unicode, bignum)                                                                                                     |
-| Big integers beyond 2^53-1 (number-precision testing)                                   | `pnpm run seed -- trace-tree --observations 1 --payload-style bignum`                                                                                                                                                                |
-| See all scenarios and flags                                                             | `pnpm run seed -- list --json`                                                                                                                                                                                                       |
-| Predict without writing                                                                 | add `--dry-run`                                                                                                                                                                                                                      |
+| I need... | Command |
+| --- | --- |
+| A very complex observation tree (v3) | `pnpm run seed -- trace-tree --observations 5000 --depth 12 --breadth 500` |
+| The same tree readable in the v4 events UI | add `--v4` (writes `events_full`; `events_core` fills via MV) |
+| A super tough session (v3 legacy session view) | `pnpm run seed -- long-session --traces 300 --observations-per-trace 8` |
+| Many traces for list/filter performance | `pnpm run seed -- many-traces --count 100000 --days 14` |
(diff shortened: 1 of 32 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #14526, opened 2026-06-24
feat(editor): magic paste for JSON editors (LFE-10488)

nkabardin:
## TL;DR
Pasting raw text into any JSON editor in the app used to break the JSON. This adds a client-side "magic paste" that keeps it valid automatically, with a one-tap **Paste raw** escape hatch — now a default for every editable JSON editor.

## Why
A non-technical user pastes a long blob (a chapter, a log, a transcript) into a JSON field. If the caret is inside a string value, the unescaped quotes / backslashes / newlines / control chars make the JSON invalid; if the field is empty, a bare blob isn't JSON at all. The data is private, so the usual workaround — an external JSON-escape tool — isn't acceptable. So pasting just… breaks, with no obvious fix. This started from the dataset-item input/output/metadata fields but applies to every JSON editor.

## What
A new CodeMirror extension (`createJsonMagicPasteExtension`) is wired into `CodeMirrorEditor` as a **built-in for editable `mode="json"` editors**, so it covers dataset items, corrected-output, dataset/experiment metadata, playground tool/schema definitions, model config, prompt config, etc. It applies the **deterministic, lossless** transform for the context:

- **Caret inside a JSON string** → escape the pasted fragment (`"` `\` newlines, tabs, control chars).
- **Blank field** whose paste isn't already valid JSON → wrap it as a quoted JSON string.
- Everything else (already-valid JSON, plain text needing no escaping, structural mid-document pastes, multi-cursor, file pastes, mid-escape carets) → untouched; files still go to the media handler.

After a transform, a small passive control appears at the paste site — **"Escaped to keep JSON valid · Paste raw"** (reachable by mouse and by ⇧⌘V / Ctrl+Shift+V) — auto-dismissing on the next edit; the whole paste is one transaction, so ⌘/Ctrl+Z reverts it too.

The transform is **conservative** — it only acts in the two cases above and otherwise defers to normal paste — so no existing JSON editor regresses (it's never worse than today's paste).

## Result
JSON stays valid on paste across every editable JSON editor, 100% client-side (no external calls). Verified in the browser (dataset-item editors + the newly-covered dataset-metadata editor) and covered by unit tests.

<details>
<summary>Implementation, scope & verification</summary>

**Files**
- `web/src/components/editor/jsonMagicPaste.ts` — the extension + pure helpers (`planMagicPaste`, `escapeForJsonStringBody`, `isInsideJsonString`, `isValidJson`, `selectionStaysInJsonString`) + the "Paste raw" tooltip (a `showTooltip` state field + a `Mod-Shift-v` keymap).
- `web/src/components/editor/jsonMagicPaste.clienttest.ts` — 29 unit tests (escape/wrap/context/cross-string/multi-cursor/mid-escape).
- `web/src/components/editor/CodeMirrorEditor.tsx` — one line: enable it for editable `mode="json"`.
- `web/src/components/editor/fileDropPaste.ts` — export `extractTransferFiles` so magic paste and the media handler share one file-detection definition.

**Mechanism**
- Paste interception via `EditorView.domEventHandlers({ paste })` (same pattern as `fileDropPaste.ts`); files defer to the media handler; IME composition defers to native paste.
- Escaping is `JSON.stringify(text).slice(1, -1)`; "inside a string" / "selection stays in one string" / "mid-escape" use a small JSON string-state scanner (handles unterminated strings, escaped quotes, and dangling backslashes — cases the Lezer tree misses).
- Multi-cursor pastes defer to native per-cursor paste.

**Affordance / a11y**
- Rendered below the paste (can't be clipped); follows the caret on long pastes; survives cursor moves, clears on the next edit or on blur; "Paste raw" is keyboard-reachable (⇧⌘V / Ctrl+Shift+V) with `aria-keyshortcuts`; live region announces the status only; tooltip paints its own surface from app tokens for audited contrast.

**Scope**
- Promoted from a per-form opt-in to a `CodeMirrorEditor` default for editable JSON editors; the dataset-item forms no longer need bespoke wiring (the media drop/paste handler stays). Conservative design means no editor regresses.

**Verification**
- `pnpm --filter web run test-client` (29/29), `lint`, `typecheck` all green.
- Browser (Playwright): empty-field wrap, inside-string escape, "Paste raw" revert (mouse + keyboard), tooltip visibility incl. long pastes, blur dismissal, the edit-dialog path, and the newly-covered **New dataset** metadata editor — all produce valid JSON.
</details>

### DIFF
--- a/web/src/components/editor/CodeMirrorEditor.tsx
+++ b/web/src/components/editor/CodeMirrorEditor.tsx
@@ -32,6 +32,7 @@ import {
 import { lightTheme } from "@/src/components/editor/light-theme";
 import { darkTheme } from "@/src/components/editor/dark-theme";
 import { autoScrollOnSelectionDrag } from "@/src/components/editor/autoScrollOnSelectionDrag";
+import { createJsonMagicPasteExtension } from "@/src/components/editor/jsonMagicPaste";
 
 // Custom language mode for prompts that highlights mustache variables and prompt dependency tags
 const promptLanguage = StreamLanguage.define({
@@ -509,6 +510,10 @@ export function CodeMirrorEditor({
         : []),
       ...(mode === "json" ? [json()] : []),
       ...(mode === "json" && linterEnabled ? [linter(jsonParseLinter())] : []),
+      // Magic paste: when editing JSON, escape pasted text inside a string (or
+      // wrap a blob pasted into a blank field) so the JSON stays valid, with a
+      // "Paste raw" escape hatch. Conservative — defers to normal paste otherwise.
+      ...(editable && mode === "json" ? [createJsonMagicPasteExtension()] : []),
       ...(mode === "prompt" ? [promptSupport, promptLinter] : []),
       ...(lineWrapping ? [EditorView.lineWrapping] : []),
       ...(additionalExtensions ?? []),
--- a/web/src/components/editor/fileDropPaste.ts
+++ b/web/src/components/editor/fileDropPaste.ts
@@ -3,9 +3,13 @@ import { type Extension } from "@codemirror/state";
 
 /**
  * Collects files from a drop/paste transfer, falling back to `items` for
- * sources (e.g. clipboard images) that don't populate `files`.
+ * sources (e.g. clipboard images) that don't populate `files`. Exported so other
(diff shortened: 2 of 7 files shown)


nkabardin:
@claude review

PR #14528, opened 2026-06-24
feat(web): reflect Parquet export-tuning override in blob-storage settings (LFE-10463)

niklassemmler:
## What does this PR do?

Surfaces the ClickHouse-native **Parquet** blob-export format (added in #14509, LFE-10463) in the Blob Storage Integration settings page.

Parquet is an **internal, DB-set** `exportTuning.parquet` knob with no UI write path. When set, the worker already overrides the file format — it always produces a `.parquet` artifact and ignores `fileType`/`compressed`. This PR makes the settings form **reflect that override read-only** so the UI doesn't misrepresent what will be exported. It is a deliberately lightweight, removable interim surface — if Parquet later becomes a first-class `fileType` option, this reflection goes away.

When the override is active, the form:
- locks the **File Type** selector to "Parquet" with an explanatory description,
- hides the **Gzip Compression** toggle (Parquet compresses internally),
- drops the enrichment-only price columns (`input_price`, `output_price`, `total_price`) from the **Model** field group — they're absent from Parquet exports because the ClickHouse-native path runs no JS enrichment.

Implementation:
- **shared** (`analytics-integrations/index.ts`): derive a `parquetDescription` per export field group (only the model group differs), alongside the existing `legacyDescription`.
- **web** (`blobstorage-integration/types.ts`): `parquetEnabledFromTuning()` reads the persisted `exportTuning` column; the settings page keys the three UI changes off it.

No Postgres enum/schema migration, no public API change, no Fern change, no worker change.

Fixes # (LFE-10463)

## Type of change

## Mandatory Tasks

## Checklist

- Added client tests: `parquetEnabledFromTuning` truthiness + the Model-group parquet description (drops prices; other groups unchanged). 19 client tests pass.
- `pnpm turbo run typecheck lint --filter=web --filter=@langfuse/shared` pass.
- Browser review of the read-only reflection (requires an integration with `exportTuning.parquet=true` in the DB) is still recommended before merge — not run in this environment.

### DIFF
--- a/packages/shared/src/features/analytics-integrations/index.ts
+++ b/packages/shared/src/features/analytics-integrations/index.ts
@@ -117,10 +117,25 @@ const legacyDescriptionForGroup = (
     : "Not included in the legacy observations export";
 };
 
+// The Parquet export path (ClickHouse-native FORMAT Parquet) runs no JS
+// enrichment, so the model-price columns (MODEL_ENRICHMENT_FIELDS) are absent.
+// Drop them from the model group description; all other groups are unchanged.
+const parquetDescriptionForGroup = (
+  group: ObservationFieldGroupFull,
+): string => {
+  const base = EXPORT_FIELD_GROUP_LABELS[group].description;
+  if (group !== "model") return base;
+  return base
+    .split(", ")
+    .filter((field) => !MODEL_ENRICHMENT_FIELDS.includes(field))
+    .join(", ");
+};
+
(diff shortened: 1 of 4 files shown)


niklassemmler:
* **#14528** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #14530, opened 2026-06-24
feat(peek): unify the trace detail surface into one adaptive, resizable drawer

nkabardin:
## TL;DR
The table **peek** is now a **resizable, width-adaptive** drawer that **Expand**s in place (swipe-down on mobile, click-outside to close) — and the scope that grew underneath it: the peek and the standalone trace page now render **one shared trace-detail surface** instead of two diverging copies. As the peek narrows, its header adapts to **its own width** (not the screen), folding controls into a tidy menu rather than crowding. Closes **LFE-10118**, folds in **LFE-5877**.

## Why (the story)
There were **three** ways to look at the same trace, and they'd drifted apart:
- the **peek** (`?peek=`) — a fixed-width side panel, minimal chrome, no star/publish/delete;
- an **in-place fullscreen** toggle — same panel, wider, but un-shareable (no URL);
- **"open in current tab"** — which navigated to `/traces/<id>`, a *different component* with the delete button, a breadcrumb, and **no way back**.

So "expand" and "open the page" were a jarring UI swap, and the peek and the page were two implementations of the same thing. That fork was an early mistake that grew out of control — and it's exactly what blocks improving the peek: you can't make it faster or richer while every change has to be made twice.

## What
**One peek, one surface, that adapts to its size.**
- **One Expand control** — widens the peek to max width (**viewport − sidebar**, so the nav stays visible), reflected in the **URL** (`peekView=expanded`) so it's shareable and reload-safe. Dragging the handle to the edge does the same thing. Replaces the ephemeral fullscreen *and* both old open-in-tab buttons; the standalone-page view is still one click away via **Open in new tab**.
- **Responsive shell** — docked, resizable panel on desktop; a `vaul` **swipe-down drawer** on mobile. **Click-outside closes**; clicking another row still **switches** the peeked item in place (power-user behavior preserved).
- **Width-adaptive header** — measures the *peek's own width* and, as it narrows, reduces in least-painful order: the trace actions fold into a labeled **"…" menu**, the type badge drops to an **icon** (never truncates to "Tr…"), the prev/next nav goes **compact** (arrows only, K/J in the tooltip), then open-in-tab folds. Correct on first paint, not just after a nudge.
- **Unified detail surface** — the peek and the standalone `/traces/<id>` page now share one data fetch, one body, one title, and one action set (star/publish/delete). They're thin shells around the same core, so the peek finally has parity with the page and future work lands **once**.
- **Modernized state** per `frontend-large-feature-architecture` (the exemplar we're setting): a per-mount Zustand store for the widget width, the URL for the expanded view, localStorage for the persisted width preference.

## On breaking the "small PR" rule
Our big-frontend-feature skill says *don't bundle a large rewrite*. We deliberately broke that here, with eyes open: **the duplication was the bug**, and there's no incremental path that doesn't first remove the fork. We ran a multi-perspective review (a risk lens, an architecture lens, and a QA-edge-case lens) plus several follow-up reviewer rounds, and fixed every P0/P1 surfaced.

## Result
Verified live across both surfaces: resize (stops exactly at the **sidebar edge** — no overshoot/snap-back, and no jump when grabbing the handle while expanded), Expand ↔ URL, reload/back, mobile swipe, click-outside vs row-switch, nested split resize, the **width-adaptive header** (fold → menu, badge → icon, nav → compact across the drag range, and compact on refresh), shared star/publish/delete with delete-closes-peek, and unified fast tooltips. **31** focused unit tests; lint / typecheck / prettier green.

Also fixes a **pre-existing** v4 bug the unification surfaced: on the events path an access error rendered as *"Trace not found"* — now correctly *"you don't have access to this trace"*.

<details>
<summary>Engineering detail — state, adaptation, fixes, a11y, what's deferred</summary>

**State altitudes**
- `store/peekPanelStore.ts` — per-mount vanilla Zustand store (lazy `useState`), owns only the widget width + transient drag state, mutated through named `actions`; primitive selectors.
- `actions/resizePeekPanel.ts` — the drag workflow (window listeners → store actions; commits a width or flips the expanded flag on pointer-up). The expand threshold and width cap are **sidebar-aligned**, so a drag stops at `viewport − sidebar` instead of overshooting onto the sidebar and snapping back; a pointer leaving the window clamps cleanly; the drag seeds its draft from the current state so merely *pressing* the handle while expanded doesn't jump.
- `usePeekPanelState.ts` — integration boundary: derives width (widget vs expanded, capped at the sidebar edge via CSS `min()`), measures the sidebar offset live, wires drag/keyboard.
- **Expanded** lives in the URL (`peekView=expanded`, `router.replace` so toggling doesn't spam history), cleared on close by `usePeekNavigation`.

**Width-adaptive header** (`PeekHeader.tsx` + `peekHeaderOverflow.ts`)
- A pure, unit-tested planner keeps the title above a minimum width and applies reductions in order — fold actions → badge to icon → compact nav → fold open-in-tab — based on the measured peek width.
- The header width equals the peek width, so it doesn't change when controls settle; a second observer on the control cluster re-triggers the plan, so a refreshed narrow peek is correct on the **first paint**.
- The "…" overflow is a real labeled menu (full-width, whole-row clickable) reusing the existing controls — Share keeps its URL popover, Delete its confirm. All header controls share one fast Radix tooltip (no native `title`); Share suppresses its tooltip while its popover is open.

**Unified surface:** `useTraceDetailData` (beta-aware fetch), `TraceDetailBody`, `traceDetailTitle`, `TraceDetailActions` — all shared by `TracePage` and the peek wrappers. `isNotFound` now requires no-data **and** no-error, so a 500/UNAUTHORIZED on the v4 path is no longer mislabeled as a missing trace.

**Other notable fixes**
- Dismissal is target-based and never closes for a target inside the peek (`data-peek-content`) — fixes a crash where the inner trace split's pointer-capture was misread as an outside click and tore the panel group down mid-drag.
- Delete-from-peek invalidates broadly (the peek is hosted over many different lists) and closes in place; the page still navigates.
- Nested-table state (filters/sort/pagination) survives the desktop↔mobile breakpoint cross; the `actions` subtree is keyed so folding moves DOM without remounting (an in-progress delete confirmation survives a resize).
- star / publish / delete share one ghost icon style; resize handle straddles the left edge and exposes `aria-valuemin/max/now`; soft left-cast shadow lifts the panel.

**Deferred follow-ups (now cheap, since the core is shared):** collapse the `<Trace context>` branching; a `panelStyle`-only subscriber so a 60fps drag doesn't re-render the shell; retire the now-dead `expandConfig`/`expandPeek` plumbing; experiment-item peek actions.
</details>

### DIFF
--- a/web/src/components/publish-object-switch.tsx
+++ b/web/src/components/publish-object-switch.tsx
@@ -5,11 +5,17 @@ import {
   PopoverContent,
   PopoverTrigger,
 } from "@/src/components/ui/popover";
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from "@/src/components/ui/tooltip";
 import { useV4Beta } from "@/src/features/events/hooks/useV4Beta";
 import { usePostHogClientCapture } from "@/src/features/posthog-analytics/usePostHogClientCapture";
 import { useHasProjectAccess } from "@/src/features/rbac/utils/checkProjectAccess";
 import { api } from "@/src/utils/api";
 import { copyTextToClipboard } from "@/src/utils/clipboard";
+import { cn } from "@/src/utils/tailwind";
 import { trpcErrorToast } from "@/src/utils/trpcErrorToast";
 import { type RouterInput } from "@/src/utils/types";
 import { CheckIcon, Globe, Link, Share2 } from "lucide-react";
@@ -23,6 +29,8 @@ export const PublishTraceSwitch = (props: {
   size?: "icon" | "icon-xs";
   /** When set, render as a full-width labeled menu item instead of an icon. */
   label?: string;
+  /** Hover tooltip for the icon button (suppressed while the popover is open). */
+  tooltip?: string;
 }) => {
   const { isBetaEnabled } = useV4Beta();
   const capture = usePostHogClientCapture();
@@ -106,6 +114,7 @@ export const PublishTraceSwitch = (props: {
       isPublic={props.isPublic}
       size={props.size}
       label={props.label}
+      tooltip={props.tooltip}
       onChange={(val) => {
         capture("trace_detail:publish_button_click");
         return mut.mutateAsync({
@@ -168,6 +177,7 @@ const Base = (props: {
   disabled?: boolean;
   size?: "icon" | "icon-xs";
   label?: string;
+  tooltip?: string;
 }) => {
   const [isCopied, setIsCopied] = useState(false);
   const [isOpen, setIsOpen] = useState(false);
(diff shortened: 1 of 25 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

PR #14537, opened 2026-06-25
feat(trace): show subtree wall-clock duration on observations (LFE-10475)

nkabardin:
## TL;DR

A parent observation that kicks off **async / fire-and-forget** work shows a misleadingly short duration — its own span ends before its children finish, so the children outlive the parent and its duration badge understates the real elapsed time. This PR surfaces the **subtree wall-clock duration** as a distinct `∑` badge next to the own-span badge, shown whenever the two differ. It also fixes a tree tooltip that was outright lying.

## Why

A customer reported that a parent span (an async workflow dispatcher) shows e.g. **0.04s**, even though the work it triggers runs for seconds. The badge today is `end − start` of the observation's **own** OTEL span — for async children that outlive the parent, that number is misleading. The ask (consensus on the thread): *show both the own-span duration and the subtree's wall-clock duration, when they differ.*

To make it worse, the tree already rendered a tooltip **"Aggregated duration of all child observations"** on nodes with children — but the value shown was the node's **own** span, not the subtree. The UI claimed subtree-aggregated and showed span-own.

## What

- **New `∑` subtree badge** next to the own-span badge — the wall-clock duration of the node and all its descendants (`first descendant start → last descendant end`), the same way the trace-level latency is calculated.
- **Shown on any visible difference** from the own span — i.e. whenever the subtree would render a *different* value at the duration formatter's precision (~0.01s). The subtree always contains the own span (so it can only be ≥ own); synchronous parents that cover their children show a single number, async ones show both. No arbitrary magic thresholds. Users who don't want durations at all use the existing "show duration" toggle.
- **Both surfaces:** the tree (`SpanContent`) and the timeline (`TimelineBar`), kept consistent (identical own-span basis).
- **Accurate, one-line tooltips:** own-span vs subtree wall-clock are clearly labeled; the mislabeled "aggregated" tooltip is fixed.
- **No new setting** — it appears automatically under the existing "show duration" toggle.

## Result

In an async trace, the dispatcher node now reads `0.04s  ∑ 4.09s` — own span *and* the real elapsed time of the work it triggered — in both the tree and the timeline. Verified live in the browser across both views.

<details>
<summary>Mechanism, units, threshold & verification</summary>

**Aggregation.** `subtreeWallClockDurationMs` is computed bottom-up during tree construction in `tree-building.ts`, mirroring the existing `totalCost` aggregation: each node tracks the earliest start / latest end across itself and all descendants (a missing `endTime` contributes only its start). Purely client-side — no backend/API change. The field is left undefined on the synthetic TRACE wrapper node (which already renders trace-level latency), so no badge appears there.

**Units.** Kept in **milliseconds** end-to-end (observations store latency in ms, traces in s — keeping ms avoids the mismatch), divided to seconds only at `formatIntervalSeconds`.

**When it shows.** `getSubtreeDurationOverflowMs(ownMs, subtreeMs)` returns the subtree duration only when `formatIntervalSeconds(subtree) !== formatIntervalSeconds(own)` — any difference visible at the rendered precision, and never two identical-looking numbers. A missing own-span duration is treated as 0, so a node with no recorded end still surfaces its subtree duration. The tree and timeline use an identical own-span basis (`endTime − startTime`, falling back to `latency`), so they always agree.

**Tests.** `tree-building.clienttest.ts` covers the aggregation (async child outliving parent, deep nesting, multi-sibling spans, synchronous-contained subtree, missing-endTime fallback, zero-duration leaf, events-based roots) and the visibility rule (sub-precision hidden, 10ms shown, minute-scale resolution, missing-own fallback, leaf no-op).

**Reproducing locally.** `trace-tree` gained an opt-in `--async-parents` flag (root + hub end immediately while their subtree keeps running):

\`\`\`
pnpm run seed -- trace-tree --observations 60 --depth 6 --breadth 12 --async-parents --v4
\`\`\`

**Verification.** Lint, typecheck (web + shared), and client tests pass; reviewed both the tree and timeline in a real browser against the seeded async trace.
</details>


nkabardin:
@claude review

PR #14538, opened 2026-06-25
fix(dashboard): show per-series summary values in time-series chart legend (LFE-10498)

nkabardin:
## TL;DR

Home-dashboard time-series legends now show a per-series **summary** next to each series name — for the charts where that number is unambiguous: additive metrics (event counts, token/cost totals). Empty-vs-zero handling is correct: a real `0` shows, a series with no data shows no number.

## Why

On the home screen the multi-series chart legends listed each series by name only, with no at-a-glance summary, and the reported behavior conflated "no data" with a real `0`. The trap to avoid: `0` is a **real** value and must be shown, whereas a series with no data point must not be given an invented number.

A first attempt also summed non-additive series (latency percentiles, average scores), which review correctly flagged as misleading — a sum of percentiles or averages is meaningless. An `"average"` mode was then tried, but review caught a second, subtler problem: the shared dashboard pipeline (`extractTimeSeriesData` / `fillMissingValuesAndTransform`) pads missing `(bucket × series)` pairs with real numeric `0`s, and `Number.isFinite(0)` is `true`, so those padding zeros deflate any average computed at the chart layer. Computing a correct average would require null-padding non-additive series in that shared pipeline — out of scope for this paper-cut.

## What

- `getDimensionSummaries(data)` (chart-library utils) sums each series' **finite** numeric metric values and returns the total, or `null` when the series has **no** finite data point. A series whose values sum to `0` keeps its `0` (real value); non-finite values and the histogram tuple shape are ignored.
- New `legendSummary` prop (`"sum" | "none"`, default `"none"`) on `Chart` / `LineChartTimeSeries` / `AreaChartTimeSeries`. The summary is shown **only** when `legendSummary="sum"` **and** the per-series value is non-`null` (so `0` shows, missing does not).
- Only additive home-dashboard charts opt in: **Observations by time** and **Model Usage**. Non-additive charts (Model latencies, Scores) keep name-only legends — i.e. their pre-existing behavior — rather than a fabricated or deflated number.

## Result

- "Observations by time" legend reads e.g. `DEFAULT 400`, reconciling with the card's `400` headline; "Model Usage" reads `gpt-3.5-turbo $60,000.00`.
- Real `0` summaries render; series without data add no misleading value.
- Model latencies and Scores legends are unchanged from before this PR (name only).
- Verified live across all four chart types.

<details>
<summary>Implementation detail & verification</summary>

**Mechanism.** The legend already iterated unique dimensions; the value is computed once per render (memoized) from the same `DataPoint[]` the chart consumes, and only when `legendSummary !== "none"`. The render predicate is `summary !== null` (not falsy) so a legitimate `0` renders — the crux of the fix. Values reuse the chart's compact `metricFormatter`/unit handling, so a cost series shows `$…`, a count shows `400`, etc.

**Why summary is gated per call-site.** By the time data reaches the chart the aggregation has already happened (each point is already a count, a percentile, or an average), so the chart can't tell additive from non-additive on its own. An explicit per-call-site `legendSummary` declares it; the default is `"none"` so a new caller never silently gets a misleading summary. Restricting to additive metrics also sidesteps the pipeline's `0`-padding, which would otherwise deflate a chart-layer average.

**Tests.** `getDimensionSummaries` unit tests in `utils.clienttest.ts` cover: summing per series; the zero-is-real invariant (a `0`-sum series kept as `0`, not dropped); a series with only non-finite points reported as `null`; finite-plus-NaN mixing; and ignoring dimensionless rows and the histogram tuple shape. `pnpm vitest --project client` green (16/16). Lint, prettier, and typecheck pass.

**Live check.** Logged into the seeded home dashboard via the browser and inspected the legend DOM: additive sums reconcile with card headlines (`DEFAULT 400` = headline `400`; `$60,000.00`), real `0`s render, and the non-additive Model latencies / Scores legends show name only (no number).

</details>


nkabardin:
@claude review

PR #14572, opened 2026-06-25
feat(trace): rebuild Timeline as a two-pane waterfall (LFE-10527)

nkabardin:
## TL;DR
The **Timeline** view's parent→child connectors were dead — and the root cause was architectural, not a CSS tweak: hierarchy (depth) and time (gantt) were crammed onto one horizontal axis. This PR rebuilds the Timeline as a **two-pane waterfall** (a fixed name-gutter + a time-aligned chart) and layers on the UX the new structure unlocks. Connectors now connect at every depth, and the timeline can finally be made wide.

## Why
The connectors were assembled at render time by CSS absolute divs at hardcoded offsets, hoping two unrelated coordinate systems would line up. They couldn't: the depth-indented tree elbow and the time-offset bar drifted apart by the node's start time — a gap that grew without bound (measured 21px → 223px down the deep-nested trace). The Tree view was unaffected; only the Timeline conflated the two axes.

## What
**Two coordinate systems → two panes** (`TimelineRow` split into `TimelineGutterRow` + `TimelineBar`):
- **Gutter (depth):** indented span-name tree with connectors. Sticky — never scrolls horizontally. Connectors are computed rail lines (parent center→bottom + child top→center meet at the row boundary), so they're continuous at every depth.
- **Chart (time):** the gantt bars under the time scale; owns the only horizontal scrollbar.
- Vertical scroll is mirrored between panes; the scale is transform-synced; classic-scrollbar widths are reserved on the gutter (bottom) and scale (right) so all three share one scrollable extent.

**UX the structure unlocks:**
- Whole-row **hover/selection** (caption + chart together), visible on the light theme; selection adds an accent bar + ring.
- **Resizable** name gutter; **collapsible detail panel** — drag the splitter past its threshold to hide it (timeline goes full-width); an edge tab and selecting a node both reopen it.
- **Tree | Timeline** segmented icon switch that collapses to icons-only at narrow *panel* width; the minor toolbar tools fold into a **"⋯" overflow menu** when space is tight.
- **Storybook** stories for the presentational pieces.

> ⚠️ Touches the shared `TraceLayoutDesktop` splitter (detail-panel min size + collapse), so the **Tree view** inherits the resize/collapse behavior too.

## Result
Connectors link parent→child at every depth and sit on the bar center; only the chart scrolls horizontally (gutter pinned, vertical synced, scale aligned under scroll); the detail panel collapses so the timeline can be dragged full-width. Works in the full-page Timeline and the table peek.

<details>
<summary>Verification + notes</summary>

Measured in-browser on deep-nested + child-before-root seeded traces (`view=timeline`):
- **Connectors** continuous at every depth (elbow → badge ≈ 0; guide segments contiguous across rows).
- **Scroll:** gutter horizontal range = 0 (pinned); chart scrolls; vertical mirrored; 0s tick == 0s bar at scroll 0 and mid-scroll.
- **Collapse:** detail panel resizes below the old 50% floor down to a 360px floor, then snaps shut; reopens via the edge tab and on node selection (incl. re-selecting the same node).
- **Responsive header:** switch labels and the "⋯" overflow toggle on the panel's own width (container query), pixel-aligned with the toolbar.
- **Storybook:** `TimelineScale`, `TimelineBar`, `TimelineGutterRow` (connector states + a stacked tree showcase from the real flattening) — 16 stories render green.
- lint + typecheck pass; pre-commit prettier + `turbo run lint` across all packages pass.

Review nits from automated review addressed (pointercancel teardown on the resizer; streaming in-flight dashed cue; equal pane scrollbar extents on both axes; badge no-shrink at the gutter floor; reopen on re-select). Scrollbar-extent fixes are 0-impact on macOS overlay scrollbars and only matter on classic-scrollbar platforms.

The journey here was several earlier commits (repair connectors → split waterfall → two-pane → polish); they squash-merge into the above.
</details>

Related: #14539 (LFE-10496, Timeline gantt origin — separate, merged).


nkabardin:
@claude review

PR #14579, opened 2026-06-26
feat(api): add experiments public API

wochinge:
## Summary

- Adds public `GET /api/public/experiments` and `GET /api/public/experiment-items` endpoints backed by events-table queries, with cursor pagination, optional field groups, filters, and score hydration.
- Updates the experiment list contract to return one representative event row per experiment, remove `itemCount`, and omit the unreleased single-experiment detail endpoint.
- Keeps Zod schemas, Fern source, generated OpenAPI, and server/repository tests aligned with the final public API surface.

## Linear

- [LFE-10502](https://linear.app/langfuse/issue/LFE-10502/implement-public-api)

## Major Decisions

- Experiment summaries page over event rows ordered by `project_id`, minute-bucketed `start_time`, `experiment_id`, `start_time`, and `span_id`, then use `LIMIT 1 BY project_id, experiment_id` to return one row per experiment.
- Cursor pages exclude experiments seen in the one-day lookback before the cursor timestamp to avoid duplicate experiments when an experiment has multiple events around a page boundary.
- Experiment and experiment-item cursors both stay at version `1` because this public API has not been released yet; the two endpoints still parse different cursor payload shapes internally.

## Review Focus

- ClickHouse pagination semantics for experiment summaries, especially the ordering tuple, `LIMIT 1 BY`, and one-day lookback exclusion.
- Public API contract alignment across route handlers, Zod schemas, Fern source, and generated OpenAPI.
- Score hydration for experiment summaries and experiment items, including timestamp bounds and per-subject `scoreLimit` behavior.


web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,193 @@
...
+  name?: string[];
+  datasetId?: string[];
+  advancedFilters?: EventsTableFilterState;
+  cursor?: ExperimentSummariesCursor;
+  includeMetadata: boolean;
+  includeScores: boolean;
+  scoreLimit?: number;
+};
+
+export async function queryExperimentSummariesForPublicApi(
```
wochinge: @sumerman This is only the first endpoint but would appreciate some quick initial feedback
- the cursor pagination with aggregates is a bit awkward as we can in theory only paginate correctly _if_ we filter after the grouping (which would be super bad for performance). So I'm now applying the cursor on an event level - risking that an experiment is repeated on later pages. Do you have any ideas that would improve the situation?
example:

```
experiment A - item 1 - 8:00am
experiment B - item 1 - 8:05am
-- cursor
experiment A - item 2 - 9:00am
```

- could you please have a look at the structure of the repository? I'm a bit confused what parts go into the repository, what parts go into the `event-query-builder`. For now I only moved things to the query builder If I expect them to be re-used. 

- what do you think of the scores query? I checked and it increases read data only slightly (for queries covering the last year it was about 3% more given that experiment level scores are very rare. The alternative would be 2 separate queries.

wochinge: will actually split the join into 2 queries. For list experiment-items it's much more efficient and for list experiments it doesn't matter so much. Definitely easier to reason about.

sumerman: 1. I think you can apply it twice before and after aggregation. Technically you only need project id and time before the aggregation. 

2. Structure looks reasonable at the moment. My rule of thumb is to try and keep as much SQL as possible out of repository. If it is very reusable it goes into a query builder, otherwise into a query-fragments. 

3. will get back to you on that

wochinge: > I think you can apply it twice before and after aggregation. Technically you only need project id and time before the aggregation.

I'll try that out on Monday and check the performance implication

> f it is very reusable it goes into a query builder, otherwise into a query-fragments.

Okay, will have a look for candidates on Monday

wochinge:
@claude REVIEW

web/src/features/experiments/server/public/repository.ts
wochinge: > I think you can apply it twice before and after aggregation. Technically you only need project id and time before the aggregation.

That doesn't work. If you already filter based on the cursor time in the `where`, then the groupBy doesn't see any prior events to that timestamp anymore.
Would only work if we prefilter in the `where` based on the users's `fromTimestamp` and apply the cursor timestamp in the groupby. However, that is really nasty for deep pages, isn't it?

> f it is very reusable it goes into a query builder, otherwise into a query-fragments.

In my opinion fine as is. There is a bunch of sql in the score repository but that's the same case for the other things in the score repo.

web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,463 @@
...
+            lastStartTime: params.cursor.lastStartTime,
+            lastTraceId: params.cursor.lastTraceId,
+            lastId: params.cursor.lastId,
+            lastExperimentId: params.cursor.lastExperimentId,
+          }
+        : undefined,
+    )
+    .orderBy(experimentItemOrderBy("e"))
+    .limitBy("e.span_id", "e.project_id")
+    .limit(params.limit);
```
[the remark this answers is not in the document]
wochinge: Same as observation v2.

fern/apis/server/definition/experiments.yml
```diff
@@ -0,0 +1,186 @@
...
+          limit:
+            type: optional<integer>
+            docs: Number of experiments to return per page. Maximum 100, default 50.
+          scoreLimit:
+            type: optional<integer>
+            docs: Number of scores to return per experiment when `fields=scores` is requested. Maximum 50, default 50.
+          cursor:
+            type: optional<string>
+            docs: Versioned base64url cursor from the previous response page.
+          fromTime:
```
wochinge: renamed it from `fromStartTime` to `fromTime` to make it agnostic on which field in the table we apply it

fern/apis/server/definition/experiments.yml
```diff
@@ -0,0 +1,197 @@
...
+        type: datetime
+        docs: |
+          Start of the experiment, i.e. the earliest event within the
+          requested time range. Clipped to `fromStartTime` when the
+          experiment started before the requested range.
+      endTime:
+        type: datetime
+        docs: |
+          End of the experiment, i.e. the latest event end within the
+          requested time range.
```
wochinge, line 134: added this on top of the proposal.
Because of the way the query is shaped now pages only include events whose _last_ event was part of the time window.

Example:
- Experiment A: starts at 9am, last event 11am
- Experiment B: starts and end at 10am

With page size 1:
- we first select the experiments in the last page (based on descending) -> Experiment A has last timestamp, but startedTime 9am
- B ends up on second page (with startTime 10am)

by surfacing the `endTime` we make this behavior transparent

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -41,14 +41,47 @@ export interface SplitQueryBuilder extends QueryWithParams {
...
+export type OrderByColumnsOptions = { eventTableAlias?: string };
+
+const buildOrderByClause = (
+  entries: OrderByEntry[],
+  options: OrderByColumnsOptions = {},
+) => {
+  if (!entries.length) {
+    return undefined;
+  }
+
```
sumerman, line 81: Lost a comment

```suggestion
    // When ordering by start_time, prepend project_id and toStartOfMinute(e.start_time)
    // to match the table PRIMARY KEY: (project_id, toStartOfMinute(start_time), xxHash32(trace_id))
```

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -41,14 +41,47 @@ export interface SplitQueryBuilder extends QueryWithParams {
...
+  const columns: string[] = [];
+  const { eventTableAlias } = options;
+  const startTimeColumn = eventTableAlias
+    ? `${eventTableAlias}.start_time`
+    : undefined;
+  const startTimeEntry = startTimeColumn
+    ? entries.find((e) => e.column.replace(/"/g, "") === startTimeColumn)
+    : undefined;
+
+  if (startTimeEntry && eventTableAlias) {
```
sumerman: this is super automagical and fragile behavior for a helper function that is called from 2 different contexts. I would much rather make it an explicit boolean option and compute "has start time in the sort order" at the call site, while keeping startColumn selection (line 70) local to this function.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -552,6 +609,100 @@ abstract class AbstractQueryBuilder {
     return this;
   }
 
+  /**
+   * Add exact event-level start time lower bound.
+   */
+  withExactTimeFrom(startTimeFrom?: string | null): this {
```
sumerman: why add a pair of fully custom methods instead of building on top of e.g. `applyFilters` ?

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -552,6 +609,100 @@ abstract class AbstractQueryBuilder {
...
+  }
+
+  /**
+   * Add cursor support for experiment APIs.
+   *
+   * The cursor is applied to raw event rows before any optional aggregation.
+   * This keeps pagination on the same event-level ordering key for experiment
+   * summaries and experiment items.
+   */
+  withCursor(cursor?: {
```
sumerman: Both cursor implementations are far too specific to experiments to be a part of query builder. Observations `applyCursorPagination(` lives in a corresponding repository.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -552,6 +609,100 @@ abstract class AbstractQueryBuilder {
...
+  }): this {
+    return this.when(Boolean(cursor), (b) => {
+      if (!cursor) return b;
+
+      return b
+        .whereRaw(
+          // The plain start_time bound is redundant with the tuple comparison
+          // but lets the primary index prune granules newer than the cursor;
+          // the mixed tuple alone is not index-usable.
+          "e.start_time <= {lastTime: DateTime64(6)} AND (toStartOfMinute(e.start_time), e.start_time, e.experiment_id, e.span_id) < (toStartOfMinute({lastTime: DateTime64(6)}), {lastTime: DateTime64(6)}, {lastExperimentId: String}, {lastId: String})",
```
sumerman: I would expect `toStartOfMinute` to make any difference in this position. Could make some difference for the outer `e.start_time <= {lastTime: DateTime64(6)} `

sumerman: Generally, since it is redundant perhaps start_time pruning alone would be enough.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -552,6 +609,100 @@ abstract class AbstractQueryBuilder {
...
+          `e.experiment_id NOT IN (
+  SELECT e2.experiment_id
+  FROM events_core e2
+  WHERE e2.project_id = {projectId: String}
+    AND e2.experiment_id != ''
+    -- don't list experiments that were part of the previous page - that means
+    -- no newer item exists
+    AND e2.start_time >= {lastTime: DateTime64(6)}
+    -- experiments shouldn't take more than that so we checked back long
+    AND e2.start_time < {lastTime: DateTime64(6)} + ${cursor.lookbackInterval}
```
sumerman: I would very strongly prefer NOT to splice this as a string here since it is possible to express via query parameter.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -552,6 +609,100 @@ abstract class AbstractQueryBuilder {
...
+    AND (toStartOfMinute(e2.start_time), e2.start_time, e2.experiment_id, e2.span_id) >= (toStartOfMinute({lastTime: DateTime64(6)}), {lastTime: DateTime64(6)}, {lastExperimentId: String}, {lastId: String})
```
sumerman: here as well, I don't think clickhouse query analysis is smart enough to deal with `toStartOfMinute` inside of a tuple, but I might be wrong.

packages/shared/src/server/repositories/scores.ts
```diff
@@ -699,6 +699,187 @@ export const getScoresForObservations = async <
...
+  traceIds,
+  observationIds,
+  min,
+  toTimestamp,
+  scoreLimit,
+}: {
+  projectId: string;
+  traceIds: string[];
+  observationIds: string[];
+  min: Date;
```
sumerman: I understand that you are bounding the lookup window to experiment timeframe, but, as Claude pointed out, scores repository generally uses an offset. I think it is worth at least mentioning in a comment that this function behaves differently and time bounds are precise.

packages/shared/src/server/repositories/scores.ts
```diff
@@ -699,6 +699,187 @@ export const getScoresForObservations = async <
...
+                  AND (s.observation_id IS NULL OR s.observation_id = '')
+                  AND s.session_id IS NULL
+                  AND s.dataset_run_id IS NULL
+                )
+              )
+            ORDER BY s.event_ts DESC
+            LIMIT 1 BY s.id
+          ) s
+          ORDER BY s.event_ts DESC
+          LIMIT {scoreLimit: UInt32} BY s.trace_id
```
sumerman, line 806: Is it true that trace ids are never shared between experiments? (I suppose it is 🤔 )

web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,435 @@
...
+      // page experiment's first and last event.
+      "max(e.start_time) AS cursor_time",
+      "argMax(e.trace_id, (e.start_time, e.span_id)) AS cursor_trace_id",
+      "argMax(e.span_id, (e.start_time, e.span_id)) AS cursor_span_id",
+      // Counts item root spans to match what /experiment-items paginates.
+      "uniqIf(e.span_id, e.span_id = e.experiment_item_root_span_id) AS item_count",
+    )
+    .withExactTimeFrom(fromTime)
+    .withExactTimeTo(toTime)
+    .whereRaw(`e.experiment_id IN (${pageQuery})`, pageParams)
```
sumerman: could be a opportunity for a generic `withSubquery` that incapsulates `
  const { query: pageQuery, params: pageParams } =
    pageQueryBuilder.buildWithParams();`

web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,435 @@
...
+    .orderBy(
+      "ORDER BY toStartOfMinute(cursor_time) DESC, cursor_time DESC, experiment_id DESC, cursor_span_id DESC",
```
sumerman: Not sure it makes a difference to use `toStartOfMinute(cursor_time) DESC` over aggregated data.

web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,435 @@
...
+    publicApiExperimentItemSimpleFilterMappings,
+    params.advancedFilters,
+    publicApiExperimentItemColumnMappings,
+    publicApiExperimentItemColumnDefinitions,
+  );
+
+  const queryBuilder = new EventsQueryBuilder({ projectId: params.projectId })
+    .selectFieldSet(
+      "publicApiExperimentItemCore",
+      ...(params.includeDataset
```
sumerman: I feel like `when` + multiple `selectFieldSet` would be easier to read

web/src/features/experiments/server/public/repository.ts
```diff
@@ -0,0 +1,435 @@
...
+      }),
+  });
+
+  return rows;
+}
+
+export async function queryExperimentItemsForPublicApi(
+  params: QueryExperimentItemsParams,
+) {
+  const rows = await queryExperimentItemRowsForPublicApi(params);
```
sumerman, line 475: Do you have any plans to make experiments compatible with batch or blob exports? If so, it might make sense to make scores join streaming friendly right away.

packages/shared/src/server/repositories/scores.ts
wochinge: Would be very weird if it'd be the case.
But also don't think it would matter. If an experiment item references that trace, then the trace and its scores belong to that that trace and hence to the experiment.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: I was mirroring `withStartTimeFrom` -> can derive all of them from `applyFilters` 👍🏻

web/src/features/experiments/server/public/repository.ts
wochinge: Given that we only do this 1 I wouldn't generalize. In my opinion the relatively raw sql also the advantage of being very understandable compared to our hand-rolled primitives

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
sumerman: withStartTimeFrom has special logic :)

web/src/features/experiments/server/public/repository.ts
sumerman: We do CTEs very often I think in the future this could be an even more convenient primitive. But I can also do this after you PR lands.

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: yep, saw that now as well when I tried to migrate it as well :-)

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: I only added the `alias` part tbh - can make it explicit though

wochinge: Actually: `orderByColumns` actually wants that magic behavior, no? So either we adapt all callers of that or we leave it like this imo.

sumerman: Previously it was more contained: only one place and if you have start_time it prepended `toStartOfTheMinute(start_time)` now the condition is more complex: `startTimeEntry && eventTableAlias)` and it is called from two different call-sites.

web/src/features/experiments/server/public/repository.ts
wochinge: not at the moment. Would defer it for now (also because the `join` was problematic when I tried it out)

web/src/features/experiments/server/public/repository.ts
wochinge: done

web/src/features/experiments/server/public/repository.ts
wochinge: verified and dropped it 👍🏻

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: done

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: verified and dropped it 👍🏻

packages/shared/src/server/repositories/scores.ts
wochinge: done 👍🏻

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: verified and dropped it 👍🏻

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: moved 👍🏻

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: done

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: done

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
wochinge: brought it back

hassiebp (approved):
LGTM

PR #14612, opened 2026-06-28
fix: use sha256 for bullmq repeat keys

### LINKED ISSUE #14611: bug: worker fails in Node FIPS mode because BullMQ repeatable jobs default to MD5 hashing
### Describe the bug

The Langfuse worker can fail in Node.js FIPS mode because BullMQ repeatable-job key generation defaults to `md5`.

Langfuse `v3.195.0` depends on BullMQ `^5.76.3`:

https://github.com/langfuse/langfuse/blob/v3.195.0/packages/shared/package.json

In BullMQ `5.76.3`, repeatable jobs use `settings.repeatKeyHashAlgorithm` when provided, otherwise they fall back to `md5`:

https://github.com/taskforcesh/bullmq/blob/v5.76.3/src/classes/repeat.ts#L25-L30

That value is later passed to `crypto.createHash(...)`. In Node.js FIPS mode, `md5` is rejected.

BullMQ exposes a supported option for this:

https://github.com/taskforcesh/bullmq/blob/v5.76.3/src/interfaces/advanced-options.ts#L10-L14

Langfuse already centralizes BullMQ queue/worker options here:

https://github.com/langfuse/langfuse/blob/v3.195.0/packages/shared/src/server/redis/redis.ts#L258-L267

and workers pass those options through here:

https://github.com/langfuse/langfuse/blob/v3.195.0/worker/src/queues/workerManager.ts#L137-L151

However, Langfuse does not currently set `settings.repeatKeyHashAlgorithm`, so BullMQ falls back to `md5`.

### To reproduce

Run the Langfuse worker in an environment where Node.js is running with FIPS mode enabled, then initialize code paths that create/update BullMQ repeatable jobs.

Expected failure shape:

```text
Error: error:0308010C:digital envelope routines::unsupported
```

or an equivalent OpenSSL/FIPS error from `crypto.createHash("md5")`.

### Expected behavior

Langfuse should be able to run in Node.js FIPS mode without requiring downstream image mutation or patching files under `node_modules`.

A likely fix would be for Langfuse's central BullMQ option helper to pass a FIPS-safe hash algorithm, for example:

```ts
settings: {
  repeatKeyHashAlgorithm: "sha256",
}
```

If changing the default could affect existing repeatable-job keys, this could also be exposed as an environment variable, while documenting the migration/compatibility behavior.

### Environment

- Langfuse worker: `3.195.0`
- BullMQ: `5.76.3`
- Runtime: Node.js FIPS mode / hardened container image

Leundai:
Fixes langfuse/langfuse#14611.

## Summary

* Configure centralized BullMQ queue options with `settings.repeatKeyHashAlgorithm = "sha256"` so repeatable-job registration does not fall back to MD5.
* Keep BullMQ worker options typed separately from queue options; workers do not receive the queue-only repeat hash setting.
* Centralize repeatable schedule `{ pattern, key }` metadata in `packages/shared/src/server/redis/repeatableJobs.ts`.
* Add explicit legacy `repeat.key` values for existing repeatable schedules so deployments with already-registered MD5 repeat keys update the existing schedule instead of creating duplicate SHA-256 schedules.
* Add regression coverage for queue/worker option shaping and the legacy repeat-key constants.

## Why

BullMQ 5.76.3 defaults repeatable-job key hashing to `md5` when `settings.repeatKeyHashAlgorithm` is unset. Node.js FIPS mode rejects `crypto.createHash("md5")` with `ERR_OSSL_EVP_UNSUPPORTED`, so Langfuse workers can crash during repeatable schedule registration in FIPS-enabled runtimes.

Changing the default hash algorithm alone would change repeatable-job metadata keys for existing Redis data. This PR therefore supplies the historical BullMQ MD5 repeat keys explicitly for current schedules, avoiding runtime MD5 while preserving existing repeatable-job keys.

## Verification

* `pnpm --filter worker exec vitest run src/__tests__/bullmqVersionCheckOptions.test.ts`
* `pnpm exec prettier --check packages/shared/src/server/redis/redis.ts packages/shared/src/server/queues.ts packages/shared/src/server/redis/cloudUsageMeteringQueue.ts packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts packages/shared/src/server/redis/postHogIntegrationQueue.ts packages/shared/src/server/redis/mixpanelIntegrationQueue.ts packages/shared/src/server/redis/dataRetentionQueue.ts packages/shared/src/server/redis/dlqRetryQueue.ts packages/shared/src/server/redis/eventPropagationQueue.ts packages/shared/src/server/redis/meteringDataPostgresExportQueue.ts packages/shared/src/server/redis/blobStorageIntegrationQueue.ts packages/shared/src/server/redis/coreDataS3ExportQueue.ts worker/src/__tests__/bullmqVersionCheckOptions.test.ts --experimental-cli`
* `pnpm --filter @langfuse/shared run typecheck`
* `pnpm --filter worker run typecheck`
* `pnpm --filter @langfuse/shared run build`
* `pnpm --filter @langfuse/shared run lint`
* `pnpm --filter worker run lint`
* Pre-commit hook: `prettier --check "**/*.{js,jsx,ts,tsx,css}" --experimental-cli` and `turbo run lint`


worker/src/__tests__/bullmqVersionCheckOptions.test.ts
```diff
@@ -87,7 +88,7 @@ const importRedisModule = async (
   return { redisInstances, redisModule };
 };
 
-describe("BullMQ Redis version check options", () => {
+describe("BullMQ Redis options", () => {
```
Steffen911, line 91: Could you also rename the file itself given the new purpose?

packages/shared/src/server/redis/redis.ts
```diff
@@ -272,14 +275,19 @@ const getBullMQOptionsForRedisConnection = (
...
     enableOfflineQueue: false,
     ...redisQueueRetryOptions,
   });
 
   return connection
-    ? getBullMQOptionsForRedisConnection(queueName, connection)
+    ? {
+        ...getBullMQBaseOptionsForRedisConnection(queueName, connection),
+        settings: {
+          repeatKeyHashAlgorithm: "sha256",
```
Steffen911, line 288: Can we make this configurable using an environment variable instead of changing the default here? I think keeping the standard would be preferred with an _option_ to switch in case compliance requires it?

packages/shared/src/server/redis/repeatableJobs.ts
```diff
@@ -0,0 +1,91 @@
...
+type BullMQLegacyRepeatableJobName = "BlobStorageIntegrationHourlyJob";
+
+// BullMQ's historical md5 repeatable-job keys for existing schedules. Supplying
+// them explicitly avoids runtime md5 in FIPS mode without duplicating schedules
+// that were already registered with BullMQ's previous md5 default.
+export const BullMQRepeatableJobOptions = {
+  [QueueJobs.CloudUsageMeteringJob]: {
+    pattern: "5 * * * *",
+    key: "377ab885ab66e6719c7ead551191bee4",
+  },
```
Steffen911, line 29: Can you add more detail on how this key is derived? I'm fine with pulling this into a separate utility, but given the other comment I'd prefer if md5 remains the default.

Leundai:
Closing this in favor of #16252.

The FIPS problem in #14611 is still valid, but #16223 replaced repeatable jobs with BullMQ Job Schedulers and now owns the recurring-job migration. Rebasing this PR would reintroduce the deprecated path and conflict with that design.

#16252 keeps the scheduler architecture, preserves normal behavior by default, and adds a guarded FIPS rollout. It will not register a second scheduler when legacy repeatable schedules remain.

PR #14617, opened 2026-06-29
perf(filters): load event filter options lazily, per-facet on demand (LFE-10519)

nkabardin:
## TL;DR
The v4 events filter sidebar + search bar bulk-loaded **every** facet's options on mount — one ClickHouse scan computing `approx_top_k` for ~18 columns **plus up to 3 separate score-name queries**, including high-cardinality `userId`/`sessionId` that aren't even shown by default. This PR loads them **lazily, per facet, on demand, and caches each one independently**: a fresh load requests **5 columns and 0 score queries**, and opening a facet fetches **only that facet** (the rest are served from cache).

## Why
Loading all facet options is the dominant cost of opening the events table, and most of it is wasted — the user sees a handful of facets and rarely touches `userId`/`sessionId`/scores. (Continues the "time-cut" work that already scoped filterOptions to the visible time range.) Goal: be lazy and gentle on the backend — only load the facets the user actually sees/uses, fetch each one once, and show a loading state while it streams in.

## What
`useEventsFilterOptions` gains a `lazy` mode that requests only the eagerly-visible facets (`environment`/`name`/`type`/`level`/`isRootObservation` = sidebar default-expanded ∪ search-bar suggestion fields) and loads everything else on demand:

- **Sidebar** — a facet's options load when its section is expanded (an effect on `queryFilter.expanded`; also covers active filters, which auto-expand on mount).
- **Search bar** — a field's values load when the caret enters its value stage. The pure completion planner emits `requestColumns`; the composer relays it. `scores.`/a score value requests the score-name columns.
- **Ask AI** — the grounding columns load when AI mode opens (so the prompt still sees real values).

**Independent per-facet caching.** One bulk query covers the eager set; every on-demand facet is its **own** cached query (`api.useQueries`, merged via `combine`). Opening a new facet fetches **only that column** — already-loaded facets are served from cache and never re-fetched, and re-collapsing/re-opening is free.

**Per-facet loading state.** The hook returns a precise `loadingColumns` set (requested-but-not-arrived, gated on an in-flight fetch) that drives the sidebar skeleton for exactly those facets — never for never-enumerated facets like `metadata`, and never forever on error. The search-bar dropdown's "Loading values…" row is keyed **per field** and settles to empty on a terminal error (symmetric with the sidebar).

**Backend contract.** `getEventFilterOptions` returns **only the requested columns** (previously every column with `[]` for unrequested ones, which made a lazily-loadable facet indistinguishable from a loaded-but-empty one). When everything is requested (the default, non-lazy path) all keys are present as before — monitors, widgets, dashboards, and session detail are unchanged.

## Result
- **5 columns + 0 score queries** on first load vs **~18 columns + 3 score queries** before.
- Opening a facet fetches **just that facet**; `userId`/`sessionId`/scores never load until shown or typed; re-opening is cache-served.
- Sidebar, search-bar suggestions, and Ask-AI grounding all still work; loading state is per-facet; errors settle to empty.

<details>
<summary>Mechanism, safety &amp; verification</summary>

- **The ClickHouse win.** The column-scoped query builds `approx_top_k` only for the requested columns and gates the (separate) score-name queries on whether score columns are requested. Per-facet caching means each facet is one single-column scan, fetched once.
- **Stable merge.** Eager bulk query + per-column `useQueries` combined via `combine` (react-query-memoized), so the merged payload identity is stable between changes (no downstream churn). The lazy column set only grows; collapsing never narrows it.
- **Planner stays pure.** Lazy-loadability is derived from the field registry (`exactOption`/`arrayOption` ⇒ `f.id` is the filter-options column), so no events-specific knowledge is injected into the grammar. `name` is eager and `id` has no server option list, so neither shows an endless loading row.
- **Error handling.** `loadingColumns` (sidebar) is gated on `isFetching`; the planner (bar) is gated on the hook's `isError` (`optionsErrored`). A terminal fetch error settles both to the empty state rather than skeletoning forever (no auto-retry: `staleTime: Infinity`).
- **Scoped to v4 events table.** `lazy` is opt-in; v3 and the session/widget/monitor filter-option consumers keep the bulk behavior.
- **Verified live** (network tab): first load → `[environment, name, type, level, isRootObservation]`; expand User ID → `[userId]` only; expand Provided Model Name → `[providedModelName]` only (renders `gpt-4o`); re-collapse + re-open User ID → no request; type `sessionId:` in the bar → `[sessionId]` only; type `scores.` → score-name columns, dropdown shows real names.
- **Tests.** Unit coverage in `completions.clienttest.ts` for per-field/score lazy loading, request emission, and error settling; the `getEventFilterOptions` subset test asserts unrequested columns are absent; search-bar (160) + filter-integration + sidebar-persistence + data-table-controls suites pass; web typecheck + lint clean.

</details>

Related (same filter layer, open): #14515 / #14562 (nested AND/OR), #14443 (Ask AI grounding), #14529 (selected options at top).


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #14570, opened 2026-06-25
feat(charts): in-view charts for the v4 events table ("any view is a chart")

nkabardin:
**Ticket:** [LFE-11023](https://linear.app/langfuse/issue/LFE-11023)

## TL;DR
The v4 events table can now become a **chart**. A **Table ⇄ Chart toggle** turns the current view into a live, in-view-configurable chart (chart type × metric × aggregation × breakdown), reusing the dashboard widget/chart system — plus **Ask AI → chart** and **Add to dashboard**.

## Why
Exploring the events table, you can't visualize the data you're already looking at without leaving for the dashboard builder and rebuilding your filters. The `chart-library` and the query engine already exist — so the events view should chart *exactly what you filtered to*, and let you drop that chart onto a dashboard in one step.

## What
- **Table ⇄ Chart toggle** on the full-page v4 events view — URL-driven, reversible, deep-linkable.
- **In-view "Visualize" panel:** chart type (line / area / bars / ranked / pie / number) × metric × aggregation × breakdown.
- **Same data as the table:** reads the live filters + time range; renders identically to the equivalent dashboard widget.
- **Filter-compatibility disclosure:** filters the chart's aggregate query can't apply (measures, scores, comments, metadata, `has:`, free-text) are shown **"not applied"** — dimmed + reason on hover — in the search bar *and* the sidebar, instead of hiding the chart. Forwards the observations dimensions the widget builder exposes (LFE-10751); scores/comments follow LFE-10756 / LFE-10757.
- **Ask AI → chart** (Cloud + AI-features + Bedrock gated) and **Add to dashboard** (via the existing `dashboardWidgets.create` + `SelectDashboardDialog`).

## Result
Flip to a chart, reconfigure in place, ask AI to build it, or drop it onto a dashboard — without leaving the events view or rebuilding filters.

<details>
<summary>Engineering detail (reuse, WYSIWYG parity, verification)</summary>

**Reuse, don't fork.** Data via the existing `dashboard.executeQuery` (`view:"observations"`, `version:"v2"` — the v4 events read path); rendering via the shared `chart-library`; placement via `dashboardWidgets.create` + `SelectDashboardDialog` + the dashboard's `?addWidgetId=` flow. No new endpoint, chart stack, or attach path. `features/chart-view/**` owns the toggle, panel, and pure data mappers (`buildChartQuery` / `rowsToDataPoints`, `coerceConfig`); `chartFilterCompatibility.ts` is the single source of truth for what forwards vs. shows "not applied".

[...]

PR #14618, opened 2026-06-29
feat(dashboard): chart polish, hover interactivity, and cross-chart sync (LFE-10549)

nkabardin:
## TL;DR

A charts overhaul for [LFE-10549](https://linear.app/langfuse/issue/LFE-10549): **polish** (de-animate, even ticks, one consistent legend), **interactivity** (hover-proximity line highlight + a crosshair synced across charts on the same timeline), wired into the **real home + custom dashboards**, plus a safe **perf** foundation. Phase-0 analysis found recharts v3 is fine — the issues were usage/consistency — so this fixes those, the modern way.

## Why

Charts hadn't been revisited since the Tremor→recharts migration: twitchy animations, uneven date ticks, a duplicated legend, and dense "spaghetti" charts that were hard to read. The fix for density is *interaction* (hover to read a line precisely; sync the time crosshair across panels), not throwing data away.

## What

**Polish (chart-library)**
- **De-animate** the Area/Line/Bar/Pie/Histogram series (only the tooltip was before).
- **Even x-axis ticks** via a computed numeric interval — replaces `interval="preserveStartEnd"`+`minTickGap`, whose width-dependent dropping skipped ticks unevenly.
- **One shared legend** (de-dups line/area, gives the bar time-series a legend) that **wraps** so it never clips; action-faithful aria labels. Additive metrics show a `sum`; non-additive (latency, scores) show the bare name (a sum is meaningless; an unlabeled non-sum reads ambiguously).

**Interactivity**
- **Hover proximity highlight:** the single line nearest the cursor is emphasized and the rest dim (a 2nd only if it overlaps within 1px); away from any line it renders normally; scoped to the chart under the cursor; the selected line's **tooltip row** is bolded to match.
- **Synced crosshair:** charts sharing a `syncId` move one crosshair + tooltip together; tooltips can escape a chart's bounds.

**Integration**
- **Home dashboard:** the time-series charts (traces/observations, model usage, generation latency, scores) share a `syncId` → synced crosshair.
- **Custom dashboards:** `DashboardWidget` passes `dashboardId` as `syncId` → all time-series widgets on a dashboard sync.

**Perf (safe foundation)**
- `React.memo(Chart)` so a scheduler-driven page re-render doesn't reconcile the recharts subtree when inputs are unchanged; `ChartScores` memoizes its data + config to make that effective.

## Result

Snappier, consistent charts; dense charts readable by hovering; and hovering any chart on a dashboard moves the time marker across all of them. **Verified live** (Playwright, real cursor) on the seeded home dashboard (hover → crosshair + tooltip on all 4 time-series charts at the same timestamp) and a custom dashboard.

<details>
<summary>Perf rationale & verification</summary>

**Why not memoize more aggressively:** the data-fetching chart components are intentionally NOT wrapped in `React.memo` — the query scheduler propagates "you may fetch now" via a page-level re-render, so a memoized chart with stable props would never pick up its fetch slot. The bigger lever (the scheduler re-rendering the whole page on every queue tick → move `canFetch` to per-query subscriptions) touches query orchestration / ClickHouse concurrency and is left as a measured follow-up. recharts v3's internal store already softens reconciliation, so measure before investing further.

**Verification:** `pnpm --filter web typecheck`, `eslint`, `prettier`, chart-library client tests, and CI fast checks green. Browser-verified the polish + proximity + synced crosshair on the home and a custom dashboard; confirmed no load regression (charts fetch + render; nothing stuck).

**Storybook:** `ChartInteractivity` (synced dashboard + spaghetti/proximity) and `LineChartTimeSeries` stories cover the behaviors.

Related: #14538 (the original `legendSummary`), #14570 (the separate "any view is a chart" prototype).
</details>


nkabardin:
@claude review

nkabardin:
@claude review

PR #14621, opened 2026-06-29
feat(agent): Add langfuse skills

bezbac:

This PR wires two Langfuse-specific skills (error-analysis and CLI reference) into the in-app agent by loading them from markdown files at build time and passing them to the Mastra `Agent` constructor. It also ships a developer utility (`sync-raw-skills.mjs`) for pulling the latest skill markdown from an upstream GitHub repository.

- **Skill loading**: A new `skills.ts` module reads `.md` files via static imports (handled by webpack `asset/source` / Turbopack `raw-loader` / a custom Vitest plugin), parses YAML frontmatter with a hand-rolled parser, and constructs Mastra skill objects at module-load time.
- **Sync script**: `scripts/in-app-agent/sync-raw-skills.mjs` fetches all `.md` files from `langfuse/skills` on GitHub, filters to those with `required_access: LANGFUSE_PROJECT_INTERFACE`, and writes them to `skills/raw/`. A `--check` flag enables CI validation that local files match upstream.
- **Dependency bump**: `@mastra/core` is upgraded from `1.37.1` to `1.46.0` alongside the changes.
</details>

Safe to merge — the runtime path is straightforward and well-tested; the concerns are around developer tooling and future-proofing.

The core change (passing skills to the Mastra Agent) is minimal, tested, and works correctly with the current skill files. The hand-rolled YAML frontmatter parser handles the two current files correctly but will silently produce wrong values if future skill files use quoted strings with colons or other non-trivial YAML. The disconnect between the sync script (dynamic) and the skills registry (hardcoded static imports) means developers must remember a two-step update. The sync script's lack of GitHub token support could block CI --check runs against a private repo.

web/src/ee/features/in-app-agent/server/skills.ts (fragile YAML parser and manual import list) and scripts/in-app-agent/sync-raw-skills.mjs (unauthenticated GitHub API calls).
</details>

`````markdown

---

### Issue 1 of 3
web/src/ee/features/in-app-agent/server/skills.ts:53-73
**Hand-rolled YAML parser will silently misparse non-trivial frontmatter**

`parseFrontmatter` matches only bare top-level keys (`^([A-Za-z0-9_-]+):\s*(.*)$`) and appends any other line as a continuation of the previous key. Quoted strings containing `:`, real block scalars (`|`, `>`), and multi-level nesting will all be silently mangled. As new skill files are added from the remote repo via `sync-raw-skills.mjs`, this parser could silently produce wrong `name`/`description` values if upstream authors use slightly different YAML conventions. Using a lightweight YAML parser (e.g., `js-yaml`) for the frontmatter block would make this robust.

### Issue 2 of 3
web/src/ee/features/in-app-agent/server/skills.ts:12-15
**New files synced by `sync-raw-skills.mjs` won't become active skills automatically**

`LANGFUSE_IN_APP_AGENT_SKILL_MARKDOWN` is a hardcoded list of two static imports. When `sync-raw-skills.mjs` downloads additional skill files from the upstream `langfuse/skills` repo into `skills/raw/`, those files are placed on disk but never imported here and never reach the `Agent`. A developer who runs the sync script has no indication that `skills.ts` also needs a new static import and array entry. Consider adding a build-time assertion that every `.md` file in `skills/raw/` (excluding `README.md`) is present in this array, so the gap fails loudly rather than silently.

### Issue 3 of 3
scripts/in-app-agent/sync-raw-skills.mjs:25-40
**GitHub API requests are unauthenticated, which will fail for private repos and hit rate limits in CI**

Both `listRemoteMarkdownFiles` and `fetchText` send no `Authorization` header. The GitHub API returns 404 for private repositories without auth, and anonymous requests are rate-limited to 60/hr. If `langfuse/skills` is private or the `--check` mode runs in CI on a busy day, the script will fail with a non-descriptive 404 or 429. Reading an optional `GITHUB_TOKEN` environment variable and forwarding it as `Authorization: Bearer ${process.env.GITHUB_TOKEN}` when present would cover both cases without requiring a token in non-CI contexts.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(agent): Add langfuse skills"](https://github.com/langfuse/langfuse/commit/738c137c0b39f55d1fc07afd37d28f84b0138d9f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40512563)</sub>

### CODE REVIEW

File: web/src/ee/features/in-app-agent/server/skills.ts [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+type LangfuseInAppAgentSkillDefinition = {
+  name: string;
+  description: string;
+  instructions: string;
+};
+
+const LANGFUSE_IN_APP_AGENT_SKILL_MARKDOWN = [
+  errorAnalysisMarkdown,
+  cliMarkdown,
+] as const;
```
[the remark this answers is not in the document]
bezbac: This is fine

scripts/in-app-agent/sync-raw-skills.mjs
```diff
@@ -0,0 +1,220 @@
...
+
+const isCheckMode = process.argv.includes("--check");
+
+const getGitHubHeaders = () => ({
+  Accept: "application/vnd.github+json",
+  Authorization: process.env.GITHUB_TOKEN
+    ? `Bearer ${process.env.GITHUB_TOKEN}`
+    : undefined,
+  "User-Agent": "langfuse-sync-in-app-agent-raw-skills",
+});
```
nimarb, line 34: do we need to set the tokens here? should be public

bezbac: It is public but the rate limits are different. This was a greptile suggestion: [https://github.com/langfuse/langfuse/pull/14621#discussion_r3491930147](<https://github.com/langfuse/langfuse/pull/14621#discussion_r3491930147>)

web/src/ee/features/in-app-agent/prompts/in-app-agent-system-prompt.txt
```diff
@@ -25,6 +25,7 @@ IMPORTANT: You should NOT answer with unnecessary preamble or postamble (such as
 
 <tools>
 Use the docs tools to find relevant general information about Langfuse or best practices.
+For questions about best practices or specific workflows (e.g. setting up evals), use the available Langfuse skills when they apply.
```
nimarb, line 28:
```suggestion
For questions about best practices or specific workflows (e.g. setting up evals), use the available Langfuse skills when they apply.
If the skill ever tells you to reference the docs, always use the langfuse-docs-mcp server to access those docs.
```

scripts/in-app-agent/sync-raw-skills.mjs
nimarb, line 1: do we add all skills here? i think we should have a list of "blocked" skills at the very top which is easy to edit with a reason for why we leave it out. for example, i would not add:
* ci-cd.md, we cannot make CICD experiment changes from our agent yet
* cli.md, we cannot use the CLI from the agent
* skill-feedback.md, the agent doesn't have network access to submit feedback

bezbac: I talked to Lotte and Annabell about this. They marked which skills require which capabilities and we are only pulling the ones that are relevant for the in-app agent. I think maintaining a list here is not a good idea if this information is already included in the skill itself.

web/src/ee/features/in-app-agent/server/skills/generated/raw/cli.md
nimarb, line 1: would remove, see comment above

web/src/ee/features/in-app-agent/server/skills/raw/README.md
```diff
@@ -0,0 +1 @@
+These skills are pulled from https://github.com/langfuse/skills/tree/main, never update them manually.
```
nimarb:
```suggestion
These skills are pulled from https://github.com/langfuse/skills/tree/main, never update them manually.
Update them by running `scripts/in-app-agent/sync-raw-skills.mjs`, following the README.md in that folder.
```

web/src/ee/features/in-app-agent/server/skills.ts
```diff
@@ -0,0 +1,73 @@
...
+type LangfuseInAppAgentSkillDefinition = {
+  name: string;
+  description: string;
+  instructions: string;
+};
+
+const LANGFUSE_IN_APP_AGENT_SKILL_MARKDOWN = [
+  { fileName: "error-analysis.md", markdown: errorAnalysisMarkdown },
+  { fileName: "cli.md", markdown: cliMarkdown },
+] as const;
```
nimarb: perhaps this is actually the better location to set the skills, we can import this variable also in the sync script to just have 1 place of maintenance. or put this const into it's own file for clarity.

bezbac: I'll look into autogenerating it in the .mjs file 👍 . I think we should not actually maintain a whitelist.

web/package.json
```diff
@@ -174,6 +174,7 @@
     "uuid": "^14.0.0",
     "vaul": "^1.1.2",
     "vis-network": "^9.1.13",
+    "yaml": "^2.9.0",
```
nimarb, line 177: i think we have https://www.npmjs.com/package/js-yaml already in deps, can we use that? otherwise, yaml also fine

nimarb (changes requested):

web/package.json
bezbac: yaml is already pulled in as a transitive dep, this is correct. But e18n advises to use yaml over js-yaml: [https://e18e.dev/docs/replacements/js-yaml](<https://e18e.dev/docs/replacements/js-yaml>)

web/src/ee/features/in-app-agent/server/skills/generated/raw/cli.md
bezbac: I added this part back to the system prompt:
> If the user asks you to perform an action you cannot, see if the action is available in the UI or via the CLI, and suggest that to the user.
> - Explain to the user how they can perform the action themselves in the UI (use the docs for this if needed).
> - If the action is available via the CLI, suggest that the user can ask their own agent (Claude, Codex or similar) to perform the action for them using the CLI, for that they should use the Langfuse skill: https://github.com/langfuse/skills. When suggesting this, provide a prompt the user can use as a code block.

For this, I think it makes sense to keep the skill. If we see it confuses the models, we can remove it in the future :)

nimarb (approved):

PR #14691, opened 2026-07-01
fix(trace): stop v4 tree/search rows overlapping with many scores (LFE-10591)

nkabardin:
## TL;DR

In the v4 ("fast") trace view, a node with **lots of scores** made rows **overlap** — worst right after **Collapse all**. Root cause was a virtualization mis-keying, not the score display. Fixed with a one-line `getItemKey` per list, plus a compact `+N` score cap so dense rows stay tidy. Verified in the browser: a reproducible **25px** overlap → **0**.

## Why

A customer reported (at a conference, high visibility) that the v4 trace **Tree** glitches with many scores: rows render on top of each other, and hitting **Collapse all** makes it worse. The v3 ("slow") tree renders the same data correctly.

Each tree/search row is rendered by `SpanContent`, whose score badges **wrap over several lines** — so rows have dynamic, widely-varying heights (I measured 55–204px in a narrow pane). Both the Tree (`VirtualizedTree`) and the search results (`VirtualizedList`) virtualize with `@tanstack/react-virtual` using **dynamic per-row measurement**, but they keyed the measurement cache **by index** (the default) while keying the DOM rows **by node id**.

On any reorder — Collapse all, expanding/collapsing a mid-tree node, or typing in search — React reuses a row's DOM element **by id** without resizing it, so the virtualizer never re-measures and keeps the **previous** occupant's height at that index. The `translateY(start)` offsets then drift out of sync with the real heights and rows overlap. The more scores a node has, the taller/more variable the rows, and the worse it gets.

## What

1. **Correctness** — give both virtualizers a `getItemKey` returning the node/item id, so each measurement **travels with its node** across reorders and offsets stay correct. This alone eliminates the overlap without changing what's shown.
2. **Density** — cap inline scores at 3 groups per row with the existing `+N` hover pill (matching the timeline's `TimelineBar`), so a node with many scores is a compact one-line row instead of a tall wrapping grid. The full set stays available via the pill and the node's **Scores** tab.
3. **Repro** — extend the `trace-tree` seeder with `--scores-per-node N` so the "lots of scores" shape stays cheaply reproducible.

## Result

Reproduced the exact bug locally (deep v4 trace, 12 scores/node, narrow pane): collapsing a few mid-tree nodes produced a **persistent 25px overlap across 4 row pairs**. After the fix: **0 overlapping pairs**, holding through collapse/expand, scroll, and live search filtering.

<details>
<summary>Mechanism, verification & scope detail</summary>

**Why index-keying breaks:** `@tanstack/react-virtual` caches item sizes by `getItemKey(index)` (default `index => index`). The row DOM is keyed `key={node.id}`. When the flattened list reorders, React reconciles by id and **reuses** the element — no size change → the internal `ResizeObserver` never fires → the size stored at that index is stale (it belongs to whichever node was there before). `getItemKey: index => flattenedItems[index].node.id` makes the cache follow the node, so a reordered row's correct measured height is reused.

**Measured evidence (Playwright + computed geometry):**
- Before: expanded/scrolled = clean; after collapsing mid nodes = `maxOverlap 25px`, 4 overlapping pairs, persistent.
- After (getItemKey + cap): `0` overlapping pairs across collapse/expand, a full scroll-stress pass, and progressive search filtering; rows now a uniform 55–108px.

**Scope:** both `VirtualizedTree` (tree) and `VirtualizedList` (search results) share the pattern and both render `SpanContent`, so both are fixed. The **Timeline** view is unaffected — it uses a fixed 26px row height with no dynamic measurement, and already caps inline scores.

**Tests:** this is a real-layout / DOM-measurement bug (jsdom reports 0 heights, so it can't be unit-tested meaningfully); verification is the browser repro above. The seeder flag keeps it reproducible for future regressions.
</details>


nkabardin:
@claude review

PR #14694, opened 2026-07-01
fix: encode clickhouse blob storage object keys

### LINKED ISSUE #14457: [Bug]: Observation/trace ids used verbatim as S3 object keys; non-S3-safe ids (base64 with '=') fail on MinIO
LangFuse v3 builds the blob-storage key as `<projectId>/observation/<id>/<uuid>.json`
using the client-supplied observation id unescaped. When an upstream (here LiteLLM's
Responses API) supplies an id containing base64 padding/charset, MinIO rejects the
upload with `XMinioInvalidObjectName: Object name contains unsupported characters`. The
ingestion POST has already 200'd, so the event is lost with no caller-visible error and
nothing reaches ClickHouse.

### Observed

- Caller: LiteLLM `langfuse` success-callback logging a Responses-API generation whose
  id is `resp_<base64url>=` (base64 padding).
- Storage: external MinIO, S3 blob storage.
- `langfuse-web` logs:

```
error Failed to upload JSON to S3 <projectId>/observation/time-..._resp_<base64>=/<uuid>.json Object name contains unsupported characters.
error Failed to upload event to S3
Error: Failed to upload events to blob storage, aborting event processing
```

`langfuse-worker` stays clean (nothing was ever enqueued), and the trace never appears
in the UI despite a 200 on ingestion.

### Suggestion

URL-encode / sanitize / hash externally-supplied ids before using them as object keys,
so a hostile or non-S3-safe id degrades to a safe key rather than a silently dropped
event. AWS S3 tolerates these ids; MinIO (a very common self-hosted target) does not, so
the behavior is storage-backend-dependent today.

### Environment

- LangFuse v3 (langfuse-k8s chart), ClickHouse + MinIO (S3) blob storage
- Upstream id source: LiteLLM proxy Responses API (`resp_<base64>=`)

stutimohindra:
## What does this PR do?

Encodes blob storage object key path segments for ClickHouse upserts and standard ingestion event uploads.

Observation, trace, and event IDs can be supplied by upstream integrations and may contain characters rejected by some S3-compatible backends. In particular, LiteLLM Responses API IDs may contain base64 padding such as `=`, which can cause MinIO uploads to fail with `XMinioInvalidObjectName`.

This PR URL-encodes user-supplied entity IDs and generated event/file IDs before building blob storage object keys, while preserving the original unencoded IDs in ClickHouse metadata and queue payloads.

Fixes #14457

## Type of change

## Mandatory Tasks

## Testing

- `pnpm test`
- `eslint src/server/repositories/clickhouse.ts --max-warnings 0`
- `eslint src/server/ingestion/processEventBatch.ts --max-warnings 0`
- `eslint src/__tests__/server/repositories/clickhouse-upsert.servertest.ts --max-warnings 0`
- `eslint src/__tests__/server/unit/blobStorageKey.servertest.ts --max-warnings 0`


stutimohindra:
@claude review

packages/shared/src/server/ingestion/processEventBatch.ts
```diff
@@ -251,14 +252,14 @@ export const processEventBatch = async (
           // write time and rides the queue payload as `fileKey`. Sanitize
           // here so `/`, `\`, control bytes, and over-budget lengths can't
           // reroute the write or overflow NAME_MAX.
-          key: safeBlobFilenameStem(event.id, ".json"),
+          key: safeBlobFilenameStem(encodeURIComponent(event.id), ".json"),
```
Steffen911, line 255: Unfortunately, this will break update flows that we support for existing SDKs. The worker-side relies on a consistent prefix for all updates for the same entity identifier. If we roll this out, then partial updates that were written _before_ the new version will not be reflected in newly processed records and may cause duplicates.

I think this needs to be solved upstream where the instrumentations use a restricted set of characters. 

If we want to include a warning in the response body or at least via a log message this seems good though.

Additionally, this problem should go away with newer Langfuse SDK versions that are OTel based as those produce 32-char alphanumeric IDs that are fully safe.

Steffen911:
Given the reasoning in https://github.com/langfuse/langfuse/pull/14694#discussion_r3524614111 this approach seems problematic and therefore I'll close the PR. Thank you @stutimohindra for giving it a try though and for your contribution!

PR #14712, opened 2026-07-02
feat(agent): Add sandbox

bezbac:

This PR introduces a sandboxed code execution environment for the in-app agent feature. It adds two sandbox providers — a Docker-based one for local development and a Lambda MicroVM provider for production — exposing `read`, `write`, `edit`, and `bash` tools to the agent. Conversation sandbox state (session ID, provider type) is persisted in the database, and a write-lock mechanism prevents new turns to sandbox conversations older than 8 hours.

- **New sandbox providers**: `dangerous-docker` (dev-only) launches isolated Docker containers; `lambda-microvm` starts AWS Lambda MicroVMs, both exposing a small HTTP server that validates paths and runs operations.
- **State persistence**: `InAppAgentSandboxProvider` enum and `sandboxProvider` column are added to `InAppAgentConversation`, allowing provider sessions to be resumed across requests.
- **Tool-call file sync**: Prior non-sandbox tool call results are reconstructed from persisted events and mounted read-only into the sandbox's `tool_calls/` directory each turn.
</details>

The PR is safe to merge; all three findings are non-blocking quality issues rather than data-loss or correctness bugs.

The core sandbox logic — path escaping, provider lifecycle, state persistence, write-lock enforcement — is well-structured and tested. The edit first-occurrence behavior and missing body-size cap are real but low-impact gaps. Dynamic imports inside function bodies are a deliberate lazy-loading choice that conflicts with the team convention.

packages/in-app-agent-sandbox-runtime/src/server.ts has two issues worth addressing before relying on the edit tool at scale; web/src/ee/features/in-app-agent/server/sandbox/config.ts and providers/docker.ts have the import placement concern.
</details>

`````markdown

---

### Issue 1 of 3
web/src/ee/features/in-app-agent/server/sandbox/config.ts:48-49
**Dynamic import inside function body**

`createInAppAgentSandboxProvider` uses `await import("./providers/docker")` at call time rather than at module load time. The comment notes this is intentional for dev-only lazy loading, but it directly violates the team rule that imports must live at the top of the module. The same pattern appears in `docker.ts` at the `createDockerSandboxProvider` factory where `dockerode` is also dynamically imported inside the function body.

### Issue 2 of 3
packages/in-app-agent-sandbox-runtime/src/server.ts:253-261
**`edit` silently replaces only the first occurrence**

`String.prototype.replace()` replaces the first match only. If `oldText` appears more than once in the file the response still returns `{ replaced: true }`, giving the agent no signal that it edited the wrong occurrence. The agent may then loop incorrectly, believing the targeted span was not modified. Consider either throwing (or returning `{ replaced: false, ambiguous: true }`) when multiple occurrences are detected, so the agent can supply more specific context.

### Issue 3 of 3
packages/in-app-agent-sandbox-runtime/src/server.ts:418-430
**No request body size limit**

`readJsonBody` accumulates the entire incoming body into a string with no ceiling. A large `write` payload (or a malformed/oversized request) will buffer unboundedly before the Zod schema even runs. Since the sandbox container has finite memory, this could exhaust it and crash the process. A simple guard — e.g. tracking accumulated bytes and throwing once they exceed a reasonable ceiling (say 10 MB) — would protect against accidental or adversarial oversized payloads.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["Regenerate migration"](https://github.com/langfuse/langfuse/commit/52796d0c14ca57052872a3b3d8210b56d7dbfd15) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43016861)</sub>

**Context used:**

**Learned From**

### CODE REVIEW

**Comment by bezbac** [comment]
@greptileai

File: packages/shared/prisma/schema.prisma
Comment (bezbac) [inline packages/shared/prisma/schema.prisma:1]: Remove these changes, assume the provider is stable.

File: packages/shared/prisma/migrations/20260709135752_in_app_agent_sandbox_state/migration.sql
Comment (nimarb) [inline packages/shared/prisma/migrations/20260709135752_in_app_agent_sandbox_state/migration.sql:1]: remove as discussed

File: packages/in-app-agent-sandbox-runtime/Dockerfile
Comment (bezbac) [inline packages/in-app-agent-sandbox-runtime/Dockerfile:1]: I extracted a list of tools from the agentic coding langfuse and will make sure these work within the sandboxes:
```
grep
tail
head
jq
find
sed
echo
printf
mkdir
cp
mv
uniq
sort
cut
wc
cat
date
awk
lsof
ls
env
printenv
```

bezbac: All of these are available through BusyBox + jq is manually installed.

nimarb (approved):

PR #14722, opened 2026-07-02
perf(api): two-phase trace lookup for GET /api/public/sessions/:sessions

Steffen911:

This PR introduces a two-phase ClickHouse lookup for `GET /api/public/sessions/:sessionId` to avoid reading wide `input`/`output` columns across all partitions when the query is filtered only by `session_id`. It also adds a `fromTimestamp` query parameter that controls the trace timestamp lower bound, defaulting to `session.createdAt - 2 days`.

- **Phase 1** runs a narrow `SELECT id, min(timestamp) GROUP BY id` to collect trace IDs and their minimum timestamps for a session. **Phase 2** calls `getTracesByIds` per 500-ID chunk using the per-chunk minimum timestamp to enable ClickHouse partition pruning.
- The `fromTimestamp` parameter is wired through the Fern API spec, Zod schema, and handler; when omitted, the handler substitutes the 2-day lookback default, which is a behavioral change from the previous behavior of returning all traces regardless of timestamp.
</details>

Safe to merge only after verifying that no production callers rely on traces backdated more than 2 days before session creation, or after adding a mechanism that signals to clients when the response is incomplete.

The core two-phase ClickHouse optimization is logically correct and the tests cover the intended scenarios. The concern is the default 2-day lower bound: before this PR, the endpoint returned every trace for a session unconditionally; now any trace with a timestamp older than session.createdAt - 2 days is silently dropped from the response with no indication to the caller.

web/src/pages/api/public/sessions/[sessionId].ts — the default timestamp lower bound is a behavioral change for every existing consumer of this endpoint who does not pass fromTimestamp.
</details>

`````markdown

---

### Issue 1 of 2
web/src/pages/api/public/sessions/[sessionId].ts:52-58
**Silent breaking change: traces silently dropped for existing clients**

Before this PR, `getTracesBySessionId` was called with no timestamp filter, so the response always included every trace for the session. Now a lower bound of `session.createdAt - 2 days` is applied by default. Any existing API consumer calling this endpoint without `fromTimestamp` will silently receive an incomplete list of traces if the session has backdated traces (e.g. OTEL backfills, offline agents). The API response contains no `truncated` flag or `totalCount` that would signal the omission, so callers have no way to detect the data loss.

### Issue 2 of 2
packages/shared/src/server/repositories/traces.ts:314-330
**Stale version returned when a trace's latest event has a timestamp below the lower bound**

The chunk-level lower bound passed to `getTracesByIds` is `Math.min(...chunk.map(t => t.min_timestamp))`. If a trace has its highest `event_ts` row stored at a `timestamp` below the user-provided lower bound, that row is excluded from both phases and the query silently returns an older version. This limitation pre-existed the timestamp-filter design but becomes more likely to affect real datasets now that the default 2-day lower bound is applied unconditionally.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(api): fromTimestamp override for se..."](https://github.com/langfuse/langfuse/commit/cb4ee31e5052b86574ea3774eff3c5f7f5f1dadf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41373928)</sub>

### DIFF
--- a/fern/apis/server/definition/sessions.yml
+++ b/fern/apis/server/definition/sessions.yml
@@ -50,6 +50,16 @@ service:
         sessionId:
           type: string
           docs: The unique id of a session
+      request:
+        name: GetSessionRequest
+        query-parameters:
+          fromTimestamp:
+            type: optional<datetime>
+            docs: |
+              Optional lower bound for the traces returned with the session (ISO 8601).
+              Only traces with a timestamp on or after this value are included.
+              Defaults to the session's createdAt minus 2 days, so traces backdated
+              further need an explicit fromTimestamp.
       response: commons.SessionWithTraces
 types:
   PaginatedSessions:
--- a/packages/shared/src/server/repositories/traces.ts
+++ b/packages/shared/src/server/repositories/traces.ts
@@ -15,6 +15,7 @@ import { orderByToClickhouseSql } from "../queries";
 import { shouldSkipObservationsFinal } from "../queries/clickhouse-sql/query-options";
 import { LISTABLE_SCORE_TYPES } from "../../domain/scores";
 import { OrderByState } from "../../interfaces/orderBy";
+import chunk from "lodash/chunk";
 import snakeCase from "lodash/snakeCase";
 import { FilterState } from "../../types";
 import {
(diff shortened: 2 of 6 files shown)


Steffen911:
Closing as moving to hot cached instance resolves it.

PR #14716, opened 2026-07-02
fix(trace): comfortable peek defaults, unify inner-split persistence (LFE-10601)

nkabardin:
## TL;DR
The trace **peek** opened with worse defaults the bigger your screen — a very wide tree and a cramped info panel — and the tree/info ratio reset between tabs. This makes the peek open at a **comfortable, screen-aware default** (info wide, tree bounded, list still navigable) and **remembers your inner split in `localStorage`** like it already does the outer width. Scoped to the peek; the full-page trace view is untouched.

## Why
Two independent, separately-persisted resize systems stack in the peek:
1. **Outer width** — `0.5 × viewport`, so on a 2560px+ monitor the peek was ~1280px+ and kept growing.
2. **Inner tree ↔ info split** — a *share-based* 60% tree / 40% info, persisted in **per-tab `sessionStorage`**.

Because the tree was a *percentage of an already-large peek*, "bigger screen → bigger tree": at 2560px the tree was ~768px and info a cramped ~512px. And per-tab storage is why the split felt inconsistent between tabs and between people (one long-lived tab vs. many "open in new tab" tabs).

## What
- **Outer width — cap the default's px growth.** Default = `min(0.5·vw, PEEK_MAX_DEFAULT_WIDTH_PX)`, floored at the drag minimum. Normal laptops keep 50vw (zero change ≤ ~2800px); wide monitors stop ballooning so the underlying list stays navigable. User drags still go up to 90%.
- **Inner split (peek only) — anchor to comfort, give extra to info.** Give the info panel a comfortable target and hand the tree the remainder, clamped to a comfortable band, so extra width on a wide peek flows to the *content* (info), not the *index* (tree).
- **Unify persistence → `localStorage`.** The peek's inner split now persists cross-tab under a peek-scoped group id, matching the outer width. The full-page trace view keeps its own share-based, per-tab layout.

## Result (verified in-browser across widths)
| Viewport | Peek width | Tree | Info |
|---|---|---|---|
| 2560 | 1280 (50%) | **459** (was ~768) | **819** (was ~512) |
| 3440 | **1400** (capped, was 1720) | 460 | 940 |
| 1512 (laptop) | 756 (50%, unchanged) | 339 | 415 |
| 1280 | 640 | 278 (min-area) | 360 |
| 1100 | 550 | 260 (rail-min + horizontal scroll preserved) | 360 |

[...]

PR #14735, opened 2026-07-02
fix(trace): avoid minifier-dropped const in peek collapse boundary

vasyl-stanislavchuk:
## TL;DR

Opening a trace **peek** threw `ReferenceError: COLLAPSED_RAIL_BOUNDARY_PX is not defined` — but only in the **production build**, so dev never caught it. The SWC minifier constant-folded one call site of a small helper and **dropped the single-use module const it read**, while a second (dynamic) call site kept a dangling reference. Inlining the value as a **function-local const** removes the top-level binding the minifier could drop. Behavior is unchanged. Regression from #14716 (LFE-10601).

## Why

`db9ee72` (#14716) introduced a single-use, module-scoped const read **only** inside `collapsedShareMaxForWidth`, which runs in two places:

```ts
const COLLAPSED_RAIL_BOUNDARY_PX =
  (COLLAPSED_PANEL_PX + NAVIGATION_PANEL_MIN_PX) / 2;   // single-use, top-level
function collapsedShareMaxForWidth(groupWidthPx: number): number {
  return (COLLAPSED_RAIL_BOUNDARY_PX / groupWidthPx) * 100;
}
const COLLAPSED_LAYOUT_SHARE_MAX_PCT =
  collapsedShareMaxForWidth(BOTH_PANELS_MIN_WIDTH_PX); // module init — CONSTANT arg (621)

// …and at render, in peek mode — DYNAMIC arg:
isPeekMode && peekWidthPx > 0
  ? collapsedShareMaxForWidth(peekWidthPx)
  : COLLAPSED_LAYOUT_SHARE_MAX_PCT;
```

The SWC production minifier **constant-folds** the module-init call (`260 + 360 + 1 = 621` is compile-time), removing every init-time reference to the const. The dynamic render-time call can't be folded, so its body is kept — but the minifier, now seeing the const as unused outside the folded-away path, **drops the binding while leaving the reference dangling** (the un-mangled name in the error confirms it was treated as a free global).

That asymmetry is why it fails **only when opening a trace peek** (`?peek=…`), never on normal page load or the full-page trace view (which use the folded literal).

## What

- Inline the boundary as a **function-local** const, so there is no top-level binding for the minifier to drop. Byte-for-byte identical arithmetic — restores the safe shape the pre-#14716 code had.
- Swept `web/src` for the same shape (a small arithmetic helper reading a single-use module const, folded at a constant call but kept at a dynamic call elsewhere). This was the **only** instance: `buildFilterMappings` (dashboard) has no dynamic call site and returns non-foldable objects; `peekPanelStore` consts are literals / exported / multi-use.

## Linear

- None (regression from LFE-10601, #14716)

## Review focus

- **Behavior unchanged** — pure refactor of a numeric helper, no logic or UX change.
- The real gate is the **production build** (the minifier path this targets); dev never reproduced it.

<details>

- `turbo run lint` across all packages — passes (pre-commit hook).
- `TraceLayoutDesktop.tsx` typechecks clean.
- Correct by construction: the identifier no longer exists anywhere, so nothing can drop it or leave it dangling.

</details>

### DIFF
--- a/web/src/components/trace/components/_layout/TraceLayoutDesktop.tsx
+++ b/web/src/components/trace/components/_layout/TraceLayoutDesktop.tsx
@@ -116,10 +116,12 @@ const NOOP_LAYOUT_STORAGE = {
 // small-share) nav on a wide peek is mis-seeded as collapsed and its content
 // unmounts for a frame until `onResize` corrects it. Boundary = midpoint between
 // the 40px rail and the smallest open min (nav 260px), as a share of the width.
-const COLLAPSED_RAIL_BOUNDARY_PX =
-  (COLLAPSED_PANEL_PX + NAVIGATION_PANEL_MIN_PX) / 2;
+// Kept LOCAL (not a module const): as a single-use top-level const it tripped
+// the SWC prod minifier — the binding was dropped but a dangling reference kept,
+// throwing a `ReferenceError` on peek open.
 function collapsedShareMaxForWidth(groupWidthPx: number): number {
-  return (COLLAPSED_RAIL_BOUNDARY_PX / groupWidthPx) * 100;
+  const boundaryPx = (COLLAPSED_PANEL_PX + NAVIGATION_PANEL_MIN_PX) / 2;
+  return (boundaryPx / groupWidthPx) * 100;
 }
 // Full-page fallback (its container width isn't known at seed time): the boundary
 // as a share of the narrowest both-open width, matching the prior heuristic.


vasyl-stanislavchuk:
Closing — superseded by the full revert of #14716 on `main` (5a661a3c5, "revert: peek comfortable defaults — prod crash (LFE-10640)"). That revert removes the `COLLAPSED_RAIL_BOUNDARY_PX` const / `collapsedShareMaxForWidth` helper this PR patched and restores the pre-#14716 inline `COLLAPSED_LAYOUT_SHARE_MAX_PCT`, so the crash is already resolved on main and this targeted fix has nothing left to change.

**Note for re-landing #14716:** the prod crash (`ReferenceError: COLLAPSED_RAIL_BOUNDARY_PX is not defined`, peek only) came from a single-use, module-scoped const read only inside `collapsedShareMaxForWidth`, which was called both at module-init with a constant arg (SWC constant-folds it) and at render with a dynamic peek width. The minifier folded the init call, dropped the now-"unused" const binding, and left the dynamic call's reference dangling. When the feature is re-landed, compute the boundary as a **function-local** const (or otherwise avoid a single-use top-level const feeding a helper that has both a folded and a dynamic call site) so the minifier has no top-level binding to drop.

PR #14737, opened 2026-07-02
fix(events): reveal experiment environments when filtering by experiment (LFE-10644)

nkabardin:
## TL;DR

Filtering the v4 events table by an experiment (`experimentName` / `experimentId` / `experimentDatasetId`) could **never return a row**: experiment-enriched events always live in the `sdk-experiment` / `langfuse-prompt-experiment` environments, and the table's *implicit* hidden-environments default silently excludes exactly those. A positive experiment filter now lifts the two experiment environments out of the implicit exclusion. Plus a new `experiment-run` seeder scenario that makes the shape reproducible in one command.

## Why

[LFE-10644](https://linear.app/langfuse/issue/LFE-10644): an events-table view filtered by `name any of […] AND experimentName any of […]` returned 0 results where ~82 were expected — while the filter dropdown itself offered that experiment name as an option.

Root cause: the SDKs (JS ≥ 5, Python ≥ 3.9) stamp `environment: sdk-experiment` **together with** the `langfuse.experiment.*` attributes on every in-process span of an experiment run — the run root directly, children via context propagation ([langfuse-js `propagation.ts`](https://github.com/langfuse/langfuse-js/blob/main/packages/core/src/propagation.ts), langfuse-python `_client/propagation.py`). UI prompt experiments behave the same way in `langfuse-prompt-experiment`. So on the write side, *any* row that can match an experiment filter also carries a hidden-by-default environment: the combination "experiment filter + default environment exclusion" is unsatisfiable by construction. The filter-options API doesn't apply the exclusion, so the UI kept offering filter values that could never match.

## What

- `managedEnvironmentPolicy` now resolves the hidden-environments set **against the rest of the filter state**: a positive experiment filter (`any of` / `contains` / `is not null`, in id or display-name column form) removes the experiment environments from the implicit exclusion. Negative experiment filters (`none of`, `does not contain`, `is null`) and user-authored environment selections keep today's behavior.
- The lifted `none of` shape is recognized as system-shaped by the strip step, so the `updateFilter` round-trip (which feeds *effective* state back into `setFilterState`) doesn't leak it into persisted explicit state — and the full exclusion returns as soon as the experiment filter is removed.
- A system-shaped exclusion materialized into a shared URL re-canonicalizes the same way instead of staying unsatisfiable.
- New seed CLI scenario `experiment-run`: SDK-shaped experiment events (`experiment-item-run` root + app-named children) with an `--unenriched-children` variant for the cross-process shape.

## Result

The ticket's exact filter shape returns the expected rows (verified live: 40/40 seeded child spans). The default view is unchanged — all six internal environments stay hidden when no experiment filter is active.

**Confirmed against production data**: for an affected experiment run, a raw `events.all` probe without the environment filter returned all of the run's enriched rows (roots *and* children — in-process propagation had enriched everything), every one of them in `environment: sdk-experiment`. The implicit default exclusion was the only thing hiding them; there is no write-side enrichment gap for this class of runs.

<details>
<summary>Mechanism, verification, and test coverage</summary>

### Why fix the implicit default (and not the query or the write path)

- The compiled ClickHouse SQL for the ticket payload is structurally correct (`e."experiment_name" IN (…) AND e."environment" NOT IN (…)`) — the query builder is not at fault.
- The `environment none of [6]` clause in the failing request is not user intent: it is the sidebar's implicit default (`DEFAULT_SIDEBAR_HIDDEN_ENVIRONMENTS`), merged into every events query. The fix therefore belongs where the default is applied. Explicit user environment selections are never overridden.
- Enriching cross-process spans on the write side was previously explored and rejected (#14290): rewriting direct-written rows creates duplicate rows since reads don't uniformly deduplicate.

### Live verification (seeded local stack)

1. `pnpm run seed -- experiment-run --items 20` → 60 events (20 `experiment-item-run` roots + 40 app-named children, all enriched, env `sdk-experiment`).
2. Ticket-shaped URL (`name any of [4 app names]` + `experimentName any of [run]`): **0 rows before, 40 rows after**; the emitted `events.all` payload now carries `environment none of [langfuse-llm-as-a-judge, langfuse-code-eval, langfuse-natural-language-filter, langfuse-evaluation]`.
3. Fresh state without filters: payload carries the full 6-environment exclusion — default hiding intact.
4. Round-trip: toggling another facet while the experiment filter is active leaks no environment filter into the URL; clearing the experiment filter restores the full exclusion on the next request.

### Test coverage

`web/src/features/filters/lib/managedEnvironmentPolicy.clienttest.ts` (18 tests): lift on positive filters (id + display-name + text-filter forms), no lift on negative filters, explicit env filter precedence, strip of both system shapes (incl. after experiment-filter removal), URL-materialized re-canonicalization, user `any of` never touched, and a parity guard that `EXPERIMENT_FILTER_COLUMNS` stays in sync with the events-table experiment columns. All 125 existing filter-feature tests pass.

</details>


nkabardin:
Closing: this special-cases one scenario (experiment filters lifting the hidden-environment default). The underlying problem — facet filter options computed without the table's other applied filters — is general and will be addressed separately.

PR #14738, opened 2026-07-02
ci: fail builds on minifier-dropped bindings in the client bundle (LFE-10645)

nkabardin:
## TL;DR

`next build` can ship a bundle that throws `ReferenceError` in production only: the SWC minifier inside Turbopack sometimes **deletes a binding that live code still references** (that's what crashed the trace peek — LFE-10640, reverted in #14736, worked around in #14735). This PR adds a ~5-second post-build **client-bundle scanner** to CI that fails the build on any such dropped binding — and on Node-only globals leaking into browser chunks. It immediately proved itself by finding **23 real dropped bindings** in our current bundle, which this PR also fixes.

> **Reviewer note:** this PR's first CI run (at the scanner-only commit) is *intentionally red* — the gate lands first so it demonstrably catches the pre-existing miscompilations; the fix commit turns it green.

## Why

Opening a trace peek in production threw `ReferenceError: COLLAPSED_RAIL_BOUNDARY_PX is not defined`. The root cause was not our code: a documented, still-unfixed SWC minifier soundness bug (byte-identical broken output on every `@swc/core` from 1.3.107, Jan 2024, through 1.15.43, current latest) deletes a `const` when a rare five-condition shape occurs — the build succeeds, `next dev` is unminified, so nothing before real production traffic could catch it.

A source-level lint cannot express this (two of the five trigger conditions are minifier-internal heuristics), and it could never see dependencies. Scanning the **emitted assets** can: a deleted binding always leaves an identifier that is referenced but declared nowhere, and scope analysis finds every one of them.

## What

- `scripts/scan-client-bundle.mjs` — parses every emitted client asset under `web/.next/static` (espree + eslint-scope + globals, the PR's only new root devDependencies and the entire `pnpm-lock.yaml` diff) and reports identifiers referenced but never declared. Exempt: `typeof x` probes, references structurally gated by a `typeof` guard in the same ternary/logical/if (scoped to the guarded subtree, never the whole file), and a short allowlist where every entry carries a reason. Node globals are deliberately **not** exempt — bare `require`/`process`/`Buffer` in a browser chunk is a guaranteed crash the gate must catch. Exit 1 on findings; exit 2 on anything that would otherwise silently degrade the gate. ~5s for 613 files. Locally: `pnpm run scan:client-bundle`.
- Wired into the `e2e-tests` job right after the prod build.
- **Fix for the findings it caught:** all 23 sat inside `prettier/plugins/typescript`, which embeds a copy of the TypeScript compiler that the minifier miscompiles (e.g. `2&e?nullNodeConverters:createNodeConverters(y)` with the declaration of `nullNodeConverters` deleted — the literal LFE-10640 ternary shape). Each was a latent `ReferenceError` behind the eval-template **Format** action. The editor's format path now uses Prettier's **babel-ts** parser, which covers the same TypeScript surface without embedding the compiler (eval templates ban imports/exports anyway). The directly-imported `typescript` chunk used for live editor diagnostics scanned clean and stays unchanged.
- **Docs:** one Verification bullet in `.agents/AGENTS.md` (the gate + where the canonical fix lives) and one Web Convention in `web/AGENTS.md` (never client-import `prettier/plugins/typescript`; use `babel-ts`).

## Result

- The exact LFE-10640 bundle fails the scan (verified against a rebuild of the pre-revert state: `COLLAPSED_RAIL_BOUNDARY_PX` flagged in 8 chunks, declared in none); the fixed bundle passes.
- 23 latent `ReferenceError` landmines removed from the production bundle.
- Any future minifier-dropped binding — first-party or vendored — and any Node/CJS leakage into client code fails CI at PR time with the identifier and location named, instead of shipping to production.

<details>
<summary>Mechanism, repro, scanner design, and verification detail</summary>

### The five-condition shape (all required; verified by flipping each)

1. A binding whose initializer needs evaluation (`const A = 40 + N` — literal or plain-alias initializers are compiled correctly);
2. exactly one function reads it (two readers → correct output);
3. that reader is small enough to inline (padded body → correct);
4. one call of the reader is constant-foldable and its result stored in a non-exported binding (unused or exported → correct);
5. the folded result is referenced by the same surviving function that receives the inlined body (no co-reference → correct).

The constant-folder erases the only counted read, the inliner copies the reader's body (with the reference) into the live call site, and dead-binding elimination then deletes the declaration using stale usage counts. Minimal repro (breaks identically on `@swc/core` 1.3.107 → 1.15.43):

```js
const N = 260;
const A = 40 + N;
function f(w) { return A / w; }
const S = f(621);
export function dyn(w) { return f(w) + S; }
// minifies to: let n=300/621;export function dyn(t){return A/t+n}
// dyn(100) → ReferenceError: A is not defined
```

Related upstream: swc-project/swc#2934 (known-issues meta lists this family), closed-unmerged swc-project/swc#9107, and the 2022 twin vercel/next.js#39538.

### Scanner design notes

- Detection core: `eslint-scope`'s `globalScope.through` gives exactly the references no scope in the file resolves. Known globals = `globals.browser + worker + builtin`, **minus** Chrome-only names (`ai`, `model`, `scheduler`, …) that would mask a Firefox/Safari-only crash, and **without** `globals.node` — Node/CJS leakage into a client chunk must fire, not pass. (The tightened set immediately surfaced a real latent case: Next.js's own scheduler ships two unguarded `setImmediate` calls in a client chunk — upstream, process-gated at runtime, now the one precisely-commented allowlist entry for it.)
- Guard awareness: `"function"==typeof require?require:…` and friends are execution-gated probes, exempted structurally and only inside the guarded subtree — so the same name dropped elsewhere in the file still fires. This keeps the allowlist down to genuinely ambient names.
- Fail-closed on every silent-degradation path: a parse failure of any file, zero JS files at the expected path (build-output restructure), and a non-directory argument all exit non-zero — and a parse failure still prints the findings from the files that did parse. The walk is recursive with deterministic ordering, so subdirectory-emitted chunks can't silently narrow the gate.
- Honest limits (in the script header): the guarantee covers the *ReferenceError manifestation* of the dropped-binding class; a dropped binding whose name collides with a real window property (`status`, `length`, …) degrades silently, and wrong-value folds need runtime exercise. CI scans the e2e job's build as a proxy — release images minify the same sources separately with different inlined `NEXT_PUBLIC_*` constants.

### Verification

- Scanner red on the reconstructed LFE-10640 bundle and on the pre-fix bundle (23 findings, all `prettier/plugins/typescript`); clean (613 files, 0 findings, ~4.5s) after the babel-ts swap.
- 11-case fixture suite: dropped binding fires with `file:line:column`; bare `require`/`module`/`process`/`Buffer` fire; ternary/logical/if typeof-guard idioms stay clean; an unguarded Node-only name fires; a top-level-await chunk parses cleanly (module-first); `regeneratorRuntime` self-registration write is exempt while a read fires; Chrome-only names fire; a parse failure prints surviving findings and exits 2; a file argument and an empty directory exit 2.
- Eval-template editor exercised against a production build (`next build` + `next start`): live type diagnostics flag an injected type error; Format reformats `const   ok    =     1   +    2;` → `const ok = 1 + 2;`.
- `code-eval-template-validation.clienttest.ts` 16/16 (exercises real Prettier through the new parser); web lint + typecheck green.

</details>

Analysis and full evidence log: LFE-10645. Seed crash: LFE-10640 · #14736 (revert) · #14735 (workaround).


nkabardin:
@claude review

PR #14771, opened 2026-07-03
feat(search-bar): suggest observed metadata keys with types (LFE-10666)

nkabardin:
## TL;DR

Typing `metadata.` in the filter search bar now suggests real metadata keys with their observed types (`metadata.hej` · `number`), and `metadata.<key>:` suggests the first few observed values — built client-side from rows the user has already loaded and persisted per project in localStorage. Zero new API surface.

## Why

Metadata filtering is hard: the API does not enumerate metadata keys, so the bar's `metadata.` stage completed from free typing only — the user had to know their keys (and values) by heart. Backend metadata-structure analysis is deferred, so this is the low-hanging frontend win: the table already loads row metadata; analyze it opportunistically and remember what was seen.

## What

- **`lib/metadata-paths.ts`** (pure): samples the first 30 visible rows per fetch (same sampling as the AI-context path), records their **top-level metadata keys** with the observed JSON value type plus the first few distinct scalar values. A key seen with more than one type becomes `mixed` and drops its type hint; null observations register the key without demoting a known type.
- **`store/observedMetadataStore.ts`**: zustand `persist` store, one global `Record<projectId, …>` map (the `globalDateRangeStore` shape, so project switches can't clobber each other). Bounded on every axis: ≤ 200 keys/project, ≤ 5 values/key, ≤ 1024 values/project (hard backstop; the effective bound is 200 × 5), value length ≤ 60, ≤ 20 projects (LRU eviction), `version` key that resets the cache on schema change; a no-change merge skips the localStorage write entirely.
- **Injection**: keys merge into the observed-options map under `metadata`, values under `metadata.<key>` — exactly where the completion planner already looked (`keyPathOptions` + the metadata value stage), so ranking, quoting of grammar-char keys, and pick behavior are all inherited. The type rides a new optional `type?` on `ObservedValue`, shown as the option detail.

**Design note — top-level keys only, never flattened nested dot-paths.** Metadata is stored as a flat `Map(String, String)`: nested object values are JSON-encoded strings under their top-level key, and `StringObjectFilter` matches the *literal* top-level key ("we can only filter on the first level"; the metadata view's filter shortcut resolves the top-level key for the same reason). A flattened `metadata.scope.name` suggestion would build a filter that matches nothing. Object-valued keys are suggested with type `object` (their content matches via `contains`); dotted OTel-attribute keys (`gen_ai.request.model`) are literal keys and suggest/filter exactly like nested paths would. True nested filtering stays a backend follow-up.

**Design note — values are skip-don't-truncate, scalars only.** Only string/number/boolean leaves are collected (object/array stored forms could not round-trip into a matching `=` filter), and a value longer than the cap is dropped entirely — a truncated suggestion would insert a filter that confidently matches nothing. Eviction is first-N-distinct (no counts): frequency tracking would churn localStorage on every fetch.

Types are display-only: metadata filters lower to `stringObject` regardless (numeric metadata comparisons stay rejected). Accepted caveat: metadata the user has never loaded is never suggested.

## Result

Typing `metadata.` on the v4 events tables opens an "Observed keys" section with typed suggestions; `metadata.<key>:` opens "Observed values"; prefix typing ranks and arms Enter; accepted picks build working `stringObject` filters; the map survives reloads and is strictly project-scoped.

<details>

- Unit: 27 new tests (`metadata-paths.clienttest.ts`, `observedMetadataStore.clienttest.ts`) covering type merging (mixed-absorbing, null upgrades), value collection (distinct, per-key cap, scalar-only, skip-don't-truncate), caps (keys/values/projects/LRU/key-length, total-value backstop against drifted persisted state), `__proto__`/empty-key hardening, and the merge-into-observed helper; 1 new completions test for the type detail. Full search-bar suite: 220 passed.
- Browser (localhost:3034, seeded demo project): key dropdown shows `metadata.langgraph_node · string`, `metadata.seed · number`, etc.; loading trace-tree rows grows the map with `scope`/`attributes` (`object`) and dotted `node.depth` (`number`); `metadata.langgraph_node:` offers `critic`/`generator`/`retriever`/`planner`; picking `retriever` commits `?filter=metadata;stringObject;langgraph_node;=;retriever` and returns exactly the matching rows; `metadata.node.depth:3` (dotted key) returns matching rows; reload keeps all suggestions; the persist v1→v2 migrate resets and repopulates cleanly; a second project shows none (no cross-bleed).
- Perf: analysis runs once per fetch on ≤ 30 rows with a 200 KB per-row parse guard; idempotent merges skip the store write, so re-renders and refetch echoes settle without localStorage churn.
- `pnpm --filter web run typecheck`, `pnpm run lint` clean.

</details>

Related: search-bar README owns the full contract (new "Metadata key suggestions" section documents this seam).


nkabardin:
@claude review

PR #14789, opened 2026-07-04
fix(prompts): namespace cache key by version/label

### LINKED ISSUE #14788: bug: prompt cache key collides version numbers with numeric labels, returning the wrong prompt
### Describe the bug

The server side prompt cache builds its Redis key without distinguishing a version lookup from a label lookup, so a version number and a numeric label with the same text share one cache entry and return each other's content.

In packages/shared/src/server/services/PromptService/index.ts, getCacheKey returns:

  `${prefix}:${params.version ?? params.label}`

where prefix is `prompt:{projectId}:{epoch}:{promptName}`. version is a number and label is a user provided string. Labels are allowed to be purely numeric (PROMPT_LABEL_REGEX = /^[a-z0-9_\-.]+$/ in packages/shared/src/features/prompts/constants.ts). So a request for version 3 and a request for label "3" produce the identical key `prompt:{projectId}:{epoch}:{name}:3` and share one cached value. Whichever request populates the key first wins, and the other request gets a cache hit with the wrong prompt version.

This is active in default deployments. LANGFUSE_CACHE_PROMPT_ENABLED defaults to "true" (packages/shared/src/env.ts) and prompt caching was turned on by default, and Redis is a standard part of a v3 deployment. The resolved prompt path (resolve defaults to true) is the one that uses the cache.

What I expected: fetching a prompt by version N and fetching it by a label whose text happens to be "N" should each return their own prompt version. A numeric label should never be able to return a version pinned prompt, or the reverse.

### Steps to reproduce

1. Run Langfuse with Redis and default prompt caching (LANGFUSE_CACHE_PROMPT_ENABLED unset or "true").
2. Create a prompt named "greeting":
   - version 3 with content A
   - version 7 with content B, and add the custom label "3" to version 7 (a numeric label is accepted).
3. Fetch by version first so it populates the cache:
   GET /api/public/v2/prompts/greeting?version=3
   Returns content A and caches it under key ...:greeting:3
4. Fetch by the numeric label:
   GET /api/public/v2/prompts/greeting?label=3
   Expected: content B (version 7).
   Actual: content A (version 3), served from the colliding cache entry.
5. The failure is symmetric. If you request label "3" first, a later version=3 request returns version 7's content instead.

Root cause is the untyped `version ?? label` concatenation in getCacheKey. A fix is to prefix the selector with its type, for example `version:{n}` versus `label:{s}`, so the two namespaces cannot collide. Old format keys simply miss and repopulate, then expire by TTL.

Self-hosted

_No response_

_No response_

### Additional information

- Impact: prompt management is a core feature and teams pin prompts by version or by label for production. This can silently serve a different prompt version than requested, which changes model behavior.
- Scope: affects both self hosted and Langfuse Cloud, since the code path and the default on caching are shared.
- Trigger condition: a prompt that has a version number N and also a custom label whose text is "N" on some version. The default "production" label does not collide because it is not numeric.
- Suggested fix: discriminate the cache key by lookup type in getCacheKey (version: vs label:). I am happy to open a PR with the fix and a PromptService regression test that asserts version=N and label="N" return distinct content from cache.
- No screenshot needed; this is reproducible purely from two API calls.

Yes

G26karthik:
## What does this PR do?

Fetching a prompt by version and by a numeric label with the same text returns the wrong prompt because both map to the same cache entry.

Fixes #14788.

### The problem

`PromptService.getCacheKey` currently builds the Redis key as:

```ts
`${prefix}:${params.version ?? params.label}`
```

`version` is a number and `label` is a string, but they are concatenated without any type information. Since labels are allowed to be purely numeric (`PROMPT_LABEL_REGEX = /^[a-z0-9_\-.]+$/`), a lookup for `version=3` and a lookup for `label="3"` both produce the same cache key:

```text
prompt:{projectId}:{epoch}:{name}:3
```

Whichever request populates the cache first wins, and the other request receives a cache hit with the wrong prompt content.

Prompt caching is enabled by default (`LANGFUSE_CACHE_PROMPT_ENABLED` defaults to `"true"`), so this issue is reachable in a standard deployment rather than only in an opt-in configuration.

### The fix

Namespace the cache key by lookup type so version- and label-based lookups can never collide:

```ts
const selector =
  typeof params.version === "number"
    ? `version:${params.version}`
    : `label:${params.label}`;

return `${prefix}:${selector}`;
```

Keys now look like:

```text
...:{name}:version:3
...:{name}:label:3
```

This change is backwards compatible. Existing cache entries in the old format will simply miss on the next lookup, be repopulated using the new key format, and the stale entries will expire naturally via TTL. No cache flush or migration is required. This also aligns with the epoch-based cache invalidation introduced in #12363.

### Testing

Added a regression test in `promptCache.servertest.ts` that verifies:

- A lookup by `version=N` and a lookup by `label="N"` generate distinct cache keys.
- The two lookups do not reuse each other's cached entry.
- Updated the existing cache key assertion to match the new key format.

This PR fixes a cache key collision in `PromptService` where looking up a prompt by `version=N` and by `label="N"` (a numeric string) both produced the same Redis key, causing one to return the wrong cached value. The fix namespaces the cache key by lookup type (`version:N` vs `label:N`), making the keys structurally distinct.

- **Core fix** (`packages/shared/src/server/services/PromptService/index.ts`): `getCacheKey` now branches on `typeof params.version === "number"` to emit either `version:{N}` or `label:{val}`, eliminating the collision. Old-format keys expire naturally via TTL with no migration required.
- **Tests** (`web/src/__tests__/server/promptCache.servertest.ts`): Two new regression tests verify that version and label lookups produce distinct Redis keys and that a label-based lookup cannot reuse a version-based cache entry. The existing cache-key assertion is updated to match the new format.
</details>

Safe to merge — the change is a surgical two-line fix to a private helper with no side effects on callers, and old cache entries expire via existing TTL without any required manual intervention.

The discriminated union on `PromptParams` guarantees the else branch always has a string `label`, the key format change is backwards-compatible (stale entries expire naturally), and both new regression tests directly exercise the previously-colliding code path. No correctness concerns were found.

</details>

```mermaid
flowchart TD
    A[getPrompt called] --> B{cacheEnabled?}
    B -- No --> DB[(Database lookup)]
    B -- Yes --> C[getCacheKey]
    C --> D{typeof version === 'number'?}
    D -- Yes --> E["key = prefix:version:{N}"]
    D -- No --> F["key = prefix:label:{val}"]
    E --> G[Redis GET]
    F --> G
    G --> H{Cache hit?}
    H -- Yes --> I[Return cached prompt]
    H -- No --> DB
    DB --> J{prompt found?}
    J -- No --> K[Return null]
    J -- Yes --> L[getCacheKey again]
    L --> M{typeof version === 'number'?}
    M -- Yes --> N["write key = prefix:version:{N}"]
    M -- No --> O["write key = prefix:label:{val}"]
    N --> P[Redis SET with TTL]
    O --> P
    P --> I
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[getPrompt called] --> B{cacheEnabled?}
    B -- No --> DB[(Database lookup)]
    B -- Yes --> C[getCacheKey]
    C --> D{typeof version === 'number'?}
    D -- Yes --> E["key = prefix:version:{N}"]
    D -- No --> F["key = prefix:label:{val}"]
    E --> G[Redis GET]
    F --> G
    G --> H{Cache hit?}
    H -- Yes --> I[Return cached prompt]
    H -- No --> DB
    DB --> J{prompt found?}
    J -- No --> K[Return null]
    J -- Yes --> L[getCacheKey again]
    L --> M{typeof version === 'number'?}
    M -- Yes --> N["write key = prefix:version:{N}"]
    M -- No --> O["write key = prefix:label:{val}"]
    N --> P[Redis SET with TTL]
    O --> P
    P --> I
```

</a>
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(prompts): namespace cache key by ver..."](https://github.com/langfuse/langfuse/commit/2750af76b52ee6703d06da8044e87bd5e3f879ab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41840009)</sub>


hassiebp:
Thanks for raising this, merged in https://github.com/langfuse/langfuse/pull/15040 due to the failing test

PR #14811, opened 2026-07-06
fix(playground): scroll to and focus new message on Add message (LFE-6864)

nkabardin:
## TL;DR
In the Playground, clicking **Add message** with a long system prompt did nothing visible — the new input was appended far below the fold and the view didn't move. Now, after adding a message (any role, plus Placeholder), the new input **scrolls into view and is focused**, so you can type immediately. Because `ChatMessages` is shared, this also fixes the **New Prompt (chat) editor**.

Fixes LFE-6864.

## Why
There was already a "scroll to bottom on new message" effect in `ChatMessages`, but it was effectively a no-op. Two root causes:

1. **Wrong scroll strategy / timing.** It set `scrollAreaRef.scrollTop = scrollElement.scrollHeight` synchronously in a `useEffect`. The freshly-appended row hosts a CodeMirror editor that mounts *asynchronously*, so `scrollHeight` was read before the new row had laid out — and in the Prompt editor the inner `overflow-auto` div isn't even the element that scrolls (an outer container does), so writing its `scrollTop` did nothing at all.
2. **No focus.** The add handlers discarded the new message returned by `addMessage`, so there was no way to target the new row for focus.

## What
- `ChatMessages` keeps a small registry of each message's row + editor refs (populated by `ChatMessageComponent`, reusing the refs it already holds).
- After `addMessage` returns the new message, a `scrollToMessage(id)` helper `scrollIntoView({ block: "nearest" })`s the new row and focuses its editor. It runs on `requestAnimationFrame` and **retries over a short bounded window** until the editor is mounted and actually holds focus — robust against the async editor mount without guessing a fixed delay.
- The role dropdown additionally prevents Radix's close-time focus restoration (`onCloseAutoFocus`) so focus can land on the new editor instead of snapping back to the trigger.

## Result
Add via the **Message** button, the **role dropdown** (User/Assistant/System/Developer/Tool), and **Placeholder** all scroll the new input into view and focus it. Verified live in the browser for every path; the shared New Prompt (chat) editor behaves the same.

<details>
<summary>Verification detail</summary>

Driven in a real browser against a long system prompt:

- **Message button:** new editor focused (active element inside the new `.cm-editor`) and within the viewport.
- **Dropdown roles (System, Assistant, …):** new editor focused + in view. Without the `onCloseAutoFocus` guard + focus retry, focus fell back to the popover then `<body>` — now it lands on the editor.
- **Placeholder:** new editor focused + in view.
- **New Prompt (chat) editor:** typing after Add lands the caret in the new message's editor, which is scrolled into view.

Lint / typecheck / prettier all clean (pre-commit `turbo run lint` across all packages passed).

</details>

This PR fixes the \"Add message\" scroll-and-focus behavior in the Playground's `ChatMessages` component. The previous approach was a no-op because it set `scrollTop` synchronously before CodeMirror had laid out, and the outer container did not scroll anyway.

- Replaces the broken `scrollTop` effect with a ref registry (`rowRefsById` Map) populated by each `ChatMessageComponent` via a new `registerRow` prop; a bounded `requestAnimationFrame` retry loop in `scrollToMessage` handles the asynchronous CodeMirror mount and Radix focus teardown competition.
- Captures the return value of `addMessage` in all three add paths (regular button, role dropdown, placeholder) and calls `scrollToMessage(newMessage.id)` so the new row scrolls into view and its editor receives focus.
- Adds `onCloseAutoFocus={(e) => e.preventDefault()}` to the role dropdown to prevent Radix from stealing focus back to the trigger after an item is selected — however this prevention fires on all closes (including Escape/click-outside), dropping keyboard focus to `<body>`.
</details>

The core scroll-and-focus fix works correctly, but the blanket onCloseAutoFocus suppression introduces a keyboard accessibility regression on the role dropdown.

The retry-loop and ref-registry approach is well-designed and resolves the stated bug. However, onCloseAutoFocus fires unconditionally: a keyboard user who opens the dropdown and presses Escape without selecting a role now loses focus entirely instead of returning to the trigger button.

web/src/components/ChatMessages/index.tsx — specifically the DropdownMenuContent onCloseAutoFocus handler.
</details>

```mermaid
sequenceDiagram
    participant User
    participant AddMessageButton
    participant ChatMessages
    participant ChatMessageComponent
    participant CodeMirror

    User->>AddMessageButton: clicks Message or role dropdown item
    AddMessageButton->>ChatMessages: addMessage returns newMessage
    AddMessageButton->>ChatMessages: scrollToMessage(newMessage.id)
    Note over ChatMessages: requestAnimationFrame(attempt)
    ChatMessages->>ChatMessageComponent: render with registerRow
    ChatMessageComponent->>ChatMessages: registerRow useEffect editorRef null
    ChatMessageComponent->>CodeMirror: mount editor
    CodeMirror-->>ChatMessageComponent: onEditorMount
    ChatMessageComponent->>ChatMessages: registerRow editorRef populated
    loop rAF retry up to 21 frames
        ChatMessages->>ChatMessages: check editorRef.current.view
        ChatMessages->>ChatMessageComponent: scrollIntoView nearest
        ChatMessages->>CodeMirror: view.focus
        alt view.hasFocus
            ChatMessages->>ChatMessages: return done
        end
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant User
    participant AddMessageButton
    participant ChatMessages
    participant ChatMessageComponent
    participant CodeMirror

    User->>AddMessageButton: clicks Message or role dropdown item
    AddMessageButton->>ChatMessages: addMessage returns newMessage
    AddMessageButton->>ChatMessages: scrollToMessage(newMessage.id)
    Note over ChatMessages: requestAnimationFrame(attempt)
    ChatMessages->>ChatMessageComponent: render with registerRow
    ChatMessageComponent->>ChatMessages: registerRow useEffect editorRef null
    ChatMessageComponent->>CodeMirror: mount editor
    CodeMirror-->>ChatMessageComponent: onEditorMount
    ChatMessageComponent->>ChatMessages: registerRow editorRef populated
    loop rAF retry up to 21 frames
        ChatMessages->>ChatMessages: check editorRef.current.view
        ChatMessages->>ChatMessageComponent: scrollIntoView nearest
        ChatMessages->>CodeMirror: view.focus
        alt view.hasFocus
            ChatMessages->>ChatMessages: return done
        end
    end
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/components/ChatMessages/index.tsx:277-283
**`onCloseAutoFocus` blanket prevention breaks keyboard dismiss**

`onCloseAutoFocus` fires on every close — including when the user presses Escape or clicks outside without selecting an item. In those cases Radix's default focus-return to the trigger is prevented, so focus is silently dropped to `<body>`. A keyboard user who opens the dropdown and reconsiders (Escape) now has no focused element and must Tab through the whole page to get back to the button.

The fix would be to only suppress the default when an add-message action actually ran. One approach: track a flag (`didAddMessage`) that is set in the `onClick` handler and cleared in `onCloseAutoFocus`, then call `event.preventDefault()` only when that flag is true.

### Issue 2 of 2
web/src/components/ChatMessages/ChatMessageComponent.tsx:260-266
**Initial `registerRow` call passes refs before the editor has mounted**

The `useEffect` runs after the first render and registers `{ rowRef, editorRef }` immediately, but at that point `editorRef.current` is still `null` because CodeMirror hasn't mounted yet. The `handleEditorMount` callback then re-registers with the populated ref. The retry loop in `scrollToMessage` compensates for this, but the initial registration is functionally a no-op and only adds noise to the map. Consider registering only from `handleEditorMount` (and keeping the cleanup in the `useEffect`) to make the flow cleaner and avoid relying on the retry to cover the window where `editorRef.current === null`.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(playground): scroll to and focus new..."](https://github.com/langfuse/langfuse/commit/2705c9e5086252a55da29717959f78d08c4d969b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42067769)</sub>

### DIFF
--- a/web/src/components/ChatMessages/ChatMessageComponent.tsx
+++ b/web/src/components/ChatMessages/ChatMessageComponent.tsx
@@ -38,6 +38,11 @@ import {
   useOptionalMessageSearchPageId,
 } from "./MessageSearch";
 
+export type MessageRowRefs = {
+  rowRef: RefObject<HTMLDivElement | null>;
+  editorRef: RefObject<ReactCodeMirrorRef | null>;
+};
+
 type ChatMessageProps = Pick<
   MessagesContext,
   | "deleteMessage"
@@ -45,7 +50,13 @@ type ChatMessageProps = Pick<
   | "availableRoles"
   | "toolCallIds"
   | "replaceMessage"
-> & { message: ChatMessageWithId; index: number };
+> & {
+  message: ChatMessageWithId;
+  index: number;
+  // Lets the parent ChatMessages track this row's DOM + editor refs so it can
+  // scroll to and focus a freshly added message (LFE-6864).
+  registerRow?: (id: string, refs: MessageRowRefs | null) => void;
+};
 
 const ROLES: ChatMessageRole[] = [
   ChatMessageRole.User,
@@ -94,6 +105,7 @@ export const ChatMessageComponent: React.FC<ChatMessageProps> = ({
   availableRoles,
   index: _index,
   toolCallIds,
+  registerRow,
 }) => {
   const [roleIndex, setRoleIndex] = useState(1);
   const playgroundContext = useOptionalPlaygroundContext();
@@ -243,7 +255,19 @@ export const ChatMessageComponent: React.FC<ChatMessageProps> = ({
     unregisterMessageTarget,
   ]);
 
+  // Keep the parent ChatMessages registry in sync so it can scroll to and
+  // focus this row when it is the newly added message (LFE-6864).
+  useEffect(() => {
+    registerRow?.(message.id, { rowRef, editorRef });
+
+    return () => {
+      registerRow?.(message.id, null);
+    };
+  }, [message.id, registerRow]);
+
   const handleEditorMount = useCallback(() => {
+    registerRow?.(message.id, { rowRef, editorRef });
+
     if (!pageId || !registerMessageTarget) {
       return;
     }
@@ -252,7 +276,7 @@ export const ChatMessageComponent: React.FC<ChatMessageProps> = ({
       rowRef,
       editorRef,
     });
-  }, [message.id, pageId, registerMessageTarget]);
+  }, [message.id, pageId, registerMessageTarget, registerRow]);
 
   return (
     <Card
--- a/web/src/components/ChatMessages/ChatMessageComponent.tsx
+++ b/web/src/components/ChatMessages/ChatMessageComponent.tsx
@@ -256,10 +256,12 @@ export const ChatMessageComponent: React.FC<ChatMessageProps> = ({
   ]);
 
   // Keep the parent ChatMessages registry in sync so it can scroll to and
-  // focus this row when it is the newly added message (LFE-6864).
+  // focus this row when it is the newly added message (LFE-6864). The actual
+  // registration happens in handleEditorMount once CodeMirror has mounted and
+  // editorRef.current is populated — registering here on first render would
+  // store a null editor (CodeMirror mounts async). This effect only handles
+  // unregistering the row when it unmounts.
   useEffect(() => {
-    registerRow?.(message.id, { rowRef, editorRef });
-
     return () => {
       registerRow?.(message.id, null);
     };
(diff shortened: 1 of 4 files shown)


nkabardin:
@claude review

PR #14855, opened 2026-07-07
fix(trace): comfortable info-favoring default + persistent split for full-page trace view (LFE-10729)

nkabardin:
## TL;DR

The standalone full-page trace view now opens with the info panel as the larger panel (tree clamped to a comfortable 340–460px band) and remembers a dragged split across reloads and tabs — the same computed-default + localStorage machinery the peek view got in LFE-10601 (#14716, #14766). The nav-header's responsive breakpoints are retuned so the Tree/Timeline switcher no longer clips at the new default widths.

## Why

LFE-10601 fixed the peek view, but the full-page trace page kept its old defaults: a fixed 60/40 share (tree wider than the info panel, ballooning on wide monitors) and per-tab `sessionStorage` (a dragged split silently reset in every new tab). Users read the info panel — the tree is the index — so the default had it backwards, and the "my layout keeps resetting" complaints were the sessionStorage scoping.

The new tree default (340–460px) also exposed a stale set of nav-header container-query breakpoints: they predate the playback transport in that row, so at 380–434px everything stayed inline but no longer fit, clipping the Tree/Timeline switcher — the old 60% default was simply too wide to ever land there.

## What

Layout (`TraceLayoutDesktop.tsx`):

- Generalized the peek's `computePeekNavPercent` → `computeNavPercent` and drive the full-page default from the viewport width: the tree gets a clamped comfortable band (340–460px), the info panel gets everything else.
- Full-page split now persists in `localStorage` (was `sessionStorage`), matching the peek.
- Bumped the full-page group id `trace-layout-v2` → `trace-layout-v3` so stale saved 60/40 layouts don't mask the new default.
- Made the collapse-seed threshold width-aware for the full-page too (previously peek-only), so a persisted open-but-small-share tree on a wide monitor isn't mis-seeded as collapsed on restore.

Nav header (`TracePanelNavigationHeader.tsx`):

- Retuned the container-query breakpoints to the row's measured content minimums: switcher labels hidden < 440px (row needs ~434px with them), minor tools fold into "…" < 360px (icons-only needs ~352px), and the search input's min-width narrows < 300px so the row still fits at the 260px panel minimum. Arithmetic documented inline so the next toolbar addition retunes instead of clipping.

Peek behavior and annotation mode (nav starts collapsed, no layout saving) are unchanged.

## Result

Verified in the browser at 2560/1920/1440/1100px viewports:

| Viewport | Tree (default) | Info (default) |
|---|---|---|
| 2560px | 427px | 1948px |
| 1440px | 401px | 854px |
| 1100px | 383px | 532px |

Drag → reload → new tab all restore the chosen split from localStorage; the peek still opens at its LFE-10601 computed default under its own `trace-layout-peek-v1` key; annotation queues still start with the tree collapsed and persist nothing. Header verified at tree widths 280/340/396/420/480px: no overflow in any band — labels ≥440px, icons-only 360–440px, tools folded <360px.

<details>
<summary>Engineering detail</summary>

- The default is expressed as a computed *percentage* `defaultLayout` (not a px `defaultSize`) for the same reason as the peek: the library resolves px defaults against whatever group width it first measures, which is non-deterministic mid-mount (see #14716).
- The full-page container width is approximated by `window.innerWidth` (sidebar offset ignored) — the tree is width-capped, so the overestimate only makes the tree slightly narrower, still inside its comfortable band. Same trade-off the peek makes with its drawer width.
- The `navigationDefaultSize`/`detailDefaultSize` context plumbing is gone: both views are driven by the computed layout, so the panel `defaultSize` is a static SSR-only fallback (nav 40%, detail fills the remainder).
- Collapse-state seeding on restore uses the width-aware boundary for both views now; the fixed `COLLAPSED_LAYOUT_SHARE_MAX_PCT` remains only as the SSR fallback. Without this, a 400px open tree saved on a 2560px monitor (≈16% share) would be mis-seeded as collapsed on reload.
- The nav-header breakpoints are pure CSS container queries (no JS measurement); all elements in the row are fixed-size icon buttons or min-width-bounded inputs, so the content minimums are font-independent and the thresholds deterministic across browsers.
- Lint (`Tasks: 8 successful, 8 total`) and web typecheck pass.

</details>

### DIFF
--- a/web/src/components/trace/components/_layout/TraceLayoutDesktop.tsx
+++ b/web/src/components/trace/components/_layout/TraceLayoutDesktop.tsx
@@ -23,51 +23,51 @@ import { useViewPreferences } from "@/src/components/trace/contexts/ViewPreferen
 import { useSelection } from "@/src/components/trace/contexts/SelectionContext";
 import { resolveEffectiveWidthFraction } from "@/src/components/table/peek/store/peekPanelStore";
 
-// v2: the default split gives the trace (tree/timeline) the central space with a
-// slimmer detail panel. Bumped so a stale saved layout doesn't mask it. Used by
-// the full-page trace view; the peek uses its own id/storage/default (below).
-const RESIZABLE_PANEL_GROUP_ID = "trace-layout-v2";
-// Peek gets a distinct group so its computed-percentage default and localStorage
-// persistence don't leak into the full-page view (which stays share-based, per
-// tab). Same panels, different layout scope (LFE-10601).
+// Full-page trace view's layout group. v3: the full-page view gets the same
+// computed info-favoring default + localStorage persistence as the peek
+// (LFE-10729); bumped from v2 so stale saved layouts from the old 60/40-share
+// era don't mask the new default.
+const RESIZABLE_PANEL_GROUP_ID = "trace-layout-v3";
+// Peek keeps a distinct group id so the two surfaces remember their sizes
+// independently — a split dragged in a ~50vw peek is rarely right for a
+// full-page trace, and vice versa. Same panels, different layout scope
+// (LFE-10601).
 const PEEK_RESIZABLE_PANEL_GROUP_ID = "trace-layout-peek-v1";
 const RESIZABLE_PANEL_HANDLE_ID = "trace-layout-handle";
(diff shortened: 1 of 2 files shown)


nkabardin:
@claude review

PR #14880, opened 2026-07-08
feat(blob-export): make Parquet the default export format for all projects (LFE-10748)

niklassemmler:
## What does this PR do?

Makes the Parquet blob-storage export generally available for all projects.

- **Removes the whitelist.** Deletes the in-code project allowlist (`parquetFileType.ts` + its client test) that gated the first-class `PARQUET` fileType, and the tRPC/REST enforcement that referenced it.
- **Parquet is the new default.** Default file type switched from JSONL/CSV to `PARQUET` at the application layer: the zod form schema default, the upsert service CREATE fallback, and the settings-form defaults. No DB schema change — the column default is never hit by application code (every create supplies an explicit `fileType`), so changing it would be inert and only add deploy/rollback surface.
- **Selectable via the public API.** `PARQUET` is added to the create/update request `fileType` enum (zod + Fern + regenerated `web/public/generated/api/openapi.yml`), and the REST handler's Parquet downgrade guard is removed so the format can be freely set and changed via the API.

The settings form still reflects Parquet's real behavior (locks gzip off, trims the enrichment-only model-price columns from the field-group descriptions) — only the whitelist gating was removed.

### Fields not available with Parquet

Parquet exports stream raw bytes from the query engine and skip the JS enrichment step, so the observation columns `input_price`, `output_price`, and `total_price` are omitted from Parquet exports (present in CSV/JSON/JSONL). Traces and scores are unaffected. Docs follow-up: **LFE-10748**.

## Type of change

## Mandatory Tasks

## Checklist

- [x] Typecheck and lint pass (`pnpm turbo run typecheck lint --filter=web --filter=worker`).
- [x] Updated tests for the removed whitelist / new default; tRPC server tests pass locally (41/41). The HTTP public-API and MinIO worker suites run in CI (green).
- [x] Public API contract updated in Fern and the served OpenAPI spec regenerated (`fern-api export`); generated SDKs are produced by CI.


niklassemmler:
* **#14880** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #14899, opened 2026-07-08
feat(public-api): expose isRootObservation on Observations v2 API

### LINKED ISSUE #14868: bug: app-root observations are not exposed as roots in the Observations v2 API (non-null parentObservationId, no isRoot filter)
### Describe the bug

When an app-root observation's OpenTelemetry parent span is not exported to Langfuse (e.g. a FastAPI server span excluded by `LangfuseSpanProcessor` filtering), the SDK correctly marks the observation `langfuse.internal.is_app_root` and Langfuse derives trace-level input/output from it, and the UI shows it under "Is Root Observation". But the public `GET /api/public/v2/observations` still returns a **non-null** `parentObservationId` for that observation — pointing at the parent span that was never ingested — and the only public root filter, `parentObservationId is null`, does not consult the app-root signal.

Expected: an app-root observation should be discoverable as a root through the public Observations v2 API — either its `parentObservationId` is nulled on ingestion when the referenced parent was never ingested, or an `isRootObservation` field/filter is exposed. Actual: these traces (all of our current production traffic) are invisible to any consumer that lists traces by root observation, so we cannot page recent traces one-row-per-trace via the recommended v2 endpoint.

### Steps to reproduce

1. Instrument a FastAPI service with `opentelemetry-instrumentation-fastapi` and the Langfuse Python SDK sharing the global OpenTelemetry `TracerProvider`.
2. Start a Langfuse observation inside a request handler, so its OTel parent is the auto-instrumented HTTP server span.
3. Let `LangfuseSpanProcessor` filter out the server span (the default for non-Langfuse spans). The SDK marks the flow observation an app root (`LangfuseSpanProcessor._mark_app_root_candidate` sets `langfuse.internal.is_app_root = True` because the span is exported but its parent is not).
4. Query `GET /api/public/v2/observations?parentObservationId=` (empty string = "is null"). The trace's root observation is **not** returned, even though the UI shows it as a root and trace-level input/output is populated.
5. Note there is no other public signal to filter on: `langfuse.internal.as_root` / `langfuse.internal.is_app_root` are not surfaced in the API response or in metadata. The legacy `GET /api/public/traces` does return these traces, but it is being deprecated on Cloud and returns HTTP 422 ("This legacy endpoint can be slow. Please migrate to the high-performance Observations API v2...") unless given a narrow date window.

Langfuse Cloud

*No response*

* langfuse (Python SDK): 4.7.1
* OpenTelemetry Python SDK: 1.38.0
* Integration: `opentelemetry-instrumentation-fastapi` (HTTP server span) + `LangfuseSpanProcessor` on the global TracerProvider, which filters the server span out of the Langfuse export.

### Additional information

Proposed fixes (either resolves it):

1. Preferred: on ingestion, when an observation is an app root (`is_app_root`, or its referenced parent was never ingested), null out `parentObservationId` so `parentObservationId is null` treats it as a root.
2. Expose the app-root signal in Observations v2: an `isRootObservation` boolean field in the response **and** a filterable column, so consumers can query root observations without relying on `parentObservationId` nullness.

Related issues:

* langfuse/langfuse#12896 (closed; added SDK app-root detection, but not public-API root exposure).
* langfuse/langfuse#12984 (open; same family, framed around trace I/O).
* langfuse/langfuse#9970 (closed; OTel).

No

wei-juncheng:
## Problem

Closes #14868.

When an app-root observation's OpenTelemetry parent span is never exported to
Langfuse (e.g. a FastAPI server span filtered out by `LangfuseSpanProcessor`),
the SDK marks it `langfuse.internal.is_app_root` and Langfuse treats it as a
trace root internally. But `GET /api/public/v2/observations` still returns a
**non-null, dangling** `parentObservationId` for it, and the only public root
filter — `parentObservationId is null` — does not consult the app-root signal.
Consumers listing traces one-row-per-trace via the recommended v2 endpoint
never see these traces.

## Approach

Chose the issue's **Option 2** (expose the app-root signal) over Option 1
(null out `parentObservationId` on ingestion), because Option 1 destroys real
OTel parent data irreversibly, can't reliably decide "parent will never
arrive" at ingestion time given out-of-order spans, and doesn't help already
ingested data.

While implementing, it turned out the `isRootObservation` **filter was already
functional** — the internal column `(parent_span_id = '' OR is_app_root)` was
defined and already accepted by the v2 `filter` param. It was just
undocumented, untested at the public-API layer, missing from the response, and
not surfaced through the MCP observation tools. So this PR is mostly about
surfacing and documenting it.

### Public API v2

- **Query builder** — add `is_root_observation` to `EVENTS_FIELDS` (reusing
  `eventsTableIsRootObservationSql`) and to the `basic` field group.
- **Read/domain/convert** — thread the field through the events read-record
  schema, `EventsObservationSchema`, and `convertEventsObservation`.
- **Response contract** — add `isRootObservation` to `APIObservationV2`;
  exclude it from the strict v1 `transformDbToApiObservation` (same pattern
  as `bookmarked`/`public`).
- **Fern** — document `isRootObservation` and `hasParentObservation`,
  including the dangling-parent semantics (an app root can have both
  `hasParentObservation = true` and `isRootObservation = true`).

### MCP observation tools

- **Filter allowlist** — add `isRootObservation` to
  `OBSERVATION_MCP_ALLOWED_EVENTS_TABLE_FILTER_COLUMNS`, so `listObservations`
  accepts it as a filter column and `getObservationFilterSchema` advertises it.
  Previously only `hasParentObservation` was allowed, which misses app roots
  with dangling parents.
- **Filter values** — add `isRootObservation` to the value-discovery column
  list in `getObservationFilterValues` (with boolean `true`/`false`
  normalization), and widen the `getEventFilterValuePage` column union
  accordingly.
- **Field metadata** — add the MCP observation field metadata entry (required
  by the typed field map).

## Usage

```
GET /api/public/v2/observations?filter=[{"type":"boolean","column":"isRootObservation","operator":"=","value":true}]
```
returns app roots (dangling parent) alongside true roots; each row's
`isRootObservation` distinguishes them without relying on
`parentObservationId` nullness. The same filter column is now usable from the
MCP `listObservations` / `getObservationFilterValues` tools.

## Tests

- **Targeted v2 test** — true root, app root (dangling parent + `is_app_root`),
  and child: asserts the filter both ways, that the raw `parentObservationId`
  is preserved, and that the response carries `isRootObservation`.
- **Field-group contract test** — added `isRootObservation` to the generic
  `basic`-group and non-core field lists so future placement/removal mistakes
  are caught.
- **MCP tests** — filter-schema exposure, filter-values boolean counts (app
  root counted as a root), and an end-to-end `listObservations` filter run.

## Verification

- `pnpm run lint` — `Tasks: 8 successful, 8 total`
- `pnpm run typecheck` — `Tasks: 8 successful, 8 total`
- `npx fern-api check` — `All checks passed`
- v2 observations API — `Tests 38 passed (38)`
- MCP read tools — `Tests 153 passed (153)`
- MCP public-api tools — `Tests 14 passed (14)`
- v1 observations regressions — list `Tests 24 passed (24)`, single `Tests 32 passed (32)`

> Fern SDK regeneration is left to the post-merge `sdk-api-spec` CI job (it
> runs with `FERN_TOKEN`); the definition was validated locally with
> `fern check`.

This PR exposes the pre-existing `isRootObservation` signal on the v2 observations API, enabling consumers to discover app-root observations (those whose OTel parent span was filtered out and never ingested) without relying on `parentObservationId` being null.

- **Query builder & schema**: `is_root_observation` is added to `EVENTS_FIELDS` (reusing `eventsTableIsRootObservationSql`), the `basic` field group, the ClickHouse read schema, and threaded through the converter, domain schema, and `APIObservationV2` Zod type.
- **API surface**: `isRootObservation` is deliberately excluded from the v1 `transformDbToApiObservation` transformer (same destructure-to-omit pattern used for `bookmarked`/`public`), documented in Fern, and added to the MCP field metadata and allowed filter columns.
- **Tests**: New regression tests cover the true-root, app-root (dangling parent + `is_app_root`), and child cases, asserting both filtering behaviour and the presence of `isRootObservation` in the response.
</details>

Safe to merge. The change adds a read-only, backward-compatible field to the v2 API; it does not mutate any stored data and cannot affect v1 callers.

The field is threaded consistently through every layer (SQL → read schema → converter → domain → API schema → docs), the underlying SQL expression was already tested and used for filtering before this PR, v1 callers are explicitly shielded by the destructure-to-omit pattern, and the new tests cover all three cases (true root, app root with dangling parent, child) with both filter directions.

</details>

```mermaid
sequenceDiagram
    participant Client
    participant API as v2 Observations API
    participant QB as EventQueryBuilder
    participant CH as ClickHouse (events table)

    Client->>API: "GET /api/public/v2/observations?fields=basic&filter=[isRootObservation=true]"
    API->>QB: build query (FIELD_SETS.basic includes isRootObservation)
    QB->>CH: "SELECT ..., toBool(e.parent_span_id = '' OR e.is_app_root = true) AS is_root_observation"
    CH-->>QB: rows with is_root_observation
    QB-->>API: EventsObservationRecordReadType[]
    API->>API: convertEventsObservation() maps is_root_observation to isRootObservation
    API->>API: APIObservationV2 Zod parse
    API-->>Client: "{ data: [{ id, parentObservationId, isRootObservation: true, ... }] }"
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Client
    participant API as v2 Observations API
    participant QB as EventQueryBuilder
    participant CH as ClickHouse (events table)

    Client->>API: "GET /api/public/v2/observations?fields=basic&filter=[isRootObservation=true]"
    API->>QB: build query (FIELD_SETS.basic includes isRootObservation)
    QB->>CH: "SELECT ..., toBool(e.parent_span_id = '' OR e.is_app_root = true) AS is_root_observation"
    CH-->>QB: rows with is_root_observation
    QB-->>API: EventsObservationRecordReadType[]
    API->>API: convertEventsObservation() maps is_root_observation to isRootObservation
    API->>API: APIObservationV2 Zod parse
    API-->>Client: "{ data: [{ id, parentObservationId, isRootObservation: true, ... }] }"
```

</a>
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(public-api): expose isRootObservati..."](https://github.com/langfuse/langfuse/commit/6e998a9cb84a0940eecd06b1932f8e0c4c9db3b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42760430)</sub>

### DIFF
--- a/fern/apis/server/definition/commons.yml
+++ b/fern/apis/server/definition/commons.yml
@@ -252,6 +252,13 @@ types:
       sessionId:
         type: optional<nullable<string>>
         docs: The session ID associated with the observation
+      isRootObservation:
+        type: optional<boolean>
+        docs: |
+          Whether the observation is a root of its trace. True when the observation has no parent,
+          or when it is an app root whose OpenTelemetry parent span was never ingested into Langfuse
+          (e.g. filtered out by the SDK span processor). Use this instead of relying on
+          `parentObservationId` being null to discover trace roots.
 
       # Time fields (field group: time)
       completionStartTime:
--- a/fern/apis/server/definition/observations.yml
+++ b/fern/apis/server/definition/observations.yml
@@ -17,7 +17,7 @@ service:
         ## Field Selection
         Use the `fields` parameter to control which observation fields are returned:
         - `core` - Always included: id, traceId, startTime, endTime, projectId, parentObservationId, type
-        - `basic` - name, level, statusMessage, version, environment, bookmarked, public, userId, sessionId
+        - `basic` - name, level, statusMessage, version, environment, bookmarked, public, userId, sessionId, isRootObservation
         - `time` - completionStartTime, createdAt, updatedAt
         - `io` - input, output
         - `metadata` - metadata (truncated to 200 chars by default, use `expandMetadata` to get full values)
@@ -130,6 +130,10 @@ service:
               - `version` (string) - Version tag
               - `userId` (string) - User ID
               - `sessionId` (string) - Session ID
+
+              ### Trace-Tree Position
+              - `isRootObservation` (boolean) - True for trace roots: observations without a parent, plus app roots whose OpenTelemetry parent span was never ingested (e.g. filtered out by the SDK span processor). Use this to list one root per trace even when `parentObservationId` is a dangling reference.
+              - `hasParentObservation` (boolean) - True when the observation carries a non-empty `parentObservationId`. Note: an app root can have both `hasParentObservation = true` (dangling parent) and `isRootObservation = true`.
 
               ### Trace-Related Fields
               - `traceName` (string) - Name of the parent trace
--- a/packages/shared/src/domain/observations.ts
+++ b/packages/shared/src/domain/observations.ts
@@ -121,6 +121,9 @@ export const EventsObservationSchema = ObservationSchema.extend({
   tags: z.array(z.string()).nullable().optional(),
   bookmarked: z.boolean().optional(),
   public: z.boolean().optional(),
+  // True when the observation is a trace root: no parent, or an app root
+  // whose OTel parent span was never ingested (is_app_root).
+  isRootObservation: z.boolean().optional(),
 });
 
 export type EventsObservation = z.infer<typeof EventsObservationSchema>;
--- a/packages/shared/src/eventsTable.ts
+++ b/packages/shared/src/eventsTable.ts
@@ -417,6 +417,7 @@ const OBSERVATION_MCP_ALLOWED_EVENTS_TABLE_FILTER_COLUMN_IDS = [
   "metadata",
   "traceTags",
   "hasParentObservation",
+  "isRootObservation",
 ] as const satisfies readonly EventsTableColumnId[];
 
 export type ObservationMcpAllowedEventsTableFilterColumn =
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -2,6 +2,7 @@ import {
   OBSERVATION_FIELD_GROUPS_PUBLIC_API,
   ObservationFieldGroupPublicApi,
 } from "../../../domain/observation-field-groups";
+import { eventsTableIsRootObservationSql } from "../../../eventsTable";
 import { OBSERVATIONS_TO_TRACE_INTERVAL } from "../../repositories/constants";
 import { FilterList, StringFilter } from "./clickhouse-filter";
 
@@ -113,6 +114,8 @@ const EVENTS_FIELDS = {
   environment: 'e.environment as "environment"',
   type: "e.type as type",
   parentObservationId: 'e.parent_span_id as "parent_observation_id"',
+  // App-root observations (dangling parent, is_app_root=true) count as roots
+  isRootObservation: `toBool(${eventsTableIsRootObservationSql}) as "is_root_observation"`,
   name: "e.name as name",
   level: "e.level as level",
   statusMessage: 'e.status_message as "status_message"',
@@ -375,6 +378,7 @@ const FIELD_SETS = {
     "public",
     "userId",
     "sessionId",
+    "isRootObservation",
   ],
   time: ["completionStartTime", "createdAt", "updatedAt"],
   model: ["providedModelName", "internalModelId", "modelParameters"],
--- a/packages/shared/src/server/repositories/definitions.ts
+++ b/packages/shared/src/server/repositories/definitions.ts
@@ -124,6 +124,7 @@ export const eventsObservationRecordReadSchema =
     tags: z.array(z.string()).optional(),
     bookmarked: z.boolean().optional(),
     public: z.boolean().optional(),
+    is_root_observation: z.boolean().optional(),
   });
 export type EventsObservationRecordReadType = z.infer<
   typeof eventsObservationRecordReadSchema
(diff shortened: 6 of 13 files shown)


BorisSmorodin:
Independent confirmation of the same shape from a different trigger, in case it helps scope this.

The dangling parent does not only come from a parent span filtered by `LangfuseSpanProcessor`. It also appears when the caller pins only the trace id — `trace_context={"trace_id": ...}` with no `parent_span_id`. To attach a span to an existing trace, OpenTelemetry requires a parent `SpanContext`, so the SDK builds one with a random span id:

```python
int_parent_span_id = (
    int(parent_span_id, 16) if parent_span_id
    else RandomIdGenerator().generate_span_id()
)
```

That span is never exported, the observation is therefore an app root, and `GET /api/public/v2/observations` returns a non-null `parentObservationId` for it — exactly the shape described here.

Observed against a self-hosted Langfuse 3.225.9 with langfuse 3.15.0: a root observation created this way comes back with `parentObservationId` pointing at an id that is present neither among the trace's observations nor anywhere in the project.

This will keep arriving: 3.15.0 is the latest stable 3.x, and `_create_remote_parent_span()` is unchanged in 4.15.4.

Our case is distributed tracing across processes — the run identifier doubles as the trace id, so the trace id is pinned deliberately and there is no real parent span to name. The app-root marker keeps the UI correct (the node is drawn as a root), so the cost lands only on API consumers: anything that lists a trace by its root observation has to special-case a parent that does not exist.

PR #14934, opened 2026-07-09
feat(trace): graph view mode switcher — aggregated vs expanded as-it-ran DAG (LFE-10676)

nkabardin:
## TL;DR

The trace **graph view** gets a mode switcher: the existing **Aggregated** graph (repeated steps collapse into one `name (3/3)` node), plus a new **Expanded** "as it ran" mode — one node per call, so three `litellm_request` calls are three distinct nodes and loops unroll into an acyclic DAG.

## Why

Customers like the graph view but find the aggregated form hard to follow: pattern-grouping renders repeated calls as cycles, and not everyone reads that. Several asked for the full process as a DAG, in execution order, as an alternative (redundancy with the tree view is expected — that's what they ask for). Linear: LFE-10676.

## What

- **Mode switch** (segmented pill, `Aggregated | Expanded`) overlaid top-left on the graph canvas, styled like the Tree/Timeline switch; collapses to icons on narrow panels. Selection persists as a trace view preference (`ViewPreferencesContext`, localStorage, sanitized against stale values).
- **Expanded mode is hierarchy-first and framework-agnostic**: it builds from the raw observations (minus EVENTs) — the instrumented parent-child structure is the source of truth, so **every call becomes a node**, including the LLM/tool calls inside LangGraph nodes that the aggregated path's normalization filters out. Edges: a call hangs off its parent (descent into a subtree) or off the sibling(s) that actually finished before it started (fork/join; timing is only used to order siblings *within* one parent's scope — no step inference, no framework metadata). A step-sequence variant was prototyped and dropped after live comparison.
- **No label suffixes**: repeated names stay distinct as separate nodes; the layout's run order tells you which iteration is which (the tree doesn't number repeats either, and this avoids confusion with the aggregated `(3/3)` cycling counter). Aggregated keeps the framework-nodes view and its cycling counter unchanged.
- **Left→right wrapped layout** for expanded mode: unrolled runs are long thin chains; ELK `MULTI_EDGE` wrapping snakes them across rows so fit-zoom stays legible (wrap capped at 300 nodes — the wrapper recursion is per-layer and stack-overflows on very long chains; a plain ribbon takes over beyond). Aggregated stays top-down, unchanged. `__start__`/`__end__` anchors are pinned to the first/last layer.
- **Guardrails**: a 10k edge budget (parallel joins are combinatorial; past it the view shows a "too complex" notice instead of freezing ELK), allocation-free sibling scan (36 ms at the 5000-observation panel cap).
- **The tree/graph split ratio now persists** (localStorage, `useDefaultLayout` — the LFE-10601/LFE-10729 model): dragging the graph panel bigger sticks across reloads, with peek and full-page each keeping their own group. Only the open ratio is saved (the collapsed 28px rail is guarded out; collapse state persists separately per-project), and reopening a collapsed-at-mount panel restores the saved ratio atomically via the group's `setLayout`.
- Node id = observation id in expanded mode, so tree↔graph selection maps 1:1 — clicking any node (including a nested LLM/tool call) selects exactly that observation; observations without their own node highlight their nearest ancestor in the graph.

## Result

Aggregated is byte-identical in behavior. Expanded verified live on LangGraph-style refine-loop traces (metadata + timing-only paths): loops unroll in order with children rendered, selection syncs both directions down to individual child calls, mode persists across reloads, stale prototype localStorage values degrade to the default, 200-node expanded graphs lay out in ~330 ms.

<details>
<summary>Engineering detail & verification</summary>

### Builder
- `buildExpandedGraph(data, ancestry)` in `web/src/features/trace-graph-view/buildExpandedGraph.ts`; returns the same `{graph, nodeToObservationsMap}` pair as the aggregated builder, so everything downstream (ELK layout, renderer, selection) is mode-agnostic.
- Dedupes duplicate observation ids, drops incoming synthetic system rows and re-derives `__start__`/`__end__` from the built edges (sources/sinks; sinks capped to the root-level run so bushy trees don't converge on `__end__`).
- Sibling edges are the transitive reduction of the happened-before interval order; an instant / still-running / same-millisecond frontier falls back to one edge from the latest-ending predecessor so chains of instants stay chains (never orphaned onto `__start__`).
- Parent resolution walks ancestor chains through observations missing from the graph (EVENTs) via the unfiltered data.

### Selection / playback
- URL `?observation=` sync: direct id hit in expanded mode; fallback walks up `parentObservationId` to the nearest id present in the graph. Aggregated path unchanged. The click-echo ref is cleared on mode switch so selection re-resolves into the new node-id space.
- Playhead glow projection is identity in expanded mode (observation id = node id).

### Verification
- Unit: `web/src/__tests__/buildExpandedGraph.clienttest.ts` — 15 tests (distinct repeated-name nodes, sibling chain/fork/join, transitive reduction, instant/un-ended/same-ms predecessors, ancestry walk, orphan grouping, id dedupe, system-row replacement, edge-budget bail, 5000-chain under budget, identity map, empty input). `Tests 15 passed (15)`; existing graph tests pass.
- Live (Playwright, dev server): aggregated regression on both seeded shapes (langgraph trace still shows framework nodes only); expanded renders all 24 calls of a 3-turn refine loop including nested LLM/tool children; clicking a child node selects exactly that observation; mode persists across reload; stale-enum fallback; icon collapse below 340px canvas; zero page errors cycling modes on fresh load.
- `pnpm run lint` (`Tasks: 8 successful, 8 total`), `pnpm --filter web run typecheck` clean.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

PR #14939, opened 2026-07-09
fix(filters): deselecting a facet value excludes it instead of selecting the complement (LFE-10717)

nkabardin:
## TL;DR

Unchecking one value in a sidebar facet (e.g. one user in Sessions → User IDs) used to persist the **complement** — `any of [all ~999 other options]` — which blows past the HTTP header budget on refresh (**431**, blank page) and returns **wrong results** for multi-valued columns. A deselect now persists a compact `none of [unchecked]`, and a defensive cap keeps any oversized filter state out of the URL entirely.

## Why

A customer excluded their own user on the Sessions table: the deselect worked in-session, but on refresh the URL carried every *other* user ID (`userIds;arrayOptions;any of;[100+ ids]`) → HTTP 431 → no data (LFE-10717).

It wasn't just a URL-size bug. For a multi-valued column, `any of [everyone-except-X]` ≠ `none of [X]`: a session with users X **and** Y still matched, so the "excluded" user's multi-user sessions kept showing. And since facet options are top-N-capped (1000 for user IDs), the materialized complement silently excluded every user beyond the cap too. The complement is O(option-count); the user's intent is O(what they touched).

## What

- **Deselect-from-implicit-all → `none of [deselected]`** (`resolveCheckboxOperator`). Explicit `any of` / `all of` selections are never inverted — the original author's caveat about general inversion stands; only the "everything was checked, user unchecked N" gesture converts, where it is exactly `none of [those N]`.
- **Unified checked=kept display**: arrayOptions `none of` now renders as everything-but-excluded checked with the NONE badge lit — the model stringOptions facets always had. Exclusions that fall out of the (top-N, time-scoped) option list are carried over on later interactions and keep the facet's Clear affordance alive.
- **SOME/ALL/NONE toggle carries the value list** (`any of [a,b]` ⇄ `none of [a,b]`), and label-click "Only X" replaces an active exclusion with `any of [X]`.
- **Excluded rows pin to the top** of long option lists (extends the LFE-10494 pinning: the applied filter is what gets surfaced, and under NONE that's the unchecked rows).
- **URL safety net**: serialized filter state above `MAX_URL_FILTER_QUERY_LENGTH` (4000 chars) never reaches the URL — it lives in the existing per-tab sessionStorage mirror. Same-tab refreshes keep working; a copied link degrades to "no filter" instead of a dead 431 page. Protects every large arrayOptions facet (tags, labels) against pathological selections.

## Result

On a Sessions table with 200+ user IDs: uncheck one user → URL is `?filter=userIds;arrayOptions;;none of;user_10` (~50 chars), survives refresh with data, and multi-user sessions containing that user are correctly excluded. Old URLs and saved views with `none of` filters keep identical query semantics (display flips to the unified model).

<details>
<summary>Mechanism, decisions, verification</summary>

### Semantics decisions (made with maintainer)

- **Unified checked=kept model** over keeping the old NONE-mode convention (checked=excluded): the same persisted filter shape can only display one way, and the old convention would make the primary gesture look broken (the box you just unchecked re-checks itself under a NONE badge). Pre-existing manual-NONE filters keep identical semantics; their checkbox display flips to the complement.
- **"Only" in NONE mode** now means only: `any of [X]`, not `none of [X]` (which is everything-*except*-X — the opposite of the affordance).
- Guard cap (4000 chars) and its scope (`urlAndSessionStorage` state location only — the one with a fallback store; plain `url` tables have small facets) were implementation calls.

### Edge cases covered

- Re-checking the last excluded value clears the filter (no lingering empty `none of`).
- Exclusions outside the current option list (time range changed, top-N cap) are preserved on interaction and keep `isActive`/Clear on — for stringOptions facets too, which previously showed an active none-of filter as a fully-idle facet when its value wasn't in the option list.
- Unchecking everything keeps accumulating exclusions (`none of [all]`), consistent with the exclusion model; the pre-existing stringOptions "uncheck last → reset" top-guard path is unchanged.
- The oversized-URL guard is wired through all three URL touch points (write, optimistic-pending clear, canonicalization) so a hand-pasted oversized legacy URL is evicted to sessionStorage instead of wedging the optimistic state.

### Adversarial review pass

Two independent skeptic reviews (product + technical) ran against the first commit; everything actionable is fixed in the follow-up commit:

- **"All" label-click on the last kept value** under an active none-of filter derived "exclude every option" from the emptied checked set (both reviews flagged it; reproduced in a test). Now re-selects everything.
- **NONE toggle with nothing deselected** persisted a vacuous `none of []`. Now a no-op — deselecting enters NONE mode by itself.
- **stringOptions facets** now get the excluded-row pinning too (operator is exposed for them; the SOME/ALL/NONE toggle stays arrayOptions-only). Their none-of filters previously buried the just-unchecked row behind "Show more".
- **Oversized evictions/canonicalization use `replaceIn`** — pushing turned Back into a rewrite loop on legacy oversized links.

Flagged-and-accepted consequences (semantics-preserving, worth a changelog note):

- SOME⇄NONE toggling carries the stored value list, so the visible checked set flips to its complement (`any of [a,b]` ⇄ `none of [a,b]` — a clean involution; the alternatives are the 431 bug or broken round-trips).
- Filters saved earlier via the manual NONE mode keep identical query semantics; their checkbox display flips to the unified model (the old muscle memory still works: clicking the value you want excluded still excludes it — it's checked now, and unchecking excludes).
- Unchecking *every* value of an arrayOptions facet accumulates `none of [all known values]` (shows only rows without any known value — a real query for multi-valued columns); stringOptions keeps its pre-existing last-uncheck-resets behavior.
- An exclusion whose value left the (top-N/time-scoped) option list is not individually listed in the facet (it is carried on interactions, keeps the facet's Clear affordance, and shows in the URL); surfacing it as an unchecked pinned row is a possible fast-follow.
- The URL-budget eviction is silent (console.warn). Legacy giant `any of` bookmarks still 431 server-side — nothing client-side can prevent that; this PR stops *creating* such URLs.

### Verification

- TDD: 14 failing tests first, then the fix. New gesture-level suite (`sidebarArrayOptionsExclusion.clienttest.tsx`) drives the real hook through uncheck/re-check/Only/toggle journeys; URL-guard suite covers apply/restore/resume-after-oversized; pure-function and pinning tests updated.
- Full web client suite: `Test Files 141 passed | 1 skipped (142)`, `Tests 1927 passed | 79 skipped (2006)`. Lint clean, `build:check` clean (clienttest files are type-checked there).
- Browser-verified on a seeded instance (200+ users, 100-session pool, multi-user sessions): deselect → compact URL → refresh 200 with data; NONE⇄SOME toggle shows exactly the ClickHouse-verified session sets (excluded user's 7 sessions absent under NONE, present under SOME); `traceTags` arrayOptions and `traceName`/`level` stringOptions facets on the v4 tables behave as expected.

</details>

This PR fixes a semantic bug (LFE-10717) where unchecking a value from an `arrayOptions` facet materialized the complement as `any of [all-other-N]`, which caused HTTP 431 errors at large option counts (~1000 user IDs) and produced wrong results for multi-valued rows. Deselects from the implicit-all state now persist as `none of [deselected]`, and a 4000-char URL guard evicts oversized states to session-storage only.

- **Core semantic fix** in `resolveCheckboxOperator`: the "no existing filter + uncheck" gesture now produces `none of [deselected]` instead of `any of [remaining]`, with stale exclusions (values that fell out of the top-N option list) carried forward on subsequent interactions.
- **Unified display model** in `computeSelectedValues`: `arrayOptions "none of"` now renders as its complement (everything-but-excluded checked), the same way `stringOptions "none of"` always has, so the deselect gesture visually round-trips.
- **URL safety net** via `toUrlFilterQuery` applied at all three URL touch points (write, pending-clear, canonicalization), with `replaceIn` used for evictions to prevent a Back-button rewrite loop on legacy oversized links.
</details>

Safe to merge; the changes are well-scoped and the new `none of` deselect path has thorough unit and gesture-level coverage.

The three interconnected changes — semantic operator derivation, unified checkbox display, and URL budget guard — are each independently testable and independently tested. The `resolveCheckboxOperator` rewrite covers the stale-exclusion carry-over, the empty-exclusion-set teardown, and the explicit-operator preservation paths, all with new unit tests. The gesture suite drives the real hook through uncheck/re-check/Only/toggle/stale-exclusion journeys. The URL guard is wired at all three URL touch points and its test suite verifies apply, same-tab-restore, and resume-after-oversized sequences. No mutations to server-side query logic or DB schema.

No files require special attention; `useSidebarFilterState.tsx` is the most complex touched file but its new paths are all covered by the added tests.
</details>

```mermaid
flowchart TD
    A["User unchecks a facet value\n(onChange called)"] --> B{"existingFilter?"}

    B -- "none (implicit-all)" --> C["resolveCheckboxOperator\n!existingFilter branch"]
    C --> D["deselected = availableValues − checked\nreturn none of [deselected]"]

    B -- "arrayOptions none of [...]" --> E["resolveCheckboxOperator\nnone-of branch"]
    E --> F["carriedExclusions = existingFilter.value ∩ ¬available\ndeselected = available − checked\nreturn none of [carried + deselected]"]

    B -- "arrayOptions any of [...]" --> G["return any of [values]\n(positive selection preserved)"]
    B -- "arrayOptions all of [...]" --> H["return all of [values]"]

    D --> I{"finalValues.length === 0?"}
    F --> I
    I -- "Yes" --> J["Remove filter (implicit-all)"]
    I -- "No" --> K["Persist filter state"]

    K --> L["encodeFiltersGeneric(filters)"]
    L --> M{"stateLocation =\nurlAndSessionStorage?"}
    M -- "Yes" --> N{"encoded.length >\nMAX_URL_FILTER_QUERY_LENGTH\n(4000)?"}
    N -- "Yes (oversized)" --> O["urlQuery = ''\nsetUrlFiltersQuery(null, replaceIn)\nsetStoredFiltersQuery(encoded)"]
    N -- "No" --> P["setUrlFiltersQuery(encoded)\nsetStoredFiltersQuery(encoded)"]
    M -- "No (url only)" --> Q["setUrlFiltersQuery(encoded)"]

    O --> R["console.warn oversized state\nSession storage keeps filter\nCopied link degrades to no-filter"]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A["User unchecks a facet value\n(onChange called)"] --> B{"existingFilter?"}

    B -- "none (implicit-all)" --> C["resolveCheckboxOperator\n!existingFilter branch"]
    C --> D["deselected = availableValues − checked\nreturn none of [deselected]"]

    B -- "arrayOptions none of [...]" --> E["resolveCheckboxOperator\nnone-of branch"]
    E --> F["carriedExclusions = existingFilter.value ∩ ¬available\ndeselected = available − checked\nreturn none of [carried + deselected]"]

    B -- "arrayOptions any of [...]" --> G["return any of [values]\n(positive selection preserved)"]
    B -- "arrayOptions all of [...]" --> H["return all of [values]"]

    D --> I{"finalValues.length === 0?"}
    F --> I
    I -- "Yes" --> J["Remove filter (implicit-all)"]
    I -- "No" --> K["Persist filter state"]

    K --> L["encodeFiltersGeneric(filters)"]
    L --> M{"stateLocation =\nurlAndSessionStorage?"}
    M -- "Yes" --> N{"encoded.length >\nMAX_URL_FILTER_QUERY_LENGTH\n(4000)?"}
    N -- "Yes (oversized)" --> O["urlQuery = ''\nsetUrlFiltersQuery(null, replaceIn)\nsetStoredFiltersQuery(encoded)"]
    N -- "No" --> P["setUrlFiltersQuery(encoded)\nsetStoredFiltersQuery(encoded)"]
    M -- "No (url only)" --> Q["setUrlFiltersQuery(encoded)"]

    O --> R["console.warn oversized state\nSession storage keeps filter\nCopied link degrades to no-filter"]
```

</a>
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(filters): harden exclusion edge case..."](https://github.com/langfuse/langfuse/commit/bfdb1a059d76107142f5591fcb3bc097d766dbd4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42997215)</sub>


nkabardin:
@claude review

PR #14972, opened 2026-07-10
feat(dashboards): break down tags per tag instead of per tag-set (LFE-10834)

nkabardin:
## TL;DR

Breaking down a widget by **Tags** now yields **one bucket per tag** instead of one bucket per distinct tag-array (`"seed, session-shapes, agent"`). When the same widget filters `Tags any of [a, b]`, the breakdown shows **exactly the a and b buckets**. Exploded breakdowns are **capped at 25 buckets** (top-N by metric). And because buckets of an exploded breakdown overlap, **aggregates across buckets are made honest**: the pie center shows a deduplicated total, pivot totals that would double-count are suppressed, and affected charts carry an ⓘ notice. See **Limitations**.

## Why

A customer wanted the simplest possible tags widget: pick 2 tags, get a chart with the count for those two tags. The natural config — filter `Tags any of [a, b]` + breakdown by Tags — produced combinatorial buckets instead: the query grouped by the whole `Array(String)` column, so every distinct tag-set became its own bucket, comma-joined into labels that truncate identically in charts.

Before (filter `any of [agent, chat]` + breakdown by Tags):

| bucket | count |
|---|---|
| `seed, session-shapes, agent` | 32 |
| `seed, session-shapes, chat` | 32 |
| `support, chat, session` | 8 |

After:

| bucket | count |
|---|---|
| `chat` | 40 |
| `agent` | 32 |

Review feedback then flagged the second-order problem: per-tag buckets **overlap** (an entity with several tags counts once per tag), so cumulative visualizations — the pie center "Total", pivot grand totals, a stacked bar's height — showed sums exceeding the real entity count. We surveyed how Elasticsearch/Kibana, Druid/Superset, Amplitude, Mixpanel, PostHog, Sentry, Grafana, and Metabase handle multi-valued breakdowns: exploding per value is the norm where the feature exists at all, nobody disables chart types, and the honesty patterns in the wild are an **independently computed deduplicated total** (Mixpanel) and **structural removal of totals that would double-count** (PostHog). This PR follows those patterns rather than an opt-in explode checkbox — an opt-in would make the useless combo-buckets the default again.

## What

- Enroll `tags` in the existing `explodeArray` mechanism (already used by `toolNames`/`calledToolNames`) across all six views that declare it (v1 traces/observations/scores, v2 events-traces/events-observations/events-scores).
- Support `explodeArray` on `aggregationFunction` dimensions (v2 events views): the inner SELECT aggregates the per-entity array, the outer SELECT `arrayJoin`s it.
- Couple positive array filters to exploded breakdowns: an `any of` / `all of` filter on the same column restricts the exploded elements, so the breakdown shows exactly the selected values (`none of` keeps excluding entities without restricting elements).
- **Bucket cap (25)** for every query grouping by an exploded array dimension: non-timeseries charts always get a `LIMIT` (explicit `row_limit` clamped to 25; pivot tables keep theirs), timeseries get a top-25-series `HAVING` subquery.
- **Honest cross-bucket aggregates** (frontend only, no engine changes):
  - pie/donut center shows the **true deduplicated total** via a companion no-breakdown query (the Big Number query), not the sum of overlapping slices
  - pivot **grand totals are dropped** and **subtotals that aggregate across an exploded dimension** stay as group headers with dashed-out values; subtotals *within* one exploded bucket remain (they are honest distinct counts)
  - charts with an exploded breakdown carry a small corner **ⓘ notice**: "A trace with multiple tags is counted once in each matching bucket…"

## Limitations — read before reviewing

1. **Hard cap of 25 buckets on exploded chart breakdowns.** Non-timeseries charts always get `LIMIT ≤ 25` (top-N by the first metric); an explicit `row_limit` above 25 is **silently clamped**. The widget builder's "Breakdown Row Limit" spinner still offers 0–1000 — a UI-side clamp for array dimensions is a follow-up.
2. **Timeseries breakdowns show only the top-25 series**; series beyond that are silently dropped, and the cap subquery costs roughly one extra inner scan.
3. **Existing saved widgets and pivot tables that break down by tags change output** — combo buckets become per-tag buckets, pies get deduplicated center totals, pivot grand totals disappear. No migration or notice.
4. **Bucket values remain membership counts** — a trace with both selected tags still counts once in each bucket, and a **stacked bar's visual height** still sums overlapping series (no number can fix a visual sum; the ⓘ notice is the mitigation there). Pie slice percentages are shares of the membership sum, not of the entity total (Mixpanel-consistent).
5. **The pie's honest total costs one extra (cheap, no-breakdown) query**, fired only for pie widgets with an exploded breakdown.
6. **Untagged traces disappear from tags breakdowns** (`arrayJoin` of an empty array yields no rows — same as `toolNames`).
7. **Public Metrics API semantics change** for `dimensions: [{"field": "tags"}]`: one row per tag with a scalar string value, capped at 25 for chart-type queries. Zod contract unchanged — silent data change; deserves a changelog note.
8. **Filter coupling applies to positive operators only** (`any of` / `all of`, union across filters); `none of` doesn't restrict surviving elements.
9. **`toolNames` / `calledToolNames` breakdowns inherit all of the above** — the rules are generic for exploded array dimensions.
10. **Pivot tables are exempt from the 25 hard cap** (they keep their explicit `row_limit`) — 25 is a chart readability number and tables scroll. Flag if pivots should be capped too.

## Result

Filter two tags + breakdown by Tags → two honest bars. Pie of an overlapping breakdown → center shows the real entity count (verified live: slices 137 + 104, center 137, not 241). Pivot by tags → per-tag rows, no lying Total row. Every affected chart states the per-value counting via the ⓘ notice.

<details>
<summary>Details: engine mechanics, honest-totals mechanics, verification</summary>

### Engine mechanics
- Engine changes are additive: the outer-explode branch only activates for dimensions with both `aggregationFunction` and `explodeArray`; the filter-coupling and the bucket cap only activate for exploded dimensions. Scalar dimensions are untouched.
- The timeseries series cap ranks by the first metric over the whole window (stable series set across buckets) and reuses the inner query including the filter-coupling restriction.

### Honest-totals mechanics (frontend)
- `web/src/features/widgets/utils/exploded-breakdown.ts` — detection (`explodeArray` flag from the shared view declarations) + notice wording; consumed by WidgetForm (preview) and DashboardWidget (saved widgets).
- The pie's true total is the same query with `dimensions: []` — the existing Big Number path; no new endpoint.
- Pivot suppression lives in `pivot-table-utils.ts`: grand total dropped when any pivot dimension is exploded; a subtotal is suppressed iff a *deeper* dimension is exploded (it then aggregates across that dimension), rendered as a dashed group header.

### Verification
- Failing-first integration tests (real ClickHouse) for the engine work: per-tag breakdown, `any of`/`all of`/`none of` coupling, chart cap, pivot row_limit, timeseries top-25 series — v1 (`queryBuilder.servertest.ts`) and v2 (`dashboard-v1-v2-consistency.servertest.ts`).
- Client tests for the honest-totals layer: exploded-dimension detection + notice wording, pivot grand-total drop, level-aware subtotal suppression (`exploded-breakdown.clienttest.ts`, `pivot-table-utils.clienttest.ts`).
- Suites: queryBuilder 115, v1/v2 consistency 38, metrics API + dashboards 103, shared 17, worker monitors 34, widget client tests 301. Lint exits clean workspace-wide.
- Verified live in the widget builder: pie center 137 vs slice sum 241 with overlapping tags, ⓘ tooltip, pivot without Total row; earlier: per-tag bars, time-series legends, Big Number unchanged.

</details>


marksalpeter (changes requested):
**TLDR;** 
1. Can this be opt in (eg with a 'explode' checkbox)? 
2. Can we disable charts with cumulative totals (eg pivot tables, pice chars, and stacked bar chart time series) when tags are exploded.
3. Alternatively can we remove the totals or add a warning to the chart when this is enabled?

https://github.com/user-attachments/assets/34a00c23-f749-4f86-be40-881b4a93cd49

nkabardin:
> **Note:** this comment is from **Claude (the agent working this branch), not Nikita** — posted as research context + a record of what the commit does. The direction call stays with Nikita and Mark.

What `2c6ff68` ships (frontend only, explode semantics unchanged): pie center = deduplicated entity total via a companion no-breakdown query; pivot grand totals dropped and subtotals crossing an exploded dimension dashed out; corner ⓘ notice on exploded breakdowns.

Research summary (verified against local clones + primary docs):
- **Explode per value:** Elasticsearch/Kibana, Druid (under Superset), Amplitude, Mixpanel, PostHog cohort breakdowns. None warn on-chart or disable chart types; the honesty patterns are Mixpanel's independently-deduplicated total shown next to overlapping segments ("This is correct."), Amplitude's counting-mode toggle, and PostHog structurally removing totals that would double-count.
- **Never explode:** Grafana (labels are single-valued; group-by keys on the whole value) and Metabase (arrays second-class; group by whole array).
- **Designed away:** Sentry — tags are single-valued key→value per event.

On the follow-up points: agreed the dedup center total means slices no longer sum to it — that's the Mixpanel trade-off; removing the total instead is equally defensible and a one-line change. Group-by **metadata key** (mutually exclusive values, Sentry-style) would sidestep the overlap entirely for that field class — expressible in the current engine, but needs an efficiency check on the metadata Map columns first. Whether tags breakdowns should explode at all, revert to whole-set buckets, or something else is the open product decision for you and Nikita — the engine now supports either.

PR #14977, opened 2026-07-10
refactor(blobstorage): split settings into container + disposable form layers

niklassemmler:
## What does this PR do?

Stacked on #14964. Restructures the blob storage settings into a **state layer** and a **disposable form layer**, fixing two draft-leak regressions the previous slice introduced by replacing the guarded reset with an unconditional `keepDirtyValues` re-sync.

**Draft-lifetime model**: draft state lives exactly as long as one entity identity. `BlobStorageIntegrationContainer` owns the async inputs (project, cloud region, availability derivation), all four mutations, and the entity-action buttons (Validate / Run Now / Reset); it gates rendering behind a skeleton until inputs resolve, then mounts `BlobStorageIntegrationForm` with `key = projectId + (configured|new)`. The form takes `initialValues` once via `defaultValues` and emits values only through `onSubmit` — no reactive `values`, no `resetOptions`, no reset `useEffect`, no tRPC imports.

Fixed regressions:
- **Delete flow**: Reset → delete → refetch previously re-populated the blank form from retained dirty values, so one Save could silently re-create the deleted integration. Now delete flips the key `configured → new` and the form remounts blank, guaranteed one layer up.
- **Cross-project carryover**: unsaved credentials typed under project A survived a client-side project switch and could be saved under project B. Now the key includes `projectId`, so the draft is discarded on switch.

Declared behavior deltas (by design):
- Clean fields no longer live-update from background refetches while the form is mounted; the form is a draft scoped to the entity identity it was opened for. Same-entity refetches (5s running/queued status poll, post-save invalidation) keep the key stable and never touch the draft — mid-save typing survives, and validation errors persist across refetches instead of being partially re-synced.
- Post-create, the form remounts from the saved row, which also clears react-hook-form's retained dirty flags.
- The form is not rendered until the integration query and project have resolved (skeleton placeholder), replacing the "keepDirtyValues protects half-typed input during async load" mechanism — there is no mid-flight reset at all.

Not addressed here (tracked separately): stale async feedback after a project switch (a Validate fired under project A resolving after a switch shows its toast on project B) — the container survives the switch, so this redesign does not change it.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] Documentation update

## How should this be tested?

[...]

PR #14980, opened 2026-07-10
feat(blobstorage): drift banner + stale-feedback guards for settings form

niklassemmler:
## What does this PR do?

Stacked on #14977. Ships the three agreed follow-ups from the adversarial review of the blob storage settings refactor; the disposable form layer is untouched — everything lands in the container.

1. **Stale cross-project mutation feedback guarded.** The container survives a client-side project switch, so a mutation fired under project A that resolves after switching used to toast on project B's screen. Callbacks now compare the projectId they were fired with against the live prop and skip UI feedback when stale. Cache invalidation stays unguarded on purpose — it is keyed per projectId, so refreshing A's cache entry in the background remains correct.
2. **Real component test coverage.** Container-level tests with the tRPC boundary mocked (loading gate, delete-flow remount, project-switch isolation, post-create remount, stale-toast guard, drift banner, own-save suppression) plus form-level tests for conditional fields (provider switch, gzip toggle, custom start date) and the LFE-10296 blocked-save path. This addresses the review finding that the previously cited tests were pre-existing pure-helper tests that could not fail on component mistakes.
3. **Concurrent-edit drift banner (non-destructive).** The form freezes a snapshot at mount; the container records the user-configurable field values that snapshot was built from. A refetch whose form-bound values differ on the same entity identity (someone else saved) shows a banner — deliberately NOT keyed on `updatedAt`, which Prisma bumps on every worker status write (review finding; a timestamp signal would banner on every active sync) — the draft is kept intact until the user explicitly clicks "Reload form (discards unsaved edits)", which bumps a key epoch and remounts from the latest values. The user's own save is recognized via the mutation's returned `updatedAt` (marked in-flight from `onMutate` so racing poll responses can't flash the banner); any refetch at-or-after that value whose form-bound values match the save is adopted silently by rebaselining the drift snapshot only — the mounted draft is never remounted, so typing that continued past Save survives; the explicit Reload click clears any pending own-save expectation. Identity changes rebaseline the snapshot so the banner never fires against a stale baseline. Note: drift is detected on whatever refetch occurs (5s running/queued status poll, post-mutation invalidation) — the page intentionally does not poll otherwise.

## Type of change

- [ ] Documentation update

## How should this be tested?

- `pnpm --filter web run test-client src/features/blobstorage-integration` — `Tests 54 passed (54)` (16 component tests: 9 form + 7 container; 38 pre-existing helper tests).
- `pnpm turbo run lint typecheck --filter=web` — `Tasks: 6 successful, 6 total`.
- Manual browser verification on local seed data: saved an integration, forced queued status so the 5s poll runs, dirtied the bucket field, then bumped `bucket_name`/`updated_at` in Postgres to simulate another tab → banner appeared within one poll cycle with the draft intact; "Reload form" showed the external values and cleared the banner; a subsequent own save (with the poll still running) adopted silently with no banner and the form remounted from the saved row.

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md)
- [ ] My changes require a change to the documentation
- [ ] I have updated the documentation accordingly

This PR adds concurrent-edit drift detection and stale-mutation guards to `BlobStorageIntegrationContainer`. When a background refetch brings a newer `updatedAt` for the same entity, a non-destructive banner is surfaced and the user's draft is preserved until they explicitly reload. Mutation callbacks for the user's own save now set a `pendingSelfSave` flag so the resulting `updatedAt` bump silently remounts the form rather than triggering the banner.

- `BlobStorageIntegrationContainer.tsx`: introduces `snapshot`/`pendingSelfSave` state, render-phase derive-from-props rebaseline on identity change, `projectIdRef` stale-guard for `update` and `validate` mutation callbacks, epoch-keyed form remount, and the drift `Alert` UI.
- Two new test files cover the container's state machine (loading gate, project switch, delete, post-create, drift banner, own-save, stale mutation guard) and four additional form-level behaviors (S3-compatible field reveal, gzip toggle, custom-date export mode, blocked export-source validation).
</details>

Safe to merge; the state machine correctly handles all main lifecycle events and is well-tested.

The drift detection design is sound and the test suite is thorough. Two narrow edge cases exist: the 5 s background poll could briefly flash the drift banner on the user's own save, and a stuck pendingSelfSave flag could silently absorb a later concurrent edit without surfacing the banner.

BlobStorageIntegrationContainer.tsx — specifically the pendingSelfSave / updatedAtChanged interaction and the render-phase state transitions.
</details>

`````markdown

---

### Issue 1 of 2
web/src/features/blobstorage-integration/components/BlobStorageIntegrationContainer.tsx:104-112
**Narrow race: poll refetch can arrive before `onSuccess` sets `pendingSelfSave`**

`utils.blobStorageIntegration.invalidate()` is called on line 120, and `setPendingSelfSave(true)` is called immediately after — both in the same synchronous `onSuccess` callback. This is the safe common path. However, if the 5 s background poll fires AND resolves *after* the save completes on the server but *before* the mutation's `onSuccess` has been processed by React, the new `updatedAt` arrives while `pendingSelfSave` is still `false`, causing `hasDrift = true` and briefly flashing the drift banner — which then self-corrects once `onSuccess` does run and the epoch bumps. The banner should not appear for the user's own save.

### Issue 2 of 2
web/src/features/blobstorage-integration/components/BlobStorageIntegrationContainer.tsx:104-110
**`pendingSelfSave` stuck true can silently consume a concurrent edit**

If `setPendingSelfSave(true)` is set in `onSuccess` but the subsequent refetch (triggered by `invalidate()`) fails or returns the same `updatedAt` (e.g., optimistic caching), `pendingSelfSave` stays `true` indefinitely. In that state, the next external edit that arrives via the background poll (`updatedAtChanged = true`, `pendingSelfSave = true`) will satisfy the condition on line 104 and silently bump the epoch, remounting the form without ever showing the drift banner to the user. A timeout-based reset of `pendingSelfSave` (e.g., using a `useEffect` with a TTL) would close this window.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(blobstorage): drift banner + stale-..."](https://github.com/langfuse/langfuse/commit/3f74f05159fdcf9d32e3896731efc54312395927) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43290169)</sub>

### DIFF
--- a/web/src/features/blobstorage-integration/README.md
+++ b/web/src/features/blobstorage-integration/README.md
@@ -158,9 +158,16 @@ changes (project switch, create, delete) remount the form; same-entity
 refetches (5s status poll, post-save invalidation) keep the key stable
 and therefore can never wipe or leak a draft. Consequence, by design:
 clean fields do not live-update from background refetches while the
-form is mounted. Field-group components receive the typed `control`
-prop and subscribe to individual fields with `useWatch`; there is no
-local Zustand store and no feature `useEffect`.
+form is mounted. Instead, a refetched `updatedAt` differing from the
+mounted snapshot (someone else saved) surfaces a non-destructive
+drift banner whose explicit "Reload form" click bumps a key epoch;
+the user's own save adopts the bumped `updatedAt` silently and
+remounts from the saved row. Mutation callbacks compare their
+fired-with projectId against the live prop and skip toasts when
+stale (container survives client-side project switches). Field-group
+components receive the typed `control` prop and subscribe to
+individual fields with `useWatch`; there is no local Zustand store
+and no feature `useEffect`.
 
 **Performance boundaries**: a keystroke or select change rerenders only
 the field-group component watching that field. The form layer reads no
--- a/web/src/features/blobstorage-integration/README.md
+++ b/web/src/features/blobstorage-integration/README.md
@@ -158,11 +158,16 @@ changes (project switch, create, delete) remount the form; same-entity
 refetches (5s status poll, post-save invalidation) keep the key stable
 and therefore can never wipe or leak a draft. Consequence, by design:
 clean fields do not live-update from background refetches while the
-form is mounted. Instead, a refetched `updatedAt` differing from the
-mounted snapshot (someone else saved) surfaces a non-destructive
-drift banner whose explicit "Reload form" click bumps a key epoch;
-the user's own save adopts the bumped `updatedAt` silently and
-remounts from the saved row. Mutation callbacks compare their
+form is mounted. Instead, a refetch whose **user-configurable field
+values** differ from the mounted snapshot (someone else saved)
+surfaces a non-destructive drift banner whose explicit "Reload form"
+click bumps a key epoch. Drift deliberately does NOT key on
+`updatedAt`: Prisma's `@updatedAt` bumps on every worker status write
+(runStartedAt, lastSyncAt, …) exactly while the 5s poll is active, so
+timestamps would banner on every sync. `updatedAt` is used only to
+recognize the user's own save (the mutation's returned value; any
+refetch at-or-after it adopts silently and remounts from the saved
+row). Mutation callbacks compare their
 fired-with projectId against the live prop and skip toasts when
 stale (container survives client-side project switches). Field-group
 components receive the typed `control` prop and subscribe to
--- a/web/src/features/blobstorage-integration/README.md
+++ b/web/src/features/blobstorage-integration/README.md
@@ -165,9 +165,11 @@ click bumps a key epoch. Drift deliberately does NOT key on
 `updatedAt`: Prisma's `@updatedAt` bumps on every worker status write
 (runStartedAt, lastSyncAt, …) exactly while the 5s poll is active, so
 timestamps would banner on every sync. `updatedAt` is used only to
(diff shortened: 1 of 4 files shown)


niklassemmler:
* **#14980** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14980?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14980?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14964** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14964?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#14977](https://github.com/langfuse/langfuse/pull/14977) <a href="https://app.graphite.com/github/pr/langfuse/langfuse/14977?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
Too much effort for too little gain

PR #14920, opened 2026-07-09
fix(audit-log): Include userId in in-app agent api-key audit logs

bezbac:
<details open><summary><h3>Greptile Summary</h3></summary>

This PR enriches audit logs created via API key authentication so that in-app-agent MCP session keys carry the originating user's ID, enabling clearer attribution in the audit trail.

- The `auditLog` function is refactored into three explicit early-return branches; the `apiKeyId` branch now fetches the key row and populates `userId` from `createdByUserId` when `isInAppAgentKey` is `true`.
- A new server test file verifies both that the `createdByUserId` is stored on the key and that it propagates to the audit log record; the existing MCP tool test is updated to spy on `$transaction` instead of `auditLog.create`.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

Safe to merge — the change is narrowly scoped to the API-key audit-log path and does not touch session or userId paths.

The logic is correct: `createdByUserId` is read and conditionally set as `userId` only for in-app-agent keys, and the null-to-undefined conversion is handled properly. Tests cover both the key-creation and audit-log propagation cases.

No files require special attention.
</details>

<details><summary><h3>Sequence Diagram</h3></summary>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
sequenceDiagram
    participant Caller
    participant auditLog
    participant DB

    alt apiKeyId path (new)
        Caller->>auditLog: "auditLog({ apiKeyId, ... })"
        auditLog->>DB: $transaction start
        DB-->>auditLog: tx
        auditLog->>DB: "tx.apiKey.findUnique({ id: apiKeyId })"
        DB-->>auditLog: "{ isInAppAgentKey, createdByUserId }"
        note over auditLog: if isInAppAgentKey, userId = createdByUserId
        auditLog->>DB: "tx.auditLog.create({ ..., userId? })"
        DB-->>auditLog: created
        auditLog->>DB: $transaction commit
        auditLog-->>Caller: done
    else session path (unchanged)
        Caller->>auditLog: "auditLog({ session, ... })"
        auditLog->>DB: "auditLog.create({ userId: session.user.id, ... })"
        DB-->>auditLog: created
        auditLog-->>Caller: done
    else userId path (unchanged)
        Caller->>auditLog: "auditLog({ userId, ... })"
        auditLog->>DB: "auditLog.create({ userId, ... })"
        DB-->>auditLog: created
        auditLog-->>Caller: done
    end
```

</a>
[...]

PR #15038, opened 2026-07-13
chore(web): typecheck test files in CI and fix latent test type errors

wochinge:
## Summary

Follow-up to #15031 (broken main/production deploys). Root problem: **nothing type-checks web test files.** The CI typecheck job used `tsconfig.build.json`, which excluded `*.servertest.*`/`*.clienttest.*`; CI test builds set `NEXT_IGNORE_BUILD_ERRORS=true`; and even the production Docker build's `next build` check silently filters diagnostics in `__tests__/` directories (`next/dist/lib/typescript/runTypeCheck.js` drops them by path regex — verified by experiment). 760 type errors had accumulated invisibly, and two of them escaped far enough to break deploys.

Three changes:

- **`turbo.json`**: `NEXT_IGNORE_BUILD_ERRORS` is now part of the build task's cache key. Without this, a build cached with the check skipped can replay as a cache hit in a context that expects type-checking — silently green. Prerequisite for ever running a flag-off build in CI.
- **web ESLint**: new `no-restricted-imports` pattern banning relative imports that escape into other workspace packages (`../../../../packages/shared/...`). These bypass the exports map (which points at `dist/`) and pull shared *source* into web's TS program, where web's next-auth `User` augmentation breaks `customSsoProvider` — the exact #14920 failure mode, now caught in the editor and by the existing lint job/pre-commit hook. Static imports only: the blob-storage servertest's dynamic relative imports are a deliberate workaround (vitest loads shared's CJS dist through Node's require cache as a second module instance, so env-mutation tests need a Vite-transformed source copy) and are documented in the rule comment.
- **`web/tsconfig.build.json`**: test excludes removed, so the existing tsgo typecheck job now covers test-file bodies — strictly stronger than the prod build's check, closing the blind spot for good. The 760 latent errors this surfaced are fixed across 94 test files: session fixtures completed to the current next-auth `User` shape (`hasTraces`, `createdAt`, `aiFeaturesEnabled`/`aiTelemetryEnabled`, new feature flags), discriminant narrowing before union property access, `makeAPICall<T>` generics for error-branch responses, and `as unknown as T` casts where tests intentionally send invalid payloads. **No production code changed**; where a fix was more than type-level (fixture completions, `headers: {}` in `createInnerTRPCContext`), the affected test files were run — several hundred tests pass across the touched files. `web/tsconfig.json` also loses its vestigial `outDir` (nothing emits web via tsc; with tests in the program it produced ~400 bogus TS6059 rootDir errors that Next's check never sees).

Verification: fresh `tsgo -p tsconfig.build.json --noEmit --skipLibCheck` → 0 errors; `eslint . --max-warnings 0` → exit 0; full `pnpm --filter web run build` (the exact prod Docker step) → ✓ Compiled successfully.

Known pre-existing issues surfaced while verifying (not addressed here): `media.servertest.ts` deletes the shared seed project's media rows mid-run and can race `dataset-item-media-references.servertest.ts` in the same invocation; `event-repository.servertest.ts` has 2 failures in `getLatestSdkVersionInfoFromEvents` that reproduce on a clean tree; `ResourceSpan` in `OtelIngestionProcessor` doesn't type the protobuf-JSON buffer / hex-ID inputs it actually handles at runtime.

This PR closes a longstanding type-checking blind spot in the web package: test files were excluded from `tsconfig.build.json`, `NEXT_IGNORE_BUILD_ERRORS` masked errors in CI, and Next.js's own production type-check silently skipped `__tests__` paths — meaning 760 latent type errors had accumulated undetected. The PR fixes all three gaps without changing any production code.

- **`turbo.json`**: Adds `NEXT_IGNORE_BUILD_ERRORS` to the build task's env, ensuring cached unchecked builds can't replay as type-checked ones.
- **`web/eslint.config.mjs`**: Adds a `no-restricted-imports` pattern that catches static relative imports escaping into other workspace packages via `../…/packages|ee|worker/`; dynamic imports are explicitly left unchecked (and documented) to support the blob-storage vitest env-mutation pattern.
- **`web/tsconfig.build.json`**: Removes the test-file excludes so `tsgo --noEmit` now covers test bodies; the 760 surfaced errors are fixed by completing session fixtures, adding discriminant narrowing before union property access, and adding `makeAPICall<T>` generics.
</details>

Safe to merge — no production code changed; all edits are test fixtures, type annotations, and CI/build config.

The tsconfig and turbo.json changes are structurally sound and well-explained. The ESLint rule closes the exact failure mode from the linked incident. The 760 type fixes are mechanical and well-commented. Two minor observations (hard-coded workspace roots in the regex, `as unknown as string` casts in test fixtures) do not affect runtime behavior.

web/eslint.config.mjs — the restricted-import regex names three workspace roots explicitly; worth verifying coverage is complete as the monorepo grows.
</details>

`````markdown

---

### Issue 1 of 2
web/eslint.config.mjs:192-212
**ESLint regex covers only three workspace roots**

The pattern `^(\\.\\./)+(packages|ee|worker)/` guards `packages/`, `ee/`, and `worker/`. If the monorepo ever gains a sibling workspace root at a different path (e.g. `apps/`, `infra/`, or a future `services/`) a relative escape through it would bypass the rule entirely. Worth confirming these three names cover all current and planned workspace roots, or switching to a broader negative-anchor like `^(\.\./){3,}` that triggers on any path going three or more levels up from `web/src`.

### Issue 2 of 2
web/src/__tests__/server/traces-trpc.servertest.ts:388-400
**`input`/`output` cast works around a type gap in `createTrace`**

The cast `{ query: "..." } as unknown as string` silences the compiler, but the underlying `createTrace` factory declares `input`/`output` as `string` while the tests (correctly) pass objects that the ClickHouse client serializes. The cast is a workaround, not a fix — a narrower solution would be to widen the `createTrace` factory types to `string | Record<string, unknown>` (or `unknown`) so future callers get accurate IDE feedback without needing a cast. This is cosmetic for now because the casts are well-commented, but worth tracking.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore(web): typecheck test files in CI a..."](https://github.com/langfuse/langfuse/commit/05c8d26c83a8c8afa2fd86e38fe51af495558528) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43798308)</sub>


wochinge:
@claude review

PR #15058, opened 2026-07-13
fix(web): fall back when crypto.randomUUID is unavailable on non-secure origins (LFE-10858)

### LINKED ISSUE #15008: bug: Frontend crashes with client-side exception on browsers where crypto.randomUUID() is unavailable
### Describe the bug

When accessing a self-hosted Langfuse instance from another device on the same LAN, the frontend fails to load on certain browsers with a client-side exception.

The page displays:

`Application error: a client-side exception has occurred while loading <host-ip> (see the browser console for more information).`

The browser console reports an error related to:

`TypeError: crypto.randomUUID is not a function
`
The application becomes unusable after the page loads.

Interestingly, the issue appears browser/environment dependent:

Firefox: reproducible even after enabling the relevant HTTP/security flag.
Chrome: works after enabling the corresponding insecure/private network access flag.
Accessing the same self-hosted instance locally via localhost (through port forwarding on the host machine) works without issues.

This suggests the frontend assumes crypto.randomUUID() is available without a fallback, leading to compatibility issues in certain browser/network configurations.

### Steps to reproduce

1. Deploy Langfuse as a self-hosted instance.
2. Connect another device on the same LAN.
3. Open the Langfuse UI using the server IP and port (e.g. `http://<ip>:<port>`).
4. Log in using valid credentials.
5. Create or upload a trace.
6. Observe the client-side exception

Self-hosted

v3.195.0 OSS

_No response_

### Additional information

The frontend should either:

- gracefully fall back when` crypto.randomUUID()` is unavailable, or
- detect unsupported environments and provide a meaningful error instead of crashing.

The issue only occurs when accessing the instance remotely over the LAN using the server IP. Accessing the same instance locally via localhost (using port forwarding) works correctly.

Yes

nkabardin:
## TL;DR

Self-hosted Langfuse opened over plain HTTP (e.g. `http://<lan-ip>:3000`) white-screens the moment anything calls `crypto.randomUUID()`, because that API only exists in secure contexts. This PR installs a uuid-backed `crypto.randomUUID` polyfill at the client entry point — covering our code **and every bundled dependency, permanently** — and adds a documented `safeRandomUUID()` helper for direct id generation in client code.

Fixes langfuse/langfuse#15008 — thanks @RHarish1 for the detailed report (and for offering to contribute a fix!).

## Why

`crypto.randomUUID` (and `crypto.subtle`) are secure-context-only Web APIs: present on HTTPS and `localhost`, `undefined` over plain HTTP to a LAN IP — a very common self-hosting setup. Any unguarded call throws `TypeError: crypto.randomUUID is not a function` and takes down the app with "Application error: a client-side exception has occurred".

langfuse/langfuse#14498 already fixed the two call sites in our own code (dataset item media upload, shipped in v3.196.0 — the reporter was on v3.195.0). But the crash class remained open: any dependency bump can reintroduce an unguarded call that a static grep of our sources will never see, and it only manifests for self-hosters on HTTP.

## What

* `web/src/polyfills/crypto-random-uuid.ts` — installs `crypto.randomUUID` when missing, as the **first import of** `_app.tsx`, so it runs before any other module can call it. The fallback forces uuid's `crypto.getRandomValues` path (available in non-secure contexts) and is written so it cannot recurse into itself.
* `web/src/utils/safe-random-uuid.ts` — `safeRandomUUID()`, the documented helper for direct id generation in client code; the two existing ad-hoc `uuidv4()` call sites from langfuse/langfuse#14498 now use it, replacing their copy-pasted explanatory comments with one canonical home.
* Unit tests for both (install-when-missing, keep-native, module-import side effect, native-vs-fallback branches).

Related prior art: langfuse/langfuse#14498 (dataset media ids/hash without secure-context crypto), [#7227](<https://github.com/langfuse/langfuse/issues/7227>) (clipboard fallback for non-secure contexts — same API family).

<details><summary>Mechanism, pitfall found while testing, and verification detail</summary>

### Recursion pitfall (caught by the new tests)

uuid v14 checks `crypto.randomUUID` **lazily at call time** (`if (!buf && !options && crypto.randomUUID) return crypto.randomUUID();` — v11 bound it once at module load). A naive polyfill `crypto.randomUUID = () => uuidv4()` therefore recurses infinitely: `uuidv4()` sees the polyfill and calls it back. The installed fallback is `() => uuidv4({})` — passing an options object makes uuid take its `getRandomValues` path unconditionally, so the polyfill can never re-enter itself regardless of uuid's binding style.

### Why first-import placement

ES module evaluation is depth-first in import order, so the side-effect import at the top of `_app.tsx` executes before every other app module (and their dependency graphs). A module-scope `crypto.randomUUID` call anywhere downstream is already covered.

### Audit of bundled dependencies (current main)

All client-bundled packages that reference `crypto.randomUUID` guard it today: `react-hook-form`, `next-auth`, `@sentry/core`, `uuid`, `mermaid`. The polyfill exists so the next bump that doesn't guard cannot regress self-hosters.

### Verification (all in a genuine non-secure context: `http://192.168.1.19:3058`, `isSecureContext === false`, `crypto.randomUUID`/`crypto.subtle` undefined natively)

* **Reproduced the mechanism on main first**: `crypto.randomUUID()` throws the reporter's exact `TypeError`; confirmed the two langfuse/langfuse#14498 call sites are the only literal client-side usages in our sources today (rest are server-side `node:crypto`).
* **Production build** (`next build` + `next start`, the bundle-sensitive environment) verified in **Firefox and Chrome**: login → traces → seeded trace detail → dataset new-item dialog → real media upload (SHA-256 via the pure-JS fallback since `crypto.subtle` is unavailable) → item submit → sweep of home/dashboards/sessions/prompts/playground/settings. Zero uncaught page errors; `typeof crypto.randomUUID === "function"` and returns valid v4 — polyfill demonstrably active in the minified bundle.
* **Dev/turbopack** smoke-checked the same way (polyfill active, no errors).
* `pnpm run scan:client-bundle`: `clean: every referenced identifier is declared or a known global` (623 files).
* Client tests: `Test Files 2 passed (2)`, `Tests 8 passed (8)`. Web lint and typecheck pass.

### Graceful degradation note

The ticket also asked for a meaningful message instead of a white screen in "truly unsupported" environments. With the polyfill, the only remaining requirement is `crypto.getRandomValues`, which is available in non-secure contexts in every browser Langfuse supports — the crash class this ticket describes is gone, so no additional error-boundary UX was added here.

</details>

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

This PR installs a `crypto.randomUUID` polyfill at the client entry point (`_app.tsx`) to prevent white-screen crashes on self-hosted Langfuse instances served over plain HTTP, where the secure-context-only Web API is unavailable. It also introduces a `safeRandomUUID()` utility as the canonical helper for direct UUID generation in client code, replacing two ad-hoc `uuidv4()` call sites from the prior fix (langfuse/langfuse#14498).

* `crypto-random-uuid.ts` patches `globalThis.crypto.randomUUID` at module-evaluation time using `uuidv4({})` — the options-object trick avoids infinite recursion with uuid v14's lazy `crypto.randomUUID` check, and the first-import placement in `_app.tsx` ensures coverage of all downstream modules and bundled dependencies.
* `safe-random-uuid.ts` wraps `uuidv4()` (which itself falls back to `crypto.getRandomValues` in non-secure contexts) so call sites don't need to reason about import order or polyfill presence.
* Unit tests cover install-when-missing, keep-native, no-op on undefined, and module-scope side-effect scenarios for both utilities.

</details>

Safe to merge — the change is narrowly scoped to UUID generation in non-secure HTTP contexts and cannot regress HTTPS deployments.

The polyfill is gated behind typeof target.randomUUID === 'function', so HTTPS/localhost instances that already have the native API are completely unaffected. The recursion pitfall (uuid v14's lazy crypto.randomUUID read) is correctly neutralised by passing {} as options. Both the polyfill and the safeRandomUUID helper include tests that cover install-when-missing, keep-native, and module-scope side-effect scenarios, including the uuid v11/v14 binding difference. The first-import position in \_app.tsx is enforced by a code comment and guarantees coverage of all bundled dependency call sites.

No files require special attention. The implementation is self-contained across the two new utility files and their tests.

</details>

<details><summary>Flowchart</summary>

[%%{init: {'theme': 'neutral'}}%% flowchart TD A\[Browser loads \_app.tsx\] --> B\[First import: crypto-random-uuid.ts\] B --> C{globalThis.crypto.randomUUID defined?} C -- Yes, HTTPS/localhost --> D\[Keep native — no-op\] C -- No, plain HTTP --> E\[Install polyfill: crypto.randomUUID = uuidv4 with options\] E --> F\[All subsequent module imports\] D --> F F --> G{Dependency calls crypto.randomUUID} G -- polyfill installed --> H\[uuidv4 with options takes getRandomValues path\] G -- native present --> I\[Native crypto.randomUUID\] F --> J{Our code calls safeRandomUUID} J --> K\[uuidv4 no options\] K --> L{crypto.randomUUID available at call time?} L -- Yes native or polyfill --> M\[Uses it\] L -- No --> N\[Falls back to crypto.getRandomValues\] H --> O\[Valid UUID v4\] I --> O M --> O N --> O](<#gh-light-mode-only>) [%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%% flowchart TD A\[Browser loads \_app.tsx\] --> B\[First import: crypto-random-uuid.ts\] B --> C{globalThis.crypto.randomUUID defined?} C -- Yes, HTTPS/localhost --> D\[Keep native — no-op\] C -- No, plain HTTP --> E\[Install polyfill: crypto.randomUUID = uuidv4 with options\] E --> F\[All subsequent module imports\] D --> F F --> G{Dependency calls crypto.randomUUID} G -- polyfill installed --> H\[uuidv4 with options takes getRandomValues path\] G -- native present --> I\[Native crypto.randomUUID\] F --> J{Our code calls safeRandomUUID} J --> K\[uuidv4 no options\] K --> L{crypto.randomUUID available at call time?} L -- Yes native or polyfill --> M\[Uses it\] L -- No --> N\[Falls back to crypto.getRandomValues\] H --> O\[Valid UUID v4\] I --> O M --> O N --> O](<#gh-dark-mode-only>)

</details>

Reviews (1): Last reviewed commit: ["fix(web): fall back when crypto.randomUU..."](<https://github.com/langfuse/langfuse/commit/5d3de7062eb25882b7546e268468af429597c4dd>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=43981138>)


nkabardin:
@claude review

PR #15083, opened 2026-07-14
fix(worker): stream core data S3 exports with keyset pagination

Steffen911:

This PR replaces the all-at-once in-memory fetch (`Promise.all` + `findMany` for every table) with a keyset-paginated streaming export: each table is streamed in 1,000-row pages via an `AsyncGenerator`, uploaded to S3 with `uploadFileBuffered`, and run with `pLimit(3)` concurrency so the Prisma pool stays available to other worker jobs. The `DataRetentionJob` cron is shifted from 3:15 am to 2:45 am to avoid competing for the same connections.

- **Keyset pagination (`createTableJsonlStream`)**: uses `cursor + skip: 1` on a stable `orderBy: { id: \"asc\" }` (or composite key) per table; memory footprint is now O(page size) instead of O(table size).
- **Concurrency and error handling**: `Promise.allSettled` lets all 11 tables attempt export even if one fails; failures are counted and logged, though only the first error is re-thrown.
- **Schedule migration**: `removeRepeatable(\"15 3 * * *\")` fires async on first queue instantiation to clean up the old BullMQ key before installing the new `\"45 2 * * *\"` pattern.
</details>

Safe to merge; the core pagination logic is correct and well-tested, with two minor quality gaps worth addressing in a follow-up.

The keyset streaming implementation is correct and the tests cover all meaningful edge cases (empty tables, full pages, partial last pages, error propagation). Two gaps remain: when multiple table exports fail simultaneously only the first error is rethrown, and `billingMeterBackup` is the only table without an explicit `select`, leaving it exposed to future schema additions that could silently widen the export surface.

The failure-aggregation block in `coreDataS3ExportProcessor` (around the `failures[0].reason` throw) and the `billingMeterBackup` `fetchPage` definition both deserve a second look before the next schema change or alerting integration lands.
</details>

`````markdown

---

### Issue 1 of 2
worker/src/queues/coreDataS3ExportQueue.ts:449-457
**Only first failure re-thrown; others silently dropped**

When multiple table exports fail, `failures[0].reason` is thrown but the remaining errors are not surfaced beyond the count in the aggregate log. BullMQ will retry the job with only that one error visible in the retry payload, and the root causes of the other failures are lost (per-table `logger.error` lines still fire, but the thrown error aggregates nothing). Consider throwing an `AggregateError(failures.map((f) => f.reason), ...)` so all failure reasons are preserved in the retry / alerting context.

### Issue 2 of 2
worker/src/queues/coreDataS3ExportQueue.ts:277-312
**`billingMeterBackup` fetches all columns — no `select` guard**

Every other table export uses an explicit `select` to control which columns are exported and to deliberately exclude sensitive fields (e.g., `password`, `authConfig`). The `billingMeterBackup` entry omits `select` entirely, so any column added to that model in the future will appear in the S3 export without a conscious decision to include it. Aligning this with the other tables by adding an explicit `select` would make the exported surface area intentional and safe across schema changes.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: revert retention queue settings"](https://github.com/langfuse/langfuse/commit/faa5893ade395c9c605f4842665dc94c938e4f01) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44154992)</sub>

### DIFF
--- a/packages/shared/src/server/redis/dataRetentionQueue.ts
+++ b/packages/shared/src/server/redis/dataRetentionQueue.ts
@@ -35,25 +35,12 @@ export class DataRetentionQueue {
 
     if (DataRetentionQueue.instance) {
       logger.debug("Scheduling jobs for DataRetentionQueue");
-      // Remove the old 3:15am cron pattern - BullMQ keys repeatable jobs by
-      // name + pattern, so changing the pattern creates a second schedule
-      // while the old one keeps firing.
-      DataRetentionQueue.instance
-        // eslint-disable-next-line @typescript-eslint/no-deprecated -- Existing repeatable-job cleanup; job scheduler migration should be handled separately.
-        .removeRepeatable(QueueJobs.DataRetentionJob, {
-          pattern: "15 3 * * *",
-        })
-        .catch((err) => {
-          logger.error("Error removing legacy DataRetentionJob schedule", err);
-        });
       DataRetentionQueue.instance
         .add(
           QueueJobs.DataRetentionJob,
           {},
           {
-            // every day at 2:45am, 30min before the core data S3 export to
-            // avoid contending for the same worker postgres connection pools
-            repeat: { pattern: "45 2 * * *" },
+            repeat: { pattern: "15 3 * * *" }, // every day at 3:15am
           },
         )
         .catch((err) => {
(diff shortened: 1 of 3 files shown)


Steffen911:
@claude review

PR #15120, opened 2026-07-15
feat(otel): upload media during worker ingestion

hassiebp:
## Motivation

Raw OTEL spans can contain inline base64 media. Unlike Langfuse SDK ingestion, raw OTEL ingestion has no SDK-side media upload step, so large media blobs would otherwise be serialized, queued, and persisted inside event payloads. This increases ingestion latency, worker memory/CPU usage, and ClickHouse event size.

This PR adds a minimal, opt-in worker fallback that extracts embedded media from normalized OTEL events, uploads it through the existing media service, and persists a compact Langfuse media reference instead.

Linear: [LFE-10952](<https://linear.app/langfuse/issue/LFE-10952/extract-media-from-otel-spans-during-worker-ingestion>)

## Approach

* Disabled by default behind `LANGFUSE_OTEL_MEDIA_UPLOAD_ENABLED`.
* Runs after OTEL normalization, only when the batch is written through the native direct events-table path.
* Processes each normalized direct event in place and inspects its non-null `input`, `output`, and `metadata` fields.
* Detects embedded `data:*;base64,...` URIs and provider-specific Anthropic, Vertex, Gemini, AI SDK v6, and AI SDK v7 media shapes.
* Uploads candidates sequentially through the shared media service, which provides content-hash deduplication and trace/observation linking.
* Replaces content only after a successful upload/link. Invalid media or upload failures leave the original value in place and do not fail ingestion.
* Yields to the event loop before decoding each detected media candidate; no yield or upload work occurs when no candidate is found.

```mermaid
flowchart TD
    A["Normalized OTEL batch"] --> B{"Native direct events-table write?"}
    B -- "No: legacy or eval-only" --> C["No media processing"]
    B -- "Yes" --> D{"Feature enabled?"}
    D -- "No" --> C
    D -- "Yes" --> E["Keep events with input, output, or metadata"]
    E --> F["Process each field in place"]
    F --> G["Create and persist event records"]
```

```mermaid
flowchart TD
    A["Direct-event field: input, output, or metadata"] --> D{"Value shape"}

    D -- "Ordinary string" --> E["Scan once for ;base64,"]
    E --> F["No marker: return unchanged"]
    E --> G["Marker: forward-scan Data URIs"]

    D -- "JSON-looking string" --> H["Check complete provider-specific key combination"]
    H --> I["No match: return without JSON.parse"]
    H --> J["Match: JSON.parse and traverse"]

    D -- "Object or array" --> K["Iterative traversal, maximum depth 10"]
    K --> L["Inspect string leaves and provider media shapes"]

    G --> M{"Valid media candidate?"}
    J --> M
    L --> M
    M -- "No" --> N["Keep original value"]
    M -- "Yes" --> O["Yield, decode, hash, and upload sequentially"]
    O --> P["Replace with Langfuse media reference after success"]
```

## Predicted performance impact

The feature is opt-in and limited to native direct writes. For legacy `ingestionEvents`, eval-only batches, and any batch with the flag disabled, the added cost is only the existing direct-write decision plus one flag/branch check; media processing, payload scans, child spans, and media metrics are not invoked.

For native direct writes with the flag enabled, the worker makes one linear pass over the event array to discard events without any of the three media-bearing fields. It allocates one filtered array but no per-field target objects. Processing cost is then linear in the normalized fields inspected:

| No-media field shape | Work performed | Warm local benchmark per 1 MiB field |
| -- | -- | -- |
| Plain string | One full `;base64,` marker scan plus constant-time prefix checks | \~17 us |
| Generic JSON string | Marker scan plus provider-key pre-detection; no parsing | \~82 us |
| JSON string with common `data` key only | Additional key checks; `data` alone does not trigger parsing | \~119 us |
| Structured value with 100 child objects | Full iterative traversal; each string leaf is byte-counted and marker-scanned | \~163 us |
| Full provider-shape false positive | `JSON.parse` plus structured traversal | \~618 us |

Measurements are a warm Node.js 24.6 microbenchmark of the exact `transformMediaPayload` function, reported as the median of seven runs with 300 calls each. They exclude target construction and tracing/metric-library overhead and are intended as order-of-magnitude estimates, not production latency guarantees.

Expected common case:

* Langfuse SDK-originated OTEL `input` and `output` are normally strings, so each takes the cheap single-scan path.
* Normalized `metadata` is a structured object containing user, resource, scope, and retained span attributes. It is traversed, but is normally shallow and small.
* Large structured input/output is mainly produced by third-party normalization paths such as Genkit, Flue, GenAI span events, Logfire, TraceLoop, and OpenInference.
* A direct event is processed exactly once before `createEventRecord` serializes it for ClickHouse. If the same event is also used for evaluations, the evaluation scheduler receives that already-processed record without a second media pass.
* Legacy `ingestionEvents` and eval-only batches intentionally do not process media in V1.
* There is intentionally no in-memory upload cache; durable media deduplication happens in the shared media service.
* If media is detected, base64 decoding, hashing, database linking, and S3 upload dominate detector overhead. Candidates are processed sequentially to bound peak decoded-media memory.

Each enabled processing pass creates one OTEL child span and records aggregate detection/upload metrics. Structured values and strings that enter a detection path additionally record checked-byte metrics, allowing CPU/event-loop changes to be correlated with detector workload.

## Observability

Existing raw-span size observability remains available through:

* `langfuse.ingestion.otel.event.byte_length`
* `langfuse.ingestion.otel.oversized_span`

Media processing adds outcome, byte-length, bytes-removed, detection-path, checked-byte, batch-size, and processing-duration telemetry. No Datadog monitor is added in this repository.

## Explicit non-goals

* Media/file size caps.
* Whole-field attachment fallback.
* Detection in the web container or rewriting the raw OTEL S3 object.
* Worker threads or parallel media decoding/uploads.

## Impacted packages

* `@langfuse/shared`: shared media detection, upload, MIME/extension, and OTEL processing helpers.
* `worker`: opt-in direct-event media processing and telemetry.
* `web`: consumes the shared media helpers instead of maintaining a separate implementation.

## Verification

* `pnpm --filter @langfuse/shared run test MediaPayloadProcessor.test.ts OtelMediaProcessor.test.ts mediaService.test.ts` — `Tests  27 passed (27)`
* `pnpm --filter worker run test processOtelMedia.test.ts scheduleObservationEvals.test.ts` — `Tests  19 passed (19)`
* `pnpm exec dotenv -e .env.dev.example -- pnpm --filter web run test media.servertest.ts`
* `pnpm run lint` — `Tasks: 7 successful, 7 total`
* `pnpm run typecheck` — `Tasks: 7 successful, 7 total`
* GitHub CI on the current PR head.


packages/shared/src/server/otel/OtelMediaProcessor.ts
```diff
@@ -0,0 +1,652 @@
...
+  "traceloop.entity.output",
+  "output.value",
+  "final_result",
+  "output",
+  "gen_ai.output.messages",
+  "gen_ai.tool.call.result",
+  "genkit:output",
+  "tool_response",
+]);
+
```
hassiebp, line 11: We are missing handling for media in langfuse metadata keys

packages/shared/src/server/otel/OtelMediaProcessor.ts
```diff
@@ -0,0 +1,652 @@
...
+async function processMediaString(
+  value: string,
+  context: ProcessContext,
+): Promise<string> {
+  if (isMediaReference(value)) return value;
+
+  let valueWithMediaReferences = value;
+  const dataUris = [...new Set(value.match(DATA_URI_PATTERN) ?? [])];
+  for (const dataUri of dataUris) {
```
hassiebp: From what I currently see, we're running the data URI pattern on every attribute value that is a string or is inside of a list of strings. This can be potentially very expensive on the event loop. Can we do an optimization here where we narrow down the candidates before we run regex on them? Discuss our options.

packages/shared/src/server/otel/OtelMediaProcessor.ts
```diff
@@ -0,0 +1,652 @@
...
+}
+
+async function processMediaString(
+  value: string,
+  context: ProcessContext,
+): Promise<string> {
+  if (isMediaReference(value)) return value;
+
+  let valueWithMediaReferences = value;
+  const dataUris = [...new Set(value.match(DATA_URI_PATTERN) ?? [])];
```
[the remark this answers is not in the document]
hassiebp: Agreed — this is a valid finding. An `includes("data:")` guard would reduce average work, but it would not remove the worst-case behavior on attacker-controlled strings. I propose replacing the global regex with a linear scanner that locates `data:`, validates the MIME and `;base64,` delimiter, then advances over base64 characters once. That preserves embedded data-URI support without regex backtracking. I would add an adversarial repeated-`data:` regression test with the implementation.

packages/shared/src/server/otel/OtelMediaProcessor.ts
hassiebp: These keys are currently handled implicitly: `mediaFieldForAttribute` maps every key that is not classified as input/output to `metadata`, so `langfuse.trace.metadata` and `langfuse.observation.metadata` are processed today. However, that fallback is too broad and makes the behavior unclear. I agree we should make metadata explicit by adding the Langfuse metadata keys/prefixes (including the aliases recognized by the ingestion processor), returning `undefined` for unrelated attributes, and adding top-level plus prefixed-metadata tests.

packages/shared/src/server/otel/OtelMediaProcessor.ts
hassiebp: Yes. I see three options:

1. Add a cheap marker guard before the regex. This improves the common case, but still visits every string and does not fix CodeQL's worst case.
2. Explicitly route only known input/output keys plus explicit metadata keys/prefixes, then use cheap marker checks inside those values. This is my recommendation: unrelated attributes become O(1) key checks and are never scanned or parsed.
3. Replace the generic walker with provider-specific path handlers. This can be fastest, but duplicates more SDK schema knowledge and is brittle as provider shapes evolve.

I would combine option 2 with a linear data-URI scanner. Separately, the JSON transform can be reduced to one async walker with a mutable `changed` flag and a small declarative table for Anthropic/Vertex/AI SDK shapes. That removes the `WeakSet` (JSON.parse cannot produce cycles), `TransformResult`, and the current transform/structured/replace layering while retaining Python/JS SDK coverage and fail-open replacement.

packages/shared/src/server/otel/OtelMediaProcessor.ts
hassiebp: Implemented in b6bd9e012. Metadata detection is now explicit for `langfuse.trace.metadata`, `langfuse.observation.metadata`, `langfuse.metadata`, and `ai.telemetry.metadata`, including dotted child keys. Unrelated attributes are skipped and covered by regression tests.

packages/shared/src/server/otel/OtelMediaProcessor.ts
hassiebp: Implemented in b6bd9e012. The hot path now classifies the attribute key before inspecting its value, skips unrelated attributes, and delegates relevant strings to a focused payload transformer. The OTEL processor dropped from 652 to 300 lines, and duplicate embedded data URIs are processed only once.

packages/shared/src/server/otel/OtelMediaProcessor.ts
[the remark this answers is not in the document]
hassiebp: Implemented in b6bd9e012. The global data-URI regex is gone; detection now uses a linear character scanner with explicit base64 validation. An adversarial regression test covers 10,000 repeated `data:` prefixes.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -318,6 +322,43 @@ export const otelIngestionQueueProcessorBuilder = (
...
+                telemetrySdkLanguage: null,
+              };
+        const mediaTargets = createOtelMediaTargets({
+          ingestionEvents: events,
+          eventInputs,
+        });
+
+        // Media extraction only needs normalized fields. Drop the expanded raw
+        // OTEL tree before decoding or uploading potentially large payloads.
+        parsedSpans = [];
```
hassiebp: I don't like this optimization, limited upside and mutating

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -318,6 +322,43 @@ export const otelIngestionQueueProcessorBuilder = (
       });
       const events: IngestionEventType[] =
         await processor.processToIngestionEvents(parsedSpans);
+      let eventInputs:
```
hassiebp: why are we not calling processToEvents here already?

worker/src/queues/otelIngestionQueue.ts
hassiebp: Addressed in add73700c. The `parsedSpans = []` mutation was removed; the raw parsed spans remain available for the later SDK-info check.

worker/src/queues/otelIngestionQueue.ts
hassiebp: Addressed in add73700c. `processToEvent(parsedSpans)` is now called immediately after `processToIngestionEvents(parsedSpans)`, before optional media processing, and the lazy optional `eventInputs` path was removed.

packages/shared/src/server/media/mediaService.ts
```diff
@@ -0,0 +1,266 @@
...
+    contentType,
+    contentLength: contentBytes.length,
+  });
+  await linkMediaToTraceOrObservation({
+    projectId,
+    traceId,
+    observationId,
+    mediaId,
+    field,
+  });
```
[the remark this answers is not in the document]
hassiebp: Fixed in 75ce38990. New media is now linked only after the S3 upload and successful status update, so upload failures leave no trace/observation link. The regression test asserts both failure behavior and success call ordering.

packages/shared/src/server/media/mediaService.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  const uploadStartedAt = Date.now();
+  await getS3MediaStorageClient(mediaBucket).uploadFile({
+    fileName: bucketPath,
+    fileType: contentType,
+    data: Readable.from([contentBytes]),
+  });
```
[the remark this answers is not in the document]
hassiebp: Fixed in 75ce38990. `linkMediaToTraceOrObservation` now runs only after S3 upload and the successful media status update. On S3 failure the original error is rethrown and no link row is created; covered by a regression test.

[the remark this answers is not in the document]
hassiebp: Fixed in 75ce38990. The worker upload path now defers linking until after S3 upload and status persistence succeed, eliminating the permanently dangling relation on fail-open ingestion. The failure regression test verifies that no link or success update occurs.

packages/shared/src/server/media/mediaService.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+    throw error;
+  }
+
+  recordIncrement("langfuse.media.upload_http_status", 1, {
+    status_code: 200,
+  });
+  recordHistogram("langfuse.media.upload_time_ms", uploadTimeMs, {
+    status_code: 200,
+  });
```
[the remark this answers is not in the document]
hassiebp: Fixed in 75ce38990. S3 upload exceptions now emit `langfuse.media.upload_http_status` and `langfuse.media.upload_time_ms` using the AWS HTTP status when available and `500` as the fallback, then rethrow. Covered with a 503 regression test.

packages/shared/src/server/media/MediaPayloadProcessor.ts
```diff
@@ -0,0 +1,528 @@
...
+      if (value.charCodeAt(headerCursor) === 59) break;
+      headerCursor += 1;
+    }
+
+    if (headerCursor === value.length) break;
+    const semicolon = headerCursor;
+    if (!value.startsWith(BASE64_MARKER, semicolon)) {
+      cursor = semicolon + 1;
+      continue;
+    }
```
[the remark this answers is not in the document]
hassiebp: Fixed in 75ce38990. The linear scanner now continues through media-type parameters until `;base64,`, while retaining comma termination and nested `data:` restart behavior. A `data:text/plain;charset=utf-8;base64,...` regression test verifies the normalized `text/plain` candidate.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -318,6 +322,28 @@ export const otelIngestionQueueProcessorBuilder = (
...
+      // This also records original OTEL span-size telemetry before media
+      // references replace any inline data.
+      const eventInputs = processor.processToEvent(parsedSpans);
+      const mediaUploadEnabled =
+        env.LANGFUSE_OTEL_MEDIA_UPLOAD_ENABLED === "true";
+
+      if (mediaUploadEnabled) {
+        const mediaTargets = createOtelMediaTargets({
+          ingestionEvents: events,
+          eventInputs,
```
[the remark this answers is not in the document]
hassiebp: Fixed in 198a16833. Media target selection is now split by actual persistence destination: normalized ingestion events are processed only when legacy tables are written, and `eventInputs` are processed only when the direct `events_full` write is active. Legacy-only eval processing no longer scans/uploads or links discarded event-input payloads. Added focused coverage showing legacy and direct-event targets can be selected independently; worker media tests pass (5/5).

packages/shared/src/server/media/MediaPayloadProcessor.ts
```diff
@@ -0,0 +1,542 @@
...
+
+  if (isMediaReference(value)) return { value, bytesRemoved: 0 };
+
+  if (value.includes(BASE64_MARKER)) {
+    const withDataUriReferences = await replaceDataUris(value, params);
+    if (
+      withDataUriReferences.value !== value ||
+      value.includes(DATA_URI_PREFIX)
+    ) {
+      return withDataUriReferences;
```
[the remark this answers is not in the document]
hassiebp: Fixed in 198a16833. After Data URI replacement, stringified JSON now continues into the cheap provider-shape check and structured transformation instead of returning early. A regression test with both a Data URI and an Anthropic-style raw-base64 object verifies both are replaced; shared media tests pass (14/14).

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -510,6 +533,22 @@ export const otelIngestionQueueProcessorBuilder = (
         return;
       }
 
+      if (mediaUploadEnabled && shouldWriteToEventsTable) {
```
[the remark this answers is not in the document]
hassiebp: Good catch — observation eval scheduling serializes `eventInputs` to eval S3 even without a direct ClickHouse write, so the old gate left inline media in that consumed representation. Fixed in d13636271 by processing event inputs when either eval configs or direct persistence consume them. Added a four-case decision test; worker media tests pass (9/9), and lint/typecheck both pass (7/7).

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -510,6 +535,30 @@ export const otelIngestionQueueProcessorBuilder = (
...
       }
 
+      // eventInputs are a separate normalized representation consumed by both
+      // direct persistence and observation evals, which upload it to eval S3.
+      if (
+        shouldProcessOtelEventInputMedia({
+          enabled: mediaUploadEnabled,
+          hasEvalConfigs,
+          shouldWriteToEventsTable,
+        })
```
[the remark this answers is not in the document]
hassiebp: Fixed in e1f28153e. Direct writes still prepare the normalized batch eagerly, while eval-only observations now use an injected prepareObservationForUpload hook that runs only after executability, filter, and sampling checks find a match and immediately before the S3 upload. Added regression tests proving preparation runs for a matching config and is skipped when no config matches. Validation: 27 tests passed; lint and typecheck each report 7 successful tasks.

worker/src/features/otel-media/processOtelMedia.ts
```diff
@@ -0,0 +1,217 @@
...
+ * directly and therefore needs media replacement before event records are
+ * created. Eval-only payloads are prepared later, after filter and sampling
+ * checks confirm that the observation will be uploaded.
+ */
+export function shouldProcessOtelEventInputMedia(params: {
+  enabled: boolean;
+  shouldWriteToEventsTable: boolean;
+}): boolean {
+  return params.enabled && params.shouldWriteToEventsTable;
+}
```
Steffen911: Drop

worker/src/features/otel-media/processOtelMedia.ts
```diff
@@ -0,0 +1,217 @@
...
+      continue;
+    }
+
+    addMediaTargets({
+      targets,
+      body,
+      traceId,
+      observationId,
+    });
+  }
```
Steffen911: This should also go away then, right?

worker/src/features/otel-media/processOtelMedia.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  return targets;
+}
+
+/**
+ * Builds media targets for one already-normalized body. The function itself
+ * does not mutate `body`; returned targets retain its reference, so processing
+ * those targets later mutates the body's input, output, and metadata fields.
+ */
+export function createOtelMediaTargetsForBody(params: {
```
Steffen911: Drop

worker/src/features/otel-media/processOtelMedia.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/**
+ * Runs instrumented OTEL media processing when configured.
+ *
+ * Successful replacements mutate each target body in place. Missing storage
+ * configuration or unexpected processing errors are logged and swallowed so
+ * media extraction cannot reject the enclosing OTEL ingestion job.
+ */
+export async function processOtelMediaIfEnabled(params: {
```
Steffen911: I think having a singleton here with an enabled true/false and a projectId constructor might be preferable unless this is the only external entrypoint.

worker/src/features/otel-media/processOtelMedia.ts
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Builds media targets for one already-normalized body. The function itself
+ * does not mutate `body`; returned targets retain its reference, so processing
+ * those targets later mutates the body's input, output, and metadata fields.
+ */
+export function createOtelMediaTargetsForBody(params: {
+  body: Record<string, unknown>;
+  traceId: string;
+  observationId?: string;
```
Steffen911: I would also track projectIds here for tenant isolation.

worker/src/features/otel-media/processOtelMedia.ts
Steffen911: That way we can also check if media is configured and may return early on everything else.

packages/shared/src/server/media/MediaPayloadProcessor.ts
```diff
@@ -0,0 +1,584 @@
+import { isMediaContentType, type MediaContentType } from "../../domain/media";
```
Steffen911, line 1: I have no idea what happens in this entire code block. Seems either like it's really complicated or obscurely designed. I'd trust that it makes sense to you, is well tested, and can be easily disabled or refactored if it becomes necessary.

Having a high level overview or description on how processOtelMedia / OtelMediaProcessor and MediaPayloadProcessor share their work would be appreciated, because for me the naming is too close to understand whether the separation is arbitrary or has an actual foundation.

packages/shared/src/server/otel/OtelMediaProcessor.ts
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * Detects and uploads embedded media for a set of normalized OTEL fields.
+ *
+ * Targets and candidates are processed sequentially to bound peak decoded
+ * media memory. Successful replacements mutate each target's `body` in place.
+ * Invalid values and failed uploads are left unchanged; their outcomes are
+ * reflected in the returned counters instead of failing the whole run.
+ */
+export async function processOtelMedia(params: {
```
Steffen911, line 83: The projectId isolation seems right to me. As there is no permanent state in all components, I don't see how we could arrive at cross-project leakage of media assets. So this part should be good.

worker/src/features/otel-media/processOtelMedia.ts
hassiebp: Refactored to only one caller, thus not yet deploying singleton

packages/shared/src/server/media/MediaPayloadProcessor.ts
hassiebp: Probably both - the detection logic is indeed quite complicated as I decided to roll a linear parser instead of relying on regex. 

I added a better description here. Also, by dropping the `ingestionEvents` path, the logic has simplified a lot: https://github.com/langfuse/langfuse/pull/15120/commits/0885b68e41889ba64392ea4a71a98e01510c157f

PR #15123, opened 2026-07-15
fix(trace): keep tree and timeline layouts readable on extremely deep traces (LFE-10959)

nkabardin:
## TL;DR

A customer trace with ~1,400 observations chained into a single parent line (every LLM call parented onto the previous call) collapsed both trace-detail layouts: tree rows were pushed tens of thousands of pixels off-viewport and the timeline gutter + time axis degenerated into unreadable clutter. Indentation is now capped to the available width (deeper rows render flat at the cap), the time axis picks time-nice steps up to multi-day scales, and hour-scale tick labels read as durations. Shallow traces render exactly as before.

## Why

A customer reported an "everything is basically broken" trace: ~1,400 sequential generations over ~2.8h, mis-instrumented so each call is the sole child of the previous one — tree depth 1,400. Both layouts assume bounded depth:

- **Tree**: 20px indent per level, uncapped → rows offset up to ~28,000px; flex crushed row content into 1-character vertical strips, and deeper rows rendered as a completely blank pane.
- **Timeline gutter**: 14px/level inside a 160–560px overflow-hidden pane → names truncated into nothing after ~14 levels.
- **Time axis**: step sizes clamp at 500s, so a ~10,000s trace rendered ~21 tick labels ~45px apart, overlapping into garbage.

## What

- New shared `computeMaxVisualDepth` helper (`_shared/visual-depth.ts`): converts the measured container/gutter width into a maximum *visual* depth, keeping a reserved content width free. Rows past the cap render flat at the cap level; connector spines that would dangle at the cap are suppressed. The cap only bites when indentation would eat into the content reserve — everyday traces are untouched.
- Tree: `VirtualizedTree` measures its container (ResizeObserver, stores the derived integer cap) and passes the cap through `TreeNodeMetadata`; the node wrapper clamps indentation and connectors.
- Timeline: the cap derives from the (resizable) gutter width; widening the gutter reveals more structure.
- Axis: `PREDEFINED_STEP_SIZES` extended with time-nice steps (10m, 15m, 20m, 30m, 45m, 1h … 24h) plus a whole-day fallback beyond that, so tick spacing never drops below the 100px design grid for any duration; ticks ≥ 60s format via `formatIntervalSeconds` ("25m 00s", "1h 30m 00s") instead of "1500.00s".

## Result

At depth 1,400, both views stay fully readable and navigable at any scroll position: tree rows show name/latency/∑-duration/tokens/cost everywhere, gutter names survive, and the axis shows ~9 clean duration ticks. Normal traces are pixel-identical to before.

<details>
<summary>Reproduction, mechanism, verification</summary>

**Repro**: `pnpm run seed -- deep-chain --v4` (new seeder scenario added in this PR) reconstructs the reported shape — 1,401 sequential generations in one parent chain, matching timings (~5.7s p50 calls, ~0.8s gaps, ~2.8h total), usage and cost distributions — then open the seeded trace in tree and timeline view.

**Caps**: tree `{indent 20px, reserve 220px, min 8, max 32}`, gutter `{indent 14px, reserve 120px, min 6, max 32}`. Stored as the derived integer cap so resize only re-renders when the cap actually moves. Unmeasured containers fall back to the max cap (still bounded, never the legacy unbounded behavior).

**Connector semantics at the cap**: a capped node's children render at the same indent (not one level right), so the parent's downward child-spine is suppressed there; ancestor rails are sliced to the visual depth. Below the cap nothing changes.

**Also fixes** the per-row DOM cost at extreme depth: previously each visible row emitted one indent div per ancestor level (~1,400 divs × ~40 visible rows); now bounded by the cap (≤ 32).

**Verification**: failing tests first (`calculateStepSize` previously returned 500s for a 10,000s trace → ~45px tick spacing), then fix; `Tests  52 passed (52)` across `timeline-calculations.clienttest.ts` + new `visual-depth.clienttest.ts`; web `typecheck` and `eslint` clean. Verified live in Firefox (the reporting browser) against the seeded depth-1,401 trace — tree top/deep-scroll, timeline top/deep-scroll — and against a normal 60-observation branching trace (tree + timeline) to confirm zero visual regression.

</details>

This PR fixes catastrophic layout breakage in trace views when a trace has extremely deep nesting (e.g., 1,400 sequential generations each parented on the previous). Both the tree view and timeline gutter were unbounded in their indentation, causing content to render tens of thousands of pixels off-screen. The time axis also had no step sizes above 500 s, producing ~21 overlapping tick labels on a 2.8-hour trace.

- **Visual depth cap** (`visual-depth.ts`): a new `computeMaxVisualDepth` helper converts available container width into a max visual depth; both the tree (`VirtualizedTree`/`VirtualizedTreeNodeWrapper`) and timeline gutter (`TimelineGutterRow`) now clamp indentation and suppress dangling child spines at the cap, keeping row names readable at any real depth.
- **Time axis** (`timeline-calculations.ts`): `PREDEFINED_STEP_SIZES` extended with time-nice entries from 10 min to 24 h, plus a whole-day fallback for multi-day traces; tick labels ≥ 60 s now format as human durations ("25m 00s", "1h 30m 00s") via `formatIntervalSeconds`.
- **Seeder scenario** (`deep-chain.ts`): new seed scenario reproduces the reported 1,401-generation chain, with statistically faithful timings, usage, and cost distributions, to support manual regression testing.
</details>

Safe to merge — the changes are self-contained layout fixes with no backend or data-model impact, and tests confirm the corrected step-size and depth-cap behaviour.

The depth-capping, connector-suppression, and step-size extension logic all check out. The only note is a pre-existing duplicate div in the root-node connector section that has no visual effect and is not introduced by this PR. The seeder scenario accurately reconstructs the reported failure shape. Both tree and timeline paths are covered by new unit tests.

No files require special attention; the one flagged item in VirtualizedTreeNodeWrapper.tsx is harmless and pre-dates this change.
</details>

`````markdown

---

### Issue 1 of 1
web/src/components/trace/components/_shared/VirtualizedTreeNodeWrapper.tsx:142-145
**Duplicate root-node downward connector (pre-existing)**

The "Root node downward connector" block is a strict subset of the general child-spine block directly above it: when `depth === 0`, `childrenAreCapped = (0 >= maxVisualDepth)` is always false (since `maxVisualDepth >= minDepth >= 6`), so both conditions fire together for the root node and produce two identical absolutely-positioned `div`s overlapping at the same coordinates. There's no visible rendering bug because they stack on top of each other, but it's redundant DOM. This `!childrenAreCapped` guard was added consistently to both by this PR, which is correct — a light follow-up could simply remove the second block.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): keep tree and timeline layou..."](https://github.com/langfuse/langfuse/commit/9ce4d533602a68cf226724a84e7c34d3d5ef9185) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44490107)</sub>


nkabardin:
@claude review

PR #15124, opened 2026-07-15
fix(sessions): bound trace-card I/O in the session detail view (LFE-10958)

nkabardin:
## TL;DR

Session detail cards fetched and rendered **every observation's full I/O** — customers reported multi-GB browser memory and frozen/never-loading session pages. Cards are now bounded previews: fields within 300K chars render exactly as before; anything larger ships a 4K preview head with its true size and two escape hatches ("Open in trace view" at that observation, "Download I/O" without rendering). A card can no longer grow with payload size.

## Why

The v4 session detail view defaults to the "All observations with I/O" preset. Its per-trace card query (`sessions.observationsForTraceFromEvents`) ran with `truncated: false`, so every scrolled trace pulled full input/output/metadata for all its observations and rendered everything inline through the pretty view. Customers with large payloads reported 5–8 GB browser memory and session pages that freeze or never load; a chronic Sentry issue (`[useParsedObservation] Worker error`, hundreds of users) sits on the same render path.

Extending the virtualized `AdvancedJsonViewer` into cards — the direction PR #11010 deferred — turns out to be architecturally incompatible: the virtualized viewer must own a fixed-height scroll container, while session cards must stay content-height because the outer list virtualizer measures them. Bounding what a card fetches and renders is the compatible fix.

## What

- **Size-capped I/O select** (ClickHouse `events_full`): fields ≤ 300K chars return whole — this matches the pretty view's existing client-side parse cap (`deepParseJson` maxSize), so under-cap cards are behavior-identical. Larger fields return a 4K head plus true `inputLength`/`outputLength` and truncation flags. Metadata values are capped with the same policy (`metadataTruncated` flag).
- **Per-card bounds**: at most 50 observations per card (notice + trace link when more exist) and a 2M-char cumulative I/O budget so many just-under-cap observations can't add up to an unbounded response.
- **Truncated observation UI**: bounded monospace preview + true size ("535.9K characters — showing the first 4K") + "Open in trace view" (opens the peek at that exact observation, where the virtualized JSON viewer and worker parse live) + "Download I/O" (fetches the raw payload via a new session-scoped procedure and saves it to a file without rendering).
- **New `sessions.observationFullIOFromEvents`**: session-authorized (public sessions keep working, unlike trace-authorized `events.batchIO`); the query scopes by `session_id` so a session grant can't read observations outside that session.
- **Dedupe hardening**: the synthetic trace-row dedupe now also compares true lengths, so equal preview heads of different payloads never false-dedupe.

## Result

Measured locally (seeded session, 120 traces with ~0.3–1.6 MB payloads each, identical scroll journey, Chromium via CDP):

| | before | after |
|---|---|---|
| I/O transferred per card | ~800 KB | ~212 KB (bounded; whales cap at 4K heads) |
| long-task time per card | ~109 ms | ~34 ms |
| worst case per card | grows with payload | **fixed ceiling** (~2M chars) |

Sessions with small/normal payloads are pixel-identical (verified: zero truncated blocks, pretty chat renders as before). Firefox pass: scrolling the pathological session stays responsive (5 ms rAF gap), no page errors.

<details>
<summary>Engineering details</summary>

**Query layer** — `EventsQueryBuilder.selectIOWithSizeCap(inlineChars, previewChars)` emits `if(lengthUTF8(e.input) <= N, e.input, leftUTF8(e.input, P))` plus `lengthUTF8()` length columns. `lengthUTF8` is computed in the query rather than reading the materialized `input_length` columns so no assumption is made about every deployment's `events_full` schema (the full column is read by this select anyway — the win is response size/network/browser memory, not ClickHouse-side I/O). Size-capped mode forces `events_full` (true lengths + full under-cap values) and survives the `positionInTrace` CTE path, which only filters span ids.

**Length zip** — lengths are attached to converted observations **by row order**, never by id: un-merged ReplacingMergeTree versions of one span can return multiple rows sharing an id, and an id-keyed zip could attach one version's lengths to another version's payload.

**Budget trim** — runs in the router over the returned rows (order-stable, startTime ASC); the budget check runs before adding a row's own size, so the first observation always keeps whatever the per-field cap allowed. Trimmed rows get their flags set and keep their true lengths.

**Download** — plain tRPC client call (no React Query cache) so the full payload isn't retained after saving; I/O is embedded as raw strings because parsing multi-MB JSON just to pretty-print it risks the same freeze the card avoids. The repository query bounds the read with ±1s around the observation's `start_time` (primary-key pruning) and `ORDER BY event_ts DESC LIMIT 1` for a deterministic latest version.

**Peek deep-link** — the session page's peek config mirrors `row.observationId` into the `?observation=` param (read by the trace detail's SelectionContext); the annotation queue's new-tab `openPeek` appends the same param.

**Annotation queue** — uses the same cards (`hideTracePanel`); truncated observations there open the trace page in a new tab (no in-page peek exists in that flow).

**Tests** — query-builder SQL unit tests; servertests against real ClickHouse covering full/preview fields, true lengths, metadata capping, the 50-observation page + `hasMoreObservations`, the cumulative budget trim, and the session-scoping of the full-I/O procedure (wrong session ⇒ NOT_FOUND); client tests for the card's render decision, peek handoff, and download flow.

**Verification** — `pnpm run lint` (7/7), `pnpm run typecheck` (7/7), targeted vitest suites green (5+5+6 new, 5+4 existing session servertests), before/after browser measurements via Playwright + CDP (heap, longtasks, response bytes), UI pass in Chromium + Firefox with screenshots.

**Out of scope / follow-ups** — the legacy (non-v4) `SessionPage` still fetches full trace-level I/O via `traces.byId`; the trace detail's own large-payload rendering limits are tracked separately (LFE-10152). A pre-existing off-by-one in the `positionInTrace` ranking CTE (un-merged span versions shift `ROW_NUMBER` ranks; needs dedup before ranking, not after) surfaced during review and is deferred as a follow-up.

</details>

This PR fixes runaway browser memory and frozen session pages by capping what the `observationsForTraceFromEvents` card query returns: fields ≤ 300K chars come back whole; larger fields are served as a 4K preview head with true size metadata and two escape hatches (trace-view peek, raw download). A new `observationFullIOFromEvents` procedure handles session-scoped full-payload downloads without caching them in React Query.

- **ClickHouse layer**: `EventsQueryBuilder.selectIOWithSizeCap` emits conditional `if(lengthUTF8 ≤ N, full, leftUTF8(preview))` for I/O and metadata, with true length columns returned for UI formatting.
- **Router layer**: adds a 50-observation per-card limit (plus a `hasMoreObservations` flag) and a 2M-char cumulative budget trim to prevent many just-under-cap fields from adding up.
- **UI layer**: `SessionObservationIO` routes observations through `IOPreview` when within cap, and a bounded monospace preview with escape-hatch buttons when server-truncated.
</details>

Safe to merge for the common case (small/normal payloads, which are pixel-identical to before), but the 2M cumulative budget trim — the secondary safety net for many near-cap observations — does not engage for valid JSON payloads, which are precisely the payloads that cause the memory problem this PR targets.

The per-field 300K inline cap is enforced correctly at the ClickHouse layer and bounds worst-case card size to ~30MB (50 obs × 2 fields × 300K), a dramatic improvement over multi-GB. But the router's cumulative budget check gates on typeof input === 'string', and shouldJsonParse: true converts valid JSON strings to objects before that check runs — so the budget accumulates 0 for every observation with parseable JSON, and the trim never fires. The existing server test covers only non-JSON repeat-char inputs and therefore misses this. For customers whose sessions consist of standard LLM traces (JSON chat arrays, tool calls), the budget provides no protection.

web/src/server/api/routers/sessions.ts (budget trim logic at lines 884–920); the corresponding server test in web/src/__tests__/server/sessions-observations-io-cap.servertest.ts should cover valid JSON inputs to verify the budget actually fires.
</details>

`````markdown

---

### Issue 1 of 3
web/src/server/api/routers/sessions.ts:884-920
**Budget trim silently no-ops for valid JSON payloads**

The cumulative budget check and the per-observation trim both gate on `typeof observation.input === "string"`, but this query runs with `shouldJsonParse: true`. `enrichObservationsWithModelData` calls `convertEventsObservation` → `renderIO` → `parseJsonPrioritised`, which parses any valid JSON string into a JS object. When that happens, `typeof input` is `"object"`, so `returnedIOChars = 0` for the observation and `inputSliced = false` — the budget accumulates nothing and the slice never fires.

In the primary real-world scenario (LLM traces with large JSON chat arrays or tool calls), all 50 near-cap observations slip through the budget check uncounted and unsliced, and the card response can reach 50 × 600K ≈ 30MB instead of the intended 2M-char ceiling. The existing server test at line 637 avoids this because it seeds `"b".repeat(290_000)`, which is not valid JSON and therefore comes back as a raw string.

### Issue 2 of 3
packages/shared/src/server/repositories/events.ts:548-564
**Zip-by-index relies on an undocumented 1:1 contract across two enrichers**

`enriched[i]` is zipped with `observationRecords[i]` on the assumption that both `enrichObservationsWithModelData` and `enrichObservationsWithTraceFields` are strict 1:1 maps with no reordering or filtering. The comment explains the reasoning, but neither function enforces or documents this guarantee. If either is ever changed to deduplicate, sort, or skip certain records, the wrong `inputLength`/`outputLength` values would silently attach to the wrong observations, producing incorrect truncation flags with no runtime error.

### Issue 3 of 3
web/src/components/session/TraceEventsRow.tsx:1081-1083
**Hardcoded display count will drift from the router's authoritative limit**

`SESSION_CARD_OBSERVATIONS_NOTICE_COUNT = 50` cannot be imported from the router (server/client boundary), so if `SESSION_OBSERVATIONS_PER_TRACE_LIMIT` is changed in `sessions.ts`, the "Only the first 50 observations are shown here" message will show the wrong number. The comment acknowledges this, but it is easy to miss in practice.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(sessions): bound trace-card I/O in t..."](https://github.com/langfuse/langfuse/commit/c21aa7cf148093154f56505950d8e5c28fd5eaf9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44490198)</sub>

### DIFF
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.test.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.test.ts
@@ -40,8 +40,13 @@ describe("EventsQueryBuilder.selectIOWithSizeCap", () => {
     expect(query).toContain(
       "arrayMap(v -> if(lengthUTF8(v) <= 300000, v, leftUTF8(v, 4000)), arrayReverse(e.metadata_values))",
     );
+    // The flag only fires for a key's winning value (a shadowed duplicate
+    // must not raise it), and the shipped weight counts every capped value.
     expect(query).toContain(
-      "arrayExists(v -> lengthUTF8(v) > 300000, e.metadata_values) as metadata_truncated",
+      "arrayExists((v, i) -> lengthUTF8(v) > 300000 AND arrayFirstIndex(n -> n = e.metadata_names[i], e.metadata_names) = i, e.metadata_values, arrayEnumerate(e.metadata_values)) as metadata_truncated",
+    );
+    expect(query).toContain(
+      "arraySum(arrayMap(v -> if(lengthUTF8(v) <= 300000, lengthUTF8(v), 4000), e.metadata_values)) as metadata_length",
     );
     // The default full-value metadata expression must not also be present.
     expect(query).not.toContain(
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.test.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.test.ts
@@ -5,6 +5,57 @@ import {
   EventsQueryBuilder,
 } from "./event-query-builder";
 
+describe("EventsQueryBuilder.selectIOWithSizeCap", () => {
+  const build = () =>
+    new EventsQueryBuilder({ projectId: "test-project" })
+      .selectFieldSet("base", "calculated", "metadata")
+      .selectIOWithSizeCap(300_000, 4_000)
+      .whereRaw("e.trace_id = {traceId: String}", { traceId: "trace-1" })
+      .buildWithParams();
+
+  it("returns full fields under the cap and a preview head above it", () => {
+    const { query } = build();
+
+    // lengthUTF8() is computed in the query on purpose: the materialized
+    // input_length/output_length columns cannot be assumed present on every
+    // deployment's events_full, and the full column is read here anyway.
+    expect(query).toContain(
+      "if(lengthUTF8(e.input) <= 300000, e.input, leftUTF8(e.input, 4000)) as input",
+    );
+    expect(query).toContain(
+      "if(lengthUTF8(e.output) <= 300000, e.output, leftUTF8(e.output, 4000)) as output",
+    );
+  });
+
+  it("exposes the true lengths so callers can detect previews", () => {
+    const { query } = build();
+
+    expect(query).toContain("lengthUTF8(e.input) as input_length");
+    expect(query).toContain("lengthUTF8(e.output) as output_length");
+  });
+
+  it("caps metadata values with the same policy and flags truncation", () => {
+    const { query } = build();
+
+    expect(query).toContain(
+      "arrayMap(v -> if(lengthUTF8(v) <= 300000, v, leftUTF8(v, 4000)), arrayReverse(e.metadata_values))",
+    );
+    expect(query).toContain(
+      "arrayExists(v -> lengthUTF8(v) > 300000, e.metadata_values) as metadata_truncated",
+    );
+    // The default full-value metadata expression must not also be present.
+    expect(query).not.toContain(
+      "mapFromArrays(arrayReverse(e.metadata_names), arrayReverse(e.metadata_values)) as metadata",
+    );
+  });
+
+  it("reads events_full (true lengths + full under-cap values)", () => {
+    const { query } = build();
+
+    expect(query).toContain("FROM events_full");
+  });
+});
+
 describe("buildEventsFullTableSplitQuery", () => {
   const buildBase = () =>
     new EventsQueryBuilder({ projectId: "test-project" })
--- a/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
@@ -1160,9 +1160,9 @@ export class EventsQueryBuilder extends BaseEventsQueryBuilder<
       }
     }
 
-    // Size-capped metadata: same per-value policy as I/O, plus a flag so
-    // callers can surface that values were capped. Mirrors the arrayReverse
-    // (last-wins) semantics of the default metadata expression.
+    // Size-capped metadata: same per-value policy as I/O, plus a truncation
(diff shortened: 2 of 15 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15155, opened 2026-07-16
fix(trace): size-gate the unvirtualized JSON IO view (LFE-10989)

nkabardin:
**TL;DR** — The trace/observation **JSON** IO tab renders the whole payload through an unvirtualized JSON viewer on the main thread. On multi-megabyte IO this freezes the tab for seconds and can crash the renderer. We now size-gate that view: above **2 MB per field** we skip the parse and full render and show a bounded preview + raw download, pointing users to the views that scale. Normal IO is untouched. Part of the large-traces resilience epic (LFE-10152).

**Why** — A large trace's IO (~20 MB input in our large-traces seed) sent the plain JSON view into a multi-second main-thread freeze and, in memory-constrained browsers, a renderer crash. The view parses the payload synchronously and then builds ~one DOM node per JSON token, and the pipeline makes several full O(size) copies (pre-parse + `structuredClone` + internal re-parse + unicode-decode) before rendering. Sibling surface to the already-shipped session-IO bounding (#15124) and the trace-graph budget (#15153).

**What** — A small pure helper decides, from a cheap serialized-length probe, whether a field (input/output/metadata) exceeds a **2 MB** render cap. Over the cap, the JSON view skips both the synchronous parse and the unvirtualized render and instead shows a bounded preview head plus escape hatches: a **raw download** (raw text for string fields, pretty JSON for objects), and a pointer to the **Formatted** (lazy table) and **JSON Beta** (virtualized) views. Under the cap, nothing changes.

**Result** — On the ~20 MB seed, switching to JSON view no longer freezes for seconds or risks a crash in Firefox or Chrome; it shows a bounded card with a working download in well under a second. Verified live in both browsers; gating logic + show-logic unit/component-tested; typecheck + lint clean.

<details><summary>Mechanism, threshold, and verification</summary>

- **Path:** `IOPreview → IOPreviewJSONSimple → PrettyJsonView(currentView="json") → JSONView (react18-json-view)`, unvirtualized. `json-beta` (AdvancedJsonViewer) is already virtualized; this view was not gated. The over-limit path skips `deepParseJson`/`parsePreservingPrecision` (field → `undefined`) **and** renders `LargeJsonFieldFallback` instead of the viewer — both, not one.
- **Threshold — 2 MB/field** (`JSON_VIEW_RENDER_CHAR_LIMIT`, UTF-16 `.length`): ~4–6× the codebase's own safe main-thread parse bounds (`IOPreviewPretty`/`PrettyJsonView` `maxSize` 300–500 KB), still 100–1000× above normal KB-scale IO (never gated), and below the size where the unvirtualized render is perceptibly janky. Large payloads stay reachable via Formatted, JSON Beta, and raw download.
- **Show-logic:** for non-over-limit fields the condition collapses bit-for-bit to the original (`hideIfNull` null still hidden, empty object still shown); an over-limit field with `hideIfNull` now surfaces the fallback (intended). `CorrectedOutputField` gets no baseline for an over-limit output (acceptable degradation, noted in code).
- **Download:** string fields download as raw `text/plain` (`.txt`); objects as `JSON.stringify(v,null,2)` (`.json`) — the huge-IO seed is a base64 string, so raw text avoids over-encoding.
- **Verification:** real login, opened the large-IO trace, switched to JSON view — main-thread heartbeat showed a ~3.2s Firefox / ~1.65s Chrome freeze with the gate off vs. sub-second with it on, no crash, fallback + working raw download in both browsers. 13 unit/component tests (boundary + edge cases + fallback selection + null-hiding + raw-string download).

</details>

This PR size-gates the unvirtualized JSON IO view in `IOPreviewJSONSimple`: fields above a 2 MB character threshold skip `deepParseJson` and the `react18-json-view` render, and instead show a bounded 4000-char preview plus a raw download button via the new `LargeJsonFieldFallback` component.

- **Size probe and parse guard** (`jsonViewSizeGate.ts`, `IOPreviewJSONSimple.tsx`): A cheap `getJsonStringSize` probe (O(1) for strings, O(size) for objects via `JSON.stringify`) is memoized per field; over-limit fields short-circuit `deepParseJson` and route to the fallback. The `hideIfNull` visibility logic is correctly updated so over-limit fields aren't silently dropped.
- **Fallback component** (`LargeJsonFieldFallback.tsx`): Renders a dashed-border card with a truncated preview, a prose hint pointing at Formatted/JSON-Beta views, and a download button that writes strings as raw `.txt` and objects as pretty `.json`.
- **Tests**: 13 unit/component tests cover boundary conditions, null-hiding semantics, string-field raw download, and fallback vs `PrettyJsonView` selection.
</details>

The change is well-scoped and low-risk: the hot path (fields under 2 MB) is untouched, and the over-limit path is an entirely additive fallback with no mutations to shared state.

The implementation is solid for the primary use case (large string payloads). The two minor findings — a redundant full serialization in the preview memo and a silent empty-file download on serialization error — are both on the rare, already-gated path and don't affect correctness for the cases the fix targets.

LargeJsonFieldFallback.tsx — the previewText useMemo and the onDownload error path are worth a second look for large object (non-string) payloads.
</details>

```mermaid
flowchart TD
    A[IOPreviewJSONSimple renders] --> B[useMemo: getJsonStringSize per field]
    B --> C{size > 2 MB?}
    C -- No --> D[effectiveX = parsedX ?? deepParseJson]
    C -- Yes --> E[effectiveX = undefined\nskip deepParseJson]
    D --> F{showInput/Output/Metadata?}
    E --> G{showInput/Output/Metadata?\n inputTooLarge overrides hideIfNull}
    F -- Yes --> H[PrettyJsonView\n react18-json-view]
    G -- Yes --> I[LargeJsonFieldFallback]
    I --> J[safeStringify up to 4000 chars preview]
    I --> K[Download button\nstring → .txt\nobject → pretty .json]
    K --> L[safeStringify full value\ncreate Blob → anchor click]
    I --> M[PostHog capture on download]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[IOPreviewJSONSimple renders] --> B[useMemo: getJsonStringSize per field]
    B --> C{size > 2 MB?}
    C -- No --> D[effectiveX = parsedX ?? deepParseJson]
    C -- Yes --> E[effectiveX = undefined\nskip deepParseJson]
    D --> F{showInput/Output/Metadata?}
    E --> G{showInput/Output/Metadata?\n inputTooLarge overrides hideIfNull}
    F -- Yes --> H[PrettyJsonView\n react18-json-view]
    G -- Yes --> I[LargeJsonFieldFallback]
    I --> J[safeStringify up to 4000 chars preview]
    I --> K[Download button\nstring → .txt\nobject → pretty .json]
    K --> L[safeStringify full value\ncreate Blob → anchor click]
    I --> M[PostHog capture on download]
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/components/trace/components/IOPreview/components/LargeJsonFieldFallback.tsx:43-48
**Full serialization before 4000-char slice**

For non-string fields above the limit, `safeStringify(value)` fully materializes the entire object as a string before slicing to `PREVIEW_DISPLAY_CHARS`. The parent's `useMemo` already called `JSON.stringify` once for the size probe, so this is a second full in-memory serialization of potentially a 20 MB object on the initial render. For the primary use case (base64 string payload) this is O(1) because `isString` is true, but for large JSON-object payloads it allocates two large transient strings back-to-back. Consider either passing the serialized string from the parent's probe down as a prop, or using an early-terminating serializer that stops after `PREVIEW_DISPLAY_CHARS` characters.

### Issue 2 of 2
web/src/components/trace/components/IOPreview/components/LargeJsonFieldFallback.tsx:54-59
**Silent empty download on serialization failure**

When `safeStringify` catches an error (e.g., a non-serializable object that somehow passed the size gate returning 0), it returns `""`, and the download button writes an empty file without any notification to the user. Since `getJsonStringSize` also returns 0 on serialization failure, such a value would never be gated in the first place (it's treated as size 0, so `inputTooLarge` stays false and this fallback is never shown). The empty-download path is therefore unreachable in practice. However, if the guard relationship ever drifts, the UX failure would be invisible. Consider logging a warning or showing an inline error message when `content.length === 0`.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): size-gate the unvirtualized ..."](https://github.com/langfuse/langfuse/commit/c5687923ea04c09785020be5e06fc6e925971a50) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44754477)</sub>

### DIFF
--- a/web/src/components/trace/components/IOPreview/IOPreviewJSONSimple.clienttest.tsx
+++ b/web/src/components/trace/components/IOPreview/IOPreviewJSONSimple.clienttest.tsx
@@ -1,0 +1,145 @@
+/**
+ * The JSON view must never hand a multi-MB field to the unvirtualized
+ * react18-json-view (LFE-10989). These tests pin the branch selection in
+ * IOPreviewJSONSimple: over-limit fields render the bounded fallback and skip
+ * PrettyJsonView entirely, while normal fields render exactly as before.
+ */
+import { fireEvent, render, screen } from "@testing-library/react";
+
+// PrettyJsonView is the unvirtualized render path we must NOT reach for
+// over-limit fields — mock it so its presence is observable by test id.
+vi.mock("@/src/components/ui/PrettyJsonView", () => ({
+  PrettyJsonView: (props: { title?: string }) => (
+    <div data-testid="pretty-json-view">{props.title}</div>
+  ),
+}));
+
+vi.mock("./components/CorrectedOutputField", () => ({
+  CorrectedOutputField: () => <div data-testid="corrected-output" />,
+}));
+
+vi.mock("@/src/features/posthog-analytics/usePostHogClientCapture", () => ({
+  usePostHogClientCapture: () => vi.fn(),
+}));
+
+// deepParseJson is the only runtime import from the shared barrel here; stub it
+// so the test stays light and deterministic (identity is fine for these cases).
+vi.mock("@langfuse/shared", () => ({
+  deepParseJson: (value: unknown) => value,
+}));
+
+import { IOPreviewJSONSimple } from "./IOPreviewJSONSimple";
+import { JSON_VIEW_RENDER_CHAR_LIMIT } from "./lib/jsonViewSizeGate";
+
+const FALLBACK_TEXT = /too large to render in JSON view/i;
+
+describe("IOPreviewJSONSimple size gating", () => {
+  it("renders the fallback and NOT PrettyJsonView for an over-limit field", () => {
+    const huge = "x".repeat(JSON_VIEW_RENDER_CHAR_LIMIT + 1);
+    render(
+      <IOPreviewJSONSimple
+        input={huge}
+        hideOutput
+        hideIfNull
+        showCorrections={false}
+        projectId="p"
+        traceId="t"
+      />,
+    );
+
+    // Fallback shown for the input field...
+    expect(screen.getByText(FALLBACK_TEXT)).toBeInTheDocument();
+    expect(
+      screen.getByRole("button", { name: /Download Input/i }),
+    ).toBeInTheDocument();
+    // ...and the unvirtualized viewer was never mounted.
+    expect(screen.queryByTestId("pretty-json-view")).not.toBeInTheDocument();
+  });
+
+  it("downloads a string field as raw text (.txt), not JSON-wrapped", () => {
+    // The huge-IO seed is a base64 string; JSON.stringify-ing it would add
+    // quotes + escapes and over-encode the download. The fallback must write
+    // the raw text.
+    const rawString = "eyJhIjoxfQ==" + "A".repeat(JSON_VIEW_RENDER_CHAR_LIMIT);
+    const blobCalls: { part: string; type: string }[] = [];
+    let downloadName = "";
+
+    // Capture the exact content handed to the Blob (jsdom's Blob has no
+    // readable .text() in this env, so intercept at construction).
+    class MockBlob {
+      type: string;
+      constructor(parts: BlobPart[], opts?: BlobPropertyBag) {
+        this.type = opts?.type ?? "";
+        blobCalls.push({ part: String(parts[0]), type: this.type });
+      }
+    }
+    vi.stubGlobal("Blob", MockBlob);
+    vi.stubGlobal("URL", {
+      ...URL,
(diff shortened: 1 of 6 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

PR #15159, opened 2026-07-16
fix(trace): gate multi-MB single strings in the Pretty IO view (LFE-10991)

nkabardin:
**TL;DR** — `deepParseJson`'s size guard only fires for objects, so a multi-megabyte top-level **string** slips past every parse bound in the Pretty IO view and drags the tab through several full-length main-thread passes plus an unvirtualized JSON viewer holding the whole payload. `PrettyJsonView` now detects a top-level string over 2M chars and renders a bounded preview + raw download instead — matching the JSON-tab size gate (LFE-10989). Real KB-scale traces are untouched. Part of the large-traces resilience epic (LFE-10152).

**Why** — A giant single-line string I/O (logged prose/base64/blob) hits the Pretty (default) view, where `deepParseJson`'s size guard is object-only (`packages/shared/src/utils/json.ts` — `typeof json === "object"`). So the `maxSize` bounds in `IOPreviewPretty`/`PrettyJsonView` are no-ops for a plain string: it flows through repeated O(n) passes (`deepParseJson` + `decodeUnicodeInJson`, the `isMarkdownContent` `JSON.stringify` probe, then the same again inside the always-mounted hidden `JSONView`) plus an unvirtualized react18-json-view. The cell truncates to 2000 chars visually, but the work + memory are unbounded, and the "expand …N more characters" affordance would dump the full multi-MB string into one DOM node.

**What** — In `PrettyJsonView`, detect a top-level string over `LARGE_STRING_RENDER_CHAR_LIMIT = 2_000_000` (O(1), from raw props) **before** any parse/decode, and short-circuit the parse, the markdown probe, the table data, the body, and the JSON-collapse button to a `LargeStringFallback` (4000-char preview head, no stringify + raw `.txt` download of the full string). Under the cap, or for objects, nothing changes.

**Result** — On seeded 6/30 MB single-string traces, the Pretty view no longer mounts the unvirtualized viewer or dumps the payload into the DOM; it shows a bounded preview + working download and stays responsive in Firefox + Chrome (30 MB main-thread blocking ~2380 → ~1756 ms; DOM bounded to the 4000-char preview vs the whole string). Normal KB-scale traces render exactly as before.

<details><summary>Mechanism, threshold, scope, tests</summary>

- **Gate placement:** `largeStringValue` is the first `useMemo` in `PrettyJsonView`; every heavy path (parse/decode, `isMarkdownContent` probe, `baseTableData`, the hidden `JSONView` mount, the collapse button) is gated on it. The unvirtualized `JSONView` (which itself re-runs `decodeUnicodeInJson(deepParseJson(...))`) is never mounted for an over-limit string — that's most of the win.
- **Threshold 2 MB** = same as the JSON-tab gate (LFE-10989 #15155); ~4–6× the codebase's 300–500K parse bounds, 100–1000× above real KB-scale I/O.
- **Render-layer scope:** this bounds the render layer. A >2M string that reaches the client **without** pre-parsing would still incur the upstream main-thread `deepParseJson` cost (data layer) — consistent with #15155's Formatted-view escape hatch; bounding that is separate LFE-10152 work. In practice >100K payloads are parsed off-main-thread by the observation worker, so the upstream calls are skipped.
- **Complementary to #15155** (JSON tab / `IOPreviewJSONSimple`): this closes the Pretty-view gap #15155 leaves open; separate files, no conflict; a shared constant/fallback is a reasonable post-merge consolidation.
- **Tests:** 4 unit (`largeStringGate` boundary + type exclusions) + 2 component (`PrettyJsonView` renders `LargeStringFallback`, mounts neither `JSONView` nor the lazy table, no DOM dump; small I/O still mounts the viewer) + the 15 existing PrettyJsonView logic tests. Download fires a PostHog `trace_detail:large_string_field_download` capture (mirrors #15155). Typecheck + lint (7/7) clean.

</details>

This PR adds a render-layer size gate for multi-megabyte top-level plain strings in `PrettyJsonView`. Previously, `deepParseJson`'s `maxSize` guard only fired for object payloads, so a 2M+ char string would run through several O(n) main-thread passes (parse, decode, markdown probe, a `JSON.stringify` call inside `isMarkdownContent`) and mount the unvirtualized `react18-json-view` holding the full payload.

- **New gate** (`largeStringGate.ts`): `isLargeRenderString` checks `typeof value === \"string\" && value.length > 2_000_000` in O(1) before any parsing — the first `useMemo` in `PrettyJsonView` — and short-circuits all downstream work (markdown probe, `baseTableData`, the hidden `JSONView`, the collapse button) to a `LargeStringFallback` that renders a 4 000-char preview head plus a raw-text download.
- **`LargeStringFallback`** (`LargeStringFallback.tsx`): self-contained component that slices the preview via `useMemo`, constructs a `Blob` download without re-serializing the string, and fires a PostHog `trace_detail:large_string_field_download` capture.
- **Tests**: four unit tests pin the boundary/type behaviour of `isLargeRenderString`, and two component tests assert that an over-limit string mounts neither `JSONView` nor the table and keeps the DOM footprint under 10 K characters.
</details>

Safe to merge. The gate is logically correct and all downstream heavy paths are properly short-circuited. The only edge case identified (JSON-encoded string receiving the fallback before its outer quotes are stripped) applies to a narrow caller pattern and would merely show a preview with a leading quote — not a crash or data loss.

The gate logic, fallback rendering, PostHog integration, and tests are all well-constructed. Two minor concerns: an edge case where a JSON-encoded large string surfaces with surrounding quotes in the preview/download, and a single inconsistent `!largeStringValue` guard where every other occurrence uses `=== null`. Neither affects the correctness of the performance guard for the targeted scenario.

web/src/components/ui/PrettyJsonView.tsx — the `largeStringValue` useMemo that reads from either `props.parsedJson` or `props.json` deserves a second look to confirm no caller ever passes a raw JSON-encoded string large enough to trigger the gate before `deepParseJson` can unwrap it.
</details>

```mermaid
flowchart TD
    A[PrettyJsonView receives props.json / props.parsedJson] --> B{largeStringValue useMemo\nisLargeRenderString?}
    B -- "YES: string > 2M chars" --> C[largeStringValue = raw string]
    B -- "NO" --> D[largeStringValue = null]

    C --> E[parsedJson useMemo\nreturns largeStringValue as-is\nskip decodeUnicodeInJson]
    D --> F[parsedJson useMemo\nnormal path: deepParseJson\n+ decodeUnicodeInJson]

    C --> G[isChatML: false\nearly-exit for non-object]
    C --> H[isMarkdown / markdownContent\nshort-circuit: false/undefined]
    C --> I[baseTableData: skipped\nlargeStringValue === null guard]
    C --> J[shouldUseTableView: false]

    E --> K{body}
    F --> K

    K -- "largeStringValue !== null" --> L[LargeStringFallback\npreview first 4000 chars\ndownload button]
    K -- "null: normal path" --> M[loading skeleton]
    K -- "null: normal path" --> N[emptyValueDisplay]
    K -- "null: normal path" --> O[MarkdownView]
    K -- "null: normal path" --> P[JsonPrettyTable hidden\n+ JSONView always-mounted]

    L --> Q[PostHog capture on download\ntrace_detail:large_string_field_download]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[PrettyJsonView receives props.json / props.parsedJson] --> B{largeStringValue useMemo\nisLargeRenderString?}
    B -- "YES: string > 2M chars" --> C[largeStringValue = raw string]
    B -- "NO" --> D[largeStringValue = null]

    C --> E[parsedJson useMemo\nreturns largeStringValue as-is\nskip decodeUnicodeInJson]
    D --> F[parsedJson useMemo\nnormal path: deepParseJson\n+ decodeUnicodeInJson]

    C --> G[isChatML: false\nearly-exit for non-object]
    C --> H[isMarkdown / markdownContent\nshort-circuit: false/undefined]
    C --> I[baseTableData: skipped\nlargeStringValue === null guard]
    C --> J[shouldUseTableView: false]

    E --> K{body}
    F --> K

    K -- "largeStringValue !== null" --> L[LargeStringFallback\npreview first 4000 chars\ndownload button]
    K -- "null: normal path" --> M[loading skeleton]
    K -- "null: normal path" --> N[emptyValueDisplay]
    K -- "null: normal path" --> O[MarkdownView]
    K -- "null: normal path" --> P[JsonPrettyTable hidden\n+ JSONView always-mounted]

    L --> Q[PostHog capture on download\ntrace_detail:large_string_field_download]
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/components/ui/PrettyJsonView.tsx:804-808
**Gate fires on JSON-encoded string representation before parse**

When a caller passes `props.json` as a raw JSON string that encodes a large string value (e.g. `'"aaaa…a"'` — 2 000 002 chars including the outer quotes), `isLargeRenderString` fires on the JSON form before `deepParseJson` can strip the surrounding double-quotes. The `LargeStringFallback` then receives and previews `"aaaa…` (with the leading `"` quote), and the downloaded `.txt` file contains the JSON-quoted form rather than the clean string. The threshold means this only matters for strings already very close to 2 M characters, but it could confuse users who see a preview that starts and ends with a literal `"`. A comment noting this behaviour — or an early pre-strip for the quoted case — would prevent surprises.

### Issue 2 of 2
web/src/components/ui/PrettyJsonView.tsx:1460
All other checks use `largeStringValue === null`; this condition uses `!largeStringValue`. Both are equivalent here (the value is always `null` or a non-empty string), but the inconsistency can confuse readers who expect a uniform pattern. Prefer `largeStringValue === null` for clarity.

```suggestion
              {!shouldUseTableView && !isMarkdownMode && largeStringValue === null && (
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test(trace): component test for the larg..."](https://github.com/langfuse/langfuse/commit/eb7b9ca60df5741485452323766bfc23ca0e67af) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44768141)</sub>


nkabardin:
@claude review

PR #15189, opened 2026-07-17
fix(experiments): sum item cost across full observation subtree

wochinge:
## Summary

- The experiments grid's per-item cost query read a single root-span's own `total_cost` instead of aggregating cost across the item's full observation subtree. Root spans in the v4 events pipeline are typically cost-free wrapper nodes (real cost lives on nested generations), so the per-row cost showed $0 even when the experiment's aggregate cost was correct.
- Fixed the query to sum cost via a window function (`SUM(e.total_cost) OVER (PARTITION BY experiment_item_id, experiment_id)`) scoped to the same bounded item/experiment set the query already filters on, and moved root-row selection from `WHERE` into `ORDER BY`/`LIMIT BY` so the window function still sees sibling (child) rows before the root row is picked.
- Added a regression test seeding items with different cost distributions (cost-only-on-child, split root/child, root-only), asserting both the per-item fix and that summed per-item costs match the independently-computed experiment-level aggregate (`getExperimentMetricsFromEvents`).

## Linear

- [LFE-11025](https://linear.app/langfuse/issue/LFE-11025/ramp-experiments-table-shows-non-aggregated-costs)

## Major Decisions

- None.

## Review Focus

- **Query shape**: the fix moves the `experiment_item_root_span_id = span_id` narrowing out of `WHERE` and into `ORDER BY (span_id = experiment_item_root_span_id) DESC, start_time DESC` + `LIMIT 1 BY (experiment_item_id, experiment_id)`, so the window function sees every observation in scope before the root row is picked. The query already preselects a bounded page of item IDs upstream, so this stays bounded by nature.
- ClickHouse query profiling before/after shows no meaningful cost regression: same row count read (4,456,536 rows), elapsed 0.121s → 0.104s, bytes read 737.74 MB → 759.32 MB (~3% more, negligible).
  - Before: 
    CleanShot 2026-07-17 at 11 48 20
  - After: 
    CleanShot 2026-07-17 at 11 49 14

This PR fixes a bug where the experiments grid showed `$0` per-item cost because the query read only the root span's own `total_cost` — but root spans in the v4 events pipeline are cost-free wrapper nodes with real cost living on child generations. The fix introduces a window function (`SUM(e.total_cost) OVER (PARTITION BY experiment_item_id, experiment_id)`) and moves root-span selection from `WHERE` to `ORDER BY (span_id = experiment_item_root_span_id) DESC` + `LIMIT 1 BY`, so the window aggregates across the full subtree before the root row is selected.

- **`query-fragments.ts`**: Extracts a new `eventsExperimentsForItems` helper that omits the `experiment_item_root_span_id = span_id` filter, then rebuilds `eventsExperimentsRootSpans` as a thin wrapper around it, preserving existing callers.
- **`experiments.ts`**: Switches `getExperimentItemsFromEvents` to use `eventsExperimentsForItems` + the window-function cost column, and adds a two-column ORDER BY so `LIMIT 1 BY` still lands on the root span.
- **Test**: Seeds three items with distinct cost distributions (child-only, split, root-only) and cross-checks per-item sums against the independently-computed experiment-level aggregate from `getExperimentMetricsFromEvents`.
</details>

Safe to merge; the query rework is logically sound and verified by profiling data showing no performance regression.

The window-function approach is correct: ClickHouse evaluates window expressions before LIMIT BY, so the aggregate sees all child observations before the root row is selected. The ORDER BY prioritising `(span_id = experiment_item_root_span_id) DESC` reliably lifts the root span to position 1 in each partition. The one minor shift worth awareness is that `SUM()` over an all-NULL partition returns 0 rather than NULL, which changes the downstream `totalCost` value from `null` to `0` for items with no recorded cost anywhere in their subtree.

The `experiments.ts` change is the core query modification; the NULL-to-zero coercion is the only subtle difference from the previous behaviour.
</details>

```mermaid
sequenceDiagram
    participant UI as Experiments Grid
    participant Repo as getExperimentItemsFromEvents
    participant CH as ClickHouse

    UI->>Repo: fetch items (projectId, baseExperimentId, ...)
    Repo->>CH: QUERY 1 — get paged item IDs\n(eventsExperimentsRootSpans)
    CH-->>Repo: [item_id, ...]
    Repo->>CH: "QUERY 2 — fetch data for all experiments\n(eventsExperimentsForItems)\nSELECT SUM(total_cost) OVER (PARTITION BY item_id, exp_id)\nORDER BY (span_id = root_span_id) DESC, start_time DESC\nLIMIT 1 BY item_id, exp_id"
    Note over CH: Window fn sees ALL child observations\nbefore LIMIT BY selects root span row
    CH-->>Repo: one row per (item_id, exp_id) with aggregated total_cost
    Repo-->>UI: ExperimentItemData[] (correct cost)
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant UI as Experiments Grid
    participant Repo as getExperimentItemsFromEvents
    participant CH as ClickHouse

    UI->>Repo: fetch items (projectId, baseExperimentId, ...)
    Repo->>CH: QUERY 1 — get paged item IDs\n(eventsExperimentsRootSpans)
    CH-->>Repo: [item_id, ...]
    Repo->>CH: "QUERY 2 — fetch data for all experiments\n(eventsExperimentsForItems)\nSELECT SUM(total_cost) OVER (PARTITION BY item_id, exp_id)\nORDER BY (span_id = root_span_id) DESC, start_time DESC\nLIMIT 1 BY item_id, exp_id"
    Note over CH: Window fn sees ALL child observations\nbefore LIMIT BY selects root span row
    CH-->>Repo: one row per (item_id, exp_id) with aggregated total_cost
    Repo-->>UI: ExperimentItemData[] (correct cost)
```

</a>
</details>

`````markdown

---

### Issue 1 of 1
packages/shared/src/server/repositories/experiments.ts:1092
**NULL total_cost becomes 0 after aggregation**

`SUM(total_cost)` in ClickHouse returns `0` (not `NULL`) when every row in the partition has a `NULL` `total_cost`. Before this change, the root span's own `total_cost` was returned directly — preserving `NULL` when absent. After the change, items whose entire observation subtree has no recorded cost will now surface `totalCost: 0` instead of `totalCost: null` downstream. In practice this is unlikely to matter (a $0 cost is treated the same as a missing cost in the UI), but worth noting if any caller uses a strict null-check to mean "no cost data available".

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(experiments): sum item cost across f..."](https://github.com/langfuse/langfuse/commit/dc6c0a45767fe72e216092df4a658f76505b0e6d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45050850)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/experiments.ts
+++ b/packages/shared/src/server/repositories/experiments.ts
@@ -1074,15 +1074,10 @@ export const getExperimentItemsFromEvents = async (
   ];
 
   // ========== QUERY 2: Fetch data for ALL experiments ==========
-  // total_cost is summed across the item's full observation subtree via a
-  // window function - the root span itself usually carries no cost of its
-  // own, so reading e.total_cost directly would just return 0. WHERE only
-  // scopes to project/experiment/item; root-row selection happens in
-  // ORDER BY/LIMIT BY below, so the window function still sees sibling rows.
-  // The partition includes trace_id so items with multiple repetitions
-  // (until we model them properly, LFE-8965) sum only the selected
-  // iteration's own subtree, not every repetition's cost combined.
-  const queryBuilderData = eventsExperimentsForItems({
+  // Collapses un-merged ReplacingMergeTree span versions to the newest one
+  // (by event_ts) before summing cost, so a duplicate version isn't counted
+  // twice.
+  const dedupBuilder = eventsExperimentsForItems({
     projectId,
     experimentItemIds: itemIds,
     experimentIds: allExperimentIds,
@@ -1092,24 +1087,56 @@ export const getExperimentItemsFromEvents = async (
       "e.experiment_id as experiment_id",
       "e.level as level",
       "e.start_time as start_time",
-      "sum(e.total_cost) OVER (PARTITION BY e.experiment_item_id, e.experiment_id, e.trace_id) as total_cost",
-      "if(isNull(e.end_time), NULL, date_diff('millisecond', e.start_time, e.end_time)) as latency_ms",
+      "e.end_time as end_time",
+      "e.total_cost as total_cost",
       "e.span_id as observation_id",
       "e.trace_id as trace_id",
+      "e.experiment_item_root_span_id as experiment_item_root_span_id",
     )
-    // We must deterministically return the latest row for each experiment_item_id, experiment_id pair until we model repetitions (LFE-8965).
-    // Uses raw orderBy() rather than orderByColumns(), which auto-prepends a
-    // toStartOfMinute(start_time) primary-key-read-order prefix whenever a
-    // start_time entry is present - that prefix would outrank the root-flag
-    // tiebreak below whenever a child observation starts in a later minute
-    // bucket than its root, causing LIMIT BY to keep the child row instead.
+    .orderBy("ORDER BY e.event_ts DESC")
+    .limitBy("e.span_id");
+
+  // total_cost sums the item's full observation subtree (root spans usually
+  // carry no cost of their own); partitioned by trace_id too so items with
+  // multiple repetitions (LFE-8965) sum only the selected iteration.
+  // ORDER BY uses a raw clause, not orderByColumns(), to avoid its
+  // auto-prepended toStartOfMinute(start_time) prefix outranking the
+  // root-flag tiebreak.
+  const queryBuilder = new CTEQueryBuilder()
+    .withCTE("deduped_events", {
+      ...dedupBuilder.buildWithParams(),
+      schema: [
+        "item_id",
+        "experiment_id",
+        "level",
+        "start_time",
+        "end_time",
+        "total_cost",
+        "observation_id",
+        "trace_id",
+        "experiment_item_root_span_id",
+      ],
+    })
+    .from("deduped_events", "d")
+    .selectColumns(
+      "d.item_id",
+      "d.experiment_id",
+      "d.level",
+      "d.start_time",
+      "d.observation_id",
+      "d.trace_id",
+    )
+    .select(
+      "sum(d.total_cost) OVER (PARTITION BY d.item_id, d.experiment_id, d.trace_id) as total_cost",
+      "if(isNull(d.end_time), NULL, date_diff('millisecond', d.start_time, d.end_time)) as latency_ms",
+    )
     .orderBy(
-      "ORDER BY (e.span_id = e.experiment_item_root_span_id) DESC, e.start_time DESC",
+      "ORDER BY (d.observation_id = d.experiment_item_root_span_id) DESC, d.start_time DESC",
     )
-    .limitBy("e.experiment_item_id, e.experiment_id");
+    .limitBy("d.item_id, d.experiment_id");
 
   const { query: dataQuery, params: dataParams } =
-    queryBuilderData.buildWithParams();
+    queryBuilder.buildWithParams();
 
   const rows = await queryClickhouse<ExperimentItemEventsDataReturnType>({
     query: dataQuery,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by wochinge** [comment]
@claude review

**Comment by wochinge** [comment]
@claude review

File: packages/shared/src/server/repositories/experiments.ts [hunk 1]
```diff
@@ -1073,7 +1074,10 @@ export const getExperimentItemsFromEvents = async (
   ];
 
   // ========== QUERY 2: Fetch data for ALL experiments ==========
-  const queryBuilderData = eventsExperimentsRootSpans({
+  // Collapses un-merged ReplacingMergeTree span versions to the newest one
```
wochinge: @sumerman this part only

sumerman: Do we actually have duplicates?

packages/shared/src/server/repositories/experiments.ts
```diff
@@ -1083,17 +1087,56 @@ export const getExperimentItemsFromEvents = async (
...
+    .orderBy("ORDER BY e.event_ts DESC")
+    .limitBy("e.span_id");
+
+  // total_cost sums the item's full observation subtree (root spans usually
+  // carry no cost of their own); partitioned by trace_id too so items with
+  // multiple repetitions (LFE-8965) sum only the selected iteration.
+  // ORDER BY uses a raw clause, not orderByColumns(), to avoid its
+  // auto-prepended toStartOfMinute(start_time) prefix outranking the
+  // root-flag tiebreak.
+  const queryBuilder = new CTEQueryBuilder()
```
sumerman: start_time doesn't seem to be used at all for filtering. it should be

packages/shared/src/server/repositories/experiments.ts
```diff
@@ -1083,17 +1087,56 @@ export const getExperimentItemsFromEvents = async (
...
       "e.total_cost as total_cost",
-      "if(isNull(e.end_time), NULL, date_diff('millisecond', e.start_time, e.end_time)) as latency_ms",
       "e.span_id as observation_id",
       "e.trace_id as trace_id",
+      "e.experiment_item_root_span_id as experiment_item_root_span_id",
     )
-    // We must deterministically return the latest row for each experiment_item_id, experiment_id pair until we model repetitions (LFE-8965)
-    .orderByColumns([{ column: "e.start_time", direction: "DESC" }])
-    .limitBy("e.experiment_item_id, e.experiment_id");
+    .orderBy("ORDER BY e.event_ts DESC")
```
sumerman: I would probably prefer if this was based on the table ordering plus event_ts as a tiebreaker. By definition of this table an identity of an event is `(project_id, toStartOfMinute(start_time), xxHash32(trace_id), span_id, start_time)` therefore ORDER BYs and LIMIT BYs ideally should match it unless there's a good reason

packages/shared/src/server/repositories/experiments.ts
wochinge: Not sure to be honest. I added it as good practice to avoid weird downstream errors.

sumerman: Design-wise it is possible, but should be extremely rare.

wochinge: okay, let's drop it then - rids us of a bunch of complexity then

packages/shared/src/server/repositories/experiments.ts
wochinge: dropped the deduplication so this is no longer relevant

packages/shared/src/server/repositories/experiments.ts
wochinge: Was a previously existing issue but fixed as part of this

PR #15190, opened 2026-07-17
feat(sessions): cleaner more compact view

nimarb:

This PR ships the "Compact Session View" (Modern Session) as a cloud-only Feature Preview, replacing the existing per-trace card layout with a minimap sidebar and a single virtualized conversation feed. It also adds `IOPreviewContentMode` to let observers filter the feed to conversation-only messages or raw data/tool calls, and extracts action buttons into a shared `SessionTraceActionButtons` component.

- **New `ModernSession` layout**: a left-panel minimap lists all traces with scroll-spy sync; the right panel virtualizes trace content using `source="modern"` absolute positioning instead of CSS transforms to avoid sticky-element offset bugs.
- **Content mode filtering**: `shouldRenderMessageForContentMode` gates system messages, tool/function roles, and pure-JSON messages per mode; `parseChatML` is extracted from the hook into a pure function so pre-computed results can be passed through the component tree without re-parsing.
- **ClickHouse `observation_count`**: a new `groupUniqArrayIf` field in `EventsAggregationQueryBuilder` counts non-synthetic child spans and surfaces the count in the minimap without an extra query.
</details>

Safe to merge behind the Feature Preview flag; the new layout is additive and the legacy card path is untouched.

The minimap label derives its count from a fixed multiplier that routinely misrepresents actual tool-call counts, and a no-op view-toggle handler in data mode may surface a broken button. Neither blocks the feature but both warrant cleanup before GA.

ModernSession.tsx (misleading toolDataCount label) and ChatMessage.tsx (no-op view-toggle handler in data mode)
</details>

`````markdown

---

### Issue 1 of 3
web/src/components/session/ModernSession.tsx:61
**Misleading `toolDataCount` heuristic**

`observationCount * 2` has no reliable relationship to the actual number of tool-call or data items a trace contains. A trace with 5 purely text observations would display "Tool calls & data · 10" in the collapsible trigger (line 146), while a trace with 2 observations and 8 tool calls would display "4". Because `observationCount` counts non-synthetic spans (from ClickHouse), the × 2 multiplier doesn't correspond to any real content property. Consider omitting the count until the actual item count is available from the loaded observations, or dropping the numeric suffix entirely from the collapsed label.

### Issue 2 of 3
web/src/components/session/ModernSession.tsx:23
**Loose `any` type for `openPeek` row argument**

The `row: any` parameter bypasses TypeScript's type checker for the second argument passed to `openPeek`. At the two call sites the argument is an `EventSessionTrace`, so the type can be tightened.

```suggestion
import { type EventSessionTrace } from "@/src/components/session/sessionDetailPageTypes";
type OpenPeek = (id: string, row: EventSessionTrace) => void;
```

### Issue 3 of 3
web/src/components/trace/components/IOPreview/components/ChatMessage.tsx:113-120
**No-op `handleOnValueChange` may expose a broken toggle**

`handleOnValueChange={() => {}}` is passed to `MarkdownJsonViewHeader` for the tool-calls branch of the `data` content mode. If `MarkdownJsonViewHeader` renders a pretty/JSON view toggle button wired to this handler (as it does in non-data contexts), the button will appear in the UI but do nothing when clicked. Since the view is already fixed to showing tool invocations via `ToolCallInvocationsView`, consider either omitting the prop or explicitly disabling the toggle, whichever `MarkdownJsonViewHeader` supports.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into feature/lfe-104..."](https://github.com/langfuse/langfuse/commit/4d029b122addb4369e092e6fd9c9e748380d19d4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45077637)</sub>


nimarb:
@claude review

PR #15201, opened 2026-07-17
chore(clickhouse): propagate query_id to streaming queries and log_comment to all queries (LFE-11043)

niklassemmler:
## What does this PR do?

Brings the streaming ClickHouse query path to parity with the blob exporter for query attribution (LFE-11043), and closes the remaining `log_comment` gaps so every ClickHouse query carries attribution.

**query_id propagation** — `queryClickhouseStream`, `queryClickhouseStreamRawText`, `queryClickhouseWithProgress`, and `queryClickhouseExecRaw` now generate the `query_id` client-side (`randomUUID()`) and pass it to the ClickHouse client, instead of only reading it from the response:

- Errors thrown **before or without a response** (e.g. a query killed at start by the memory OvercommitTracker) now carry `[query_id: …]` in the message — previously the id was only known after a successful response, so exactly the failures that needed it lost it. In `queryClickhouseExecRaw` the pre-response `enrichWithQueryId` call referenced a `queryId` that was always `undefined` at that point; enrichment now happens exactly once, in the outer catch.
- `ch.queryId` is set on the query span up front, so failed-query spans carry it too.
- `system.query_log` / `system.processes` stay pollable by a known id for the whole query lifetime.

Batch export failure logs (`[BATCH EXPORT] Getting data from DB and transform failed`) include the error message, so the query id is now recoverable from Datadog even after `system.query_log` retention expires.

**log_comment coverage** — the last raw `clickhouseClient()` calls without attribution now set `log_comment` via `buildClickHouseLogComment` (`surface: worker`, `route: background-migration.<name>`): the `SHOW TABLES` validation guards and the Postgres→ClickHouse migration inserts in `worker/src/backgroundMigrations/**`.

Fixes [LFE-11043](https://linear.app/langfuse/issue/LFE-11043)

## Type of change

- [x] Chore (tooling, dependencies, CI, workflows, repo upkeep, or other maintenance work)

## Mandatory Tasks

This PR improves ClickHouse observability by generating a `query_id` client-side (via `randomUUID()`) before each streaming query so that span attributes and error messages always carry a stable ID — even when a request fails before ClickHouse returns a response. It also backfills `log_comment` on raw `clickhouseClient()` calls in background migration classes that were bypassing the shared wrapper.

- `queryClickhouseStream`, `queryClickhouseStreamRawText`, and `queryClickhouseExecRaw` now generate and forward `query_id` to ClickHouse, set `ch.queryId` on the span upfront, and append `[query_id: <uuid>]` to thrown errors via the new `enrichWithQueryId` helper.
- Background migration files (`migrateTracesFromPostgresToClickhouse`, `migrateObservationsFromPostgresToClickhouse`, `migrateScoresFromPostgresToClickhouse`, `migrateDatasetRunItemsFromPostgresToClickhouseRmt`, `migrateEventLogToBlobStorageRefTable`, `backfillEventsFullFromDatasetRunItems`, and `backfillBase`) now include `log_comment` on their direct `SHOW TABLES` and `insert` calls.
- New server tests verify that both pre-stream and mid-stream errors from `queryClickhouseStream` and `queryClickhouseStreamRawText` expose a UUID-shaped `[query_id: ...]` in the thrown error message.
</details>

Safe to merge; the changes are additive observability improvements with no risk to query execution or data integrity.

The core change — moving `query_id` assignment before the request — is correct and well-tested for the three streaming functions it targets. Two gaps are worth a follow-up: `queryClickhouseWithProgress` was not updated to the same pattern, and `queryClickhouseExecRaw` can produce double `[query_id: ...]` in non-resource error messages because both the inner `.catch` and the outer `catch` call `enrichWithQueryId` for non-resource errors. Neither affects production behavior beyond slightly noisy error strings.

`packages/shared/src/server/repositories/clickhouse.ts` — the `queryClickhouseWithProgress` function and the double-enrichment path in `queryClickhouseExecRaw` warrant a second look.
</details>

`````markdown

---

### Issue 1 of 2
packages/shared/src/server/repositories/clickhouse.ts:716-764
**`queryClickhouseWithProgress` skipped in query_id propagation**

`queryClickhouseWithProgress` is a streaming async generator with the same error surface as `queryClickhouseStream`, but it was not updated in this PR. It still uses a server-assigned `query_id` (set after the response via `sendClickhouseQuery`), so if a request fails before ClickHouse responds, the span attribute `ch.queryId` is never set and errors are not enriched with `[query_id: ...]`. The same client-generated UUID pattern applied to the other streaming functions should be applied here.

### Issue 2 of 2
packages/shared/src/server/repositories/clickhouse.ts:447-501
**Double `[query_id: ...]` appended for non-resource errors in `queryClickhouseExecRaw`**

The inner `.catch` on the `exec()` call already calls `enrichWithQueryId(error, queryId)` and then `wrapIfResourceError`. For resource errors (memory limit, timeout, etc.), `wrapIfResourceError` returns a `ClickHouseResourceError`, which the outer `catch` re-throws immediately via `if (error instanceof ClickHouseResourceError) throw error` — so only one enrichment. But for non-resource errors (e.g. table-not-found, connection reset), `wrapIfResourceError` returns the plain enriched error, which the outer `catch` then passes through `enrichWithQueryId` a second time, producing messages like `"...table not found [query_id: abc] [query_id: abc]"`. The other streaming functions (`queryClickhouseStream`, `queryClickhouseStreamRawText`) avoid this by not calling `enrichWithQueryId` in the inner `.catch` — only in the outer one.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore(clickhouse): propagate query\_id to..."](https://github.com/langfuse/langfuse/commit/11d18206cc96f0e0aa7243769dbecf9ef05eed01) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45111520)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/clickhouse.ts
+++ b/packages/shared/src/server/repositories/clickhouse.ts
@@ -471,8 +471,16 @@ export async function queryClickhouseExecRaw(
     );
 
     // The span outlives this function (it covers the consumer's read). Forward
-    // source errors so the consumer sees them.
-    res.stream.on("error", (error) => guardedStream.destroy(error));
+    // source errors (e.g. mid-transfer connection resets) so the consumer sees
+    // them, enriched like the other error paths of this function.
+    res.stream.on("error", (error) =>
+      guardedStream.destroy(
+        ClickHouseResourceError.wrapIfResourceError(
+          enrichWithQueryId(error, queryId),
+          normalizedTags,
+        ),
+      ),
+    );
     // `.pipe()` only wires src→dest, so destroying guardedStream (e.g. the
     // worker's pipeline aborting on an upload failure) would leave the live CH
     // body streaming into an unread socket — pinning a connection slot and query
--- a/packages/shared/src/server/repositories/clickhouse.ts
+++ b/packages/shared/src/server/repositories/clickhouse.ts
@@ -264,14 +264,17 @@ export async function* queryClickhouseStream<T>(
     kind: SpanKind.CLIENT,
   });
 
-  let queryId: string | undefined;
+  // Client-generated so failures before/without a response still carry a
+  // query_id on errors and spans; system.query_log stays pollable by id.
+  const queryId = randomUUID();
 
   try {
     setSpanQueryAttributes(span, opts.query);
+    span.setAttribute("ch.queryId", queryId);
 
     const res = await context
       .with(trace.setSpan(context.active(), span), () =>
-        sendClickhouseQuery({ ...opts, format: "JSONEachRow", span }),
+        sendClickhouseQuery({ ...opts, format: "JSONEachRow", span, queryId }),
       )
       .catch((error) => {
         throw ClickHouseResourceError.wrapIfResourceError(
@@ -279,9 +282,6 @@ export async function* queryClickhouseStream<T>(
           normalizedTags,
         );
       });
-
-    queryId = res.query_id;
-    span.setAttribute("ch.queryId", queryId);
 
     for await (const rows of res.stream<T>()) {
       for (const row of rows) {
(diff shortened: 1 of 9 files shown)


niklassemmler:
* **#15201** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/15201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/15201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
@claude review

PR #15202, opened 2026-07-17
fix(traces): clean up annotation queue items when their trace is deleted (#12852)

### LINKED ISSUE #12852: bug: Data retention (TTL) deletes traces referenced by annotation queue items, leaving orphaned items
### Describe the bug

  Traces that are referenced by annotation queue items are deleted by the data retention (TTL) mechanism without any check, leaving orphaned `AnnotationQueueItem` records in PostgreSQL that point to non-existent traces in ClickHouse.

  When a user opens one of these orphaned annotation queue items in the UI, they see:

  > **Trace not found** — Path: `traces.byIdWithObservationsAndScores`

  ### Root cause

  The TTL deletion logic in `handleDataRetentionProcessingJob.ts` and `BatchDataRetentionCleaner` deletes traces purely by age (`timestamp < cutoffDate`)
  without checking whether any `AnnotationQueueItem` references them via `objectId`.

  Since `AnnotationQueueItem.objectId` is a plain string with no foreign key to ClickHouse traces, there is no referential integrity — the deletion proceeds
  silently and the queue items become orphaned.

  ### Impact

  - Annotation queue items marked as PENDING can never be completed since their underlying data is gone
  - Completed annotation queue items lose their trace context, making reviews un-auditable
  - No warning is shown to users when setting retention on projects with active annotation queues
  - No cleanup mechanism exists to remove orphaned items

  ### Suggested solutions

  1. **Exclude referenced traces from TTL deletion** — Before deleting traces by age, check if any `AnnotationQueueItem` references them and skip those
  2. **Cascade cleanup** — When traces are deleted by TTL, also delete (or mark as expired) the corresponding `AnnotationQueueItem` records
  3. **UI/UX improvement** — Show a clear "This trace was deleted by data retention" message instead of a generic "Trace not found" error
  4. **Warn on retention config** — When a user sets data retention on a project with active annotation queues, warn them that queued items may be affected

  ### Related

  - #10332 — Similar UI error when annotation queue API accepts invalid `objectId` at creation time (different root cause, same symptom)

### Steps to reproduce

  1. Create an annotation queue and add traces to it
  2. Set the project's data retention to N days
  3. Wait N days (or set a short retention for testing)
  4. Open the annotation queue in the UI
  5. Items that referenced expired traces show "Trace not found"

Self-hosted

v3.150.0

_No response_

### Additional information

_No response_

Yes

hassaanch23:
## What does this PR do?

Annotation queue items reference traces via `objectId` with **no foreign key** to the ClickHouse traces they point at. When a trace is deleted, nothing removes the referencing `AnnotationQueueItem` rows, so they become orphans pointing at data that no longer exists. Opening such an item in the review UI fails with:

> **Trace not found** — Path: `traces.byIdWithObservationsAndScores`

Tracing the deletion code, **three** separate paths delete traces and all of them leak queue items:

1. **Per-project data retention (TTL)** — the reported issue. `handleDataRetentionProcessingJob` deletes traces older than the cutoff purely by age and never touches the referencing queue items.
2. **Manual / batch trace deletion** — `processPostgresTraceDelete` already cleans up `jobExecution` rows for the deleted traces, but not annotation queue items, so single/batch/cleaner deletions orphan items too.
3. **Batch retention cleaner** — `BatchDataRetentionCleaner` (opt-in) bulk-deletes expired traces straight from ClickHouse and touches no Postgres at all.

### Fix

Cascade-delete the `AnnotationQueueItem` rows that reference a trace whenever that trace is deleted, across all three paths, via two small shared helpers in `packages/shared/src/server/repositories`:

- `deleteAnnotationQueueItemsByObjectIds({ projectId, objectType, objectIds })` — the Postgres delete.
- `getExpiredAnnotationQueueTraceIds(projectId, beforeDate)` — resolves which of a project's queue-referenced traces are past the cutoff (annotation queues are a small, curated set, so this stays cheap), built on a new `getTraceIdsOlderThan` ClickHouse query.

Wiring:

- **Manual path** (`processPostgresTraceDelete`): delete queue items for the exact trace ids being removed.
- **Per-project TTL** (`handleDataRetentionProcessingJob`) and **batch cleaner** (`BatchDataRetentionCleaner`): resolve the expiring referenced trace ids **before** the ClickHouse delete removes them, then delete the queue items **after** the trace data is gone — so a failed ClickHouse delete never prematurely drops still-valid items.

### Tests

`worker/src/__tests__/annotationQueueTraceDeletionCleanup.test.ts` covers all three paths:

- Manual deletion removes the referencing item and leaves unrelated items intact.
- Every item referencing the same deleted trace is removed (multiple queues).
- Deletion is project-scoped: a same-`objectId` item in another project is untouched, while the same-project item is removed.
- OBSERVATION-type items are left untouched (documents the intentional scope).
- Per-project TTL and the batch cleaner each remove items whose trace is expired while keeping items whose trace is still within retention.

### Scope / follow-ups

Scoped to `TRACE`-type queue items — the reported case and the dominant usage. Two same-class gaps are called out in code comments as deliberate follow-ups rather than bundled here:

- **`OBSERVATION`-type items**: trace deletion also deletes the trace's observations, so observation-referencing items can orphan the same way. Handling them needs an observation → trace resolution the manual path doesn't currently have.
- **`SESSION`-type items**: a session is derived from its traces (no ClickHouse `sessions` table), so it only "disappears" once *all* its traces are deleted — a different liveness check.

Project deletion already cascades queue items via the Postgres FK (`onDelete: Cascade`), so that path is unaffected.

Fixes #12852

## Type of change

## Mandatory Tasks

## Checklist

- [x] PR title follows Conventional Commits.
- [x] Added tests that prove the fix is effective (`annotationQueueTraceDeletionCleanup.test.ts`, covering all three deletion paths).
- [x] Commented the non-obvious areas (ordering rationale, scope boundaries).
- [x] No documentation changes required.
- [ ] Lint and the full test suite are validated by CI on this PR.

This PR adds cascade-delete logic for `AnnotationQueueItem` rows that reference traces deleted via any of three paths: manual/batch deletion, per-project data retention TTL, and the batch retention cleaner. Two shared helpers (`deleteAnnotationQueueItemsByObjectIds` and `getExpiredAnnotationQueueTraceIds`) are introduced and wired into each path.

- The manual-deletion path (`processPostgresTraceDelete`) correctly deletes queue items for the explicit trace IDs being removed and is safe on retry.
- The TTL path (`handleDataRetentionProcessingJob`) and batch cleaner (`BatchDataRetentionCleaner`) use a pre-compute/post-delete strategy: queue trace IDs are resolved before ClickHouse deletion, then the queue items are removed after. However, in both places the secondary lookup is not wrapped in error handling — a failure there can block the primary trace deletion, and a failure in the post-delete cleanup creates permanently unrecoverable orphans on retry (because `getTraceIdsOlderThan` returns empty for already-deleted traces).
</details>

Merging is risky for the TTL and batch-cleaner paths: a transient Prisma error during the pre-deletion lookup can block all trace deletion for a project, and a failure in the post-deletion queue cleanup leaves queue items permanently orphaned after the traces are already gone.

The manual-deletion path is clean and correctly implemented. The retention paths place a new Prisma query in the critical path of ClickHouse trace deletion without error handling, and the post-delete queue cleanup cannot be recovered on retry once traces have been removed from ClickHouse.

worker/src/ee/dataRetention/handleDataRetentionProcessingJob.ts and worker/src/features/batch-data-retention-cleaner/index.ts both need try/catch around the annotation queue lookup and the subsequent delete so failures are logged rather than propagated.
</details>

```mermaid
sequenceDiagram
    participant Job as Retention Job / Batch Cleaner
    participant PG as PostgreSQL (Prisma)
    participant CH as ClickHouse

    Note over Job,CH: Pre-deletion: resolve expiring queue trace IDs
    Job->>PG: getExpiredAnnotationQueueTraceIds()
    PG-->>Job: referencedTraceIds[]
    Job->>CH: getTraceIdsOlderThan(referencedTraceIds, cutoff)
    CH-->>Job: expiredTraceIds[]

    Note over Job,CH: Primary deletion
    Job->>CH: deleteTracesOlderThanDays / executeBatchDelete
    CH-->>Job: OK

    Note over Job,PG: Post-deletion cleanup
    Job->>PG: deleteAnnotationQueueItemsByObjectIds(expiredTraceIds)
    PG-->>Job: deletedCount

    Note over Job,CH: Manual path (processPostgresTraceDelete)
    Job->>PG: jobExecution.deleteMany(traceIds)
    Job->>PG: deleteAnnotationQueueItemsByObjectIds(traceIds)
    Job->>CH: processClickhouseTraceDelete(traceIds)
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Job as Retention Job / Batch Cleaner
    participant PG as PostgreSQL (Prisma)
    participant CH as ClickHouse

    Note over Job,CH: Pre-deletion: resolve expiring queue trace IDs
    Job->>PG: getExpiredAnnotationQueueTraceIds()
    PG-->>Job: referencedTraceIds[]
    Job->>CH: getTraceIdsOlderThan(referencedTraceIds, cutoff)
    CH-->>Job: expiredTraceIds[]

    Note over Job,CH: Primary deletion
    Job->>CH: deleteTracesOlderThanDays / executeBatchDelete
    CH-->>Job: OK

    Note over Job,PG: Post-deletion cleanup
    Job->>PG: deleteAnnotationQueueItemsByObjectIds(expiredTraceIds)
    PG-->>Job: deletedCount

    Note over Job,CH: Manual path (processPostgresTraceDelete)
    Job->>PG: jobExecution.deleteMany(traceIds)
    Job->>PG: deleteAnnotationQueueItemsByObjectIds(traceIds)
    Job->>CH: processClickhouseTraceDelete(traceIds)
```

</a>
</details>

`````markdown

---

### Issue 1 of 4
worker/src/ee/dataRetention/handleDataRetentionProcessingJob.ts:92-95
**Secondary lookup blocks primary trace deletion**

`getExpiredAnnotationQueueTraceIds` issues a Prisma query before any traces are deleted. If that query fails (transient DB overload, connection pool exhaustion, etc.), the entire job throws and the primary ClickHouse deletion at line 101 never runs. Repeated failures on the next retry cycle could delay retention indefinitely for the affected project. Because the annotation queue cleanup is best-effort relative to the main deletion, this lookup should not block it — wrap the lookup and the subsequent delete in a `try`/`catch` that logs but does not re-throw.

### Issue 2 of 4
worker/src/ee/dataRetention/handleDataRetentionProcessingJob.ts:119-130
**Queue-item delete failure after ClickHouse delete creates permanent orphans**

If `deleteAnnotationQueueItemsByObjectIds` throws after ClickHouse traces have already been removed (line 101–114), the job fails and BullMQ retries. On the retry, `getExpiredAnnotationQueueTraceIds` calls `getTraceIdsOlderThan`, which queries ClickHouse for those trace IDs — but they no longer exist, so it returns `[]`. The `if (expiredReferencedTraceIds.length > 0)` guard is never entered, and the queue items are permanently orphaned — the exact bug this PR aims to fix. Wrapping lines 120–129 in a `try`/`catch` that only logs (and does not re-throw) prevents the job from failing and avoids this unrecoverable state.

### Issue 3 of 4
worker/src/features/batch-data-retention-cleaner/index.ts:214-225
**`Promise.all` over per-project lookups blocks `executeBatchDelete`**

If `getExpiredAnnotationQueueTraceIds` rejects for any project in the `workloads` list, the `await Promise.all(...)` at line 216 rejects, and `executeBatchDelete` at line 227 never runs for any project. A single project's Prisma issue therefore prevents batch trace deletion across all projects in this run. The queue cleanup should be best-effort — wrapping this block (and the subsequent `cleanupOrphanedAnnotationQueueItems` call) in a `try`/`catch` that logs but continues keeps the primary deletion path independent of the secondary cleanup.

### Issue 4 of 4
worker/src/features/batch-data-retention-cleaner/index.ts:234-236
**`cleanupOrphanedAnnotationQueueItems` failure after batch delete causes unrecoverable orphans**

If `cleanupOrphanedAnnotationQueueItems` throws after `executeBatchDelete` has already removed the traces, the `PeriodicExclusiveRunner` schedules a fresh execution. On the next run, `getExpiredAnnotationQueueTraceIds` returns `[]` for those projects because their traces are gone, so the cleanup is never reattempted and the queue items remain orphaned permanently. Catching the error inside `cleanupOrphanedAnnotationQueueItems` per-project (or around the entire call here) and logging rather than propagating avoids this unrecoverable state.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(traces): clean up annotation queue i..."](https://github.com/langfuse/langfuse/commit/699aa59591e5788e6998bf5845c6867303bdb9b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45112235)</sub>

### DIFF
--- a/packages/shared/src/server/repositories/annotation-queue-items.ts
+++ b/packages/shared/src/server/repositories/annotation-queue-items.ts
@@ -1,0 +1,63 @@
+import { AnnotationQueueObjectType, prisma } from "../../db";
+import { getTraceIdsOlderThan } from "./traces";
+
+/**
+ * Delete annotation queue items that reference any of the given objects.
+ *
+ * Annotation queue items point at traces/observations via `objectId` with no
+ * foreign key to the underlying ClickHouse data. When that data is deleted
+ * (manual/batch trace deletion or data retention) the queue items would otherwise
+ * be left behind as orphans that render "Trace not found" in the review UI.
+ * Callers pass the ids that are being removed so the matching items are cleaned up
+ * in the same operation. See langfuse/langfuse#12852.
+ *
+ * @returns the number of deleted annotation queue items.
+ */
+export const deleteAnnotationQueueItemsByObjectIds = async ({
+  projectId,
+  objectType,
+  objectIds,
+}: {
+  projectId: string;
+  objectType: AnnotationQueueObjectType;
+  objectIds: string[];
+}): Promise<number> => {
+  if (objectIds.length === 0) return 0;
+
+  const { count } = await prisma.annotationQueueItem.deleteMany({
+    where: {
+      projectId,
+      objectType,
+      objectId: { in: objectIds },
+    },
+  });
+
+  return count;
+};
+
+/**
+ * Resolve which TRACE-type annotation queue items in a project reference traces
+ * older than `beforeDate` (i.e. traces about to be removed by data retention).
+ *
+ * Returns the expiring trace ids, which callers pass to
+ * `deleteAnnotationQueueItemsByObjectIds` AFTER the traces are deleted. It is
+ * resolved while the traces still exist so we never drop items for traces that
+ * survive the cutoff. Annotation queues are a small, curated set, so listing the
+ * referenced ids and filtering them by the cutoff stays cheap. Shared by the
+ * per-project retention job and the batch retention cleaner. See
+ * langfuse/langfuse#12852.
+ */
+export const getExpiredAnnotationQueueTraceIds = async (
+  projectId: string,
+  beforeDate: Date,
+): Promise<string[]> => {
+  const items = await prisma.annotationQueueItem.findMany({
+    where: { projectId, objectType: AnnotationQueueObjectType.TRACE },
+    select: { objectId: true },
+  });
+
+  const referencedTraceIds = [...new Set(items.map((item) => item.objectId))];
+  if (referencedTraceIds.length === 0) return [];
+
+  return getTraceIdsOlderThan(projectId, referencedTraceIds, beforeDate);
+};
--- a/packages/shared/src/server/repositories/index.ts
+++ b/packages/shared/src/server/repositories/index.ts
@@ -27,3 +27,4 @@ export * from "./job-executions";
 export * from "./daily-metrics";
 export * from "./dataset-runs";
 export * from "./score-configs";
+export * from "./annotation-queue-items";
--- a/packages/shared/src/server/repositories/traces.ts
+++ b/packages/shared/src/server/repositories/traces.ts
@@ -918,6 +918,38 @@ export const hasAnyTraceOlderThan = async (
   return rows.length > 0;
 };
 
+/**
+ * Given a set of candidate trace ids, returns the subset whose trace timestamp is
+ * older than `beforeDate`. Used by data retention to find annotation queue items
+ * that reference traces about to be removed, so those items can be cleaned up
+ * instead of being left as orphans. See langfuse/langfuse#12852.
+ */
+export const getTraceIdsOlderThan = async (
+  projectId: string,
+  traceIds: string[],
+  beforeDate: Date,
+): Promise<string[]> => {
+  if (traceIds.length === 0) return [];
+
+  const rows = await queryClickhouse<{ id: string }>({
+    query: `
+      SELECT DISTINCT id
+      FROM traces
+      WHERE project_id = {projectId: String}
+      AND id IN ({traceIds: Array(String)})
+      AND timestamp < {cutoffDate: DateTime64(3)}
+    `,
+    params: {
+      projectId,
+      traceIds,
+      cutoffDate: convertDateToClickhouseDateTime(beforeDate),
+    },
+    tags: { projectId },
+  });
+
+  return rows.map((row) => row.id);
+};
+
 export const deleteTracesOlderThanDays = async (
   projectId: string,
   beforeDate: Date,
--- a/worker/src/__tests__/annotationQueueTraceDeletionCleanup.test.ts
+++ b/worker/src/__tests__/annotationQueueTraceDeletionCleanup.test.ts
@@ -1,0 +1,194 @@
+import { expect, describe, it } from "vitest";
+import { randomUUID } from "crypto";
+import { Job } from "bullmq";
+import {
+  createTrace,
+  createTracesCh,
+  createOrgProjectAndApiKey,
+} from "@langfuse/shared/src/server";
+import { AnnotationQueueObjectType, prisma } from "@langfuse/shared/src/db";
+import { processPostgresTraceDelete } from "../features/traces/processPostgresTraceDelete";
+import { handleDataRetentionProcessingJob } from "../ee/dataRetention/handleDataRetentionProcessingJob";
+import { BatchDataRetentionCleaner } from "../features/batch-data-retention-cleaner";
+
+/**
+ * Regression coverage for langfuse/langfuse#12852.
+ *
+ * Annotation queue items reference traces via `objectId` with no foreign key to
+ * ClickHouse. Deleting the referenced trace (manually or via data retention) used
+ * to leave the queue item behind, so opening it in the UI failed with
+ * "Trace not found". These tests assert the items are cleaned up with the trace
+ * across every trace-deletion path: manual/batch deletion, the per-project
+ * retention job, and the batch retention cleaner.
+ */
+describe("annotation queue cleanup on trace deletion (#12852)", () => {
+  const createQueueWithItem = async (
+    projectId: string,
+    objectId: string,
+    objectType: AnnotationQueueObjectType = AnnotationQueueObjectType.TRACE,
+  ) => {
+    const queueId = randomUUID();
+    await prisma.annotationQueue.create({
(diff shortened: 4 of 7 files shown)


hassaanch23:
Rebased onto current `main` — this was conflicting after ~191 commits of drift, so it should be reviewable again now.

One real conflict, in `worker/src/features/batch-data-retention-cleaner/index.ts`, from the `workloads` → `observedWorkloads`/`selectedWorkloads` refactor. Resolved onto **`selectedWorkloads`**, since that's the leading `PROJECT_LIMIT` slice that actually gets deleted in the pass — scoping the annotation-queue lookup to `observedWorkloads` would have dropped queue items for projects whose traces survive the cutoff. Also dropped a now-stale `percentile` import that `main` had already removed.

No behavior change from the original PR otherwise. Verified locally on the rebased branch:

- `turbo run typecheck lint --filter=worker --filter=@langfuse/shared` → `Tasks: 7 successful, 7 total`
- `prettier --check` on all changed files → `All matched files use Prettier code style!`

The regression tests in `annotationQueueTraceDeletionCleanup.test.ts` are integration tests (real Postgres + ClickHouse), so I've left those to CI rather than asserting them from an environment without the infra.

PR #15230, opened 2026-07-20
fix(trace): gate oversized fields in the JSON Beta IO viewer (LFE-10847)

nkabardin:
## TL;DR
Large traces could still freeze the tab in the **JSON Beta** IO viewer. It virtualizes the DOM but builds the full node tree on the main thread first, so a payload with ~1M nodes (a big conversation / deeply nested JSON) wedges the browser before virtualization helps. This gates oversized fields on **node count** — over-limit fields show the same bounded preview + download fallback the plain JSON view already uses, so the trace still opens and the tab never freezes.

## Why
A customer reported a trace that crashes the browser (part of the large-traces resilience epic, LFE-10152). Digging in on the reproduction:

- The **plain JSON** view is already gated by *character count* (LFE-10989) and the **Formatted** view lazy-collapses, so both survive today.
- The **JSON Beta** viewer is the remaining gap: it has **no size gate**. It renders only visible rows, but `buildMultiSectionTree` still materializes *every* node on the main thread up front. A ~20 MB structured-JSON field is ~1M nodes → the tab wedges. This matches user reports of freezing when switching to the JSON view on large conversations.
- A character limit is the wrong metric for this viewer: a multi-MB *single string* (e.g. a base64 data-URI) is a single node and renders instantly as a media chip — a char gate would wrongly hide it.

## What
- Gate the Beta viewer on **node count** (`countJsonRows > JSON_VIEW_RENDER_ROW_LIMIT`, 50k). Over-limit fields skip the parse/decode/tree-build entirely and render the existing bounded `LargeJsonFieldFallback` (4k-char preview + raw download); everything else renders exactly as before.
- Fix the fallback guidance, which pointed users at the JSON Beta view that could itself freeze — it now points only at the lazy **Formatted** view.

## Result
Verified locally against seeded 20 MB payloads (v4):

| Field | Nodes | JSON Beta before | JSON Beta after |
|---|---|---|---|
| Structured JSON | ~1,000,000 | tab freezes | bounded fallback + download, no freeze |
| base64 data-URI | ~15 | renders (media chip) | renders (media chip) — unchanged |

`countJsonRows` on the 1M-node payload is ~23 ms, so the gate itself is cheap; the freeze was the tree build it now skips.

<details>
<summary>Mechanism, scope & verification</summary>

**Root cause.** `IOPreviewJSON` fed each field straight into `MultiSectionJsonViewer`, whose eager `buildMultiSectionTree` is O(node count) on the main thread. Virtualization bounds the *DOM*, not the *build*, so node count — not payload size — is what freezes the tab.

**Fix shape.** Per field: parse once, `countJsonRows`, and if over the limit render the fallback and exclude the field from the viewer's `sections` (so the tree builder never sees it). Non-gated fields keep the shared viewer (search, jump-to, virtualization). The header/nav are hidden when every field is gated. Reuses the existing `probeJsonField` + `LargeJsonFieldFallback` from the plain-view gate — one fallback UI, two size metrics (chars for the unvirtualized view, nodes for the virtualized one).

**Threshold.** 50k rows = 15× the viewer's own virtualization threshold (3333), far below the measured ~1M-node freeze; a 50k-node tree builds in tens of ms and virtualizes fine. Ordinary large traces are never gated.

**Tests.** New `IOPreviewJSON.clienttest.tsx` pins the branch selection: over-limit field → fallback and never handed to the viewer; huge single string → not gated (one node); mixed → only the over-limit field is gated; normal I/O → viewer as before. Confirmed the two gate assertions fail without the change.

**Verification.** `pnpm --filter web run typecheck` clean; `turbo run lint` 7/7; new + sibling gate tests 18 passed. Browser-checked both payloads in all three IO views on a local dev server.

**Scope.** Frontend resilience only — the complementary backend cap on how much IO is sent to the client is tracked separately.
</details>

This PR closes a browser-freeze gap in the **JSON Beta** IO viewer by adding a node-count gate (`JSON_VIEW_RENDER_ROW_LIMIT = 50 000`). Fields that exceed the limit are diverted to the existing `LargeJsonFieldFallback` (preview + download) before the O(N) tree-build in `MultiSectionJsonViewer` ever runs, preventing tab freezes on million-node payloads.

- Parses each field once with `deepParseJson`, counts rows via `countJsonRows`, and gates before the Unicode-decode and tree-build steps; non-gated fields continue through the existing viewer path unchanged.
- Hides the search/nav toolbar when all fields are gated (no viewer sections), and updates the fallback guidance to remove the now-incorrect pointer to \"JSON Beta.\"
- Adds four focused client tests that pin the gate logic: over-limit → fallback only; huge single string (one node) → viewer; mixed → only gated field uses fallback; normal I/O → viewer.
</details>

Safe to merge; the fix prevents a real browser freeze with no regressions to the normal rendering path.

The gate logic is correct and well-tested. Two minor issues exist: the Copy button in mixed mode silently omits gated fields from the clipboard payload, and `probeJsonField` receives the raw prop rather than the deeply-parsed value so the character count shown in the fallback can be misleading for double-encoded JSON strings.

The copy handler and the `probeJsonField` call site in `IOPreviewJSON.tsx` are worth a second look.
</details>

```mermaid
flowchart TD
    A[Field received input/output/metadata] --> B[deepParseJson → parsedValue]
    B --> C[countJsonRows on parsedValue]
    C --> D{rows > 50k?}
    D -- Yes --> E[inputTooLarge = true]
    D -- No --> F[inputTooLarge = false]
    E --> G[effectiveField = undefined]
    E --> H[probeJsonField on raw input]
    H --> I[LargeJsonFieldFallback - preview + download]
    F --> J[decodeUnicodeInJson → effectiveField]
    J --> K[Add to MultiSectionJsonViewer sections]
    K --> L{sections.length > 0?}
    L -- Yes --> M[Render header/nav + viewer]
    L -- No --> N[Render fallbacks only, no header/nav]
    I --> N
    I --> M
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[Field received input/output/metadata] --> B[deepParseJson → parsedValue]
    B --> C[countJsonRows on parsedValue]
    C --> D{rows > 50k?}
    D -- Yes --> E[inputTooLarge = true]
    D -- No --> F[inputTooLarge = false]
    E --> G[effectiveField = undefined]
    E --> H[probeJsonField on raw input]
    H --> I[LargeJsonFieldFallback - preview + download]
    F --> J[decodeUnicodeInJson → effectiveField]
    J --> K[Add to MultiSectionJsonViewer sections]
    K --> L{sections.length > 0?}
    L -- Yes --> M[Render header/nav + viewer]
    L -- No --> N[Render fallbacks only, no header/nav]
    I --> N
    I --> M
```

</a>
</details>

`````markdown

---

### Issue 1 of 2
web/src/components/trace/components/IOPreview/IOPreviewJSON.tsx:301-315
**Copy silently omits gated fields in mixed rendering mode**

When at least one field is under the row limit (so `hasViewerSections` is `true` and the toolbar renders), the Copy button is shown. But over-limit fields have `effectiveInput/Output/Metadata === undefined`, and `JSON.stringify({ input: undefined })` omits those keys entirely — so the copied JSON is silently incomplete. A user pasting it elsewhere won't see the gated fields at all. At minimum the button tooltip could reflect this ("Copy visible sections"), or gated fields could be included from their probe serialization.

### Issue 2 of 2
web/src/components/trace/components/IOPreview/IOPreviewJSON.tsx:189-201
**`probeJsonField` serializes the raw prop, but the row-count gate runs on the deeply-parsed value**

`countJsonRows` is called on `inputParsed` (`deepParseJson(input)`), which expands any embedded JSON strings into their full object trees. `probeJsonField` is then called with the original raw `input` prop. When a field is a string containing valid JSON (the "v4 session events" case mentioned in the comment), `deepParseJson` can turn it into a large object tree that trips the gate, while `probeJsonField` sees a short string and reports its `isString: true` length. The fallback would then display, for example, "100 characters — too large to render in JSON view" even though the actual content is tens of thousands of nodes. Passing `inputParsed` to `probeJsonField` instead would keep the node-count and the displayed size consistent.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): gate oversized fields in the..."](https://github.com/langfuse/langfuse/commit/7098a2a6ae225baa7d9c3287c9fbf41f5a31f2d9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45627775)</sub>

### DIFF
--- a/web/src/components/trace/components/IOPreview/IOPreviewJSON.clienttest.tsx
+++ b/web/src/components/trace/components/IOPreview/IOPreviewJSON.clienttest.tsx
@@ -101,11 +101,34 @@ describe("IOPreviewJSON node-count gating", () => {
       screen.getByRole("button", { name: /Download Input/i }),
     ).toBeInTheDocument();
     // ...and its section carries NO data, so the viewer builds no tree for it.
-    expect(screen.getByTestId("section-input")).toHaveAttribute(
+    // The gated section uses a collapse-isolated key (`__oversized`), never the
+    // plain field key, so it cannot inherit a persisted collapsed state.
+    expect(screen.getByTestId("section-input__oversized")).toHaveAttribute(
       "data-hide-data",
       "true",
     );
     expect(screen.queryByTestId("data-input")).not.toBeInTheDocument();
+  });
+
+  it("gates under a collapse-isolated key, not the plain field key", () => {
+    // The fallback's download escape hatch lives in the section footer, which
+    // the viewer hides when the section is collapsed and persists that collapse
+    // per section key. Reusing the plain field key would let a collapse from an
+    // earlier trace (where the field rendered normally) silently hide the
+    // fallback here. A distinct `__oversized` key keeps the two states separate.
+    render(
+      <IOPreviewJSON
+        input={manyRows()}
+        hideOutput
+        hideIfNull
+        showCorrections={false}
+        projectId="p"
+        traceId="t"
+      />,
+    );
+
+    expect(screen.getByTestId("section-input__oversized")).toBeInTheDocument();
+    expect(screen.queryByTestId("section-input")).not.toBeInTheDocument();
   });
 
   it("reports the node/row count as the reason, not a character count", () => {
@@ -165,7 +188,7 @@ describe("IOPreviewJSON node-count gating", () => {
     expect(
       screen.getByRole("button", { name: /Download Input/i }),
     ).toBeInTheDocument();
-    expect(screen.getByTestId("section-input")).toHaveAttribute(
+    expect(screen.getByTestId("section-input__oversized")).toHaveAttribute(
       "data-hide-data",
       "true",
     );
@@ -173,7 +196,7 @@ describe("IOPreviewJSON node-count gating", () => {
     expect(screen.getByTestId("data-output")).toBeInTheDocument();
     const sections = screen.getAllByTestId(/^section-/);
     expect(sections.map((el) => el.getAttribute("data-testid"))).toEqual([
-      "section-input",
+      "section-input__oversized",
       "section-output",
     ]);
   });
--- a/web/src/components/trace/components/IOPreview/IOPreviewJSON.clienttest.tsx
+++ b/web/src/components/trace/components/IOPreview/IOPreviewJSON.clienttest.tsx
@@ -2,20 +2,37 @@
  * The JSON Beta viewer virtualizes the DOM but still builds the full node tree
  * on the main thread, so a field with too many nodes (a large conversation /
  * deeply nested JSON) freezes the tab even here (LFE-10847). These tests pin the
- * node-count gate in IOPreviewJSON: over-limit fields render the bounded
- * fallback and are NEVER handed to MultiSectionJsonViewer, while normal fields
- * (including a huge single string, which is only one node) render as before.
+ * node-count gate in IOPreviewJSON: over-limit fields render as a `hideData`
+ * section (so the viewer builds no tree) whose footer is the bounded fallback,
+ * while normal fields (including a huge single string, which is only one node)
+ * render with their data as before.
  */
 import { render, screen } from "@testing-library/react";
 
-// MultiSectionJsonViewer is the tree-building path we must NOT reach for
-// over-limit fields — mock it and expose the section keys it receives.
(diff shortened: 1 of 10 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

PR #15239, opened 2026-07-20
feat(api): stream trace observation IO as bytes (LFE-11081)

nkabardin:
## TL;DR

Adds one additive, backend-only endpoint that streams a **single** trace-observation IO field (`input` | `output` | `metadata`) to the browser **as raw bytes**, piping ClickHouse's response body straight through without buffering the whole value in Node. This is the **transport primitive** for a future lazy/streaming large-JSON viewer. Purely additive; no existing endpoint or behavior changes. Spike/RFC-grade.

## Why

Very large single fields (multi-hundred-MB `input`/`output`) crash the browser today: the current read path materializes the field and the client fetches + `JSON.parse`s it in one shot (seed context: **LFE-10847** big-trace browser crashes, **LFE-10152** large-trace findings). The viewer we want (**LFE-11081**) needs to pull a field as a *byte stream* it can page/parse incrementally. That requires a byte-oriented transport first — this PR — before any UI work.

## What

**New endpoint (session-authed, project-scoped):**

```
GET /api/traces/:traceId/observations/:observationId/io/:field?projectId=...&startTime=<ISO-8601>
```

- `field` ∈ `input | output | metadata`. Responds `Content-Type: application/json; charset=utf-8`.
- Sits next to the existing `GET /api/traces/:traceId/download` route and reuses its **exact** authorization.

**Where the IO comes from:** v4 observation IO lives in the ClickHouse `events_full` table as raw `String` columns (`input`, `output`); `metadata` is reconstructed from the `metadata_names`/`metadata_values` arrays. This is the same source the `events.batchIO` read path uses. It is **not** offloaded to S3 — the field bytes are in ClickHouse — so we can stream straight from CH.

**How it streams:** a new repository fn `streamObservationIOFieldFromEventsTable` selects the one field with `FORMAT RawBLOB` and hands back ClickHouse's live HTTP body as a `Readable` (via the existing `queryClickhouseExecRaw`, the same machinery the Parquet blob-export uses). The route pipes that stream to the response with backpressure — **the field never fully materializes in the Node process**.

## Security model (the #1 review concern)

- **Authenticated + strictly project-scoped.** Authorization reuses the canonical `getAuthorizedTrace` (project membership, public trace, public session, or admin + admin-access webhook) — identical to the trace download route. No new auth surface is invented.
[...]

PR #15265, opened 2026-07-21
feat(json-viewer): async virtualized lazy JSON renderer (LFE-11080)

nkabardin:
## TL;DR

The frontend of the real fix for "big trace crashes the browser": a JSON viewer that never parses, builds, or holds more than what is expanded and visible — so it renders arbitrarily large JSON without freezing the tab. This PR is the reusable stack (byte engine → async source → tree model → **virtualized renderer**), written **once** against an async seam so the same renderer drives an in-memory value today and a Worker-backed ~1 GB stream later, unchanged. The trace-view integration is stacked on top in **#15322**.

## Why

The current JSON Beta viewer virtualizes the *DOM* (only visible rows paint) but still builds the **entire** node tree up front, on the main thread, before any paint. A ~20 MB structured payload is ~1M nodes → the build freezes the tab; `JSON.parse` alone blows heap 5–8.5× and hard-walls at V8's ~512 MB string cap. The shipped gate (#15230) is an interim stopgap that *refuses* to render oversized fields. This is the actual fix: **cost proportional to what is expanded/visible, never to total size.**

## What

One engine, one model, one renderer:

```
raw UTF-8 bytes ─▶ ByteJsonIndexEngine ─▶ AsyncJsonSource ─▶ TreeRowModel ─▶ React renderer
```

- **Byte engine** — a UTF-8 byte indexer (never a whole-doc JS string, never a whole-doc `JSON.parse`). `load()` is O(1); a container is scanned **once** on expand and its columnar child-offset table cached (pages after the first are O(page)); values are materialized by slicing + `TextDecoder`, precision preserved (bigint / raw string). A `ByteScanner` seam lets a WASM hot-loop drop in later.
- **Async source + tree model** — a nodeId-keyed async seam (`root` / `childrenPage` / `getValue`) and the one flatten/expand/paginate implementation over it. Revision-stamped windows for stale-read detection; `getValue` error envelope; wide containers page with a "load more" row.
- **Renderer** (`lazy/react/`) — a per-mount vanilla Zustand store owns the async model lifecycle and all actions; `@tanstack/react-virtual` positions row shells and reports the visible range back; rows are view-only and memoized. The `buildModel` seam is where a Worker source plugs in with no renderer change.

## Result

Demoable in Storybook (`AdvancedJsonViewer/LazyJsonViewer`): a 20,000-element array keeps only ~65 rows in the DOM (visible + overscan), expand/collapse/load-more/copy all work, and a 400-deep chain stays readable. Store logic + async-race correctness are covered by client tests; verified in-browser. The trace-view integration (#15322) already renders real oversized trace fields (258.5K-row input, 129.9K-row output) through this renderer.

<details>
<summary>Engineering detail — correctness under a real async source, and what's next</summary>

Adversarially reviewed and hardened for the day a Worker source replaces the in-process one (genuinely async, out-of-order responses):
- each windowed fetch merges at **its own** captured offset (out-of-order windows can't land at the wrong indices);
- the fetch trigger reacts to range **growth in place** on expand, not just scroll;
- structural mutations are **serialized** (the tree mutation isn't reentrant — concurrent expands would double-page children), with a per-node pending indicator;
- byte-engine failures are captured to Sentry (`area: json-viewer`); copy handles bigint + truncated (>cap) leaves; deep indent is capped.

**Stacked follow-ups:** #15322 integrates this into IOPreview (main-thread, replacing the JSON-Beta dead-end). Later: the Worker source for the streamed GB tail (consuming the #15239 byte endpoint), in-viewer find, size tiering + instrumentation, and retiring the gate. Two seam-level gaps for the Worker path: `RowModel` has no dispose/cancel, and no progress signal.

Design notes: `web/src/components/ui/AdvancedJsonViewer/LAZY_TREE_DESIGN.md` and `lazy/react/README.md`.
</details>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15279, opened 2026-07-21
feat(web): banner to prompt reload on a new deploy (LFE-10978)

nkabardin:
## TL;DR
A persistent, dismissible **"Langfuse just got an update — Reload"** notification that appears when a tab has been open across a deploy. It's a floating, frosted-glass pill (center-top, Apple-notification style), not a layout-pushing bar. Users refresh on their own terms; we never auto-reload. This is **§4.9 part 1 (banner)** of the Sentry-noise workstream; chunk-load / worker-retry resilience is a deliberate sibling follow-up.

## Why
Frequent deploys + long-lived tabs leave users running a stale JS bundle. When the app then tries to load a code-split chunk that no longer exists, it 404s and mints stale-fingerprint error-tracker noise — plus a broken experience. A visible "reload to update" prompt shrinks both. (Infra can't retain old chunks server-side, so the banner *is* the mitigation, not a nicety.)

## What
- **Detection reuses the existing seam.** The tRPC `buildIdLink` now records the `x-build-id` response header on **every** response (success and error, not just the 404 path) into a small external store read via `useSyncExternalStore` — no polling, no effect state-sync.
- **Conservative mismatch predicate:** "update available" only when both the running build id (`NEXT_PUBLIC_BUILD_ID`) and the observed one are present **and differ**. Either missing (e.g. a self-hosted build with no build id) stays silent.
- **Rolling-deploy robust:** during a rollout a tab sees responses from both old and new pods, in any order, with opaque build hashes. The store asks only "have we ever seen a build id ≠ ours?" — so the signal is **sticky** (an old-pod response can't flip it back off) and **flap-free** (re-observing a seen id does nothing).
- **Floating pill, not a bar:** renders into the top-most overlay `toast` layer, floating center-top over content (no chrome shift). It **offsets below** any full-width top banner (e.g. the payment banner) via `top-banner-offset`, so it doesn't overlap.
- **Clean, single entrance:** a grace period holds the banner out of the app's startup churn (hydration flicker + StrictMode double-mount) so its animation plays exactly once; `fill-mode-both` prevents a start-frame flash. Both the settle-gate and the once-per-appearance analytics guard live in module scope, so the banner survives remounts (e.g. AppLayout ↔ MinimalLayout) without re-hiding or double-counting.
- **Never auto-reloads** (would discard unsaved work — open annotations, editors). Dismiss hides it for the session; a genuinely newer build id re-shows it.
- **Analytics:** `version_update:banner_shown` (once per appearance) + `reload_clicked` / `dismissed`.

## How to see it
- **Storybook:** story `Default` in `VersionUpdateBannerView.stories.tsx` renders the pill in isolation (Reload + dismiss ×).
- **Live:** keep the app open on one build, deploy another (responses start returning a different `x-build-id`) — the pill appears ~5s after the page settles. Locally, force it by serving a build id from the tRPC seam different from `NEXT_PUBLIC_BUILD_ID`.

<details>
<summary>Files, design decisions & verification</summary>

**Feature module** `web/src/features/version-update/`:
- `versionUpdateStore.ts` — `isVersionMismatch(running, observed)` predicate + `createVersionUpdateStore(getRunningBuildId)` factory (injected running id → testable) + the app-wide singleton wired to `process.env.NEXT_PUBLIC_BUILD_ID`. Sticky `updateAvailable`, `dismissed`, a seen-ids set (flap-free), and `markShownReported()` (banner_shown once-per-appearance guard, module-level so it survives remounts). Cached boolean snapshot for `useSyncExternalStore`; `getServerSnapshot` always `false`.
- `useAppSettled.ts` — module-level startup grace-period gate (factory + singleton external store); the settled flag survives banner remounts so the grace period never restarts.
- `useVersionUpdateAvailable.ts` — the store hook.
- `VersionUpdateBannerView.tsx` — presentational, props-only (`onReload`/`onDismiss`) → renders in isolation. Frosted-glass pill, `top-banner-offset`, entrance animation with `fill-mode-both`.
- `VersionUpdateBanner.tsx` — connected: `isVisible = updateAvailable && appSettled`, portals into the `toast` layer, reports analytics, wires reload/dismiss.

**Wiring:** `api.ts` `buildIdLink` feeds the store on every response; `AuthenticatedLayout.tsx` mounts `<VersionUpdateBanner />`.

**Verification:** typecheck + lint clean; **30 client tests pass** (predicate, store report/dismiss/sticky/flap + markShownReported, the settle gate, and connected-banner render/analytics/gating). Entrance animation verified in Chrome + Firefox (single clean play); the remount + offset fixes address the two review nits.
</details>

### DIFF
--- a/web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
+++ b/web/src/components/layouts/app-layout/variants/AuthenticatedLayout.tsx
@@ -13,6 +13,7 @@ import { SidebarPresenceProvider } from "@/src/components/nav/sidebar-presence";
 import { Toaster } from "@/src/components/ui/sonner";
 import { Layer } from "@/src/components/ui/layer";
 import { TopBannerProvider } from "@/src/features/top-banner";
+import { VersionUpdateBanner } from "@/src/features/version-update";
 import { AppContentWithRightDrawer } from "../right-drawer/AppContentWithRightDrawer";
 import { ThemeToggle } from "@/src/features/theming/ThemeToggle";
 import {
@@ -173,6 +174,7 @@ export function AuthenticatedLayout({
           <SidebarProvider>
             <div className="flex h-dvh w-full flex-col">
               <PaymentBanner />
+              <VersionUpdateBanner />
               <div className="pt-banner-offset flex min-h-0 flex-1">
                 <AppSidebar
                   navItems={navigation.mainNavigation}
--- a/web/src/features/posthog-analytics/usePostHogClientCapture.ts
+++ b/web/src/features/posthog-analytics/usePostHogClientCapture.ts
@@ -232,6 +232,10 @@ export const events = {
     "compare_run_added",
     "compare_run_removed",
   ],
+  // Version-update reload notification (LFE-10978). `banner_shown` fires once
+  // per appearance; the two actions measure the reload-vs-dismiss split. No
+  // props carry user content.
+  version_update: ["banner_shown", "reload_clicked", "dismissed"],
   notification: ["click_link", "dismiss_notification"],
   toast: ["report_issue", "dismiss"],
   tag: [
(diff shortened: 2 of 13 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15303, opened 2026-07-22
feat(components): one-line self-measuring OverflowActionBar (LFE-11067)

nkabardin:
## TL;DR
The engine for the mobile "second top bar" you picked: a **one-line action bar that measures itself and spills overflow into a badge-counted `⋯` menu**, growing items back as space allows. Pinned actions (Ask AI) never overflow. This PR is the reusable primitive + Storybook; wiring it into the mobile shell (and hoisting each page's toolbar into it) is the next slice.

## Why
The mobile top was overloaded because page actions (time range, refresh, view toggle, filters, columns, export, Ask AI, …) just wrap onto several rows. We want them on exactly one line, with the rest one tap away — the pattern Grafana/Superset/Sentry all converge on (measured overflow into a "More" menu).

## What
`OverflowActionBar` takes an ordered list of actions and:
- renders the prefix that fits, spills the rest into a `⋯` dropdown with a **count badge**;
- **grows items back inline** when width increases (bidirectional);
- keeps `pinned` actions (Ask AI) always inline;
- measures **container width** (ResizeObserver), not the viewport — so it reacts when the filter sidebar opens.

## Result
Strictly one line at every width. Verified in Storybook across 240–680px: actions progressively move into `⋯(n)`, Ask AI stays put, the `Filters (2)` badge shows inline then moves into the menu. Typecheck + lint green.

<details>
<summary>Mechanism & notes</summary>

- A hidden, `aria-hidden`, absolutely-positioned measurement row renders every item once to capture natural widths; a `useLayoutEffect` computes the fitting prefix (reserving space for the trigger + pinned items) and sets `visibleCount`. The visible row is `invisible` until the first measurement lands, so there's no flash of the un-trimmed set.
- Filters intentionally are **not** a special case here — they're just another action (a `Filters (n)` button) that can overflow; where the filter *panel* itself lives on mobile is a separate decision.
- Overflowed actions expose an `overflowLabel` + `onSelect` for their menu representation. Complex controls (e.g. the date picker) would stay pinned or open their own modal — the research-backed "date picker → modal on phone" refinement lands with integration.
- Not yet wired into a page — that's deliberate (validate the mechanic in isolation first). Stories: `Default`, `Overflowing`, `AcrossWidths`.

</details>

This PR adds a self-measuring one-line action bar for mobile toolbars. The main changes are:

- A reusable component that moves trailing actions into a counted overflow menu.
- Support for pinned actions that remain inline.
- Container-based resize measurement for growing actions back into the row.
- Storybook examples across desktop and mobile widths.
</details>

Dynamic and effectful actions can behave incorrectly and need fixes before production integration.

- The hidden measurement row mounts action controls twice.
- Width changes inside an action can leave the overflow state stale.
- Overflowable actions without handlers become enabled no-op menu entries.

web/src/components/OverflowActionBar.tsx
</details>

`````markdown

---

### Issue 1 of 4
web/src/components/OverflowActionBar.tsx:176
**Hidden Controls Mount Twice**

The measurement row mounts every supplied `content` node in addition to its visible instance. When an action contains a date picker, dialog, or another effectful control, the hidden copy can duplicate IDs, subscriptions, requests, and portals because `aria-hidden` and `invisible` only hide DOM output.

### Issue 2 of 4
web/src/components/OverflowActionBar.tsx:70
**Content Changes Keep Stale Widths**

This calculation does not rerun when an action changes width without changing the container or `actions` references. A stateful date label, growing count badge, or newly loaded font can therefore leave too many items inline, and `overflow-hidden` clips the row instead of moving the item into the menu.

### Issue 3 of 4
web/src/components/OverflowActionBar.tsx:144
**Overflow Action Becomes Inert**

A non-pinned action may omit `onSelect`, but this path still renders it as an enabled menu item. At widths where that action overflows, selecting it closes the menu without invoking the working inline control, so the action silently stops functioning.

### Issue 4 of 4
web/src/components/OverflowActionBar.tsx:68
**Serialized Keys Hide Action Changes**

Joining keys with `|` makes distinct lists such as `['view|table', 'export']` and `['view', 'table|export']` share the same effect dependency. If a caller switches between those valid key sets while preserving the array partition references, the overflow calculation can keep the previous visible count and show the wrong action set.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(components): OverflowActionBar — on..."](https://github.com/langfuse/langfuse/commit/ed9872213e5ce11a33fd5c48b16075663c9b8013) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46290157)</sub>

**Context used:**

**Learned From**

### DIFF
--- a/web/src/components/OverflowActionBar.stories.tsx
+++ b/web/src/components/OverflowActionBar.stories.tsx
@@ -1,0 +1,159 @@
+import { fn } from "storybook/test";
+import {
+  BarChart3,
+  Bookmark,
+  Columns3,
+  Download,
+  Filter,
+  RefreshCw,
+  Sparkles,
+  Table,
+} from "lucide-react";
+
+import preview from "../../.storybook/preview";
+import { Button } from "@/src/components/ui/button";
+import { OverflowActionBar, type OverflowAction } from "./OverflowActionBar";
+
+const meta = preview.meta({
+  component: OverflowActionBar,
+});
+
+/** A realistic Traces-page action set: date range, refresh, view toggle,
+ *  filters (with a count), views, columns, export, and a pinned "Ask AI". */
+const sampleActions = (): OverflowAction[] => {
+  const barButton = (label: string, icon: React.ReactNode) => (
+    <Button variant="outline" size="sm" className="gap-1.5" onClick={fn()}>
+      {icon}
+      {label}
+    </Button>
+  );
+  return [
+    {
+      key: "date",
+      content: barButton("Past 1 day", null),
+      overflowLabel: "Past 1 day",
+      onSelect: fn(),
+    },
+    {
+      key: "refresh",
+      content: barButton("Off", <RefreshCw className="size-3.5" />),
+      overflowLabel: (
+        <>
+          <RefreshCw className="size-4" /> Auto-refresh
+        </>
+      ),
+      onSelect: fn(),
+    },
+    {
+      key: "view",
+      content: barButton("Table", <Table className="size-3.5" />),
+      overflowLabel: (
+        <>
+          <BarChart3 className="size-4" /> Chart view
+        </>
+      ),
+      onSelect: fn(),
+    },
+    {
+      key: "filters",
+      content: (
+        <Button variant="outline" size="sm" className="gap-1.5" onClick={fn()}>
+          <Filter className="size-3.5" />
+          Filters
+          <span className="bg-primary text-primary-foreground ml-0.5 flex h-4 min-w-4 items-center justify-center rounded-full px-1 text-[0.625rem] leading-none">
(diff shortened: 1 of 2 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
Closing for now: OverflowActionBar isn't consumed anywhere yet (net-new primitive built ahead of the page-header rewire). The reviews surfaced real robustness gaps (double-mounting effectful content in the measurement row, stale widths on dynamic-width actions, inert overflow actions) that are best designed against a real consumer. Will revive + harden this when the header-rewire follow-up actually wires it up.

PR #15305, opened 2026-07-22
feat(search-bar): dogfood the Ask AI prompt as a managed Langfuse prompt (LFE-11004)

nkabardin:
> Stacked on #15292 (prompt detangle) — this PR's base is that branch, so the diff shows only the managed-prompt layer. Retargets to `main` once #15292 merges.

## TL;DR
Our v4 "Ask AI" filter prompt lived only in code, so we could not iterate it in our own product. This makes it a **managed Langfuse prompt** in our internal AI-features project: the app fetches and compiles it at runtime, links the exact prompt version onto every traced generation, and falls back to the in-code prompt whenever the managed one is unavailable. The prompt becomes editable in the Langfuse UI, and every generation's trace shows which version produced it — the foundational step for the whole Ask-AI instrumentation workstream (LFE-10716).

## Why
We self-trace every Ask AI generation into our internal AI-features project, but the prompt itself could only be changed in code — we were building an eval/observe loop around a prompt we could not iterate in the product. The legacy v3 filter already dogfoods a managed prompt; the v4 bar did not. This closes that gap.

## What
- New `resolveFilterPrompt.ts` fetches the `search-bar-filter` chat prompt from the AI-features project (same client the v3 path uses) and compiles it with **code-generated** `{{catalog}}` / `{{nullable_ids}}` / `{{current_datetime}}` — the catalog stays code-derived from the field registry, so it can never drift from the grammar.
- The managed prose becomes the source of truth (edit it in the UI); the in-code `buildFilterSystemPrompt` remains the **fallback** for self-hosted, missing keys, or any fetch/compile error. The path **never throws** — a managed-prompt outage degrades to today's behavior.
- The prompt version is linked onto the trace when the managed prompt served the request.
- A repo-versioned seed (`search-bar-filter.prompt.json`) + a human-run, per-region sync script provision the initial prompt. The seed reproduces today's prompt **byte-for-byte** (locked by a regression test), so the first push is behavior-neutral; drift is intentional thereafter.

## Result
The prompt is now dogfooded through our own prompt management, with per-generation version links, and no behavior change on rollout.

<details>
<summary>Review (skeptic: FIX → addressed) & verification</summary>

- **Seed ↔ code equivalence** verified byte-for-byte (16550 bytes) and locked by a regression test that recompiles the committed seed against the registry and asserts equality — catches any future seed prose / score-column change before it ships.
- **Fallback** covers every failure mode (no keys, self-hosted, 404, 5xx, network timeout, compile error, wrong prompt type) → code path, never a 500.
- **Decoupled the managed-prompt read from telemetry consent** (matches v3): reading our own prompt sends no org data out, so `aiTelemetryEnabled` now gates only the trace write + version link, not whether the improved prompt is used.
- **Cold-start latency capped** (`fetchTimeoutMs` + `maxRetries: 0`); the SDK cache is stale-while-revalidate so steady state adds ~0.
- **Prompt name single-sourced** (sync script derives it from the seed; a test ties the code constant to the seed) so a rename can't silently fall back.
- Typecheck / lint / targeted tests green. No eval harness yet (LFE-11005), so residual prompt-shape reasoning is code-level, not measured.

**Merge/deploy note:** push the seed to all regions **with or before** the deploy that carries this merge — otherwise, between merge and seeding, every eligible generation logs a "prompt not found" fallback warning (behavior is still correct, just noisy).
</details>

Related: LFE-10716 (workstream), #15292 (detangle — base of this PR).

This PR moves the v4 Ask AI filter prompt into managed Langfuse prompts. The main changes are:

- Runtime prompt fetching and compilation with a local fallback.
- Registry-derived catalog, nullable fields, and request-time datetime variables.
- Managed prompt version links on traced generations.
- A versioned prompt seed and regional synchronization script.
- Tests covering fallback behavior and seed equivalence.
</details>

Malformed managed chat messages can bypass the fallback and fail Ask AI requests.

- Fetch and compile exceptions correctly return the local prompt.
- The compiled result is checked only for being an array, not for containing valid chat messages.
- A bad live prompt version can therefore cause a provider or message-mapping error.

web/src/features/search-bar/server/resolveFilterPrompt.ts
</details>

`````markdown

---

### Issue 1 of 1
web/src/features/search-bar/server/resolveFilterPrompt.ts:132-142
**Malformed Chat Escapes Fallback**

The guard accepts any array, including an empty array or entries with unsupported roles or non-string content. A malformed live prompt edit can therefore reach `generateLangfuseAIText` and fail the request instead of using the local fallback; validate the compiled message structure before returning `usedPrompt`.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(search-bar): decouple managed Ask AI..."](https://github.com/langfuse/langfuse/commit/fa8d8fafb94989e19d7884e98c570e3358e4eb92) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46293579)</sub>


nkabardin:
@claude review

PR #15315, opened 2026-07-22
fix(search): don't apply incomplete filter tokens to the query (LFE-11017)

nkabardin:
## TL;DR

Typing a partial filter into the trace search bar and stopping mid-way (e.g. `type`, a field you started filtering on) silently applied it as a full-text search and returned **"No results"** — with no error shown. This makes a lone bare field-name token an **incomplete filter**: it renders red (the existing error affordance) and is **excluded from the query** until you finish it (or quote it to search as literal text). Multi-word free-text search is untouched.

## Why

The bar treats bare text as a full-text `searchQuery`. A single word that is actually a **field name** (`type`, `level`, `env`, …) is almost always a filter the user started and hasn't finished — not free text. But it parsed as valid free text, so on the next commit (blur/Enter) it lowered to `searchQuery=type`, AND-joined with any existing filter, and wiped the results to "No results" with **no signal**. The colon form `type:` already errored (red chip); the bare form did not — an inconsistent, silent failure.

## What

A lone unquoted free-text token that resolves to a field name is now flagged as an **incomplete filter** at the single validation gate (`validate.ts` → `validateQuery`), which drives all three consumers at once: the red chip, the commit block, and the red input border. This mirrors the existing dangling dot-prefix guard (`metadata.`) already in the adapter, added for the identical "don't silently set searchQuery to a partial token" reason.

- **Scoped to the sole free-text token**, so a deliberate multi-word phrase that happens to contain a field word (`type error`, `session timeout`) still searches as free text — those lower to one contiguous-substring `searchQuery`, not a filter.
- **Quoting is the escape hatch**: `"type"` searches for the literal word, the same way reserved keywords (`and`/`or`/`not`) already work. The serializer force-quotes a field-name free-text value so a legacy/bookmarked `searchQuery=type` round-trips back as valid literal text instead of re-deriving as a red chip.

## Result

- Partial token `type` → **not applied** (URL `search` param stays empty) and shown red with a validation indicator.
- Complete `type:chat` → applies as before.
- `type:` (colon, no value) → keeps its existing "Missing value" error.
- Free text `hello` and phrases like `type error` → still apply as full-text search.

<details>
<summary>Mechanism, decision, and live verification</summary>

**Mechanism.** `semanticDiagnostics` (whole-AST) collects free-text nodes; if there is exactly one, it is unquoted, and it resolves to a field, it emits an `error`-severity diagnostic on that token's span. Because `validateQuery` is the single gate behind `planCommit` (commit), `deriveComposerSegments` (chip classification), and the store's `draftValid` (input border), one check covers all three — no drift. The mirror half lives in `serialize`'s free-text branch: a field-name value is force-quoted so `filterStateToQueryText → serialize → parse` stays stable and valid.

**Applied-vs-error decision.** The task allowed either "don't apply" or "show a validation error". The existing model for `type:` does **both** (red + commit-blocked), so this matches it: the token is both excluded from the query and surfaced as an error. Scoping to the single free-text token is the deliberate choice that keeps multi-word free-text phrases working.

**Tests.** Added unit coverage in `validate.clienttest.ts` (lone field word flagged; repro `name:foo type`; `type:` unchanged; phrase `type error` not flagged; quoted `"type"` valid; complete expression valid; field-word free-text round-trips quoted). Added `type`/`level` to the round-trip property harness `freeTextValues` (`searchBarInvariants.clienttest.ts`) to lock the serialize↔parse mirror (INV-3). All 257 search-bar unit tests pass; typecheck + lint clean.

**Live (Chrome, v4 traces).** Reproduced the exact flow. Free text `hello` → committed `?search=hello` (free text still applies). Bare `type` on blur → `search` stays empty (not applied) and the token renders as a red dashed pill with the bar's red border + error icon. Phrase `type error` → committed `?search=type error`, not flagged. Screenshot confirms the red chip.

</details>

This PR prevents incomplete field tokens from being applied as free-text searches. The main changes are:

- Adds a validation error for lone, unquoted field names.
- Quotes field-name text during serialization to preserve literal searches.
- Extracts a shared quoting helper.
- Adds validation and round-trip tests.
</details>

This looks safe to merge.

- No blocking issues found in the changed code.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(search): don&#39;t apply incomplete filt..."](https://github.com/langfuse/langfuse/commit/b940f73d39377c39e5422cf2734ee3bdf9dd38ca) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46342416)</sub>

**Context used:**

**Learned From**

### DIFF
--- a/web/src/features/search-bar/lib/langQ.ts
+++ b/web/src/features/search-bar/lib/langQ.ts
@@ -19,7 +19,7 @@
 
 import type { ASTNode, CompareOp, FilterNode, Span, TextNode } from "./ast";
 import { canonicalKey, operatorIssue, resolveField } from "./fields";
-import { NEEDS_QUOTES, unquote } from "./quoting";
+import { NEEDS_QUOTES, quote, unquote } from "./quoting";
 
 // Re-exported for back-compat: quoting primitives now live in the shared
 // dependency-free `quoting.ts` (so `fields.ts` can use them too).
@@ -854,8 +854,7 @@ export function serializeValue(value: string): string {
     value.startsWith("!") ||
     RESERVED_BARE_TOKEN.test(value)
   ) {
-    // Escape \ before " — must be the exact inverse of unquote.
-    return `"${value.replace(/\\/g, "\\\\").replace(/"/g, '\\"')}"`;
+    return quote(value);
   }
   return value;
 }
@@ -924,8 +923,15 @@ export function serialize(ast: ASTNode | null): string {
       return serializeFilter(ast);
     case "text":
       // Always route through serializeValue so bare keywords (AND/OR/NOT) and
-      // leading-operator/hyphen free text get quoted and reparse as text.
-      return serializeValue(ast.value);
+      // leading-operator/hyphen free text get quoted and reparse as text. A lone
+      // word that resolves to a field name is treated as an incomplete filter by
+      // the validator (LFE-11017), so it must serialize QUOTED too — otherwise a
+      // committed free-text searchQuery like `type` (e.g. a legacy URL/saved
+      // view) re-derives as a red chip instead of a valid literal search. Mirror
+      // invariant with validate.ts's incomplete-field-token check.
+      return resolveField(ast.value) !== null
+        ? quote(ast.value)
+        : serializeValue(ast.value);
     case "not": {
       const child = ast.child;
       if (child.kind === "filter") return `-${serializeFilter(child)}`;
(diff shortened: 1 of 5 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15320, opened 2026-07-22
ci(observability): register Sentry release + deploy marker on prod deploy (LFE-14421)

nkabardin:
## TL;DR
Register a **Sentry release + deploy marker** on every prod deploy, so we can finally ask Sentry *"what did this deploy introduce?"* — per region, in the right Sentry org/project automatically. No app-code change, best-effort (never blocks a deploy).

## Why
Sentry events already carry a `release` tag — `Sentry.init({ release: process.env.NEXT_PUBLIC_BUILD_ID })`, and the prod Docker build bakes `NEXT_PUBLIC_BUILD_ID = github.sha`. But **nothing creates the release _object_** in Sentry, so the Releases dashboard is empty and we get none of the payoff: the **Regressions** view ("first seen in `<release>`"), **deploy markers** on the timeline, **resolve-in-release**, and **auto-reopen** when a fixed issue recurs in a later release. Today, separating pre- vs post-deploy issues means eyeballing `firstSeen` timestamps — which is exactly the wall I hit triaging the v3.224.0 deploy.

This is a Sentry-noise-workstream lever (LFE-10974): sharper triage and regression detection = signal, not noise.

## What
A best-effort **`Register Sentry release`** step in the reusable prod-deploy workflow (`_deploy_ecs_service.yml`), using `getsentry/action-release` (SHA-pinned to v3.7.0). Release identifier = `github.sha` (matches `NEXT_PUBLIC_BUILD_ID` and the event `release` tag).

The deploy workflow already fans out **per environment** (`prod-eu`, `prod-us`, `prod-hipaa`, `prod-jp`) with per-environment `vars.SENTRY_ORG` / `vars.SENTRY_PROJECT` / `secrets.SENTRY_AUTH_TOKEN`, so the release lands in the **correct Sentry org + project per region automatically** — the US org's `langfuse` / `langfuse-us` projects, and the separate EU org — with zero hardcoding.

## Result
Once merged and deployed, each region's Sentry gets a finalized release + a production deploy marker per build, unlocking the Regressions/deploy views. No source maps are uploaded (they're served publicly, #15277). No application code changes.

## ⚠️ One config prerequisite
Each environment's `SENTRY_AUTH_TOKEN` must have access to **that environment's** Sentry org. **The EU org is separate**, so its environment needs an EU-org-scoped token if the current one is single-org. The step is `continue-on-error`, so a missing/mis-scoped token just **skips that region's release** rather than failing the deploy — safe to merge before the token is provisioned.

<details>
[...]

PR #15324, opened 2026-07-22
fix(observability): re-enable Sentry source-map upload w/ debug IDs (LFE-14423)

nkabardin:
## TL;DR
Prod Sentry stack traces are fully minified and untriageable. The cause is an **accidental regression in the Sentry v8→v10 upgrade (#8934)** — upload was silently disabled ~10 months ago. Re-enable source-map **upload with debug IDs**; this restores behaviour that already worked across all regions/orgs/projects under v8. One-config change; public serving (for DevTools) stays.

## Why (root cause)
Verified on a live prod event: frames are `app:///_next/static/chunks/<hash>.js`, functions `cc` / `ue`, no source context.

`git` history shows the real cause: the **v8→v10 upgrade (#8934, Sep 2025)** changed
```
- hideSourceMaps: true            // v8: upload to Sentry, then hide from the public bundle
+ sourcemaps: { disable: true }   // v10: do NOT upload at all
```
Those aren't equivalent — the correct v10 translation of `hideSourceMaps: true` is `deleteSourcemapsAfterUpload: true`. So **upload was accidentally turned off**, and Sentry stacks have been minified ever since. (#15277 later added public map serving on the theory Sentry would fetch them via `sourceMappingURL`; it can't — the SDK rewrites frames to the `app:///` scheme, which isn't a fetchable URL, so those public maps only help browser DevTools.)

## What
Re-enable the plugin's upload in `web/next.config.mjs` (`sourcemaps: { disable: true }` → `sourcemaps: { deleteSourcemapsAfterUpload: false }`). The plugin injects **debug IDs** into the JS + maps and uploads the maps; Sentry matches map→event by the embedded id, independent of URLs and the `app:///` rewrite. Unlike the old `hideSourceMaps`, we keep serving maps publicly (`productionBrowserSourceMaps`) for DevTools, so we don't delete them after upload.

## Is this safe for our multi-region Sentry setup? (2 orgs, 3 projects)
**Yes — because this exact upload already worked across that topology under v8**, until #8934 disabled it. It reads the *same* per-region `SENTRY_ORG` / `SENTRY_PROJECT` / `SENTRY_AUTH_TOKEN` the v8 upload used — these are **environment-scoped** in GitHub (each of `prod-eu`, `prod-us`, `prod-hipaa`, `prod-jp` has its own), so each region's build uploads to its own org/project with its own token. The separate EU org already has its own env-scoped token. Nothing is hardcoded; no cross-contamination (each build targets exactly one org/project). Source maps are compiled code — no PII/HIPAA concern in uploading. And upload errors are **non-fatal** (logged, build continues), so a mis-scoped region degrades to "no symbolication for that region," never a failed deploy.

<details>
<summary>Details & verification</summary>

- **Build engine**: Webpack (`next build`, no `--turbopack`) + `@sentry/nextjs@^10` → debug-ID injection + upload fully supported.
- **Release binding**: `SENTRY_RELEASE = NEXT_PUBLIC_BUILD_ID = github.sha`, so uploaded artifacts tie to the same release the events are tagged with (and to #15320's release/deploy marker).
- **Non-prod**: builds without `SENTRY_AUTH_TOKEN` skip upload gracefully.
- **Post-merge verification** (the one residual, since it's been off ~10 months and the topology is "a bit of a mess"): after a prod deploy, confirm in **each** region's Sentry project that a fresh client error symbolicates to real filenames/functions — paying special attention to the **separate EU org** and to each env's `SENTRY_PROJECT` matching its `NEXT_PUBLIC_SENTRY_DSN` project (events + uploaded maps must be the same project).

Trade-off: upload + `widenClientFileUpload` add some build time — the accepted cost for legible Sentry stacks.
</details>

Linear: LFE-14423 (sub-task of LFE-10974). Restores upload disabled by #8934; complements #15320 (release + deploy marker). Supersedes the source-maps intent of #15277 (public serving stays for DevTools).


nkabardin:
@claude review

nkabardin:
@claude review

PR #15350, opened 2026-07-23
feat(ingestion): move oversized direct-event fields to media

hassiebp:
🟡 **Live preview: [pr-15350.preview.langfuse.com](https://pr-15350.preview.langfuse.com)**

## What

- move observation input, output, and each persisted metadata value above a configurable per-field UTF-8 byte limit to media when writing directly to `events_full`
- gate the entire overflow path behind `LANGFUSE_OBSERVATION_FIELD_OVERFLOW_ENABLED`; default to disabled (`false`)
- configure the limit with `LANGFUSE_OBSERVATION_FIELD_SIZE_LIMIT_BYTES`; default to 2 MiB (`2097152` bytes)
- replace successfully uploaded values with the standard media tag carrying `source=field_size_limit`, so media-aware trace views show the specialized warning and original-file link
- leave legacy `observations` writes and staging-based dual writes through `observations_batch_staging` unchanged for this minimal V1
- do not add an aggregate metadata or endpoint-size cap
- forward the opt-in flag and size limit to the worker in Docker Compose, while using the container-internal MinIO endpoint for worker uploads

## Scope

Field overflow is disabled by default. Set `LANGFUSE_OBSERVATION_FIELD_OVERFLOW_ENABLED=true` on the worker to enable it.

When enabled, field overflow is intentionally limited to records written through `IngestionService.writeEventRecord()`, including direct event-table ingestion. The legacy observation merge and its staging-based dual-write propagation do not invoke the overflow processor. Consequently, deployments running the default legacy write mode are not protected by this V1.

Usage, cost, tool normalization, and evaluation enrichment run on the original values before the direct event record is transformed. Since legacy observation rows retain their original values, no sentinel-aware usage calculation is needed.

The worker overflow feature is a stateless event-record transformation called directly at the write boundary. It owns field extraction, media processing, logging, metrics, and reconstruction of the immutable persisted record; `IngestionService` does not carry a processor as instance state.

Sub-limit fields are measured with `Buffer.byteLength()` without allocating content buffers. A `Buffer` is created only after a field is known to require upload.

The overflow path uses the existing media upload contract and preserves its current content-type and reupload semantics. It does not add a separate collision-protection option.

## Failure behavior

When enabled, overflow upload or media-link failures are fail-open for V1: emit a structured warning and failed outcome metrics, persist the original oversized value, and continue the ingestion job. This intentionally means an oversized `events_full` value can still be written while media storage is unavailable.

## Validation

- shared overflow processor: `Tests  7 passed (7)`
- shared media service: `Tests  4 passed (4)`
- worker event-record overflow feature, including disabled behavior: `Tests  4 passed (4)`
- worker full-module mock regression: `Tests  18 passed (18)`
- ingestion direct-versus-legacy scope regression: `Tests  9 passed (9)`
- token/cost regression suite: `Tests  28 passed (28)`
- web media utility: `Tests  14 passed (14)`
- web media view: `Tests  1 passed (1)`
- web media layout regression: `Tests  1 passed (1)`
- Docker Compose defaults: flag `false`, limit `2097152`, worker media endpoint `http://minio:9000`
- Docker Compose overrides: flag `true`, limit `123456`
- `pnpm run lint`: `Tasks: 7 successful, 7 total`
- `pnpm run typecheck`: `Tasks: 7 successful, 7 total`
- pre-commit formatting: `All matched files use Prettier code style!`

Linear: https://linear.app/clickhouse/issue/LFE-14407/set-a-hard-cap-on-large-observation-field-payloads


packages/shared/src/domain/observation-field-spill.ts
```diff
@@ -0,0 +1,14 @@
...
+export const OBSERVATION_FIELD_SIZE_LIMIT_EXCEEDED_KEY =
+  "_langfuse_field_size_limit_exceeded_file";
+
+export const OBSERVATION_FIELD_SIZE_LIMIT_MEDIA_SOURCE = "field_size_limit";
+
+export function createObservationFieldSizeLimitSentinel(
+  mediaReference: string,
+): string {
+  return JSON.stringify({
+    [OBSERVATION_FIELD_SIZE_LIMIT_EXCEEDED_KEY]: mediaReference,
```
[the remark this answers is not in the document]
hassiebp: Fixed in 521a73a25: successful spills now persist the bare `@@@langfuseMedia:...|source=field_size_limit@@@` reference directly. Added a client integration test that passes that bare reference through `classifyMediaValue()` into `JsonMediaTag` and asserts the specialized warning.

worker/src/services/IngestionService/index.ts
```diff
@@ -438,10 +443,36 @@ export class IngestionService {
...
+      traceId: eventRecord.trace_id,
+      observationId: eventRecord.span_id,
+      fields: {
+        input: eventRecord.input,
+        output: eventRecord.output,
+        metadata: eventRecord.metadata_values,
+      },
+    });
+    const persistedMetadataValues = Array.isArray(spillResult.fields.metadata)
+      ? spillResult.fields.metadata
```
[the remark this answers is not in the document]
hassiebp: Acknowledged as an intentional scope boundary for this V1. The cap applies only to observation `input`, `output`, and individual observation metadata values at ingestion. `experiment_item_expected_output`, trace fields, historical rows, and accepted fail-open values are outside that contract; expanding the event-propagation/backfill path would broaden the feature beyond the ticket. The PR description and Linear update now state the scoped fields and fail-open tradeoff explicitly.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const result = await spillOversizedObservationFields({
+    fields,
+    maxFieldBytes: env.LANGFUSE_OBSERVATION_FIELD_SIZE_LIMIT_BYTES,
+    upload: async ({ field, contentBytes }) => {
+      if (!env.LANGFUSE_S3_MEDIA_UPLOAD_BUCKET) {
+        throw new Error("Media upload bucket is not configured");
+      }
+
+      return uploadMediaForTrace({
```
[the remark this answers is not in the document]
hassiebp: Fixed in f833b458c. The Compose worker now defaults LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT to http://minio:9000, while the web service explicitly defaults it to the browser-reachable http://localhost:9090. docker compose config resolves those exact per-service endpoints, and an explicitly supplied env value still overrides both.

worker/src/services/IngestionService/index.ts
```diff
@@ -963,6 +994,23 @@ export class IngestionService {
...
+      traceId: finalObservationRecord.trace_id || finalObservationRecord.id,
+      observationId: finalObservationRecord.id,
+      fields: {
+        input: finalObservationRecord.input,
+        output: finalObservationRecord.output,
+        metadata: finalObservationRecord.metadata,
+      },
+    });
+    finalObservationRecord.input = spillResult.fields.input ?? undefined;
+    finalObservationRecord.output = spillResult.fields.output ?? undefined;
```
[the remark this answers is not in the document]
hassiebp: Fixed in 3e8145dfe. Field-size-limit media references are now recognized during manual tokenization, the previously stored usage bucket is preserved when only the opposite side is updated, and total usage/cost are recomputed. The regression test covers a stored spilled input followed by a new output update.

docker-compose.yml
```diff
@@ -44,9 +44,10 @@ services:
       LANGFUSE_S3_MEDIA_UPLOAD_REGION: ${LANGFUSE_S3_MEDIA_UPLOAD_REGION:-auto}
       LANGFUSE_S3_MEDIA_UPLOAD_ACCESS_KEY_ID: ${LANGFUSE_S3_MEDIA_UPLOAD_ACCESS_KEY_ID:-minio}
       LANGFUSE_S3_MEDIA_UPLOAD_SECRET_ACCESS_KEY: ${LANGFUSE_S3_MEDIA_UPLOAD_SECRET_ACCESS_KEY:-miniosecret} # CHANGEME
-      LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT: ${LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT:-http://localhost:9090}
+      LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT: ${LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT:-http://minio:9000}
```
hassiebp, line 47: why do we edit this file at all?

hassiebp: This feature makes the worker perform media uploads for the first time. Inside the worker container, `http://localhost:9090` points to the worker itself rather than MinIO, so every oversized-field spill in the shipped Compose setup would fail open and keep the large ClickHouse value. The worker therefore needs the internal `http://minio:9000` default. The web service explicitly overrides it back to `http://localhost:9090`, because its presigned media URLs must remain reachable from the host browser. A user-supplied `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT` still overrides both defaults.

worker/src/services/IngestionService/index.ts
```diff
@@ -430,6 +435,8 @@ export class IngestionService {
   /**
    * Writes an event record directly to the events_full table.
    * A materialized view auto-populates events_core from events_full.
+   * Legacy observation writes and staging-based dual writes intentionally do
+   * not use field spilling.
```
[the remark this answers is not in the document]
hassiebp: This is intentional for the minimal V1 agreed in LFE-14407. The change is deliberately scoped to direct `events_full` writes through `writeEventRecord()`; legacy `observations` writes and staging-based dual writes remain unspilled. The PR description now calls out that default legacy-mode deployments are not protected by this V1, and the regression test locks in that boundary. No code change needed for this thread.

packages/shared/src/server/media/ObservationFieldSpillProcessor.ts
```diff
@@ -0,0 +1,175 @@
...
+  const {
+    field,
+    serializedValue,
+    fallbackValue,
+    maxFieldBytes,
+    upload,
+    onUploadError,
+  } = params;
+  const contentBytes = Buffer.from(serializedValue, "utf8");
+  const originalBytes = contentBytes.length;
```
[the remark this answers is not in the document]
hassiebp: Accepted and fixed in `36b929098`. `spillValue` now uses `Buffer.byteLength()` for the threshold check and only allocates `contentBytes` after the value is known to be oversized. Added a regression test covering input, output, and metadata: it failed before the change with three `Buffer.from` calls and now passes. Validation: shared spill processor `Tests 7 passed (7)`, worker spill feature `Tests 3 passed (3)`, targeted web media consumers `Tests 14 passed (14)` and `Tests 1 passed (1)`, lint/typecheck `7/7` each.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,107 @@
...
+        projectId,
+        traceId,
+        observationId,
+        field,
+        contentType: MediaContentType.TXT,
```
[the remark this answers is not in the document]
hassiebp: Confirmed and fixed in `67bcc60e4`. `uploadMediaForTrace` now rejects an existing same-hash media row with a different content type before any row/object mutation; the field-spill caller catches that error and follows the documented fail-open path. `upsertMediaRecord` also includes `contentType` in its guarded update to cover concurrent inserts. The regression test resolved before the fix (demonstrating the overwrite path) and now rejects without calling insert, update, upload, or link. Validation: combined shared media suites `Tests 12 passed (12)`, worker spill feature `Tests 3 passed (3)`, web media utility `Tests 14 passed (14)`, lint/typecheck `7/7` each.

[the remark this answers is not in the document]
hassiebp: Follow-up after full CI: the first fix made content-type immutability global and correctly failed the existing public media retry/reupload contract (`media.servertest.ts`). `baa15b535` narrows the protection to field-spill uploads via `protectExistingContentType: true`; existing public upload callers retain their established overwrite/retry behavior. The guarded upsert still covers concurrent same-hash inserts for spills. Added unit coverage for both protected rejection and default cross-type reupload. Local validation: shared media suites `Tests 13 passed (13)`, worker spill feature `Tests 3 passed (3)`, lint/typecheck `7/7` each. The exact server integration test could not run locally because Postgres is not available; replacement CI is running it.

packages/shared/src/server/media/mediaService.ts
```diff
@@ -44,6 +44,8 @@ export async function upsertMediaRecord(params: {
   uploadBucket: string;
   contentType: MediaContentType;
   contentLength: number;
+  /** Prevents an existing same-hash media row from being retyped. */
+  protectExistingContentType?: boolean;
```
hassiebp: I think we can drop this - overly defensive.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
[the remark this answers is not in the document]
hassiebp: Follow-up per product decision: removed `protectExistingContentType` and the related media-service changes to keep V1 narrow. Existing media collision and reupload semantics remain unchanged; the overflow path now relies on the current fail-open upload behavior. This intentionally accepts the rare cross-type, same-byte edge case for V1.

docker-compose.yml
hassiebp, line 1: why are we editing this file? remove

.env.dev.example
hassiebp, line 1: remove changes

worker/src/env.ts
```diff
@@ -365,6 +368,8 @@ const EnvSchema = z.object({
     .default("false"),
   LANGFUSE_S3_MEDIA_UPLOAD_SSE: z.enum(["AES256", "aws:kms"]).optional(),
   LANGFUSE_S3_MEDIA_UPLOAD_SSE_KMS_KEY_ID: z.string().optional(),
+  LANGFUSE_OBSERVATION_FIELD_SIZE_LIMIT_BYTES:
+    langfuseObservationFieldSizeLimitBytesSchema,
```
hassiebp: why is this not simply defined inline? simplify

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,107 @@
...
+      field: outcome.field,
+      outcome: outcome.outcome,
+    });
+    recordDistribution(
+      "langfuse.ingestion.observation_field_overflow.original_bytes",
+      outcome.originalBytes,
+      { field: outcome.field, outcome: outcome.outcome },
+    );
+    recordDistribution(
+      "langfuse.ingestion.observation_field_overflow.persisted_bytes",
```
hassiebp: return and report the removed bytes instead to be analogous to media processing in ingestion pipeline

docker-compose.yml
hassiebp: Addressed in 73c785f2e. Removed the redundant size-limit declaration and kept only the required endpoint routing: the worker resolves MinIO internally at http://minio:9000 while web keeps browser-reachable signed URLs at http://localhost:9090. Verified with docker compose config.

.env.dev.example
hassiebp: Addressed in 73c785f2e. Removed the redundant .env.dev.example declaration; the worker runtime schema remains the source of the configurable 2 MiB default.

worker/src/env.ts
hassiebp: Addressed in 73c785f2e. The Zod validation is now defined inline in worker/src/env.ts and the one-use shared schema was removed.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
hassiebp: Addressed in 73c785f2e. Overflow outcomes now return bytesRemoved, failed uploads report zero, and metrics emit a single observation_field_overflow.bytes_removed distribution, matching the existing media-processing convention. Covered by shared and worker unit tests.

worker/src/env.ts
```diff
@@ -365,6 +368,11 @@ const EnvSchema = z.object({
     .default("false"),
   LANGFUSE_S3_MEDIA_UPLOAD_SSE: z.enum(["AES256", "aws:kms"]).optional(),
   LANGFUSE_S3_MEDIA_UPLOAD_SSE_KMS_KEY_ID: z.string().optional(),
+  LANGFUSE_OBSERVATION_FIELD_SIZE_LIMIT_BYTES: z.coerce
+    .number()
+    .int()
+    .positive()
+    .default(OBSERVATION_FIELD_SIZE_LIMIT_BYTES),
```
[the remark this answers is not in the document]
hassiebp: Addressed in 6c7854df7. Compose now forwards both the configurable size limit and the opt-in feature flag to the worker. Verified defaults are false / 2097152 and explicit overrides resolve to true / 123456.

hassiebp:
@claude review

packages/shared/src/server/media/ObservationFieldOverflowProcessor.ts
```diff
@@ -0,0 +1,176 @@
+import {
```
hassiebp: why does replaceOverSizedObservationFieldsWithMedia live in its own file in shared? why is that file named differently? Why is this not sumply in processObservationFieldOverflow.ts or as a separate file in the observation-field-overflow directory?

packages/shared/src/domain/observation-field-overflow.ts
```diff
@@ -0,0 +1,3 @@
+export const OBSERVATION_FIELD_SIZE_LIMIT_BYTES = 2 * 1024 * 1024;
```
hassiebp: same here, why does this not simply live in worker observation-field-overflow dir, why so complicated?

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  const overflowResult = await processObservationFieldOverflow({
+    projectId: eventRecord.project_id,
+    traceId: eventRecord.trace_id,
+    observationId: eventRecord.span_id,
+    fields: {
+      input: eventRecord.input,
+      output: eventRecord.output,
+      metadata: eventRecord.metadata_values,
```
hassiebp: i think we can simplify a lot by typing metadata here as metadata_values directly as we only pass in eventRecord.metadata_values anyway

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,106 @@
...
+  recordDistribution,
+  recordIncrement,
+  replaceOversizedObservationFieldsWithMedia,
+  uploadMediaForTrace,
+} from "@langfuse/shared/src/server";
+import { MediaContentType } from "@langfuse/shared";
+
+import { env } from "../../env";
+
+export async function applyObservationFieldOverflow(
```
hassiebp, line 31: do we have sufficient error handling? we should never fail ingestion due to this, and I can currently see only upload error handling but no processor error handling

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,106 @@
...
+      metadata: eventRecord.metadata_values,
+    },
+  });
+  const persistedMetadataValues = Array.isArray(overflowResult.fields.metadata)
+    ? overflowResult.fields.metadata
+    : [];
+
+  return {
+    ...eventRecord,
+    input: overflowResult.fields.input ?? undefined,
```
hassiebp: is there a way to type this more elegantly?

worker/src/features/observation-field-overflow/processObservationFieldOverflow.test.ts
```diff
@@ -0,0 +1,199 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { EventRecordInsertType } from "@langfuse/shared/src/server";
+
+const mocks = vi.hoisted(() => ({
```
sumerman, line 4: I am usually somewhat skeptical of standalone heavily mocked tests and ask my agent to review them to see if it makes sense to incorporate them into an existing integration suite.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -615,7 +615,7 @@ export const otelIngestionQueueProcessorBuilder = (
           // Step 3: Write to events table (independent of eval scheduling)
           if (shouldWriteToEventsTable) {
             try {
-              ingestionService.writeEventRecord(eventRecord);
+              await ingestionService.writeEventRecord(eventRecord);
```
sumerman, line 618: why add these awaits all over?

worker/src/features/internal-tracing/createInternalEventsWriter.ts
```diff
@@ -39,7 +39,7 @@ async function writeInternalEventInputs(params: {
 
   if (v4WritesToEventsTable(env)) {
     for (const eventRecord of eventRecords) {
-      service.writeEventRecord(eventRecord);
+      await service.writeEventRecord(eventRecord);
```
sumerman, line 42: same here. I would assume @Steffen911 left these without `await` for a reason

worker/src/services/IngestionService/tests/IngestionService.integration.test.ts
```diff
@@ -173,7 +173,9 @@ describe("Ingestion end-to-end tests", () => {
       true,
     );
 
-    eventRecords.forEach((record) => ingestionService.writeEventRecord(record));
+    await Promise.all(
```
sumerman, line 176: same here, I would also assume the line below performs actual waiting.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,138 @@
...
+  if (env.LANGFUSE_OBSERVATION_FIELD_OVERFLOW_ENABLED !== "true") {
+    return eventRecord;
+  }
+
+  try {
+    const mediaBucket = env.LANGFUSE_S3_MEDIA_UPLOAD_BUCKET;
+    const input =
+      eventRecord.input == null
+        ? eventRecord.input
+        : await overflowValue(
```
sumerman: why do this sequentially?

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,138 @@
...
+            eventRecord,
+            "output",
+            eventRecord.output,
+            mediaBucket,
+          );
+    const metadataValues: string[] = [];
+    for (const value of eventRecord.metadata_values) {
+      metadataValues.push(
+        await overflowValue(eventRecord, "metadata", value, mediaBucket),
```
sumerman: wow, if I am reading this right, this is potentially `length(keys(metadata))*latency(s3Put)` easily waiting anywhere from a couple of seconds to minutes.

sumerman (commented):
I don't think this should be merged. The way I am reading it, this code waits until each individual field is uploaded as a media before proceeding. This could easily balloon latencies on heavy payloads. 

I would run oversized entries detection as a first pass, followed by S3 upload in batches of N (configurable).

worker/src/features/observation-field-overflow/processObservationFieldOverflow.test.ts
hassiebp: fair point in general, but in this case I think having this boundary is useful as we can focus behavior tests to cover the byte boundaries, upload failures etc

The `IngestionService` tests separately cover the wiring and direct-events scope

worker/src/queues/otelIngestionQueue.ts
hassiebp: writeEventRecord was previously sync, with the overflow processing we have to await the potential S3 upload before enqueueing to CH writer

Should not be a problem from what I see

worker/src/features/internal-tracing/createInternalEventsWriter.ts
hassiebp: please see previous comment

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
hassiebp: - I think in general we have a tradeoff here between memory overhead and wallclock time of the job
- I intentionally decided to go the sequential processing path (exactly as done for media processing as well in ingestion pipeline) as creating the upload payload with Buffer.from(value) allocates another full copy of each oversized field. Processing the fields sequentially limits an observation to one active overflow buffer/upload at a time. 
- Downside is that we are accepting additional processing latency for the unusual case where one observation contains several oversized fields
- Also code is a bit easier to digest without the additional optimization of concurrent processing / upload

In general I think it's fine to go for simpler approach for this PR

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
hassiebp: please see previous comment

worker/src/queues/otelIngestionQueue.ts
sumerman: Apologies, I missed the signature change.

worker/src/services/IngestionService/index.ts
```diff
@@ -438,10 +441,14 @@ export class IngestionService {
...
-  public writeEventRecord(eventRecord: EventRecordInsertType): void {
+  public async writeEventRecord(
+    eventRecord: EventRecordInsertType,
+  ): Promise<void> {
+    const persistedRecord = await applyObservationFieldOverflow(eventRecord);
+
     this.clickHouseWriter.addToQueue(
       TableName.EventsFull,
-      withSerializedEventByteLength(eventRecord),
+      withSerializedEventByteLength(persistedRecord),
```
Steffen911, line 451: Does this mean that we also exclude the extracted IO from the event size? I feel that this makes sense for actual media processing, but not necessarily for big input/output/metadata extractions. 🤔 Unsure about that one, but somehow it feels a bit odd.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
sumerman: - In my limited experience I saw them come in 3. Large input, large output and one large metadata field. 
- As you may remember I was advocating for batching in media extraction as well. We already have very problematic spikes.

worker/src/queues/otelIngestionQueue.ts
Steffen911: I somewhat liked the semantic of the write being a sync action that just adds something to the CHWriter Queue. It made it pretty obvious that no writing happened at that point and there aren't many transformations happening anymore. 

Why doesn't this fit into createEventRecord ? For the evaluation parts?

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,138 @@
...
+  type MediaField,
+  OBSERVATION_FIELD_SIZE_LIMIT_MEDIA_SOURCE,
+} from "@langfuse/shared";
+
+import { env } from "../../env";
+
+export async function applyObservationFieldOverflow(
+  eventRecord: EventRecordInsertType,
+): Promise<EventRecordInsertType> {
+  if (env.LANGFUSE_OBSERVATION_FIELD_OVERFLOW_ENABLED !== "true") {
```
Steffen911, line 34: You could/should check whether a media bucket is configured here.

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
```diff
@@ -0,0 +1,138 @@
...
+    return eventRecord;
+  }
+
+  try {
+    const mediaBucket = env.LANGFUSE_S3_MEDIA_UPLOAD_BUCKET;
+    const input =
+      eventRecord.input == null
```
Steffen911: Deliberate `==`? I'd rather have full explicitness with `===`.

hassiebp: Yeah as this will capture both null and undefined

worker/src/features/observation-field-overflow/processObservationFieldOverflow.ts
hassiebp: We do in line 70 👍🏽

worker/src/services/IngestionService/index.ts
hassiebp: Yes, I discussed with Max and we would like to have the `sum(media['byte_length']) + sum (events_core['event_bytes'])` as the relevant total bytes ingested value, thus we need to exclude here

https://clickhouse-inc.slack.com/archives/C0B6DBVRPKJ/p1784704658640459?thread_ts=1784636714.956729&cid=C0B6DBVRPKJ

PR #15387, opened 2026-07-23
fix: keep bare string dataset item inputs as strings

### LINKED ISSUE #15342: bug: Langfuse.get_dataset() causes digital strings beign casted to int
### Describe the bug

I created a dataset, in which some items' inputs were like `"%d+"` (for example, `"123456"`).
The items were displayed as strings in the UI.
But when I accessed the dataset in my code with `Langfuse.get_dataset()`, I discovered that digital strings became integers.

### Steps to reproduce

To reproduce the bug one may use this simple code:
```python
from dotenv import load_dotenv
from langfuse import get_client

load_dotenv(".env")
langfuse = get_client()

ds_name = "TestingTypesDataset" 
langfuse.create_dataset(
    name=ds_name
)

str_item = "123456"
print(f'Item type: {type(str_item)}')
langfuse.create_dataset_item(dataset_name=ds_name, input=str_item)

ds = langfuse.get_dataset(ds_name)
print(f'Item type after retrieving the dataset: {type(ds.items[0].input)}')
```
The output is as follows:

> Item type: <class 'str'>
> Item type after retrieving the dataset: <class 'int'>

Self-hosted

3.223.0

langfuse==4.14.1

### Additional information

In the web UI the item is displayed as string.
Image

No

prithidevghosh:
## What does this PR do?

Creating a dataset item through the public API with a bare string input like `"123456"` stored it as the number `123456`, so reading the dataset back through the SDK returned an `int` instead of the `str` that went in. The UI still rendered it as a string, which made it look like a retrieval bug, but the value was already a number in Postgres.

Fixes langfuse/langfuse#15342

### Cause

`DatasetItemValidator.normalize()` ran every string through `parseJsonPrioritised()`:

```ts
if (typeof data === "string") {
  const parsedJson = parseJsonPrioritised(data);
  parsed = (parsedJson === undefined ? data : parsedJson) as Prisma.InputJsonValue;
}
```

That is correct for tRPC, where the dataset item form submits JSON-encoded strings like `'{"key":"value"}'`. It is wrong for the public API, where the HTTP body parser has already decoded the payload, so a string is a real string. Any value that happens to look like a JSON literal got parsed a second time, turning `"123456"`, `"1.5"`, `"true"` and `"null"` into a number, a boolean and null.

One function was serving two callers with opposite needs and had no way to tell them apart.

### Fix

Parsing is now opt-in through `normalizeOpts.parseJsonStrings`, so the decision is made by the caller that knows whether its values are still encoded:

| Caller | Parses | Why |
| -- | -- | -- |
| tRPC `createDatasetItem`, `createManyDatasetItems`, `updateDatasetItem` | yes | the item form submits JSON-encoded strings |
| worker `processAddObservationsToDataset` | yes | reads `input`/`output` from ClickHouse `Nullable(String)` columns |
| Public API (`createDatasetItemForApi`) and MCP | no | the HTTP body parser has already decoded the payload |

Two call sites are worth flagging for review. `updateDatasetItem` previously passed no `normalizeOpts` at all and was relying on the old implicit default, so it needed the opt-in to keep working. The worker batch action genuinely depends on the parse, since ClickHouse returns those columns as strings — I confirmed the column types rather than assuming.

Behaviour is unchanged everywhere except the public API and MCP, which is the path in the issue.

## Type of change

## Mandatory Tasks

## Tests

Failing tests were written first, and confirmed to fail before the fix:

* `packages/shared/src/server/services/DatasetService/DatasetItemValidator.test.ts` (new) covers both paths: already-parsed values are stored verbatim, opted-in JSON strings are still decoded. 8 of the 17 assertions failed before the fix.
* `web/src/__tests__/server/datasets-api.servertest.ts` adds a public API round-trip for `"123456"`, `"1.5"`, `"true"` and `"null"`. To check these actually catch the regression rather than passing vacuously, I reverted the fix and confirmed all four flip to failing, then restored it.
* The two existing `datasets-schema-validation` unit tests that feed JSON strings directly to the validator now pass `parseJsonStrings: true`, since they model the tRPC path.

Local runs: `typecheck` and `lint` both `7 successful, 7 total`; `DatasetItemValidator` 17/17; `datasets-api` 27/27; `datasets-schema-validation` 48/48; `dataset-items-repository` 46/46; `dataset-service` 35/35; worker `batchAction.test.ts` 15/15.

Also verified end to end against a local instance: an item posted via the public API with input `"123456"` is stored as a JSON `string` and reads back as `str`, while items created and edited through the UI form are still stored as parsed objects.

This PR makes dataset-item JSON-string parsing caller-controlled.

* Adds `parseJsonStrings` normalization configuration and preserves literal strings by default.
* Opts the tRPC and observation batch-action paths into parsing while leaving public API and MCP values decoded.
* Adds validator and public API regression coverage for JSON-literal-looking strings.

</details>

<details><summary>Confidence Score: 4/5</summary>

The PR appears safe to merge, with the non-blocking source-file encoding issue corrected to keep the new test reviewable as text.

Caller-specific parsing preserves the existing tRPC and worker behavior while fixing public API string coercion; the only accepted concern is that an actual NUL byte makes the new test binary to Git.

packages/shared/src/server/services/DatasetService/DatasetItemValidator.test.ts

</details>

```markdown

---

### Issue 1 of 1
packages/shared/src/server/services/DatasetService/DatasetItemValidator.test.ts:68
**Keep the test source text-diffable**

This string literal contains an actual NUL byte, so Git classifies the TypeScript file as binary and hides the entire test from normal diffs. Using the `\u0000` source escape exercises the same runtime character while keeping reviews and future merges text-based.
```

</details>

Reviews (1): Last reviewed commit: ["fix: keep bare string dataset item input..."](<https://github.com/langfuse/langfuse/commit/f8cb0577578276782d5207129d8b979150b8e3ea>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=46834796>)

**Context used:**

* Knowledge Base — [Datasets, Dataset Runs, and Experiments](<https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/datasets-experiments.md>)
* Knowledge Base — [Batch Actions and Data Exports](<https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/batch-actions-exports.md>)

### DIFF
--- a/packages/shared/src/server/repositories/dataset-items.ts
+++ b/packages/shared/src/server/repositories/dataset-items.ts
@@ -35,7 +35,23 @@ import {
 import { postgresSearchCondition } from "../queries";
 import { TracingSearchType } from "../../interfaces/search";
 
-const emptyNormalizeOpts: { sanitizeControlChars?: boolean } = {};
+/**
+ * Normalization options for dataset item payloads.
+ *
+ * `parseJsonStrings` must only be set by callers whose values are still
+ * JSON-encoded strings: tRPC (the web form submits strings) and the worker
+ * batch action (ClickHouse returns input/output as String columns). Callers
+ * holding already-parsed values — the Public API and MCP, where the HTTP body
+ * parser has already decoded the payload — must leave it unset, so a bare
+ * string such as "123456" is stored verbatim instead of being decoded a second
+ * time into the number 123456 (#15342).
+ */
+export type DatasetItemNormalizeOpts = {
+  sanitizeControlChars?: boolean;
+  parseJsonStrings?: boolean;
+};
+
+const emptyNormalizeOpts: DatasetItemNormalizeOpts = {};
 const emptyValidateOpts: { normalizeUndefinedToNull?: boolean } = {};
 
 /**
@@ -231,9 +247,7 @@ export async function createDatasetItem(props: {
   metadata?: string | unknown | null;
   sourceTraceId?: string;
   sourceObservationId?: string;
-  normalizeOpts?: {
-    sanitizeControlChars?: boolean;
-  };
+  normalizeOpts?: DatasetItemNormalizeOpts;
   validateOpts?: {
     normalizeUndefinedToNull?: boolean;
   };
@@ -292,7 +306,7 @@ export async function upsertDatasetItem(
     sourceTraceId?: string;
     sourceObservationId?: string;
     status?: DatasetStatus;
-    normalizeOpts?: { sanitizeControlChars?: boolean };
+    normalizeOpts?: DatasetItemNormalizeOpts;
     validateOpts: { normalizeUndefinedToNull?: boolean };
   } & IdOrName,
 ): Promise<DatasetItemDomain & { datasetName: string }> {
@@ -604,7 +618,7 @@ export async function deleteDatasetItem(props: {
 export async function createManyDatasetItems(props: {
   projectId: string;
   items: CreateManyItemsPayload;
-  normalizeOpts?: { sanitizeControlChars?: boolean };
+  normalizeOpts?: DatasetItemNormalizeOpts;
   validateOpts?: { normalizeUndefinedToNull?: boolean };
   allowPartialSuccess?: boolean;
 }): Promise<
--- a/packages/shared/src/server/repositories/dataset-items.ts
+++ b/packages/shared/src/server/repositories/dataset-items.ts
@@ -197,27 +197,22 @@ function toDomainType<
   return withDatasetName as any;
 }
 
+/**
+ * Merge the non-IO fields (source ids, status) of an update with the existing
+ * item, keeping the existing value for anything the caller omitted. IO fields
+ * (input/expectedOutput/metadata) are merged inside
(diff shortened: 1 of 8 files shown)


prithidevghosh:
@nkabardin requesting your review on this pr.
this change stops the API turning a dataset item input of type "string" to type "int". 
as mentioned in the issue #15342

prithidevghosh:
@wochinge @nkabardin could you please check this pr?

nkabardin:
@claude review

prithidevghosh:
@nkabardin i have added a fix pr, can you please rerun the review.
i tried it myself by mentioning claude but it seems to be not working.

nkabardin:
@claude review

prithidevghosh:
@nkabardin please help running the review again, did just a small comment change. 
I hope, we would be good to merge after this :)

nkabardin:
@marliessophie would you take a look?

prithidevghosh:
@marliessophie can you review this pr please?

PR #15395, opened 2026-07-24
feat(dashboards): copyable drill-in breakdown labels and clearer widget copy action (LFE-10962)

nkabardin:
## TL;DR
Breakdown-chart category labels (truncated user IDs, model names, etc.) are now interactive: hover a label and a card shows the **full value** with a **copy** button and a **"View filtered table"** link that jumps straight to the underlying rows. Also disambiguated the widget kebab **"Copy to clipboard" → "Copy widget as JSON"** (and "Copy card as JSON" on preset cards).

## Why
A user reported two paper-cuts on dashboard breakdown widgets: long category values get truncated to `cmmtj4wmo00nt…` with no way to read the full value or find which rows it represents, and the "Copy to clipboard" menu item was ambiguous ("what does this copy — the widget? the data? in what format?").

## What
- The horizontal-bar Y-axis label is now a real interactive element (an SVG `foreignObject` hosting an HTML button) instead of plain SVG text, so it can open a **hover card** (opens on hover *and* keyboard focus; stays open when you move into it so the copy button + link are usable). The card's link reuses the widget's existing "View as table" deep-link machinery, extended to pin the hovered category value on top of the widget's own filters.
- The widget-definition copy action is now self-describing: "Copy widget as JSON" where it copies the full widget definition, "Copy card as JSON" on Langfuse-maintained preset cards (which copy a preset reference).

## Result
Breakdown labels are readable and actionable; the copy action reads correctly at a glance. Verified live end-to-end: opening the card on a userId breakdown shows the full value + copy, and "View filtered table" lands on the observations table filtered to exactly that user (`userId = …` + the dashboard's env + date range).

<details>
<summary>Mechanism, safety &amp; verification</summary>

- `buildCategoryTableHrefs` (extracted + unit-tested) types each pin against the **destination** table column (string → `=`, stringOptions/arrayOptions → `any of`); dimensions that can't be expressed (metadata, numeric version) get no link rather than a wrong one.
- The null/empty bucket is rendered as a sentinel `"n/a"` (now a shared `MISSING_DIMENSION_LABEL` constant); it is explicitly **excluded** from drill-in links — otherwise the (often largest) no-value bar would pin `column = "n/a"` and land on zero rows.
- Label card = Radix `HoverCard` portaled to the `popover` overlay layer (never clipped by the chart SVG); the native `<title>` fallback is retained for a bare mouseover. The link is a `next/link` `<Link>` (SPA nav, cmd/ctrl-click opens a new tab) with the analytics `capture()` firing before navigation.
- Only wired for `HORIZONTAL_BAR` widgets; analytics payloads carry ids only (no raw dimension value / PII).
- `web` lint + typecheck green; `buildTableFilterHref` suite 23 passed; full client sweep 2532 passed.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

nmtrang29:
*(Comment posted by Claude on Trang's behalf.)*

Heads-up: the widget copy-action rename now lives in #16296 as part of a broader widget-menu consistency pass — the label there is plain **"Copy widget" / "Copy card"** (plus a copied-toast with a Cmd/Ctrl+V hint), per Trang's call. To avoid a conflicting rename of the same strings, feel free to drop the "Copy widget as JSON" bit from this PR — the drill-in breakdown-label work here is unaffected and still stands on its own.

PR #15403, opened 2026-07-24
refactor(widgets): move WidgetForm to react-hook-form and fix histogram silent-revert (LFE-10699)

nkabardin:
## TL;DR
Rewrote the 2,380-line `WidgetForm` (17 `useState`, 8 sync `useEffect`, no form library) onto **react-hook-form + zod** — **zero `useEffect`, zero `useState`**, one form field per concept — and fixed the histogram "silent revert" bug in the process. Same UI, byte-identical save payload (proven by round-trip parity tests), live-verified across create / edit / pivot / histogram / view-switch / import.

## Why
The widget builder had grown a web of 8 effects that reactively rewrote each other's state: switching chart type could snap your aggregation, changing a measure could silently flip your chart to a Big Number, and a stale pivot sort was cleared behind your back. The worst offender (the reported bug): picking a histogram-incompatible measure/chart combo would **silently revert** to Sum with no explanation — it read as broken. This was fragile, hard to reason about, and it's the foundation the next two dashboard levers (a create/edit **modal**, and **per-series colors**) build on, so it was worth doing right first.

## What
One composed zod schema is the source of truth. The dual single-metric/pivot representation collapses into `metrics[]` / `dimensions[]` arrays that already match the save payload. One `useForm` + one `useWatch` in the controller derive everything once; each control is a small subcomponent bound to exactly one nested field. Cross-field rules (histogram compatibility, pivot cardinality, breakdown support) live in a zod `superRefine`; cross-slice cascades live in three parent handlers fired from the initiating event — never in an effect. Invalid states can no longer silently disable Save: they either **heal** to a valid config or surface an inline message.

The histogram bug is fixed three ways: the histogram aggregation is only offered on the histogram chart, chart/measure changes resolve compatibility synchronously with your click, and any residual incompatible state is a visible validation message — never a silent snap.

## Result
Same builder, same saved output, but zero effects, one-concept-one-field state, and the histogram footgun gone. 258 widget tests green (incl. round-trip parity + heal cases); lint, typecheck, and a full production build pass. Live-verified: create→save persists; auto-name/description pre-fill and live-update; building a Histogram then switching View **heals to Big Number with Save enabled** (the previously-reported dead-end).

<details>
<summary>Mechanism &amp; verification</summary>

- **State**: `makeWidgetFormSchema(viewVersion)` with a flat `chart` record (not a discriminated union, to avoid churning RHF field arrays on type change). `toDefaultValues`/`toSavePayload` are pure adapters; the save object is copied verbatim from the old `handleSaveWidget` — round-trip parity tests (including a non-empty-filter fixture) prove it's byte-identical.
- **Effects → 0**: one pure `normalizeWidgetFormValues` (reusing the existing, unit-tested `resolveAggregationAndChartType` fixed point) heals the (chart, aggregation) pair and drops incompatible dimensions. It runs at three seams — `toDefaultValues` (so edit/import never *mount* invalid), the import `form.reset`, and the three cascade handlers (`onViewChange`/`onChartTypeChange`/`onMeasureChange`). A valid widget is a fixed point (parity preserved); an invalid stored/imported one heals to NUMBER exactly as the old mount effects did. Beta/edit resets are keyed remounts; pivot-sort sanitize, auto-name/description, and view-version are derived in render.
- **Auto-name**: value = `override || suggestion`, isolated in `NameField`/`DescriptionField` so the presentation is swappable in one place (a follow-up decision).
- **Validation**: superRefine messages render inline under the relevant control and next to a disabled Save, restoring the intent of the old toasts.
- **Verified**: 258 tests (parity, heal-on-seed, superRefine invariants, empty-slot) + the existing render and `resolveAggregationAndChartType` suites kept green; lint / typecheck / `next build` pass; and a live browser pass over create, save, histogram-heal-on-view-change, and the post-save dashboard flow.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15404, opened 2026-07-24
feat(session): mobile layout — reclaim viewport for session content (LFE-11067)

nkabardin:
## TL;DR
On mobile the session detail page gave the actual session content only ~40% of the screen. This adds the missing mobile branch: page actions collapse to one scrollable row, and the session's metadata/controls bar becomes a default-collapsed accordion — so the virtualized trace feed reclaims the height. **Desktop is byte-identical.**

## Why
There was **no mobile layout** for the session view — the desktop layout was just shrunk to phone width. At ~390px, two non-collapsible headers stacked above the content:
1. `MobilePageTitle` rendered the session's 6–9 action buttons as **two** `flex-wrap` rows, wrapping into ~200–300px of chrome.
2. The session's own sticky metadata/controls bar (LLM-calls preset, Saved Views, "Filter observations", trace/cost/user/score stats) wrapped into another ~100–140px.

The `flex-1` virtualized trace feed got only the small remainder. The trace-detail view already solved this class of problem with a `useIsMobile()` branch + a collapsible accordion; this mirrors that.

## What
- **`MobilePageTitle`** (shared, all mobile detail pages): merge `actionButtonsRight` + `actionButtonsLeft` into **one horizontally-scrollable row** (`flex-nowrap` + `[&>*]:shrink-0` + `min-w-0 overflow-x-auto`). Actions keep their natural size and the row scrolls sideways within its own box instead of wrapping or causing page-level horizontal overflow. The title keeps its own line (does not regress the header-crush fix). Pages with few actions never need to scroll.
- **Session content** (`web/src/components/session/index.tsx`): a new in-file `SessionControlsBar` wrapper renders the sticky metadata/controls bar. On desktop it returns the original `<div>` with the caller-supplied className (byte-identical DOM). On mobile it collapses into a **default-collapsed accordion** — a sticky summary header (trace count · cost) + chevron toggle; the expanded body holds the full bar. Plain `useState` + a click handler, no effects. Applied to both the **v4 events default view** (`LoadedSessionEventsPage`) and the **legacy `SessionPage`**. The `parentRef` virtualized feed stays `flex-1` and takes the reclaimed height.

## Result
On mobile the session content region reclaims the height previously eaten by the two stacked headers. Desktop (≥768px) renders exactly as before.

<details>
<summary>Mechanism, scope, and verification</summary>

- **Desktop byte-identical:** `SessionControlsBar` with `isMobile=false` early-returns `<div className={desktopClassName}>{children}</div>` — same class string, same children, same DOM node. The `summary` prop is computed but unused on desktop.
- **ModernSession untouched:** the v4 call passes `isMobile={isMobile && !isModernSessionEnabled}`, so the admin-only Feature Preview always takes the desktop branch.
- **Preserved (load-bearing):** per-trace lazy virtualization (`LazySessionTraceEventsRow` + IntersectionObserver), the sticky per-card trace header, and the v4 data path (`sessions.*FromEvents`). None are touched.
- **No new effects:** accordion open state is `useState(false)` + an `onClick` toggle, matching the repo rule and the trace-view mobile pattern.
- **a11y:** the accordion toggle is a `Button` with `aria-expanded`; the summary count uses `truncate` + a `title`.
- **Checks:** `pnpm --filter web run typecheck` → exit 0, 0 TS errors. `pnpm --filter web run lint` → exit 0. Pre-commit hook: prettier `All matched files use Prettier code style!` and turbo lint `Tasks: 7 successful, 7 total`.

</details>

### DIFF
--- a/web/src/components/layouts/mobile-page-title.tsx
+++ b/web/src/components/layouts/mobile-page-title.tsx
@@ -93,14 +93,17 @@ export const MobilePageTitle = ({
         )}
       </div>
 
-      {actionButtonsRight && (
-        <div className="mt-2 flex flex-wrap items-center gap-1">
+      {/* Page actions live on ONE horizontally-scrollable row. Detail pages
+          (e.g. a session) hand us 6-9 buttons across both clusters; at phone
+          width, wrapping them stacks 200-300px of chrome above the content and
+          starves the actual page. `flex-nowrap` + `[&>*]:shrink-0` keeps each
+          action its natural size and lets the row scroll sideways within its
+          own `min-w-0`/`overflow-x-auto` box instead of wrapping or forcing
+          page-level horizontal overflow. Pages with a couple of actions never
+          need to scroll. */}
+      {(actionButtonsRight || actionButtonsLeft) && (
+        <div className="mt-2 flex min-w-0 flex-nowrap items-center gap-1 overflow-x-auto [&>*]:shrink-0">
           {actionButtonsRight}
-        </div>
-      )}
-
-      {actionButtonsLeft && (
-        <div className="mt-2 flex flex-wrap items-center gap-1">
           {actionButtonsLeft}
         </div>
       )}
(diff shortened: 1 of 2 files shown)


nkabardin:
@claude review

PR #15405, opened 2026-07-24
feat(in-app-agent): Append-only agent run events

LysanderKie:
## Summary

- replace foreground delete-and-rewrite persistence with fenced append-only event writes
- append completed units immediately; only unfinished redirect tool lifecycles stay buffered, so the Postgres event stream stays useful for incremental consumers
- persist and hydrate visible reasoning while excluding it from model replay, and tolerate unknown stored event types on read
- share one `flushPendingRunEvents` between the handler and the persistence tests, and log appends fenced by a terminal run instead of dropping events silently
- keep the implementation local to the existing web in-app-agent module

## Ordering and sequencing

- preserve existing event rows untouched (sequence numbers and timestamps); allocate new, strictly monotonic sequences under the conversation lock
- fence appends after a run reaches a terminal state
- the live SSE stream to the client is unchanged, so the display ordering from #15312 is unaffected by construction
- completed text, tool, activity, reasoning, and terminal units are never delayed by an unresolved redirect; a redirect lifecycle always lands in a single append so read-side filtering stays correct. The only persisted-order change: a render-only redirect result is stored at its completion time, so after a refresh its chip can render after sibling output it preceded live
- preserve replay and strict sequence ordering if an append-written history is later rewritten by the legacy writer during rollback

## Reasoning and redirects

- preserve visible reasoning across refreshes, merge reasoning deltas per append, and exclude reasoning and encrypted reasoning from model replay
- collapse thinking phases that hydration renders adjacently into a single thought block (display-only); live rendering is unchanged since streamed phases are separated by the text and tool content between them
- persist successful redirect actions as sanitized render-only results so they survive refresh, excluded from model replay
- discard failed redirect lifecycles/results so they leave neither broken display actions nor replay input
- ignore unknown persisted event types for forward-compatible reads while keeping current-writer event handling exhaustive with `assertUnreachable`

## Testing

- `pnpm --filter web run test in-app-agent-persistence.servertest.ts in-app-agent-api-route-auth.servertest.ts`: `Tests 49 passed (49)`
- `pnpm run lint`: `Tasks: 7 successful, 7 total`; `pnpm run typecheck`: `Tasks: 7 successful, 7 total`
- GitHub CI: `all-ci-passed`
- browser-verified on the PR preview: a real agent turn produced reasoning blocks, tool calls, and a redirect action; after a hard refresh the reasoning blocks hydrate with their content, the redirect chip survives and navigates to the trace, and the user message appears exactly once

Tests added: 2, rollback compatibility for mixed append-only and legacy writers, and collapsing adjacent hydrated reasoning into one thought block. Existing persistence tests were extended for sequence stability, reasoning hydration/replay, unknown events, redirect interleaving, and terminal fencing, and now exercise the production flush helper directly.

## Related work

- [LFE-14371](https://linear.app/clickhouse/issue/LFE-14371/append-only-event-persistence-current-foreground-path-chapter-2)
- based on the run-lifecycle schema from #15291
- does not touch the live-stream ordering fixed in #15312 (the branch is rebased onto a main that includes it)

## Notes

- No shared-package, schema, migration, queue, feature-flag, or worker changes.


LysanderKie:
@claude review

LysanderKie:
@claude review

PR #15408, opened 2026-07-24
feat(trace): mobile tabbed trace view — Tree/Timeline/Graph/Info (LFE-11067)

nkabardin:
## TL;DR
The mobile trace view is a tab bar now — **Tree · Timeline · Graph · Info** — each a full-height surface. Tap an observation in any navigator and it jumps to **Info** with that observation's detail. Desktop is untouched (byte-identical); every change is gated to the mobile path.

## Why
On mobile the peek drawer and full page both render the same `<Trace>`, which fell into an accordion: a ~384px-capped "Navigation" section stacked over a cramped detail, with **no way to switch tree ↔ timeline** and the **graph jammed into a 256px slot** — effectively unusable on a phone.

## What
A dedicated mobile tab host replaces the accordion, built on the existing `TabsBar` primitive. The four tab bodies reuse the **shared, unchanged** components (`TraceTree`, `TraceTimeline`, the graph wrapper, and `TracePanelDetail`). Cross-navigator selection already converges on the `?observation=` URL param, so "select → show detail" is wired on top of that with a tiny mobile-only layout context.

## Result
- Tree/Timeline/Graph are full-height and switchable; Graph uses the whole viewport.
- Selecting a node (tree, timeline, or graph) lands you on Info with its detail.
- Tab + selection state live in the URL (`?mobileTab`, `?observation`), so the view is shareable and Back works.
- Desktop rendering and behavior are unchanged.

<details>
<summary>Engineering detail (mechanism, desktop-safety, adaptations, follow-ups)</summary>

**Tab state.** A `?mobileTab=tree|timeline|graph|info` URL param (use-query-params) — shareable and back-able; never mirrored into local state. The **default** is computed once at mount: `info` if `?observation=` is already set (deep link / row click that targeted an observation), else `tree`. If the Graph tab is hidden (`!shouldShowGraph`, reused from `TraceContent`) and the URL asks for `graph`, it falls back to `tree`. Explicit tab taps write with `pushIn` (Back returns to the previous tab); the auto-switch to Info writes with `replaceIn` so one user action stays one history entry.

**Selection → Info (two-pronged, both additive).**
1. *Effect* — the one sanctioned selection→tab effect watches `useSelection().selectedNodeId` (the `?observation=` param) and sets the tab to `info` on a genuine change. This covers all three navigators, **including the graph**, which writes `?observation=` directly rather than through the shared setter. It's ref-guarded to skip mount (a plain deep link is handled by the default, keeping URLs clean).
2. *Imperative `switchToInfoTab()`* — for the **same-node re-tap** case, where the param (and thus the effect) doesn't change. Exposed via a small `useMobileLayoutContextOptional()` that mirrors `useDesktopLayoutContextOptional()`. `TraceTree` and `TraceTimeline` call it **alongside** their existing `layout?.expandDetailPanel()`; the graph wrapper routes its `onObservationSelect` to it. All three call-sites are `mobileLayout?.…` — **null (a no-op) on desktop**.

**Desktop byte-identical.** Mobile-only files: `TraceLayoutMobile.tsx` (reworked into the tab host + context) and the `MobileTraceContent` branch in `Trace.tsx`. The shared navigators only gain an *optional* hook whose provider is absent on desktop, so the desktop code paths are provably unchanged. The now-dead `TracePanelNavigationLayoutMobile.tsx` (old graph accordion) is removed.

**Timeline mobile fit.** The name gutter defaults to its **minimum** (~160px) when rendered under the mobile context; the chart pane keeps its own `overflow-auto`, so the waterfall scrolls **within the tab** rather than overflowing the page at 360px. A full single-column timeline redesign is a deliberate follow-up.

**Graph mobile fit.** Rendered in a full-height tab body (256px slot dropped); its own pan/zoom owns the canvas.

**Deferred / ambiguous decisions.**
- **Search** has no mobile entry point yet (its input lives in the desktop-only navigation header). Kept desktop-only for this PR; `SearchContext` is untouched and no search UI renders on mobile. Noted in code.
- The min-observation-level "hidden observations" notice + control were part of the removed mobile nav layout and have no mobile toggle; bucketed with search as a follow-up.
- Inactive tabs unmount (Radix default) to keep a single heavy subtree live at a time on constrained devices; the tradeoff is navigator scroll/zoom resets on revisit — an accepted v1 behavior (follow-up: preserve navigator state).

**Verification.** `pnpm --filter web run typecheck` → exit 0. `pnpm --filter web run lint` → exit 0 (`turbo run lint`: `Tasks: 7 successful, 7 total`). No browser run here — verified live by the reviewer against seeded v4 traces.
</details>


nkabardin:
@claude review

PR #15429, opened 2026-07-24
fix(mobile): assistant no-autofocus + useIsMobile correct on first render (LFE-11067)

nkabardin:
## TL;DR
On mobile the in-app assistant no longer auto-focuses its composer (the keyboard was springing up and burying the panel), and `useIsMobile()` is now correct on the first render so gating mount-only behavior on it actually works.

## Why
Auto-focusing the textarea on open triggers the mobile keyboard, which eats half the viewport and scrolls the panel into an unreadable state before the user wants to type. Gating that on `useIsMobile()` didn't work at first because the hook returned `false` on its first render (it resolved the media query in a `useEffect`) — so `autoFocus` fired before the flag flipped. Rather than document/keep a synchronous-read workaround, fix the hook.

## What
- **`useIsMobile()`**: rewritten with `useSyncExternalStore` — reads the media query synchronously, so it's correct on the first client render (SSR assumes desktop, hydration corrects). No effect. 16 existing consumers get the corrected value sooner.
- **Assistant**: gate both the on-open `autoFocus` and the post-turn `setInputRef` refocus on `useIsMobile()` (the latter otherwise re-sprang the keyboard after a quick-action turn). Desktop keeps auto-focusing.

## Result
Verified live: at 390px the input is not focused on open (focus stays on `<body>`, no keyboard); at 1280px it still focuses; mobile layout renders normally.

<details>
<summary>Notes</summary>

- `useSyncExternalStore` is the same pattern already used elsewhere for shared browser prefs; `getServerSnapshot` returns `false` (no viewport on the server) → no hydration mismatch.
- Only mobile programmatic focus is suppressed; desktop and mid-conversation behavior are unchanged.
- `pnpm --filter web run typecheck` → exit 0; eslint on both changed files → exit 0.

</details>


nkabardin:
@bezbac @LysanderKie FYI

just a curious observation, all interesting stuff is in description, no need to review

bezbac (approved):

nkabardin:
@claude review

PR #15430, opened 2026-07-24
fix: reduce batch cleaner query concurrency and stagger worker startup

sumerman:
## Summary

- Upon worker containers restart multiple instances may put unnecessary heavy load on CH 
- Hold the per-table Redis lock across ClickHouse counting, deletion, and failure recovery.
- Renew the lock before deletion and during recovery, and fail fast when it is unavailable.
- Stagger batch cleaner startup to reduce simultaneous load.


sumerman:
@claude review

PR #15435, opened 2026-07-24
feat(seeder): add `env` subcommand to list environments + trace counts

### LINKED ISSUE #15434: feat(seeder): add `pnpm run seed -- env` subcommand to list environments + trace counts
## Problem

`packages/shared/scripts/seeder/cli-main.ts` exposes `doctor` and `list` as subcommands but has no way to inspect what an environment looks like for a project. The CLI accepts `--environment <env>` on every scenario, but you cannot tell from the CLI:

- which environments currently exist for the seed project
- how many traces each environment has

Debugging env-related issues (e.g. `LFE-14651` "trace environment set by first-ingested span, not root span", or the `JSON-string metadata` flow in `environment-repository.servertest.ts`) currently requires going to the web UI or running a raw ClickHouse query — both of which the seeder AGENTS.md explicitly tells contributors not to do.

## Motivation

The seeder is heavily used by humans and coding agents to populate test data (`pnpm run seed -- trace-tree --v4`, `support-agent`, `long-session`, etc.). When a run produces wrong env attribution, or when verifying a seeder scenario didn't accidentally collide with a previous run's env, the only current options are:

- open the web UI and navigate to the project
- hand-write a ClickHouse query

Both are slower than the rest of the seeder workflow (which the AGENTS.md says should never need ad-hoc inserts or raw queries).

## Background

- `packages/shared/src/server/repositories/environments.ts` already exports `getEnvironmentsForProject` (returns distinct environment strings; used by `web/src/features/projects/server/projectsRouter.ts`).
- The existing function has no count. Counting requires a separate query.
- The seeder's `doctor` command pattern is the right shape to mirror: a thin wrapper that runs a query, supports `--json`, and exits 0/1.
- A dedicated `project_environments` table existed but was dropped in migration `0045_drop_project_environments`. Today envs are a column on `traces` / `observations` / `scores` — so a count needs to read from those tables, gated by `LANGFUSE_MIGRATION_V4_WRITE_MODE` like the existing function.

## Current behavior

```bash
pnpm run seed -- trace-tree --v4
# runs the scenario, prints the JSON summary, prints deep links
# no way to verify after-the-fact which envs the run produced,
# no way to compare against a prior run's envs
```

## Expected behavior

```bash
pnpm run seed -- env                        # human-readable table
# ENV                  TRACES
# default              1247
# production           312
# staging-eu           98

pnpm run seed -- env --json                # machine output
# {"projectId":"...","environments":[{"environment":"default","count":1247}, ...]}
```

## Proposed solution

1. **Repository**: add `getEnvironmentsWithCountsForProject` in `packages/shared/src/server/repositories/environments.ts`. Mirror the v3/v4 split already in `getEnvironmentsForProject`. The count column should be the number of distinct trace ids per environment (cheaper than `count()` over all rows and matches the env-semantic the UI shows).
2. **Seeder CLI**: add an `env` subcommand in `cli-main.ts` that calls the new function and prints a two-column table by default, or a JSON object with `--json`. Follow the existing `doctor` / `list` shape: parse args, run query, print, return 0/1.
3. **Docs**: add one row to the seeder's `AGENTS.md` need→command table and one line to the `README.md` examples block.
4. **Tests**: extend `web/src/__tests__/server/repositories/environment-repository.servertest.ts` (and the events-only variant) with a small assertion that the new function returns counts in the expected shape. The seeder CLI subcommand itself is a thin wrapper, so no dedicated seeder test is needed.

## Alternatives considered

- **Add the env listing to `doctor` instead of a new subcommand**: rejected — `doctor` is for stack health (PASS/WARN/FAIL with fix commands), not data inspection. Mixing them makes `doctor` output noisier and breaks the existing CI scripts that grep for PASS/FAIL.
- **Compute the count in the seeder by querying `traces` directly**: rejected — duplicates the v3/v4 routing logic that already lives in `getEnvironmentsForProject`, and risks drift if the v4 events schema changes.
- **Re-create the dropped `project_environments` table**: rejected — the team explicitly dropped it in migration `0045` (replaced by an MV in a later migration per the seeder docs; out of scope here).

## Scope

- One new repository function (~30 lines, parallel to the existing one).
- One new seeder subcommand handler (~50 lines, parallel to the `doctor` / `list` handlers).
- One entry in the seeder docs (~2 lines each).
- One test extension (~10 lines).

No public API change. No env-var change. No schema change. No migration. Pure additive CLI surface.

## Acceptance criteria

- `pnpm run seed -- env` prints a sorted table of `(environment, trace_count)` for the project.
- `pnpm run seed -- env --json` prints the same data as JSON on stdout.
- `--project <id>` overrides the default seed project id, matching the existing `doctor --project <id>` flag.
- Exits 0 on success, non-zero on ClickHouse error.
- v3 (`legacy` write mode) and v4 (`dual` / `events_only`) paths return the same data they do today for `getEnvironmentsForProject`.
- New test passes alongside the existing `environment-repository` tests.

## Backward compatibility

Pure addition. No existing CLI surface changes. The new function is additive in the shared server barrel; no consumer of `getEnvironmentsForProject` is affected.

## Risks

- The seeder already imports the shared server barrel, which means loading the CLI triggers Prisma + env validation. The new subcommand inherits the same preflight behavior as `doctor` (relying on the `cli.ts` env-precheck bootstrap). If ClickHouse is down, the query throws and the subcommand exits non-zero — same failure mode as any other subcommand that queries ClickHouse.
- The new function is called from a `package.json` script path that doesn't currently touch ClickHouse outside of scenarios. A misbehaving query in `pnpm run seed -- env` will surface in DX setups; the existing seeder tests for `getEnvironmentsForProject` cover the underlying routing, so the risk is contained.

Harsh23Kashyap:
### Problem

The seeder CLI accepts `--environment` on every scenario but has no way to inspect what an environment looks like for a project. Today the only options for "which envs exist and how big are they" are the web UI or a hand-written ClickHouse query — both of which the seeder AGENTS.md tells contributors not to do.

### Fix

Adds `pnpm run seed -- env` which queries ClickHouse for distinct environments on a project alongside a per-environment distinct-trace count, and prints either a two-column table or `--json`.

```
$ pnpm run seed -- env
ENV                  TRACES
default              1247
production           312
staging-eu           98
```

```
$ pnpm run seed -- env --json
{"projectId":"...","environments":[{"environment":"default","count":1247}, ...]}
```

`--project <id>` overrides the default seed project id, matching the existing `doctor --project <id>` flag.

### Implementation

- **`getEnvironmentsWithCountsForProject`** added in `packages/shared/src/server/repositories/environments.ts`. Same v3 (legacy `traces`) / v4 (`events_core`) split as the existing `getEnvironmentsForProject`. Count is `count(DISTINCT id)` (legacy) or `count(DISTINCT trace_id)` (v4) — one row per distinct trace per environment, matching the env-semantic the UI shows. Default environment is always present, even with zero rows.
- **`env` subcommand** added in `packages/shared/scripts/seeder/cli-main.ts` after the `list` handler. Follows the same shape: `parseArgs` → call the repository function → `printEnv(projectId, rows, json)` → return 0.
- **Usage** block in `cli-main.ts` updated to list `env` next to `doctor` and `list`. The "unknown scenario" error now also lists `env`.
- **Tests** added in `web/src/__tests__/server/repositories/environment-repository.servertest.ts` (legacy mode) and `…-events-only.servertest.ts` (v4 mode) — parallel to the existing `getEnvironmentsForProject` tests. Asserts the default env is always present and the count matches the inserted rows (including distinct-trace semantics in v4 where two events with the same `trace_id` count once).
- **Docs** updated in `packages/shared/scripts/seeder/{AGENTS,README}.md` and `.agents/skills/seed-test-data/SKILL.md` — one line in each command-examples block.

### Why this belongs

- The seeder AGENTS.md is explicit: "do not write ad-hoc ts-node scripts or raw ClickHouse inserts". The `env` subcommand is the canonical way to answer this question from the CLI.
- No new public API, no schema change, no migration. Pure additive CLI surface; the new repository function is an additive export from the shared server barrel.
- Mirrors the existing `doctor` and `list` subcommands in shape and verbosity.

### Verification

- `pnpm --filter @langfuse/shared run lint` — clean.
- `pnpm --filter @langfuse/shared run typecheck` — only the pre-existing 3 errors in `src/features/monitors/service/helpers.ts` (confirmed unrelated on `upstream/main` HEAD).
- `pnpm run seed -- help` — shows the new `env` subcommand in usage.
- `pnpm run seed -- env` — reaches the new handler (the ClickHouse query fails locally only because the dev stack is not running; a non-`unknown` error path confirms the routing).
- `pnpm run seed -- foo-bar` — error message correctly lists `env` in the available commands.

The two new `*.servertest.ts` test cases run against a real Postgres + ClickHouse (the test environment already in CI); they are added next to the existing `getEnvironmentsForProject` cases in the same files and follow the same `LANGFUSE_MIGRATION_V4_WRITE_MODE` gating.

### Risks

- New ClickHouse query (`count(DISTINCT trace_id)` on `events_core` in v4 mode, or `count(DISTINCT id)` on `traces` in legacy mode). Both are simple aggregations scoped by `project_id`; the existing `getEnvironmentsForProject` proves this routing works at scale.
- The `env` subcommand requires the same env-var precheck as every other seeder subcommand (handled by `cli.ts` before the new handler runs). No new env vars.

Fixes #15434


nkabardin:
@claude review

Steffen911:
slop

PR #15440, opened 2026-07-25
fix(web): recover from chunk loading stalls and worker errors (#15373)

### LINKED ISSUE #15373: bug: Javascript file not loading
### Describe the bug

When opening the Langfuse UI and trying to navigate to the pages Home or Tracing or Prompts, the app hangs at the Loading animation. From the inspect page of the browser, I see the reason being a Javascript file `_next/static/chunks/0f40q95x0p4fx.js` that is pending and doesn't load. If I try to open that file directly, it opens though, so it exists.

Image

### Steps to reproduce

I have self-hosted Langfuse 3.221.1 on a GCP VM using Docker Compose (following [this guide](https://langfuse.com/self-hosting/deployment/docker-compose))

Then I created one organisation, and then one project. When entering into that project, the app hangs on the loading animation. I have waited over two hours, it just never loads.

Image

I have checked with another script that traces are ingested correctly, and if I interrogate the Clickhouse database they do show up there.

Self-hosted

3.221.1

_No response_

### Additional information

_No response_

No

vivekmaddy16:
## What does this PR do?

Fixes #15373

This PR resolves an issue where the Langfuse UI hangs indefinitely at the Loading animation (`<Spinner />`) during client-side page navigation on self-hosted environments. 

Key changes:
1. **Worker Error & Timeout Fallback**: Updated `useParsedTrace` and `useParsedObservation` to clean up broken worker instances on `onerror` and added a 5-second timeout safeguard that falls back to synchronous JSON parsing (`syncParseTraceData` / `syncParseObservationData`) rather than leaving pending promises.
2. **Next.js Static Chunk Load Resilience**: Created `chunkLoadResilience` utility to intercept failed or stalled `/_next/static/chunks/*.js` script loading and automatically retry fetching the chunk with a retry flag.
3. **App Initialization**: Registered `setupChunkLoadResilience` on client app initialization in `_app.tsx`.

## Type of change

## Mandatory Tasks

## Checklist


eeshsaxena (commented):
Clearing `workerInstance` inside `onerror` means the next parse builds another worker from the same URL that just failed. `getOrCreateWorker` only constructs when the instance is null, so a chunk that fails to load turns into one new `Worker` per parse call, each one firing `onerror` again.

That interacts badly with the flush, because `pendingCallbacks` is module scoped and shared across worker generations. Each `onerror` runs `pendingCallbacks.forEach(cb => cb({ id: "", error: ... }))` and then `clear()`, so an error from a worker that has already been replaced will abort parses belonging to its replacement and push them onto the sync path even though nothing was wrong with them.

A latch avoids both. Something like a module level `workerUnavailable` flag, set next to `terminate()` in `onerror`, with `getOrCreateWorker` returning null while it is set. Callers already treat null as "parse synchronously", so later parses skip the worker, the timeout, and the respawn entirely.

On the timeout: a parse that has been in the worker for 5s is a large payload, and the fallback then parses that same payload synchronously on the main thread while the worker keeps chewing on it. That is the case the worker exists for, so for the payloads that actually hit 5s this makes the UI freeze longer rather than shorter, and does the work twice. Scaling the timeout with the estimated size (you already have `estimateSize`) or setting it much higher would keep it as a stall guard without catching the merely-slow case.

Both files have the same shape, so whatever you settle on applies to `useParsedObservation.ts` and `useParsedTrace.ts` alike.

PR #15459, opened 2026-07-27
feat(billing): CHB provider resolver, plan resolution

Steffen911:
🟡 **Live preview: [pr-15459.preview.langfuse.com](https://pr-15459.preview.langfuse.com)**

Per CHB integration plan §3.1–§3.3, §3.7, §3.9:
- cloudConfig.clickhouse block (written by CHB webhook + checkout only)
- shared getBillingProvider / hasPaidBillingState resolver with LANGFUSE_CLOUD_BILLING_CHB_CUTOFF_DATE routing (unset = feature off)
- chbCatalogue: planCode <-> Plan mapping, order keys, product-id bridge
- getPlan: clickhouse.planCode branch (manual override > CHB > Stripe > hobby)
- worker: threshold paid-gate extended via hasPaidBillingState; defensive non-Stripe skip + metric in the usage-metering loop
- env vars declared in web/worker schemas, turbo globalEnv, and .env.*.example files


packages/shared/src/interfaces/billingProvider.ts
```diff
@@ -0,0 +1,70 @@
...
+ * unparsable) means the CHB path is off entirely. Read lazily so web and
+ * worker share one implementation and tests can vary the env — format
+ * validation happens loudly in each app's env schema, so an invalid value
+ * here only ever fails closed to Stripe.
+ */
+export function getChbCutoffDate(): Date | null {
+  const raw = process.env.LANGFUSE_CLOUD_BILLING_CHB_CUTOFF_DATE;
+  if (!raw) return null;
+  const parsed = new Date(raw);
+  return Number.isNaN(parsed.getTime()) ? null : parsed;
```
Steffen911: We should use the `env` object that is validated from the env.ts file instead of accessing process.envs directly.

packages/shared/src/interfaces/billingProvider.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Resolve which billing provider owns an organization.
+ *
+ * Decision order (see CHB integration plan §3.1):
```
Steffen911: Remove references to the integration plan in comments. They should stand on their own instead of requiring an external doc. I have no intention of actually merging the integration plan.

packages/shared/src/interfaces/cloudConfigSchema.ts
```diff
@@ -23,6 +23,41 @@ export const CloudConfigSchema = z.object({
...
+  clickhouse: z
+    .object({
+      // ClickHouse Organization ID. Required and uuid-validated: parseDbOrg
+      // nulls the whole cloudConfig on parse failure, so strictness here is
+      // only safe because both writers (checkout response, webhook handler)
+      // validate the uuid before persisting. Change the stored schema first,
+      // writers second, if the id format ever loosens.
+      organizationId: z.uuid(),
+      bundleId: z.string().nullish(),
+      planCode: z.string().nullish(), // "core" | "pro" | "team" | "enterprise"; kept loose like stripe.subscriptionStatus
```
Steffen911: Why don't we reuse or map the cloudConfigPlans that we have for the `plan` on top-level here? Shouldn't they map? Then we can easily warn if there's a discrepancy. Or where would this validation happen?

web/src/features/entitlements/server/getPlan.ts
```diff
@@ -33,6 +35,18 @@ export function getOrganizationPlanServerSide(
             throw new Error(`Unhandled plan case: ${exhaustiveCheck}`);
         }
       }
+      // ClickHouse Billing plan via plan code
+      if (cloudConfig.clickhouse?.planCode) {
```
Steffen911: It feels like having an enum here makes more sense. If we get an unexpected code from the webhook, we should reject the webhook instead of storing anything in the DB. This relates to my cloudConfig comment earlier that it would be better to have confirmation. Then we can also perform an exhaustive check here.

worker/src/env.ts
```diff
@@ -24,6 +24,11 @@ const EnvSchema = z.object({
 
   STRIPE_SECRET_KEY: z.string().optional(),
 
+  // ClickHouse Billing cutoff, shared with web via the provider resolver in
+  // @langfuse/shared (getBillingProvider). The worker only consults it in the
+  // defensive usage-metering guard; unset = CHB routing off.
+  LANGFUSE_CLOUD_BILLING_CHB_CUTOFF_DATE: z.iso.datetime().optional(),
```
Steffen911: Let's make this simply a date. We don't want ti to be effective as of a date time. Makes it easier to reason about if there's a clear UTC-date cutline.

.env.dev-azure.example
```diff
@@ -98,3 +98,12 @@ LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=dual_write
...
 LANGFUSE_BACKGROUND_MIGRATION_V4_ENABLE_HISTORIC_BACKFILL=false
+
+# ClickHouse Billing (CHB) integration (Langfuse Cloud only). Cutoff routes
+# first-time upgrades on/after this instant to CHB; unset = feature off.
+# LANGFUSE_CLOUD_BILLING_CHB_CUTOFF_DATE=
+# CLICKHOUSE_BILLING_BASE_URL=
+# CLICKHOUSE_BILLING_SERVICE_TOKEN=
+# CLICKHOUSE_BILLING_WEBHOOK_SIGNING_SECRET=
+# CLICKHOUSE_BILLING_METRICS_API_KEY=
+# CLICKHOUSE_BILLING_EVENT_BUS_URL=
```
Steffen911: Let's drop them from the dev examples that don't need it. Keep it to .env.dev.example and .env.prod.example only.

packages/shared/src/interfaces/billingProvider.ts
Steffen911: Fixed in 00da897d61.

Importing the shared \`env\` object here would not work: \`billingProvider.ts\` is re-exported from \`packages/shared/src/index.ts\`, the client-safe barrel, and \`packages/shared/src/env.ts\` runs \`EnvSchema.parse(process.env)\` over the full server schema at module load. Nothing in that barrel pulls it in today, so importing it would drag the server env into client bundles.

Inverted it instead — \`getChbCutoffDate\` is gone and \`getBillingProvider(org, { cutoff, now })\` takes the cutoff as an argument, so each app passes the value from its own validated env schema. The usage-metering job is the only caller today. Shared no longer touches \`process.env\` at all.

Side effect: the \`turbo.json\` \`globalEnv\` entry is also gone. It only existed because the explicit \`process.env\` member access tripped \`turbo/no-undeclared-env-vars\` (off for web, active in shared) — and since \`globalEnv\` feeds the global task hash, changing the cutoff would have busted the entire turbo cache.

packages/shared/src/interfaces/billingProvider.ts
Steffen911: Fixed in e6cca7c947 — both references (\`§3.1\` here, \`§3.4\` in \`chbCatalogue.ts\`) now carry their reasoning inline. You deleted the plan doc yourself in eaf2558b5b, which my series is rebased onto.

worker/src/env.ts
Steffen911: Fixed in 66deaa8a20 — \`z.iso.date()\` (\`YYYY-MM-DD\`) in both `worker/src/env.ts` and `web/src/env.mjs`, changed together so the two schemas cannot drift. \`new Date()\` on a date-only string is UTC midnight per spec, so the cutline is one unambiguous instant.

.env.dev-azure.example
Steffen911: Fixed in 00f8e13af7 — the block now exists only in \`.env.dev.example\` and \`.env.prod.example\`. Dropped from \`.env.dev-azure.example\`, \`.env.dev-oci.example\`, \`.env.dev-redis-cluster.example\` and \`.env.test.example\`.

Also dropped the five \`CLICKHOUSE_BILLING_*\` declarations from \`web/src/env.mjs\` and both surviving examples. Nothing in this PR reads them; the PRs that add the client, webhook and metrics API can declare what they actually use.

packages/shared/src/interfaces/cloudConfigSchema.ts
Steffen911: Fixed in 6c8599dfc2. They do map — CHB codes are the lowercased *paid* tiers of \`cloudConfigPlans\`; there is deliberately no code for the free tier, because a free org just carries no bundle.

Both now live in shared next to \`cloudConfigPlans\` (they had to move out of \`chbCatalogue.ts\` — shared cannot import from web), with a \`satisfies Record<ChbPlanCode, Plan>\` clause so a code without a plan is a compile error rather than a runtime warning.

On where validation happens: strict rejection belongs at the webhook, and PR 07 will use the bare \`z.enum(chbPlanCodes)\` on the inbound payload so an unknown code is never stored. The *stored* schema here uses \`.catch(null)\` rather than a bare enum on purpose — \`parseDbOrg\` nulls the **entire** \`cloudConfig\` on any parse failure, so one drifted code would also discard the org's \`plan\` override, \`rateLimitOverrides\` and \`stripe.customerId\` — and an org without a Stripe customer id drops out of the usage-metering job's selection and silently stops being billed. \`.catch(null)\` keeps the blast radius to the one field.

Flagging that as a decision rather than a detail: if you would rather have the loud failure, the alternative is a bare \`z.enum\` plus a metric on the \`parseDbOrg\` failure path — otherwise the failure is not actually loud, just destructive.

web/src/features/entitlements/server/getPlan.ts
Steffen911: Fixed in 6c8599dfc2 — this is now an exhaustive lookup, and the unknown-code path is gone rather than handled.

\`chbPlanCodeToPlan\` is a total \`Record<ChbPlanCode, Plan>\` in shared, and \`cloudConfigSchema\` validates stored codes against the same enum, so there is no \`null\` return left to branch on. Agreed on rejecting at the webhook: PR 07 will validate the inbound payload with the bare enum so an unrecognized code never reaches the DB. See the reply on the \`cloudConfigSchema\` thread for why the stored schema keeps a \`.catch(null)\` rather than a bare enum.

One thing that fell out of this: I restructured the branch to key on \`clickhouse.organizationId\` instead of \`planCode\`. The old shape fell through to the Stripe branch when the code was unrecognized, so an org carrying both a drifted CHB code and a stale \`stripe.activeProductId\` resolved to the *paid* Stripe plan. New test in \`web/src/__tests__/server/unit/getPlan.servertest.ts\` fails with \`expected 'cloud:team' to be 'cloud:hobby'\` against the old shape — I verified that by reverting the branch locally. That also closes the two claude[bot] threads on this file.

web/src/features/entitlements/server/getPlan.ts
```diff
@@ -33,6 +35,18 @@ export function getOrganizationPlanServerSide(
...
+        }
+        // Unknown plan codes fail open to the free tier, never to a paid tier.
+        logger.error(
+          `getOrganizationPlanServerSide: unknown CHB plan code ${cloudConfig.clickhouse.planCode}, resolving to cloud:hobby`,
+        );
+      }
+
       // stripe plan via product id
       if (cloudConfig.stripe?.activeProductId) {
         const stripePlan = mapStripeProductIdToPlan(
```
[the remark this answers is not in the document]
Steffen911: Confirmed and fixed in 6c8599dfc2, structurally rather than with the suggested one-line \`return\`.

The branch now keys on \`clickhouse.organizationId\` instead of \`planCode\`, so a CHB org can never reach the Stripe branch — the same CHB-wins precedence \`getBillingProvider\` enforces. Combined with a total \`chbPlanCodeToPlan\` lookup and an enum on the stored \`planCode\`, the unmapped-code branch you describe no longer exists.

Reproduced before fixing: \`web/src/__tests__/server/unit/getPlan.servertest.ts\` fails with \`AssertionError: expected 'cloud:team' to be 'cloud:hobby'\` when the branch is reverted to the field-keyed shape. Your note that the file had no test coverage at all was correct; it does now.

web/src/features/entitlements/server/getPlan.ts
```diff
@@ -33,6 +35,18 @@ export function getOrganizationPlanServerSide(
...
+      // ClickHouse Billing plan via plan code
+      if (cloudConfig.clickhouse?.planCode) {
+        const chbPlan = mapChbPlanCodeToPlan(cloudConfig.clickhouse.planCode);
+        if (chbPlan) {
+          return chbPlan;
+        }
+        // Unknown plan codes fail open to the free tier, never to a paid tier.
+        logger.error(
+          `getOrganizationPlanServerSide: unknown CHB plan code ${cloudConfig.clickhouse.planCode}, resolving to cloud:hobby`,
+        );
```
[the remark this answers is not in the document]
Steffen911: Fixed in 6c8599dfc2. Your read of the two call sites was right — \`findDbKeyOrThrow\` on the Bearer path has no Redis caching, so this would have logged per request.

Rather than downgrade to \`warn\`, the log is gone: \`chbPlanCodeToPlan\` is a total \`Record<ChbPlanCode, Plan>\` and \`cloudConfigSchema\` validates stored codes against the same enum, so there is no unmapped-code branch to log from. The \`logger\` import is out of \`getPlan.ts\` entirely, which also removes the asymmetry with the silent \`stripe.activeProductId\` fallthrough you flagged.

Steffen911:
@claude review

worker/src/env.ts
Steffen911: Follow-up in 27bda821c2: the date-only string is now parsed into a \`Date\` inside both env schemas, so consumers get the instant instead of repeating the ternary. Non-scalar transforms were already the pattern in both files (three comma-separated lists in worker, six transforms in web).

Verified the chain rather than assuming it: \`"2026-01-01"\` parses to \`2026-01-01T00:00:00.000Z\`, unset yields \`null\`, and empty strings, full datetimes and garbage are all rejected. Confirmed separately that t3-env's \`createEnv\` really applies the transform (a throwaway test asserted \`instanceof Date\` through the proxy, then was deleted). Blank values cannot reach the transform anyway: web sets \`emptyStringAsUndefined\`, worker runs \`removeEmptyEnvVariables\`.

One caveat worth knowing, not introduced here: under \`DOCKER_BUILD=1\` both apps bypass validation (\`skipValidation\` in web, \`process.env as any\` in worker), so during image builds this field is the raw string while its type says \`Date | null\`. That already applies to every coerced and transformed field in both schemas, and nothing reads this one at build time.

PR #15309, opened 2026-07-22
feat(billing): clickhouse billing integration

Steffen911:

PR #15461, opened 2026-07-27
feat(billing): CHB project events to event bus

Steffen911:
🟡 **Live preview: [pr-15461.preview.langfuse.com](https://pr-15461.preview.langfuse.com)**

CHB needs to know which projects exist per organization to poll the billing metrics API, so project create/delete/transfer emit `LANGFUSE_PROJECT_CREATED` / `LANGFUSE_PROJECT_DELETED` to CHB's event bus. Wired into both the tRPC router and the admin API, since either can create or delete a project.

### Transport: cross-account EventBridge, not HTTP

The bus turned out to be a cross-account Amazon EventBridge bus, not an HTTP endpoint. ClickHouse provisioned it as a bus resource policy that trusts our whole AWS account and whitelists the allowed event types **by `detail-type`**, and asked us to grant our own task role `events:PutEvents` against it. The bearer-token POST this started as could never have authenticated against that, so delivery is now `PutEvents`:

- the event type travels as `detail-type`, because that is what CHB's policy filters on — not only inside `Detail`
- `PutEvents` answers `200` with a per-entry error when the bus rejects an event, so `FailedEntryCount` rather than the HTTP status is the real result
- auth is SigV4 from the task role; no shared secret is involved

Declares `CLICKHOUSE_BILLING_EVENT_BUS_ARN` (the ARN also carries the bus region — CHB's regions do not match ours, and the ARN is the only place that region appears). `CLICKHOUSE_BILLING_SERVICE_TOKEN` is declared for CHB's REST API in the next PR in the stack, not for the bus.

The Langfuse-side IAM grant is a separate infrastructure change: `events:PutEvents` scoped to the single bus ARN, on the `web` task role only, enabled in staging while production stays unset.

### Sync starts at checkout

Nothing is mirrored to CHB until checkout persists `cloudConfig.clickhouse.organizationId` — before that CHB has no organization to attribute a project to, and the vast majority of orgs never check out at all. `resolveChbOrganizationId` is the single gate enforcing that for every outbound event.

`backfillChbProjectEvents` closes the gap that creates: it sends `LANGFUSE_PROJECT_CREATED` for the projects an org already had, so CHB does not end up metering only post-checkout projects. The call site is on the first checkout (stack 8/8); it is deliberately first-checkout-only, since a retry reuses the same CH organization.

### Failure behavior

Best-effort, no queue: the bus is an always-on managed endpoint, and at-most-once is acceptable because a lost delete is benign (the metrics API returns zeros for deleted projects) and a lost create is repaired by CHB's own backfill pipeline. Emits never throw into the caller, so project create/delete latency and success are unaffected, and a failed backfill can never fail checkout. `langfuse.billing_events.emit_failed` — tagged `source:request` or `source:backfill` — is the signal that these assumptions stopped holding.

Split from #15309 (stack 5/8). Its parent #15460 has landed, so this now targets `main` directly.


web/src/features/projects/server/projectsRouter.ts
```diff
@@ -210,6 +218,14 @@ export const projectsRouter = createTRPCRouter({
...
+      // billable now, not when the async hard-delete worker finishes.
+      emitChbProjectEvent({
+        type: "LANGFUSE_PROJECT_DELETED",
+        orgId: ctx.session.orgId,
+        projectId: input.projectId,
+      });
+
       const projectDeleteQueue = ProjectDeleteQueue.getInstance();
       if (!projectDeleteQueue) {
         throw new TRPCError({
```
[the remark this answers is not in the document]
Steffen911: Good catch — real gap, fixed in 7e905e5a87. `transfer` now emits `LANGFUSE_PROJECT_DELETED` for the source org and `LANGFUSE_PROJECT_CREATED` for the destination, after the transaction and cache invalidation. Each emit still no-ops unless that org carries CHB state, so a Stripe-to-Stripe transfer stays silent.

Regression test: "emits a delete for the source org and a create for the destination on transfer" in `projectsRouter-chbProjectEvents.servertest.ts` — 0 emits instead of 2 against the previous shape. The module also got its own unit test, including the no-op-without-CHB-state invariant the fix leans on.

One correction to the finding: the plan doc is not part of this stack anymore (deleted before #15459 merged), so there is no emit-point list to reconcile against — the call sites are the contract now.

.env.prod.example
```diff
@@ -386,3 +386,8 @@ LANGFUSE_ENABLE_BLOB_STORAGE_FILE_LOG=true
 # LANGFUSE_CLOUD_BILLING_CHB_CUTOFF_DATE=
 # Bearer secret for GET /api/billing/metrics; unset = endpoint disabled.
 # CLICKHOUSE_BILLING_METRICS_API_KEY=
+# CHB's EventBridge bus for project lifecycle events, authenticated by IAM from
+# the task role; unset = no events emitted.
+# CLICKHOUSE_BILLING_EVENT_BUS_ARN=
+# Bearer token for CHB's REST API (not the event bus).
+# CLICKHOUSE_BILLING_SERVICE_TOKEN=
```
Steffen911: Do we need this in this PR? I thought the event bus arn + updating the auth role should be sufficient.

web/src/env.mjs
```diff
@@ -415,6 +415,20 @@ export const env = createEnv({
...
+    CLICKHOUSE_BILLING_EVENT_BUS_ARN: z
+      .string()
+      .regex(
+        /^arn:aws:events:[a-z0-9-]+:\d{12}:event-bus\/.+$/,
+        "must be an EventBridge bus ARN (arn:aws:events:<region>:<account-id>:event-bus/<name>)",
+      )
+      .optional(),
+    // Bearer token we present to CHB's REST API (checkout sessions, bundles,
+    // invoices). Not used for the event bus, which authenticates via IAM.
+    CLICKHOUSE_BILLING_SERVICE_TOKEN: z.string().optional(),
```
Steffen911: Probably this can come via a later PR?

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
```diff
@@ -0,0 +1,257 @@
...
+ * CHB needs to know which projects exist per organization to poll our billing
+ * metrics API. Delivery is a direct best-effort cross-account EventBridge
+ * PutEvents from web, no queue: the bus is an always-on managed endpoint;
+ * absorbing brief unavailability is CHB's responsibility. At-most-once is
+ * acceptable by design — a lost PROJECT_DELETED is benign (the metrics API
+ * returns zeros for deleted projects by contract), a lost PROJECT_CREATED
+ * delays metering for one project until CHB's own backfill pipeline catches
+ * it. If `langfuse.billing_events.emit_failed` ever shows real loss, swap this
+ * helper's internals for a queue — call sites keep the same signature.
+ *
```
Steffen911, line 24: Make sure that there's an error log message whenever an event is not accepted and a created vs delete lost metric that we track. Initially, this should work and if it surfaces repeatedly via alerts, we can look into an internal queue here.

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
```diff
@@ -0,0 +1,257 @@
...
+    client: new EventBridgeClient({
+      region,
+      requestHandler: {
+        requestTimeout: EVENT_BUS_REQUEST_TIMEOUT_MS,
+        throwOnRequestTimeout: true,
+      },
+    }),
+  };
+  return cachedClient.client;
+};
```
Steffen911: We should only have one region per deployment. I.e. a global singleton should be fine with no need to make it a map of clients.

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
```diff
@@ -0,0 +1,257 @@
...
+  let failed = 0;
+
+  try {
+    const chbOrganizationId = await resolveChbOrganizationId(params.orgId);
+    if (!chbOrganizationId) return { sent, failed };
+
+    const projects = await prisma.project.findMany({
+      where: { orgId: params.orgId, deletedAt: null },
+      select: { id: true },
+    });
```
Steffen911, line 279: Do we need to paginate this call? What's the prisma behaviour here?

.env.prod.example
Steffen911: Agreed — dropped from this PR in [`970bdc4`](https://github.com/langfuse/langfuse/commit/970bdc46f91ff19713042aafe27ec4ebdf6d43ea). Nothing here reads it any more now that the bus authenticates via SigV4 from the task role, so the ARN plus the IAM grant really is sufficient, as you said.

It moves to #15462, where `chbApiClient.ts` is the only consumer: [`593f49f`](https://github.com/langfuse/langfuse/commit/593f49fed81fc6b5ef1442b41ac4199f9db16445).

web/src/env.mjs
Steffen911: Yes — moved to #15462 in [`593f49f`](https://github.com/langfuse/langfuse/commit/593f49fed81fc6b5ef1442b41ac4199f9db16445) and removed from this PR (schema + `runtimeEnv` + both `.env` examples) in [`970bdc4`](https://github.com/langfuse/langfuse/commit/970bdc46f91ff19713042aafe27ec4ebdf6d43ea).

`CLICKHOUSE_BILLING_EVENT_BUS_ARN` is now the only env var this PR declares. Worth noting it also carries the bus region, since CHB's bus regions don't line up with ours and the ARN is the only place that region appears.

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
Steffen911: The error log was already in place: a rejected event throws out of `sendChbProjectEvent`, and both callers catch it into `logger.error` — after backoff exhausts, so retries don't generate noise.

The created-vs-delete split was genuinely missing, though: the metric only carried `unit` and `source`, so a lost create and a lost delete were indistinguishable. Added an `event_type` tag on every `emit_failed` path in [`970bdc4`](https://github.com/langfuse/langfuse/commit/970bdc46f91ff19713042aafe27ec4ebdf6d43ea). They fail differently enough to be worth separating on the alert — a lost create under-meters an org, a lost delete keeps billing a project that no longer exists.

Agreed on holding off on a queue until the alert says otherwise.

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
Steffen911: Right — collapsed to a lazy singleton in [`970bdc4`](https://github.com/langfuse/langfuse/commit/970bdc46f91ff19713042aafe27ec4ebdf6d43ea).

The region key was pointless: the ARN comes from env and can't change at runtime, so the comparison never varied and it only implied a flexibility that doesn't exist.

web/src/ee/features/billing/server/chb/chbProjectEvents.ts
Steffen911: No pagination needed. Prisma applies no default `take` to `findMany`, so it returns every matching row — and a cap here would be the more dangerous option, because it would silently under-backfill with nothing to signal it. Only `id` is selected, so the result stays small even for a very large org. I noted both points in the docstring.

The real cost was the send loop, not the query: it was one `PutEvents` call per project. Batched in tens (the PutEvents limit) in [`970bdc4`](https://github.com/langfuse/langfuse/commit/970bdc46f91ff19713042aafe27ec4ebdf6d43ea), so fifty projects cost five sequential calls instead of fifty.

One thing that came out of that: retries now carry only the entries the bus is still rejecting. `PutEvents` fails per entry, so resending a whole batch because one entry was throttled would duplicate the ones that already landed. A rejection also carries its failed-entry count now, so the metric reflects events actually lost rather than whole batches.

Steffen911:
@claude review

Steffen911:
@claude review

PR #15462, opened 2026-07-27
feat(billing): implement CHB api client

Steffen911:
🟡 **Live preview: [pr-15462.preview.langfuse.com](https://pr-15462.preview.langfuse.com)**

`ChbApiClient` is the only module that knows the CHB wire format: bundle reads and mutations, checkout and portal sessions, invoices. zod-parsed responses that ignore unknown fields, a typed 409 `ChbPaymentRequiredError`, and idempotency keys via the extended `IdempotencyKind` machinery.

`ChbBillingService` exposes the same router-facing surface as the Stripe `BillingService`, so the `cloudBillingRouter` procedures can dispatch to either provider without a tRPC contract change. Checkout persists `clickhouse.organizationId` with Stripe-state and manual-plan interlocks.

Re-adds a much smaller `chbCatalogue`: main deleted it when the plan-code → plan mapping moved to shared, and the only piece that cannot live there is the Stripe product-id bridge the plan-selection mutations still need. Order keys are read off `stripeProducts` rather than duplicated, so an upgrade cannot be classified differently depending on which provider bills the org.

## Auth: Auth0 client credentials, not a shared secret

CHB's billing-api authenticates inbound requests with Auth0 M2M — it verifies the bearer token's issuer, audience and RS256 signature against its tenant's JWKS on every route except the health probes (ClickHouse/control-plane#38250). A static shared secret is rejected outright, so the `CLICKHOUSE_BILLING_SERVICE_TOKEN` this PR originally carried had no possible use and is gone.

`chbAccessToken` instead exchanges the credentials of CHB's non-interactive `Langfuse Billing` client for an access token scoped to the `billing-api` audience. It caches the token, renews five minutes ahead of expiry so a request cannot arrive carrying a token that expired in transit, and single-flights the grant so a burst of billing requests does not stampede Auth0. `ChbApiClient` replays a request once on a 401: CHB did no work, and the replay carries the same idempotency key, which turns a revoked token or a rotated signing key into a retry rather than a user-visible billing failure.

New env vars, all required together — `createChbApiClientFromEnv` fails closed so a partly configured deployment never reaches CHB unauthenticated:

- `CLICKHOUSE_BILLING_BASE_URL`
- `CLICKHOUSE_BILLING_AUTH0_DOMAIN`
- `CLICKHOUSE_BILLING_AUTH0_CLIENT_ID`
- `CLICKHOUSE_BILLING_AUTH0_CLIENT_SECRET`
- `CLICKHOUSE_BILLING_AUTH0_AUDIENCE` (defaults to `billing-api`)

Rollout is ordered: CHB enforces auth as soon as billing-api deploys, so the `Langfuse Billing` client id and secret have to be in our secret store before that.

## Tests

69 web `server-unit` tests across the client, the token provider, the billing service and the catalogue. They pin the decisions that are ours rather than CHB's, so they stay meaningful while the wire format is still being reconciled: auth headers and redirect refusal, path-prefix-preserving url joins, bundle-id escaping, 409 → `ChbPaymentRequiredError` → `PRECONDITION_FAILED`, upgrade-now vs downgrade-at-cycle-end, the checkout interlocks, and the no-bundle fallbacks that keep hobby orgs and org deletion working.

Two things they turned up: an unparseable CHB invoice timestamp reached the invoice table as `NaN`, and `IdempotencyKind` declared a `chb.checkout.create` kind that nothing can generate. Both fixed.

Split from #15309 (stack 6/8). Rebased onto main now that #15461 has merged.


Steffen911:
What to doubt here, roughly in order:

**The 401 replay.** `ChbApiClient` retries once when CHB answers 401. My argument is that a 401 means CHB did no work and the replay carries the same `Idempotency-Key`, so it is safe — but that rests on CHB rejecting at the middleware, before the route touches anything. It does today (`m2mAuthMiddleware` runs before `routes`), and this breaks if CHB ever starts returning 401 from inside a handler. Worth deciding whether we want the retry at all versus surfacing the failure.

**Renewing 5 minutes early on a 24h token.** Arbitrary. It only has to exceed the worst-case request latency plus clock skew between us and Auth0. The odd-looking `Math.max`/`Math.min` in `mintToken` exists so a short-lived token does not re-grant on every call; if CHB never issues anything shorter than an hour, that branch is dead weight and could go.

**Credentials in env rather than a fetched secret.** `CLICKHOUSE_BILLING_AUTH0_CLIENT_SECRET` is a long-lived secret in the task environment, same as our other provider secrets. Fine by precedent, but this one is rotated by CHB, not us, and a rotation is a deploy on our side.

**Fail-closed on partial config.** `createChbApiClientFromEnv` returns null unless the base url and all three credentials are set. Null means "CHB unavailable" to callers, which #15464 turns into a refusal rather than a fallback to Stripe. Check that is the behaviour we want for a misconfigured region.

**One base url, three CHB regions.** Staging exposes billing-api in us-east-2, eu-central-1 and ap-southeast-1 (ClickHouse/data-plane-configuration#82814). A single env var per deployment covers it only if each Langfuse region should talk to exactly one CHB region. Nobody has written down that mapping yet.

**The test wire format.** The response schemas are still deliberately loose, and the tests assert our own mapping decisions rather than CHB's payload shape — on purpose, so they do not have to be rewritten when the contract lands. That also means they will not catch a CHB field rename. The client tests do assume CHB rejects on the middleware and that our own url join is what I claim; both are worth a skim.

**`isChbUpgrade` on an unknown plan code.** An unrecognised code resolves to order key 0, so switching *from* an unknown code reads as an upgrade and applies immediately. Test pins it. It is the conservative direction, but it is a guess.

**Local dev has no story.** CHB mints mock JWTs under `STAGE=local` and we cannot produce those, so the CHB path is staging-and-up only. Not addressed here.

Steffen911:
@claude review

Steffen911:
Follow-up on the checkout idempotency question — what a reviewer should doubt in the latest commit:

**The claim is a raw `UPDATE`, the first raw SQL against `cloud_config` in the repo.** Everything else reads the column through Prisma and `parseDbOrg`. I went raw because the guard has to be a `WHERE` predicate to be single-winner, and Prisma's JSON path filters can't reliably express "key is absent" on Postgres. Worth doubting whether that's the right precedent to set here, versus adding a real column for the CH organization id and getting a unique constraint for free — which is what I'd actually want long term.

**The `jsonb_set` merge is hand-rolled and the unit tests mock it out entirely.** I verified it directly against Postgres instead — six `cloud_config` shapes (NULL, `{}`, unrelated keys, a `clickhouse` object without an id, an id already set, a JSON-`null` id) plus two racing transactions claiming 1 and 0 rows. But that evidence lives in this comment, not in CI, so a later edit to the statement has no test that would catch a regression.

**Losing the race returns `CONFLICT` and leaves an orphaned CH organization behind.** The payment can't misroute, which was the point, but we still created something in CHB that nobody references. I log the id rather than clean it up — there is no CHB delete-organization call in the client, and I don't know if one exists.

**The idempotency key trusts CHB's replay semantics, which are still unconfirmed.** If CHB errors on same-key-different-params rather than returning the cached response, the `planCode` in the key is what keeps a plan switch from becoming a user-visible failure. If it errors on same-key-*same*-params, a genuine double-click gets an error instead of the original URL. The client also still sends the key both as an `Idempotency-Key` header and in the request body, which I'd want CHB to confirm is right.

**The retry-mismatch check is a guess about severity.** If CHB hands back a different organization id than the one we asked it to reuse, I fail the request. Refusing is the safe read, but if CHB legitimately re-homes an organization, this turns a recoverable state into a hard block on checkout for that org.

Steffen911:
@claude review

Steffen911:
@claude review

Steffen911:
Three follow-ups on the last review pass, and one thing I am deliberately not changing.

**Token buffer (`47e1ab9`, first of three).** The cached-lifetime formula was `max(expires_in - margin, min(expires_in, margin))`, which pins to the margin itself for any TTL between one and two margins — a 310s token was treated as fresh for 300s, leaving 10s before real expiry instead of the 300s the margin exists to hold. Now the margin is capped at half the token's life, so the buffer shrinks with the token rather than collapsing. Reachable cost was small (Auth0 issues 24h tokens here, and the client re-mints on a 401), but the code was claiming a margin it did not keep. The new test fails against the old formula; the existing short-token test now advances 25s instead of 30s, because a 60s token caches for 30s rather than its full life.

**`mapScheduled` (same push).** The comment claimed `"immediate"` changes never linger as pending state; the code never read `scheduled.when` and resolved a date from `startDate ?? period.endDate`. So an immediate cancellation echoed back on a bundle with a live period rendered as "cancels at end of cycle" — the invariant only held when no date resolved at all. One-line guard plus a test that fails without it. Still display-only and still gated on CHB echo behaviour nobody has confirmed, but the comment is now true.

**What I am not fixing: the checkout `safeParse`.** The review is right that it cannot fire for the case its comment describes — `ChbApiClient` hard-parses the same `z.uuid()` field, so a malformed id from CHB throws inside the client before the service sees it. I am keeping the block anyway: it is the gate on the *write*, not on the wire, and what it protects against is the stored schema and the response schema drifting apart while the CHB contract is still loose. Deleting it trades a cheap check for a `parseDbOrg` failure that nulls the whole `cloudConfig`. What I did change is the comment and the test name, which both claimed to catch a ZodError they never see.

Doubt that skip if you think the wire schema and the stored schema should just be the same type, in which case the block really is dead and should go. Also worth doubting the halved margin: a 60s token now re-grants twice per lifetime instead of once, which is fine at our request volume but is a real change in Auth0 call pattern if CHB ever shortens TTLs a lot.

Checks: `Tasks: 8 successful, 8 total` (lint), `Tasks: 8 successful, 8 total` (typecheck), and the four CHB unit suites — `Tests 10 passed (10)`, `Tests 40 passed (40)`, `Tests 23 passed (23)`, `Tests 7 passed (7)`. No preview run: none of this is user-visible without CHB credentials, which previews do not carry.

PR #15463, opened 2026-07-27
feat(billing): CHB inbound webhook processor

Steffen911:
🟡 **Live preview: [pr-15463 app preview](https://pr-15463.preview.langfuse.com)**

`clickhouse-webhook` route + handler, the single writer of `cloudConfig.clickhouse`: HMAC-SHA256 verification → Redis `SET NX` dedupe → region fan-out via JSONB `organizationId` lookup → monotonic `lastEventCreatedAt` ordering guard → `attachedplan.{created,updated,scheduled,cancelled}` effects → cached-API-key invalidation + audit log.

`bundle.created` backfill-emits project events for projects that predate checkout. It seeds no default spend alerts: the spend-alert job computes spend from the org's own Stripe subscription, which a CHB org does not have, so seeded alerts would show in billing settings and never fire.

Wire format follows the ClickHouse control plane's signed-webhook dispatcher: one `X-CHB-Signature: t=<unix seconds>,v1=<hex>[,v1=<hex>...]` header, HMAC-SHA256 over `<t>.<raw body>`, verified over the bytes as received. Several `v1` values appear while CHB rotates its key and any one matching our secret passes, so a rotation needs nothing on our side beyond swapping `CLICKHOUSE_BILLING_WEBHOOK_SIGNING_SECRET` during the overlap. The envelope is `{eventId, type, occurredAt, data}` with the owning ClickHouse `organizationId` inside `data`. Event types and `data` follow the billing integration spec: `attachedplan.created|updated|scheduled|cancelled`, `data.id` for the attached plan (stored as `clickhouse.bundleId`), `payment.dueDate` (stored as `nextPaymentDate`), and the `scheduled` union (`cancel` with `endDate`, `upgrade`/`downgrade` with `planCode` and `startDate`). An unknown plan code is refused with a 500 so the dispatcher keeps retrying until the mapping ships. Still open: where `organizationId` and the event time land once the emitter is settled, see the question in the comments.

Declares `CLICKHOUSE_BILLING_WEBHOOK_SIGNING_SECRET`, which this is the first code to read.

Three review findings are fixed in this PR, each with a test that fails against the shape they were found in:

- `bundle.created` cleared `cloudFreeTierUsageThresholdState` unconditionally, un-blocking ingestion for a suspended org whose initial payment was pending or failed. Now gated on `payment.status === "active"`, the same gate `bundle.updated` and the Stripe path use.
- The Redis dedupe claim is taken by the check itself, before dispatch, and dispatch was unguarded — so a transient Prisma failure 500'd with the claim held, CHB's retry was dropped as a duplicate, and the event was lost for the full 24h TTL. Dispatch is wrapped now, and a failed apply releases the claim before we ask for the retry.
- The interlock refusing to write CHB state onto a Stripe org could never fire (the org is resolved *by* its clickhouse org id). Replaced with the contradiction that matters: a live `stripe.activeSubscriptionId` alongside CHB state means the customer is billed twice.

Split from #15309. No longer stacked: rebased onto main now that #15462 has merged, which dropped the commits this branch was carrying as duplicates of already-merged work (the client and service, and the project-event commits from #15461). What is left is the webhook alone — seven files, no dependency changes, and no file in common with #15464, so the two can merge in either order.

Only the webhook signing secret is new here; the Auth0 credentials it used to declare came with #15462.


web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,507 @@
...
+    orgColumns: {
+      // First paid subscription anchors the billing cycle on the bundle start
+      cloudBillingCycleAnchor: data.startDate
+        ? new Date(data.startDate)
+        : startOfDayUTC(new Date()),
+      // Un-suspend: the org is now a paying customer
+      cloudFreeTierUsageThresholdState: null,
+    },
+  });
+
```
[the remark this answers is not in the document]
Steffen911: Fixed in 8468764a45. `handleBundleCreated` now computes the same `isPaidAndCurrent` gate as `handleBundleUpdated` and only clears `cloudFreeTierUsageThresholdState` when it holds; the bundle itself (and the cycle anchor) is still recorded either way, and `bundle.updated` lifts the suspension when the payment goes active.

Regression tests in `chbWebhookHandler.servertest.ts`: "leaves the free-tier suspension in place on bundle.created with a pending initial payment" and "…with no payment block at all", plus the positive "un-suspends on bundle.created only once payment is active".

Note the no-payment-block case is now also gated, which the finding did not call out explicitly — an envelope without payment info is exactly as unproven as a pending one, and it is the shape most likely to show up if CHB fires `bundle.created` before the first charge settles.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,507 @@
...
+      "NX",
+    );
+    // ioredis returns null when NX prevented the write → key existed
+    return redis ? result === null : false;
+  } catch (error) {
+    logger.warn("[CHB Webhook] Redis dedupe check failed, processing event", {
+      eventId,
+      error,
+    });
+    return false;
```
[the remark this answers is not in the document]
Steffen911: Confirmed and fixed in 8468764a45. The dispatch is now wrapped, and a failed apply calls `releaseEventClaim(event.id)` (Redis `DEL`) before returning the 500, so CHB's retry is processed rather than swallowed. The claim is deliberately kept on the paths that are decisions rather than failures — unknown org (region fan-out), out-of-order event, and the Stripe-contradiction refusal — since a retry cannot change those outcomes.

Regression test: "releases the dedupe claim when applying the event fails" (500 + `DEL chb-webhook-event:evt_1`), with "keeps the dedupe claim when the event applied cleanly" as the counterpart so the release cannot silently become unconditional.

I took the release-on-failure option rather than claiming only after success: the claim has to be taken before the work to be a mutual exclusion at all, otherwise two concurrent deliveries of the same event both pass the check.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,507 @@
...
+
+  // Interlock (must never happen): never write CHB state onto a Stripe org.
+  if (getBillingProvider(parsedOrg) === "stripe") {
+    logger.error(
+      `[CHB Webhook] Org ${parsedOrg.id} resolves to the Stripe provider, refusing to apply ${event.type}`,
+    );
+    traceException(
+      `[CHB Webhook] Org ${parsedOrg.id} resolves to the Stripe provider, refusing to apply ${event.type}`,
+    );
+    return NextResponse.json({ received: true }, { status: 200 });
```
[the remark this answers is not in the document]
Steffen911: Right, it was unreachable — `getBillingProvider` returns `"clickhouse"` on `cloudConfig.clickhouse.organizationId`, which is precisely how `getOrgByChbOrganizationId` found the org.

Rather than delete the interlock, 8468764a45 points it at the contradiction that actually matters: an org carrying a live `stripe.activeSubscriptionId` alongside CHB state is being billed by both providers, and applying more CHB state would bury that. It refuses, logs, and pages, and stays a 200 because no retry can fix it. Gated on `activeSubscriptionId` rather than `customerId` — a lingering customer id without a subscription is not double billing, and refusing on it would strand legitimate events.

Test: "refuses to apply CHB state onto an org with a live Stripe subscription" (no org update, `traceException` called).

Steffen911:
Rebased onto main and unstacked. This branch had been carrying six commits that were duplicates of work already merged — the client and service from #15462, and the project-event commits from #15461 — which is what made it conflict with main. Dropping them leaves the webhook commit alone: seven files, no dependency changes.

What to doubt about the rebase:

**The only conflicts were additive env blocks.** `web/src/env.mjs` and both `.env.*.example` files wanted the Auth0 credentials (from main) and the webhook signing secret (from here) in the same place; I kept both sides. Nothing else conflicted, which is worth a skim precisely because it is surprising — the webhook handler was written against the pre-Auth0 service and still compiles unchanged against the merged one. It turns out it never touched the client at all: its only import from the billing surface is `createDefaultSpendAlertsForPlan`, and it reaches Postgres directly for everything else.

**It no longer declares `CLICKHOUSE_BILLING_SERVICE_TOKEN`.** That was the shared-secret auth #15462 removed in favour of Auth0 M2M; this branch was still declaring it as a leftover of the duplicated commits. Only the signing secret is new here now.

Verified on the rebased branch: `Tasks: 8 successful, 8 total` (typecheck), `Tests 16 passed (16)` (`chbWebhookHandler.servertest`), and all eight CHB suites green with this and #15464 both merged onto main. No file in common with #15464, so order does not matter.

Steffen911:
@claude review

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,561 @@
...
+    );
+  }
+
+  // Backfill-emit LANGFUSE_PROJECT_CREATED for all existing projects: projects
+  // created before checkout would otherwise be invisible to CHB metering.
+  // Best-effort -- CHB's own backfill pipeline is the backstop.
+  const projects = await prisma.project.findMany({
+    where: { orgId: parsedOrg.id, deletedAt: null },
+    select: { id: true },
+  });
```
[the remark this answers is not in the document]
Steffen911: Confirmed and fixed in https://github.com/langfuse/langfuse/pull/15463/commits/401dd6dd1c.

The organization update in `persistAndPropagate` is the commit point: once it lands, this event's `occurredAt` is the stored `lastEventCreatedAt` and the ordering guard drops any retry of it as already applied. A 500 from anything after that point only asks CHB for a retry that can never reach the failed step, so post-commit steps must not fail the request.

Everything after the update is now best-effort: the project backfill (moved into `backfillProjectEvents`) and the API-key cache invalidation, which had the same shape — it runs a Prisma query too, and a miss is bounded by the 5-minute cache TTL. Both log and page instead of returning 500. Failures at or before the update still return 500, where the retry does help.

Tests: "keeps a failed project backfill out of the retry path" and "keeps a failed API-key cache invalidation out of the retry path" — both 200 with the dedupe claim kept and `traceException` called; both returned 500 before the change.

Steffen911:
@claude review

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,631 @@
...
+      user: { id: "clickhouse-webhook" },
+      orgId: parsedOrg.id,
+    },
+    orgId: parsedOrg.id,
+    resourceType: "organization",
+    resourceId: parsedOrg.id,
+    action: `BillingService.chb.${event.type}`,
+    before: parsedOrg.cloudConfig,
+    after: updatedCloudConfig,
+  });
```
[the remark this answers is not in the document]
Steffen911: Fixed in https://github.com/langfuse/langfuse/pull/15463/commits/c083c730ce1a414a022c00d83b70c495a7271c7d. `auditLog` is now awaited under the same post-commit guard as the cache invalidation: a failed audit insert logs and pages, and the request still returns 200.

Test: "keeps a failed audit log write out of the retry path". Writing it exposed a leak in the suite — `vi.clearAllMocks()` keeps `mockRejectedValue` implementations, so the invalidation rejection from the previous test was bleeding into the next one and made the new test pass before the fix. `beforeEach` now pins both mocks back to resolving.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+        break;
+      case "bundle.updated":
+        await handleBundleUpdated(parsedOrg, event);
+        break;
+      case "bundle.scheduled":
+        await handleBundleScheduled(parsedOrg, event);
+        break;
+      case "bundle.cancelled":
+        await handleBundleCancelled(parsedOrg, event);
+        break;
```
lightningspirit: Please, take a look to https://linear.app/clickhouse/issue/BIL-5791/billing-integration-langfuse-clickhouse-billing.


```suggestion
      case "attachedplan.created":
        await handleBundleCreated(parsedOrg, event);
        break;
      case "attachedplan.updated":
        await handleBundleUpdated(parsedOrg, event);
        break;
      case "attachedplan.scheduled":
        await handleBundleScheduled(parsedOrg, event);
        break;
      case "attachedplan.cancelled":
        await handleBundleCancelled(parsedOrg, event);
        break;
```

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+// stays permissive about unknown fields so CHB can extend the payload without
+// a deploy here.
+const ChbWebhookEventSchema = z.object({
+  eventId: z.string().min(1),
+  type: z.string(),
+  occurredAt: z.iso.datetime({ offset: true }),
+  data: z.object({
+    // ClickHouse Organization ID owning the bundle
+    organizationId: z.uuid(),
+    bundleId: z.string().nullish(),
```
lightningspirit: It's `id` not `bundleId`

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+    planCode: z.string().nullish(),
+    startDate: z.string().nullish(),
+    nextPaymentDate: z.string().nullish(),
```
lightningspirit: I don't see this property in the spec. Did we change it in the last few days?

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+    payment: z
+      .object({
+        status: z.string().nullish(),
+        provider: z
+          .object({
+            customerId: z.string().nullish(),
+          })
+          .nullish(),
```
lightningspirit, line 90: This property `nextPaymentDate` is actually `payment.dueDate` here in this object.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
Steffen911: Done in https://github.com/langfuse/langfuse/pull/15463/commits/c739eaa13aba1742d764d98a9bed74e4ff7f19eb: the switch, the handlers, the audit actions, and the tests now use `attachedplan.created|updated|scheduled|cancelled`.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
Steffen911: Done in https://github.com/langfuse/langfuse/pull/15463/commits/c739eaa13aba1742d764d98a9bed74e4ff7f19eb: the payload schema reads `data.id`. We keep storing it as `cloudConfig.clickhouse.bundleId` for now because `getBillingProvider`, the billing UI, and `chbBillingService` on main read that key; renaming the stored key to `attachedPlanId` goes with repointing the REST client to `/attachedplan`, which is a separate PR.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
Steffen911: Not a spec change — `nextPaymentDate` is the name of our stored field and it leaked into the wire schema. Removed from the payload schema in https://github.com/langfuse/langfuse/pull/15463/commits/c739eaa13aba1742d764d98a9bed74e4ff7f19eb; the stored value is now fed from `payment.dueDate`.

web/src/ee/features/billing/server/chb/chbWebhookHandler.ts
Steffen911: Done in https://github.com/langfuse/langfuse/pull/15463/commits/c739eaa13aba1742d764d98a9bed74e4ff7f19eb: `payment.dueDate` maps onto the stored `nextPaymentDate`. The payload schema now also carries `payment.provider.name` and the spec's `scheduled` union (`cancel` with `endDate`, `downgrade`/`upgrade` with `planCode` and `startDate`).

Steffen911:
@lightningspirit @agilliland one open question before this can merge: the envelope.

The spec puts `eventId`, `createdAt`, and `organizationId` at the top level next to `type` and `data`. The signed-webhook dispatcher that will actually deliver these rebuilds the body from the EventBridge event: `eventId` is the EventBridge event id, `type` the detail-type, `occurredAt` the EventBridge time, and `data` the raw `detail`. So the spec's top-level fields can only arrive inside `data`. If the events are published through the control plane's event-bus SDK they land one level deeper still, at `data.payload.*`, with `type` = `BILLING_ATTACHEDPLAN_CREATED`. Today that SDK schema only has the created event with `{organizationId, planCode, createdAt}`, and its producer has no caller yet.

This PR assumes `type` is `attachedplan.created|updated|scheduled|cancelled`, `organizationId` sits at `data.organizationId`, and `occurredAt` drives the ordering guard. Could you confirm:

1. Which emitter produces the Langfuse webhook: the event-bus SDK (then detail-type and `data` follow its envelope) or a dedicated emitter that puts the spec's event into `detail` with `detail-type` = `attachedplan.*`?
2. Where `organizationId` ends up on the wire, and whether the spec's `createdAt` is the dispatcher's `occurredAt`.
3. When the updated, scheduled, and cancelled events land on the CHB side, and whether the created event will carry the full `data` from the spec (`id`, `payment`, `startDate`) rather than the current three fields.

Whatever the answer, the change here is a few lines; I would rather not guess it.

agilliland:
Yeah, i'll let @lightningspirit confirm, but we should expect that structure of the webhook to be roughly ...

```js
// EventBride event structure
{
  eventId: 'some-unique-identifier',

  ... other eventbridge event attributes

  // Control Plane event structure in the 'data' attribute
  data: {
    id: 'unique-control-plane-event-identifier',
    source: 'control-plane-api',
    timestamp: 1788430679,

    // unique payload shape for the Control Plane event
    payload: {
      eventType: 'SOME_TYPE_OF_EVENT',

      ... whatever other details are relevant to the event
    }
  }
}
```

So I would expect the real control plane event data to come from `data.payload.*` in the received webhook

agilliland:
@Steffen911 I'm finalizing an e2e test for the signed webhooks in our dev environment and once i've confirmed everything is working then we'll be good to enable the webhooks in staging for you guys. Should be ready Monday for sure.

If you know the url you want to use for the staging webhook then drop it in here and i'll get a PR queued up.

Steffen911:
We intend to use [`https://staging.langfuse.com/api/billing/clickhouse-webhook`](<https://staging.langfuse.com/api/billing/clickhouse-webhook>) in our staging environment.

agilliland:
Perfect. So I've gone ahead and provisioned the webhooks now, they won't start delivering until you've gathered the secret and got your code setup on the Langfuse side.

You'll need to get the secret value to use for signing from here using SOPS `aws/iac/environments/non-prod/cp-staging-account/ap-southeast-1/signed_webhooks/secrets.yaml` (it's the same secret in all 3 regions)

Once you have the secret then add it to your infra as appropriate and use the signing code example to make sure you are verifying the signed request ... https://github.com/ClickHouse/data-plane-infrastructure/blob/main/aws/iac/modules/control_plane/signed_webhooks/README.md

You can see a working example of this in a TS app here which is what I used to validate this end-to-end ... https://github.com/ClickHouse/control-plane/blob/main/apps/control-plane-dummy-app/src/signedWebhook.ts

This is the final PR that we will use to enable the flow of the webhooks, so just let us know when the above is in place and we can green light the delivery ... https://github.com/ClickHouse/data-plane-infrastructure/pull/25494

PR #15466, opened 2026-07-27
feat(filters): refine event facet counts from the row query's filter state (LFE-14489)

nkabardin:
## TL;DR

The events-table filter facets now recompute their **counts against the currently-applied filter set**, not just the time window — you can no longer pick a facet value that then returns "No results". Frontend wiring for the already-merged backend (#15424). The core design decision: facet counts and table rows **share one filter source**, so they structurally cannot disagree.

Closes **LFE-14489**.

## Why

Facet counts were scoped only by the time range. With `environment = development` active, the Status facet still advertised project-wide counts (e.g. `ERROR 334`) — click it and the table comes back near-empty. That "the number lied" confusion is the friction tracked in LFE-10644.

## Key decisions

1. **One filter source — the counts refine from the exact `FilterState` that scopes the rows.** The events table previously kept a second, vestigial decoder of the `?filter=` URL param for the facet query (wrong table's column config → filters on events-only columns silently dropped; no embed scoping; a hardcoded type default the rows never applied). That decoder is deleted; the facet query now consumes the sidebar's `effectiveFilterState` — including the app-root default, the implicit environment policy, and the user/session/prompt embed scoping — verbatim.

2. **Sidebar filter hook split into core + presentation; no store.** `useSidebarFilterState` conflated route state (URL/session/memory decode, defaults, policy, writer) with option-aware presentation (facet actions, UIFilter list). The apparent cycle — "the sidebar hook needs the options; the options query needs the filter state" — was false: the state half never reads the options payload. It's now `useSidebarFilterStateCore` + `useSidebarFilterPresentation`, composed as before for every other table (public API unchanged), and composed *around* the facet-options query by the events table. Filter state stays URL-owned route state — a client store would fight the URL as source of truth.

3. **A facet never filters itself (pure query planner).** Refining a facet by its own condition would collapse it to the selected values and break the `none of` complement math. `lib/facet-query-plan.ts` plans the fetch: facets carrying their own active filter are pulled out of the shared bulk query and re-issued self-excluded; everything else is cross-refined in one scan. Column id/label canonicalization lives here too (embed-scope filters are label-keyed, e.g. `User ID`).

4. **The score catalog is never refined.** The score-name columns are time-bounded project metadata the backend deliberately doesn't narrow — and the search bar's score-type routing plus its AI score-name validation must keep seeing every score name, or a filtered-out score would be treated as unknown and its filter dropped.

5. **Explicit `startTimeFilter` vs `refiningFilter` hook inputs.** Previously one mixed filter array; splitting them means a time-window move re-keys every facet query while a filter edit re-keys only the refined ones — and it removed the render-identity workarounds the mixed array required.

## Result

Verified live in the browser (seeded project, 90-day window):
- Applying `environment = development` (286 events) recomputes the Status facet `DEFAULT 8K / ERROR 334 / WARNING 181 → 275 / 10 / 1` (sum = 286), while the Environment facet still lists **all four** environments.
- On a fresh landing, the facet query's filter is **byte-identical** to the row query's filter.
- On a user detail page, facet counts are scoped to that user's events — Status showed `22 / 2 / 1`, exactly the 25 rows in the table.

<details>
<summary>Architecture, mechanism & verification</summary>

**One-way flow** (per `frontend-large-feature-architecture`): route state → pure query plan → server options → presentation.

```
useSidebarFilterStateCore ──► planEventFacetQueries ──► events.filterOptions ──► useSidebarFilterPresentation
(decode once, events config)     (pure, tested)            (react-query)          (UIFilters with counts)
        │
        └── same effectiveFilterState ──► row query
```

**Planner contract** — `planEventFacetQueries({refiningFilter, eagerColumns, lazyColumns})` → `{bulk: {columns, filter}, perColumn: [{column, filter}]}`. Lazily-requested facets are per-column queries with their own react-query cache entries and self-exclude like any other facet. `useEventsFilterOptions` just executes the plan.

**Known, by-design behaviors** (not regressions): a non-participating filter (input/output/comment fields, position-in-trace) drops counts from all facets per the backend contract; a facet value with no rows under the *other* active filters drops out of that facet's list; a facet briefly skeletons when it first gains/loses its own filter (it moves between the bulk and its own query key).

**Checks** — `pnpm --filter web run typecheck` (0 errors), `pnpm --filter web run lint` (7/7 packages), clienttests across events/filters/search-bar/table `Tests 564 passed (564)` (incl. 10 planner tests + hook wiring tests), plus the live browser passes above.
</details>

### DIFF
--- a/web/src/components/session/index.tsx
+++ b/web/src/components/session/index.tsx
@@ -1019,7 +1019,7 @@ const LoadedSessionEventsPage: React.FC<{
 
   const { filterOptions, isFilterOptionsPending } = useEventsFilterOptions({
     projectId,
-    oldFilterState: timeFiltersForOptions,
+    startTimeFilter: timeFiltersForOptions,
   });
   const typedFilterOptions = filterOptions as EventFilterOptions;
 
--- a/web/src/features/events/components/EventsTable.tsx
+++ b/web/src/features/events/components/EventsTable.tsx
@@ -486,10 +486,8 @@ export default function ObservationsEventsTable({
     projectId,
   });
 
-  // Route-state half of the sidebar filters. Runs BEFORE the facet-options
-  // query so its effectiveFilterState — the same state that scopes the row
-  // query — can refine the facet counts (LFE-14489). The option-aware UI half
-  // (useSidebarFilterPresentation) runs after the options arrive.
+  // Route-state half of the sidebar filters, ahead of the facet-options query
+  // so the same state that scopes the rows can refine the counts (LFE-14489).
   const filterStateOptions: UseSidebarFilterStateOptions = useMemo(() => {
     const baseOptions = {
       implicitDefaultConfig: DEFAULT_SIDEBAR_IMPLICIT_ENVIRONMENT_CONFIG,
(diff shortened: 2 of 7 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

PR #15478, opened 2026-07-27
fix(redis): keep sentinel connections alive when BullMQ's blocking watchdog fires mid-resolution

### LINKED ISSUE #13880: bug: Redis Sentinel connections permanently broken after data node failover (missing sentinelRetryStrategy)
### Describe the bug

When a Redis data node restarts (e.g. during a failover), the `langfuse-worker` pod enters a broken state and stops processing all queues. The only recovery is a manual pod restart.

All queue workers (DataRetentionProcessingQueue, EvalExecutionQueue, etc.) emit sentinel unreachable errors simultaneously, then go completely silent — ioredis has given up retrying and the connection is permanently broken.

**Expected behavior:** the worker self-heals once sentinels finish their failover (typically a few seconds), without requiring a pod restart.

**Root cause:** `sentinelRetryStrategy` is never configured anywhere in Langfuse (`packages/shared/src/server/redis/redis.ts`, `createRedisSentinelInstance`). ioredis uses its default strategy, which gives up after a bounded number of attempts and leaves the `Redis` instance in a permanently broken state.

Example logs before the worker goes silent:

```
2026-05-27T10:05:15.059Z error  DataRetentionProcessingQueue error All sentinels are unreachable. Retrying from scratch after 10ms. Last error: Connection is closed.
Error: All sentinels are unreachable. Retrying from scratch after 10ms. Last error: Connection is closed.
    at connectToNext (.../ioredis/built/connectors/SentinelConnector/index.js:64:31)
    at connectToNext (.../ioredis/built/connectors/SentinelConnector/index.js:117:24)

2026-05-27T10:05:15.059Z error  Redis sentinel error All sentinels are unreachable. ...
2026-05-27T10:05:15.059Z error  EvalExecutionQueue shard 0 error All sentinels are unreachable. ...
```

### Steps to reproduce

1. Deploy Langfuse worker with Redis Sentinel (`REDIS_SENTINEL_ENABLED=true`)
2. Wait for the worker to be healthy and processing queues
3. Restart a Redis data node (`kubectl rollout restart statefulset/<redis-statefulset>` or delete a pod)
4. Observe worker logs — within seconds all queues emit `All sentinels are unreachable` errors
5. Errors eventually stop (ioredis gives up), worker is now silently broken and processes no jobs
6. Restart the worker pod → recovers immediately

Self-hosted

3.167.4 (Helm chart 1.5.26)

- ioredis: 5.10.1
- Redis setup: 2 Redis data nodes + 3 Sentinel nodes (Spotahome Redis Operator / RedisFailover CRD)
- `REDIS_SENTINEL_ENABLED=true`, `REDIS_SENTINEL_NODES=<single-service-endpoint>:26379`

### Additional information

**Affected code:** `packages/shared/src/server/redis/redis.ts`, function `createRedisSentinelInstance`

```ts
const instance = new Redis({
  sentinels,
  name: env.REDIS_SENTINEL_MASTER_NAME,
  // ...credentials...
  // ⚠️ no sentinelRetryStrategy — ioredis default gives up after ~10 attempts
  ...defaultRedisOptions,
  ...additionalOptions,  // has retryStrategy + reconnectOnError, but NOT sentinelRetryStrategy
  ...tlsOptions,
});
```

`redisQueueRetryOptions` (used by every queue worker) correctly sets `retryStrategy` and `reconnectOnError` for READONLY errors, but `sentinelRetryStrategy` — the separate callback ioredis uses when it exhausts all sentinel addresses during a reconnect round — is never set anywhere.

**Proposed one-line fix** inside `createRedisSentinelInstance` (before `...additionalOptions` so callers can still override):

```ts
sentinelRetryStrategy: (retries: number) => Math.min(retries * 200, 10_000),
```

This backs off up to 10 s between retry rounds but never stops retrying, allowing the worker to self-heal once the sentinel failover completes.

No

Steffen911:
BullMQ's Worker.waitForJob arms a watchdog that calls bclient.disconnect(true) when a blocking command gets no response in time. In sentinel mode the client spends whole seconds in ioredis status "connecting" with no live socket while it resolves the master through the sentinels (initial connect and every reconnect, e.g. during a failover). A disconnect landing in that window can never emit a socket "close" event, so retryStrategy is never consulted: the in-flight connect() rejects and ioredis parks the client in the terminal "end" state with all further errors silenced. Worker pods then stay detached from Redis until restarted.

Reproduced with a local sentinel topology (master + replica + 3 sentinels, ioredis 5.10.1 + bullmq 5.76.3): a \~45s total outage reliably killed every worker blocking connection \~6s in, and they never recovered after Redis returned. With the guard, the same scenario recovers fully.

Suppress exactly that call shape - disconnect(reconnect=true) while the client is resolving the master - and propagate the guard through duplicate(), which is how BullMQ creates its blocking clients. During resolution no command is on the wire to unstick, and the pending resolution already retries on its own, so the suppressed call loses nothing. All other disconnect paths (shutdown, quit, BullMQ's own disconnect-then-reconnect flows) pass through unchanged.

Fixes langfuse/langfuse#13880

### DIFF
--- a/packages/shared/src/server/redis/redis.test.ts
+++ b/packages/shared/src/server/redis/redis.test.ts
@@ -1,15 +1,19 @@
-import { describe, expect, it, vi } from "vitest";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
 import type { Redis } from "ioredis";
 
-vi.mock("../../env", () => ({
-  env: {
-    REDIS_CLUSTER_ENABLED: "false",
-    REDIS_ENABLE_AUTO_PIPELINING: "false",
-    REDIS_SENTINEL_ENABLED: "false",
-  },
+const mockEnv = vi.hoisted(() => ({
+  REDIS_CLUSTER_ENABLED: "false",
+  REDIS_ENABLE_AUTO_PIPELINING: "false",
+  REDIS_SENTINEL_ENABLED: "false",
+  REDIS_SENTINEL_MASTER_NAME: undefined as string | undefined,
+  REDIS_SENTINEL_NODES: undefined as string | undefined,
 }));
 
-import { scanKeys } from "./redis";
+vi.mock("../../env", () => ({
+  env: mockEnv,
+}));
+
+import { createNewRedisInstance, scanKeys } from "./redis";
 
 type ScanCall = [string, "MATCH", string, "COUNT", number];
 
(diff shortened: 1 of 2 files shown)


Steffen911:
Replaced with https://github.com/langfuse/langfuse/pull/15512

PR #15480, opened 2026-07-27
fix(filters): use legacy environments in dual mode

hassiebp:
## Summary

- make dual-write deployments source environment options from the legacy tracing tables used by the v3 filters
- keep events-only deployments sourced from `events_core`
- add focused coverage for both routing modes

## Root cause

The shared environment-options lookup switched dual mode to `events_core` in #14959. The v3 traces/observations facets still consume that lookup while filtering the legacy tables with case-sensitive values, so an events value such as `PROD` could not match the corresponding legacy `prod` rows.

## Impacted package

- `@langfuse/shared`

## Verification

- `pnpm exec dotenv -e .env.dev.example -e .env.test.example -- pnpm --filter @langfuse/shared run test environments.test.ts` — 2 passed
- `pnpm run lint` — Tasks: 7 successful, 7 total
- `pnpm exec dotenv -e .env.dev.example -e .env.test.example -- pnpm --filter @langfuse/shared run typecheck` — passed
- `pnpm exec dotenv -e .env.dev.example -e .env.test.example -- pnpm --filter web run typecheck` — passed
- `pnpm exec dotenv -e .env.dev.example -e .env.test.example -- pnpm --filter worker run typecheck` — passed

Linear: [LFE-14403](https://linear.app/clickhouse/issue/LFE-14403/environment-filter-shows-prod-instead-of-prod)

### DIFF
--- a/packages/shared/src/server/repositories/environments.test.ts
+++ b/packages/shared/src/server/repositories/environments.test.ts
@@ -1,0 +1,57 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { env } from "../../env";
+import { getEnvironmentsForProject } from "./environments";
+import { queryClickhouse } from "./clickhouse";
+
+vi.mock("./clickhouse", () => ({
+  queryClickhouse: vi.fn(),
+}));
+
+const originalWriteMode = env.LANGFUSE_MIGRATION_V4_WRITE_MODE;
+
+afterEach(() => {
+  vi.mocked(queryClickhouse).mockReset();
+  (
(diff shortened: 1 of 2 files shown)


Steffen911:
@hassiebp The idea was to have the more performant accesses in the dual mode. We can also make it dependent on the v4 opt in behaviour of the client which would probably be the cleanest option if we're able to pass that along.

Steffen911:
This hides the error but doesn't address the actual issue. Will create a separate PR for this.

Steffen911:
Replaced with https://github.com/langfuse/langfuse/pull/15499

PR #15486, opened 2026-07-27
feat(events): outlier chart strip above the trace table (LFE-14451)

nkabardin:
**Ticket:** [LFE-14451](https://linear.app/clickhouse/issue/LFE-14451)

## TL;DR
A compact, Firefox-devtools-style **bar strip between the search bar and the trace/events table** that surfaces **outliers over time** — every bar is the *worst single event* (max cost / latency / tokens) in its time bucket. **Click a spike → the table narrows to that window; browser Back returns.** Always on: expanded by default where there's space (up to **three charts** at ≥400px each), closed it becomes a "Pulse" toolbar button next to Columns (closed is the mobile default). This is the v4 substitute for sortable per-trace aggregated cost/tokens, which is too expensive to compute in ClickHouse.

## Why
Finding the expensive or slow outlier in the events table used to mean a per-trace aggregate sort that ClickHouse can't afford. A "worst event per time slice" strip answers the same question in one cheap aggregate query — and turns it into navigation: see the spike, click it, land on the rows that made it.

## What
- **Always-on band at table width** just under the search bar (the facet sidebar keeps its full height). Closing is an X on the strip; closed, it becomes a "Pulse" toolbar button next to Columns (closed is the mobile default). Open/closed and all chart picks persist per user.
- **One full-width chart (Cost default) with a mode dropdown — Cost / Latency / Tokens / Split.** Split shows three charts when they fit (~400px each), two (Cost + Latency) on smaller widths, and leaves the menu below that; a stored Split preference degrades to Cost without being overwritten. Every chart states its aggregate: cost switches **max / total**, latency **max / p95 / avg** (all series in the same single scan), tokens is max. 40px tall, square bars with 1px gaps resting on a baseline, sqrt height scale (10–40× outliers don't flatten the base load — picked off real renders in Storybook), values on hover only (empty buckets read an honest zero), cursor-tracking tooltip, sparse Firefox-style gridline ticks. Charts span the full time domain — no trailing whitespace.
- **Space-adaptive binning:** the strip measures its width and picks the finest `dashboard.executeQuery` granularity preset that fits (~5px/bar); fractional bar slots always fill the width exactly. **Zero backend change** — one existing-endpoint query returns count + all three metric maxes per bucket, so switching metrics or the multi-chart layout never refetches.
- **Click/drag/tap-to-navigate:** a bar click (or mobile tap) drills into that bucket; a **Grafana-style horizontal drag — mouse or touch — selects a bucket-snapped span**, highlighted in sync across all charts, and applies it on release (the strip-level slice of LFE-14532; the app-wide select-on-chart for recharts surfaces follows separately). Every navigation writes an absolute `?dateRange=` with a history push — table, facets, and strip all re-derive; **Back restores the outer view natively**; the transient zoom is deliberately *not* persisted as the project's default range.
- Filters forward exactly like the in-view chart (#15187, `toChartFilters`); unsupported ones (incl. `is root observation`) are silently ignored, matching the analytics surfaces. Honest empties: buckets with events but no metric data render subtle activity ticks, never fake-zero bars; empty buckets are not clickable.
- Hidden on prompt-version-scoped tables (that page-scope filter can't forward and nothing would disclose the mismatch) and in full chart-view mode.
- Ships an **`outlier-traffic` seeder scenario**: diurnal traffic over N days with deterministic cost/latency/token spikes and hour-long latency incidents, mirrored into v4 events — the strip's test bed.

## Result
Open the traces view, scan the skyline, click the spike, read the rows, hit Back. Iterated live with product review: verified end-to-end in the browser (desktop 3-up/2-up/1-up, collapse/expand, mobile collapsed-bar default, tooltip, drill-in URL, Back, no default-range leakage) over 90 days of seeded outlier traffic.

<details>
<summary>Engineering detail (architecture, review fixes, follow-ups, verification)</summary>

**Preparer/visualiser split** per the charts manifesto: `lib/binning.ts` owns every decision (granularity ladder, epoch densification matching ClickHouse's `toStartOfInterval` grid, executeQuery row mapping, formatting) as pure unit-tested functions; `OutlierBarStrip` renders rects and decides nothing. Storybook stories are the design surface and remain as showcases.

**Adversarial-review + live-testing fixes baked in:** WITH FILL filler rows are dropped in the row mapper (ClickHouse fills the non-nullable tokens measure with `0`, which would have drawn a phantom clickable baseline across empty buckets — verified against live ClickHouse); `placeholderData` keeps previous bins on auto-refresh so the persistent band never flashes; clicks on truly empty buckets are ignored; a stale hover index after a data-shape change no longer crashes (`dense[i]` returning `undefined` slipped past a `!== null` check); the width-measuring wrapper stays mounted across collapse/expand (`useElementSize` observes only on mount, so unmounting it froze the measurement).

**Known limits / follow-ups:**
- After a drill-in below ~30 minutes the finest preset (`minute`) yields few, fat bars — finer granularity tokens (`10s`, `30s`, also `6h`/`12h`) are a proposed follow-up backend ticket (a prototype of a client-driven-bin endpoint exists in closed PR #15472 for reference).
- The in-view chart (#15187) shares the prompt-version-scope blind spot this PR gates the strip against; flagged as a pre-existing follow-up.
- Next slices: app-wide select-on-chart for recharts surfaces (LFE-14532), vertical resize, mobile swipe + touch polish.
- No analytics instrumentation yet (matches the chart-view toggle); to be added with the usage-analytics pass.

**Verification.** Live browser flows as above. Unit: `binning.clienttest.ts` `Tests  7 passed (7)` (granularity picking, epoch densification, fill-row drop); stories as tests `Tests  8 passed (8)`; web + shared typechecks clean; full lint green. Net server diff: zero (an earlier bespoke-endpoint commit was reverted in-branch after the design pivot); the seeder scenario is dev-only tooling.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15487, opened 2026-07-27
feat(agent): Silent mode for tools

bezbac:
🟡 **Live preview: [pr-15487.preview.langfuse.com](https://pr-15487.preview.langfuse.com)**

Adds optional silent execution for in-app-agent MCP tools.
- Extends eligible MCP tool schemas with a `silent` argument when a sandbox is available.
- Preserves full silent results in persisted events and sandbox tool-call files while redacting streamed, displayed, replayed, and model-visible output.
- Makes current-run tool-call results available to subsequent sandbox operations in both web and worker execution paths.
- Updates manual tool approval handling to distinguish persisted results from model-facing results and adds coverage for streaming, replay, persistence, and sandbox behavior.
</details>

The PR appears safe to merge, with silent results consistently preserved for sandbox use and redacted from public and model-facing output.

The web and worker paths both track current-run tool events, persistence retains the raw result, and stream, display, replay, and approval paths consistently substitute the intended public message.
</details>

```mermaid
sequenceDiagram
  participant Model
  participant Agent
  participant MCP
  participant Events as Event persistence
  participant Sandbox
  participant Client
  Model->>Agent: "Call MCP tool with silent=true"
  Agent->>MCP: Execute without synthetic silent argument
  MCP-->>Agent: Full result
  Agent->>Events: Persist wrapped full result
  Agent-->>Model: Output saved to /workspace/tool_calls
  Events-->>Client: Redacted tool result
  Sandbox->>Events: Refresh current and prior tool calls
  Events-->>Sandbox: File containing full unwrapped result
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/langfuse/langfuse/commit/99fc45dda2209241988a1fff2ce9f32fd02e0094) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48672222)</sub>

### DIFF
--- a/packages/shared/src/in-app-agent/constants.ts
+++ b/packages/shared/src/in-app-agent/constants.ts
@@ -3,6 +3,10 @@ export const IN_APP_AGENT_REDIRECT_TOOL_NAME = "langfuse_proposeRedirect";
 
 export const IN_APP_AGENT_TOOL_REJECTION_ERROR_CODE = "tool_call_rejected";
 
+export const IN_APP_AGENT_SILENT_MCP_OUTPUT_TYPE = "silent-mcp-output";
+export const IN_APP_AGENT_SILENT_MCP_OUTPUT_MESSAGE =
+  "Output saved to /workspace/tool_calls";
+
 export const IN_APP_AGENT_MCP_USER_AGENT = "langfuse-in-app-agent";
 
 // Header used only by Langfuse's server-side in-app agent when it calls the
--- a/packages/shared/src/in-app-agent/server/agent.ts
+++ b/packages/shared/src/in-app-agent/server/agent.ts
@@ -23,7 +23,9 @@ import {
   createSandboxTools,
   createRedirectActionTool,
   filterInAppAgentAvailableLangfuseMcpTools,
+  getPublicInAppAgentMcpToolResultContent,
   type InAppAgentUserAccess,
+  withOptionalSilentMcpOutput,
   withInAppAgentToolApproval,
 } from "./tools";
 import { LANGFUSE_IN_APP_AGENT_SKILLS } from "./skills";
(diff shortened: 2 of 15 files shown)


bezbac:
@claude review

LysanderKie:
Had a closer look, mostly around how this interacts with my background execution. Feature overall looks right and I'm glad it covers the worker path too as far as I see. Few comments after taking a deeper look and aligning with my mental model of background execution and the path forward:

- `getPublicInAppAgentMcpToolResultContent` is called from `agent.ts` (foreground SSE), `watch.ts#toPublicEvent` (background SSE), `getConversationMessagesFor DisplayFromEvents`, and `sanitizeConversationMessagesForReplay`. 
From what I can see there are two other readers still don't redact: `getConversationMessages` is used by
`submitFeedback` and by `maybeInferAndPersistConversationTitle` (however those are safe as they only operate on user/assistant messages. 
I think this collapses to two flows:
-> messages: Can we put the redact logic in `getMessagesFromPersistedEvents`. It's the core funnel, so display, replay, title and feedback all inherit it.
-> events: `watch.ts#toPublicEvent` already is the "make this client-safe" function. Can `agent.ts` call it instead of inlining a second copy. Best case foreground and background emit identical streams, two implementations of the same redaction is something I would like to avoid (also not super critical as I hopefully foreground will be gone next week)

- `currentRunSandboxToolCallEvents` buffers the whole token stream: The push is unconditional (and we push the DELTA chunks one-by-one from what I know, so the array might have a significant size)  in an array named after tool calls. However, `getSandboxToolCallFiles` only reads `TOOL_CALL_START`/`ARGS`/`RESULT`, so we could already filter on push-side (and I think we scan this buffer on every read). 
Worth noting the read side too: `service.ts` calls `getToolCallFiles()` inside `ensureSession()`, which runs on every sandbox read/write/edit/bash. So today each sandbox op rescans the full array, rebuilds the set.
Less important as temporary: Also, this block is now duplicated in `executeInAppAgentRun.ts`. Can we centralize this and then share the logic?

- Sandbox filenames: Can we key on `toolCallId` instead of timestamps (as I guess every tool call has one return file right?), as this would also let us return the exact path in the redaction message instead of just the directory, which right now forces an additional ls call on read?

- Storage question, not blocking but would be great to align: We have no S3 anywhere in the in-app-agent code, the JSONB event row is the only store. That was fine while every persisted result was also a result the model had consumed, so size was "implicitly" capped by the context window. Silent mode removes that cap by design: the premise is "too big for context" and the we "persist in full".
The only consumer of the raw bytes is `getSandboxToolCallFiles`, so we're storing something in PG which main purpose is to become a file to be read in the sandbox. Two things follow:
-> `getConversationEvents` selects the full `event` JSONB for every event, unbounded, and it sits on the hydration path that runs on every drawer open and refresh. We fetch, then redact.
-> There's no "data retention. Conversation delete is a soft delete and I can't find any `inAppAgentEvent.deleteMany` or cleanup job, so these rows live until the project cascade (at least at the moment). 
Do you plan to add S3 support for the sandbox?

PR #15494, opened 2026-07-27
ci(preview): share one GitHub environment for PR preview deployments

wochinge:
## Summary

- PR previews recorded a native GitHub deployment under a per-PR `pr-<n>` environment. GitHub environments are durable config objects, so that grew an unbounded list of them, one per PR ever opened. Deployments now land in a single shared `PR Preview` environment, keyed per PR by the deployment `task` (`preview-pr-<n>`) so cleanup can still scope a query to one PR while the immutable SHA stays the deployment ref.
- Cleanup now **deletes** deployment records instead of only marking them inactive — with a shared environment, retired-but-present records would otherwise accumulate on the Deployments page forever. `auto_inactive` is off and `transient_environment` is false, since other PRs share the environment and must stay active.
- Adds `labeled` / `unlabeled` triggers so manually re-adding the `preview` label after a teardown recreates the deleted deployment record. The auto-labeler still can't trigger a build (GITHUB_TOKEN anti-recursion), so the `opened` event remains the normal entry point.
- Scopes both workflows' concurrency groups. `preview-build` and `preview-deactivate` deliberately share a group so a close cancels an in-flight build, but a group is claimed at **queue** time, before job `if`s are evaluated — so a run triggered by an unrelated label would have joined the group, cancelled a live build, and then skipped every job. Only runs that will do work now join `preview-build-<n>`; the rest get a throwaway per-run group.

## Major Decisions

- **One shared environment over per-PR environments.** GitHub can't delete environments with `GITHUB_TOKEN` (that needs repo administration), so per-PR environments could only ever accumulate. The tradeoff: the repo Environments page now shows a single `PR Preview` whose "latest deployment" is whichever PR deployed most recently, so per-PR discoverability there is lost. The in-PR **View deployment** button — the actual entry point — is unaffected.
- **Delete records rather than deactivate.** Inactive records are the conventional way to render a retired environment, but that convention assumes the environment is per-preview. Sharing one environment makes deletion the only way to stop advertising dead URLs.
- **Conditional concurrency group over dropping the shared group.** Keeping the cross-workflow cancel is worth the expression complexity; the alternative (removing the shared group) would leave a build racing a close.

## Review Focus

- **The concurrency group strings must stay byte-identical.** `preview-build.yml` and `preview-deactivate.yml` both resolve their true branch to `format('preview-build-{0}', github.event.pull_request.number)`. If those two literals ever drift, cancel-on-close silently stops working — no error, just a build that outlives the PR. Worth confirming the pair by eye.
- **The `env:` / task literals are a three-file contract.** `"PR Preview"` and `preview-pr-<n>` appear in all three preview workflows. A drift makes `listDeployments` match nothing, so cleanup logs success while dead records pile up — silent in both directions. Flagged with a comment at each site; there's no cross-file constant mechanism in Actions to enforce it.
- **`labeled` widens who can start a build.** Adding a label needs only *triage*, not write access. The built ref is still write-authored and the OIDC trust scope is unchanged, so there's no code-injection delta — but the trigger is no longer strictly write-gated.

### DIFF
--- a/.agents/skills/langfuse-previews/SKILL.md
+++ b/.agents/skills/langfuse-previews/SKILL.md
@@ -36,9 +36,9 @@ Pushing updates it; closing the PR tears it down.
 ## Using a preview
 
 - **Spin up** — open a same-repo PR. It's auto-labeled and builds (~5 min); a
-  bot comment then posts the preview URL and login, and a GitHub deployment
-  (environment `pr-<n>`) gives the PR a **View deployment** button. No manual
-  step, no label to add.
+  bot comment then posts the preview URL and login, and a deployment in the
+  shared GitHub `PR Preview` environment gives the PR a **View deployment**
+  button. No manual step, no label to add.
 - **Log in** — use the credentials in the **bot's PR comment** (the source of
   truth). The demo project's shared seed identity is `demo@langfuse.com` /
   `password`, with API keys `pk-lf-1234567890` / `sk-lf-1234567890` — shared and
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -42,7 +42,8 @@ concurrency:
   # building runs join `preview-build-<n>`, which preview-deactivate.yml matches
   # verbatim to cancel on close; the rest get a throwaway per-run group.
   group: >-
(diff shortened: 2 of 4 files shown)


wochinge:
@claude review

PR #15499, opened 2026-07-27
fix(otel): normalize environment on the direct events_full write path

Steffen911:
The direct OTel events write (processToEvent -> createEventRecord -> events_full) passed the raw environment attribute through, while the legacy path normalized it via the public ingestion schema. A span with deployment.environment = "PROD" was stored as "PROD" in events_core but "prod" in the legacy tracing tables, so environment facets and filters diverged between v3 and v4 views (LFE-14403).

Normalize in OtelIngestionProcessor.extractEnvironment with the same public/internal split as createIngestionEventSchema: public values are lowercased, charset/length-validated, and lose the reserved "langfuse" prefix; internal processors keep the "langfuse-*" namespace so internal traces stay excluded from user views and the eval-recursion guard.

The public prefix strip is now repeated instead of single-pass, keeping normalization idempotent (values pass through the schema twice on the OTel path) and closing the reserved-namespace bypass where "langfuselangfuse-x" survived as "langfuse-x".

Ref LFE-14403


Steffen911:
@claude review

PR #15513, opened 2026-07-28
fix(otel): map foreign observation levels onto the Langfuse enum

Steffen911:
## Problem

The OTel ingestion path handles `langfuse.observation.level` inconsistently between its two write targets (LFE-14547, sibling of LFE-14403):

- **Events path:** `processToEvent` passes the raw attribute into `events_full.level` (unconstrained `LowCardinality(String)`), so values like `INFO`, `warning`, or `SUCCESS` land verbatim and break level filters (~484k `INFO` rows on prod-eu, ~21k `WARN` on prod-us).
- **Legacy path:** observations are re-parsed against the strict `ObservationLevel` enum; an invalid level fails `safeParse` and the whole observation is **silently dropped** from the legacy tables. In dual mode the same span exists in v4 events views but is missing from v3.

## Fix

Translate common OTel/logging vocabularies at the extraction seam in `OtelIngestionProcessor` (both the events and legacy outputs):

- `WARN`, `WARNING` → `WARNING`
- `ERROR`, `FATAL`, `CRITICAL` → `ERROR`
- `DEBUG`, `TRACE`, `VERBOSE` → `DEBUG`
- `DEFAULT`, `INFO`, `LOG`, `OK`, `SUCCESS`, `NOTICE` → `DEFAULT`
- unknown → no value, so the existing status-derived fallback applies (`status.code === 2` → `ERROR`)

The classic `/api/public/ingestion` contract keeps its strict enum (per-event error back to the SDK); only the OTel adapter translates. Once the processor emits canonical values, the legacy enum parse passes and the observation dropping stops.

Historical rows are normalized via a Cloud-only data patch tracked in LFE-14547 (self-hosted gets the forward fix only).

## Impacted packages

- `packages/shared` (OTel ingestion processor)
- `web` (test only)

## Verification

- `pnpm --filter web run test src/__tests__/server/api/otel/otelMapping.servertest.ts` — `Tests  222 passed | 1 skipped (223)` (new test written failing-first: `expected 'warning' to be 'WARNING'`)
- `pnpm --filter @langfuse/shared run test src/server/otel` — `Tests  42 passed (42)`
- `pnpm --filter worker run test src/queues/__tests__` — `1 failed | 111 passed`; the failure is a pre-existing timezone-sensitive `experiment_item_version` assertion that fails identically on unmodified `main`
- `pnpm tc` — `Tasks: 7 successful, 7 total`

Ref LFE-14547


Steffen911:
@claude review

PR #15517, opened 2026-07-28
feat(traces): show approximate row count in the traces table (LFE-11057)

nkabardin:
## TL;DR
The v4 traces table now shows an **approximate match count** — `Total ≈ X` — next to the pagination controls, so you can see roughly how many traces your current filters + time range return. It's a cheap HyperLogLog estimate fetched asynchronously; the table never waits on it.

## Why
A user asked to see how many traces match their query. v4 deliberately dropped the *precise* count from the traces table: a distinct-trace count over the events table (`uniqExact`) is expensive and had to go. But "how big is my result set?" is a common, reasonable question — so the plan (agreed with the backend folks) was to bring back an **approximate** count that's cheap enough to run on every load.

## What
- A new lightweight tRPC query (`events.approxCount`) runs a single ClickHouse `uniq(trace_id)` (HyperLogLog) over the **same filters, time range, and search** the row query uses, so the number matches what the table shows.
- The frontend fetches it **async and non-blocking** (`keepPreviousData`, gated exactly like the row query) — the table render never waits on it. While the first value loads, a subtle spinner shows.
- Rendered as `Total ≈ X` at the bottom-right, just left of the pagination controls, in muted styling, with a tooltip explaining it's an HLL estimate. The `≈` marker makes clear it's approximate, never exact.

## Result
Bottom-right of the traces table: `Total ≈ 402`. Verified live against seeded demo data — the endpoint returns 24,466 (no filter), 354 (level=ERROR), and 24,255 (type=GENERATION), each matching the ground-truth ClickHouse `uniq(trace_id)` exactly, and the footer scopes to the active time range (402 for the default window vs 24,466 all-time).

<details>
<summary>Mechanism, caveats & verification</summary>

**Backend**
- `event-query-builder.ts`: new `uniqueTraces` field set = just `uniq(e.trace_id)` — a single approximate aggregate with **no** precise `count(*)`/`uniqExact` over the scan.
- `repositories/events.ts`: new `unique-trace-count` select branch + `getApproxUniqueTraceCountFromEventsTable()` helper, reusing the existing `buildEventsObservationRowSelection` filter→WHERE planner (same translation the row + facet-count queries use).
- `eventsService.ts` / `eventsRouter.ts`: `getApproxEventTraceCount()` + an `events.approxCount` procedure that mirrors `countAll` (applies comment filters, normalizes orderBy, instrumented).

**Frontend**
- `useEventsTableData.ts`: a separate `events.approxCount` query, enabled whenever the table rows are (disabled in chart mode), with `keepPreviousData` so filter tweaks don't flash back to a loading state.
- `data-table.tsx` / `data-table-pagination.tsx`: optional `approxTotalCount` / `isApproxTotalCountLoading` pagination props; when present, render `Total ≈ X` near the controls. Purely additive — other tables that don't pass them are unchanged.

**Caveats**
- Approximate by design: HyperLogLog, typically within a few percent of the true count. It counts distinct **traces**, which suits the tracing view (the default view is one root observation per trace).
- Cheap: one `uniq()` per load; no precise count, no `uniqExact`, no `FINAL`.

**Verification**
- Playwright (Chrome) against the running dev server on seeded data: `Total ≈ 402` renders bottom-right; endpoint values match ClickHouse `uniq(trace_id)` exactly across three filter sets (all / ERROR / GENERATION), proving filter-awareness; time-range scoping confirmed (402 windowed vs 24,466 all-time).
- Added a unit test asserting the `uniqueTraces` field set emits `uniq(e.trace_id)` and **not** `count(*)` / `uniqExact`.
- `typecheck` + `lint` clean; full pre-commit lint across all packages green.
</details>


nkabardin:
@claude review

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.test.ts
```diff
@@ -21,6 +21,21 @@ describe("EventsAggregationQueryBuilder", () => {
...
+  it("selects a single approximate uniq(trace_id) with no precise count()", () => {
+    const { query } = new EventsQueryBuilder({ projectId: "test-project" })
+      .selectFieldSet("uniqueTraces")
+      .buildWithParams();
+
+    // HLL estimate over trace_id — the "Total ≈ X" footer count.
+    expect(query).toContain('uniq(e.trace_id) as "unique_trace_count"');
+    // Deliberately cheap: no precise count() over the scan (that would be the
+    // expensive count the v4 traces table dropped).
+    expect(query).not.toContain("count(*)");
```
sumerman: these two asserts are uselsess

packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts
```diff
@@ -208,6 +208,10 @@ const FIELD_SETS = {
   // Aggregates
   count: ["count"],
   countWithUniqueTraces: ["count", "uniqueTraceCount"],
+  // Approximate distinct-trace count only (HLL) — a single cheap `uniq()` with
+  // no precise `count()` over the scan. Powers the traces-table footer
+  // "Total ≈ X".
+  uniqueTraces: ["uniqueTraceCount"],
```
sumerman: I don't think HLL over trace id is what you really want. Whenever you table is showing more than one observation per trace it is not going to be even approximately correct. Perhaps you want uniq over span_id?

sumerman (dismissed):
I belieave counting can and should piggyback on the filter-options query instead of introducing a brand new scan

packages/shared/src/server/queries/clickhouse-sql/event-filter-options.test.ts
```diff
@@ -0,0 +1,59 @@
...
+} from "./event-filter-options";
+
+const base = {
+  projectId: "test-project",
+  filter: [] as FilterState,
+  columns: ["level", "type"] as const,
+  limit: 1000,
+};
+
+describe("buildEventsFilterOptionsForColumnsQuery — approximate total count", () => {
```
sumerman: please prompt it to extend existing API tests instead of adding random query shape tests that don't meaningfully assert anything.

packages/shared/src/server/queries/clickhouse-sql/event-filter-options.ts
```diff
@@ -384,14 +393,19 @@ export const buildEventsFilterOptionsForColumnsQuery = (params: {
   const { query: aggregatedOptionsQuery, params: aggregatedOptionsParams } =
     aggregatedOptionsBuilder.buildWithParams();
 
+  // Approx total rides one extra sentinel row so the result shape stays {column, value, count}.
+  const approxTotalCountRow = includeApproxTotal
+    ? `,\n      [tuple('${EVENTS_APPROX_TOTAL_COUNT_MARKER}', '', toUInt64(approx_total_count), toInt64(0))]`
```
sumerman, line 398: I would have preferred if an actual tuple defintion was a constant defined above together with others. 

While we are at it would be great to explain tuple format: `option-name, option-value, value occurrence count, sorting order`. My bad I didn't do it initially.

sumerman (commented):
A couple more pieces of feedback. These aren't blocking, but I really really really would like a proper test before it gets merged.

packages/shared/src/server/queries/clickhouse-sql/event-filter-options.test.ts
nimarb: @nkabardin please follow up on this on monday, thank you!

PR #15540, opened 2026-07-28
feat(events): Follow-up UX adjustments to the Pulse outlier strip 

nmtrang29:
## Summary

UX adjustments to the Pulse outlier strip above the trace table (#15486):

**Always visible**
- The X close button, the toolbar "Pulse" reopen button, and the open/closed localStorage state are removed — the strip simply always shows (also on mobile, which previously defaulted to closed). The toolbar's `preColumnsSlot` prop existed only for the reopen button and is removed with it.
- The associated `pulse:closed` / `pulse:reopened` analytics events drop with the buttons; `pulse:drill_in`, `pulse:mode_switch`, `pulse:aggregation_switch`, and `pulse:preview_pinned` (from #15535) are kept with payloads adapted to the simplified shape.

**Simplified metrics**
- **Cost** plots `sum` only (was max/sum); single-option metrics render no aggregation label.
- **Latency** offers **p95 (default)** and **p50** — the old `avg` was replaced by a *true* `p50` executeQuery aggregation, not renamed.
- **Tokens** metric and the **Split** view are removed; the mode dropdown is just Cost / Latency. Stored legacy preferences (`split`, `tokens`, `max`, `avg`) degrade safely via the zod settings schema + registry fallback.

**Brush discoverability & visual polish**
- Crosshair cursor over the plot (the Grafana/Datadog "this surface supports range selection" affordance).
- Drag-selection band: stronger fill (0.12 → 0.18) plus crisp 1px edge lines — the flat faint fill was easy to miss over sparse bars.
- Sans-serif for the title/aggregation dropdowns and tooltips (one-off ahead of a design-system pass); triggers 11px → 13px; separator dot dropped; more headroom between label row and plot (mt-1 → mt-2).
- Query error state reads "No Data".

## Testing

- `binning.clienttest.ts`: `Tests  13 passed (13)` — registry derivation, WITH FILL handling, aggregation-column selection, and legacy-aggregation fallback updated for the new registry
- `pnpm run lint`: `Tasks: 7 successful, 7 total` (pre-commit on every commit)
- Browser-reviewed on local dev against seeded data (90-day window): strip render, mode/agg dropdowns, drag-selection band, drill-in, crosshair
- One review round applied; the unrelated dev-environment turbopack.root fix moved to #15541 per review feedback. No other findings (tenant isolation unchanged, no CH/PG/API/seeder surface)

## Screenshots

1280x720, seeded demo data, 90-day window. Before = main (`187686a1e`), after = this branch.

| | Before | After |
|---|---|---|
| **Strip** (mono "Cost · max" + X close → sans "Cost", no X) | strip before | strip after |
| **Drag-selection band** (0.12 flat fill → 0.18 fill + 1px edges) | drag band before | drag band after |
| **Tooltip** (mono → sans; same bucket also shows cost max → sum: $0.003 → $23.77) | tooltip before | tooltip after |

_Asset branch `pr-15540-assets` — delete after merge._

### DIFF
--- a/web/next.config.mjs
+++ b/web/next.config.mjs
@@ -3,7 +3,6 @@
  * for Docker builds.
  */
 await import("./src/env.mjs");
-import { fileURLToPath } from "node:url";
 import { withSentryConfig } from "@sentry/nextjs";
 import { env } from "./src/env.mjs";
 
@@ -104,11 +103,6 @@ const nextConfig = {
     },
   },
   turbopack: {
-    // Pin the workspace root: in a nested git worktree Next.js infers the
-    // OUTER checkout's root (outermost lockfile wins), which breaks the
-    // relative resolveAlias below. fileURLToPath, not URL.pathname — the
-    // latter percent-encodes spaces and breaks on Windows drive letters.
-    root: fileURLToPath(new URL("..", import.meta.url)),
     resolveAlias: {
       "@langfuse/shared": "./packages/shared/src",
     },
--- a/web/next.config.mjs
+++ b/web/next.config.mjs
@@ -3,6 +3,7 @@
  * for Docker builds.
  */
 await import("./src/env.mjs");
+import { fileURLToPath } from "node:url";
 import { withSentryConfig } from "@sentry/nextjs";
 import { env } from "./src/env.mjs";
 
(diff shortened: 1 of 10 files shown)


web/next.config.mjs
```diff
@@ -103,6 +104,11 @@ const nextConfig = {
     },
   },
   turbopack: {
+    // Pin the workspace root: in a nested git worktree Next.js infers the
+    // OUTER checkout's root (outermost lockfile wins), which breaks the
+    // relative resolveAlias below. fileURLToPath, not URL.pathname — the
+    // latter percent-encodes spaces and breaks on Windows drive letters.
+    root: fileURLToPath(new URL("..", import.meta.url)),
```
nkabardin: this probably added by agent mistake?

nmtrang29: _(Reply written by Claude on Trang's behalf.)_

Deliberate, not a mistake — but you're right that it doesn't belong in this PR. It pins `turbopack.root` because `next dev` run from a git worktree nested inside the main checkout (`.claude/worktrees/…`) infers the **outer** checkout's workspace root (outermost lockfile wins), which breaks the relative `resolveAlias` for `@langfuse/shared` — that's how this branch was being reviewed locally.

Since it's a dev-environment fix unrelated to the Pulse strip, it's now split out into #15541 and this branch no longer touches `next.config.mjs`.

PR #15545, opened 2026-07-28
fix(mcp): honor X-Forwarded-Host behind trusted proxies (opt-in)

### LINKED ISSUE #15534: Consider `X-Forwarded-Host` header for MCP Server
## Problem Description

The Langfuse Model Context Protocol (MCP) Server validates incoming connections by inspecting the HTTP `Host` header against allowed host configurations. 

However, in many production cloud environments (such as AWS ALB, Cloudflare, Traefik, Nginx reverse proxies, or Kubernetes Ingresses), the original `Host` header is modified or replaced with the internal service name/IP address when forwarding traffic to the upstream container. Instead, standard proxy architectures relay the original client host in the **`X-Forwarded-Host`** header.

Because the MCP server currently relies solely on the `Host` header:
* Requests passing through reverse proxies/ingress controllers fail host verification.
* Operators are forced to alter standard proxy rules or hardcode host rewrites, which goes against best practices for many cloud infrastructure providers.

## Proposed Solution

Update the MCP Server's request verification logic to respect proxy headers:

1. **Fallback/Fallback order:** Inspect `X-Forwarded-Host` prior to or as a fallback to `Host` (e.g., `req.headers['x-forwarded-host'] || req.headers['host']`).
2. *(Optional / Recommended)* **Trust Proxy Setting:** If applicable (e.g., Express/Fastify backends), expose or respect standard `TRUST_PROXY` / proxy settings so header spoofing risks are mitigated when behind trusted reverse proxies.

## Additional Context & Standard References

* **RFC 7239 / De-facto Standard:** `X-Forwarded-Host` is the widely accepted standard header for identifying the original host requested by the client in HTTP proxying.
* **Impact:** Affects users deploying MCP servers in Kubernetes, AWS ECS/EKS behind ALBs, or behind Cloudflare/Traefik API gateways.

Mohith26:
Fixes #15534, reported by @thepabloaguilar. @bezbac I see you were assigned recently, happy to close this if you already have a fix in flight.

The MCP server validates requests against a Host allowlist for DNS-rebinding protection. Reverse proxies and ingresses (ALB, Cloudflare, Traefik, nginx) rewrite Host to the internal upstream name and carry the original host in X-Forwarded-Host, so requests through standard proxy setups fail host verification.

This honors X-Forwarded-Host (first comma-separated value, mirroring the existing x-forwarded-proto handling in getRequestOrigin), falling back to Host when absent.

Since X-Forwarded-Host is client-settable, trusting it unconditionally would defeat the DNS-rebinding protection. It is therefore only honored behind an explicit opt-in, a new `LANGFUSE_MCP_TRUST_FORWARDED_HEADERS` env var (default off), meant for deployments behind a trusted proxy that overwrites the header. X-Forwarded-Proto is intentionally not consumed since host validation here is scheme-agnostic.

Tests cover the forwarded path (incl. Origin and comma-separated lists), Host fallback, rejection of non-allowlisted forwarded hosts, and the header being ignored when the flag is off.

### DIFF
--- a/web/src/__tests__/server/unit/mcp-security.servertest.ts
+++ b/web/src/__tests__/server/unit/mcp-security.servertest.ts
@@ -1,6 +1,7 @@
 const mockEnv = vi.hoisted(() => ({
   env: {
     LANGFUSE_MCP_ALLOWED_HOSTS: [] as string[],
+    LANGFUSE_MCP_TRUST_FORWARDED_HEADERS: undefined as string | undefined,
     NEXTAUTH_URL: "https://langfuse.example.com",
     NODE_ENV: "production",
   },
@@ -19,6 +20,7 @@ describe("MCP request security", () => {
     mockEnv.env.NEXTAUTH_URL = "https://langfuse.example.com";
     mockEnv.env.NODE_ENV = "production";
     mockEnv.env.LANGFUSE_MCP_ALLOWED_HOSTS = [];
+    mockEnv.env.LANGFUSE_MCP_TRUST_FORWARDED_HEADERS = undefined;
   });
 
   it("allows an exact additional host from LANGFUSE_MCP_ALLOWED_HOSTS", () => {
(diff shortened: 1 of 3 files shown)


bezbac:
Thanks for putting this together!

I think the approach in this PR is definitely interesting, but I'd like to make sure we fully understand the problem space before we commit to a particular solution.

There are a few possible directions we could take here (including making host header verification configurable), and I'd prefer to continue the discussion in the issue first to make sure we're solving the right problem rather than immediately implementing one specific approach.

I'll hold off on reviewing or merging this for now while we gather a bit more feedback on the GitHub issue, but I appreciate you taking the time to put together the implementation.

bezbac:
Hi @Mohith26, thanks again for the contribution!

I'm going to close this PR for now since we ended up merging the simpler approach of making Host validation configurable in https://github.com/langfuse/langfuse/pull/15584.

If we see a concrete need for `X-Forwarded-Host`-based validation in the future, we can definitely revisit this approach. Thanks again for taking the time to put this together!

PR #15556, opened 2026-07-29
fix(traces): push score names into the traces table scores CTE

### LINKED ISSUE #15553: bug: traces table score filters aggregate every score name in the project, exhausting ClickHouse memory
### Describe the bug

Filtering the Traces table by a **score value** makes `getTracesTableGeneric` build a `scores_avg` CTE that aggregates **every score name in the project**, then discards all but the filtered name in the outer `WHERE`. On projects with high score volume this CTE dominates the cost of the statement and can exhaust the ClickHouse server memory budget, taking unrelated queries down with it.

The CTE is bounded only from below:

```sql
FROM scores s FINAL
WHERE project_id = {projectId: String}
  AND s.timestamp >= {traceTimestamp: DateTime64(3)} - INTERVAL 1 HOUR   -- no upper bound
GROUP BY project_id, trace_id, name, data_type, string_value
```

The missing upper bound looks deliberate (scores can be attached long after their trace, so bounding at `toTimestamp` would drop late annotations and eval scores). The consequence is that **narrowing the UI date range barely helps** — the scan is anchored at the window start and always runs to "now". Selecting an *older* 24 h window is more expensive than a recent one.

`name`, meanwhile, is never filtered — even though it is the third column of the sort key:

```sql
ORDER BY (project_id, toDate(timestamp), name, id)
```

### Impact observed

Self-hosted, `3.185.0` and reproduced on `3.224.1`, ClickHouse 25.2 (3 replicas, 25 GiB limit → `max_server_memory_usage` 22.5 GiB), `scores` 92 GB / `observations` 480 GB.

```
Code: 241. DB::Exception: (total) memory limit exceeded: would use 12.13 GiB
(attempt to allocate chunk of 8372224 bytes), current RSS: 22.49 GiB, maximum: 22.50 GiB.
OvercommitTracker decision: Query was selected to stop by OvercommitTracker:
While executing AggregatingTransform. (MEMORY_LIMIT_EXCEEDED)
(comment: {"surface":"trpc","route":"traces.countAll","projectId":"..."})
```

Failures also land on `FillingRightJoinSide` — the CTE is the build side of the `LEFT JOIN`, and hash joins have no spill path.

### Proposed fix

Restrict the CTE to the score names the filters reference. This is sound for `select` values that do not project the CTE:

| select | route | uses `s.scores_avg` |
| -- | -- | -- |
| `count` | `traces.countAll` | filter only |
| `rows` | `traces.all` | filter only |
| `identifiers` | — | filter only |
| `metrics` | `traces.metrics` | **projects** `scores_avg` / `score_categories` / `score_booleans` |

Every score filter (`NumberObjectFilter`, `CategoryOptionsFilter`, `BooleanObjectFilter`) carries a `key` — the score name — and only ever inspects array entries carrying that key, so entries under other names cannot change the outcome. That holds for negated operators too (`none of`, `<>`), because traces left without a CTE row are filled by the `LEFT JOIN` with empty arrays — the same value an unrestricted CTE would have produced for a trace matching none of the filtered names.

The pushdown must be skipped when ordering by a score column (the full per-trace aggregate is needed) and when any score filter does not name a score (e.g. a `NullFilter`).

Since `name` is in the sort key, this converts a full-project scan into a primary key range scan. No new index is required.

### Steps to reproduce

1. A project with a large number of scores across several score names.
2. Open the Traces table, set a date range that ended some days ago.
3. Add a numeric score filter, e.g. `Scores (numeric)` → `<some score name>` → `<` → `0`.
4. Observe `traces.all` / `traces.countAll` in the ClickHouse query log: the `scores_avg` CTE has no `name` predicate and no upper time bound.

Self-hosted

3.224.1 (originally found on 3.185.0)

Yes — PR follows.

thspinto:
## What does this PR do?

Restricts the `scores_avg` CTE in `getTracesTableGeneric` to the score names the filters actually reference.

Filtering the Traces table by a score value currently makes that CTE aggregate **every score name in the project**, and the outer `WHERE` then discards all but the filtered name. The CTE is bounded only from below:

```sql
FROM scores s FINAL
WHERE project_id = {projectId: String}
  AND s.timestamp >= {traceTimestamp: DateTime64(3)} - INTERVAL 1 HOUR   -- no upper bound
GROUP BY project_id, trace_id, name, data_type, string_value
```

The absent upper bound looks deliberate — scores can be attached long after their trace, so bounding at `toTimestamp` would drop late annotations and eval scores. A consequence is that narrowing the UI date range barely helps: the scan is anchored at the window start and always runs to "now", so an *older* 24 h window costs more than a recent one.

`name` is never filtered, even though it is the third column of the sort key:

```sql
ORDER BY (project_id, toDate(timestamp), name, id)
```

so a `name` predicate is a primary key range scan. No new index is required.

Fixes #15553

## Production measurements

Self-hosted, ClickHouse 25.2, 3 replicas, 25 GiB container limit (`max_server_memory_usage` 22.5 GiB). One project, `scores` 92 GB / `observations` 480 GB. Project has **69 distinct score names**; the three largest account for ~72% of score volume, and the name being filtered is **0.04%** of it.

For one real Traces-table query (single score filter, 2-day trace window, user filter), measured against the live dataset:

| Stage | Before | After | Factor |
|---|---:|---:|---:|
| Scores scanned | 65,412,538 | 25,265 | **2,589×** |
| Join build rows (`FillingRightJoinSide`) | 28,587,067 | 3,436 | **8,320×** |
| Inner GROUP BY groups (`AggregatingTransform`) | 60,332,535 | 3,436 | **17,558×** |

**With the pushdown, the same query completes and returns rows:**

| | |
|---|---:|
| Peak memory | **51.14 MiB** |
| Rows read | 598.9 K |
| Bytes read | 88.2 MiB |
| Duration | **336 ms** |

## Why this is sound

The pushdown is applied only where the CTE is not projected:

| `select` | route | uses `s.scores_avg` |
|---|---|---|
| `count` | `traces.countAll` | filter only |
| `rows` | `traces.all` | filter only |
| `identifiers` | — | filter only |
| `metrics` | `traces.metrics` | **projects** `scores_avg` / `score_categories` / `score_booleans` |

It is additionally skipped when ordering by a score column (the full per-trace aggregate is needed) and when any score filter does not name a score (e.g. a `NullFilter`, which can match through any name).

Every score filter — `NumberObjectFilter`, `CategoryOptionsFilter`, `BooleanObjectFilter` — carries a `key` holding the score name, and each only ever inspects array entries carrying that key. Entries under other names cannot change the outcome. This holds for negated operators too (`none of`, `<>`): traces left without a CTE row are filled by the `LEFT JOIN` with empty arrays, which is exactly the value an unrestricted CTE produces for a trace matching none of the filtered names. `join_use_nulls` is not set anywhere in the repo, so the default of filling type defaults applies.

## Type of change

## Verification

- `packages/shared` build (`tsc`) — clean
- `web` typecheck — clean
- `eslint` on both changed files — clean
- `prettier` — clean
- Repo-wide `format:check` + `lint` (pre-commit hook) — 7/7 tasks pass
- `traces-ui-table.servertest.ts` against live ClickHouse + Postgres — **9 passed** (7 pre-existing + 2 new)

Two tests added: one asserting that only the correctly-scored trace matches, with the `rows` and `count` paths agreeing; one covering the negated `<>` operator on `score_booleans`, which is the case where the pushdown could plausibly change results.

Confirmed via `system.query_log` that the pushdown fires only when intended — the new tests emit `s.name IN (...)`, the pre-existing tests (no score filters) do not.

## Mandatory Tasks

Narrows the Traces table score aggregation to score names referenced by typed filters when full score data is not otherwise required.
- Collects and deduplicates names from numeric, categorical, and boolean score filters.
- Skips pushdown for metrics queries, score-based ordering, and filters without a specific score name.
- Adds server tests covering positive numeric filtering, count/row consistency, and negated boolean filtering for traces without the named score.
</details>

The pushdown is limited to non-projecting, non-score-ordered query modes, includes every referenced typed score key, and is disabled when a score filter cannot supply a name; the affected filter predicates only depend on entries for their own keys.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(traces): push score names into the t..."](https://github.com/langfuse/langfuse/commit/956015412f44285411916ee2dc88ef5028bac281) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48257770)</sub>

**Context used:**

### DIFF
--- a/packages/shared/src/server/services/traces-ui-table-service.ts
+++ b/packages/shared/src/server/services/traces-ui-table-service.ts
@@ -8,6 +8,9 @@ import {
   StringFilter,
   StringOptionsFilter,
   DateTimeFilter,
+  NumberObjectFilter,
+  CategoryOptionsFilter,
+  BooleanObjectFilter,
 } from "../queries/clickhouse-sql/clickhouse-filter";
 import {
   getProjectIdDefaultFilter,
@@ -274,10 +277,57 @@ async function getTracesTableGeneric(props: FetchTracesTableProps) {
       f.field === "timestamp" && (f.operator === ">=" || f.operator === ">"),
   ) as DateTimeFilter | undefined;
 
-  const requiresScoresJoin =
-    tracesFilter.find((f) => f.clickhouseTable === "scores") !== undefined ||
+  const ordersByScores =
     findUiColumnMapping(tracesTableUiColumnDefinitions, orderBy?.column)
       ?.clickhouseTableName === "scores";
+
+  const requiresScoresJoin =
+    tracesFilter.find((f) => f.clickhouseTable === "scores") !== undefined ||
+    ordersByScores;
+
+  // Restrict the scores CTE to the score names the filters actually reference.
+  //
+  // `scores` is ORDER BY (project_id, toDate(timestamp), name, id), so a `name`
+  // predicate is a primary key range scan. Without it the CTE aggregates every
+  // score name in the project from `traceTimestamp - 1 hour` onwards — there is
+  // no upper bound, because scores may be attached long after their trace — and
+  // the outer query then discards all but the filtered names. On projects with
+  // high score volume that CTE is the dominant cost of the whole statement, both
+  // as GROUP BY state and as the right side of the LEFT JOIN.
+  //
+  // Only sound when the CTE is not projected and not ordered on: `metrics`
+  // selects s.scores_avg / s.score_categories / s.score_booleans, and ordering by
+  // a score column needs the full per-trace aggregate. Score filters that do not
+  // name a score (e.g. a NullFilter on the column) disable the pushdown, since
+  // they can match a trace through any name.
+  //
+  // Equivalence holds for negated operators too ("!=", "none of", "<>"): each of
+  // these filters only ever inspects array entries carrying its own key, so
+  // entries for other names cannot change the outcome. Traces left without a CTE
+  // row are filled by the LEFT JOIN with empty arrays — the same value they would
+  // have had from an unfiltered CTE that matched none of their scores.
+  const scoreFilterNames: string[] = [];
+  let everyScoreFilterNamesAScore = true;
+  tracesFilter.forEach((f) => {
+    if (f.clickhouseTable !== "scores") return;
+    if (
+      f instanceof NumberObjectFilter ||
+      f instanceof CategoryOptionsFilter ||
+      f instanceof BooleanObjectFilter
+    ) {
+      scoreFilterNames.push(f.key);
+    } else {
+      everyScoreFilterNamesAScore = false;
+    }
+  });
+
+  const pushedDownScoreNames =
+    select !== "metrics" &&
+    !ordersByScores &&
+    everyScoreFilterNamesAScore &&
+    scoreFilterNames.length > 0
+      ? Array.from(new Set(scoreFilterNames))
+      : null;
 
   const requiresObservationsJoin =
     tracesFilter.find((f) => f.clickhouseTable === "observations") !==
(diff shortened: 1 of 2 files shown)


sumerman:
@thspinto thank you for your effort, I will go with my own implementation to cover v4 surface as well https://github.com/langfuse/langfuse/pull/15591

PR #15585, opened 2026-07-29
perf(worker): read the dataset item once per eval variable extraction

irontaek:
## What

`extractVariablesFromTracingData` issues one `prisma.datasetItem.findFirst` **per dataset_item variable**, and every one of those queries has an identical `WHERE`:

```ts
for (const variable of variables) {
  ...
  const prismaField = snakeToCamel(safeInternalColumn.id);
  const datasetItem = await prisma.datasetItem.findFirst({
    select: { [prismaField]: true },     // only this differs
    where: {
      id: datasetItemId,                 // same every iteration
      projectId,                         // same every iteration
      ...(datasetItemValidFrom ? { validFrom: datasetItemValidFrom } : { validTo: null }),
    },
  });
```

So an eval template with N dataset_item variables reads **the same row N times**, each time selecting a different column. A template mapping `input` and `expected_output` — which is what the existing test in `worker/src/__tests__/evalService.test.ts` ("extracts variables from a dataset item") sets up — does two round trips for one row.

## How

Collect the mapped columns before the loop and do the lookup once, selecting the union:

```ts
const datasetItemColumns = new Set<string>();
// ...resolve each dataset_item variable's column exactly as the loop does...

let datasetItemRow = null;
let datasetItemFetched = false;
const getDatasetItem = async () => {
  if (datasetItemFetched) return datasetItemRow;
  datasetItemFetched = true;
  datasetItemRow = await prisma.datasetItem.findFirst({
    select: Object.fromEntries([...datasetItemColumns].map((c) => [c, true])),
    where: { /* unchanged */ },
  });
  return datasetItemRow;
};
```

Two properties worth calling out:

- **Still selects only the mapped columns.** I deliberately did not switch to selecting the whole row — the dataset item payload columns can be large, and the current code is right to fetch narrowly. The union of mapped columns is the same data that was being fetched before, just in one round trip instead of N.
- **Still lazy.** The query only fires if a dataset_item variable actually reaches that branch, so calls with no dataset item variables issue no query, exactly as today.

Everything else — the column-resolution guard, the "dataset item not found" error, `parseDatabaseRowValue` — is untouched.

## About the comment I removed

The function's header said:

> `// We do not cache dataset items as Postgres is cheaper than ClickHouse.`

I want to be explicit that I read it and am not overriding it lightly. That line explains why dataset items don't get a `Map` cache like traces and observations do, and as a *caching* decision it's reasonable — Postgres lookups are cheap relative to the ClickHouse reads the other caches protect.

This change isn't caching across different lookups, though. Every dataset_item variable in a single call resolves the *same row with the same WHERE*; the only thing that varied was which column got selected. Collapsing identical queries into one is a different thing from deciding whether to memoize distinct ones — and it costs nothing at N=1, which is why I think it's worth doing even given that comment. Happy to restore the line (reworded) if you'd rather keep the rationale documented.

## Testing

`worker/src/__tests__/evalService.test.ts` → `test variable extraction` already covers this precisely. The first test maps two dataset_item columns (`input`, `expected_output`) from one item and asserts both extracted values, which is exactly the multi-variable path this change collapses.

I ran that block against a local Postgres + ClickHouse (docker compose, migrations applied):

```
npx vitest run src/__tests__/evalService.test.ts -t "test variable extraction"

Test Files  1 passed (1)
     Tests  16 passed | 86 skipped (102)
```

Identical on a clean `main` (16 passed) and with this change applied, so the covered behaviour is unchanged — including `extracts variables from a versioned dataset item using datasetItemValidFrom`, which exercises the `validFrom` branch of the WHERE.

`tsc --noEmit` in `worker` reports 0 errors.

## Impact

Evals run per trace in the worker, so this removes N−1 Postgres round trips per eval execution for every template with more than one dataset item variable. It is not a hot-loop micro-optimisation — the win is round trips, which is why there is no benchmark table here; the metric is query count, and it goes from N to 1.

Consolidates dataset-item variable extraction into one lazy Prisma lookup per evaluation call while preserving narrow column selection and existing project/version constraints.
- Precomputes the union of mapped dataset-item columns.
- Reuses the fetched row across all dataset-item variables.
- Leaves trace and observation extraction unchanged.
</details>

The consolidated lookup derives its selected fields through the same validated mapping and column-resolution logic used during extraction, while retaining the existing dataset ID, project, and version filters.
</details>

<sub>Reviews (1): Last reviewed commit: ["perf(worker): read the dataset item once..."](https://github.com/langfuse/langfuse/commit/847974276637cd91494ce60b79c5a62fcea5122e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48337566)</sub>


irontaek:
Closing this to keep my open PR count down — it has been sitting well past this repo's usual review turnaround, and I would rather not leave unsolicited changes cluttering the queue. The change is self-contained and still applies; happy to reopen if a maintainer wants to take a look.

PR #15638, opened 2026-07-30
perf(worker): skip tokenization when cost details are provided

Steffen911:
🟡 **Live preview: [pr-15638.preview.langfuse.com](https://pr-15638.preview.langfuse.com)**

# Description

Tokenization currently runs even when the user already provided cost details. Those tokenized counts can never affect costs: `calculateUsageCosts` short-circuits on the first provided cost point ("If user has provided any cost point, do not calculate any other cost points") and ignores computed usage entirely. So for cost-providing users we were paying the full tokenization cost — worker-thread CPU, and in the worst case the multi-megabyte-payload timeouts behind the LFE-9339 event-loop incidents — to produce token counts that only ever served display.

## Change

`IngestionService.getUsageUnits` now skips manual tokenization when `provided_cost_details` contains at least one non-null value (same predicate `calculateUsageCosts` uses to decide costs are user-managed). The observation ingests with provided costs as-is and **blank** `usage_details`.

## Safety analysis (is `provided_cost_details` ever initialized implicitly?)

Checked every construction site — none fabricates cost details:

- **Legacy ingestion** (`IngestionService` observation path): starts `{}`; keys only added via explicit `!= null` checks on `usage.inputCost/outputCost/totalCost` and a null-filtered spread of `body.costDetails`.
- **Events path** (`internalTraceEvents.ts`): `asNumberRecord(body.costDetails)` keeps only finite numbers and returns `undefined` when empty; snapshots propagate earlier provided costs, which is desired (a cost provided on any event of the observation marks costs as user-managed).
- **OTel** (`OtelIngestionProcessor.extractCostDetails`): defaults to `{}`; only populated from explicit cost attributes (`langfuse.observation.cost_details`, `gen_ai.usage.cost`, `llm.cost.total`).
- **ClickHouse read-merge**: `provided_cost_details` round-trips from the existing row; values may arrive as strings, which the `!= null` predicate treats the same way `calculateUsageCosts` already does.

## Behavior change

Observations with provided costs but no provided usage no longer get estimated token counts — `usage_details` stays empty (absence over estimates, consistent with the LFE-9339 direction). This also applies on updates: if a generation was tokenized on create and a later event adds cost details, the merged record ends with blank usage details, matching the rule "provided costs ⇒ usage stays blank".

`usage_pricing_tier_*` is now explicitly left unset for records without usage: review caught that `matchPricingTier` ran against the truthy `{}` and stamped the default tier from a fabricated all-zero usage vector (a pre-existing quirk on the ERROR-level path, made more common by this PR). Tier matching is now gated on non-empty `usage_details`. Costs are unaffected either way (provided cost points always short-circuit `calculateUsageCosts`).

Also carries a cherry-pick of #15644 (zizmor ignore for the gh-aw compiled workflow) so `Check GitHub Actions security` passes — that failure comes from main via #15639 and hits every PR; identical patch, git dedups whichever merges first.

## Impacted packages

- `worker`

## Verification

- `pnpm --filter worker run test calculateTokenCost.unit.test.ts` → `Tests  29 passed (29)`; the new test (`should skip tokenization and leave usage details blank when cost details are provided`) was written first and failed against the old behavior with tokenized `usage_details {input: 2, output: 3, total: 5}` instead of `{}`.
- `pnpm --filter worker run typecheck` → clean.
- `pnpm run lint` → see pre-commit hook run (`turbo run lint`) on the commit.

Note: independent of #15627 (skip-on-async-failure), but both touch `getUsageUnits`; whichever merges second has a trivial conflict in the `Pick<...>` type list.


Steffen911:
@claude review

Steffen911:
Re the review finding on `usage_pricing_tier_*`: confirmed and fixed in f50b5c02c — tier matching is now gated on non-empty `usage_details` (per the suggested fix), so cost-providing observations (and the pre-existing ERROR-level path) leave the tier unset instead of stamping the default tier against an all-zero usage vector. The new test pins `usage_pricing_tier_id/name` as `null` for this case.

Also note 41cda7126: cherry-pick of #15644 (zizmor `ref-version-mismatch` ignore for the gh-aw compiled workflow) so the `Check GitHub Actions security` check can pass here — that failure comes from main via #15639 and blocks every PR; git dedups the identical patch whichever PR merges first. Happy to drop it from this branch if you'd rather merge #15644 first.

PR #15641, opened 2026-07-30
fix(ci): pin docker/login-action by tag so gh-aw emits a version comment

wochinge:
🟡 **Live preview: [pr-15641.preview.langfuse.com](https://pr-15641.preview.langfuse.com)**

## Summary

- **Unblocks every PR in the repo.** #15639 merged while its checks were still running, and it landed a `ci-runtime-analyst.lock.yml` whose `docker/login-action` pin carries the SHA as its version comment (`@650006... # 650006...`). zizmor flags that as `ref-version-mismatch` (medium), and since it scans *all* workflow files, the finding now fails `Check GitHub Actions security` on **any** PR — which is a required check for merging to main.
- Main itself looks green only because `zizmor.yml` sets `advanced-security: true` on `push`, which uploads SARIF to code scanning instead of gating. `pull_request` runs gate at `min-severity: low`, so the breakage is PR-only and was invisible on the merge commit.
- Root cause is a gh-aw compiler behaviour change: v0.82.0 reverse-looked-up the tag for a SHA-pinned `uses:` in a custom step and emitted `# v4.2.0`; v0.83.5 and v0.84.0 emit the SHA instead. The source `.md` was correct all along.
- Fix references the action by **tag** and lets gh-aw do the pinning — already how `pnpm/action-setup` is declared two lines above in the same file.

## Major Decisions

- **Tag in source, SHA in compiled output** rather than editing the generated lock (which is marked DO NOT EDIT and is rewritten by every `gh aw compile`, so the failure would silently return). Adding a `docker/login-action` entry to `actions-lock.json` by hand was also tried and does not change the emitted comment.
- **No zizmor suppression.** The finding is legitimate and cheap to fix properly; suppressing a security audit to unblock a merge would be the wrong trade, and this repo requires explicit approval to widen lint/config overrides.
- **Not reverting to gh-aw v0.82.0**, which does not exhibit the bug but pins firewall 0.27.13 — that release's AI-credits pricing table does not know `claude-sonnet-5`, and the AWF api-proxy rejects unpriced models with HTTP 400.

## Review Focus

- **Which code runs is unchanged.** The compiled workflow is still pinned to the identical commit `650006c6eb7dba73a995cc03b0b2d7f5ca915bee`; only the trailing comment differs. `actions-lock.json` now records the tag → SHA mapping, so the pin stays reproducible rather than resolved at compile time.
- Worth a second opinion on whether tag-in-source is the idiom this repo wants for gh-aw custom steps, since it is now consistent within the file but differs from the SHA-pinned style used in hand-written workflows like `pipeline.yml`.
- `gh aw compile` also warns that `docker/login-action@v4.2.0` is outdated (latest `v4.4.0`). Deliberately left alone here to keep this change a no-op on behaviour.

This PR restores valid version metadata for the Docker login action without changing the executed code.
- Changes the gh-aw source workflow to reference `docker/login-action@v4.2.0`.
- Records the tag-to-SHA mapping in the gh-aw actions lock file.
- Regenerates the compiled workflow with the same action SHA and the corrected `v4.2.0` comment.
</details>

The PR appears safe to merge because the generated workflow continues to execute the identical pinned Docker login action commit.

The source tag, lock-file mapping, generated manifest, and compiled workflow are consistent, and the runtime `uses` reference remains pinned to the pre-existing SHA.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ci): pin docker/login-action by tag ..."](https://github.com/langfuse/langfuse/commit/aa6d04ad00e1e87b6e36e7c09c9a19899f4680ae) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48711891)</sub>


wochinge:
Superseded by #15644, which lands the same fix — a `ref-version-mismatch` ignore scoped to `ci-runtime-analyst.lock.yml` in `.github/zizmor.yml`, matching the two existing gh-aw entries there.

That approach is better than this PR's: it keeps the SHA pin in the source `.md`, whereas this PR switched `docker/login-action` to a tag to coax the correct comment out of the compiler. The pin is the security control and should stay a SHA.

Verified #15644 locally against main's compiled lock file:

```
GH_TOKEN=… zizmor --min-severity low .github/workflows/
No findings to report. Good job! (19 ignored, 204 suppressed)
```

Underlying cause, for the record: gh-aw >= v0.83.5 stopped reverse-resolving the tag for a SHA-pinned `uses:` in a custom step and re-emits the SHA as the version comment; v0.82.0 emitted `# v4.2.0`. Only the comment is affected — the pinned SHA is unchanged.

PR #15653, opened 2026-07-30
chore(agent-dx): scope agent guidance to packages and validate it in CI

wochinge:
🟡 **Live preview: [pr-15653.preview.langfuse.com](https://pr-15653.preview.langfuse.com)**

## Summary

- Package-local agent guidance only reached Codex. `web/` had a hand-written `CLAUDE.md` stub; `worker/`, `ee/`, `packages/shared/`, and the seeder had none — so Claude read a root guide carrying web-specific rules while four packages' guidance stayed invisible to it. The sync script now auto-discovers every `AGENTS.md` in the tree and generates a sibling `CLAUDE.md` symlink, mirroring how the repo root already works.
- Web-only rules move out of the always-loaded root guide (`How To Work`: 51 → 33 lines). Nested guidance loads when an agent *reads a file* in that directory, which always precedes an edit, so the root pointers were paying context rent for something that arrives on its own. The `useEffect`, browser-review, charts, and screenshot bullets were each already stated in `web/AGENTS.md` or in an always-loaded skill `description`.
- `web/CLAUDE.md` no longer imports `@../AGENTS.md`, which was re-injecting all 166 root lines a second time on any web file read.
- `agents:check` now resolves the paths an `AGENTS.md` cites. It immediately caught `web/AGENTS.md` pointing at `src/features/rbac/README.md` — added in c2048de88 (Mar 2025), removed from `main` in a squashed merge, stale ever since. Repointed at the real entry points.
- CI gates the committed state. `postinstall` regenerates shims immediately before checking them, so any repairable drift was silently fixed and never reported — which is why `agents:check` has been a no-op gate since it was written. The lint job now fails if that repair dirties the tree.
- Boilerplate trimmed from all four package files: `# Codex Guidelines` → `# Agent Guidelines`, dropped the tautological intro line and the redundant "use root AGENTS.md" pointer, collapsed the Maintenance Contract from 5–7 lines to one.

## Major Decisions

- **Symlinks over `@`-import stubs.** The repo root already proves a symlinked `CLAUDE.md` works, and a symlink cannot drift from its target. The hand-written `web/CLAUDE.md` had already drifted — it carried the duplicate root import. Tracked (not gitignored) to match the root pair, so a fresh clone has guidance before `pnpm install` runs.
- **Path validation is opt-in via `--check-paths`.** `postinstall` runs shim verification only. Without the split, a stale doc path fails `pnpm i` and with it the eleven jobs that install — one typo turning eleven jobs red with `ELIFECYCLE` noise. It now fails one job with a readable message.
- **Only flag paths whose parent directory exists.** That's the signature of a moved or renamed file. An earlier "skip anything outside the repo" rule failed because `web/../langfuse-docs` resolves *inside* the repo. Trade-off: a reference into a wholly deleted directory won't be caught.
- **Search-bar and storybook rules went to `web/AGENTS.md`, not down to their components.** Both have to fire while working *elsewhere* — the search-bar contract exists precisely to stop someone forking filtering in another view. Directory-scoped loading can never reach that agent.

## Review Focus

- **`web/CLAUDE.md` changes type** (`100644` → `120000`). Tracked symlinks need `core.symlinks=true`; on Windows without developer mode git materialises them as text files containing the target path. Pre-existing for the root pair, but this takes it from 2 files to 7.
- **The CI gate is a whole-tree `git status --porcelain` check.** It sits right after install and before `cp .env.dev.example .env`; any future step that dirties the tree before that point would trip it. Verified `pnpm i` itself leaves the tree clean.
- **Discovery exclusions in `findDirectoriesWithAgentsFile`.** Dot-directories are skipped, which is what keeps the vendored `web/.agents/skills/vercel-*/AGENTS.md` bundles from becoming directory-scoped instructions. Worth a sanity check that no legitimate guidance lives under a dot-path.

This PR scopes agent guidance to individual packages and adds automated validation.
- Discovers package-local `AGENTS.md` files and maintains sibling `CLAUDE.md` symlinks.
- Adds optional path-reference validation for agent guidance.
- Updates postinstall and CI to regenerate shims and reject uncommitted drift.
- Moves web-specific guidance out of the root guide and refreshes package guidance.
</details>

The PR appears safe to merge, with no concrete changed-code defect identified.

The shim synchronization, path-validation split, committed-state CI gate, and package guidance changes form a consistent workflow without an established build or runtime failure.
</details>

```mermaid
flowchart TD
  Install[pnpm install] --> Postinstall[postinstall]
  Postinstall --> Sync[agents:sync]
  Sync --> Discover[Discover package AGENTS.md files]
  Discover --> Shims[Create or remove CLAUDE.md shims]
  Shims --> Verify[Shim verification]
  Verify --> Dirty{Working tree dirty?}
  Dirty -->|yes| FailDrift[Fail CI: generated drift]
  Dirty -->|no| CheckPaths[agents:check --check-paths]
  CheckPaths --> Result{References resolve?}
  Result -->|yes| Pass[Pass]
  Result -->|no| FailPaths[Fail CI: broken guidance path]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["chore(agent-dx): scope agent guidance to..."](https://github.com/langfuse/langfuse/commit/7ced082726422c6bc43a726a276b9b7d1ea5ba8b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48742963)</sub>


wochinge:
@claude review

PR #14667, opened 2026-07-01
fix(trace): decode unicode escapes in trace JSON download

toyayuto:
## What does this PR do?

Routes the **"Download trace as JSON"** payload through the shared `stringify` helper (which decodes `\uXXXX` escapes) instead of raw `JSON.stringify`, so downloaded trace files render non-ASCII content (Japanese, Chinese, Korean, emoji, ...) as their original characters.

This is a follow-up to #12882 (batch CSV/JSON/JSONL export), #9686 (IOTableCell in the trace table) and #13223 (PrettyJsonView in the trace detail view). The trace-detail **download** endpoint was the one remaining path still emitting raw `\uXXXX`.

Refs #10972

### Motivation

The Python SDK defaults to `ensure_ascii=True` when serialising `input` / `output` / `metadata`, so a payload like `"以上"` is stored in ClickHouse as `"以上"`. After #9686, #12882 and #13223 these escaped sequences render correctly in the trace table, in exports, and in the trace detail view — but the **"Download trace as JSON"** button (`GET /api/public/... /traces/[traceId]/download`) still used raw `JSON.stringify`, so the downloaded file contained literal `\uXXXX`, which is hard to read for users of non-Latin languages.

### Changes

**`web/src/pages/api/traces/[traceId]/download.ts`** (+6 / -1)

- Import `stringify` from `@langfuse/shared/src/server`.
- Replace `JSON.stringify(payload, null, 2)` with `stringify(payload, undefined, 2)`. `stringify` already applies `decodeUnicodeEscapesOnly` (greedy) to every string value and handles bigint — it is the same helper used by the batch export pipeline (#12882), so the download is now consistent with exports.

**`packages/shared/src/server/utils/transforms/stringify.ts`** (+ optional indent arg)

- Extract the replacer into a named `stringifyReplacer` constant (readability, no behavior change).
- Add an optional `indent?: number` parameter to `stringify`. When provided it is used directly; otherwise the existing behavior is preserved (`key === "comments" ? 2 : undefined`). This lets the trace download keep its human-readable 2-space indentation while decoding escapes. Fully backward compatible — existing callers pass only `(data, key?)`.

**`worker/src/__tests__/stringify.test.ts`** (+ 2 tests)

- `stringify` applies an explicit indent when provided.
- `stringify` decodes unicode escapes while keeping an explicit indent (mirrors the trace-download payload shape: a stringified JSON blob whose Japanese characters were escaped by `ensure_ascii=True`).

### Backward compatibility

[...]

PR #15698, opened 2026-08-03
fix(trace): decode unicode escapes in client-side trace downloads and copies

toyayuto:
## What does this PR do?

Routes the three remaining **client-side** JSON serialization paths in the trace UI through the shared `stringify` helper (which decodes `\uXXXX` escapes) instead of raw `JSON.stringify`, so downloaded/copied trace data renders non-ASCII content (Japanese, Chinese, Korean, emoji, ...) as real characters.

This is a follow-up to #9686 (IOTableCell), #12882 (batch exports), #13223 (PrettyJsonView) and #14667 (server-side trace download route). #14667 fixed `/api/traces/[traceId]/download`, but that route is only reached when the user has **V4 Beta enabled**. Users without it (the default: `session.user.v4BetaEnabled = false`) take the legacy client-side branch in `TracePanelNavigationHeader`, which still emitted raw `\uXXXX` — along with two more client-side paths found while fixing it.

Refs #10972

### Motivation

The Python SDK defaults to `ensure_ascii=True`, so a payload like `"こんにちは"` is stored as `"\u3053\u3093\u306b\u3061\u306f"`. The rendered UI decodes these, but a user on the default (non-beta) path who clicks "Download trace as JSON" gets a file full of literal `\uXXXX`.

Affected paths (all client-side, all raw `JSON.stringify` before this PR):

1. `downloadLegacyTraceAsJson` — trace download for non-beta users (`web/src/components/trace/lib/download-trace.ts`)
2. `useLogViewDownload` — Log View download + the three copy-to-clipboard branches (`web/src/components/trace/components/TraceLogView/useLogViewDownload.ts`)
3. `ChatMessage` — copy button on tool-call-only messages (`web/src/components/trace/components/IOPreview/components/ChatMessage.tsx`)

### Changes

**`packages/shared/src/utils/stringify.ts`** (moved from `src/server/utils/transforms/`)

- `stringify` / `stringifyForCsv` have no server-only dependencies (only `decodeUnicodeEscapesOnly`), so they move to the client-safe utils and are exported from the root barrel (`@langfuse/shared`). The server barrel re-exports them, so `@langfuse/shared/src/server` consumers are unchanged.
- Client and server download paths now serialize through *literally the same function*, so their output is byte-identical.

**Three client paths** — replace `JSON.stringify(data, null, 2)` with `stringify(data, undefined, 2)` (same call shape as the server download route).

**`packages/shared/src/utils/stringify.test.ts`** (moved from `worker/src/__tests__/`)

- The test moves with the implementation to a colocated shared test, following the same keep-it-in-shared feedback as the unicode helper in #12882.

**New client tests** (each proving a distinct execution path, red before the fix):

- `download-trace.clienttest.ts` — legacy download Blob content is decoded
- `useLogViewDownload.clienttest.ts` — copy (clipboard string) and download (Blob content) are decoded
- `ChatMessage.clienttest.tsx` — tool-call copy is decoded (added to the existing suite)

### Behavior notes

- **bigint**: these client paths previously threw on bigint values (raw `JSON.stringify`); they now serialize them as numbers, matching the server download.
- **Object keys**: `stringify` is replacer-based, so it decodes string *values* but not object *keys* — identical to the server download route. The rendered view (`decodeUnicodeInJson`) also decodes keys, so copied text can differ from the rendered view in keys only. Kept as-is to stay byte-identical with the server download.
- `useLogViewDownload.handleDownloadJson` is currently not wired to any UI element (`TraceLogView.tsx` only consumes `handleCopyJson`), but it is fixed and tested so it behaves correctly when connected.

### Verification

- `pnpm run lint` / `pnpm run typecheck`: 7/7 tasks successful
- Unit tests: shared `stringify.test.ts` (7 passed, 2 skipped), 3 client test files (17 passed)
- Browser-verified all three paths against seeded data containing literal `\uXXXX`:
  - Log View copy and ChatMessage tool-call copy (events mode)
  - Legacy trace download under the exact reported condition (`LANGFUSE_MIGRATION_V4_WRITE_MODE=dual`, `v4BetaEnabled: false`) — downloaded Blob has the legacy `{"trace": ...}` shape with decoded characters and no remaining `\uXXXX`

Repro for reviewers:

```bash
pnpm run seed -- trace-tree --payload-style unicode --v4
LANGFUSE_MIGRATION_V4_WRITE_MODE=dual pnpm run dev:web
# open the seeded trace with a non-beta user → Download trace as JSON
```

### Backward compatibility

- No schema or API changes; `@langfuse/shared/src/server` export surface unchanged (re-export).
- Payloads without `\uXXXX` are unaffected (`decodeUnicodeEscapesOnly` early-returns when there is no backslash).

## Type of change

## Mandatory Tasks

This PR moves the shared Unicode-aware JSON serializer to the client-safe shared package and applies it to legacy trace downloads, Log View copy/download actions, and tool-call message copying.
- Preserves the existing server import surface through a compatibility re-export.
- Adds colocated serializer tests and client tests covering each newly updated path.
- Keeps server stream transforms using the relocated implementation.
</details>

The lossy handling of intentionally literal Unicode escape text needs to be addressed before merging.

The newly routed client paths apply greedy Unicode decoding to arbitrary trace strings, so literal escape syntax in code, prompts, tool definitions, or metadata is changed in copied and downloaded data.

**Files Needing Attention:** web/src/components/trace/lib/download-trace.ts and the other client paths newly using stringify
</details>

`````markdown
### Issue 1
web/src/components/trace/lib/download-trace.ts:85
**Literal Unicode escapes are altered**

When trace input, output, tool definitions, or metadata intentionally contain literal text such as `\u0041`—for example in generated code or an escaping example—`stringify` greedily decodes it to `A`, causing the downloaded trace and the other newly routed copy/download paths to no longer preserve the original data.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): decode unicode escapes in cl..."](https://github.com/langfuse/langfuse/commit/a730c555aa8e90db5f4efb5b9a3220622adee040) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49576083)</sub>

**Context used:**

### DIFF
--- a/packages/shared/AGENTS.md
+++ b/packages/shared/AGENTS.md
@@ -37,6 +37,10 @@
 - `@langfuse/shared` via `src/index.ts`: default shared surface for
   cross-runtime types, zod schemas, table definitions, domain models, prompt
   helpers, eval/model-pricing helpers, and other frontend-safe utilities.
+  Includes the unicode-decoding JSON serialization helpers (`stringify`,
+  `stringifyForCsv` in `src/utils/stringify.ts`) used by both the server
+  trace-download route and client-side download/copy paths; the server barrel
+  re-exports them for compatibility.
 - `@langfuse/shared/src/server` via `src/server/index.ts`: server-only barrel
   for shared backend services, repositories, queue helpers/contracts, Redis and
   ClickHouse helpers, auth helpers, logger/instrumentation, ingestion helpers,
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -1,5 +1,6 @@
 export * from "./constants";
 export { decodeUnicodeEscapesOnly } from "./utils/unicode";
+export { stringify, stringifyForCsv } from "./utils/stringify";
 export * from "./interfaces/filters";
 export * from "./interfaces/orderBy";
 export * from "./interfaces/cloudConfigSchema";
--- a/packages/shared/src/server/utils/transforms/index.ts
+++ b/packages/shared/src/server/utils/transforms/index.ts
@@ -5,7 +5,10 @@ import { transformStreamToCsv } from "./transformStreamToCsv";
 import { transformStreamToJson } from "./transformStreamToJson";
 import { transformStreamToJsonl } from "./transformStreamToJsonl";
 
-export { stringify, stringifyForCsv } from "./stringify";
+// stringify lives in the client-safe utils so the web client (legacy trace
+// download, log view copy/download) serializes through the exact same helper
+// as the server-side trace download route.
+export { stringify, stringifyForCsv } from "../../../utils/stringify";
 
 export const streamTransformations: Record<
   BatchExportFileFormat,
--- a/packages/shared/src/server/utils/transforms/transformStreamToCsv.ts
+++ b/packages/shared/src/server/utils/transforms/transformStreamToCsv.ts
@@ -1,5 +1,5 @@
 import { Transform, type TransformCallback } from "stream";
-import { stringifyForCsv } from "./stringify";
+import { stringifyForCsv } from "../../../utils/stringify";
 
 const DELIMITER = ",";
 const CSV_DOUBLE_QUOTE = /"/g;
--- a/packages/shared/src/server/utils/transforms/transformStreamToJson.ts
+++ b/packages/shared/src/server/utils/transforms/transformStreamToJson.ts
@@ -1,5 +1,5 @@
 import { Transform, type TransformCallback } from "stream";
-import { stringify } from "./stringify";
+import { stringify } from "../../../utils/stringify";
 
 export function transformStreamToJson(): Transform {
   let isFirstElement = true;
--- a/packages/shared/src/server/utils/transforms/transformStreamToJsonl.ts
+++ b/packages/shared/src/server/utils/transforms/transformStreamToJsonl.ts
@@ -1,5 +1,5 @@
 import { Transform, type TransformCallback } from "stream";
-import { stringify } from "./stringify";
+import { stringify } from "../../../utils/stringify";
 
 export function transformStreamToJsonl(): Transform {
   return new Transform({
--- a/packages/shared/src/utils/stringify.test.ts
+++ b/packages/shared/src/utils/stringify.test.ts
@@ -1,8 +1,5 @@
 import { describe, it, expect } from "vitest";
-import {
-  stringify,
-  stringifyForCsv,
-} from "../../../packages/shared/src/server/utils/transforms/stringify";
+import { stringify, stringifyForCsv } from "./stringify";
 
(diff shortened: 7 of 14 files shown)


toyayuto:
One question on release lines before this drops off the radar.

This targets `main` (v4), but the affected code is also present on the `v3` maintenance branch — `downloadLegacyTraceAsJson`, the four `JSON.stringify` call sites in `useLogViewDownload`, and the same pattern in `ChatMessage.tsx` are all unchanged there.

The path that is broken is the **default** one, and on v3 there is no user-level way out of it. `LANGFUSE_MIGRATION_V4_WRITE_MODE` defaults to `legacy`, and in that mode `auth.ts` forces `v4BetaEnabled: false` for every user regardless of the value stored on the user row, while `canToggleV4` resolves to `false` so the beta toggle is never rendered. Reaching the fixed server route from #14667 therefore takes a deployment-level change — `LANGFUSE_MIGRATION_V4_WRITE_MODE=dual` **plus** `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true` — rather than a user preference. So every self-hosted v3 deployment on default settings goes through `downloadLegacyTraceAsJson`.

We have a self-hosted user on v3.221.1 (which includes #14667) hitting exactly this: downloads still contain literal `\uXXXX`, because #14667 only covered the server route (`/api/traces/[traceId]/download`) that the beta path uses. The give-away in their file was the legacy `{"trace": ..., "observations": ...}` shape, which only the client-side path produces.

Would you be open to having this on the v3 line as well? I checked the commit against `v3.224.4`: it cherry-picks with **no conflicts**, all 14 files apply cleanly, and no raw `JSON.stringify` is left in the affected files.

Happy to open the backport PR against `v3` myself, following the existing convention (`fix(trace): decode unicode escapes in client-side trace downloads and copies (v3 backport of #15698)`), or to leave it to you if you prefer to keep maintenance-line PRs in-house. Just let me know which you'd prefer.

bezbac:
@claude review

bezbac:
Hi @toyayuto, thanks for looking into this and for the detailed write-up!

I agree this makes sense.

Regarding the releases, let's get it merged into `main` first, and then I can cherry-pick it to the `v3` branch. From what you've described (and your cherry-pick test), it sounds like the backport should be straightforward.

bezbac:
Hi @toyayuto. I cherry picked the changes to v3 (https://github.com/langfuse/langfuse/pull/15801) and triggered the publishing of a new release: https://github.com/langfuse/langfuse/releases/tag/v3.225.1

toyayuto:
Thank you for the quick turnaround on the cherry-pick and release! Confirmed v3.225.1 contains the fix. Much appreciated 🙏

PR #15707, opened 2026-08-03
fix(otel): reject spans with missing or malformed ids with a 400

Steffen911:
🟡 **Live preview: [pr-15707.preview.langfuse.com](https://pr-15707.preview.langfuse.com)**

The OTLP traces endpoint accepted exports whose spans carry no usable traceId/spanId. Two shapes reached production, both from OTLP *logs* exports pointed at the traces endpoint: ResourceLogs and ResourceSpans share protobuf field numbers, so protobufjs decodes log records into spans that keep a valid instrumentation scope but lose their ids.

- absent ids: OtelIngestionProcessor.parseId calls Buffer.from(undefined) and throws ERR_INVALID_ARG_TYPE. processToIngestionEvents swallows it but processToEvent rethrows, so the queue job fails and retries all six attempts before the payload is dropped anyway. Each bad payload emitted ~12 error log lines and repeatedly tripped the OTEL Spans Processing Errors monitor.
- truncated ids: decode yields a 1- or 3-byte id, which converted silently and created a bogus trace.

Validate both at the endpoint, before the S3 upload and enqueue, and return 400. OTLP clients treat 400 as non-retryable, so this also stops the exporter-side retry loop. Rejecting the batch is not a regression for mixed batches: today a single bad span already fails the whole file in the worker.

Strings that are not hex are passed through unchanged. Hex is the only OTLP/JSON-compliant encoding, but other encodings ingest successfully today and rejecting them would break working clients for a problem we have not observed.

Emits langfuse.ingestion.otel.rejected_invalid_span_ids to size the blast radius per reason.

The PR adds request-side validation for missing and incorrectly sized OTLP span identifiers.
- Adds shared validation helpers and rejection-reason sampling.
- Rejects invalid trace exports with HTTP 400 before upload and queueing.
- Records rejection metrics and diagnostic metadata.
- Adds unit coverage for supported ID representations and malformed lengths.
</details>

The malformed nested-field crash and acceptance of zero-valued OTLP IDs should be fixed before merging.

Unvalidated JSON collection shapes can make the new validator return a server error, while length-only checks still allow reserved zero IDs to reach persistence and collide.

**Files Needing Attention:** packages/shared/src/server/otel/utils.ts, packages/shared/src/server/otel/validateOtelSpanIds.test.ts
</details>

```mermaid
flowchart TD
  A[OTLP traces request] --> B[Decode protobuf or parse JSON]
  B --> C[Validate traceId and spanId]
  C -->|Invalid| D[Record metric and return 400]
  C -->|Valid| E[Upload and enqueue]
  E --> F[OTEL ingestion worker]
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/otel/utils.ts:105-106
**Malformed collections trigger server errors**

When an OTLP/JSON request supplies a non-array, non-null `scopeSpans` or `spans` value, the nullish fallback preserves that value and `for...of` throws, causing the endpoint to return HTTP 500 instead of rejecting the malformed payload with HTTP 400.

### Issue 2
packages/shared/src/server/otel/utils.ts:57-63
**Zero-valued IDs remain accepted**

When a span contains a correctly sized all-zero trace ID or span ID, this length-only validation accepts it and ingestion persists it as the entity ID, causing unrelated traces or observations with zero IDs to collide and produce bogus merged data.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(otel): reject spans with missing or ..."](https://github.com/langfuse/langfuse/commit/3ddc240b48b13e326ac97c298c3c39e3ed26963a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49596828)</sub>


Steffen911:
@claude review

Steffen911:
@claude review

Steffen911:
@claude review

PR #15739, opened 2026-08-03
fix(ui): unify trace heatmap root aggregation

### LINKED ISSUE #15729: bug(ui): trace heatmap root aggregation diverges across four implementations
## Summary

The "root total cost / duration" values used for heatmap colour scaling in the trace detail view are computed in **four separate places, with four different formulas**. The same trace therefore gets different heatmap colours depending on which view renders it.

Two existing `TODO` comments already ask for this logic to be extracted to a shared utility, but they describe it as duplication only — as far as I can tell the fact that the copies *disagree* isn't recorded anywhere.

Observed on `main` @ `ae84d6a39` (v4.3.1).

## Affected code

| # | Location | Duration formula |
|---|---|---|
| 1 | `web/src/components/trace/lib/tree-building.ts:554-563` | `Math.max` over roots; `r.latency ? r.latency*1000 : r.endTime ? endTime-startTime : 0` |
| 2 | `web/src/components/trace/components/TraceTree.tsx:126-130` | `Math.max` over roots; `r.latency != null ? r.latency*1000 : 0` |
| 3 | `web/src/components/trace/lib/helpers.ts:489` | single root only; `tree.latency ? tree.latency*1000 : undefined` |
| 4 | `web/src/components/trace/components/TraceTimeline/index.tsx:462` | `traceDuration * 1000` (whole-trace duration) |

`rootTotalCost` has three variants too — `helpers.ts:481-488` recursively sums child costs via `calculateTreeNodeTotalCost`, while `tree-building.ts:549` and `TraceTree.tsx:118` read the pre-aggregated `r.totalCost` flat.

The two existing TODOs:

```
web/src/components/trace/lib/tree-building.ts:547
// TODO: Extract aggregation logic to shared utility - duplicated in TraceTree.tsx and TraceTimeline/index.tsx

web/src/components/trace/components/TraceTree.tsx:116
// TODO: Extract aggregation logic to shared utility - duplicated in tree-building.ts and TraceTimeline/index.tsx
```

## Concrete divergences

**a) Missing `endTime` fallback.** For a root observation with `latency == null` but a valid `endTime`, `tree-building.ts` derives duration from `endTime - startTime`; `TraceTree.tsx` yields `0`. A trace whose roots lack `latency` scales its heatmap against `0` in the tree view and against the real wall-clock span elsewhere.

**b) `latency === 0` handled inconsistently.** `tree-building.ts` uses truthiness (`r.latency ? …`), so a genuine zero-latency root falls through to the `endTime` branch. `TraceTree.tsx` uses `r.latency != null`, so zero is taken as a real value. Same input, different denominator.

**c) Max-of-roots vs. single root.** `helpers.ts` never applies `Math.max` across roots. On v4 events-based traces — which routinely have multiple roots — it scales against one root's duration while the other paths scale against the largest.

## Why this matters

The heatmap is how users visually locate slow and expensive spans, so an inconsistent denominator means the colour encoding is not comparable between the tree, the timeline, and the preview.

This class of bug has already shipped once. From `TraceTimeline/index.tsx:459-462`:

```js
// MILLISECONDS: TimelineBar heat-maps ownDurationMs against this max, and
// the tree path (TraceTree rootTotalDuration) is ms too. traceDuration is
// seconds — passing it raw inflated the heat ratio ×1000, painting every
// duration label dark red.
```

That is exactly the failure mode four independent copies invite.

## Suggested fix

Extract a single `computeRootAggregates(roots)` helper (returning `{ totalCost, totalDurationMs }`) into `web/src/components/trace/lib/`, and have all four call sites use it. The behaviour in `tree-building.ts` looks like the most complete implementation and is the natural canonical choice — it is the only variant handling both the `endTime` fallback and multiple roots.

There is a strong safety net for this: `web/src/components/trace/lib/tree-building.clienttest.ts` is ~2,350 lines and already pins the expected aggregation behaviour (e.g. `parentTotalDuration` at line 208, the `Cost Aggregation` suites at lines 390/533/705). The two divergent copies in `TraceTree.tsx` and `helpers.ts` currently have no aggregation coverage of their own, so the refactor would bring them under that existing suite.

## Suggested labels

`bug`, `ui-bug`, `feat-tracing`, `single-trace-ui`, `tech-debt`, `size:M`

---

Happy to put up a PR for this if the approach above sounds right — in particular I'd like to confirm that `tree-building.ts`'s formula is the intended canonical behaviour before standardising on it.

GautamSharma99:
## Summary

- centralize trace heatmap root cost and duration aggregation in `computeRootAggregates`
- use the same root denominator in the tree, timeline, search list, and legacy tree builder
- preserve explicit zero latency, fall back to root wall-clock duration only when latency is absent, and take the maximum across v4 roots
- sum the pre-aggregated root costs without recursively double-counting descendants

Closes #15729.

## Impacted package

- `web`

## Verification

- `pnpm --filter web exec dotenv -e '../.env.dev.example' -- vitest run --project client 'tree-building.clienttest.ts'`
  - `Tests  82 passed | 23 skipped (105)`
- `pnpm exec turbo run lint --output-logs=errors-only --log-order=stream`
- `pnpm --filter web exec dotenv -e '../.env.dev.example' -- tsc -p 'tsconfig.json' --noEmit --skipLibCheck --incremental --tsBuildInfoFile '.tsbuildinfo'`
  - exited successfully
- `pnpm exec prettier --check` on all changed files
  - `All matched files use Prettier code style!`

## Browser review

Browser verification was not run because the local seed doctor found Postgres, ClickHouse, Redis, blob storage, and the web app stopped. The changed behavior is covered by the focused client regression test and this patch does not alter layout or styling.

Centralizes trace heatmap aggregation in `computeRootAggregates`.
- Sums pre-aggregated root costs without recursively counting descendants twice.
- Uses the maximum root latency, with a wall-clock fallback only when latency is absent.
- Applies the shared cost and duration denominators across the tree, timeline, search list, and legacy tree builder.
- Adds regression coverage for multiple roots, explicit zero latency, wall-clock fallback, and cost summation.
</details>

The PR appears safe to merge, with no actionable correctness or security issues identified.

The shared helper preserves pre-aggregated root costs, explicit zero latency, and millisecond duration semantics across the migrated consumers, with focused regression coverage for the key aggregation behavior.
</details>

```mermaid
flowchart LR
  A[TreeNode roots] --> B[computeRootAggregates]
  B --> C[Sum root totalCost]
  B --> D[Max root duration]
  D --> E[Explicit latency in ms]
  D --> F[Wall-clock fallback]
  C --> G[Shared heatmap denominator]
  E --> G
  F --> G
  G --> H[Trace tree]
  G --> I[Timeline]
  G --> J[Search list]
  G --> K[Legacy builder]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ui): unify trace heatmap root aggreg..."](https://github.com/langfuse/langfuse/commit/d514789da8f0f4fc45e184f3e2ed37ef2e31c58c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49758153)</sub>

### DIFF
--- a/web/src/components/trace/components/TraceTimeline/index.tsx
+++ b/web/src/components/trace/components/TraceTimeline/index.tsx
@@ -59,6 +59,7 @@ import { useDesktopLayoutContextOptional } from "../_layout/TraceLayoutDesktop";
 import { useMobileLayoutContextOptional } from "../_layout/TraceLayoutMobile";
 import { type TreeNode } from "../../lib/types";
 import { cn } from "@/src/utils/tailwind";
+import { computeRootAggregates } from "../../lib/trace-aggregation";
 
 // Width of the left name gutter. Resizable; these bound it. Kept slim so the
 // waterfall (the point of the timeline) gets the central space.
@@ -445,21 +446,8 @@ export function TraceTimeline() {
     }
   }, []);
 
-  // Parent totals for heatmap coloring (aggregate across all roots).
-  const parentTotalCost = useMemo(() => {
-    return roots.reduce(
-      (acc, r) => {
-        if (!r.totalCost) return acc;
-        return acc ? acc.plus(r.totalCost) : r.totalCost;
-      },
-      undefined as (typeof roots)[0]["totalCost"],
-    );
-  }, [roots]);
-  // MILLISECONDS: TimelineBar heat-maps ownDurationMs against this max, and
-  // the tree path (TraceTree rootTotalDuration) is ms too. traceDuration is
-  // seconds — passing it raw inflated the heat ratio ×1000, painting every
-  // duration label dark red.
-  const parentTotalDuration = traceDuration * 1000;
+  const { totalCost: parentTotalCost, totalDurationMs: parentTotalDuration } =
+    useMemo(() => computeRootAggregates(roots), [roots]);
 
   // Score lookup: one pass over the scores instead of an O(scores) filter per
   // row per render. Two maps preserve the exact TRACE-vs-observation keying:
--- a/web/src/components/trace/components/TraceTree.tsx
+++ b/web/src/components/trace/components/TraceTree.tsx
@@ -25,6 +25,7 @@ import { cn } from "@/src/utils/tailwind";
 import type Decimal from "decimal.js";
 import { usePostHogClientCapture } from "@/src/features/posthog-analytics/usePostHogClientCapture";
 import { useTraceAnalyticsDimensions } from "../hooks/useTraceAnalyticsDimensions";
+import { computeRootAggregates } from "../lib/trace-aggregation";
 
 /**
  * Feature-scoped row container: subscribes to the row's OWN playback-active
@@ -113,22 +114,8 @@ export function TraceTree() {
     if (id) mobileLayout?.switchToInfoTab();
   };
 
-  // TODO: Extract aggregation logic to shared utility - duplicated in tree-building.ts and TraceTimeline/index.tsx
-  // Calculate aggregated totals across all roots for heatmap color scaling
-  const rootTotalCost = roots.reduce(
-    (acc, r) => {
-      if (!r.totalCost) return acc;
-      return acc ? acc.plus(r.totalCost) : r.totalCost;
-    },
-    undefined as (typeof roots)[0]["totalCost"],
-  );
-
-  const rootTotalDuration =
-    roots.length > 0
-      ? Math.max(
-          ...roots.map((r) => (r.latency != null ? r.latency * 1000 : 0)),
-        )
-      : undefined;
+  const { totalCost: rootTotalCost, totalDurationMs: rootTotalDuration } =
+    computeRootAggregates(roots);
 
   return (
     <VirtualizedTree
--- a/web/src/components/trace/lib/helpers.ts
+++ b/web/src/components/trace/lib/helpers.ts
@@ -9,6 +9,7 @@ import {
   type TraceDomain,
 } from "@langfuse/shared";
 import { type WithStringifiedMetadata } from "@/src/utils/clientSideDomainTypes";
+import { computeRootAggregates } from "./trace-aggregation";
 
 export function nestObservations(
   list: ObservationReturnType[],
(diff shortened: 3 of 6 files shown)


Diwak4r:
Rebased onto latest `main`. Upstream had relocated `web/src/components/trace/` to `web/src/features/traces/` (`lib/` -> `fns/`), so I resolved the rebase against the new layout: kept the `computeRootAggregates` change (already present in `treeBuilding.ts` via rename detection), fixed the three import paths, and dropped the now-deleted standalone `helpers.ts`. CI should re-run — requesting re-review.

Diwak4r:
Correction to my earlier note: this branch lives on @GautamSharma99's fork, so I can't push a rebase to it directly — but I worked through the resolution locally and it's a clean fix.

Root cause of the conflict: since this PR branched, langfuse/main relocated the trace UI from web/src/components/trace/ to web/src/features/traces/ (lib/ -> fns/, and components/TraceTimeline/index.tsx -> components/TraceTimeline/TraceTimeline.tsx). That's what makes the PR unmergeable, not a logic problem.

Good news: the actual fix (root-level cost/duration aggregation for the heatmap) already applies cleanly via git rename detection and is preserved in treeBuilding.ts (computeRootAggregates([tree]) in buildTraceUiData).

To resolve, update the stale import paths and drop the deleted file:
1. web/src/features/traces/components/TraceTree.tsx - import { computeRootAggregates } from ../lib/trace-aggregation -> from ../fns/traceAggregation (and useTraceAnalyticsDimensions to the @/src/features/traces/hooks/... alias).
2. web/src/features/traces/fns/traceAggregation.ts - add import { type Decimal } from decimal.js; (used by RootAggregates); keep TreeNode from ../types/treeNode.
3. web/src/features/traces/fns/treeBuilding.clienttest.ts - ./tree-building -> ./treeBuilding; getSubtreeDurationOverflowMs comes from @/src/features/traces/fns/getSubtreeDurationOverflowMs (not ./helpers, which was deleted upstream); TreeNode from ../types/treeNode; computeRootAggregates from ./trace-aggregation.
4. Delete the standalone web/src/components/trace/lib/helpers.ts - its content was merged into treeBuilding.ts upstream.

I can share the exact resolved patch if that's helpful.

PR #15741, opened 2026-08-03
test(clickhouse): cover filter SQL translation

### LINKED ISSUE #15731: test(clickhouse): clickhouse-filter.ts has no direct unit coverage
## Summary

`packages/shared/src/server/queries/clickhouse-sql/clickhouse-filter.ts` is 729 lines and sits on the path of essentially every v4 events query — tables, dashboards, and the search bar all lower their `FilterState` through it. It has no colocated unit test.

Observed on `main` @ `ae84d6a39` (v4.3.1).

## Current coverage in `clickhouse-sql/`

| File | LOC | Colocated test |
|---|---|---|
| `event-query-builder.ts` | 2150 | ✅ `event-query-builder.test.ts` |
| **`clickhouse-filter.ts`** | **729** | ❌ **none** |
| `query-fragments.ts` | 609 | ❌ none |
| `event-filter-options.ts` | 430 | ❌ none |
| `events-observation-row-selection.ts` | 347 | ❌ none |
| `factory.ts` | 307 | ❌ none |
| `fts.ts` | 201 | ❌ none |
| `search.ts` | 128 | ❌ none |
| `orderby-factory.ts` | 110 | ❌ none |
| `events-stream-query.ts` | 99 | ✅ `events-stream-query.test.ts` |
| `score-filter-pushdown.ts` | 83 | ✅ `score-filter-pushdown.test.ts` |

`clickhouse-filter.ts` is exercised only indirectly, via `packages/shared/src/server/repositories/traces.test.ts` and integration tests that reach it through a full query.

## Why this matters

- **It generates SQL.** Filter-to-SQL translation is where operator handling, parameter binding, and escaping live. Bugs here are either wrong results (silent) or injection (severe), and both are cheap to catch with table-driven unit tests over `(FilterState) → (sql, params)`.
- **It is on the hot path of the v4 migration.** The events tables are the active workstream right now, so this file is being modified under time pressure, which is exactly when a fast unit suite pays for itself.
- **Indirect coverage misattributes failures.** When an integration test fails today, it isn't immediately clear whether the fault is in the filter lowering, the query builder, or the repository. Direct tests localise that.

## Proposed approach

Add `clickhouse-filter.test.ts` alongside the existing `event-query-builder.test.ts`, following the same table-driven style already used there and in `score-filter-pushdown.test.ts`. Target the pure translation surface — no ClickHouse connection needed, so it runs in the fast unit path:

- every operator across each column type (string, stringOptions, number, datetime, boolean, arrayOptions, map/metadata)
- parameter naming and binding, including collision behaviour when the same column is filtered twice
- escaping of values containing quotes, backslashes, and `\uXXXX` sequences
- null / empty-array / empty-string handling per operator
- `AND`/`OR` composition and the empty-`FilterList` case

I'd start with the operator × column-type matrix, since that's the highest-value block and establishes the fixture shape for the rest.

## Suggested labels

`tech-debt`, `[Core] ClickHouse database`, `size:M`

---

Happy to submit this. If maintainers would rather see the suite scoped down first (e.g. operators only, deferring escaping and composition to a follow-up), say the word and I'll split it.

GautamSharma99:
## What does this PR do?

Adds direct, colocated unit coverage for `packages/shared/src/server/queries/clickhouse-sql/clickhouse-filter.ts`.

The new suite exercises the pure filter-to-SQL translation surface without starting ClickHouse or writing fixture data. It uses deterministic parameter suffixes so both generated SQL and parameter collision behavior can be asserted precisely.

### Coverage added

| Area | Covered behavior |
| --- | --- |
| String filters | Every standard operator, bound special characters, empty-string/null equivalence, and v4 `matches` validation/lowering |
| Scalar filters | Every number, datetime, and boolean operator, including numeric type overrides |
| Option filters | Every string, category, and array operator, empty arrays, flattened categories, and empty/null semantics |
| Metadata/object filters | Every string-map, number-object, and boolean-object operator; events metadata name/value arrays; ngram escaping; indexed metadata matching |
| Null filters | Both operators with ordinary and empty-string-equivalent behavior |
| Composition | Empty `FilterList`, `AND` composition, and collision-free binding when the same field is filtered twice |
| Table routing | `events_full` requirements for v4 input/output/metadata filters and core-safe negative cases |

The operator tables are derived from `filterOperators` and use exhaustive `satisfies Record<...>` mappings where applicable. Adding an operator now requires the direct SQL contract to be updated instead of relying on a distant repository failure.

This is test-only: it does not change production SQL, query structure, indexes, or ClickHouse schemas. No seed scenario is needed because the tested surface is deterministic and has no database dependency.

## Impacted package

- `@langfuse/shared`

## How was it tested?

- `pnpm --filter @langfuse/shared run test 'clickhouse-filter.test.ts'`
  - `Test Files 1 passed (1)`
  - `Tests 62 passed (62)`
- `pnpm run lint`
- `pnpm run typecheck`
- `git diff --check`

Closes #15731

Adds direct unit coverage for ClickHouse filter-to-SQL translation.

- Covers string, scalar, option, object, metadata, and null filter operators.
- Verifies parameter binding, collision-free parameter names, and empty/null semantics.
- Tests filter composition and `events_full` table-routing requirements.
</details>

The PR appears safe to merge.

The change adds deterministic unit tests without modifying production query behavior, schemas, or runtime paths, and no actionable defects were identified in the new coverage.
</details>

<sub>Reviews (1): Last reviewed commit: ["test(clickhouse): cover filter SQL trans..."](https://github.com/langfuse/langfuse/commit/be1d91020e13f07910302a113eec684e9d6ab8d1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49764835)</sub>

### DIFF
--- a/packages/shared/src/server/queries/clickhouse-sql/clickhouse-filter.test.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/clickhouse-filter.test.ts
@@ -1,0 +1,556 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  FTS_MATCH_OPERATOR,
+  filterOperators,
+} from "../../../interfaces/filters";
+
+const randomState = vi.hoisted(() => ({ next: 1 }));
+
+vi.mock("../../repositories", () => ({
+  clickhouseCompliantRandomCharacters: () => `x${randomState.next++}`,
+}));
+
+import {
+  ArrayOptionsFilter,
+  BooleanFilter,
+  BooleanObjectFilter,
+  CategoryOptionsFilter,
+  DateTimeFilter,
+  encodeBooleanScoreEntry,
+  FilterList,
+  filtersRequireEventsFull,
+  NullFilter,
+  NumberFilter,
+  NumberObjectFilter,
+  StringFilter,
+  StringObjectFilter,
+  StringOptionsFilter,
+} from "./clickhouse-filter";
+
+describe("clickhouse filters", () => {
+  beforeEach(() => {
+    randomState.next = 1;
+  });
+
+  describe("StringFilter", () => {
+    const queryByOperator = {
+      "=": "o.name = {stringFilterx1: String}",
+      contains: "position(o.name, {stringFilterx1: String}) > 0",
+      "does not contain": "position(o.name, {stringFilterx1: String}) = 0",
+      "starts with": "startsWith(o.name, {stringFilterx1: String})",
+      "ends with": "endsWith(o.name, {stringFilterx1: String})",
+    } satisfies Record<(typeof filterOperators.string)[number], string>;
+
+    it.each(filterOperators.string)("lowers the %s operator", (operator) => {
+      expect(
+        new StringFilter({
+          clickhouseTable: "observations",
+          tablePrefix: "o",
+          field: "name",
+          operator,
+          value: "needle",
+        }).apply(),
+      ).toEqual({
+        query: queryByOperator[operator],
+        params: { stringFilterx1: "needle" },
+      });
+    });
+
+    it("keeps quotes, backslashes, and unicode escape text in bound params", () => {
+      const value = String.raw`quote'\\slash\u1234`;
+      const result = new StringFilter({
+        clickhouseTable: "observations",
+        field: "name",
+        operator: "=",
+        value,
+      }).apply();
+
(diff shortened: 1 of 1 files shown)


sumerman:
See https://github.com/langfuse/langfuse/issues/15731#issuecomment-5176724228

PR #15768, opened 2026-08-04
fix(table-views): demote active system preset when user edits its filters (LFE-14699)

nkabardin:
🟡 **Live preview: [pr-15768.preview.langfuse.com](https://pr-15768.preview.langfuse.com)**

## TL;DR

On the v4 events table, applying a preset chip (e.g. Slow → "Latency over 10s") and then deleting its filter left the preset armed in session storage: the next visit to Traces/Observations silently re-applied the filter the user had just removed, and the chip stayed "active" (invisibly so in dark mode). Now any user filter edit fully deselects the active system preset, and the active chip/dropdown states are actually visible.

## Why

Applying a system preset writes its `viewId` to both the URL and session storage (so it survives navigation). Removing the preset's filter — via the search bar or the facet sidebar — only cleared the filter layer. Nothing deactivated the view, so on the next clean-URL mount the view manager's session-storage restore re-fetched the preset and resurrected the deleted filter. The only escape was explicitly deselecting the preset in the chip dropdown. On top of that, the "active" chip styling was a 5% primary tint (near-invisible in dark mode) and the dropdown computed the active row but never styled it.

## What

- `useSidebarFilterState` already tags every filter write with an origin (`user` / `saved_view` / `system`). The events table now composes an `onExplicitFilterStateChange` handler that, on a **user-origin** edit while a **system preset** is active, fully deselects it: `handleSetViewId(null, { updateType: "replaceIn" })` clears the URL param and session storage, the chip unlights immediately, and the session restore can no longer resurrect the filters.
- **Deliberately scoped to system presets.** User-saved views keep today's behavior wholesale: `appliedViewId === selectedViewId` is the load-bearing column-trust signal for the drawer's "Update view" flow, and any demotion (even session-storage-only) breaks it — silently reverting live column edits to the view's stored snapshot on the next update. Extending demotion to user-saved views first needs that signal decoupled from the session restore (noted as a follow-up below).
- Active states in `CategoryPresetChips`: the active chip now uses the filled on-state of the sibling table/chart toggle (`bg-primary text-primary-foreground`), and the active dropdown row gets an accent background plus a check indicator.

## Result

Preset applied → filter deleted → the chip deactivates at once, and navigating Traces ⇄ Observations no longer brings the filter back. Applying a view never demotes itself, user-saved views behave exactly as before, and deep-link/back-button semantics are untouched.

<details>
<summary>Mechanism, tests, and verification</summary>

### Mechanism

- The demotion decision lives in `web/src/features/events/lib/demoteViewOnUserFilterEdit.ts` (pure, unit-testable). Guards: only `origin === "user"`, only when the filters actually changed (a no-op re-commit of unchanged search-bar text does not demote), only while a **system preset** (`isSystemPresetId`) is active.
- Ordering: the filter hook is constructed before `useTableViewManager`, so the events table reaches the view controllers through a ref (same pattern as the existing `queryFilterRef`).
- History semantics: the programmatic `viewId` clear uses `replaceIn`, so the Back button does not bounce off a resurrected entry (same class as LFE-10715).
- Deliberate semantics: **any** user-origin filter edit demotes — including adding a filter on top of the preset. The chip's active state means "the table shows exactly this preset"; once the user diverges, that provenance is stale.
- Catalog guard: the changed-filters check compares with deep equality, which relies on every system preset's filters surviving the search-bar grammar round-trip (`filterStateToQueryText → planCommit`) unchanged — otherwise a preset would self-demote on a mere search-bar focus + blur. A dedicated clienttest round-trips the entire system preset catalog and asserts deep equality, so a future preset with a grammar-unstable filter shape fails CI instead of shipping the regression.
- Out of scope / follow-up: the preset's `orderBy` may linger for the page session after demotion. Extending demotion to user-saved views requires decoupling the session-restore signal from the column-trust signal first.

### Tests

- New clienttest `web/src/features/events/presetDemotionOnUserEdit.clienttest.tsx` (5 tests): documents the resurrection path (session-storage restore) as a control, then asserts full demotion + no re-apply on remount, demotion when adding a filter on top, no demotion on a no-op write, and that user-saved views are NOT demoted on user edits.
- New clienttest `web/src/features/events/systemPresetSearchBarRoundTrip.clienttest.ts` (9 tests): the catalog round-trip guard described above, iterating all system presets.
- Full web client suite green: 232 files / 2863 tests passed (includes the LFE-10715 history-writes, session-persistence, and saved-view implicit-environment regression suites).
- `tsc --noEmit` and eslint (`--max-warnings 0`) clean.

### Live verification (dev server, Chrome, dark mode)

Scripted browser run against `/project/…/observations`: applied the Slow → "Latency over 10s" preset (URL + session storage confirmed), deleted the `latency:>10` token via the search bar, confirmed `viewId` removed from URL and session storage and the chip unfilled, then navigated to `/traces` and back — no filter resurrection, search bar stays empty. Dark-mode screenshots taken of the filled active chip and the checked active dropdown row.

</details>

This PR demotes active system presets when users change event-table filters and improves the visibility of active preset chips and rows.
- Composes filter-origin notifications with table-view controller state.
- Adds a pure demotion helper that clears system preset persistence with replacement history semantics.
- Adds regression and preset grammar round-trip tests.
- Adds filled active-chip styling and a checked active dropdown row.
</details>

The explicit-state deep-link case should be fixed before merging because a filter edit can silently detach the currently selected user-saved view.

The view manager intentionally permits URL and session view IDs to diverge for explicit-state links, but the new demotion helper prefers the stale session ID and then clears both identities.

**Files Needing Attention:** web/src/features/events/lib/demoteViewOnUserFilterEdit.ts
</details>

```mermaid
flowchart TD
  A[User applies system preset] --> B[viewId stored in URL and session]
  B --> C[User opens explicit-state saved-view link]
  C --> D[URL selectedViewId is user view]
  C --> E[Session appliedViewId remains old system preset]
  D --> F[User edits filters]
  E --> F
  F --> G[Demotion prefers appliedViewId]
  G --> H[handleSetViewId null clears both IDs]
  H --> I[User saved-view association is lost]
```
</details>

`````markdown
### Issue 1
web/src/features/events/lib/demoteViewOnUserFilterEdit.ts:48
**Stale applied view wins demotion**

When a tab retains a system preset in session storage and opens a user-saved-view link with explicit table state, `appliedViewId` remains the stale preset while `selectedViewId` identifies the linked user view. Preferring `appliedViewId` causes the next filter edit to call `handleSetViewId(null)`, removing the user view's URL association and breaking its drawer and Update-view workflow.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(table-views): scope demotion to syst..."](https://github.com/langfuse/langfuse/commit/439629068d4ac2a678bc60e4998c3449e9eace4a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50040812)</sub>


nkabardin:
@claude review

nkabardin:
@claude review

PR #15770, opened 2026-08-04
fix(sentry): stop three noise families from minting a fingerprint per occurrence

nkabardin:
🟡 **Live preview: [pr-15770.preview.langfuse.com](https://pr-15770.preview.langfuse.com)**

## TL;DR

Three browser-noise families mint a **new Sentry fingerprint per occurrence** (~40 junk issues/week on the board). This adds four narrow, unit-tested rules to the established filter module (`sentryFilters.ts`, pattern from #15145/#15238): two drops keyed on exact signatures that cannot represent an app bug, one **fingerprint grouping** (kept, not dropped) for stale-chunk parse errors, and one drop for errors thrown wholly inside PostHog's session-replay recorder script. Every rule was designed from pulled production events and ships positive + negative fixtures proving real errors still pass.

## Why

- **Non-Error promise rejections** (`Non-Error promise rejection captured with value: …`): no stack, so Sentry groups by the stringified value — every new value is a new issue (LANGFUSE-5T9, 5TA). Sources: browser-extension shims, `Promise.reject(undefined)`, email-scanner artifacts.
- **Storage-access `SecurityError`** (`Failed to read the 'localStorage' property from 'Window': Access is denied…`): iframe/privacy-mode environments; our `useLocalStorage`/`useSessionStorage` hooks already catch and fall back, but the `console.error` in the catch re-captures the DOMException with a per-deploy hashed-chunk stack — four fingerprints in 14 days for one message (LANGFUSE-5TC/5TD/5TE/5TF).
- **Chunk parse `SyntaxError`s**: a stale client fetching a chunk that no longer exists after a deploy parses garbage; chunk filenames are content-hashed, so it's **one fingerprint per chunk per deploy** (LANGFUSE-5WH 15 ev, 5S7 7, 5WG 5, 5WD 3, 5WN + a long 1-event tail). The reload banner (#15279) mitigates the cause; the events themselves are non-actionable.
- **PostHog recorder internals**: rrweb DOM serialization throwing inside `/static/posthog-recorder.js` (LANGFUSE-5VY 21 ev, 5VX 19) — vendor code, no app frame in the stack.

## What

| Family | Mechanism | Signature (all conditions required) |
|---|---|---|
| Non-Error rejections | **drop** | type `UnhandledRejection` + SDK-synthesized prefix + value in an **exact denylist** (`Not implemented on this platform`, `undefined`, `Object Not Found Matching Id:…`) |
| Storage SecurityError | **drop** | type `SecurityError` + mechanism `auto.core.capture_console` (already caught by our hooks) + browser-generated getter-denial message |
| Chunk parse errors | **group** | mechanism `onerror` + type `SyntaxError` + exactly one function-less frame in `/_next/static/chunks/` → single fingerprint `stale-chunk-parse-error` |
| Recorder internals | **drop** | every attributable frame in `/static/posthog-recorder.js` (any app frame ⇒ kept) |

**What still passes (negative fixtures in `sentryFilters.clienttest.ts`):** unknown rejection values and the object-shaped rejection variant; real unhandled rejections carrying an `Error`; an **uncaught** storage `SecurityError` crashing a render (global `onerror`); cross-origin-frame `SecurityError`s; the evals editor's user-code `SyntaxError`s (`/evals/new`, `capture_console` + app frames); runtime `JSON.parse` errors; any stack mixing recorder frames with app frames.

## Result

The three families stop minting fingerprints: two families disappear as known-benign environment noise, chunk-parse errors collapse into **one** issue that would still spike visibly if a deploy ever shipped a genuinely broken chunk, and recorder-internal errors are gone. 82/82 filter tests green; lint + typecheck green.

<details>
<summary>Engineering detail: evidence, skeptic review, and deliberate limits</summary>

### Designed from wire-shape events, not stored JSON
An independent adversarial review of the first draft caught that `beforeSend` sees the **client wire shape**, not the normalized stored event:

- Parse-error frames carry the SDK's `UNKNOWN_FUNCTION` placeholder `"?"` (`_enhanceEventWithInitialFrame`); stored events show `null` only after server-side normalization. The chunk-parse predicate accepts both.
- The recorder script loads as `/static/posthog-recorder.js?v=<posthog-js version>` and the query survives into frame filenames; the predicate strips query/fragment before suffix-matching.

### Masking-risk analysis (per rule)
- **Rejections:** exact-value matching only. The codebase has no `Promise.reject(<non-Error>)`/bare `reject()` sites; a rejected real `Error` keeps its own type and never matches. The object-shaped variant (`Object captured as promise rejection with keys: …`) has carried real failures and is explicitly kept. `undefined` rejections carry zero diagnostic content (no stack, no value) — nothing actionable is lost.
- **Storage SecurityError:** the review found unguarded storage reads (e.g. a `sessionStorage.getItem` in a render-path `useMemo`) that could crash a page in storage-blocked contexts. The rule is therefore **mechanism-guarded**: only `capture_console` instances (already caught + fallen back by our hooks) are dropped; an uncaught crash arrives via `onerror` and is kept.
- **Chunk parse:** grouped, not dropped, precisely so a bad deploy stays visible as a spike on one issue. Recommended follow-up: keep the grouped issue on "archive until escalating", never archive-forever.
- **Recorder:** any stack touching an app chunk is kept. Deliberate coverage gap: the `addEventListener`-wrapped variant carries the Sentry `wrap()` frame, which in a prod bundle is an app-chunk filename — those events remain (dropping chunk-frame stacks could mask real listener errors).

### Test coverage
82 tests: each rule has fixtures replicating the pulled production event shapes (wire + normalized variants) and KEEPS-fixtures for every adjacent real-error shape listed above.

</details>

The PR narrows browser-side Sentry noise without changing application behavior.
- Drops selected non-Error promise rejections and caught storage-access denials.
- Groups anonymous Next.js chunk parse failures under one stable fingerprint.
- Drops errors attributable solely to PostHog’s session-replay recorder.
- Adds positive and adjacent negative fixtures for each filtering rule.
</details>

The PR appears safe to merge because the new filters are narrowly constrained and no concrete reachable loss of actionable telemetry was identified.

The filtering rules retain adjacent app-error shapes through mechanism, exception-type, stack-attribution, and frame-shape guards, while stale chunk failures remain visible under a stable fingerprint.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(sentry): harden filters against wire..."](https://github.com/langfuse/langfuse/commit/55aa86b358a0fc0a9ce4c9ae7341cf848dcee2f8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50043069)</sub>

### DIFF
--- a/web/instrumentation-client.ts
+++ b/web/instrumentation-client.ts
@@ -58,7 +58,7 @@ Sentry.init({
 
     // Stale-deploy / truncated chunk parse errors: collapse into ONE issue
     // instead of one per content-hashed chunk filename. Deliberately grouped,
-    // NOT dropped — a deploy that ships a genuinely unparseable chunk still
+    // NOT dropped — a deploy that ships a genuinely unparsable chunk still
     // surfaces as a spike on the single grouped issue. See
     // isStaleChunkParseErrorEvent for the rationale.
     if (isStaleChunkParseErrorEvent(event)) {
--- a/web/instrumentation-client.ts
+++ b/web/instrumentation-client.ts
@@ -2,7 +2,10 @@ import * as Sentry from "@sentry/nextjs";
 import {
   isDenylistedNoiseEvent,
   isNoisyHttpClientPollEvent,
+  isPosthogRecorderInternalEvent,
   isReactDevtoolsInternalEvent,
+  isStaleChunkParseErrorEvent,
+  STALE_CHUNK_PARSE_FINGERPRINT,
 } from "@/src/utils/sentryFilters";
 
 const isEuOrUsRegionNonHipaa =
@@ -43,6 +46,23 @@ Sentry.init({
     // to Sentry. See isDenylistedNoiseEvent for the per-rule rationale.
     if (isDenylistedNoiseEvent(event)) {
       return null;
+    }
+
+    // Drop errors thrown wholly inside PostHog's session-replay recorder
+    // script (rrweb DOM serialization failing on exotic page content). No app
+    // frame is ever present in these stacks; anything touching our code is
+    // kept. See isPosthogRecorderInternalEvent for the rationale.
+    if (isPosthogRecorderInternalEvent(event)) {
+      return null;
+    }
+
+    // Stale-deploy / truncated chunk parse errors: collapse into ONE issue
+    // instead of one per content-hashed chunk filename. Deliberately grouped,
+    // NOT dropped — a deploy that ships a genuinely unparseable chunk still
+    // surfaces as a spike on the single grouped issue. See
+    // isStaleChunkParseErrorEvent for the rationale.
+    if (isStaleChunkParseErrorEvent(event)) {
+      event.fingerprint = [STALE_CHUNK_PARSE_FINGERPRINT];
     }
 
     return event;
--- a/web/src/utils/sentryFilters.clienttest.ts
+++ b/web/src/utils/sentryFilters.clienttest.ts
@@ -3,7 +3,9 @@ import { type ErrorEvent } from "@sentry/nextjs";
 import {
   isDenylistedNoiseEvent,
   isNoisyHttpClientPollEvent,
+  isPosthogRecorderInternalEvent,
   isReactDevtoolsInternalEvent,
+  isStaleChunkParseErrorEvent,
 } from "@/src/utils/sentryFilters";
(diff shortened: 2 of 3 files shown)


nkabardin:
@claude review

PR #15772, opened 2026-08-04
fix(web): classify tRPC response JSON-parse failures as transport noise

nkabardin:
🟡 **Live preview: [pr-15772.preview.langfuse.com](https://pr-15772.preview.langfuse.com)**

## TL;DR

`TRPCClientError: JSON.parse: unexpected character…` / `Unexpected end of JSON input` events are transport noise (a non-JSON response body reached the tRPC client), not app bugs. This classifies them at the existing `handleTrpcError` seam: breadcrumb instead of Sentry capture, error toast unchanged. Negative fixtures prove real errors still capture.

## Why

Sentry issues LANGFUSE-5V0 (12 events / 6 users since 2026-07-24) and LANGFUSE-5VS (2 events / 1 user): the tRPC client throws a `TRPCClientError` wrapping a raw `SyntaxError` from `JSON.parse` on the response body. Evidence this is transport-shaped, not an app bug:

- The events carry **no tRPC error envelope** (`error.data` is absent — no `trpc.code`/`trpc.path` tags): the server never answered with a tRPC error shape.
- In the captured replay breadcrumbs, the failing fetch returned **HTTP 200 with an unparseable body** — something between the server and the `JSON.parse` call replaced or truncated the body.
- The 5V0 events all come from Firefox sessions belonging to security researchers (bug-bounty / disposable-mail accounts) probing auth and settings pages — the signature of an intercepting proxy rewriting responses. 5VS is two events 11 ms apart from one user — a truncated body on a flaky connection.

Our tRPC handler always serializes JSON, so a body that fails `JSON.parse` was corrupted in transit. The client-side event is an amplified, lower-fidelity copy of a signal the server already owns.

## What

- New named predicate `isTrpcResponseParseError` in `web/src/utils/api.ts`: a `TRPCClientError` whose `cause` is a `SyntaxError` **and** that has no parsed tRPC error envelope (`error.data`) **and** whose response status is not 414/431 — a parse failure on a request-too-large response is the app-owned oversized-GET-URL bug class (the `sendAsPostOption` machinery) and keeps capturing.
- Because `@trpc/client` drops the response meta when body parsing fails (verified against 11.13.4: the link records meta only for fulfilled requests), a small `fetch` wrapper on the links re-attaches the HTTP status to the parse `SyntaxError` so the 414/431 exclusion actually works in the live flow.
- `handleTrpcError` (the established seam from #15243) drops the transport-shaped ones from Sentry with a breadcrumb (message + response status), exactly like the expected-code suppression next to it.
- The parse-error toast copy in `trpcErrorToast.tsx` no longer claims "We've been notified" (no longer true for the suppressed class); it now reads "The request could not be completed. Please try again or contact support if this persists."

## Result

The two JSON-parse issue families stop minting Sentry events; a breadcrumb keeps the trail in any later real event from the same session.

**Does this rule hide a real error?** No — locked by negative fixtures in `api.clienttest.ts`:

- a tRPC server error envelope (e.g. `INTERNAL_SERVER_ERROR`, 500) still captures (`error.data` present → predicate is false),
- a parse failure on HTTP 414/431 (request too large — an actionable app bug) still captures, via the status annotation or link meta,
- a raw `SyntaxError` thrown by app code (not wrapped by the tRPC client) still captures via the non-TRPC branch,
- network connectivity failures and other causes are untouched.

A systemic outage where a proxy serves error pages for every response surfaces server-side (5xx/LB metrics), not through browser JSON.parse copies.

<details>
<summary>Mechanism and verification detail</summary>

- `TRPCClientError.from(cause)` (trpc client v11) wraps a thrown `SyntaxError` as `new TRPCClientError(cause.message, { cause })`, leaving `shape`/`data` undefined — verified against the installed `@trpc/client@11.13.4`. A server error response goes through the `isTRPCErrorResponse` branch instead and always carries `result.error.data`.
- The predicate reads `error.data` first, so any response that parsed into an error envelope can never be classified as a parse failure, regardless of cause.
- **Why the fetch wrapper:** empirically probed the installed client — for a 431 empty body, a 200 HTML body, and a 502 HTML body alike, the thrown `TRPCClientError` has `meta: undefined` (the link's `meta` closure is assigned in the fulfilled branch only, `httpLink` → `.catch((cause) => TRPCClientError.from(cause, { meta }))`). A status check on `error.meta.response` alone would therefore never fire in the live flow; the wrapper re-attaches the status at the only place that still has both the `Response` and the parse error. The predicate also reads `error.meta.response.status` as a fallback for links that do attach meta.
- Suppression follows the sentry-instrumentation contract: named, documented predicate + breadcrumb trail + negative fixtures (the #15243 seam-classification pattern, same file).
- Checks: `vitest --project client src/utils/api.clienttest.ts` (28 tests green, incl. wrapper tests), prettier, eslint, `tsc --noEmit` all green.

</details>

### DIFF
--- a/web/src/utils/api.clienttest.ts
+++ b/web/src/utils/api.clienttest.ts
@@ -1,12 +1,24 @@
 import { TRPCClientError } from "@trpc/client";
+import { vi } from "vitest";
 import {
   EXPECTED_TRPC_ERROR_CODES,
+  fetchWithParseErrorStatus,
   getTrpcErrorCode,
   getTrpcErrorPath,
   isExpectedTrpcClientError,
   isNetworkConnectivityError,
   isTrpcResponseParseError,
 } from "@/src/utils/api";
+
+/** A JSON.parse SyntaxError annotated with the HTTP status it was parsed
+ * from, as produced by `fetchWithParseErrorStatus`. */
+const parseErrorWithStatus = (message: string, status: number) => {
+  const cause: SyntaxError & { responseStatus?: number } = new SyntaxError(
+    message,
+  );
+  cause.responseStatus = status;
+  return cause;
+};
 
 /** Builds a TRPCClientError with the given server error shape (code/path). */
 const trpcServerError = (opts: {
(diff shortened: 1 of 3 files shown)


nkabardin:
@claude review

PR #15789, opened 2026-08-04
fix(worker): guard blob-export Parquet against Arrow 2GiB column-chunk overflow

Steffen911:
🟡 **Live preview: [pr-15789.preview.langfuse.com](https://pr-15789.preview.langfuse.com)**

## Problem

Parquet blob-storage exports on EU fail repeatedly with `ECONNRESET`; the corresponding ClickHouse queries die with:

```
std::exception. Code: 1001, type: parquet::ParquetStatusException,
e.what() = Capacity error: array cannot contain more than 2147483646 bytes, have 2149071376
```

thrown from `parquet::DictEncoderImpl<ByteArrayType>::PutByteArray` inside `DB::Parquet::writeColumnChunkBody`. Because the failure happens mid-stream on a binary `FORMAT Parquet` HTTP response, ClickHouse can only close the socket — the worker sees a bare `ECONNRESET`. A failed table export never advances `lastSyncAt`, so the affected integration retries the same window forever.

## Root cause

Two ClickHouse behaviors compose (verified against `ParquetBlockOutputFormat.cpp` / `Parquet/Write.cpp`):

1. **The row-group bytes cap cannot split a single fat chunk.** `ParquetBlockOutputFormat::consume` splits oversized staging by *rows* only; `output_format_parquet_row_group_size_bytes` merely stops further staging. The export queries' dedup sort (`ORDER BY event_ts DESC LIMIT 1 BY …`) and `FINAL` stages emit blocks bounded by `max_block_size` rows (~65k), not bytes — at multi-MB `input`/`output` payloads one chunk is multi-GiB, and it becomes a single row group despite the 128 MiB cap.
2. **Size checks are batch-granular.** The writer checks page size and the 1 MiB dictionary-fallback threshold only once per `output_format_parquet_batch_size` (default 1024) values. The affected project has ~17 MiB spans; ~125 of them inside one 1024-value batch push Arrow's dictionary memo table past 2^31 bytes before the fallback can fire.

## Fix

Tighten `BLOB_EXPORT_PARQUET_CLICKHOUSE_SETTINGS` (only consumed by the four blob-export Parquet query builders; Parquet is a per-project opt-in file type):

- `max_block_size: 8192` — keeps source chunks small so the existing 128 MiB row-group bytes cap actually binds.
- `output_format_parquet_batch_size: 16` — caps the page/dictionary size-check overshoot at 16 rows' worth of bytes (~272 MiB even at 17 MiB values).
- `output_format_parquet_max_dictionary_size: 0` — disables dictionary encoding, removing the Arrow dictionary memo-table ceiling as a failure class. Near-unique LLM payloads fell back to plain encoding anyway (dictionary ≈ data); zstd output compression recovers most of the size difference on low-cardinality columns.

Exported data is unchanged; consumers may see `PLAIN` instead of `RLE_DICTIONARY` encoding on low-cardinality columns and modestly larger files. Settings ride along per-query — no server config change, trivially revertible. The wedged EU integration self-heals on its next run since `lastSyncAt` never advanced.

## Impacted packages

- `@langfuse/shared` (constant consumed by `worker` blob-storage exports)

## Verification

- `pnpm --filter @langfuse/shared run typecheck` — clean
- `pnpm --filter @langfuse/shared run lint` — clean (pre-commit hook: `Tasks: 7 successful, 7 total`)
- `pnpm --filter worker run test src/__tests__/clickhouseExecRawParquet.integration.test.ts` against live ClickHouse — `Tests  2 passed (2)` (new settings accepted, valid Parquet bytes returned, mid-stream failure still surfaces as an error)

Tightens the ClickHouse settings used by blob-storage Parquet exports to prevent Arrow’s 2 GiB column-chunk overflow.
- Reduces source block size so the existing row-group byte cap can bind.
- Runs Parquet size checks more frequently with a smaller write batch.
- Disables dictionary encoding to remove the dictionary memo-table ceiling.
</details>

The PR appears safe to merge, with the new settings confined to the four blob-storage Parquet export paths.

The settings directly constrain the source-block and writer-batch conditions behind the reported Arrow overflow, while the investigated callers consistently request Parquet output and no concrete compatibility or correctness regression remains.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(worker): guard blob-export Parquet a..."](https://github.com/langfuse/langfuse/commit/a945ee5a7c4b3bc05098b5579eb0919eb4cee7b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50127181)</sub>


Steffen911:
@claude review

PR #15791, opened 2026-08-04
fix: trim whitespace from email in auth forms and on the server

### LINKED ISSUE #15780: bug: cannot sign in when email has leading/trailing whitespace (email is never trimmed)
### Describe the bug

Signing in with email/password fails when the email contains leading or trailing whitespace (e.g. from copy-paste, or a mobile keyboard adding a trailing space). The email is never trimmed, so the form rejects it with "Invalid email address" even though the address itself is correct — the whitespace is invisible, which makes this hard to diagnose as a user.

**Expected:** leading/trailing whitespace in the email field is trimmed before validation and lookup, as in most login forms.

**Actual:** the form shows "Invalid email address" under the Email field and never submits.

The same gap exists on sign-up, password reset, and the enterprise-SSO-required page.

### Steps to reproduce

1. Go to `/auth/sign-in`
2. In the Email field, enter a valid registered address with one trailing space, e.g. `demo@langfuse.com ` — easiest via paste from a source that includes the space. In a local dev setup the seeded user is `demo@langfuse.com` / `password`.
3. Enter the correct password
4. Click "Sign in" (or "Continue" if SSO is configured)

Result: "Invalid email address" appears under the Email field and the form does not submit. Deleting the invisible trailing space makes sign-in succeed.

Self-hosted

Reproduced against `v4.3.1` (latest release at time of filing) and `main` @ `3bca62f` — the relevant code is identical in both.

Not applicable — web UI / auth.

#### Root cause

1. **The email input is not `type="email"`**, so the browser's [value sanitization algorithm](https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)) — which strips leading and trailing whitespace for `type=email` — never runs. `web/src/pages/auth/sign-in.tsx` passes no `type`, and `web/src/components/ui/input.tsx` forwards `type` as-is, so the field renders as `type="text"`.

2. **The form schema does not trim** — `web/src/pages/auth/sign-in.tsx:59`:

   ```ts
   const credentialAuthForm = z.object({
     email: z.email(),
     password: z.string().min(8, { ... }),
   });
   ```

   With the pinned zod version (`zod: ^4.3.6` in `web/package.json`), `z.email()` rejects any leading or trailing whitespace:

   ```
   z.email().safeParse("user@example.com ")                          -> invalid, "Invalid email address"
   z.email().safeParse(" user@example.com")                          -> invalid
   z.string().trim().pipe(z.email()).safeParse("user@example.com ")  -> valid
   ```

3. **The two-step SSO flow validates the raw value too** — `sign-in.tsx:725`: `z.email().safeParse(credentialsForm.getValues("email"))`.

4. **The server does not trim either** — `web/src/server/auth.ts:103`: `where: { email: credentials.email.toLowerCase() }`. There is a `toLowerCase()` but no `trim()`, so a request that bypasses client-side validation fails the user lookup and returns "Invalid credentials".

#### Same gap elsewhere in the auth surface

All use `z.email()` with no trim, and no email input in these pages sets `type="email"`:

- `web/src/pages/auth/sign-up.tsx:49`, `:134`
- `web/src/features/auth/lib/signupSchema.ts:35`
- `web/src/pages/auth/enterprise-sso-required.tsx:24`
- `web/src/features/auth-credentials/components/ResetPasswordPage.tsx:33`
- `web/src/features/auth-credentials/components/ResetPasswordButton.tsx:27`
- `web/src/features/auth-credentials/lib/credentialsServerUtils.ts:30`, `:43` (`email.toLowerCase()` without `trim()`)

This looks like an oversight rather than a deliberate decision: `ResetPasswordButton.tsx:65` already does `email.toLowerCase().trim()`, but the guard on `:27` validates the *untrimmed* value and `handleResetPassword` returns early when it fails — so that existing trim is unreachable for whitespace input. Email is also already trimmed elsewhere in the app (#499 trimmed it in the sharing dialog), and `auth.ts` trims domains coming from env config.

#### Note for whoever fixes this

`auth.ts:87` derives the SSO-enforcement domain from the untrimmed email:

```ts
const domain = credentials.email.split("@")[1]?.toLowerCase();
```

For `user@corp.com ` this yields `"corp.com "`, which matches neither `AUTH_DOMAINS_WITH_SSO_ENFORCEMENT` nor a multi-tenant SSO config. Today that is harmless, because the user lookup on `:103` is equally untrimmed and fails anyway. But trimming only the lookup without also trimming the domain extraction would turn this into an SSO-enforcement bypass. The fix should normalize once, up front, and cover both call sites.

#### Proposed fix

Happy to open a PR. Rough shape:

- Trim in the zod schemas — `z.string().trim().pipe(z.email())`, ideally as one shared `emailSchema`, used by sign-in, sign-up, `signupSchema`, reset-password and enterprise-sso-required.
- Add `type="email"` to the email inputs, so the browser sanitizes the value too.
- Normalize once at the top of `authorize()` in `web/src/server/auth.ts` and use that value for both the domain checks and the user lookup; same for `createUserEmailPassword()`.
- Add tests for leading space, trailing space and tab on sign-in, sign-up and password reset — there is currently no test coverage for email whitespace normalization.

One open question before I start: would you prefer a single shared `emailSchema` in `packages/shared`, or keeping the normalization local to each form?

Yes

prithidevghosh:
## What does this PR do?

Signing in with email/password fails when the email has a leading or trailing space (from a paste, or a mobile keyboard that appends one): the form rejects it with "Invalid email address" even though the address is correct. The whitespace is invisible, so it's hard to diagnose. The same gap exists on sign-up, password reset, and the enterprise-SSO-required page.

Fixes langfuse/langfuse#15780

### Cause

* The email inputs are not `type="email"`, so the browser's value-sanitization (which strips surrounding whitespace for `type=email`) never runs — they render as `type="text"`.
* The form schemas use bare `z.email()`, which rejects any leading/trailing whitespace with no `.trim()` first:

  ```
  z.email().safeParse("user@example.com ")                         -> invalid
  z.string().trim().pipe(z.email()).safeParse("user@example.com ") -> valid
  ```
* The server doesn't trim either. `authorize()` and `createUserEmailPassword()` `.toLowerCase()` but never `.trim()`, so a request that bypasses the client fails the user lookup.

### Fix

* **Shared schema.** New `web/src/features/auth/lib/emailSchema.ts` exports `emailSchema` (`z.string().trim().pipe(z.email())`) and a `normalizeEmail(email)` helper (trim + lowercase). Adopted in sign-in, sign-up, `signupSchema`, password-reset and enterprise-SSO, including the two-step SSO handlers that validate `getValues("email")` directly.
* `type="email"` **on the email inputs**, so the browser strips whitespace as a second layer.
* **Server normalization, once, up front.** `authorize()` computes `normalizeEmail(credentials.email)` once and uses it for both the SSO-enforcement domain extraction and the user lookup; `createUserEmailPassword()` does the same for the duplicate check and the write.

### Why normalize once (security)

`authorize()` derives the SSO-enforcement domain from the email (`email.split("@")[1]`). Trimming only the user lookup would let `user@corp.com ` skip domain-based SSO enforcement while still matching the account — a bypass. Normalizing up front keeps the domain check and the lookup reading the same value, so this can't drift.

### Scope note

The NextAuth `signIn` callback (`auth.ts`, another bare `z.email()`) is intentionally left unchanged: it only ever sees provider-supplied OAuth emails or the already-normalized value returned by `authorize()`, never user-typed whitespace.

### Design choice

The shared schema lives in `web/src/features/auth/lib/` (next to `signupSchema.ts`) rather than `packages/shared`, since it's only consumed by web auth today. Happy to move it if a maintainer prefers.

## Type of change

## Mandatory Tasks

- [X] Make sure you have self-reviewed the code.

## Tests

The failing test was written first and confirmed red before the fix:

* `web/src/__tests__/server/unit/email-schema.servertest.ts` (new): `emailSchema` trims leading/trailing/tab whitespace, preserves case, still rejects non-emails; `normalizeEmail` trims and lowercases.
* `web/src/__tests__/server/signup-schema-unicode-name.servertest.ts`: added a case asserting `signupSchema` trims surrounding whitespace from the email (was failing before the schema change).
* `web/src/__tests__/server/credentials-server-utils.servertest.ts` (new): `createUserEmailPassword` stores a trimmed, lowercased email, and treats a whitespace-padded duplicate as the same existing user.

Local runs: `typecheck` and `lint` both `7 successful, 7 total`; the three test files `22 passed (22)`.

Also verified end to end in a browser: on `/auth/sign-in`, entering `demo@langfuse.com ` (trailing space) plus the correct password now signs in and lands on the app, where before it showed "Invalid email address" and never submitted.

The PR consistently trims user-entered email addresses across credential authentication forms and server-side credential lookup and creation, while lowercasing database keys. It also keeps SSO domain enforcement and user lookup on the same normalized value.

* Adds a shared email schema and normalization helper.
* Applies trimmed email validation and native email input types across sign-in, sign-up, password-reset, and enterprise SSO forms.
* Normalizes credential sign-in and account creation before SSO checks, duplicate checks, lookup, and persistence.
* Adds focused schema and credential-creation tests.

</details>

The PR appears safe to merge with no actionable changed-code defects identified.

Email normalization is applied consistently at the form and credential-server boundaries, and the SSO domain check and database operations consume the same normalized identity.

</details>

Reviews (1): Last reviewed commit: ["fix: trim whitespace from email in auth ..."](<https://github.com/langfuse/langfuse/commit/6a4e02c9a1c3e3e0f00361266524ba10df2994d6>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=50207095>)

**Context used:**

* Knowledge Base — [Authentication, Multi-Tenancy, and RBAC](<https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/auth-rbac.md>)

### DIFF
--- a/web/src/__tests__/server/credentials-server-utils.servertest.ts
+++ b/web/src/__tests__/server/credentials-server-utils.servertest.ts
@@ -1,0 +1,49 @@
+import { prisma } from "@langfuse/shared/src/db";
+import { createUserEmailPassword } from "@/src/features/auth-credentials/lib/credentialsServerUtils";
+import { v4 } from "uuid";
+
+describe("createUserEmailPassword", () => {
+  const createdUserIds: string[] = [];
+
+  afterAll(async () => {
+    if (createdUserIds.length === 0) return;
+    // Memberships are only created when default orgs/projects are configured;
+    // clear them defensively so the user row can be deleted regardless.
+    await prisma.organizationMembership.deleteMany({
+      where: { userId: { in: createdUserIds } },
+    });
+    await prisma.user.deleteMany({ where: { id: { in: createdUserIds } } });
+  });
+
+  it("trims and lowercases the email before storing it (#15780)", async () => {
+    const unique = v4().slice(0, 8);
+    const rawEmail = `  Whitespace.${unique}@Example.com  `;
+    const expected = `whitespace.${unique}@example.com`;
+
+    const userId = await createUserEmailPassword(
+      rawEmail,
+      "password123",
+      "Test User",
+    );
+    createdUserIds.push(userId);
+
+    const user = await prisma.user.findUnique({ where: { id: userId } });
+    expect(user?.email).toBe(expected);
+  });
+
+  it("treats a whitespace-padded duplicate as the same existing user (#15780)", async () => {
+    const unique = v4().slice(0, 8);
+    const email = `dup.${unique}@example.com`;
+
+    const userId = await createUserEmailPassword(
+      email,
+      "password123",
+      "Test User",
+    );
+    createdUserIds.push(userId);
+
+    await expect(
+      createUserEmailPassword(`  ${email}  `, "password123", "Test User"),
+    ).rejects.toThrow(/already exists/i);
+  });
+});
--- a/web/src/__tests__/server/signup-schema-unicode-name.servertest.ts
+++ b/web/src/__tests__/server/signup-schema-unicode-name.servertest.ts
@@ -95,6 +95,25 @@ describe("signupSchema name validation", () => {
     expect(result.success).toBe(false);
   });
 
+  it("trims surrounding whitespace from the email before validating (#15780)", () => {
+    for (const email of [
+      "test@example.com ",
+      " test@example.com",
+      "\ttest@example.com\t",
+    ]) {
+      const result = signupSchema.safeParse({
+        ...validBaseInput,
+        name: "André",
+        email,
+      });
+
+      expect(result.success).toBe(true);
+      if (result.success) {
+        expect(result.data.email).toBe("test@example.com");
+      }
+    }
+  });
+
   it("rejects names with disallowed punctuation", () => {
     const result = signupSchema.safeParse({
       ...validBaseInput,
--- a/web/src/__tests__/server/unit/email-schema.servertest.ts
+++ b/web/src/__tests__/server/unit/email-schema.servertest.ts
@@ -1,0 +1,50 @@
+import {
+  emailSchema,
+  normalizeEmail,
+} from "@/src/features/auth/lib/emailSchema";
+
+describe("emailSchema", () => {
+  it("trims leading, trailing and tab whitespace before validating (#15780)", () => {
+    for (const input of [
+      "user@example.com ",
+      " user@example.com",
+      "\tuser@example.com\t",
+      "  user@example.com  ",
+    ]) {
+      const result = emailSchema.safeParse(input);
+      expect(result.success).toBe(true);
+      if (result.success) {
+        expect(result.data).toBe("user@example.com");
+      }
+    }
+  });
+
+  it("accepts a clean email unchanged", () => {
+    const result = emailSchema.safeParse("user@example.com");
+    expect(result.success).toBe(true);
+    if (result.success) expect(result.data).toBe("user@example.com");
+  });
+
+  it("preserves case (the server lowercases on lookup)", () => {
+    const result = emailSchema.safeParse("User@Example.com ");
+    expect(result.success).toBe(true);
+    if (result.success) expect(result.data).toBe("User@Example.com");
+  });
+
+  it("still rejects a value that is not an email", () => {
+    for (const input of ["not-an-email ", "   ", "user@", "@example.com"]) {
+      expect(emailSchema.safeParse(input).success).toBe(false);
+    }
+  });
+});
+
+describe("normalizeEmail", () => {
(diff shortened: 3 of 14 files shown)


prithidevghosh:
@nkabardin requesting to initiate a review on this pr.

nkabardin:
@claude review

prithidevghosh:
@nkabardin please run the review again, pushed a pr, addressing the earlier review.

PR #15808, opened 2026-08-05
feat(in-app-agent): recover background runs without a browser attached (LFE-14624)

LysanderKie:
🟡 **Live preview: [pr-15808.preview.langfuse.com](https://pr-15808.preview.langfuse.com)**

## Why

Background agent runs are durable in Postgres but delivered one-shot (`attempts: 1`, `maxStalledCount: 0`), because a redelivered run can re-execute an approved mutation. That trade only holds if something else notices abandoned work, and today nothing does: `reconcileConversationRuns` is conversation-scoped and read-triggered, so a run whose worker died is only recovered while a browser is looking at it. That is exactly what background execution is meant to stop requiring, and schedule- or monitor-triggered agents have no browser at all.

The worst case is not a stuck row. The worker mints a project-scoped MCP credential per run; on SIGKILL the only retry path is snapshot hydration, so if nobody reopens the conversation, a live project credential exists forever.

GA blocker for LFE-14375.

## What

**A dedicated `in-app-agent-lifecycle-queue`**, its own worker at concurrency 1, behind `QUEUE_CONSUMER_IN_APP_AGENT_LIFECYCLE_QUEUE_IS_ENABLED`.

- **Lifecycle recovery, every 5s.** Redispatches a QUEUED run stranded between the dispatch delay and the queue timeout; terminalizes `queue_timeout`, `worker_lost` and `run_timeout`; and revokes the credential of every run it terminalizes. Bounded at 50 + 50 per tick.
- **Credential maintenance, hourly.** A pure backstop for what neither the worker's `onFinish` nor the sweep saw: a run terminalized by the read path, or a key whose run row was cascade-deleted with its project.

Selection, classification and CAS live in one batch entry point in `runLifecycle.ts`, sharing `classifyStaleRun` and a single `applyStaleRunTerminalCas` with the conversation read path, so the two cannot disagree. Only age-eligible rows are selected, so a healthy system costs two statements per tick and zero writes.

**Its own queue, not a job on the execution queue.** That queue runs 15-minute jobs at concurrency 5, so recovery scheduled there would wait behind exactly the saturation it exists to resolve; BullMQ priority orders the waiting list without preempting. Its own consumer flag so it can be killed during an incident without stopping every background run. This follows every existing cron-style queue here (`dlqRetryQueue`, `eventPropagationQueue`, `dataRetentionQueue`).

**The SSE watch becomes a pure reader.** Its 5s reconcile is removed with no attach-time replacement, and terminal credential cleanup moves off snapshot hydration. Reconciliation keeps two read-path owners that need synchronous truth (hydration, `createQueuedRun`) plus the sweep.

**Delivery fix.** `removeOnFail: true` on the execution queue, plus clearing a terminal job before redispatching. BullMQ `add()` against a retained job ID is a silent no-op, so a worker that threw before its claim CAS left the run QUEUED with a poisoned deterministic ID and recovery quietly never fired.

## Why credential maintenance is hourly, not per-minute

The first version swept every 60s. Neither of its queries can use an index — `finished_at IS NOT NULL AND mcp_api_key_id IS NOT NULL` is the complement of the partial active-run index, and `api_keys` has none on `is_in_app_agent_key`. Both plans are sequential scans, and `in_app_agent_runs` grows forever:

| Sweep | Cadence | Plan |
|---|---|---|
| Lifecycle recovery | 5s | Index Scan on `in_app_agent_runs_active_conversation_key` |
| Terminal pointer retry | was 60s | Seq Scan on `in_app_agent_runs` |
| Orphan credentials | was 60s | Seq Scan on `api_keys` |

The frequent sweep was the cheap one and the infrequent one was the full scan, 1440 times a day to almost always find nothing. The cadence also was not buying what it looked like: the secret is minted in the worker, handed to an in-process MCP client, and never transmitted anywhere but our own MCP endpoint, so a key outliving its run is an unowned credential rather than one an attacker holds.

Revoking on terminalization inverts it. The sweep already holds the row it just failed, so it revokes that key in the same pass at no extra scan, which makes the case that actually matters — a worker killed mid-run, where only the pointer knows the key exists — *faster* than before (one sweep tick instead of up to a minute), while the scanning job drops to 24 passes a day.

## Two things reviewers should push on

**Watch load is not a reason to merge this.** The ticket lists it; it does not survive the numbers. Per attached watch this removes ~11% of Postgres statements (5.6/s to 5.0/s). The remaining 89% is the interactive transaction around the 1 Hz tail, and collapsing that into a single statement (same atomicity, since one statement takes one snapshot even at Read Committed) is the real reduction. That belongs to LFE-14629 and should be tried before Redis Pub/Sub.

**Crash window B is still open.** Redispatch covers a crash between commit and enqueue, and a throw before the claim CAS. It does not cover a crash *after* job pickup but before the claim: `lockDuration` is `RUN_MAX_DURATION + 60s`, so the orphaned job holds its lock for 16 minutes, `add()` dedupes against it, and `remove()` cannot evict a locked job. The run dies at `queue_timeout` exactly as today. Closing it means shortening `lockDuration` and enabling `maxStalledCount: 1`, which changes delivery for every run and deserves its own ticket. Note the `attempts: 1` argument does not defend it: a job that never claimed cannot have executed anything.

## Notes

- No migration. The partial index `in_app_agent_runs_active_conversation_key` covers the candidate query's `finished_at IS NULL`, but carries neither `status` nor `created_at`, so every candidate needs a heap fetch. Free at tens of active runs, not free at thousands, and thousands is reachable while there is still no per-org concurrent-run ceiling. If that ceiling slips past GA, add a partial `(created_at) WHERE finished_at IS NULL` index rather than inherit the assumption.
- The foreground carve-out in the candidate query is transitional. Once foreground execution is removed after GA, no claim-less RUNNING row can exist and both the guard and the classifier's foreground branch become dead code (LFE-14376).
- `AWAITING_APPROVAL` expiry is deliberately not swept: `decideToolApproval` already enforces the 24h TTL in-transaction, so a late decision cannot execute. The residue is a stale pending badge.
- Dashboards, monitors and capacity validation stay with LFE-14629.

## Impacted packages

`@langfuse/shared`, `worker`, `web`.

## Verification

```
pnpm run lint                                              Tasks: 7 successful, 7 total
pnpm run typecheck                                         Tasks: 7 successful, 7 total
pnpm --filter worker run test src/features/in-app-agent    Tests  25 passed (25)
pnpm --filter @langfuse/shared run test in-app-agent       Tests  13 passed (13)
pnpm --filter web run test in-app-agent                    Tests  133 passed (133)
```

**Tests added: 12.** New: `worker/src/features/in-app-agent/lifecycleRecovery.test.ts` (5, real DB: heartbeat-stale terminalization vs healthy run untouched, credential revoked on terminalization, foreground exclusion, redispatch window, terminal-job clearing) and `credentialMaintenance.test.ts` (3, real DB: terminal retry, already-absent key, orphan window with referenced-key preservation). Extended: 2 fake-Prisma cases in `runLifecycle.test.ts` (mixed-batch classification issuing no per-candidate transaction, independent bounds), 1 zero-reconciliation assertion in `watch.test.ts`, 1 single-read assertion in `executeInAppAgentRun.test.ts`.

No test asserts scheduler cadence, priority or queue options; that would restate static configuration. Two were confirmed to fail against the behavior they describe: the single-read test reports 2 reads when the second `getConversationEvents` is restored, and removing the `claimedAt` guard proved that guard is defence rather than the load-bearing predicate (the comment and test rationale say so).

**Not yet run:** the manual end-to-end checklist (SIGKILL recovery, dispatch gap, saturation while all five execution slots are busy).

This PR adds a dedicated BullMQ lifecycle queue to recover stranded in-app-agent runs and maintain their per-run MCP credentials without browser activity.

- Adds periodic redispatch and stale-run terminalization backed by shared lifecycle classification and CAS logic.
- Adds hourly orphan-credential maintenance and immediate credential revocation during terminalization.
- Removes reconciliation from SSE watches and credential cleanup from snapshot hydration.
- Reuses already-loaded conversation events during worker initialization and removes failed execution jobs to permit redispatch.
</details>

This PR should not merge until abandoned foreground credentials are made recoverable and the maintenance batch is guaranteed to progress past persistently referenced keys.

The new replacement lifecycle path excludes claim-less foreground runs and then treats their credential references as proof of liveness, while the maintenance query can repeatedly consume its entire fixed batch with those skipped rows and never reach later orphan credentials.

**Files Needing Attention:** worker/src/features/in-app-agent/lifecycleSweeps.ts, packages/shared/src/in-app-agent/server/runLifecycle.ts
</details>

The lifecycle replacement leaves abandoned foreground-run credentials permanently referenced and therefore skipped, and its oldest-first pre-filtered batch can let persistent references starve later orphan credentials indefinitely.
</details>

```mermaid
flowchart TD
  Run[(In-app agent run)] --> State{Run state}
  State -->|QUEUED past delay| Redispatch[Redispatch deterministic BullMQ job]
  State -->|Claimed RUNNING stale| Terminalize[CAS to FAILED]
  Terminalize --> Revoke[Revoke MCP credential]
  State -->|Terminal pointer remains| Maintenance[Hourly credential maintenance]
  Maintenance --> Revoke
  Redispatch --> Worker[Execution worker]
  Worker -->|Claim succeeds| Running[RUNNING with heartbeat]
  Worker -->|Finishes| Cleanup[Terminal state and credential cleanup]
```
</details>

`````markdown
### Issue 1
worker/src/features/in-app-agent/lifecycleSweeps.ts:109-111
**Foreground credentials remain live**

When a web process dies after a foreground run stores its MCP credential, the global lifecycle query excludes that claim-less run and this branch treats its unfinished row as proof that the credential is still in use. The abandoned run and its project-scoped credential therefore remain live indefinitely. **How this was verified:** Foreground runs have no claim timestamp, the lifecycle selector requires one for RUNNING rows, and maintenance skips every key referenced by an unfinished run.

### Issue 2
worker/src/features/in-app-agent/lifecycleSweeps.ts:89-90
**Referenced keys starve cleanup**

If at least 100 older credentials remain referenced by unfinished runs, this query selects those same rows on every hourly pass and filters them only after applying the limit. Later orphan credentials are never examined or revoked and remain active indefinitely. **How this was verified:** The query takes the oldest 100 keys before the subsequent reference check skips rows without deleting them or advancing a cursor.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into feature/lfe-146..."](https://github.com/langfuse/langfuse/commit/6cb2e646dcfd76391218a4c9bf618b6c3196e294) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50432649)</sub>

**Context used:**


LysanderKie:
Decision to now implement this yet due to the additional complexity that adds and instead track after enabling background execution if something like this is required or not

PR #15813, opened 2026-08-05
feat(ingestion): auto-direct-write OTel traffic for post-cutoff orgs

Steffen911:
🟡 **Live preview: [pr-15813.preview.langfuse.com](https://pr-15813.preview.langfuse.com)**

## Problem

Organizations created after the v4 auto-rollout boundary are force-enabled on the v4 preview and cannot switch it off, so `events_full` is the only surface they read. If their OTLP exporter does not send `x-langfuse-ingestion-version: 4`, the batch takes the dual-write path and their data surfaces with up to ~10 minutes of delay — a correctly configured plain-OTel setup looks broken on day one.

Roughly 38% of OTel batches across production regions are still on the delayed path today (prod-us 47%, prod-eu 30%, prod-hipaa 17%, prod-jp 18%, 7-day `langfuse.ingestion.otel.write_path` counts). That figure mixes header-less plain OTel with older Langfuse SDKs; only the former is in scope here.

## Approach

A **new** cutoff rather than reusing the existing April rollout boundary, so no organization's current behavior changes. Orgs that onboarded between April and now keep the dual-write shape and are handled through docs and hand-holding.

- **New optional worker env var** `LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CREATED_CUTOFF` — an ISO date (`YYYY-MM-DD`) read as midnight UTC, validated with `z.iso.date()`. Unset disables the rule, which is the correct default when self-hosting: those deployments move the whole deployment at once via `LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=direct`. Cloud-gated on `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION`.
- **Non-Langfuse-SDK exports only.** Older Langfuse SDKs keep the dual write: their trace-level attributes surface on the synthetic root observation that only the dual path materializes, so flipping them would change the shape of data those users already read. A plain OTel exporter has no such prior expectation. Detection consults both identifying signals — the `x-langfuse-sdk-name` header and the `langfuse` instrumentation scope that pre-header SDK versions still carry, checked across every scope in the batch so collector-forwarded batches that mix SDK spans with third-party scopes stay excluded.
- **Decided entirely in the worker,** alongside the other write-path signals. The SDK-header check, the deployment override and the scope check all already resolve in `otelIngestionQueue.ts`; the web route forwards raw header values and decides nothing. The cutoff follows that pattern, so there is no queue-contract change and `web/`/`packages/` are untouched.
- **Own `direct_org_cutoff` write path,** ordered last in the precedence chain (header > env > scope > org cutoff) so the metric counts only traffic this rule genuinely moves off the dual path rather than absorbing batches a header would have routed directly anyway.

### Cost

The organization lookup runs only when no higher-precedence signal has already routed the batch and it is not Langfuse SDK traffic — so header-carrying traffic never triggers it, and nothing triggers it while the cutoff is unset. Signup dates are immutable, so results are cached in-process via `LocalCache` (the pattern `modelMatch` already uses in this path), keyed on project rather than the payload's optional `orgId` so replayed jobs resolve too. A lookup failure keeps the existing path rather than failing the batch.

## Behavior note

In `dual` write mode the legacy `traces`/`observations` tables are still written for these projects, so v3 UI and v2 API reads are unaffected. What changes for an in-scope plain-OTel project is that the synthetic trace-as-observation row (`span_id = trace_id`) that the dual-write path materializes no longer exists — the real root span is the root observation. Trace-level attributes set on only one span therefore no longer lift onto a trace-wide row. This is the intended v4 semantic, and it is why the cutoff is new rather than retroactive, and why Langfuse SDK traffic is excluded.

No docs change: the header remains the correct universal instruction and is still required self-hosted and for pre-cutoff orgs.

## Rollout

The cutoff date is deliberately unset, so this merges as a no-op until a value is configured.

## Testing

- `worker` write-path, SDK-exclusion and cutoff-predicate tests — `Tests  66 passed (66)`

---

> [!NOTE]
> **Medium Risk**
> Changes OTel ingestion routing for a cloud cohort and alters observation shape for affected plain-OTel traffic, but rollout is env-gated, SDK traffic is excluded, and failures default to the existing dual path.
> 
> **Overview**
> **Cloud-only OTel routing** adds optional `LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CREATED_CUTOFF` (ISO date, unset = no-op). On Langfuse Cloud, organizations created on or after that date can have **plain OTel** OTLP batches routed to **direct** `events_full` writes even without `x-langfuse-ingestion-version: 4`, avoiding dual-write delay for v4-only readers.
> 
> Write-path selection is centralized in **`resolveOtelWritePath`** with precedence **header → env → scope → org cutoff**, emitting **`direct_org_cutoff`** only when that rule actually moves traffic. **Langfuse SDK** batches stay on dual write when only the cutoff would apply (header/scope/sdk-name and full-batch scope scan), so older SDK trace semantics are unchanged.
> 
> The cutoff path runs a **cached Postgres** lookup of org `createdAt` by **project id** only when no higher signal already routed the batch and the traffic is not SDK-identified; lookup failures fall back to the pre-cutoff path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d5ae8bbf32a324cbf9e74e9d6c75f125db322c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


web/src/env.mjs
```diff
@@ -473,6 +473,27 @@ export const env = createEnv({
...
+    // does not send `x-langfuse-ingestion-version: 4`. Applies to non-Langfuse-
+    // SDK exports only; an older SDK keeps its established dual-write shape.
+    // A plain ISO date
+    // (YYYY-MM-DD) is enough precision and is easier to reason about than a
+    // timestamp; it is interpreted as midnight UTC. Unset disables the rule,
+    // which is the correct default for self-hosted deployments: those flip the
+    // whole deployment at once via LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR
+    // instead of rolling a tenant cohort forward.
+    LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF: z
+      .string()
```
Steffen911: Why don't we use something like [https://zod.dev/api?id=iso-dates#iso-dates](<https://zod.dev/api?id=iso-dates#iso-dates>) ?

web/src/env.mjs
```diff
@@ -473,6 +473,27 @@ export const env = createEnv({
     LANGFUSE_DISABLE_LEGACY_TRACING_IO_SEARCH: z
       .enum(["true", "false"])
       .default("false"),
+    // Cloud-only cutoff (LFE-14536) for automatic direct OTel event writes.
+    // Organizations created on or after this date are past the point where the
+    // v4 preview is force-enabled and cannot be switched off, so their OTLP
+    // traffic must reach the events table in real time even when the exporter
+    // does not send `x-langfuse-ingestion-version: 4`. Applies to non-Langfuse-
+    // SDK exports only; an older SDK keeps its established dual-write shape.
+    // A plain ISO date
```
Steffen911: Do not reference internal linear tickets. Prune them from the comments. They should be fully self-sustaining.

.env.dev-azure.example
```diff
@@ -94,6 +94,15 @@ LANGFUSE_MIGRATION_V4_WRITE_MODE=dual
...
 LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=dual_write
+# Cloud-only: organizations created on or after this ISO date (YYYY-MM-DD, read
+# as midnight UTC) get direct OTel event writes without needing the
+# `x-langfuse-ingestion-version: 4` header, because they are force-enabled on v4
+# and read the events table only. Non-Langfuse-SDK exports only; an older
+# Langfuse SDK keeps its established dual-write shape.
+# Unset (the default) disables the rule;
+# self-hosted deployments flip the whole deployment via
+# LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=direct instead.
+# LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF=
```
Steffen911: Only add the new cutoff flag to .env.dev.exmaple. We can drop it from the other .env files to keep them clean. Also, don't repeat all the comment that is already explained when declaring the env var.

web/src/features/public-api/server/otelDirectWriteCutoff.ts
```diff
@@ -0,0 +1,53 @@
...
+ * shape and keeps it, while a plain OTel exporter — the case this rule exists
+ * for — has no such prior expectation. See `resolveOtelWritePath`.
+ *
+ * Deliberately date-based: there is no per-organization or per-project v4
+ * column, only `User.v4BetaEnabled`, so signup date is the only cohort signal
+ * available. It mirrors `V4_DEFAULT_ENABLED_FROM_AT`, which gates the UI-side
+ * rollout, but keys on the single organization owning the API key rather than
+ * the oldest organization a user belongs to — ingestion has no user context.
+ */
+export const isOrgPastOtelDirectWriteCutoff = (params: {
```
Steffen911: Where do we place all the routing decisions today, e.g. around the python SDKs or the header? Do they all live in a file in web/src/features/public-api/server ? I would keep them closer to the otel processing logic. Can't we make the whole decision on the worker side in the otelIngestionQueue.ts flow?

Steffen911: Good catch — you were right, and I moved the whole decision into `otelIngestionQueue.ts` in 1ce7de1.

To answer the question directly: every other write-path signal is already decided in the worker.

| Signal | Decided in | Input |
|---|---|---|
| SDK header / ingestion-version | worker, `checkHeaderBasedDirectWrite` | raw header values off the queue payload |
| Deployment override | worker, `v4ForceDirectOtelWrite(env)` | worker env |
| Instrumentation scope | worker, `checkSdkVersionRequirements` | parsed spans |

The web route only forwards raw header strings (`sdkName`, `sdkVersion`, `ingestionVersion`) into the payload — it interprets nothing and decides nothing. So resolving the org cutoff in web and shipping a boolean verdict was the odd one out. It now sits with its siblings, and the env var moved to `worker/src/env.ts` next to `LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR`.

Consequences, all good ones:

- **The queue payload flag is gone.** No `forceDirectWrite` field, so no contract change and no rolling-deploy compatibility concern — and no need to name a flag whose meaning depended on the consumer.
- **`web/` and `packages/` are now untouched.** No `ApiAccessScope` change, no API-key cache change. The diff went from 13 files to 4.
- **It is cheaper.** The old design put `orgCreatedAt` on the shared auth scope, so it was resolved on every authenticated API request across the product to serve one decision on one route. Now the Postgres lookup runs only when no higher-precedence signal has already routed the batch and it is not Langfuse SDK traffic — so header-carrying traffic never triggers it, and nothing triggers it while the cutoff is unset. Signup dates are immutable, so the result is cached in-process with `LocalCache`, the same pattern `modelMatch` already uses in this path. Keyed on project rather than the payload's optional `orgId` so replayed jobs resolve too; a lookup failure keeps the existing path rather than failing the batch.
- It also removed a real bug — see the `types.ts` thread.

web/src/env.mjs
Steffen911: Done — switched to `z.iso.date()`. It is both less code and stricter than what I had; verified against the edge cases the bots raised in the sibling threads:

```
2026-08-06   ACCEPT
2026-02-30   reject
2026-04-31   reject
2025-02-29   reject
2024-02-29   ACCEPT
2026-13-01   reject
2026-8-6     reject
```

The declaration now lives in `worker/src/env.ts` since the decision moved worker-side.

web/src/env.mjs
Steffen911: Done. `grep -rn LFE-14536` is now empty across the tree, and the comments describe the behaviour and rationale on their own terms.

.env.dev-azure.example
Steffen911: Done — the flag is only in `.env.dev.example` now, removed from the azure/oci/redis-cluster copies, and trimmed to two lines that point at the declaration instead of repeating it:

```
# Cloud-only rollout boundary (ISO date) for automatic direct OTel writes.
# Unset disables it; see worker/src/env.ts for the full semantics.
# LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF=
```

packages/shared/src/server/auth/types.ts
```diff
@@ -22,6 +22,13 @@ const ApiKeyBaseSchema = z.object({
   // these columns existed
   createdByUserId: z.string().nullish(),
   createdByApiKeyId: z.string().nullish(),
+  // Signup date of the owning organization, used by the Cloud OTel
+  // direct-write cutoff (LFE-14536). Nullish so cache entries written before
+  // this field existed stay parseable during a rolling deploy; consumers must
+  // treat a missing value as "org age unknown" and fall back to the
+  // pre-cutoff behaviour. Entries refresh within
+  // LANGFUSE_CACHE_API_KEY_TTL_SECONDS.
+  orgCreatedAt: z.iso.datetime().nullish(),
```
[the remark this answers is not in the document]
Steffen911: Valid, and worse than described — thanks. I had claimed in the PR body that stale entries "self-heal within the 300s TTL"; that was wrong.

`fetchApiKeyFromRedis` reads with `getex(key, "EX", TTL)`, which **refreshes the TTL on every read**. An actively used API key's cache entry therefore never expires and never re-reads Postgres, so `orgCreatedAt` would have stayed absent indefinitely for exactly the busy keys this rule targets — silently keeping them on the delayed dual-write path. Fixing it in place would have needed a cache-key version bump or an invalidation sweep.

Resolved by construction instead: the decision moved into the worker (see the `otelDirectWriteCutoff.ts` thread), so there is no `orgCreatedAt` on the cache entry or the auth scope at all. `packages/shared` is untouched in the current diff. The worker reads the signup date directly with its own in-process cache.

web/src/env.mjs
```diff
@@ -473,6 +473,25 @@ export const env = createEnv({
...
+    // whole deployment at once via LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR
+    // instead of rolling a tenant cohort forward.
+    LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF: z
+      .string()
+      .regex(/^\d{4}-\d{2}-\d{2}$/, "Expected an ISO date (YYYY-MM-DD)")
+      .refine(
+        (value) => !isNaN(new Date(`${value}T00:00:00.000Z`).getTime()),
+        "Expected a valid calendar date",
+      )
+      .optional(),
```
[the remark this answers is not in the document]
Steffen911: Valid — `new Date("2026-02-30")` rolls forward rather than returning `Invalid Date`, so the refine did not do what its message claimed.

Fixed by using `z.iso.date()` (per @Steffen911's suggestion in the sibling thread) rather than the round-trip refine, since it is stricter and less code. Verified:

```
2026-08-06   ACCEPT
2026-02-30   reject
2026-04-31   reject
2025-02-29   reject
2024-02-29   ACCEPT
2026-13-01   reject
2026-8-6     reject
```

The declaration moved to `worker/src/env.ts` as part of moving the decision worker-side.

web/src/env.mjs
```diff
@@ -473,6 +473,25 @@ export const env = createEnv({
...
+
```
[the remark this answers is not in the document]
Steffen911: Confirmed and fixed — the reasoning was correct, including the point that the same faulty parse was reused at the comparison site, not just at startup validation.

Now `z.iso.date()`, which rejects `2026-02-30`, `2026-04-31` and `2025-02-29` while accepting `2024-02-29`. Since env validation rejects impossible dates at startup, the runtime `Date.parse` in the comparison can no longer receive one; it keeps an `isNaN` guard that falls back to the pre-cutoff behaviour rather than guessing.

Both the env var and the comparison moved to the worker in 1ce7de1.

Steffen911:
## Review round 1 addressed — the decision moved to the worker (`1ce7de1`)

The architectural comment was right, so this is a rewrite rather than a patch. **`web/` and `packages/` are now untouched** — the diff is 4 files, all worker-local plus one env-example line.

### Why worker-side

Every other write-path signal is already decided in the consumer:

| Signal | Decided in | Input |
|---|---|---|
| SDK header / ingestion-version | worker, `checkHeaderBasedDirectWrite` | raw header values off the queue payload |
| Deployment override | worker, `v4ForceDirectOtelWrite(env)` | worker env |
| Instrumentation scope | worker, `checkSdkVersionRequirements` | parsed spans |

The web route forwards raw header strings and decides nothing. Resolving the org cutoff in web and shipping a boolean verdict was the outlier. It now sits with its siblings, and the env var moved to `worker/src/env.ts` beside `LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR`.

This removed two problems instead of working around them:

1. **A latent bug** (greptile P1, confirmed). `fetchApiKeyFromRedis` reads with `getex(key, "EX", TTL)`, which refreshes the TTL on every read — so an actively used key's cache entry never expires and never re-reads Postgres. An `orgCreatedAt` field on that entry would have stayed absent indefinitely for precisely the busy keys this rule targets. My "self-heals within 300s" claim in the original PR body was wrong.
2. **The flag naming problem**, by deletion. No queue payload field, so no contract change and no rolling-deploy compatibility concern.

It is also cheaper: the old design resolved the signup date on every authenticated API request across the product. The lookup now runs only when no higher-precedence signal has already routed the batch and it is not Langfuse SDK traffic — so header-carrying traffic never triggers it, and nothing does while the cutoff is unset. Signup dates are immutable, so the result is cached in-process via `LocalCache` (the pattern `modelMatch` already uses here), keyed on project so replayed jobs without the optional `orgId` resolve too. A lookup failure keeps the existing path rather than failing the batch.

### Other comments

- `z.iso.date()` replaces the hand-rolled refine — stricter and less code. Rejects `2026-02-30`, `2026-04-31`, `2025-02-29`; accepts `2024-02-29`. Closes both bot findings.
- Ticket references pruned; `grep -rn LFE-14536` is empty.
- Env flag only in `.env.dev.example`, two lines, pointing at the declaration rather than repeating it.

### Verification

- `worker` write-path suite — `Tests 61 passed (61)`
- CI green on `1ce7de1`

Threads are ready to resolve.

.env.dev.example
```diff
@@ -223,6 +223,9 @@ LANGFUSE_MIGRATION_V4_WRITE_MODE=dual
 LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true
 # OTel ingestion behaviour: `dual_write` (SDK-version dispatch) or `direct`.
 LANGFUSE_MIGRATION_V4_NATIVE_OTEL_BEHAVIOUR=dual_write
+# Cloud-only rollout boundary (ISO date) for automatic direct OTel writes.
+# Unset disables it; see worker/src/env.ts for the full semantics.
+# LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF=
```
Steffen911: Let's make this an ORG_CREATED_CUTOFF ? Otherwise the semantic seems too loose to me?

Steffen911:
@claude review

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -113,6 +114,183 @@ export function checkHeaderBasedDirectWrite(params: {
...
+    return false;
+  }
+
+  // Env validation rejects a malformed cutoff at startup, so this only guards
+  // against a value reaching the comparison some other way. Treat anything
+  // unparsable as "unknown" and keep the pre-cutoff behaviour.
+  const cutoffMs = Date.parse(`${cutoff}T00:00:00.000Z`);
+  if (isNaN(cutoffMs) || isNaN(orgCreatedAt.getTime())) {
+    return false;
+  }
```
Steffen911: Is that really necessary? Why is that check desirable if prod can never reach it.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -113,6 +114,183 @@ export function checkHeaderBasedDirectWrite(params: {
...
+          ? { createdAt: project.organization.createdAt }
+          : undefined,
+        source: "postgres",
+      };
+    });
+
+    return isOrgPastOtelDirectWriteCutoff({
+      orgCreatedAt: value?.createdAt,
+      cutoff,
+      isLangfuseCloud: true,
```
Steffen911: Why do we pass this into the function to begin with if this is only ever true?

.env.dev.example
Steffen911: Renamed to `LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CREATED_CUTOFF` in 9d5ae8bbf3 so the name states what the date compares against.

worker/src/queues/otelIngestionQueue.ts
Steffen911: It wasn't — removed in 9d5ae8bbf3. `z.iso.date()` rejects a malformed cutoff at startup, and even without the guard a NaN on either side makes the `>=` comparison false, so unparsable input already fails closed to the pre-cutoff path. The invalid-`Date` unit test stays and passes against the bare comparison.

worker/src/queues/otelIngestionQueue.ts
Steffen911: Agreed — dropped in 9d5ae8bbf3. The Cloud gate now lives solely at the call site (`isProjectOrgPastOtelDirectWriteCutoff` returns false without `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION`), and the predicate just compares dates. The parameter only existed to unit-test the gating, which wasn't worth a production argument that is only ever `true`.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -113,6 +114,183 @@ export function checkHeaderBasedDirectWrite(params: {
...
+export async function isProjectOrgPastOtelDirectWriteCutoff(
+  projectId: string,
+): Promise<boolean> {
+  const cutoff = env.LANGFUSE_MIGRATION_V4_OTEL_DIRECT_WRITE_ORG_CUTOFF;
+  if (!cutoff || !env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION) {
+    return false;
+  }
+
+  try {
+    const { value } = await orgCreatedAtCache.getOrLoad(projectId, async () => {
```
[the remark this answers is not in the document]
Steffen911: Real but accepted, and now documented on the cache (9d5ae8bbf3). `projects.transfer` is a rare admin mutation, the stale window is bounded by the 60-minute TTL and self-corrects, and in dual mode both routing outcomes still land the data in the events table — worst case a transferred project temporarily keeps the previous org's write shape. A cross-service invalidation hook from web into the worker's in-process cache isn't proportionate to that.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -422,22 +610,37 @@ export const otelIngestionQueueProcessorBuilder = (
...
+        scopeBasedDirectWrite = checkSdkVersionRequirements(
           sdkInfo,
           hasExperimentEnvironment,
         );
+        // An older SDK that predates the header still identifies itself through
+        // its instrumentation scope, so both signals have to be consulted
+        // before the org cutoff may flip a batch.
+        langfuseSdkTraffic =
+          langfuseSdkTraffic ||
+          isLangfuseSdkTraffic({ scopeName: sdkInfo.scopeName });
```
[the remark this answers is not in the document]
Steffen911: Fixed in 9d5ae8bbf3: the org-cutoff exclusion now scans every instrumentation scope on every resource (`batchContainsLangfuseScope`) instead of only the first, so a collector-forwarded batch that mixes Langfuse SDK spans with third-party scopes in any order keeps its established write path. The priority-2 scope *version* check intentionally still reads the first scope — pre-existing behavior for batches the SDK itself exported, where all spans share one scope. Unit tests cover later-scope/later-resource placement and malformed `scopeSpans` shapes.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -113,6 +114,206 @@ export function checkHeaderBasedDirectWrite(params: {
...
+  try {
+    return resourceSpans.some((resourceSpan) =>
+      (resourceSpan?.scopeSpans ?? []).some((scopeSpan) =>
+        isLangfuseSdkTraffic({ scopeName: scopeSpan?.scope?.name ?? null }),
+      ),
+    );
+  } catch (error) {
+    logger.warn("Failed to scan batch scopes for Langfuse SDK spans", error);
+    return false;
+  }
```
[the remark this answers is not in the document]
Steffen911: Valid catch — the whole-scan try/catch did abort on the first malformed resource and could hide a Langfuse scope on a later one. Fixed in 5877a8ef89: the scan now guards each resource with `Array.isArray` and keeps going, and since plain JSON-parsed data cannot throw on property access the catch-all is gone entirely. Regression test added for exactly this shape (malformed resource first, Langfuse scope on a later resource).

Steffen911:
@claude review

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -113,6 +114,205 @@ export function checkHeaderBasedDirectWrite(params: {
...
+    (resourceSpan) =>
+      Array.isArray(resourceSpan?.scopeSpans) &&
+      resourceSpan.scopeSpans.some((scopeSpan) =>
+        isLangfuseSdkTraffic({ scopeName: scopeSpan?.scope?.name ?? null }),
+      ),
+  );
+}
+
+/**
+ * Whether an organization's signup date puts it past the direct-write cutoff.
```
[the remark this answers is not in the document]
Steffen911: Confirmed and fixed in a56d7a074f. `parsedSpans` is raw `JSON.parse` output (or the masking hook's return) with no shape guarantee, every other consumer guards `Array.isArray(resourceSpans)` and degrades gracefully, and `.some()` is a method invocation — so removing the try/catch had reopened this failure mode one level up: a non-array payload would crash the job and drop the batch after retries, the exact class the parseId incident taught us to avoid. Added the top-level guard plus tests for object and string payloads.

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -407,9 +610,16 @@ export const otelIngestionQueueProcessorBuilder = (
...
+      // exactly and `direct_org_cutoff` counts only the traffic that rule
+      // genuinely moves off the dual path.
+      let scopeBasedDirectWrite = false;
+      let langfuseSdkTraffic = isLangfuseSdkTraffic({
+        sdkName: job.data.payload.sdkName,
+      });
+      if (!envForcesDirect && !headerBasedDirectWrite) {
         const hasExperimentEnvironment = observations.some((o) => {
           const body = o.body as { environment?: string };
           return body.environment === "sdk-experiment";
```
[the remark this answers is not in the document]
Steffen911: Done in a56d7a074f. Extracted `isOtelDirectWriteOrgCutoffConfigured()` (date set + Cloud region) and gated the scope scan behind it, the same way the signup-date lookup already short-circuits — so deployments where the rule cannot apply (the shipped default and all self-hosted) skip the scan entirely. `langfuseSdkTraffic`'s only consumer is the cutoff branch, which is false under exactly the same condition, so the outcome is unchanged.

Steffen911:
@claude review

worker/src/queues/otelIngestionQueue.ts
```diff
@@ -422,22 +649,41 @@ export const otelIngestionQueueProcessorBuilder = (
...
+        // An older SDK that predates the header still identifies itself through
+        // its instrumentation scope, so both signals have to be consulted
+        // before the org cutoff may flip a batch. Checked across every scope in
+        // the batch, not just the first — collector-forwarded batches can mix
+        // SDK spans with third-party scopes in any order. The scan only feeds
+        // the cutoff rule, so it is skipped while that rule cannot apply.
+        langfuseSdkTraffic =
+          langfuseSdkTraffic ||
+          (isOtelDirectWriteOrgCutoffConfigured() &&
+            batchContainsLangfuseScope(parsedSpans));
```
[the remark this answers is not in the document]
Steffen911: Deliberately declining this one. The dead-work window is real but tiny in exactly the scenario it applies to: `scopeBasedDirectWrite` can only be true because the first scope of the first resource is a Langfuse scope, so `batchContainsLangfuseScope` short-circuits on the very first scope it inspects — the redundancy is one `toLowerCase().includes()` on a scope name, only for sdk-experiment batches with a cutoff configured, not a batch scan.

Adding `!scopeBasedDirectWrite` would buy that back by making `langfuseSdkTraffic` false for traffic that *is* SDK traffic, feeding `resolveOtelWritePath` (the exported, unit-tested authority for the rule — the suite asserts the `direct_scope` + `langfuseSdkTraffic: true` combination) a deliberately wrong input that only happens not to matter under today's precedence. That hidden coupling costs more than the string compare saves. The a56d7a074f gate was a different trade: it skipped genuinely full scans for every plain-OTel batch in deployments where the rule cannot apply at all.

PR #15820, opened 2026-08-06
feat(web): pnpm structure:move — a move with every importer rewritten (LFE-14804)

nkabardin:
🟡 **Live preview: [pr-15820.preview.langfuse.com](https://pr-15820.preview.langfuse.com)**

## TL;DR

`pnpm structure:move <from...> <to-dir>` moves files and folders and rewrites every importer, using the same compiler primitive VS Code's "move file" uses. The instrument panel from #15800 said what to fix; this is the hands for the mechanical half. Tooling only — no source moves in this PR.

```
pnpm structure:move --dry-run src/hooks/useFoo.ts src/features/bar/hooks
pnpm structure:move src/components/ui/AdvancedJsonViewer src/features/traces
```

## Why

Step 4 of the RFC migration plan is ~2,600 violations' worth of relocation. Done by hand that is a specifier-regex-and-pray exercise; done wrong it silently breaks `vi.mock` paths and loses `git blame`. The migration only works as many small, boring, obviously-correct PRs, so the move itself has to be free and trustworthy.

## What

`LanguageService.getEditsForFileRename` over `web/tsconfig.json` computes the rewrites — so `@/src/...` aliases, extension-less specifiers, index resolution and literal dynamic `import()` are the compiler's problem, not a regex's. Booting the service costs ~4s, every move after that is instant, hence the batch-shaped CLI.

Four properties make it safe to hand to an agent:

- **Move ≠ edit** (RFC rule 15). Rewrites land in importers. A moved file may only change where an alias self-reference has to follow the subtree it is part of; anything else — a link to something left behind, an edit outside a module specifier — aborts the whole batch, verified as a postcondition rather than assumed.
- **History preserved.** `git mv`, so `log --follow` and `blame -C` keep working.
- **Never destructive.** No `reset`, no `stash`, no `checkout --`. On failure it prints the way back, which is the inverse `structure:move`, not a reset. Idempotent: re-running an applied move is a no-op, exit 0.
- **Verifies itself.** Closes with `tsc --noEmit`, then `structure:stats --diff` so the win is visible in the terminal.

## Result

Calibration: dry-ran **and** executed the top move item — `src/components/ui/AdvancedJsonViewer` (56 files) → `src/features/traces/` — on a scratch branch, then reverted it.

| | |
| --- | --- |
| diff | 59 files changed, **9 insertions, 9 deletions** — 56 renames (55 at 100% similarity), 3 importers rewritten |
| `tsc --noEmit` | green |
| tests | `Test Files 10 passed (10)`, `Tests 176 passed (176)` |
| rule 6 (file used by one feature lives in it) | 103 → **64** |
| rule 13 (`components/ui` frozen) | 118 → **75** |
| rule 5 / rule 8 | +4 / +7 — the honest part, see below |
| total | 2692 → **2621** |
| revert | the printed inverse move, back to a byte-identical clean tree |

Two findings worth more than the tool:

**A pure move into a feature trades debt, it does not only clear it.** Once the folder lives under `src/features/traces`, its `utils/` and `lazy/` subfolders are outside the kind-folder list (rule 5 +4) and its own cross-feature imports start counting (rule 8 +7). Net is still −71, but "move it home" is step one of two — the kind-folder normalization and the feature `index.ts` are the follow-ups. Worth knowing before we queue 100 of these.

**The string blind spot is not theoretical.** The move left three `vi.mock("@/src/components/ui/AdvancedJsonViewer/…")` paths behind; six `IOPreviewJSON` tests failed under a completely green `tsc`. So the tool no longer prints a grep *hint* — it runs the grep and prints every surviving hit.

<details>
<summary>Mechanism, the batching trap, and what is deliberately not here</summary>

**The trap that makes this a script and not a loop.** The language-service host is mutable: each rename rewrites the in-memory layout and bumps the affected script versions plus the project version, so move #2 computes its edits against the tree move #1 produced. Freezing those versions is not a subtle bug — the second move's spans become offsets into stale text. Proven by running a batch of two hooks that share one importer against a frozen-version host:

```
  src/features/in-app-agent/components/InAppAgentMessage.tsx
    @/src/hooks/useElementSize → @/src/features/in-app-agent/hooks/useElementSize
    @/src/hooks/useWatchedPromiseCallback → @/src/hooks/useProjectIdFromURL
    @/src/hooks/useProjectIdFromURL → ./utils/markdown
    ./utils/markdown → ./InAppAgentMessage.module.css
    ...
```

Every import below the first edit shifts onto its neighbour. With versions bumped, the same batch produces two correct rewrites in that file.

**Move ≠ edit, precisely.** The postcondition parses each moved file before and after, blanks out every module specifier, and requires the remaining text to be byte-identical — so a non-specifier edit can never slip through. Each changed specifier is then resolved against the post-move layout: if it lands inside the move set it is a self-reference following its own subtree (`@/src/…` encodes the old path, so it has to), and it is reported separately. If it lands anywhere else the batch aborts with the offending specifiers, because that means a sibling was left behind.

**Colocated siblings.** `X.tsx` brings `X.clienttest.tsx`, `X.stories.tsx`, `X.fixtures.ts`, `.servertest`/`.test`/`.spec`/`.module` from the same directory, unless `--no-siblings`. The tag list is closed on purpose so `index.ts` never drags `index.tsx` along. `__tests__/X*` is not included — say so explicitly.

**Housekeeping.** Importer rewrites go through prettier (a longer specifier can push a line past the print width) and are staged. Edits offered for `tsconfig.json` and generated `.next/types` declarations are filtered out. `.structure-baseline.json` is untouched — baseline resets are a deliberate act in fix PRs, never a side effect of tooling.

**Deliberately absent:** renames, splits, and `structure:query` (LFE-14806). The surface staying small is the feature; anything a move cannot express cleanly is a follow-up note, not a flag.

</details>

Seed: #15800 (instrument panel) · ticket LFE-14804 · meta LFE-14748

### DIFF
--- a/web/.dependency-cruiser.js
+++ b/web/.dependency-cruiser.js
@@ -30,12 +30,19 @@ module.exports = {
     {
       name: "rfc08-features-via-index",
       comment:
-        "Rule 8: features import other features only through their index.ts.",
+        "Rule 8: features import other features only through a surface — " +
+        "`<feature>` from client code, `<feature>/server` from server code. " +
+        "A feature has two, because it is a full-stack slice: the root index " +
+        "must stay client-safe, so it can never re-export server/.",
       severity: "warn",
       from: { path: "^(src/(?:ee/)?features/[^/]+)/" },
       to: {
         path: "^src/(?:ee/)?features/[^/]+/",
-        pathNot: ["^$1/", "^src/(?:ee/)?features/[^/]+/index\\.tsx?$"],
+        pathNot: [
+          "^$1/",
+          "^src/(?:ee/)?features/[^/]+/index\\.tsx?$",
+          "^src/(?:ee/)?features/[^/]+/server/index\\.tsx?$",
+        ],
       },
     },
     {
--- a/web/AGENTS.md
+++ b/web/AGENTS.md
@@ -220,6 +220,9 @@ Sentry instrumentation skill first and decide whether it should capture at all
 - Agent browser install to the default user-level Playwright cache: `pnpm run playwright:install`
 - Build: `pnpm --filter web run build`
 - Structure-RFC violation counts: `pnpm --filter web run structure:stats`
+- Move files/folders with every importer rewritten:
+  `pnpm --filter web run structure:move <from...> <to-dir>` — never hand-edit
+  import specifiers for a move, and never `mv` a source file without it
   (see `web/scripts/structure/README.md`)
 
 ## Playbooks
--- a/web/package.json
+++ b/web/package.json
@@ -27,6 +27,7 @@
     "build-trace": "dotenv -e ../.env -- next internal trace .next/trace-turbopack",
     "storybook": "dotenv -e ../.env -- storybook dev -p 6006",
     "structure:stats": "node scripts/structure/stats.mjs",
+    "structure:move": "node scripts/structure/move.mjs",
     "build-storybook": "dotenv -e ../.env -- storybook build",
     "sfdc:backfill": "dotenv -e ../.env -- tsx src/ee/features/sfdc-sync/scripts/backfillSfdcLeads.ts"
   },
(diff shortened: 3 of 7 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15825, opened 2026-08-06
fix(dashboard): use raw value for histogram bin index, not rounded value

### LINKED ISSUE #15208: bug: silent wrong results in two score-analytics aggregations (histogram binning & pivot-table averages)
## Summary

While differential-testing Langfuse's client-side statistical code against reference libraries (`numpy`, `scikit-learn`), I found two **silent** correctness bugs — they return wrong numbers to users with no error, on the dashboards people use to read eval results:

1. **Numeric score histogram** (`createHistogramData`) counts scores in the **wrong bucket**, because it rounds each value to 2 decimals *before* binning. The rendered distribution disagrees with a standard histogram, and buckets are labeled with ranges that don't contain the values counted in them.
2. **Pivot-table subtotals/grand totals** (`calculateSubtotals`) compute averages as an **average of per-group averages** instead of a count-weighted average, so subtotal/total rows overstate or understate metrics like avg cost/latency whenever groups differ in size.

Both live in correctness-critical analytics paths. I've verified the neighbouring inter-rater stats (`calculateCohensKappa`, `calculateWeightedF1Score`, `calculateOverallAgreement`) against scikit-learn and they are **correct** (0 mismatches over 1,200 random cases) — these two are the outliers.

---

## Bug 1 — Numeric score histogram bins on rounded values

**File:** `web/src/features/dashboard/lib/score-analytics-utils.ts` → `createHistogramData` (rendered by the dashboard `NumericScoreHistogram`, produced in `dashboard-router.ts`).

The value is rounded to 2 decimals before its bin index is computed, while the bin edges are not:

```ts
const min = round(Math.min(...numericScoreValues));            // rounded
const binSize = range / bins || 1;
// ...
const shiftedValue = round(value) - min;                       // value rounded to 2dp
const binIndex = Math.min(Math.floor(shiftedValue / binSize), bins - 1);
```

### Reproduction (vs `numpy.histogram`)

A faithful port of the function was run against `numpy.histogram` (same bin count) over 1,200 random score datasets:

```
uniform 0..1 (n=50)          | value-in-wrong-bucket:185/200  numpy-mismatch:117/200
uniform 0..1 (n=200)         | value-in-wrong-bucket:200/200  numpy-mismatch:199/200
narrow 0.80..0.85 (n=40)     | value-in-wrong-bucket:200/200  numpy-mismatch:200/200
1..5 likert (n=60)           | value-in-wrong-bucket:  0/200  numpy-mismatch:  0/200
...
runs differing from numpy.histogram: 916 / 1200
```

Integer (Likert) scores are unaffected — which is why this has gone unnoticed — but any fractional scores (0–1 confidence, LLM-judge scores, cost, latency) are misbinned.

**Minimal deterministic case** (7 equal-width bins over `[0, 1]`, input `[0, 0.857, 0.86, 1]`):

| value  | bucket it's counted in (current) | bucket that contains it |
| ------ | -------------------------------- | ----------------------- |
| 0.857  | **`[0.86, 1]`** ❌               | `[0.71, 0.86]`          |

`round(0.857) = 0.86`, so `floor(0.86 / (1/7)) = 6` → the **top** bucket, even though `0.857 < 0.86`.

A second symptom of the same rounding: for narrow ranges the 2-decimal **labels collapse**, e.g. `createHistogramData([0.855, 0.857, 0.858, 0.861])` produces buckets labeled `["[0.85, 0.85]", "[0.85, 0.86]"]` — a zero-width bucket and duplicated edges.

### Fix

Bin on the raw value (`Math.floor((value - min) / binSize)`), derive edges from the raw min/max, and round only the **display labels** (with enough precision to keep edges distinct). After the fix the port matches `numpy.histogram` exactly (**0 / 1200** mismatches).

---

## Bug 2 — Pivot-table subtotals average the averages

**File:** `web/src/features/widgets/utils/pivot-table-utils.ts` → `applyAggregation` / `calculateSubtotals` (subtotal & grand-total rows in dashboard pivot-table widgets).

```ts
case "avg":
  return values.reduce((sum, val) => sum + val, 0) / values.length;   // mean of per-row means
```

Each pivot data row is already an aggregate (e.g. `avg_cost` over `count` items). Averaging those per-row averages **unweighted** only equals the true overall average when every group has the same size.

### Reproduction

The repo's own test data (and its docstring example) demonstrates it. `calculateGrandTotals` over:

```
{ count: 10, avg_cost: 0.5 }
{ count: 15, avg_cost: 0.6 }
{ count: 20, avg_cost: 0.2 }
{ count: 25, avg_cost: 0.4 }
{ count:  5, avg_cost: 0.8 }
```

| quantity                              | value      |
| ------------------------------------- | ---------- |
| **shown** (avg of averages) `Σavg/5`  | **0.5**    |
| **correct** (weighted) `Σ(avg·count)/Σcount = 32/75` | **0.4267** |

The dashboard silently overstates average cost by ~**17%** here. The existing test even encodes the wrong value:

```ts
avg_cost: 0.5, // (0.5 + 0.6 + 0.2 + 0.4 + 0.8) / 5
```

(`percentile` subtotals have the same problem and are already flagged as "a reasonable approximation" in the code.)

### Fix

Weight each row's value by its `count` when a count metric is present:
`Σ(avg_i · count_i) / Σ(count_i)`. This makes subtotal/total averages exact for the common case (a count column is present) and falls back to the previous unweighted behavior when no weights are available. The truly exact fix for percentiles is query-level (the database can recompute over raw rows), but count-weighting is a strict improvement.

---

## Methodology

Each function was ported 1:1 to JS and executed against reference implementations:
`numpy.histogram` (Bug 1), and `sklearn.metrics.{cohen_kappa_score, f1_score, accuracy_score}` for the neighbouring categorical stats (verified correct — not bugs). Repro scripts and a proposed patch with unit tests are available; happy to open a PR.

okxint:
## Problem

`createHistogramData` in `score-analytics-utils.ts` rounded each value before computing its bin index:

```ts
const shiftedValue = round(value) - min;  // rounding before binning
const binIndex = Math.min(Math.floor(shiftedValue / binSize), bins - 1);
```

For fractional scores (LLM-judge 0–1 scores, confidence values), `round(0.857)` returns `0.86`, which can cross a bin boundary — placing the value in the wrong bucket with no visible error. Integer and Likert scores are unaffected.

## Fix

Remove `round()` from the shifted value calculation. The raw value should determine the bin. The display labels on line 69 continue to use `round()` for readability.

```diff
-    const shiftedValue = round(value) - min;
+    const shiftedValue = value - min;
```

Fixes #15208.

The PR changes histogram assignment to use raw score values rather than rounded values, preserving precision around bin boundaries.
- Removes pre-binning rounding from each score.
- Keeps rounded minimum, range, and display labels.
</details>

The PR should not merge until histogram indexing handles raw values below the rounded minimum without throwing.

Using a raw score with a separately rounded minimum allows valid fractional inputs to produce bin index -1 and crash histogram generation.

**Files Needing Attention:** web/src/features/dashboard/lib/score-analytics-utils.ts
</details>

`````markdown
### Issue 1
web/src/features/dashboard/lib/score-analytics-utils.ts:74-76
**Rounded minimum creates negative index**

When the minimum score rounds upward to two decimal places, such as `0.006` becoming `0.01`, subtracting that rounded minimum from the raw value produces a negative `binIndex`. Accessing `acc[-1].count` then throws a `TypeError` and prevents the dashboard histogram from rendering.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(dashboard): use raw value for histog..."](https://github.com/langfuse/langfuse/commit/64e230b39dfd7258b611768ec339b78cb24db472) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50723843)</sub>


nimarb:
thank you @okxint , closing though as there was no reaction to greptile. please feel free to reopen

PR #15829, opened 2026-08-06
fix(blob-storage-export): keep millisecond precision in export object keys

### LINKED ISSUE #15617: bug: Blob Storage export: object keys truncated to whole seconds cause silent overwrite / data loss
### Describe the bug

When exporting traces via the **S3 Blob Storage Integration** (hourly schedule, Parquet), some export windows are silently written as **empty (0-row) files even though the traces for those windows exist in Langfuse** (they're returned by the API and present in ClickHouse). The export run reports success and writes a manifest, so nothing signals a problem. I expected every trace present in Langfuse to appear in the exported files for its window.

The loss is **silent and permanent**: the run succeeds, the incremental cursor advances past the window, and the window is never re-scanned. It is also effectively **undetectable downstream** — for Parquet the manifest records `rowCount: null`, so an overwritten (empty) window is indistinguishable from a legitimately empty one. On our hourly integration this silently drops ~7–8 windows/day.

### What's happening

The exporter [processes at most one `frequencyInterval` of data per run](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/handleBlobStorageIntegrationProjectJob.ts#L1050-L1058), and if it is still behind the `now − lagBuffer` frontier it [immediately re-queues the next chunk](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/handleBlobStorageIntegrationProjectJob.ts#L1195-L1210). So a single scheduler wake-up that is just over one interval behind emits **two chunks back-to-back**: a full-interval chunk ending at time `T`, and a sub-second "remainder" chunk ending a few milliseconds later.

The export object key for each chunk is [built from that chunk's `maxTimestamp`](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/handleBlobStorageIntegrationProjectJob.ts#L574-L576), but [`formatBlobExportTimestamp` truncates it to whole seconds](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/manifest.ts#L27-L28) (`.substring(0, 19)`). When the two chunks' `maxTimestamp` fall in the **same wall-clock second** — which the back-to-back burst above makes routine — they resolve to the **same object key**, so the near-empty remainder chunk **overwrites** the full one. On a bucket without object versioning, the full window is then unrecoverable.

> **Not the ClickHouse < 25.11 Parquet caveat.** This is a whole-file overwrite caused by a sub-second object-key collision in application code; it is independent of the ClickHouse version and would occur on 25.11+ as well. (Being on < 25.11 only makes it harder to notice, since export failures don't surface as errors.)

### Suggested fix (untested)

I don't have a local Langfuse environment to build and validate against, so please treat this as a pointer to the probable fix rather than a verified change.

Preserve millisecond precision in the export key so two chunks can never collide — [`worker/src/features/blobstorage/manifest.ts`](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/manifest.ts#L27-L28):

```diff
 export const formatBlobExportTimestamp = (date: Date): string =>
-  date.toISOString().replace(/:/g, "-").substring(0, 19);
+  date.toISOString().replace(/:/g, "-").substring(0, 23); // keep .SSS so keys are unique per chunk
```

`"2026-07-28T11:40:00.058Z"` → `2026-07-28T11-40-00-058` instead of `2026-07-28T11-40-00`. Both chunks then get distinct keys and survive; the tiny remainder file is harmless (a window's consumers read all of its files, and adjacent windows already require ID-based dedup). This key feeds both the data files and the manifest, so the change is self-consistent.

There is precedent that millisecond precision is safe: the same job already uses a full-precision timestamp for the [self-requeue `jobId`](https://github.com/langfuse/langfuse/blob/main/worker/src/features/blobstorage/handleBlobStorageIntegrationProjectJob.ts#L1195-L1210) (`${projectId}-${maxTimestamp.toISOString()}`). Only the export **file key** truncates to whole seconds — the inconsistency this closes.

*(Secondary, optional hardening: an hourly integration lands in the "just over one interval behind" state routinely because the caught-up `nextSyncAt` is derived from a lag-buffered data timestamp and interacts with the fixed scheduler grid. That's a separate improvement; the data loss is fully addressed by the key change above regardless of how often the burst occurs.)*

### Environment

Langfuse `3.221.1` (self-hosted); ClickHouse `24.3`; destination S3 + Parquet, hourly incremental export; bucket object versioning disabled. The relevant code is unchanged on current `main` (links above).

---

### Steps to reproduce

**Observable (real integration).** Configure an hourly S3 Parquet Blob Storage export against a bucket **without object versioning**, with steady trace ingestion. After a day, chain consecutive manifest `window` bounds. Windows that collapsed to a sub-second range — each immediately preceded by a one-interval gap that no manifest covers — mark overwritten full-interval chunks. Cross-check any trace whose timestamp lands in a gap: it is returned by the API but absent from the exported files.

Example (one day, hourly), showing the collapsed window and the uncovered hour:

```
manifest (hour-ending label)   window.min → window.max          width      traces file
…T10-40-00.json                09:40:00.110 → 10:40:00.058       3599.9s    full
…T11-40-00.json                11:40:00.058 → 11:40:00.447       0.389s     0 rows  ← collapsed / overwritten
…T12-40-00.json                11:40:00.447 → 12:40:00.070       3599.6s    full
```

- The surviving `…T11-40-00` file covers only a 0.389s window and is empty.
- The hour `10:40:00.058 → 11:40:00.058` is covered by **no** manifest, yet the live API returns traces in it (e.g. a trace at `10:48:50`) — those traces never reach the export.
- The collapsed window's `min` equals the previous full window's `max + interval`: the fingerprint of a full-interval chunk that wrote `…T11-40-00.parquet` and was overwritten by the sub-second chunk writing the same key.
- The pattern repeats ~7–8×/day; the uncovered hours' live traffic equals the exact `live − export` deficit.

**Deterministic (unit-level).** Drive the export job with `now − lastSyncAt` just over `interval + lagBuffer` (hourly interval, 20-min buffer). It emits `[t, t+interval]` then `[t+interval, now−lagBuffer]`, whose `maxTimestamp` differ by < 1s. Asserting that `formatBlobExportTimestamp` returns the **same** string for both reproduces the key collision (and would make a good regression test).

Self-hosted

3.221.1

_No response_

### Additional information

_No response_

No

zzyIyzz:
## What does this PR do?

Fixes #15617

The blob storage export object key stem is built with `formatBlobExportTimestamp`, which truncated the chunk `maxTimestamp` to whole seconds (`.substring(0, 19)`). A caught-up run can emit two chunks back-to-back — a full-interval chunk and a sub-second remainder chunk — whose `maxTimestamps` fall in the same wall-clock second. Both then resolve to the same object key, so the near-empty remainder chunk silently overwrites the full window's files and manifest. On buckets without versioning that window's data is unrecoverable; the reporter observed ~7–8 silently lost windows/day on an hourly integration.

## Changes

- `worker/src/features/blobstorage/manifest.ts`: keep millisecond precision in `formatBlobExportTimestamp` (`substring(0, 23)`) and replace the dot alongside colons so the stem stays key-friendly (`2026-07-28T11-40-00-058`). Both the table file names and the manifest key derive from this one function, so the change is self-consistent.
- `worker/src/__tests__/blobExportManifest.unit.test.ts`: update the existing assertions and add a regression test covering the exact collision scenario from the issue: two timestamps within the same wall-clock second must produce distinct keys, and the fixed-width stems must preserve chronological lexicographic ordering.

## Compatibility notes

- The key stem is a naming convention only — nothing in the repo parses it back, so no consumer breaks.
- Stems remain fixed-width and zero-padded, so lexicographic ordering stays chronological; the "files and manifest sort together" contract is preserved. Old second-precision keys (`...T10-20-30.json`) sort immediately adjacent to new millisecond keys (`...T10-20-30-000.json`).
- The same job already uses full-precision timestamps for its self-requeue `jobId`; this closes the inconsistency.

## Verification

- `tsc --noEmit` (worker) passes
- Prettier check passes
- Unit tests: 6/6 pass with the fix; reverting only the `manifest.ts` change makes 4 of them fail (verified), including the new collision test

Note to maintainers: I noticed this issue is assigned to @nsemmler — happy to close this if the fix is already in flight internally.

The PR preserves millisecond precision in blob-storage export object keys to prevent distinct chunks within one wall-clock second from overwriting each other.

- Updates the shared timestamp formatter to emit fixed-width, key-friendly millisecond stems.
- Adds regression coverage for uniqueness and chronological lexicographic ordering.
- Updates manifest-key expectations to the new format.
</details>

The PR appears safe to merge, with no actionable defects identified in the changed timestamp formatting or its tests.

The shared formatter is used consistently for table and manifest keys, retains fixed-width chronological ordering, and no internal consumer parses or assumes the former seconds-only key shape.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(blob-storage-export): keep milliseco..."](https://github.com/langfuse/langfuse/commit/5f5509425c9987c81a1bbc22c5952f7e3d9019f9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50748081)</sub>

**Context used:**

- Knowledge Base — [Batch Actions and Data Exports](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/batch-actions-exports.md)

### DIFF
--- a/worker/src/__tests__/blobExportManifest.unit.test.ts
+++ b/worker/src/__tests__/blobExportManifest.unit.test.ts
@@ -22,10 +22,36 @@ const file = (
 });
 
 describe("blob export manifest builders (LFE-10843)", () => {
-  it("strips colons from the timestamp stem and truncates to seconds", () => {
+  it("strips colons and dots from the timestamp stem and keeps milliseconds", () => {
     expect(
       formatBlobExportTimestamp(new Date("2026-07-10T10:20:30.123Z")),
-    ).toBe("2026-07-10T10-20-30");
+    ).toBe("2026-07-10T10-20-30-123");
+  });
+
+  it("produces distinct keys for timestamps within the same wall-clock second", () => {
+    // A caught-up run can emit a full-interval chunk and a sub-second
+    // remainder chunk back-to-back; their keys must never collide or the
+    // remainder silently overwrites the full window's files.
+    const first = formatBlobExportTimestamp(
+      new Date("2026-07-10T10:20:30.058Z"),
+    );
+    const second = formatBlobExportTimestamp(
+      new Date("2026-07-10T10:20:30.447Z"),
+    );
+    expect(first).not.toBe(second);
+    expect(
+      buildBlobExportManifestKey({
+        projectId: "proj",
+        maxTimestamp: new Date("2026-07-10T10:20:30.058Z"),
+      }),
+    ).not.toBe(
+      buildBlobExportManifestKey({
+        projectId: "proj",
+        maxTimestamp: new Date("2026-07-10T10:20:30.447Z"),
+      }),
+    );
+    // Fixed-width stems must preserve chronological lexicographic order.
+    expect(first < second).toBe(true);
   });
 
   it("builds the manifest key under the project's manifests/ prefix", () => {
@@ -35,7 +61,7 @@ describe("blob export manifest builders (LFE-10843)", () => {
         projectId: "proj",
         maxTimestamp: new Date("2026-07-10T10:20:30Z"),
       }),
-    ).toBe("team/proj/manifests/2026-07-10T10-20-30.json");
+    ).toBe("team/proj/manifests/2026-07-10T10-20-30-000.json");
   });
 
   it("treats an absent prefix as empty", () => {
@@ -44,7 +70,7 @@ describe("blob export manifest builders (LFE-10843)", () => {
         projectId: "proj",
         maxTimestamp: new Date("2026-07-10T10:20:30Z"),
       }),
-    ).toBe("proj/manifests/2026-07-10T10-20-30.json");
+    ).toBe("proj/manifests/2026-07-10T10-20-30-000.json");
   });
 
   it("assembles the manifest payload with version, window, and distinct tables", () => {
--- a/worker/src/features/blobstorage/manifest.ts
+++ b/worker/src/features/blobstorage/manifest.ts
@@ -35,9 +35,15 @@ export type BlobExportManifest = {
   files: BlobExportManifestFile[];
 };
 
-/** Shared with table file names so a run's files and manifest sort together. */
+/**
+ * Shared with table file names so a run's files and manifest sort together.
+ * Keeps millisecond precision so consecutive chunks whose maxTimestamps fall
+ * into the same wall-clock second (e.g. a caught-up full-interval chunk
+ * immediately followed by a remainder chunk) never resolve to the same
+ * object key and silently overwrite each other.
+ */
 export const formatBlobExportTimestamp = (date: Date): string =>
-  date.toISOString().replace(/:/g, "-").substring(0, 19);
+  date.toISOString().replace(/[:.]/g, "-").substring(0, 23);
 
 /** `{prefix}{projectId}/manifests/{maxTimestamp}.json` — prefix-filterable per run. */
 export const buildBlobExportManifestKey = (params: {


niklassemmler:
This approach doesn't work for us. It would change the file names of the blob export on which other users may rely.

We will address it in https://github.com/langfuse/langfuse/pull/16592 instead.

PR #15840, opened 2026-08-06
refactor(traces): migrate features/traces to the project-structure RFC (LFE-14804)

nkabardin:
🟡 **Live preview: [pr-15840.preview.langfuse.com](https://pr-15840.preview.langfuse.com)**

## TL;DR

`features/traces` migrated to the project-structure RFC. **185 files, 108 renames, 457 insertions against 1370 deletions** — the feature got smaller, not bigger.

```
traces-scoped violations   176 → 132
repo-wide total           2676 → 2592
```

Everything mechanical went through `pnpm structure:move` (#15820), so history survives: use `git log --follow` and `blame -C` on anything here. Each commit is one item from `structure:stats` and is independently green.

> **Stacked on #15820** (the tooling PR) — this PR's base is that branch, so the diff you see is traces only. Retarget to `main` once #15820 merges.

## Why this feature first

It's the largest feature in the app and it was carrying every category of debt at once, so it was the honest test of whether the RFC's rules survive contact. They mostly did. Where they didn't, the migration produced **eight RFC amendments** — four already approved in-flight and encoded in the detectors, four still needing a call. That was the real output.

## What changed

**The JSON viewer came home.** `components/ui/AdvancedJsonViewer` (56 files) was used by traces and nothing else, so it moved into `traces/components/`. That alone is rule 6: 42 → 1 and drains 43 files out of the frozen `components/ui`.

**The feature root now tells the truth.** It is `index.ts`, `TracePage.tsx`, and kind folders — nothing else:

```
components/  constants/  contexts/  fns/  hooks/  server/  stores/  types/
```

`_shared/`, `config/`, `shared/`, `lib/` and `IOPreview/lib/` are gone; `Trace.tsx`, `TraceDetailActions.tsx`, `TraceDetailBody.tsx` and `useTraceDetailData.ts` moved out of the root into `components/` and `hooks/`.

**The feature has a front door.** `index.ts` re-exports exactly what the twelve external consumers already imported — nothing added for a hypothetical future caller. It is client-safe by construction: `server/` is absent, so importing traces can never pull Prisma or ClickHouse into a browser bundle.

**1370 deletions, and most of them are dead code the sensor could not see.** A stale 283-line fork of `buildTraceUiData` (the live one is in `treeBuilding.ts`), a 105-line module whose only importer was its own test, and two exports referenced nowhere at all.

| rule | | before | after |
| --- | --- | --- | --- |
| 6 | a file used by one feature lives in it | 42 | **1** |
| 7 | no reaching into a component's internals | 13 | **10** |
| 8 | *incoming* deep cross-feature imports | 9 | **3** |
| 1 | one component per file | 10 | **6** |
| 5 | kind folders are a closed list | 7 | **3** |
| 2 / 3 / 4 / 9 / 19 | | 4 / 10 / 5 / 2 / 1 | **2 / 12 / 7 / 1 / 0** |

Rules 3 and 4 read higher because 56 files of the viewer's own pre-existing debt now count inside traces instead of hiding in `components/ui`. Nothing was hidden; the number moved address.

## Verification

`tsc --noEmit`, `pnpm run lint` (`Tasks: 7 successful`), `prettier --check` and `Tests 461 passed | 48 skipped (509)` after **every** commit, not just at the end.

<details>
<summary>The eight RFC amendments this produced, and what is still open</summary>

Full paste-ready text lives in the LFE-14804 handover. Encoded in the detectors already (approved in-flight):

- **`constants/` and `types/` are kind folders.** Traces had a bespoke `shared/` holding one constant and a `config/` holding one module purely because the closed list had no home for a value. A constant is not a function.
- **`docs/` is allowed at any level** and is not a kind folder — it holds prose, nothing imports it.
- **A `fns/` module folder groups one engine.** `utils/pathUtils.ts` has 16 related exports, `searchJson.ts` 15. The only rule-4-compliant shape was ~100 files flat in one folder, which is worse to navigate than the status quo — the opposite of the RFC's goal. A folder inside `fns/` now groups them, so each file still has one export; it may not grow kind folders of its own.
- **A feature has two surfaces**, `index.ts` and `server/index.ts`, because it is a full-stack slice. Rule 8 wanted the three server consumers of `server/legacyIoSearch` routed through the index; rule 10 forbids the client reaching `server/`. Those were unsatisfiable together — the same erosion the RFC's shared-package section documents from #15768, one layer up.

Still needing a call: a surface may only re-export feature-level modules (routing through an index *relocates* rule 7 onto the index unless what it exposes is flat); rule 15's absolute-import precondition is a rule the codebase does not yet follow, not a property it has; a type and a component may not share a name; and whether `components/` gets the same module-folder escape as `fns/` — six one-line badges that are only ever rendered together currently want six files.

**Two sensor gaps found by eye, not by the tool.** Rule 5 inspects folders only, so four loose files at a feature root were invisible. Rule 8 counts feature→feature only, so `src/components/table/peek` reaching into traces internals showed up nowhere but a survey metric. Both are the same violation as the ones that do count.

**Tooling the work forced.** `structure:move` gained a rename verb, case-only rename support (macOS calls the destination existing; TypeScript sees no rename at all under a case-insensitive host), and `git add` for a moved file's respelled self-reference. Two remaining verbs are argued for by having done without them: `structure:absolutise` (five files in one feature could not move until their relative imports became `@/src/...`) and an AST-based `structure:split` — I broke a file twice doing splits with text surgery, and `tsc` caught it both times.

</details>

Ticket LFE-14804 · meta LFE-14748 · tooling #15820 · loop template #15821


nkabardin:
@claude review

nkabardin:
@claude review

PR #15850, opened 2026-08-06
fix(agent): cancel a conversation's run when it is deleted

LysanderKie:
🟡 **Live preview: [pr-15850.preview.langfuse.com](https://pr-15850.preview.langfuse.com)**

## Why

`deleteConversation` soft-deleted a conversation regardless of its run state. Nothing reconciles a conversation nobody can open, so the orphaned run kept a worker and a slot against the per-user concurrency ceiling until its deadlines expired. A read-then-delete guard also races `startBackgroundRun`.

Pre-existing bug, found while building the assistant activity feed. Split out because it stands on its own.

## How

**Deleting cancels rather than refuses.** Deleting is an unambiguous statement of intent. Requiring the user to stop first is busywork, especially when a run is parked on an approval.

- `lockConversation` now locks and validates that the conversation still exists and is not soft-deleted. This blocks run creation and approval continuations that resume after a concurrent delete.
- `cancelConversationRunsInTransaction` owns the full transactional cancellation sweep. It shares the same private transition as Stop, including the QUEUED→RUNNING CAS fallback.
- **QUEUED** and **AWAITING_APPROVAL** runs cancel immediately. **RUNNING** cancellation remains cooperative: the worker sees `cancelRequestedAt` and winds down as it does after Stop.
- The sweep uses the full unsettled status set rather than `finishedAt: null`, because a run parked on approval already has `finishedAt` set.
- Immediately cancelled queue jobs are removed only after the database transaction commits.

## Trade-offs worth knowing

- A RUNNING run keeps its capacity slot for up to one heartbeat interval after deletion returns. This matches Stop; the alternative is blocking the UI on cooperative cancellation.
- MCP key cleanup for a RUNNING run relies on the worker's existing teardown. This is unchanged by the deletion path.

## Impacted packages

- `@langfuse/shared` — strengthens the conversation-lock invariant and centralizes race-safe cancellation in the server-only lifecycle module.
- `web` — delegates router deletion to the background-run service and removes queue jobs after a successful transaction.
- `worker` — unchanged; covered by a regression test because it consumes the shared lifecycle state.

No database schema, queue payload, or external API changes.

## Verification

| Check | Result |
| --- | --- |
| `mise exec -- pnpm run lint` | `Tasks: 7 successful, 7 total` |
| `mise exec -- pnpm --filter @langfuse/shared run typecheck` | passed |
| `mise exec -- pnpm --filter web run typecheck` | passed |
| `mise exec -- pnpm --filter worker run typecheck` | passed |
| `mise exec -- pnpm --filter @langfuse/shared run test runLifecycle.test.ts` | `Tests 3 passed (3)` |
| `mise exec -- pnpm --filter web run test in-app-agent-background-run.servertest.ts` | `Tests 23 passed (23)` |
| `mise exec -- pnpm --filter worker run test executeInAppAgentRun.test.ts` | `Tests 17 passed (17)` |

Coverage proves that a run cannot be admitted after deletion, preserves the QUEUED→RUNNING cancellation race case, and exercises router deletion with immediate approval cancellation plus cooperative RUNNING cancellation.

## Context

Part of splitting #15842, which stays open as the unstacked big-picture view and is **not** a merge candidate.

---

> [!NOTE]
> **Medium Risk**
> Changes core in-app agent lifecycle (locking, delete, and cancellation) with race-sensitive transactions; behavior is well covered by new tests but touches shared concurrency paths.
> 
> **Overview**
> **Deleting an in-app agent conversation now tears down unsettled runs instead of leaving them orphaned.** The router delegates to `deleteBackgroundConversation`, which in one transaction locks the conversation, cancels every **QUEUED**, **RUNNING**, and **AWAITING_APPROVAL** run (including approval-parked runs that already have `finishedAt`), soft-deletes the row, then drops queue jobs for runs cancelled immediately so workers do not waste slots.
> 
> **`lockConversation` is stricter:** it still takes a row `FOR UPDATE` via new `lockConversationRow`, but now rejects missing or soft-deleted conversations—so `createQueuedRun` and similar paths cannot admit work after a concurrent delete. Event append uses `lockConversationRow` only so a worker can still finish writes where appropriate.
> 
> **Cancellation logic is centralized** in `cancelRunInTransaction`, shared by single-run stop and the new `cancelConversationRunsInTransaction` sweep (same immediate cancel for queued/approval and cooperative `cancelRequestedAt` for running workers).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 209dccb0c1b2fe94325aac82ea573e9cb78d8a6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LysanderKie:
@claude review

PR #15853, opened 2026-08-06
feat(agent): float a card when a run finishes or asks for approval

LysanderKie:
🟡 **Live preview: [pr-15853.preview.langfuse.com](https://pr-15853.preview.langfuse.com)**

## Why

The badge and row indicators from #15852 already carry every state this shows. This is the push layer on top of them: you find out without going to look.

## How

**Assistant-owned, not Sonner.** The app mounts `<Toaster visibleToasts={1} />`, so a long-lived approval card parked in that single slot would mute every unrelated app alert for as long as it sits there. Raising `visibleToasts` globally would change toast behaviour for every feature.

- Results auto-hide after **8s**; an approval **waits**, because it is a question blocking a run rather than an FYI you can catch later from the badge.
- **Delivery is recorded when a card leaves, not when it appears.** A card the user never actually got — tab closed inside the result window, approval still unanswered — is therefore still undelivered on the next load and shows again.
- **Dismissing is not reading.** It records delivery and leaves the conversation unread; only looking at the conversation clears the badge.

**A card is identified by its run, not its conversation.** Keying the dismissed set by conversation silently suppressed the *next* run's card in that same conversation — the badge counted it and nothing ever appeared. Found by driving the preview in a browser.

## Review notes

- The presentational stack (`InAppAgentActivityCards`) is split from the lifecycle (`InAppAgentActivityNotifications`) so the states and their ordering render in isolation. Storybook covers them — 8 story tests including the stack cap and the open/dismiss callbacks — rather than a client test repeating the same ground.
- One drive-by fix: the host file contained a stray `NUL` byte used as a string separator, which made git treat the source as binary and its diff unreadable. It worked (split/join were symmetric) but it is now a space.

## Verification

| Check | Result |
| --- | --- |
| `pnpm run lint` | `Tasks: 7 successful, 7 total` |
| `pnpm --filter web run typecheck` | clean |
| `pnpm --filter web run test-client src/features/in-app-agent` | `Tests 122 passed (122)` |
| `pnpm --filter web run test-storybook src/features/in-app-agent` | `Tests 57 passed (57)` |

Browser-verified on the preview: card renders, X dismisses it and it stays dismissed, auto-expiry records delivery, clicking the body opens that conversation and clears the badge.

## Context

Stacked on #15852. Part of splitting #15842, which stays open as the unstacked big-picture view and is **not** a merge candidate.

### DIFF
--- a/packages/shared/src/in-app-agent/backgroundWatch.ts
+++ b/packages/shared/src/in-app-agent/backgroundWatch.ts
@@ -93,3 +93,23 @@ export function isActiveInAppAgentRunStatus(
 ): boolean {
   return IN_APP_AGENT_ACTIVE_RUN_STATUSES.includes(status);
 }
+
+/**
+ * Runs the user still has to deal with: executing, or parked on their decision.
+ *
+ * Wider than {@link IN_APP_AGENT_ACTIVE_RUN_STATUSES} on purpose. A parked
+ * approval sets `finishedAt`, so `finishedAt: null` finds executing runs but
+ * silently misses the one state that is literally waiting for the user.
+ */
+export const IN_APP_AGENT_UNSETTLED_RUN_STATUSES: readonly InAppAgentRunStatus[] =
+  [
+    InAppAgentRunStatus.QUEUED,
+    InAppAgentRunStatus.RUNNING,
+    InAppAgentRunStatus.AWAITING_APPROVAL,
+  ];
+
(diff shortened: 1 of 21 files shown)


LysanderKie:
Superseded: toast cards were folded into #15852 on `split/c-activity-feed` (rebased onto current `main`).

Issue #13330, opened 2026-04-23
bug: inline images from image_url @@@langfuseMedia:…@@@ references no longer render in Formatted chat view


crouchcd-alt:
### Describe the bug

In the trace/observation Formatted view, chat messages whose `content` is an OpenAI-style multi-modal array containing an `image_url` part with a `@@@langfuseMedia:type=...|id=...|source=...@@@` reference string no longer render the image inline between sibling `text` parts. The image only shows up in the separate "Media" panel rendered below the message.

The JSON view still shows the reference string in the expected position inside the content array, and the media panel correctly resolves the asset, so ingestion and reference parsing are fine — the breakage is purely in the Formatted view rendering path.

**Expected behavior:** given a user message like

```json
[
  { "type": "text", "text": "before" },
  { "type": "text", "text": "middle" },
  { "type": "image_url", "image_url": { "url": "@@@langfuseMedia:type=image/png|id=…|source=base64_data_uri@@@" } }
]
```

the image renders inline (via `LangfuseMediaView`) between/after the text parts in the Formatted view, and the Media strip below the message omits it (deduped).

**Actual behavior:** the `image_url` part silently renders nothing; only the separate Media strip shows the image.

### Steps to reproduce

1. Ingest a trace/observation whose `input` contains a chat message with a content array shaped like the snippet above (text parts + an `image_url` part whose `url` is a `@@@langfuseMedia:…@@@` reference string).
2. Open the trace in the UI and switch to the Formatted view of the user message.
3. Observe that only the text parts render; the image does not appear inline. The JSON view still shows the `image_url` entry, and the Media panel below the message shows the resolved asset.

### Langfuse Cloud or self-hosted?

Self-hosted

### If self-hosted, what version are you running?

v3.169.0

### SDK and integration versions

Not SDK-specific — any client that sends the documented multi-modal `image_url` content-part shape with a Langfuse media reference string reproduces it.

### Additional information

Locally, I was able to get inline rendering working again by passing the raw `content` into `MarkdownView` from `web/src/components/ui/MarkdownJsonView.tsx` instead of the zod-validated/transformed value it currently forwards. That unblocked the flow for me, but it feels like a symptom-level patch — maintainers likely have a better sense of the right long-term fix (e.g. around the transforming `MediaReferenceStringSchema` interacting with `OpenAIContentSchema`) than I do.

Happy to open a PR with whatever direction you prefer.

### Are you interested in contributing a fix for this bug?

Yes

[...]

PR #15882, opened 2026-08-07
fix(sessions): render Langfuse media tokens inline in the session view (LFE-14815, LFE-9577)

nkabardin:
🟡 **Live preview: [pr-15882.preview.langfuse.com](https://pr-15882.preview.langfuse.com)**

> **Now contains #15887 too** (LFE-9577), which was merged into this branch — so
> this PR carries both fixes. `main` has been merged in; conflicts were only in
> the seeder skill's table.

## TL;DR

Inline Langfuse media tokens rendered **nowhere** — a schema transform silently
destroyed the very field the renderer keys off. Trace detail hid it behind a
separate "Media" strip; the session view had no strip, so a customer's attached
images vanished. One fix restores inline rendering on **both** surfaces, kills a
row of `undefined` ChatML keys, and gives the session view the same media
resolution trace detail has.

| | before | after |
|---|---|---|
| Session view, Formatted | nothing — plus a table of `undefined` rows | image inline in the message |
| Trace detail, Formatted | image only in a separate "Media" strip below | image inline, strip deduped away |
| `content: [ref, ref]` (bare strings) | fell back to a `Path \| Value` table | both render inline |

Closes **LFE-14815**, **LFE-9577**, and #13330 / **LFE-13602** — all the same
transform/guard mismatch.

## Why

A customer reported that images referenced by `@@@langfuseMedia:…@@@` render in
trace detail but not in the session view, and that switching Formatted → JSON
was the only workaround.

The reported symptom turned out to be the *visible edge* of a shared renderer
defect, not a session-view bug:

1. **Inline media rendered `null` everywhere.** `MarkdownJsonView` validated
   message content with `OpenAIContentSchema` and forwarded the **parsed**
   result. But that union contains `MediaReferenceStringSchema`, which
   *transforms* the reference string into a `{type, id, source,
   referenceString}` object. `MarkdownViewer` then re-tested the part with
   `isOpenAIImageContentPart`, which validates the **untransformed** schema — so
   it returned `false`, the part matched no branch, and rendered nothing.
   Measured directly: `isOpenAIImageContentPart(raw) === true`,
   `isOpenAIImageContentPart(transformed) === false`.
2. **Trace detail masked it** by also passing a resolved `media` list, which
   renders a separate strip below the message. That strip is the image visible
   in the customer's trace-detail screenshot.
3. **The session view passed no `media` at all**, so it had no mask — nothing
   rendered, which is exactly what was reported.

Separately, every message showed a row of `undefined` ChatML keys (visible in
the report's screenshots on *both* surfaces). `ChatMlMessageSchema`'s final
transform materializes all ten schema keys, so `hasAdditionalData()` — which
counted *keys* — was effectively a constant `true`.

And a third shape was broken independently: `OpenAIContentParts` only accepted
`{type, …}` objects, so `content: ["@@@langfuseMedia:…@@@", "…"]` failed the
union and dropped the whole message to a JSON table — even though one bare
reference, or several inside one string, rendered fine.

This is flag-independent: the legacy session view and the `modernSession`
("Compact Session View") preview share `LazySessionTraceEventsRow →
TraceEventsRow → SessionObservationIO → IOPreview`.

## What

- Forward the **untransformed** content to the renderer, and type the prop as
  the schema's `z.input` so the compiler keeps it that way — the previous shape
  mismatch was invisible to types because both sides said
  `OpenAIContentSchema`.
- Accept a bare reference string as a content part. It validates through the
  canonical parser via `.refine()` but deliberately does **not** transform —
  putting a transforming schema in that union is what caused defect 1.
- Dedupe the shared media strip against what rendered inline, asking the same
  exported predicate the renderer uses, so an asset appears exactly once.
- `hasAdditionalData()` tests values instead of key presence, and the
  whole-message fallback table drops unset keys.
- Resolve observation media in the session view the way trace detail does,
  gated on the payload actually containing a reference (a session card renders
  up to 50 observations, so an ungated per-observation lookup would fan out).
- Seeder: `session-shapes --shape media` with four media-bearing message shapes.

## Result

Verified locally in both session views and in trace detail: images render
inline, zero `undefined` rows anywhere, exactly one visible media strip (on the
payload the chat renderer cannot inline), JSON mode still shows the reference
strings, and no console errors.

<details>
<summary>Mechanism, review fixes, scope and verification</summary>

### Why the type change is the load-bearing part

`z.infer<typeof OpenAIContentSchema>` (output) and `z.input<…>` (input) differ
only inside the `url` union, so a renderer written against one and fed the other
type-checks cleanly and fails at runtime. Retyping `MarkdownView.markdown`,
`parseOpenAIContentParts`, and `getRenderedInlineMediaIds` to `z.input` makes
that mismatch a compile error rather than a silent `null` render.

### Addressed from review

- **Collapsed system prompts swallowed their attachment.** The strip's dedupe
  assumes inline-renderable media rendered, but `MarkdownView` collapsed a long
  system prompt to a text preview and skipped media parts — so the asset was in
  neither place. Fixed at the renderer: collapsing hides long *text*, and an
  attachment is not text, so media parts render either way. This matches the
  existing intent (the collapse threshold is already computed from text parts
  only).
- **Concatenated tags in one array element mis-rendered.** The parser's
  `^@@@langfuseMedia:(.*)@@@$` is greedy, so two tags parsed as one and the
  `|`-split overwrote `id`/`source` — pairing tag 1's content-type with tag 2's
  id (it would have fetched a PDF and rendered it as an image). The bare-part
  schema now requires exactly one whole-string match, mirroring what
  `getStandaloneMediaReferenceStrings` already did.
- **A sparse message still tabled `undefined` rows** (e.g. a named message with
  empty content). The fallback now drops unset keys, so the "zero `undefined`
  rows" claim holds for that shape too.
- **The seeder no longer hard-fails `--shape all` without storage.** Only an
  explicit `--shape media` fails; otherwise it warns and seeds the shapes that
  need no bucket.

### Dedupe correctness

`ChatMessageList` previously deduped only string content, with a comment noting
that content-part images "still fall back to the shared media strip" — true
*because* inline rendering was broken. The dedupe is gated on
`canRenderContentAsMarkdown`, the same predicate that decides whether
`MarkdownJsonView` takes the inline path, so content that falls back to a JSON
table keeps its strip (the "link-only" seeder variant covers exactly this).

### Query fan-out

`useMedia` gained an `enabled` flag. `SessionObservationIO` enables it only when
the observation's raw I/O contains `@@@langfuseMedia:` (this path ships I/O as
raw strings) and the payload was not server-truncated. A session without media
issues no lookups.

### Verification

- `ChatMessage.clienttest.tsx` — a media `image_url` part reaches the media
  renderer; an array of bare references reaches it in order; a parsed empty
  message renders nothing; a sparse message shows no unset keys; a collapsed
  system prompt keeps its media. Every one confirmed failing against the old
  behaviour first.
- `SessionObservationIO.clienttest.tsx` — media forwarded to `IOPreview`, and
  the lookup gated on a reference being present.
- `mediaReferences.test.ts` — one whole reference accepted; concatenated,
  padded and plain strings rejected.
- After merging `main`: `Tests 3058 passed | 79 skipped (3137)` (web client),
  `Tests 877 passed | 2 skipped (879)` (shared), `pnpm run lint` and
  `pnpm run build:check` both `Tasks: 7 successful, 7 total`, and both
  typechecks clean.
- Browser (both session views + trace detail, Formatted and JSON): image
  inline, audio player for a second reference in the same message, 0
  `undefined` rows, 1 visible media strip, 0 console errors.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes shared ChatML schemas and core I/O rendering used by session and trace views; risk is mitigated by focused tests and gated media queries, but regressions in multimodal or formatted JSON display are still plausible.
> 
> **Overview**
> **Inline `@@@langfuseMedia:…@@@` tokens were disappearing** because `OpenAIContentSchema` parsed reference strings into objects while `MarkdownViewer` still validated the raw part shapes—parts matched no branch and rendered nothing. Trace detail partly masked this via a separate Media strip; the session view did not.
> 
> This PR **feeds renderers `z.input` content** (via `canRenderContentAsMarkdown` / untransformed `MarkdownView` props), adds **`MediaReferencePartSchema`** for bare reference array elements without transforming them, and **dedupes the Media strip** using the same inline-render predicate. **Collapsed long system prompts** still render non-text media parts so attachments are not hidden in both places.
> 
> **Session cards** now call **`useMedia`** when I/O contains a media prefix (and is not truncated) and pass results into **`IOPreview`**, matching trace detail.
> 
> **ChatML fallout** is tightened: `hasAdditionalData` checks values (not materialized keys), fallback JSON tables use **`withoutUnsetFields`**, and **`getRenderedInlineMediaIds`** counts bare reference parts.
> 
> **Seeder/docs:** `session-shapes --shape media` uploads fixtures to MinIO, links `observation_media`, and exercises inline / multi-ref / link-only / bare-refs / collapsed-prompt variants; `seed-media` splits upload vs trace/observation linking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6264b6a4c47d116f4a56e4ac359d1403d01bf115. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkabardin:
@claude review

_(Re-trigger posted by Claude Code on Nikita's behalf, at his request.)_

Context for this pass — the branch changed substantially since your last review:

1. **#15887 (LFE-9577) was merged into this branch**, so this PR now also carries the bare-reference content-part fix. `main` has been merged in as well.
2. All five findings from the previous round are fixed:
   - collapsed system prompts now keep rendering their media parts (the strip's dedupe assumed inline-renderable media rendered, so a collapsed prompt hid the asset in both places)
   - the whole-message fallback table drops unset ChatML keys, so a sparse message (e.g. named with empty content) no longer shows `undefined` rows
   - the seeder only hard-fails on storage for an explicit `--shape media`; `--shape all` warns and seeds the shapes needing no bucket
   - `MediaReferencePartSchema` now requires exactly one whole-string match, so two concatenated tags can no longer parse as one garbled reference (previously it paired tag 1's content-type with tag 2's id)
   - a stray screenshot accidentally committed earlier has been removed

Each fix has a test confirmed red against the prior behaviour.

nkabardin:
@claude review

_(Re-trigger posted by Claude Code on Nikita's behalf, under the standing permission he gave for this PR.)_

Heads-up: the new commit `6264b6a4c` is **seeder-only** — no product code changed since your clean pass. It adds a `collapsed-prompt` variant to `session-shapes --shape media` because the collapsed-system-prompt regression you flagged had no representable shape, so it could not be verified by hand. Verified locally in trace detail: the prompt truncates with an "Expand system prompt" toggle while the image part still renders, and it is correctly deduped out of the "Media" strip.

nkabardin:
@claude review

_(Re-trigger posted by Claude Code on Nikita's behalf, under the standing permission he gave for this PR.)_

Rebased onto `main`, which since the last pass landed **#15840** (the `features/traces` project-structure RFC migration). That moved two files this PR edits, so the merge needed real resolution rather than a text merge:

| file | what happened |
|---|---|
| `IOPreview/components/ChatMessageList.tsx` | moved to `traces/components/` — my dedupe change re-applied there, imports switched to the absolute form main now uses |
| `IOPreview/components/chat-message-utils.ts` | moved **and renamed** to `traces/fns/chatMessageUtils.ts` — `hasAdditionalData` + `withoutUnsetFields` auto-merged there; `ChatMessage.tsx` now imports from the new path |
| `SessionObservationIO.tsx` | main routes this file's traces imports through the feature surface, so `useMedia` now comes from `@/src/features/traces` and is exported from the barrel (RFC rule 8) instead of a deep import |
| `SessionObservationIO.clienttest.tsx` | main moved the mock to intercept the feature surface; since `useMedia` now resolves through that surface too, the test mocks `useMedia` there instead of at the api layer |

No behavioural change intended in this merge — the same ten assertions I verify by grep are all still in place. Green after the rebase: web client `3042 passed`, shared `887 passed`, `lint` and `build:check` both `7 successful, 7 total`, both typechecks clean.

nkabardin:
@claude review

PR #15886, opened 2026-08-07
feat(filters): share observed-metadata suggestions with the filter sidebar (LFE-11030)

nkabardin:
🟡 **Live preview: [pr-15886.preview.langfuse.com](https://pr-15886.preview.langfuse.com)**

> Follows #15881 (the pure move), now merged — this PR is rebased onto it, so the diff below is only the sidebar work.

**TL;DR** — The filter sidebar's Metadata facet was two bare text inputs. It now offers the same observed metadata keys (with their JSON type) and the same per-key observed values the search bar has had, from the same store, ranked by the same function. Both slots stay free-text.

### Why

Typing `metadata.` in the search bar suggests real keys with their type (`metadata.node.depth` · `number`) and `metadata.scenario:` offers real values. Open the sidebar's Metadata facet and you got a blank input and no idea what keys exist. Same project, same data, two different levels of help — the sidebar was simply never wired to the map the bar reads.

### What

- **One projection, two surfaces.** `observedMetadataOptions` (the pure function moved out in #15881) emits `metadata` → observed keys and `metadata.<key>` → that key's values. The bar merges it into its completion map; the sidebar merges it into its facet-options map. No second source, no second ranking.
- **Suggestions, never an enumeration.** The observed map only covers rows the user has already loaded, so a key or value it has never seen has to remain typeable. That rules out the pick-only combobox the keyed facets use for score names (which the backend *does* enumerate) — the metadata facet gets a free-text input with a ranked suggestion list under it instead. Keyboard: ↓/↑ to move, Enter to accept, Esc to dismiss.
- **Per-table opt-in.** `useFacetOptionsWithObservedMetadata(projectId, filterOptions)` is the one-liner. The v4 events table (the `/traces` and `/observations` surfaces) and the v3 observations table take it.
- **The harvest is no longer gated on the search bar.** Any surface that loads rows now contributes to the map the sidebar reads.

### Result

Applying `scenario = trace-tree` from the sidebar, both suggestions picked, verified in the browser:

`?filter=metadata;stringObject;scenario;=;trace-tree`

Checks: `pnpm typecheck` clean · `turbo run lint` 7/7 · 633 client tests green across the touched areas (`filters`, `components/table`, `search-bar`, `events`, `fns`, `stores`), including 4 new ones.

<details>
<summary>Mechanism, deliberate omissions, verification</summary>

**Contract.** `FacetOptions` names the option payload a view hands the sidebar and widens one option field: `type?`, the display-only hint a client-side suggestion source carries (server-enumerated options never set it). The keyed lookup `<column>.<key>` — already how the bar's planner finds a key's values — becomes how any keyed facet can offer them, so this is not metadata-specific plumbing. `StringKeyValueUIFilter` gains `keyDetails` and `valueOptions`, both derived by pure helpers in the presentation hook.

**Defects found after the first commit, all now covered by tests.** From driving it in a browser (jsdom could not see either): every keystroke commits the row, so the in-progress key came back through the options map as an "observed" key and was suggested to the input it came from; and the arrow-key index arithmetic never left the no-selection slot, so ↓ did nothing. From review: blur left `activeIndex` set, so returning to a field let the next Enter accept a highlight the user never made — blur now resets the whole interaction, which also stops a stale `dismissed` from surviving into whichever row React reuses the index-keyed instance for after a deletion; and `aria-selected` was compared against the input text, which the list filters out by construction, so it was never true (it now marks the keyboard-active option).

**Deliberately out of scope.**
- *Sessions.* Its Metadata facet filters `s."metadata"` — **session** metadata, a different key space from the observation metadata this store harvests. Wiring it would suggest keys that match nothing. It needs its own harvest; follow-up.
- *The legacy v3 traces table.* Its rows fetch metadata lazily per cell, so there is nothing to harvest there, and trace metadata is again a different key space from observation metadata. Reading the observation map into it would be the same category error. Follow-up alongside sessions; the user-facing `/traces` page is the v4 events table, which is wired.
- *Score-name pickers.* Left on cmdk's fuzzy scoring. Switching them to `filterRank` for ranking symmetry would narrow their matches (substring vs subsequence) — a separate, deliberate call, not a side effect of this PR.

**Verification.** Driven with Playwright against the local dev server on a project with 14 observed metadata keys: key list renders with type hints (`node.depth` · number, `scope` · object), picking fills the input, the value list for `scenario` offers `trace-tree`, picking it applies the filter above, and the free-text path still accepts an unobserved key. No console or page errors. Browser closed after the run.

</details>

---

> [!NOTE]
> **Low Risk**
> UI-only filter sidebar and client-side suggestion plumbing; no auth, API, or query semantics changes beyond richer facet options.
> 
> **Overview**
> **The Metadata facet in the filter sidebar now suggests the same observed keys, JSON type hints, and per-key values as the search bar**, using one persisted per-project store and `rankFacetOptions` ordering.
> 
> String key/value rows use a new **`SuggestingInput`** (ranked popover under a free-text field) instead of a pick-only combobox for metadata, so unknown keys and values stay typeable. **`FacetOptions`** / **`StringKeyValueUIFilter`** gain `keyDetails` and `valueOptions`; the presentation hook derives them from `metadata` and `metadata.<key>` entries.
> 
> **Events** and **observations** tables opt in via **`useFacetOptionsWithObservedMetadata`**. **`useObservedMetadataRecorder`** runs whenever rows load (not only when the search bar is open), so every loaded table grows the shared map. Four client tests lock suggestion behavior and free-text parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b614604883e26971206129a3e95e45fa5fa1197. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/web/src/components/table/data-table-controls.tsx
+++ b/web/src/components/table/data-table-controls.tsx
@@ -589,6 +589,8 @@ export function DataTableControls({
           expanded={filter.expanded}
           loading={filter.loading}
           keyOptions={filter.keyOptions}
+          keyDetails={filter.keyDetails}
+          valueOptions={filter.valueOptions}
           value={filter.value}
           onChange={filter.onChange}
           isActive={filter.isActive}
@@ -1097,6 +1099,8 @@ interface BooleanKeyValueFacetProps extends BaseFacetProps {
 
 interface StringKeyValueFacetProps extends BaseFacetProps {
   keyOptions?: string[];
+  keyDetails?: Record<string, string>;
+  valueOptions?: Record<string, string[]>;
   value: StringKeyValueFilterEntry[];
   onChange: (filters: StringKeyValueFilterEntry[]) => void;
   keyPlaceholder?: string;
@@ -2194,6 +2198,8 @@ export function StringKeyValueFacet({
   expanded: _expanded,
   loading,
   keyOptions,
+  keyDetails,
+  valueOptions,
   value,
   onChange,
   isActive,
@@ -2222,6 +2228,8 @@ export function StringKeyValueFacet({
         <KeyValueFilterBuilder
           mode="string"
           keyOptions={keyOptions}
+          keyDetails={keyDetails}
+          valueOptions={valueOptions}
           activeFilters={value}
           onChange={onChange}
           keyPlaceholder={keyPlaceholder}
--- a/web/src/components/table/key-value-filter-builder.clienttest.tsx
+++ b/web/src/components/table/key-value-filter-builder.clienttest.tsx
@@ -31,6 +31,27 @@ describe("KeyValueFilterBuilder", () => {
     expect(screen.getAllByText("missing-key")).not.toHaveLength(0);
   });
 
+  it("keeps a facet with nothing enumerated on free text while typing a key", () => {
+    // A project with no scores offers no keys, so the pick-only combobox would
+    // trap the user: it must not appear just because the row now holds a
+    // half-typed key of its own.
+    render(
+      <KeyValueFilterBuilder
+        mode="numeric"
+        keyOptions={[]}
+        activeFilters={[]}
+        onChange={noop}
+      />,
+    );
+
+    fireEvent.click(screen.getByText("Add filter"));
+    const key = screen.getByPlaceholderText("Key");
+    fireEvent.change(key, { target: { value: "a" } });
+    fireEvent.change(key, { target: { value: "accuracy" } });
+
+    expect(screen.getByPlaceholderText("Key")).toHaveValue("accuracy");
+  });
+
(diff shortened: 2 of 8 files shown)


nkabardin:
@claude review

_(Posted by Claude Code on Nikita's behalf.)_ Rebased onto latest `main` (previous base #15881 has merged) and pushed fixes for the earlier round: blur now clears the suggestion highlight and `dismissed`, and `aria-selected` marks the keyboard-active option. Please re-review the `SuggestingInput` interaction state and the sidebar wiring.

nkabardin:
@claude review

nkabardin:
@claude review

PR #15898, opened 2026-08-07
feat(web): Update V4 migration panel

nmtrang29:
🟡 **Live preview: [pr-15898.preview.langfuse.com](https://pr-15898.preview.langfuse.com)**

Reworks the v4 migration side panel + modal into an action-items checklist, decouples API-key creation from the agent prompt, and shows the delayed-ingestion pill for every delayed path.

## What changed

| Change | Before | After |
|---|---|---|
| **Header** — outcome pitch ("real-time ingestion, up to 165× faster queries"); agent CTA moved into its own section below |  |  |
| **Action items** ([LFE-14856](https://linear.app/clickhouse/issue/LFE-14856/reframe-the-cloud-migration-panel-as-a-remaining-actions-checklist)) — "What happens if I don't update?" becomes a checklist: only sections with offenders render, clean checkers collapse into one summary line, pills → dots + count badges |  |  |
| **Split by ingestion path** ([LFE-14857](https://linear.app/clickhouse/issue/LFE-14857/split-sdk-direct-otel-and-legacy-ingestion-migration-states), [LFE-14859](https://linear.app/clickhouse/issue/LFE-14859/add-trustworthy-trace-table-evidence-links-for-migration-offenders)) — Update SDK / Update OTel Instrumentation / new Upgrade Instrumentation (headerless `/api/public/ingestion` traffic); each offender row deep-links to the observations table filtered to its API key over the 14-day detection window |  |  |
| **Auto-upgrade with agents** ([LFE-14862](https://linear.app/clickhouse/issue/LFE-14862/redesign-the-coding-agent-upgrade-section-and-cta)) — own group; two-step "Copy prompt" reveals the prompt below the button, no layout jump | *(CTA lived in the header, row 1)* |  |
| **Explicit-only API keys** ([LFE-14863](https://linear.app/clickhouse/issue/LFE-14863/make-the-coding-agent-prompt-offer-langfuse-project-access-first)) — copying the prompt no longer creates keys; a permission-gated "Create API keys" action creates them once per project and renders a copyable `.env` block; secrets never enter the prompt |  |  |
| **Compare traces while you upgrade** — replaces "Want to review first?"; info hovercard, "V4" label |  |  |
| **Footer** — contact buttons → Docs · Email an engineer · Book a call links |  |  |
| **Status page hero** — counts projects needing action (+ "could not be checked" when checks fail) |  |  |

[...]

PR #15901, opened 2026-08-07
feat(web): filter events by ingestion SDK and source

nmtrang29:
🟡 **Live preview: [pr-15901.preview.langfuse.com](https://pr-15901.preview.langfuse.com)**

## Motivation

During the v4 migration, teams need to locate which app or service is still producing incompatible events. The events table already attributes every row to its ingestion API key, SDK name, SDK version, and source path (`events_core.ingestion_api_key` / `ingestion_sdk_name` / `ingestion_sdk_version` / `source`), but only the API key was filterable. When several services share one key — or one service runs two SDK versions — the key alone cannot separate them. Beyond the migration, filtering by SDK/source is generally useful for debugging instrumentation.

## What's included

**Three new filter columns on the v4 events table**, mirroring the existing `ingestionApiKey` pattern (`stringOptions`, options loaded at runtime):

| Column id | Display name | ClickHouse column |
|---|---|---|
| `ingestionSdkName` | SDK Name | `e.ingestion_sdk_name` |
| `ingestionSdkVersion` | SDK Version | `e.ingestion_sdk_version` |
| `ingestionSource` | Ingestion Source | `e.source` |

Wired end to end:

- **Column definitions** — `packages/shared/src/eventsTable.ts` + UI column mappings in `mapEventsTable.ts` so filters lower to the right ClickHouse expressions.
- **Runtime filter options** — facet-option definitions in `event-filter-options.ts` (distinct values, count-descending, empty values excluded) and the three columns added to the events `filterOptions` service + `useEventsFilterOptions` hook (lazy-loaded like the other high-cardinality facets).
- **Sidebar facets** — three categorical facets in the events filter sidebar, next to API Key.
- **Search bar** — field registry entries (same kind/syncMode as `ingestionApiKey`: `text` / `exactOption`) with aliases `sdkname`/`sdk_name`, `sdkversion`/`sdk_version`, and `source`/`ingestion_source`/`ingestionsource`. The registry-shaped invariants harness picks the new fields up automatically.

**Tightened v4-migration evidence deep links** (stacks on #15898): each SDK usage series row in the migration panel now links to the events table filtered by **key + SDK name + SDK version** instead of key only, so two SDK versions on the same key link to distinct result sets. Unattributed dimensions (`"unknown"`) fall back to key-only — "unknown" is the attribution fallback bucket, not an exact value. The delayed-OTel `source` dimension is deliberately not linked (it is a prefix match, not an exact one). The link is now built with `encodeFiltersGeneric`, the same encoder the events page decodes with.

## Verification

- `pnpm --filter web exec tsc --noEmit -p tsconfig.json` — clean (no output, exit 0)
- `pnpm --filter web run test-client src/features/v4-migration` — `Tests  62 passed (62)`
- `pnpm --filter web run test-client src/features/search-bar` — `Tests  237 passed (237)`
- `pnpm --filter web run test src/__tests__/server/event-query-builder.servertest.ts` — `Tests  33 passed (33)` (includes 6 new cases for the option queries and filter lowering of the three columns).

## Notes

- This PR is stacked on #15898 (`v4-migration-panel-ux`, now a single squashed commit on main); merge that first.
- No schema changes: all three ClickHouse columns already exist on `events_core` (migrations 0035/0042).


hassiebp:
I'm currently unsure whether this change is strictly necessary or whether the cost-benefit trade-off makes sense here.

The main cost is that we'd be adding two facets that will rarely be used: the ingestion SDK name and ingestion SDK version. The benefit is fairly narrow: during migration, it would allow some users to correctly filter the trace table for specific observations ingested with an older SDK version.

I'd also be fine with dropping this PR for now and relying solely on the detection mechanisms in the migration modal.

@nimarb

nmtrang29:
I think it's worth adding as I can imagine it this being critical during the migration to verify that everything is working as expected. Wdyt?

PR #15906, opened 2026-08-07
feat(evals): new evaluation UX

wochinge:
🟡 **Live preview: [pr-15906.preview.langfuse.com](https://pr-15906.preview.langfuse.com)**

## Summary

- new data model + migration for it (+ temp migration suite for it)
- new eval UX UI components
- v3 UX will keep on using old UX (but based on new tables)

## Linear

- [LFE-14657](https://linear.app/clickhouse/issue/LFE-14657/production-implementation-new-eval-ux)

## Major Decisions

- **Template versions newer than the referenced one are dropped.** A version no rule points at was never exercised, so we cannot guarantee it works — migrating it would make it the evaluator's head and silently promote the rule onto it. `eval_templates` is untouched, so those versions stay recoverable if we later decide to surface them.
- **Evaluators are de-duplicated on `(template_id, variable_mapping)`**, scoped per project. The variable mapping is part of the evaluator's identity, so two rules on one template with different mappings remain separate evaluators. `project_id` is part of the key because managed templates share a single `template_id` across projects, and dropping it would merge evaluators across tenants.
- We keep the legacy UX for the old v3 UI but based on the new tables

## Review Focus

- **The backfill SQL** in `packages/shared/prisma/migrations/20260807121000_add_evaluator_v2/migration.sql`. 

## Migration Verification (Production Dump)

Ran the `add_evaluator_v2` migration against a restored production dump of `job_configurations` / `eval_templates` (layered onto seed data, with synthetic `projects`/`organizations` rows added only to satisfy foreign keys the isolated dump can't otherwise resolve), then verified with the repo's `evaluator-v2` migration test suite in dump mode (`EVALUATOR_V2_MIGRATION_DATABASE_URL`) plus targeted manual queries.

**Checks:**

| Check | Result |
|---|---|
| Every `EVAL` job configuration produces exactly one evaluation rule (1:1, no drops or duplicates) | ✅ pass |
| Assignments are created only for configurations that reference a template | ✅ pass |
| Evaluators are deduplicated correctly on `(project_id, eval_template_id, variable_mapping)` | ✅ pass |
| Every rule's evaluator carries a version matching the exact template version that rule referenced | ✅ pass |
| Migration is idempotent — rerunning `prisma migrate deploy` reports no pending migrations | ✅ pass |
| Every evaluator has at least one `evaluator_version` (no orphans) | ✅ pass |

No inconsistencies found.


web/src/features/evals/v2/managedTemplatesCatalog.ts
wochinge, line 1: @annabellscha We can just put the templates here for the gallery 
Will implement the gallery Monday / Tuesday in an initial version - you can add templates before that and then edit the gallery to your liking once the initial template gallery is there

annabellscha: Thanks @wochinge; I will do a focus block on this on Wednesday afternoon!

wochinge:
@claude review

wochinge:
Migration runtime against the restored production dump: the `add_evaluator_v2` DDL + backfill completed in a few seconds (dataset scale: tens of thousands of `job_configurations` / `eval_templates` rows).

fern/apis/server/definition/unstable/evaluation-rules.yml
wochinge, line 1: I tried to keep it mostly the same for now while exposing the potential n:m mapping.

It's unstable but I don't want to break too much now without need - let's rather aim to get a stable public API out as followup. Should be fairly simple now with the clean data model

packages/shared/prisma/migration-tests/evaluator-v2/migration.test.ts
wochinge, line 1: can be be deleted before the merge

packages/shared/prisma/migrations/20260807121000_add_evaluator_v2/migration.sql
```diff
@@ -0,0 +1,341 @@
...
+      -- would hide a template family from this backfill without ever producing an evaluator
+      WHERE jc."job_type" = 'EVAL'
+        AND referenced."project_id" = template."project_id"
+        AND referenced."name" = template."name"
+    )
+  -- only select latest version of template
+  ORDER BY
+    template."project_id", template."name", template."version" DESC,
+    template."created_at" DESC, template."id" DESC
+)
```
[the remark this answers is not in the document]
wochinge: we accept that

web/src/__tests__/server/evaluator-v2-repository.servertest.ts
```diff
@@ -0,0 +1,763 @@
...
+      ).resolves.toBe(false);
+    });
+
+    it("deletes the evaluator and cascades owned records", async () => {
+      const evaluator = await createEvaluator();
+      const { assignmentId } = await createRuleAssignment({
+        evaluatorId: evaluator.id,
+      });
+
+      await expect(
```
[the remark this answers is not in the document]
wochinge: fixed

packages/shared/src/eventsTable.ts
```diff
@@ -311,6 +311,22 @@ const eventsTableColsDefinition = [
...
+  {
+    name: "Rule ID",
+    id: "ruleId",
+    type: "stringOptions",
+    internal:
+      "if(notEmpty(arrayElement(e.metadata_values, indexOf(e.metadata_names, 'rule_id'))), arrayElement(e.metadata_values, indexOf(e.metadata_names, 'rule_id')), arrayElement(e.metadata_values, indexOf(e.metadata_names, 'job_configuration_id')))",
+    options: [],
+  },
   {
     name: "Trace Tags",
```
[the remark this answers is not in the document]
wochinge: Thanks — I do not think we should extract these expressions right now. They live in deliberately separate client-safe column definitions and server table mappings, and all four copies are currently identical and correct. Sharing them would couple those layers to guard against hypothetical future drift without fixing present behavior; the adjacent Evaluator ID expressions follow the same split. I would keep this local until there is a concrete shared abstraction across table definitions and mappings.

wochinge:
@claude review

wochinge:
@claude review

packages/shared/prisma/schema.prisma
```diff
@@ -898,6 +904,56 @@ model EvalTemplate {
...
+  modelParams        Json?                        @map("model_params")
+  vars               String[]                     @default([])
+  variableMapping    Json?                        @map("variable_mapping")
+  outputDefinition   Json?                        @map("output_definition")
+  sourceCode         String?                      @map("source_code") @db.VarChar(262144)
+  sourceCodeLanguage EvaluatorSourceCodeLanguage? @map("source_code_language")
+
+  @@unique([evaluatorId, version(sort: Desc)])
+  @@map("evaluator_versions")
+}
```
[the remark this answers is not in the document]
wochinge: evaluators shouldn't be able to change their type

wochinge:
@claude review

wochinge:
@claude review

wochinge:
@claude review

web/.storybook/main.ts
```diff
@@ -22,6 +23,15 @@ const DESIGN_COMPONENT_STORIES = [
...
 // Design/Components): the token reference, one single-leaf page per element.
+// Directories that get their own sidebar section instead of the flat
+// Playground default. This is not a `stories` entry with a `titlePrefix`
+// because story titles are injected into each meta — see StoryTitleGroup.
+const STORY_TITLE_GROUPS: StoryTitleGroup[] = [
+  {
+    directory: "src/features/evals/v2/components",
+    titlePrefix: "Features/Evaluations",
+  },
+];
```
bezbac, line 34: Nice!

bezbac (commented):
I got auto-tagged because I added myself as a code-owner for storybook setup changes in https://github.com/langfuse/langfuse/pull/16420. The relevant changes here look good!

dgenio:
One versioning edge case seems worth making explicit before this lands: evaluator-v2 jobs are intentionally **not pinned to a version at dispatch**. `scheduleObservationEvals` queues only the evaluator id, and `observationEvalProcessor` resolves `versions[0]` at pickup. The execution metadata correctly records the version that actually ran, which is good for auditability, but it does not preserve one scoring definition for an already-started batch/backlog.

Concrete case: start a manual batch (or accumulate queued rule jobs) under evaluator v3, then save v4 while some jobs are still pending. Early items can be scored by v3 and later items by v4 even though they belong to the same logical batch/backlog. Per-score `evaluatorVersionId` tells us what happened afterwards, but aggregates over that run can now mix two rubric/code definitions.

Is that `latest-at-pickup` behavior intentional? If not, I think the clean invariant is to resolve/pin `evaluatorVersionId` when the batch/job is scheduled and execute that immutable version. If it *is* intentional for continuous rules, it may still be worth pinning manual batch runs (or at least ensuring aggregates are version-aware) so one evaluation run cannot silently mix scorer revisions.

wochinge:
> Is that `latest-at-pickup` behavior intentional? 

Yes it is :-)

PR #15907, opened 2026-08-07
feat(ci): select affected Vitest tests

wochinge:
🟡 **Live preview: [pr-15907 app preview](https://pr-15907.preview.langfuse.com)**

## Summary

- Select shared, web, worker, and Docker jobs in a checkout-only planning job before dependency setup or builds, then run Vitest with `--changed` inside the selected package jobs.
- Treat CI, dependency, Docker, package, Vitest, TypeScript, and Turbo configuration changes as full-suite invalidators.
- Map shared-package source changes to built CommonJS entrypoints so Vitest can conservatively select affected web, worker, and shared tests across package boundaries.
- Cover the selection policy with focused tests, including a safe full-suite fallback when Git comparison fails.

## Linear

- [LF-84](https://linear.app/clickhouse/issue/LF-84/explore-test-selection-based-on-ast)

## Major Decisions

- Use Vitest's dependency graph and custom VCS provider instead of maintaining a separate AST dependency graph.
- Gate expensive jobs with the early package-level selection while retaining Vitest's finer file-level dependency selection.
- Prefer false positives for shared-package changes by exposing concrete public entrypoints in addition to the mirrored build output.
- Use immutable pull-request and merge-group base SHAs; pushes continue to run every test.

## Review Focus

- Whether the full-suite invalidator list covers every repository-level input that can affect tests.
- Whether the conservative shared-package expansion is the right balance between safety and test reduction.
- Merge-queue base-SHA handling and the fail-safe behavior when Git metadata is unavailable.


wochinge:
Addressed the checkout-depth nit in f2db68a8b: PR and merge-group runs use full history for the triple-dot comparison, while push/workflow-dispatch runs keep depth 1. I used string '0'/'1' inputs because numeric 0 is falsy in GitHub expressions.

wochinge:
@claude review

PR #15925, opened 2026-08-09
fix(shared): add jitter to redis reconnect backoff

### LINKED ISSUE #15855: Worker: Redis reconnect storm — connections retry without jitter, ~4,800 log lines from a 15s outage
### Describe the bug

When the Redis host becomes temporarily unresolvable, `langfuse-worker` emits a
very large burst of identical connection errors in a few seconds. The per-connection
backoff appears to work correctly; what seems to be missing is **jitter**, so all
of the worker's Redis connections retry in lockstep.

I hit this on a self-hosted Docker Compose deployment where Redis is a sibling
container. A routine image bump recreates the Redis container, its DNS name
disappears for ~15 seconds, and the worker produces roughly 4,800 log lines in
that window.

### Measurements

From a single incident (Redis unavailable 16:30:43 → 16:30:58, ~15s):

- ~4,800 lines total, **2,430 within one minute**
- **20 errors inside a 35ms span**:

```
16:30:50.253  16:30:50.253  16:30:50.254  16:30:50.257  16:30:50.260
16:30:50.262  16:30:50.263  16:30:50.268  16:30:50.269  16:30:50.270
```

Each failure logs two lines:

```
Error: getaddrinfo ENOTFOUND langfuse-redis
<timestamp> error   Redis error getaddrinfo ENOTFOUND langfuse-redis
```

Individual connections do space their retries (~1.2s apart, consistent with
ioredis' default `retryStrategy`). The volume comes from many connections — the
worker appears to hold roughly one per BullMQ queue — all failing at the same
instant and therefore all retrying at the same instant.

### Expected behaviour

A brief Redis outage should produce a bounded amount of log output. Adding
jitter to the ioredis `retryStrategy` would spread reconnects out, and
optionally collapsing repeated identical connection errors would help further.

### Why this is awkward to work around

Self-hosting exposes `REDIS_COMMAND_TIMEOUT`, `REDIS_REQUEST_SOCKET_TIMEOUT_MS`
and `REDIS_BLOCKING_SOCKET_TIMEOUT_MS`, but nothing for retry or backoff policy,
so `retryStrategy` cannot be adjusted from outside the image. I have worked
around it by rate-limiting these lines in my log shipper, which is not ideal
since it also caps genuine outages.

Happy to test a patch or provide more data.

### Environment

- Langfuse: **4.1.0** (self-hosted, Docker Compose)
- Redis: 7.4.10
- Deployment: `langfuse-worker` + `langfuse-web`, Redis as a sibling container on a user-defined bridge network

Diwak4r:
## What

Adds randomized jitter to the reconnect backoff of the shared BullMQ/queue Redis connections (`redisQueueRetryOptions.retryStrategy` in `packages/shared/src/server/redis/redis.ts`) so that, during a Redis outage, the per-queue connections do not retry in lockstep. Fixes #15855 (self-hosted Docker Compose: ~4,800 identical reconnect log lines within a 15s DNS blip, 20 errors inside a 35ms span).

The exponential base delay stays clamped to 1s–20s; each retry adds up to 50% random jitter, so the effective maximum is 30s.

An earlier revision of this PR also deduplicated the `reconnectOnError` warning. That part was dropped after review: ioredis only invokes `reconnectOnError` for server error replies (`READONLY`, `MOVED`), never for the DNS/socket failures that make up a reconnect storm, so it could not reduce the reported volume.

Note for cluster-mode deployments (including Langfuse Cloud): ioredis's cluster connection pool creates node connections with `retryStrategy: null` and relies on `clusterRetryStrategy` instead, so this change only affects single-node and Sentinel deployments.

## Impacted packages

* `@langfuse/shared` (`packages/shared/src/server/redis/redis.ts`, `packages/shared/src/server/redis/redis.test.ts`)

## Verification

* `pnpm --filter @langfuse/shared run test src/server/redis/redis.test.ts` → `Tests 25 passed (25)` (jitter floor/cap/spread + `reconnectOnError` default/READONLY/MOVED behaviour)
* `pnpm --filter @langfuse/shared run lint` → exit 0; pre-commit `turbo run lint` → `Tasks: 7 successful, 7 total`
* `pnpm --filter @langfuse/shared run build:check` (tsc) → exit 0
* prettier → clean

## AI disclosure

Authored with AI assistance (original contribution by @Diwak4r; maintainer cleanup to jitter-only).


packages/shared/src/server/redis/redis.ts
```diff
@@ -37,7 +47,15 @@ export const redisQueueRetryOptions: Partial<RedisOptions> = {
...
-    logger.warn(`Redis connection error: ${err.message}`);
+    const now = Date.now();
+    if (
+      !lastReconnectErrorLog ||
+      lastReconnectErrorLog.message !== err.message ||
+      now - lastReconnectErrorLog.at >= REDIS_RECONNECT_ERROR_LOG_INTERVAL_MS
+    ) {
+      logger.warn(`Redis connection error: ${err.message}`);
+      lastReconnectErrorLog = { message: err.message, at: now };
+    }
```
Steffen911, line 58: I think that having the jitter should be sufficient. Do you see a strong reason to reduce the log volume?

Steffen911:
@claude review

Steffen911:
@Diwak4r Could you take a look at the comments? Tests look good and we can rerun them once the pending reviews are addressed.

Diwak4r:
Addressed the review feedback in 71e6f5f:

**Per-message dedupe (@claude / greptile P1 finding — correct):** The single `lastReconnectErrorLog` slot indeed failed when error messages alternated (A-B-A-B) across the ~40 concurrent queue connections, since each new message overwrote the slot and got logged. Suppression is now keyed per distinct message in a `Map<string, number>`, so back-to-back repeats *and* alternating errors collapse correctly. Verified by a regression test that drives an ENOTFOUND/ECONNREFUSED alternation and asserts one log per distinct message.

**On your question about whether reducing log volume is needed given jitter (@Steffen911):** Jitter solves the lockstep *retry* timing, but not log volume: every queue connection still runs its own reconnect attempt and calls `reconnectOnError`, so during a sustained outage with N connections you still get ~N identical warnings per retry cycle. The dedupe reduces that to one line per interval regardless of connection count. I'd argue it's worth keeping for outage readability, but it's independent of the jitter change — happy to split or drop that part if you prefer.

**Bounded state:** the map is capped at 100 keys with expiry-based pruning so a flood of unique messages can't grow it unbounded.

Tests: 29/29 pass locally (`vitest run src/server/redis/redis.test.ts`), prettier clean.

Steffen911:
@Diwak4r I think the deduplication is acceptable and I'd rather merge this as jitter only.

Steffen911:
@claude review

PR #15927, opened 2026-08-09
fix(otel): only treat parentless spans as the trace root

### LINKED ISSUE #15860: Trace-level input/output/name overwritten by last-ingested LangChain root when multiple processes share one trace id (distributed multi-agent)
### Describe the bug

In a distributed multi-agent setup, several independent processes contribute observations to the same trace id. Each process runs its own `langfuse.langchain.CallbackHandler` against a LangGraph app, and downstream processes join the caller's trace using `trace_context={"trace_id", "parent_span_id"}`.
 
The observation tree is correct: one root span, and each downstream agent's spans are correctly parented under the caller's span with the shared trace id. But the trace-level `input`, `output`, and `name` shown on the Trace record are taken from whichever process's LangChain root chain is ingested last, not from the process that actually owns the root of the trace.
 
I confirmed on the raw OTLP spans that only the originating (root) process emits trace-level attributes:
 
- Root process span has `langfuse.trace.input`, `langfuse.trace.output`, and `langfuse.trace.name` set to the originating request/response.
- Downstream (delegated) process spans do **not** emit `langfuse.trace.input`/`output`.
 
Despite that, the resulting Trace record's `input`/`output` are the downstream specialist's request/response, and without a propagated trace name the trace `name` also becomes the downstream chain's name.
 
Each process-local LangGraph root observation carries `metadata.is_langchain_root = true`. It appears the backend derives trace-level `input`/`output`/`name` from a LangChain-root observation regardless of whether that observation is the true root of the trace id (i.e., regardless of `parent_span_id` linking it beneath another process's span). The last such observation to be ingested wins, which in a fan-out is a non-deterministic specialist.
 

### Steps to reproduce

1. Stand up two services, A (orchestrator) and B (specialist), each with its own process, OTel `TracerProvider`, and `langfuse.langchain.CallbackHandler`. Export via OTLP to a collector that forwards to Langfuse with `x-langfuse-ingestion-version: 4`.
2. In A, start a root observation and set trace IO, e.g. `input = "user question"`, then run a LangGraph graph via the callback handler.
3. A calls B over HTTP. Propagate the trace so B joins the same trace id, either through W3C `traceparent` (OTel context) or by constructing B's handler with `trace_context={"trace_id": <A trace id>, "parent_span_id": <A current span id>}`.
4. In B, run a second LangGraph graph via its own `CallbackHandler`. Do not set any trace-level IO in B.
5. Open the trace in Langfuse.
 
**Expected:** the trace `input`/`output`/`name` reflect A (the process that owns the trace root).
 
**Actual:** the trace `input`/`output` reflect B's LangGraph root chain (the delegated specialist), and the trace `name` becomes B's chain name unless a trace name is explicitly propagated. The span tree itself is correct (B nested under A).
 
Notes that narrow it down:
 
- Only A emits `langfuse.trace.input`/`output` on the wire; B does not. The backend still shows B's IO.
- B's LangGraph root observation has `parent_span_id` pointing at A's span and `metadata.is_langchain_root = true`.
- Fan-out (A calls B and C concurrently) makes the chosen IO non-deterministic across runs, consistent with last-ingested-wins.
 

Langfuse Cloud

Self-hosted, OSS v3.221.1.

- langfuse (Python SDK): 4.14.3
- Integration: `langfuse.langchain.CallbackHandler` (OpenTelemetry-native, no direct REST; spans exported via OTLP/HTTP through an OpenTelemetry Collector, `x-langfuse-ingestion-version: 4`)
- langchain: 1.3.14
- langchain-core: 1.5.3
- langgraph: 1.x
- opentelemetry-sdk: 1.44.x
- Python: 3.12

### Additional information

- Cross-service trace joining works well otherwise: single trace id shared across processes, correct parent/child nesting, no phantom or unnamed trace when a trace name is propagated.
- Workarounds I tried and their outcomes:
  - Propagating `trace_name` via OTel baggage and re-applying it downstream fixes the trace `name`, but not `input`/`output`.
  - Constructing the downstream `CallbackHandler` with explicit `trace_context` does not prevent the downstream root chain's IO from becoming the trace IO.
  - Forcing delivery order (flushing the delegated process before the root finishes) does not help, which suggests the selection is not purely delivery-order based at write time.
- Desired behavior: trace-level `input`/`output`/`name` should be owned by the observation that is the actual root of the trace id (the observation with no `parent_span_id` within that trace), not by any observation merely flagged `is_langchain_root` in its own process. Alternatively, expose a supported way for an integration to opt a process-local LangChain root out of trace-level IO promotion when it is joining an existing trace.

No

Diwak4r:
## What

Fixes #15860: in a distributed multi-agent setup, trace-level `name`/`input`/`output` were taken from whichever process-local LangChain root was ingested last, not from the process that owns the trace root.

Root cause chain:
- When a process joins an existing trace (`trace_context` with `trace_id`/`parent_span_id`), the SDK starts its root chain under a remote parent span and marks it `langfuse.internal.as_root = true`. That span always carries a real `parentSpanId` (the foreign span it attaches under).
- The OTel ingestion processor computed `isRootSpan = !parentObservationId || as_root === "true"`, so the joining chain counted as a root and emitted a `trace-create` event.
- `createTraceEvent` promotes observation-level IO onto the trace record (trace-domain `extractInputAndOutput` falls back to `langfuse.observation.input/output`), and the trace name falls back to the span name. Since trace records are keyed by trace id, the last-ingested joining process overwrote the true root's name and IO.

The fix: a span with a parent span of its own is never the trace root, regardless of `as_root`. Joining processes still attach observations under their parent and can still apply explicit trace updates (`session.id`, `langfuse.trace.*` attributes) via the existing `hasTraceUpdates` path, but they can no longer overwrite the root's name and IO with their chain-level data.

## Impacted packages

- `@langfuse/shared` (`packages/shared/src/server/otel/OtelIngestionProcessor.ts`, new `packages/shared/src/server/otel/OtelIngestionProcessor.traceRoot.test.ts`)

## Verification

- New regression tests (3): true root keeps name/IO when the joining process is ingested after it; ingestion order does not matter (no last-wins); explicit trace updates from a joining process still apply without clobbering the root. All 3 pass with the fix; the first two fail without it (verified by temporarily reverting).
- Full OTel test directory in the shared package: 88/88 passed (no regressions).
- `tsc --noEmit -p packages/shared/tsconfig.json`: no errors in changed files.
- `prettier --check` on both changed files: clean.

## AI disclosure

Authored with AI assistance. The SDK-side behavior (`as_root` under a remote parent) was confirmed against langfuse-python v4.14.3 before changing the backend rule.

The PR changes OTel trace-root selection so only parentless spans can promote observation-level fields to a trace, preventing process-local roots in distributed traces from overwriting the actual trace owner.

- Removes `langfuse.internal.as_root` from the trace-root predicate when a span has a parent.
- Preserves explicit trace updates from joining processes through the existing update path.
- Adds regression coverage for ingestion ordering, root name and IO preservation, and explicit session updates.
</details>

The PR appears safe to merge, with no concrete changed-code failure identified.

The new predicate aligns trace ownership with the OTel parent relationship, while the existing non-root update path continues to apply explicit trace attributes without promoting joining spans’ observation-level name or IO.
</details>

```mermaid
flowchart TD
  Span[Ingest OTel span] --> Parent{Has parent span ID?}
  Parent -->|No| Root[Create root trace event]
  Root --> Promote[Promote trace name and IO]
  Parent -->|Yes| Updates{Has explicit trace updates?}
  Updates -->|Yes| Update[Create trace update event]
  Updates -->|No| Shallow[Create shallow event only when trace is unseen]
  Update --> Preserve[Do not promote observation-level name or IO]
  Shallow --> Preserve
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(otel): only treat parentless spans a..."](https://github.com/langfuse/langfuse/commit/937b558063746d2ba1d191295f964c544ff43d0b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51602917)</sub>

### DIFF
--- a/packages/shared/src/server/otel/OtelIngestionProcessor.traceRoot.test.ts
+++ b/packages/shared/src/server/otel/OtelIngestionProcessor.traceRoot.test.ts
@@ -1,0 +1,216 @@
+/**
+ * Regression tests for the trace-root selection in the OTel ingestion path.
+ *
+ * Issue: https://github.com/langfuse/langfuse/issues/15860
+ *
+ * In a distributed setup several processes contribute observations to one
+ * trace id. Each process-local LangChain root chain is started under a remote
+ * parent (the SDK joins the foreign trace via trace_context and sets
+ * langfuse.internal.as_root on the span), so it carries a parent span id but
+ * is still flagged as a "root". Treating such spans as the trace root promoted
+ * their observation-level input/output onto the trace record, so the trace
+ * name and IO came from whichever joining process was ingested last.
+ *
+ * The fix: a span with a parent span of its own is never the trace root, so
+ * joining processes no longer overwrite the true root's name and IO.
+ */
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+const { recordIncrementMock } = vi.hoisted(() => ({
+  recordIncrementMock: vi.fn(),
+}));
+
+vi.mock("../instrumentation", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../instrumentation")>();
+  return {
+    ...actual,
+    recordIncrement: recordIncrementMock,
+  };
+});
+
+// processToIngestionEvents awaits redis.set (seen-traces tracking); CI's
+// tests-shared job has REDIS_HOST set but no Redis server, so ioredis
+// queues the command forever and the suite times out. Stub the client.
+vi.mock("../redis/redis", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../redis/redis")>()),
+  redis: { set: vi.fn().mockResolvedValue("OK") },
+}));
+
+import {
+  OtelIngestionProcessor,
+  type ResourceSpan,
+} from "./OtelIngestionProcessor";
+
+const PROJECT_ID = "test-project-trace-root";
+
+const createProcessor = () =>
+  new OtelIngestionProcessor({
+    projectId: PROJECT_ID,
+    publicKey: "pk-test",
+    sdkName: "python",
+    sdkVersion: "4.14.3",
+  });
+
+type OtelAttribute = { key: string; value: Record<string, unknown> };
+
+const TRACE_ID = Buffer.from("0123456789abcdef0123456789abcdef", "hex");
+const ROOT_SPAN_ID = Buffer.from("0123456789abcdef", "hex");
+const JOIN_SPAN_ID = Buffer.from("fedcba9876543210", "hex");
+
+const buildSpan = ({
+  spanId,
+  parentSpanId,
+  name,
+  attributes,
+}: {
+  spanId: Buffer;
+  parentSpanId?: Buffer;
+  name: string;
+  attributes: OtelAttribute[];
+}): ResourceSpan[] => [
+  {
+    resource: {
+      attributes: [{ key: "service.name", value: { stringValue: "test-svc" } }],
+    },
+    scopeSpans: [
+      {
+        scope: {
+          name: "langfuse-sdk",
+          version: "4.14.3",
+          attributes: [
+            { key: "public_key", value: { stringValue: "pk-test" } },
+          ],
+        },
+        spans: [
+          {
+            traceId: TRACE_ID,
+            spanId,
+            parentSpanId,
+            name,
+            kind: 1,
+            startTimeUnixNano: "1752384000000000000",
+            endTimeUnixNano: "1752384001000000000",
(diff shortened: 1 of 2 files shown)


Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #15936, opened 2026-08-10
feat(web): self-hosting newsletter opt-in step in signup onboarding

jannikmaierhoefer:
🟡 **Live preview: [pr-15936.preview.langfuse.com](https://pr-15936.preview.langfuse.com)**

## What does this PR do?

Adds a post-signup onboarding step for **self-hosted** deployments that offers an opt-in to the Langfuse self-hosting newsletter ("Langfuse OSS updates"). Langfuse Cloud is unchanged.

Self-hosters previously saw **no onboarding at all**: both `pages.newUser` ([auth.ts](https://github.com/langfuse/langfuse/blob/main/web/src/server/auth.ts)) and the credentials `callbackUrl` ([sign-up.tsx](https://github.com/langfuse/langfuse/blob/main/web/src/pages/auth/sign-up.tsx)) routed to `/onboarding` only when `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` was set. So there was no in-product path to release announcements — self-hosters had to discover them on GitHub. Onboarding now runs on both deployment types and `onboarding.tsx` picks the variant: Cloud keeps the attribution question, self-hosted gets the newsletter step.

Fixes # (no issue)

### Why the signup is proxied through langfuse.com

Subscribing to a Loops mailing list requires a Loops API key, which a self-hosted instance must never carry. The marketing site already exposes the proxy its own signup forms use and holds the key, so an instance only sends an email address and a source label:

```
POST https://langfuse.com/api/productUpdateSignup   { email, list: "oss", source: "self-host-onboarding" }
```

Two deliberate details:
- The list is referenced **by name** (`list: "oss"`), never by id, so an instance cannot inject an arbitrary Loops list. No Loops identifier lives in this repo.
- The call is made **server-side**. The proxy sends no CORS headers, so a cross-origin browser request would be blocked — and this also keeps the user's browser from contacting langfuse.com at all.

### Airgap handling — no connectivity probing

1. `TELEMETRY_ENABLED=false` is read as the operator having opted out of calls to Langfuse-operated services, so the form is not offered.
2. Otherwise reachability is discovered **on explicit submit**. A failed request swaps in a link to the hosted form and never blocks onboarding.

Reachability is deliberately *not* probed on page load: `public.checkUpdate` shows what that costs on an airgapped instance, where it logs a failure on every layout mount. The unreachable path logs at `info`, not `error`, since it is expected for those deployments.

### New environment variable

```bash
LANGFUSE_DISABLE_SIGNUP_ONBOARDING=true   # default: false
```

For operators who provision users centrally and want no interstitial. A runtime **server** variable rather than `NEXT_PUBLIC_*` on purpose: self-hosters run the prebuilt Docker image, where `NEXT_PUBLIC_*` values are baked at build time and so could not be set by the audience this targets. It gates two places off one predicate (`isSignupOnboardingEnabled`): NextAuth's `newUser` page, so SSO/OAuth users are never routed there; and `onboarding.status`, which reports a disabled flow exactly like an already-completed one. Nothing is recorded when disabled.

### Storage

The opt-in rides in the existing free-form `Survey.response` JSON under the existing `USER_ONBOARDING` name — **no migration**. `newsletterOptIn: false` distinguishes declining from never being asked.

### Refactor included

Shared completion logic moved into `useCompleteOnboarding` and the card shell into `OnboardingCard` so the two variants don't duplicate ~90 lines. `completeCloudSignupOnboarding` / `getCloudSignupOnboardingStatus` were renamed to drop the now-inaccurate "Cloud".

## Type of change

## Mandatory Tasks

## Reviewer notes

**Verified locally** against a self-hosted-mode production build:

- `Tests 20 passed (20)` server-unit, `Tests 7 passed (7)` client, `Tasks: 7 successful, 7 total` for both `lint` and `typecheck`
- A **real subscribe** was exercised end-to-end against the production endpoint with a throwaway address, confirming langfuse.com already maps `list: "oss"` and accepts the new `source`. No `[onboarding.newsletter]` warn/info fired.
- The `TELEMETRY_ENABLED=false` fallback renders link-only with no dead form; `LANGFUSE_DISABLE_SIGNUP_ONBOARDING=true` bounced `/onboarding` straight to the workspace with **0** survey rows written afterwards.
- Light and dark themes both checked.

**Follow-ups intentionally not in this PR:**

- **Docs**: `LANGFUSE_DISABLE_SIGNUP_ONBOARDING` and the new step need entries in the self-hosting docs, plus the `oss-newsletter` page itself (separate `langfuse-docs` repo).
- The self-hosted step *replaces* rather than adds to the attribution question — self-hosters get no referral question. Say the word if that is wrong.
- No PostHog event: self-hosted images bake no PostHog key, and the Cloud survey is not instrumented either.
- The "set initial password" path (`ResetPasswordPage.tsx`) still sends self-hosters to `/`; left alone as out of scope.
- **The fallback link target must ship first.** `NEWSLETTER_SIGNUP_FALLBACK_URL` points at `langfuse.com/self-hosting/oss-newsletter`, the dedicated signup page. That page currently **404s** — it needs to be live before this merges, or the airgapped fallback sends users to a dead link.

**One thing worth a second opinion:** the disable flag applies wherever it is set, including Cloud. I chose predictability ("set this → no onboarding step") over a Cloud carve-out that would silently no-op, but that is a judgement call.

Adds self-hosted signup onboarding with an OSS newsletter opt-in, while retaining the existing Cloud attribution survey.

- Routes new credentials and OAuth/SSO users through deployment-specific onboarding.
- Adds a server-side newsletter proxy call with an airgap fallback and telemetry-aware availability.
- Adds a runtime operator flag to disable onboarding and refactors shared completion/card behavior.
- Persists newsletter decisions in the existing onboarding survey response.
</details>

<details><summary><h3>Confidence Score: 2/5</h3></summary>

This PR should not merge until newsletter enrollment is bound to valid recipient consent and newsletter-status failures are handled separately from an intentional opt-out.

The new authenticated endpoint can submit arbitrary third-party addresses to the mailing-list proxy, while a transient availability-query failure is misrepresented as disabled signup and can persist an incorrect decline.

**Files Needing Attention:** web/src/features/onboarding/server/onboardingRouter.ts; web/src/features/onboarding/components/SelfHostedNewsletterStep.tsx
</details>

The authenticated newsletter mutation accepts an arbitrary email address and forwards it to the external mailing-list proxy without binding it to the session identity or otherwise establishing recipient consent.
</details>

```mermaid
sequenceDiagram
  participant U as New user
  participant UI as Onboarding page
  participant RPC as Onboarding tRPC router
  participant Proxy as langfuse.com signup proxy
  participant DB as Survey storage

  U->>UI: Complete signup
  UI->>RPC: Query onboarding/newsletter status
  alt Cloud deployment
    UI->>U: Show attribution survey
    U->>RPC: Complete with referral source
  else Self-hosted and available
    UI->>U: Show newsletter opt-in
    U->>RPC: Subscribe with email
    RPC->>Proxy: "POST email, list=oss, source"
    Proxy-->>RPC: Subscription status
    U->>RPC: Complete with newsletterOptIn
  else Disabled or unreachable
    UI->>U: Show fallback/continue
    U->>RPC: "Complete with newsletterOptIn=false"
  end
  RPC->>DB: Persist USER_ONBOARDING response
  RPC-->>UI: Redirect target
```
</details>

`````markdown
### Issue 1
web/src/features/onboarding/server/onboardingRouter.ts:31-35
**Unbound newsletter recipient consent**

When an authenticated user supplies another person's email, this mutation forwards it to the external mailing-list proxy without binding it to `ctx.session.user.email` or requiring recipient confirmation, causing an unrelated address to be enrolled without the recipient initiating the request.

**How this was verified:** The authenticated procedure accepts `input.email` and passes it directly to the external signup service without using the session email.

### Issue 2
web/src/features/onboarding/components/SelfHostedNewsletterStep.tsx:160-161
**Status errors become false declines**

When the `newsletterStatus` query fails after its retries, `data` remains undefined and this expression treats the failure as disabled signup. The user sees the operator-disabled fallback, and clicking Continue permanently records `newsletterOptIn: false` even though the opt-in form failed to load.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(web): explain the newsletter step w..."](https://github.com/langfuse/langfuse/commit/5f739b6b6f3c324f71e59a44b627c695b95bcec0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54675721)</sub>

**Context used:**

- Knowledge Base — [Authentication, Multi-Tenancy, and RBAC](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/auth-rbac.md)

### DIFF
--- a/.env.dev.example
+++ b/.env.dev.example
@@ -231,6 +231,10 @@ LANGFUSE_BACKGROUND_MIGRATION_V4_ENABLE_HISTORIC_BACKFILL=false
 # Legacy tracing UI controls
 LANGFUSE_DISABLE_LEGACY_TRACING_IO_SEARCH=false
 
+# Set to `true` to skip the post-signup onboarding step and send new users
+# straight to their workspace.
+LANGFUSE_DISABLE_SIGNUP_ONBOARDING=false
+
 CLICKHOUSE_USE_LIGHTWEIGHT_UPDATE="true"
 # `auto` detects affected ClickHouse versions on startup. Set `true` to force
 # the workaround or `false` to force-disable it.
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -277,6 +277,11 @@ LANGFUSE_ENABLE_BLOB_STORAGE_FILE_LOG=true
 # Disable input/output full-text search in legacy V3 tracing tables (metadata search remains available)
 # LANGFUSE_DISABLE_LEGACY_TRACING_IO_SEARCH=false
 
+# Onboarding
+# Skip the post-signup onboarding step (self-hosting newsletter opt-in) and send
+# new users straight to their workspace. No survey response is recorded.
+# LANGFUSE_DISABLE_SIGNUP_ONBOARDING=false
+
 ### START Enterprise Edition Configuration
 
 # Allowlisted users that can create new organizations, by default all users can create organizations
--- a/web/src/__tests__/server/unit/newsletterService.servertest.ts
+++ b/web/src/__tests__/server/unit/newsletterService.servertest.ts
@@ -1,0 +1,117 @@
+const { envMock } = vi.hoisted(() => ({
+  envMock: {} as Record<string, unknown>,
+}));
+
+vi.mock("@/src/env.mjs", () => ({
+  env: envMock,
+}));
+
+import { logger } from "@langfuse/shared/src/server";
+import {
+  isNewsletterSignupAvailable,
+  subscribeToNewsletter,
+} from "@/src/features/onboarding/server/newsletterService";
+
+const setEnv = (values: Record<string, unknown>) => {
+  for (const key of Object.keys(envMock)) delete envMock[key];
+  Object.assign(envMock, values);
+};
+
+describe("newsletterService", () => {
+  let fetchMock: ReturnType<typeof vi.fn>;
+  // Spied rather than module-mocked: the shared test teardown imports this same
+  // logger, so replacing the module would break it.
+  let loggerMock: {
+    info: ReturnType<typeof vi.spyOn>;
+    warn: ReturnType<typeof vi.spyOn>;
+    error: ReturnType<typeof vi.spyOn>;
+  };
+
+  beforeEach(() => {
+    setEnv({});
+    fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
+    loggerMock = {
+      info: vi.spyOn(logger, "info").mockReturnValue(logger),
+      warn: vi.spyOn(logger, "warn").mockReturnValue(logger),
+      error: vi.spyOn(logger, "error").mockReturnValue(logger),
+    };
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  describe("isNewsletterSignupAvailable", () => {
+    it("offers the in-product signup on a plain self-hosted instance", () => {
+      setEnv({});
+      expect(isNewsletterSignupAvailable()).toBe(true);
+    });
+
+    it("does not offer it on Langfuse Cloud", () => {
+      setEnv({ NEXT_PUBLIC_LANGFUSE_CLOUD_REGION: "EU" });
+      expect(isNewsletterSignupAvailable()).toBe(false);
+    });
+
+    it("honors an operator who disabled the telemetry ping", () => {
+      setEnv({ TELEMETRY_ENABLED: "false" });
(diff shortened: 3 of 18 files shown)


jannikmaierhoefer:
@claude review

Steffen911:
@jannikmaierhoefer Do you know who built the initial onboarding questionnaire? This is a bit much frontend for me to review comfortably. Or do you have sections that seem especially relevant that we should focus on?

PR #15965, opened 2026-08-10
feat(web): make markdown render character limit configurable

DavidTraina:
## What does this PR do?

Makes the markdown render character limit configurable for self-hosters via a new `NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT` environment variable (zod-validated positive integer), defaulting to the current hardcoded `150000` — no behavior change unless explicitly set.

**Motivation:** the previous `MARKDOWN_RENDER_CHARACTER_LIMIT` constant gates markdown rendering of trace/observation I/O (consumed by `IOPreviewPretty.tsx`, `MarkdownJsonView.tsx`, and `PrettyJsonView.tsx`). `IOPreviewPretty` applies it to the combined size of input + output + all messages, so self-hosted instances tracing agents with large system prompts (e.g. an 83 KB system prompt repeated across messages) exceed it easily, and the pretty chat view silently falls back to plain text with no way to opt back into markdown. Self-hosters who are willing to accept slower `react-markdown` rendering for large payloads currently have no recourse.

**Implementation** follows the existing client env var pattern (e.g. `NEXT_PUBLIC_LANGFUSE_PLAYGROUND_STREAMING_ENABLED_DEFAULT`, `NEXT_PUBLIC_BASE_PATH`):

- `web/src/env.mjs`: new client var, `z.coerce.number().int().positive().default(150_000)`, plus the `runtimeEnv` entry.
- Consumers (`IOPreviewPretty.tsx`, `MarkdownJsonView.tsx`, `PrettyJsonView.tsx`) read `env.NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT` directly, matching how other `NEXT_PUBLIC_*` vars are consumed. `web/src/utils/constants.ts` (which existed solely to hold this one constant) is deleted rather than kept as an env re-export.
- `web/src/components/ui/markdown-render-limits.ts`: the per-string byte preempt `MARKDOWN_MAX_RENDER_BYTES` is now derived as `Math.max(150_000, <limit>)`, so raising the limit also lets individual strings above 150 KB render as markdown instead of being force-downgraded to plain text. This is safe per the module's own analysis: flat byte size never overflows the stack; the stack-overflow guard is the separate, fixed `MARKDOWN_MAX_NESTING_DEPTH` check, which is unchanged. The floor means the backstop can never be lowered below today's value.
- `web/Dockerfile`: `ARG`/`ENV` pass-through in the builder stage, per the "Also add these to web/Dockerfile" note in `env.mjs`.
- `.env.prod.example`: commented entry with the same build-time caveat as `NEXT_PUBLIC_BASE_PATH`.

**Notes for reviewers:**

- Like all `NEXT_PUBLIC_*` vars, this is inlined at build time, so it takes effect for source builds and custom images (`docker build --build-arg NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT=...`) — same limitation as `NEXT_PUBLIC_BASE_PATH`. Prebuilt images keep the `150000` default. If you'd prefer runtime plumbing so prebuilt images can configure it too, happy to rework this.
- Tests: `markdown-render-limits.clienttest.ts` gains two cases covering the `MARKDOWN_MAX_RENDER_BYTES` derivation (a raised limit raises the backstop and is honored by `exceedsMarkdownRenderLimits`; a lowered limit is floored at `150_000`), using the repo's existing `vi.mock("@/src/env.mjs", ...)` client-test pattern. The existing `PrettyJsonView.gate.clienttest.tsx` and `markdown-render-limits.clienttest.ts` suites pass unchanged with the default value.
- Docs: happy to open a companion PR to `langfuse-docs` documenting the variable if this lands.

## Type of change

## Mandatory Tasks

## Checklist

- I have added tests that prove my fix is effective or that my feature works (backstop-derivation cases in `markdown-render-limits.clienttest.ts`)

The PR replaces the fixed markdown rendering threshold with a validated public build-time setting while retaining 150000 as the intended default.

- Wires the setting through the environment schema, Docker builder, and example production configuration.
- Updates trace and JSON markdown gates to consume the setting.
- Derives the per-string markdown backstop from the configured threshold and adds derivation tests.
- Deletes the superseded constants module.
</details>

The default Docker deployment path should not be merged until omitting the new build argument reliably preserves the 150000 markdown limit.

Docker builds skip environment parsing while exporting an unset argument as an empty public value, causing client-side size comparisons to treat the default threshold as zero; the remaining test-import concern is non-blocking.

**Files Needing Attention:** web/Dockerfile, web/src/env.mjs, web/src/components/ui/markdown-render-limits.clienttest.ts
</details>

`````markdown
### Issue 1
web/Dockerfile:101-102
**Unset Docker limit becomes zero**

When a standard Docker build omits this optional build argument, the builder exports an empty `NEXT_PUBLIC_*` value while `DOCKER_BUILD=1` skips the Zod default, so the client-side comparisons treat the markdown limit as zero and disable markdown rendering for ordinary non-empty trace and observation content.

```suggestion
ARG NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT=150000
ENV NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT=$NEXT_PUBLIC_LANGFUSE_MARKDOWN_RENDER_CHARACTER_LIMIT
```

### Issue 2
web/src/components/ui/markdown-render-limits.clienttest.ts:78
**Import remains inside test helper**

The new helper performs an import from inside a function, contrary to the repository convention that imports belong at module scope; this makes module loading imperative and leaves the test inconsistent with the established import structure.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(web): make markdown render characte..."](https://github.com/langfuse/langfuse/commit/875ccfe5142e1a15eefbe3feb08b28f155ddafd2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52009377)</sub>

**Context used:**

**Learned From**


nimarb (approved):

nimarb:
@claude review

PR #15967, opened 2026-08-11
fix(batch-action): parse OTLP JSON-string metadata before dataset mapping

### LINKED ISSUE #15964: bug: dataset batch-add JSONPath mapping fails on nested metadata for OTLP-ingested observations, but resolves in the mapping Preview
### Describe the bug

When adding observations to a dataset via the batch action, a JSONPath that descends into a nested `metadata` value **resolves correctly in the mapping Preview but misses when the background job runs**, so every selected row fails.

This affects observations ingested through the OTLP endpoint. OTel span attributes are scalars, so a sender must JSON-encode any nested metadata value. Langfuse's read API parses those strings back into objects/arrays, but the batch-action worker appears to receive them still as strings — so the Preview and the executor operate on different value types for the same field.

### Steps to reproduce

**1.** Ingest a span whose metadata values are JSON strings:

```bash
curl -X POST "$LANGFUSE_HOST/api/public/otel/v1/traces" \
  -u "$PUBLIC_KEY:$SECRET_KEY" -H 'Content-Type: application/json' \
  -d '{"resourceSpans":[{"scopeSpans":[{"spans":[{
    "traceId":"b1b1b1b1b1b1b1b1b1b1b1b1b1b1b1b1","spanId":"c1c1c1c1c1c1c1c1",
    "name":"example-span","kind":1,
    "startTimeUnixNano":"1786398355000000000","endTimeUnixNano":"1786398356000000000",
    "attributes":[
      {"key":"langfuse.observation.output","value":{"stringValue":"some output text"}},
      {"key":"langfuse.observation.metadata.config","value":{"stringValue":"{\"name\":\"example\"}"}},
      {"key":"langfuse.observation.metadata.events","value":{"stringValue":"[{\"kind\":\"a\",\"text\":\"hello\"},{\"kind\":\"b\"}]"}},
      {"key":"langfuse.observation.metadata.label","value":{"stringValue":"plain-string"}}
    ]}]}]}]}'
```

**2.** Confirm the read API parses the nested values — `GET /api/public/v2/observations?fields=core,metadata&...` returns `config` as an object and `events` as an array.

**3.** Observations table → select the row → **Actions → Add to dataset** → Custom mapping, Target: Key-value map, Source: **Metadata**.

**4.** Compare the Preview against the executed job:

| JSONPath         | Preview                 | Batch job                                  |
| ---------------- | ----------------------- | ------------------------------------------ |
| `$.config.name`  | resolves to `"example"` | **miss**                                   |
| `$.events..text` | resolves to `["hello"]` | **miss**                                   |
| `$.config`       | resolves                | succeeds — returns the raw JSON **string** |
| `$.label`        | resolves                | succeeds                                   |

Error text:

```
Mapping failed - JSONPath "$.config.name" did not match any data in "metadata" (key: "...")
```

**Expected:** a path that resolves in the Preview also resolves when the job runs.

Note the pattern: single-segment access succeeds while any descent fails. That is only possible if the value is a string at execution time — JSONPath returns a string for a one-segment path, and finds nothing when asked to descend into one.

Self-hosted

v4.3.1 (docker compose). Also reproduced on v3.224.4.

Ingested via the OTLP endpoint (`/api/public/otel/v1/traces`), no Langfuse SDK involved.

### Additional information

`applyFieldMappingConfig` in `packages/shared/src/features/batchAction/applyFieldMapping.ts` is shared by both callers — `MappingPreviewPanel.tsx` and `worker/src/features/batchAction/processAddObservationsToDataset.ts` — so the mapping logic itself is common. The asymmetry looks like the input rather than the logic:

```ts
export function evaluateJsonPath(data: unknown, jsonPath: string): unknown {
  const parsed = typeof data === "string" ? parseJsonPrioritised(data) : data;
```

This parses only when the **entire** `data` is a string; individual values are never parsed. The worker types the observation as `metadata: unknown` and passes `obs.metadata` through unchanged, whereas the web read path yields values already parsed.

**Suggested fix:** normalize metadata once at a boundary shared by both callers, rather than parsing inside `evaluateJsonPath`. The deeper issue is that a Preview which can be handed a different value shape than the executor cannot function as a validation surface — which is what makes this fail silently until the job runs.

**Test gap:** `worker/src/__tests__/applyFieldMapping.test.ts` uses metadata with already-parsed values (`{ user_id: "...", tags: [...] }`). A fixture with JSON-string values would fail today.

No

chelsealong:
## What does this PR do?

Fixes #15964: adding observations to a dataset via the batch action, using a
JSONPath that descends into a nested `metadata` value, resolves correctly in
the mapping Preview but misses when the background job runs — for
observations ingested through the OTLP endpoint.

OTel span attributes are scalars, so a sender JSON-encodes any nested
metadata value. The read API (`observations.byId`, used by the mapping
Preview) parses those values back into objects/arrays via
`parseMetadataCHRecordToDomain`. The worker's `getEventsStreamForDataset`
(used by `processAddObservationsToDataset`) streamed the ClickHouse
`Map(String, String)` metadata column through unparsed, so
`evaluateJsonPath` — which only auto-parses when the *entire* input is a
string, not when individual map values are — saw a string where it expected
an object and any JSONPath descending past the first segment missed.

### Fix

Apply the same `parseMetadataCHRecordToDomain` normalization already used by
the read path to the rows yielded by `getEventsStreamForDataset`, so the
Preview and the executor operate on the same value shapes.

`packages/shared/src/features/batchAction/applyFieldMapping.ts` (the shared
mapping/JSONPath logic) is untouched — the asymmetry was in the input shape,
not the mapping logic, matching the issue's root-cause analysis.

## Type of change

## Test plan

Added `worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts`,
which mocks the ClickHouse stream and asserts `getEventsStreamForDataset`
returns metadata with JSON-string values parsed into objects/arrays (mirrors
the issue's repro: `config`/`events` become nested structures, `label` stays
a plain string).

Confirmed the test fails without the fix (reverted the source change,
reran):

```
AssertionError: expected { config: '{"name":"example"}', …(2) } to deeply equal { config: { name: 'example' }, …(2) }
 FAIL  src/__tests__/event-stream-dataset-metadata.unit.test.ts > getEventsStreamForDataset metadata parsing > parses JSON-string metadata values into objects/arrays, like the read API does
```

With the fix restored:

```
✓ src/__tests__/event-stream-dataset-metadata.unit.test.ts (1 test) 7ms
✓ src/__tests__/applyFieldMapping.test.ts (51 tests) 28ms
Test Files  2 passed (2)
     Tests  52 passed (52)
```

`pnpm --filter worker run lint`:
```
Tasks: 8 successful, 8 total
```

`pnpm --filter worker run typecheck`: clean, no errors.

Note: this sandbox has no ClickHouse/Postgres/.env available, so the
existing DB-integration-style stream tests
(`event-stream-score-filters.test.ts`,
`handleBatchActionJob-comment-filters.unit.test.ts`) don't run here either
(same pre-existing failure on unmodified `main`, unrelated to this change) —
verified by stashing this fix and reproducing the identical `CLICKHOUSE_URL`
env error on the base commit.

## AI disclosure

This PR was authored with the assistance of an AI coding agent (Claude).

Fixes #15964

This PR aligns dataset batch-action metadata with the observation read path by parsing JSON-encoded ClickHouse metadata values before field mapping.
- Normalizes streamed observation metadata through the shared metadata converter.
- Adds a focused unit test covering nested objects, arrays, and unchanged plain strings.
</details>

The PR appears safe to merge; the only identified issue is non-blocking import organization in the new test.

The production change reuses the established read-path metadata parser and the regression test covers the intended object, array, and plain-string behavior; only the test module's import placement needs cleanup.

**Files Needing Attention:** worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts
</details>

`````markdown
### Issue 1
worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts:41
**Static import follows mock setup**

The `getEventsStreamForDataset` import is placed after module-level constants and mock declarations rather than with the imports at the top, making dependency organization inconsistent with the repository convention.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(batch-action): parse OTLP JSON-strin..."](https://github.com/langfuse/langfuse/commit/730386a6af85858a4ab3957345f3393b9730b790) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52038083)</sub>

**Context used:**

**Learned From**

### DIFF
--- a/worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts
+++ b/worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 import { parseMetadataCHRecordToDomain } from "../../../packages/shared/src/server/utils/metadata_conversion";
+import { getEventsStreamForDataset } from "../features/database-read-stream/event-stream";
 
 const chainableQueryBuilder = {
   selectFieldSet: vi.fn().mockReturnThis(),
@@ -29,8 +30,6 @@ vi.mock("@langfuse/shared/src/server", () => ({
   queryClickhouseStream: mocks.queryClickhouseStream,
   logger: { info: vi.fn(), error: vi.fn(), warn: vi.fn() },
 }));
-
-import { getEventsStreamForDataset } from "../features/database-read-stream/event-stream";
 
 describe("getEventsStreamForDataset metadata parsing", () => {
   it("parses JSON-string metadata values into objects/arrays, like the read API does", async () => {
--- a/worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts
+++ b/worker/src/__tests__/event-stream-dataset-metadata.unit.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, it, vi } from "vitest";
+import { parseMetadataCHRecordToDomain } from "../../../packages/shared/src/server/utils/metadata_conversion";
+
+const chainableQueryBuilder = {
+  selectFieldSet: vi.fn().mockReturnThis(),
+  selectIO: vi.fn().mockReturnThis(),
+  buildWithParams: vi.fn(() => ({ query: "SELECT 1", params: {} })),
+};
+
+const mocks = vi.hoisted(() => ({
+  queryClickhouseStream: vi.fn(),
+}));
+
+vi.mock("../env", () => ({
+  env: { BATCH_EXPORT_ROW_LIMIT: 1_000_000 },
+}));
+
+vi.mock("@langfuse/shared/src/db", () => ({
+  prisma: {},
+}));
+
+vi.mock("@langfuse/shared/src/server", () => ({
+  buildEventsStreamQuery: vi.fn(() => ({
+    queryBuilder: chainableQueryBuilder,
+  })),
+  buildEventsBlobExportStreamQuery: vi.fn(),
+  getDistinctScoreNames: vi.fn(),
+  parseMetadataCHRecordToDomain,
+  queryClickhouseStream: mocks.queryClickhouseStream,
+  logger: { info: vi.fn(), error: vi.fn(), warn: vi.fn() },
+}));
+
+import { getEventsStreamForDataset } from "../features/database-read-stream/event-stream";
+
+describe("getEventsStreamForDataset metadata parsing", () => {
+  it("parses JSON-string metadata values into objects/arrays, like the read API does", async () => {
+    mocks.queryClickhouseStream.mockReturnValue(
+      (async function* () {
+        yield {
+          id: "obs-1",
+          trace_id: "trace-1",
+          input: null,
+          output: "some output text",
+          metadata: {
+            config: '{"name":"example"}',
+            events: '[{"kind":"a","text":"hello"},{"kind":"b"}]',
+            label: "plain-string",
+          },
+        };
+      })(),
+    );
+
+    const stream = await getEventsStreamForDataset({
+      projectId: "project-1",
+      cutoffCreatedAt: new Date(),
+      filter: [],
+      rowLimit: 100,
+    });
+
+    const rows: Array<{ metadata: unknown }> = [];
+    for await (const row of stream) {
+      rows.push(row as { metadata: unknown });
+    }
+
+    expect(rows).toHaveLength(1);
+    expect(rows[0]!.metadata).toEqual({
+      config: { name: "example" },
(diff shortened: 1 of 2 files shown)


chelsealong:
Fixed the import ordering flagged by Greptile: moved the `getEventsStreamForDataset` import in `event-stream-dataset-metadata.unit.test.ts` up with the other imports instead of after the `vi.mock` calls. Test still passes, lint clean.

PR #15982, opened 2026-08-11
fix(models): rewrite the price editor's form wiring (LFE-14948)

nkabardin:
🟡 **Live preview: [pr-15982.preview.langfuse.com](https://pr-15982.preview.langfuse.com)**

## TL;DR

A customer hit four bugs while manually pricing a model: a usage type they
couldn't type, keystrokes landing in the wrong row, Submit navigating them away,
and lost work. They are not four glitches — they are one data-model problem, and
the same four lines of `TierPriceEditor.tsx` have now been "fixed" twice with
contradictory one-liners ([#11939](https://github.com/langfuse/langfuse/pull/11939)
made the React key the price key → focus loss;
[#12344](https://github.com/langfuse/langfuse/pull/12344) reverted it to the index
→ rows reorder mid-rename). This rewrites the wiring instead: usage types become a
real field array with stable identity, prices become plain form fields with visible
validation, and the three `useEffect`s that wrote back into the form are gone.

Reading the code also turned up **silent price loss across pricing tiers** that
nobody had reported — fixed here, and pinned by a test.

## Why

Prices lived in the form as `Record<string, number>`, so the object key was at once
the row's **identity**, its **render order**, and the **value being edited**. No
React key choice can be correct on that shape:

- Renaming did `delete prices[old]; prices[new] = value`, which moves the entry to
  the **end** of the object. With `key={index}` the focused input becomes a
  different usage type on the first keystroke, and the rest of the characters land
  in other rows.
- A guard added with the first fix returned early when the typed key already
  existed, so typing `text_input_cached` past the existing `text_input` **dropped
  that keystroke** and left the field stuck. Pasting skipped the colliding
  intermediate, which is why paste "worked".
- An effect rebuilt every non-default tier's prices from the default tier's key set
  with `currentPrices[key] ?? 0`, so renaming a usage type **silently zeroed that
  price in every custom tier** and shipped it on save.
- Price rows were raw `<Input>` + `setValue` with no `FormField`/`FormMessage`, so
  zod errors on `pricingTiers.N.prices` blocked submit with **zero feedback** —
  Submit simply did nothing.

## What

1. **Usage types are one model-level field array.** The API requires every tier to
   price the same key set, so the form now models them once (`usageTypes: [{ key,
   name }]`) instead of N times. `key` is opaque form state — identity never
   derives from the edited value, and no cross-tier sync can exist. Prices are
   per-tier, keyed by that row key, and held as strings so `0.0000025` survives
   every keystroke. `Record<string, number>` exists only at the API boundary.
2. **Every field is a real `FormField` with a `FormMessage`.** Duplicate usage
   types, invalid characters, empty names, empty/negative prices and duplicate tier
   names are a `superRefine` in zod that renders inline — never a swallowed
   keystroke, never a dead Submit button.
3. **All three effects deleted.** Priority is derived from tier order at submit
   (which also fixes duplicate tier names after add/remove/add), cross-tier key
   sync no longer exists, and match-pattern prefill happens in the model-name
   field's `onChange` while the pattern is untouched.
4. **Submit is a checkpoint on edit** — toast, dialog stays open, no navigation.
   Escape with unsaved changes asks first (click-outside already never closed).

## Result

The seven acceptance cases pass by construction rather than by patch. Three
component tests pin the cases that regressed twice; run against the old
implementation they fail with the exact reported symptoms — `text_inpu_cached` for
the untypeable key, and for a single rename of the first of three rows:

```
-  "audio": 0.00003,  "text_input_cached": 0.000004,  "text_output": 0.000016
+  "audioe_p_cd": 0.00003,  "xiuch": 0.000004,  "text_outputttntae": 0.000016
```

…with all three custom-tier prices zeroed alongside.

<details>
<summary>Mechanism, decisions and verification</summary>

### Why usage types moved up a level

The ticket suggested `useFieldArray` over `Array<{ usageType, price }>` per tier.
That fixes identity but keeps N copies of one fact, so every add/rename/remove
still has to fan out across tiers — the write loop that caused the silent zeroing,
just moved into event handlers. `validatePricingTiers` already requires every tier
to share the default tier's key set, so the honest shape is one list of usage types
plus a price per (tier, usage type). Renaming then touches exactly one field, and
adding a row is the only operation that writes to other tiers at all.

### Notes

- Non-default tiers now render the usage type as text instead of a disabled input —
  the name belongs to the default tier, and a greyed input implies otherwise.
- Loading takes the **union** of keys across tiers (default tier first), so a model
  whose tiers disagree cannot silently lose a priced key; the missing cells show an
  inline error until filled.
- `PricePreview` lists the rows that currently parse and drops the "Invalid price
  entries" box, which used to flash red while you were mid-typing a key.
- Removing a usage type also drops its price entries in every tier, so a later row
  cannot inherit a deleted row's price.
- Prices are still entered per unit while the read view has a per-1K/per-1M
  selector. Out of scope here; it is why values get pasted.
- **Acceptance case 7 (tab switch / session blip tears down the page) is not this
  PR** — that is `AppLayout` + `useAuthSession` and has its own ticket (LFE-14949).
  The dirty-form guard here limits the damage; it does not fix the teardown.

### Structure

The touched files were reorganised to the [project-structure RFC](https://linear.app/clickhouse/document/langfuse-web-project-code-structure-rfc-ecbc304915d6)
(`pnpm --filter web structure:stats --scope src/features/models`: **rule 5: 2 → 1**,
total 30 → 29; every other rule unchanged, nothing new added):

- `components/pricing-tiers/` was not a kind folder (rule 5). The editors are
  `PricingSection`'s private subcomponents, so they now live in
  `components/PricingSection/{PricingSection.tsx,components/*}`.
- `UpsertModelFormDialog` gained a test file, so it becomes a folder named after
  the component.
- `utils.ts` and the new form-value module were functions outside a kind folder
  (rules 3–5) → `fns/`, one export per file: `buildFormValues`,
  `toPricingTierInputs`, `derivePriorities`, `makeUsageTypeKeys`,
  `matchPatternFor`, `parsePriceInput`, `getMaxDecimals`.
- Schemas that nothing outside `validation.ts` imports are no longer exported
  (rule 20).

Every move went through `pnpm --filter web structure:move`, so importers were
rewritten mechanically and the moves are pure renames (rule 15 — the move commit
is `1 file changed, 0 insertions(+), 0 deletions(-)`, and `git log --follow`
traces through it). Cross-directory relative imports were absolutised in their
own commit first (rule 26), because otherwise a move cannot avoid editing the
moved file. `components/test-match/` is the remaining rule-5 violation in this
feature; untouched here.

### Verification

- `pnpm --filter web run test-client src/features/models/components/UpsertModelFormDialog.clienttest.tsx` → `Tests 3 passed (3)`
- `pnpm --filter web run typecheck` → clean
- Driven in a real browser against a two-tier model, checking the DOM and the
  persisted rows after each step:
  - typing `_cached` onto `input` keystroke by keystroke → focus stays in that
    input, caret at the end, rows keep their order, and the custom tier's price for
    the renamed key stays `0.000001` (it used to become `0`)
  - `0.0000025` typed character by character survives, preview updates live
  - saved rows: `Standard` priority 0 / `Custom Tier 1` priority 1, both keyed by
    the new name
  - Submit on an edit → row updated, dialog stayed open, URL unchanged
  - duplicate usage type → inline `Duplicate usage type "…"` under the row, no
    mutation fired
  - Escape and Cancel while dirty → confirm first; "keep editing" preserves the
    edit, "discard" closes and the read view still shows the saved values

</details>

Closes LFE-14948.


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #15984, opened 2026-08-11
fix(table): refresh tables in place instead of reloading them

nkabardin:
🟡 **Live preview: [pr-15984.preview.langfuse.com](https://pr-15984.preview.langfuse.com)**

## TL;DR

Refreshing a table threw the rows away and showed skeletons. It wasn't the invalidation — a **relative time window was re-resolved into a fresh absolute timestamp on every refresh tick**, which changed the tRPC query key, which meant react-query had no cached data, which made `isPending` true, which replaced every row with a skeleton. Every 30s with auto-refresh on, and on every filter / page / sort change too.

Now the window is **anchored**, so a refresh is invalidation only and reuses the same query key: rows stay, numbers update in place, scroll position and selection survive, and the only signal is a thin progress bar above the header. Skeletons are reserved for a genuine cold load.

Measured on the same local dataset, before → after (75s watch, 30s auto-refresh, scrolled 200px down):

| | traces table | events table (v4 Tracing) |
|---|---|---|
| distinct row-query keys per refresh cycle | 4 keys / 5 fetches → **1 key / 4 fetches** | 7 / 11 → **4 / 8** (only real scope changes) |
| samples showing skeleton rows | 6 → **0** | 8 → **0** |
| shimmering metric / I/O cells (peak) | 200 → **0** | present → **0** |
| scroll position after a refresh | reset to 0 → **held** | reset to 0 → **held** |

## Why

`toAbsoluteTimeRange(timeRange)` resolves "past 24h" to `{from: now-24h, to: now}`. Three tables re-ran it on a `refreshTick` — new `from`/`to` per tick, so the tick moved the window (which is what made new rows appear) *and* re-keyed every query as a side effect. The five `invalidate()` calls were never the problem: invalidation alone keeps the previous data.

The v4 events table had a second, blunter version of the same thing: `<DataTable key={...dataUpdatedAt...}>` remounted the whole table on every successful fetch, which is why scroll position died there even when rows survived.

## What

1. **`useLiveTableDateRange`** — one anchored window per table. A relative range stays **open-ended at the top** (`to: undefined`), so rows that arrive after the anchor still match without re-resolving anything; the anchor is replaced only when the user picks another range or after it drifts past its budget (10% of the window, floor 1 min). `refreshTick` is gone from traces and events; refresh is `invalidate()` only.
2. **`placeholderData: (prev) => prev`** on the row, metrics, count and batched-I/O queries, so the key changes that remain (filter, page, sort, re-anchor) keep the current rows on screen while the new ones load.
3. **Cold load vs fetching** — `data.isLoading` now means "nothing to show"; a fetch over rows that are already there sets the new `isFetching` prop and renders a 2px bar above the header.
4. **Per-row cell state** — metric and I/O cells hold their last value; only a row we have no metrics/I/O for *yet* renders a loading cell, instead of every cell blinking whenever the batch re-keys.
5. **Events remount key removed**, and the tick-decoupled facet window folded into the anchored one (it existed only to keep the facet query key stable across ticks — the window no longer moves tick to tick, so both can share it).
6. **The progress signal reads as one event per refresh.** A refresh is several sequential queries (rows → their batched I/O → counts), and a local one resolves in ~200ms, so a naive `isFetching` binding flickers. `useAnimatedBusy` holds the flag until everything has been quiet for a settle window, rounded up to a whole animation cycle, and counts one *epoch* per refresh so the bar animates once from the left rather than restarting between stages. Same hold drives the refresh button's spinner (which also no longer disables itself — the flash of the disabled state read as a glitch).

<details>
<summary>Mechanism, trade-offs and verification</summary>

**Why anchor instead of quantising.** Any scheme that keeps re-resolving `now` changes the key sooner or later. Making the upper bound open for relative ranges is what breaks the coupling: new rows match a window that never has to move, so ticks cost nothing. The lower bound then only needs to be *honest*, not exact — it re-anchors after drifting 10% of the window (2.4h on "past 1 day", 3 min on "past 30 min"), and that transition is itself covered by `placeholderData`.

**Two deliberate consequences.** A relative window no longer carries `timestamp <= now`, so a row with a slightly future timestamp (clock skew on the client that sent it) is now visible instead of silently filtered out. And the anchored window can lag reality by up to its drift budget; the picker still says "past 1 day" because that is still what it is, ±10%.

**Where the tick survives.** The events chart and the outlier strip need a *closed* window, so theirs still advances on refresh — truncated to the minute, because a to-the-millisecond bound re-keyed the chart query on every tick and that cold load is what faded the strip out mid-refresh. Both of its ends come from one length (end = the later of the anchor and the last refresh, start = measured back from it) rather than pairing the table's anchored `from` with a moving `to`, which would let the window grow between anchors and could inverse it once the selected range is shorter than the time since the last refresh. The table's own window ignores all of this.

**Row identity was already right** (`getRowId` prefers `row.id`), so existing rows reconcile and only new rows mount — no row-rendering changes were needed to "just add lines".

**The progress bar.** A track plus a highlight a third as wide that translates across it, clipped by the track. Several details are load-bearing, each of them a bug found by looking at it:
- The highlight starts and ends fully outside the track (measured: zero visible pixels at both ends of the cycle) so the wrap back to the start is never seen — inside the box it read as a blink every cycle. The clip also keeps it from growing the table's scroll width (`scrollWidth` 2663 before and after).
- Sticky on **both** axes. Its box is only as wide as the visible area, so with a vertical anchor alone it scrolled out of view on any table wider than the viewport (at `scrollLeft` 1200 it sat at −778px instead of at the scroll port's edge).
- The cycle length lives in **one** place: the bar hands `REFETCH_SWEEP_MS` to CSS as `--table-refetch-cycle` and to the busy hook, which only releases on a whole sweep. Two independent literals would drift and cut the sweep short.
- Both edges are soft — the sweep is already running when the bar fades in, and on the way out it fades first and only then pauses — and a bar that mounts idle (every table that passes no `isFetching`) starts paused, so no animation ever runs behind an invisible bar.
- The fade deliberately avoids Tailwind's `duration-*`, which sets `--tw-duration`: `animate-*` reads the same variable, and it compiled the 1.4s sweep down to 0.2s.

**Verification** (local dev, seeded v4 + v3 data, both surfaces, before/after with the same script):
- Query keys captured off the wire across two auto-refresh ticks: identical (1 distinct key on traces, and on events only the keys for a real page/filter change).
- DOM sampled every 120ms for 75s: zero samples with skeleton rows, zero shimmering metric cells, row count constant, scroll offset constant at 200/240px. The pre-fix run of the same script shows skeletons on every tick and scroll snapped back to 0.
- Manual refresh, next page, and a filter toggle all keep the current rows up with the progress bar until the new page lands.
- Traced against live ingestion (one trace + generation every 5-15s through the public API) so refreshes had genuinely arriving rows: three consecutive auto-refreshes produced three bar windows, one animation epoch each, zero cold loads, zero window re-anchors — including the cycle where a new row arrived and the batched-I/O query re-keyed as a second stage.
- I/O columns still populate with the remount key gone.
- Unit coverage for the anchor (stable across ticks, re-anchors on drift, absolute ranges pass through, a fresh object for the same preset does not re-key) and for the busy flag (holds a fast finish for a whole cycle, spans the gap between a refresh's stages, one epoch per period).

**Not in this PR.** The other 7 tables under `components/table/use-cases/` (observations, sessions, scores, …) still hold the old pattern; the pieces they need are now shared. And when rows are added above your reading position the content still shifts by that many rows — the "N new" pill / scroll-anchoring decision is deliberately left open.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #16018, opened 2026-08-12
fix(in-app-agent): invoke sandbox lifecycle hooks

### LINKED ISSUE #15953: [bug] Sandbox provider lifecycle hooks are never invoked, leaking runtime sessions
## Summary
The sandbox provider lifecycle hooks are defined but never invoked. `createInAppAgentSandbox` returns `onTurnEnded`, but that callback only persists `providerSessionId`; it never calls the provider's `suspendSession` hook. No production caller for `suspendSession` or `terminateSession` was found.

## Code reference
- `packages/shared/src/in-app-agent/server/sandbox/service.ts:69-77` — `onTurnEnded` only calls `persistState()`.
- `packages/shared/src/in-app-agent/server/sandbox/types.ts:58-66` — documents `suspendSession` as releasing live runtime resources and `terminateSession` as permanent teardown.
- `packages/shared/src/in-app-agent/server/sandbox/providers/docker.ts:211-228` and `lambdaMicrovm.ts:284-332` — implement those hooks.

## Reproduction
1. Enable a sandbox provider and execute a sandbox tool in a conversation.
2. End the agent turn.
3. Inspect the provider: Docker containers remain running, and the Lambda provider is never asked to suspend the microVM.

## Expected behavior
After a turn, the provider lifecycle should suspend the session while retaining its persisted identifier for reuse; permanent cleanup should call `terminateSession` when the conversation/session is deleted.

## Actual behavior
Only the database state is persisted. The provider hooks are unreachable from production code, so live runtime resources are not released.

## Why it matters
A Docker container can remain running per conversation until manually removed, causing resource exhaustion. The Lambda provider's in-memory session map and remote microVM lifecycle also never receive the intended cleanup signal. The provider contract and the in-app-agent README therefore describe behavior that does not occur.

## Suggested fix
Wire `onTurnEnded` to an idempotent `suspendSession` call and add the corresponding termination path, with error handling and tests asserting both provider hooks are invoked.

uuzzrm:
## What does this PR do?

Fixes #15953

Sandbox provider lifecycle hooks were defined but never reached from production code. This change:

- suspends provider sessions once when a turn finishes, while retaining the persisted session ID;
- terminates sessions from conversation deletion, outside the DB transaction;
- preserves the session ID when termination fails so a retry can clean it up;
- prevents a late worker cleanup from restoring a session ID after deletion.

## Type of change

## Mandatory Tasks

- [x] Self-reviewed the diff.
- [x] Added regression coverage for suspend/terminate and retry behavior.
- [x] Checked formatting, lint, typecheck, and builds.

## Verification

- `11 tests passed` — `in-app-agent-sandbox.servertest.ts`
- `15 tests passed` — `in-app-agent-stream.servertest.ts`
- `pnpm --filter worker run build`
- touched-file ESLint and Prettier checks passed
- DB-backed router/worker integration tests could not run in this checkout because PostgreSQL/Docker are unavailable (`localhost:5432` was unreachable).

AI assistance was used during investigation and implementation; the patch and tests were reviewed against the repository's contribution guidance.

This PR wires sandbox suspension into worker turn completion and termination into conversation deletion while retaining failed-cleanup IDs for retries.

- Adds shared sandbox termination and single-flight suspension behavior.
- Moves deletion cleanup outside the database transaction and guards worker state writes after soft deletion.
- Adds regression coverage for suspension, termination, and failed termination.
</details>

The PR should not merge until a provider session created concurrently with deletion is terminated instead of having its persistence silently discarded.

A running worker can create a sandbox before observing cooperative cancellation, after which the deleted-row update guard loses the new session ID and only suspends the resource; provider configuration transitions can also discard persisted cleanup handles without successful termination.

**Files Needing Attention:** worker/src/features/in-app-agent/executeInAppAgentRun.ts, web/src/features/in-app-agent/server/backgroundRunService.ts
</details>

```mermaid
sequenceDiagram
  participant U as Delete request
  participant DB as Postgres
  participant W as Running worker
  participant P as Sandbox provider
  U->>DB: Read providerSessionId
  U->>DB: Set deletedAt and request cancellation
  W->>P: ensureSession()
  P-->>W: New session ID
  W->>DB: updateMany where deletedAt is null
  DB-->>W: 0 rows updated
  U->>P: Terminate snapshotted session ID
  W->>P: Suspend new session
  Note over W,P: New session ID is no longer persisted
```
</details>

`````markdown
### Issue 1
worker/src/features/in-app-agent/executeInAppAgentRun.ts:327-335
**Late session cleanup is lost**

If a running worker creates a sandbox after deletion snapshots the previous session ID but before it observes cooperative cancellation, this `updateMany` silently discards the new session ID because `deletedAt` is already set. Deletion can only terminate the earlier snapshot and worker completion only suspends the new session, leaving the provider resource allocated until provider-side expiry.

### Issue 2
web/src/features/in-app-agent/server/backgroundRunService.ts:319-334
**Unresolved provider loses cleanup handle**

If `LANGFUSE_IN_APP_AGENT_SANDBOX_PROVIDER` is unset or changed after a session was persisted, cleanup is skipped or sent to the wrong provider, but this code still clears `providerSessionId`. That orphans the MicroVM or container and removes the only persisted identifier available for a later cleanup attempt.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(in-app-agent): invoke sandbox lifecy..."](https://github.com/langfuse/langfuse/commit/8caa18b23d1aef693cc366bf5df0ccc98b45a4d0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52468200)</sub>

</details>

### DIFF
--- a/packages/shared/src/in-app-agent/server/sandbox/index.ts
+++ b/packages/shared/src/in-app-agent/server/sandbox/index.ts
@@ -1,6 +1,9 @@
 export { createLambdaMicrovmSandboxProvider } from "./providers/lambdaMicrovm";
 export { getDefaultInAppAgentSandboxProviderType } from "./config";
-export { createInAppAgentSandbox } from "./service";
+export {
+  createInAppAgentSandbox,
+  terminateInAppAgentSandboxSession,
+} from "./service";
 export type {
   InAppAgentSandbox,
   InAppAgentSandboxProviderType,
--- a/packages/shared/src/in-app-agent/server/sandbox/service.ts
+++ b/packages/shared/src/in-app-agent/server/sandbox/service.ts
@@ -1,5 +1,18 @@
 import type { InAppAgentSandbox, SandboxFile, SandboxProvider } from "./types";
 
+export async function terminateInAppAgentSandboxSession(params: {
+  provider: Pick<SandboxProvider, "terminateSession">;
+  providerSessionId?: string | null;
+}): Promise<void> {
+  if (!params.providerSessionId) {
+    return;
+  }
+
+  await params.provider.terminateSession?.({
+    sessionId: params.providerSessionId,
+  });
+}
+
 export async function createInAppAgentSandbox(params: {
   conversationId: string;
   projectId: string;
@@ -13,6 +26,8 @@ export async function createInAppAgentSandbox(params: {
 }> {
   let sessionId = params.providerSessionId ?? null;
   let sessionIsKnownActive = sessionId !== null;
+  let sessionIsSuspended = false;
+  let sessionSuspendPromise: Promise<void> | undefined;
 
   const persistState = async () => {
     await params.saveState({
@@ -23,12 +38,14 @@ export async function createInAppAgentSandbox(params: {
   const updateSessionState = async (nextSessionId: string) => {
     if (nextSessionId === sessionId) {
       sessionIsKnownActive = true;
+      sessionIsSuspended = false;
       return;
     }
 
     sessionId = nextSessionId;
     await persistState();
     sessionIsKnownActive = true;
+    sessionIsSuspended = false;
   };
 
   const ensureSession = async () => {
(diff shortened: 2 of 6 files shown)


uuzzrm:
Both findings are valid and addressed in 5e58e6b.

- P1: saveState now detects the deleted-conversation race (updateMany matches 0 rows because deletedAt is set) and terminates the newly created sandbox session immediately, so the provider resource is released instead of being left until provider-side expiry.
- P2: deleteBackgroundConversation only clears providerSessionId after a successful termination. When the provider cannot be resolved (env unset or changed), the persisted id is kept so a later cleanup attempt still has the handle.

bezbac:
Thanks for raising this and for putting the work into the PR! Our sandbox support is still quite fresh and we're still making some tweaks to it. In our case, Lambda MicroVMs already auto-suspend and have a bounded lifetime, so we don’t think the extra lifecycle handling that was added is needed here. We might re-evaluate this in the future. I’m going to close this, but really appreciate you flagging the issue 🙏

PR #16030, opened 2026-08-12
fix(trace): resolve the peeked observation off the tree (LFE-14992)

nkabardin:
🟡 **Live preview: [pr-16030.preview.langfuse.com](https://pr-16030.preview.langfuse.com)**

## TL;DR

Opening an observation that sits past the trace-detail's 10 000-observation load cap showed the
**trace** in the detail panel instead of the observation you clicked — silently, with no error. The
panel now resolves the selected observation from its **id**, not from the tree, so it shows the right
data (plus real loading / not-found states) whether or not the tree could draw it.

## Why

`MAX_OBSERVATIONS_PER_TRACE = 10_000` and rows arrive `startTime ASC`, so a big trace loses its
**chronological tail**. The panel then did this:

```tsx
const selectedNode = selectedNodeId ? nodeMap.get(selectedNodeId) : null;   // undefined past the cap
if (isObservationSelected && selectedNode) { … }                            // guard fails
return <TraceDetailView … />;                                              // ← we land here
```

`nodeMap` is built from the capped list, so past the cap the lookup misses and the panel falls
through to the trace. The existing "Observation not found" message was dead code on this path. Tree
drawability was deciding what the data panel shows — for a customer that means clicking a real
observation and being shown something else entirely.

The observation itself was never unavailable: `api.observations.byId` already exists, is used in this
same feature, and is v4-safe through the `getObservationById` routing wrapper.

## What

- New `useSelectedObservation` hook: derives "is an observation selected" from the selected id
  (the synthetic TRACE node is recognised by its id shape, no map lookup), resolves the row from the
  loaded list when present — **no extra request on the normal path** — and fetches by id otherwise.
- `TracePanelDetail` renders one of five honest outcomes: trace, observation, loading, not-found,
  error. A deleted or foreign id says so instead of falling back to the trace, and a transient
  failure is not reported as a missing observation.
- `ObservationDetailView`'s tree reach-ins (`nodeMap.get`, `roots.some`) keep degrading to
  absent/false, now deliberately: subtree metrics and root-only chrome are tree facts we genuinely
  don't have for an off-tree observation, so they are omitted rather than guessed.
- The tree and the cap are untouched — making the tree honest about the missing tail, and revisiting
  the cap itself, are the two follow-up PRs in this batch.

Seeder: `trace-tree --stride-ms N` starts each observation `index × N` ms in. Reproducing a cap
boundary needs unique, strictly increasing start times — with the default nested timing 12 000
observations share ~500 milliseconds, so which rows fall past the cap is an arbitrary tie-break.

<details>
<summary>Mechanism, verification and fixture</summary>

**One fetch, shaped for reuse.** The next PR in this batch injects the same row into the tree, so the
by-id fetch lives in one hook both surfaces can call rather than two near-identical copies. It runs
with `verbosity: "compact"` (the detail view fetches full I/O separately via `useParsedObservation`)
and `meta: { silentHttpCodes: [404] }` — the panel owns the missing-observation UX, so the global
error toast would just be alarming noise on top of it. The 404-no-retry behaviour comes from the
client's global query default.

**Cap is v4-only.** `events.byTraceId` caps the list; the v3 path
(`traces.byIdWithObservationsAndScores`) has no row cap. Both read paths were checked anyway, since
the resolution logic is shared.

**Fixture** (12 000 observations, exact boundary at index 10 000):

```bash
pnpm run seed -- trace-tree --observations 12000 --depth 14 --breadth 400 --v4 \
  --stride-ms 10 --id-prefix obscap12k
```

`…-obs-0` … `…-obs-9999` load; `…-obs-10000` … `…-obs-11999` fall past the cap and are absent from
the tree. Note that re-running with *changed* timing flags needs a fresh `--id-prefix`: `start_time`
is an events ORDER BY key, so both versions persist in the same trace.

**Verified locally** (v4 preview on, then off):

| Check | Result |
| --- | --- |
| Past-cap observation, full trace page | own detail renders — timestamp, latency, metadata, and full input/output |
| Past-cap observation, peek from the observations table | same, both surfaces share `TraceDetailBody` |
| In-cap observation | unchanged, and **zero** `observations.byId` requests |
| Bogus observation id | "Observation not found" in the panel, no trace fallback, no error toast |
| v3 read path (preview off) | in-list observation unchanged; bogus id also says not-found |

**Tests.** `TracePanelDetail.clienttest.tsx` pins the regression (the two bug cases fail against the
previous implementation with a faithful `nodeMap` mock; the trace-selection case passes on both).
`useSelectedObservation.clienttest.tsx` covers the degrade paths and the no-extra-request fast path.

```
Tests  479 passed | 48 skipped (527)   (pnpm --filter web run test-client src/features/traces)
Tasks: 7 successful, 7 total           (pnpm run lint)
```

</details>

The PR resolves selected observations independently of the capped trace tree and adds explicit loading, missing, and error states. It also adds regression coverage and a deterministic seeder timing option for exercising the 10,000-observation boundary.

- Adds `useSelectedObservation` with a loaded-list fast path and by-ID fallback.
- Updates the trace detail panel and off-tree observation behavior.
- Adds client tests for capped, missing, normal, and trace selections.
- Adds `--stride-ms` to produce deterministic chronological cap fixtures.
</details>

The valid observation-ID collision should be fixed before merging because it can still make a selected observation render as the trace.

Synthetic trace selection is inferred from an unreserved string value, so a currently valid observation ID can be misclassified and its detail view becomes inaccessible.

**Files Needing Attention:** web/src/features/traces/hooks/useSelectedObservation.ts; web/src/features/traces/fns/treeBuilding.ts
</details>

```mermaid
flowchart TD
  S[Selected node ID] --> T{Synthetic trace ID?}
  T -->|Yes| TV[Render trace detail]
  T -->|No| L{Present in loaded observations?}
  L -->|Yes| OV[Render observation detail]
  L -->|No| Q[Fetch observation by ID]
  Q -->|Success| OV
  Q -->|Pending| LD[Render loading state]
  Q -->|Not found| NF[Render not-found state]
  Q -->|Other error| ER[Render error state]
```
</details>

`````markdown
### Issue 1
web/src/features/traces/hooks/useSelectedObservation.ts:42-45
**Synthetic trace ID collision**

When an observation has the valid ID `trace-${traceId}`, this comparison classifies it as the synthetic trace node, causing the panel to render the trace detail instead of the selected observation.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): resolve the peeked observati..."](https://github.com/langfuse/langfuse/commit/0ae8dd86d52096ab615becb47d234a96e6edce27) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52553139)</sub>


nkabardin:
@claude review

PR #16032, opened 2026-08-12
fix(admin-webhook): retry after failed delivery and bound the request timeout

### LINKED ISSUE #15957: [bug] Admin access webhook failures suppress retries and can hang requests
## Summary
Admin access webhook delivery is awaited on request paths without a timeout, and failed deliveries are marked as recently sent before the request succeeds. A failed webhook therefore suppresses retries for 24 hours.

## Code reference
- `web/src/server/adminAccessWebhook.ts:19-29` — records the dedupe timestamp before delivery.
- `web/src/server/adminAccessWebhook.ts:54-81` — awaits `fetch` without an abort timeout and only logs failures.
- Callers await this helper in `web/src/server/api/trpc.ts:329,357,453,620,706`, `web/src/pages/api/dashboard/execute-query-stream.ts:91,101`, and `web/src/features/traces/server/buildTraceExport.ts:156`.
- Existing tests cover success, duplicate, and non-OK responses but not retry-after-failure or a hung endpoint.

## Reproduction
1. Configure `LANGFUSE_ADMIN_ACCESS_WEBHOOK` to an endpoint that returns HTTP 500 or never responds.
2. Trigger an admin access event.
3. Trigger the same event again within 24 hours.

## Expected behavior
A failed delivery should be retryable, and an unresponsive webhook should not block the user-facing request indefinitely.

## Actual behavior
The first attempt stores the dedupe key before `fetch`; failures do not clear it. The second attempt is skipped for 24 hours. A hanging fetch remains awaited with no application-level deadline.

## Why it matters
Security/admin notifications can be silently lost, while project access, dashboard, and export operations can stall behind an external endpoint.

## Suggested fix
Only commit the dedupe timestamp after a successful delivery (or clear it on failure), add a bounded abort timeout, and test both failure retry and hung-endpoint behavior.

WhoamiI00:
Fixes langfuse/langfuse#15957

## Problem

`sendAdminAccessWebhook` recorded the 24h dedupe timestamp **before** attempting delivery:

```ts
if (shouldSkipDueToRecentDuplicate(payload)) return; // <- writes the timestamp
await fetch(env.LANGFUSE_ADMIN_ACCESS_WEBHOOK, { ... });
```

So any failed delivery — a 500, a network error — still consumed the dedupe window, and every subsequent admin access event for the same `email:project:org` was suppressed for a full day. The notification was silently lost.

The `fetch` also had no deadline. Callers `await` this helper on user-facing request paths (5 call sites in `server/api/trpc.ts`, `pages/api/dashboard/execute-query-stream.ts`, `features/traces/server/buildTraceExport.ts`), so an unresponsive endpoint stalled those requests for as long as the connection took to die.

## Fix

* **Claim the dedupe slot up front, release it on failure.** Keeping the up-front claim preserves the existing behaviour where concurrent callers collapse onto a single delivery; releasing it on failure means a transient outage no longer suppresses the next attempt. Deleting the key is sufficient to restore the prior state — reaching a delivery attempt implies there was either no entry or one already outside the window, and both are equivalent to absent.
* **Bound each attempt at 5s** with an `AbortController`, so a hung endpoint cannot block a user-facing request.

I used `AbortController` + `setTimeout` rather than `AbortSignal.timeout()` so the deadline runs on the ambient `setTimeout` and stays controllable from fake timers — otherwise the hung-endpoint behaviour the issue asks for isn't testable.

## Tests

5 new cases in `web/src/__tests__/server/async/admin-access-webhook.servertest.ts`:

* retries on the next event after a non-ok response
* retries on the next event after `fetch` rejects
* still dedupes a **successful** delivery within the window (guards against over-correcting)
* aborts a hung endpoint instead of awaiting it indefinitely
* retries after a hung endpoint times out

All 5 were confirmed to fail against the current implementation before the fix — the two hung-endpoint cases fail by hanging until Vitest's 30s timeout, which is the bug. The existing payload assertion gained `signal: expect.any(AbortSignal)`.

## Verification

```
Tests  21 passed (21)
```

(`admin-access-webhook.servertest.ts` + `admin-access-webhook-middleware.servertest.ts`)

`pnpm --filter web run typecheck` exited 0, and `eslint` / `prettier --check` are clean on both changed files.

## Scope

No behaviour change when the webhook is unset or delivery succeeds. No new env vars, no contract changes.

The PR fixes admin-access webhook delivery by releasing provisional dedupe reservations after unsuccessful attempts and bounding outbound requests to five seconds.

* Refactors deduplication into explicit claim and release operations while retaining concurrent-delivery collapse.
* Adds an abortable delivery timeout and clears the timer after completion.
* Adds coverage for HTTP failures, rejected requests, successful deduplication, hung endpoints, and retries after timeouts.

</details>

The PR appears safe to merge, with the failure retry and bounded timeout behavior implemented consistently for the supported server runtime.

Failed deliveries now release their provisional dedupe reservation, successful deliveries retain it, and every awaited outbound request is bounded and cleaned up without introducing a concrete blocking failure.

</details>

Reviews (1): Last reviewed commit: ["fix(admin-webhook): retry after failed d..."](<https://github.com/langfuse/langfuse/commit/aa7ba76d0837a261c320ee3a14beb4d8b2c7ca4e>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=52562316>)

**Context used:**

* Knowledge Base — [Web App (`web/`)](<https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/web-app.md>)

### DIFF
--- a/web/src/__tests__/server/async/admin-access-webhook.servertest.ts
+++ b/web/src/__tests__/server/async/admin-access-webhook.servertest.ts
@@ -78,6 +78,7 @@ describe("sendAdminAccessWebhook", () => {
       headers: {
         "Content-Type": "application/json",
       },
+      signal: expect.any(AbortSignal),
     });
   });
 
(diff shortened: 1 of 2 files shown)


maxdeichmann:
This is langfuse internal infra. If things go wrong here, this is not bad.

PR #16033, opened 2026-08-12
feat(trace): tell the truth about observations past the load cap (LFE-14993)

nkabardin:
🟡 **Live preview: [pr-16033.preview.langfuse.com](https://pr-16033.preview.langfuse.com)**

## TL;DR

A trace with more observations than the detail view loads used to lose its chronological tail in
silence — the tree drew an incomplete picture, the full page raised an **error-styled toast** about
it, and the peek said **nothing at all**. Now the navigation pane carries a dismissible notice
stating the cap that was actually applied, and the observation you opened is merged into the tree
instead of being invisible in it.

[#16030](https://github.com/langfuse/langfuse/pull/16030) (LFE-14992), which owns the by-id fetch
this reuses, has since merged — `main` is merged in, so this now stands on its own.

## Why

The list is capped and ordered `startTime ASC`, so what disappears is the **end** of the trace. Two
things were wrong about how we said so:

- The only signal was a toast that reads like a failure, on the full page only. A truncated trace is
  a fact about the picture you're looking at, not an error that just happened.
- The peek had no signal whatsoever — the worse half, since that's where an observation past the cap
  is usually opened from.

And the observation itself was simply absent from the tree while its data filled the detail panel:
you could read a span that the tree next to it claimed did not exist.

## What

- **Notice in the navigation pane** (tree · timeline · search), in the same slot as the existing
  hidden-observations notice, so it shows on the full page **and** the peek. Dismissible with an `×`
  — it frees the space for the visit rather than nagging, and comes back on the next trace or reload.
  It phrases no total: the server stops counting at the cap, so we know "more than this", never how
  many. **No extra `count()` query.**
- **The selected observation joins the tree**, merged in before tree building. It carries no per-row
  label — that is noise on every scroll — so the notice states the one thing a marker would have
  protected: see below.
- The toast is gone.

The number travels **from the server response** (`events.byTraceId` now reports the cap it applied)
instead of a client-side copy of `MAX_OBSERVATIONS_PER_TRACE`. That constant lives in the server-only
barrel, and the follow-up ticket may replace the row count with a byte budget; sourcing the number
from the response keeps the copy true through that change instead of relying on someone remembering
to edit a string.

<details>
<summary>Mechanism, what the copy may and may not claim, and verification</summary>

**Where the merge happens.** `SelectionProvider` moves above `TraceDataProvider` so the selected
observation can be resolved *before* the tree is built; `Trace` then appends the fetched row to the
observation list. Everything downstream — `nodeMap`, `searchItems`, the temporal frame, the timeline
and graph — picks it up for free, and the detail panel resolves it from the list (so #16030's by-id
query goes quiet on this path).

**The copy claims a wrong position only when there is one.** `treeBuilding` nulls a
`parentObservationId` it cannot resolve, so the injected row lands at root level *only when its parent
also fell past the cap*. In practice a past-cap observation often has a loaded parent and nests
correctly — so the notice's last sentence varies:

| Case | Placement | Notice ends with |
| --- | --- | --- |
| Parent loaded (or no parent) | nests correctly | "The one you opened is loaded separately." |
| Parent also past the cap | root level | "…is loaded separately, and appears at the top level because its parent is missing too." |

Saying "may be out of position" unconditionally would be its own small lie, so the placement fact is
derived once (is the parent in the loaded list?) and carried in context.

**One regression caught in the browser and fixed:** placing the notice inside `TracePanelNavigation`
wrapped the tree in a `flex-1` div under a non-flex parent, so the virtualizer's scroll element grew
to content height and **mounted all 10 001 rows**. Using the layout's existing notice slot keeps 50
mounted rows and keeps the scroll-to-selected behaviour working.

**Verified locally** on a 12 000-observation trace (`--stride-ms` fixture from #16030):

| Check | Result |
| --- | --- |
| Past-cap observation, full page | notice + the row in the tree, scrolled into view and selected, detail panel shows its data |
| Timeline view | same row in the gutter |
| Peek from the observations table | notice (previously: nothing at all) |
| Past-cap child of a past-cap parent | root placement, and the notice says the position is not real |
| `×` dismiss | notice gone for the visit, back on reload |
| Trace under the cap | no notice, and no `observations.byId` request |

```
Tests  592 passed | 48 skipped (640)   (pnpm --filter web run test-client src/features/traces src/features/events)
Tasks: 7 successful, 7 total           (pnpm run lint)
```

</details>


nkabardin:
@claude review

PR #16037, opened 2026-08-12
feat(in-app-agent): trace complete agent turns across approvals

LysanderKie:
🟡 **Live preview: [pr-16037.preview.langfuse.com](https://pr-16037.preview.langfuse.com)**

## Summary

- Emit one Langfuse generation for every in-app-agent model invocation so tool loops, per-call context growth, token usage, and cost are visible.
- Name every generation `invoke-model` so dashboards, evaluators, and saved filters remain stable when the underlying model changes; retain the model ID on the generation's `model` attribute.
- Keep a complete user turn—including durable human-approval continuations—in one `agent-turn` trace and root observation.
- Represent approval latency explicitly with `waiting for user approval for tool <tool_name>` spans between the requesting generation and resumed tool execution.
- Capture Bedrock V3 finish usage before approval suspension so every completed model invocation retains exact token usage and cost inputs even when Mastra does not emit `onStepFinish`.
- Measure tool spans around the actual `execute()` promise, including rate-limit waits and failures, instead of relying on delayed AG-UI tool events.
- Keep complete accumulated messages, context, skills, and tools on each generation input while limiting each generation output to that model invocation.
- Generate each conversation title once by keeping title inference in the web submission path and removing the duplicate worker-completion invocation.

## Approval continuation tracing

- The first run and all approval continuations share the trace ID and root observation ID derived from the original user-message run.
- Each continuation appends a wait span, the tool decision/result, and the following generation to that same trace.
- Trace/root metadata records `approval_continuation_count`; wait spans retain the approval tool, decision, immediate `parent_run_id`, and one-based continuation number. No `parent_trace_id` is emitted because every continuation already belongs to the same trace.
- Rejected tools remain visible with `toolCallApproval: rejected`, but use a zero-duration observation because no execution occurred.
- Continuation lineage is derived by the server, persisted in the internal run request, and forwarded by the worker. The browser submits only the approval decision.
- No database migration or public API change is required.

## Failure isolation and performance

- Instrumentation is attached at model-stream, AG-UI event, and tool-execution boundaries.
- Every tracing callback is fail-open: failures are logged, later tracing is still attempted, and tracing cannot block stream forwarding, tool execution, successful results, or original tool errors.
- Trace flushing remains asynchronous and introduces no awaited network round trip in the agent loop.
- A provider-finished model invocation remains successful even if the following tool or root run fails.
- Provider finish is the single fallback when `onStepFinish` is delayed or absent; callbacks arriving after instrumentation ends are ignored instead of reopening a generation with a late update.

## Why the diff is large

- Most changed lines implement the per-invocation tracing state machine and its focused behavioral coverage.
- The remaining changes integrate model/tool boundaries, carry approval lineage across the durable web/worker boundary, and remove duplicate title inference.

## Impacted packages

- `packages/shared`: model/tool adapter instrumentation, usage normalization, trace construction, failure isolation, and internal approval-run lineage
- `web`: instrumentation, streaming, and approval-chain regression coverage
- `worker`: durable approval resume propagation, removal of duplicate worker-side title inference, and coverage

## Verification

- [x] `pnpm --filter web run test in-app-agent-instrumentation` — 33 passed
- [x] `pnpm --filter web run test in-app-agent-stream` — 18 passed
- [x] `pnpm --filter web run test in-app-agent-background-run` — 25 passed
- [x] `pnpm --filter worker run test executeInAppAgentRun` — 19 passed
- [x] `pnpm run lint` — 7/7 tasks successful
- [x] `pnpm run typecheck` — 7/7 tasks successful
- [x] Preview `da39f06dd`: normal multi-tool, approved chain, rejected approval, tool failure, per-call usage/cost, actual tool timing, and single title generation


packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -616,38 +712,266 @@ export class InAppAgentInstrumentation {
...
+    step: OpenLlmStep,
+    event: unknown,
+    options?: {
+      level?: "ERROR";
+      statusMessage?: string;
+    },
+  ) {
+    const eventRecord = isRecord(event) ? event : undefined;
+    const usageDetails = getStepUsageDetails(eventRecord?.usage);
+    const finishReason =
```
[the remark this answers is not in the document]
LysanderKie: Thanks. I’m leaving this inline intentionally. There are only two short call sites with the same precedence rules, and extracting a helper here would add another private abstraction without removing meaningful branching or improving the call sites. The provider/Mastra precedence is covered by the focused instrumentation tests.

packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -616,38 +712,266 @@ export class InAppAgentInstrumentation {
...
+        ? getStringValue(eventRecord.stepResult.reason)
+        : undefined);
+    const model =
+      getModelIdFromStepEvent(eventRecord) ?? this.model ?? undefined;
+    const output = this.getLlmStepOutput(pending.step, eventRecord);
+
+    this.enqueueObservation("generation-update", {
+      id: pending.id,
+      traceId: this.traceId,
+      parentObservationId: this.rootObservationId,
```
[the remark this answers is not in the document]
LysanderKie: Fixed in `bd879d806`. A late step finish that supplies real usage/output now updates the generation to `level: DEFAULT` and clears `statusMessage`, so a provisional abort/error label does not remain on a completed model call. Added a focused regression test for the aborted-then-late-finish sequence.

packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -616,38 +712,266 @@ export class InAppAgentInstrumentation {
...
       this.createToolObservation(toolCallId, tool, {
         metadata,
         statusMessage,
       });
       this.toolSpans.delete(toolCallId);
     }
+
+    if (forceClose) {
+      for (const [toolCallId, tool] of this.deferredToolCompletions.entries()) {
+        this.createToolObservation(toolCallId, tool, {
```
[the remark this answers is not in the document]
LysanderKie: Good catch. Removed the unreachable `forceClose` loop in `bd879d806`; `endOpenToolSpans` now unconditionally calls `flushDeferredToolCompletions()`, matching the lifecycle invariant established by `endOpenLlmStep`. The focused instrumentation suite passes with 35 tests.

packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -106,6 +116,30 @@ type AgentRunToolSpan = {
...
+  startTime: Date;
+  inputMessageCount: number;
+  completionStartTime?: Date;
+};
+type ToolExecutionTimes = {
+  startTime?: Date;
+  endTime?: Date;
+};
+// Last emitted LLM generation, so a late Mastra onStepFinish can still attach
+// usage after AG-UI complete() has already called end().
```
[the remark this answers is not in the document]
LysanderKie: This was accurate for the reviewed revision but is superseded by the execution-boundary instrumentation in `84a1521aa`. `recordToolExecutionStart` now writes `ToolExecutionTimes.startTime` immediately before the real tool `execute()` promise, while `recordToolExecutionEnd` writes the end in `finally`. The field is therefore active and intentionally retained; preview verification showed non-zero durations for both normal and approval-gated tools.

LysanderKie:
@claude review

packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -616,38 +776,284 @@ export class InAppAgentInstrumentation {
...
+              statusMessage ??
+              (typeof metadata?.error === "string"
+                ? metadata.error
+                : "LLM step closed before step finish"),
+          }
+        : {}),
+    });
+    this.openLlmStep = undefined;
+    this.flushDeferredToolCompletions();
+  }
```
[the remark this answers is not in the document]
LysanderKie: Fixed in `da39f06dd`. `endOpenLlmStep` now preserves a completed generation when the provider finish event already supplied usage, model, and output; the downstream failure remains on the root/tool observation. Added a focused regression covering provider finish followed by a run error (`36 passed`).

packages/shared/src/in-app-agent/server/agent.ts
```diff
@@ -741,6 +769,99 @@ type ExecutableInAppAgentTool = {
...
+  }
+
+  const agent = "agent" in context ? context.agent : undefined;
+  if (typeof agent !== "object" || agent === null) {
+    return undefined;
+  }
+
+  const toolCallId = "toolCallId" in agent ? agent.toolCallId : undefined;
+  return typeof toolCallId === "string" ? toolCallId : undefined;
+}
```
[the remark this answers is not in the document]
LysanderKie: Fixed in `da39f06dd`. `getToolCallId` is now exported from `tools.ts` and reused by the execution-timing wrapper, removing the duplicate context-shape parser. The stream suite passes (`19 passed`).

LysanderKie:
@claude review

packages/shared/src/in-app-agent/server/instrumentation.ts
```diff
@@ -748,18 +1154,123 @@ export class InAppAgentInstrumentation {
...
+      ? { approvalRequestedAt: new Date(approvalRequestedAt) }
+      : {}),
+    ...(approvalDecidedAt
+      ? { approvalDecidedAt: new Date(approvalDecidedAt) }
+      : {}),
+    metadata: {
+      continuation_type: "tool_approval",
+      continuation_number: continuationNumber,
+      parent_run_id: approvalRequest.runId,
+      parent_trace_id: traceId,
```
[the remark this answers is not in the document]
LysanderKie: Agreed. Removed `parent_trace_id` in `f31293865`; because every continuation belongs to the root trace, the field was self-referential and added no lineage information. Each wait span still records its one-based continuation number, approval details, and immediate `parent_run_id`, while the shared trace/root retains `approval_continuation_count`. The existing approved/rejected continuation test now asserts the exact wait-span metadata shape, and the focused suite passes (`36 passed`).

packages/shared/src/features/inAppAgent/types.ts
```diff
@@ -79,8 +79,16 @@ export const InAppAgentRunRequestSchema = z.discriminatedUnion("kind", [
   }),
   z.object({
     kind: z.literal("approvalDecision"),
-    /** Provenance for debugging only; no code path queries lineage. */
+    /** Immediate parent run used to derive durable approval-chain lineage. */
```
LysanderKie, line 82: Who is required to pass that information through? As all of that is required only for tracing, so I am afraid we are mixing concerns here.

packages/shared/src/features/inAppAgent/types.ts
```diff
@@ -79,8 +79,16 @@ export const InAppAgentRunRequestSchema = z.discriminatedUnion("kind", [
...
     parentRunId: z.string(),
+    /** Original user-message run whose trace owns the complete approval chain. */
+    rootRunId: z.string().optional(),
+    /** Stable root trace timestamp propagated across durable continuations. */
+    traceStartedAt: z.iso.datetime({ offset: true }).optional(),
```
LysanderKie, line 87: Why do we need `traceStartedAt`?

packages/shared/src/in-app-agent/server/runLifecycle.ts
```diff
@@ -271,6 +277,20 @@ export async function decideToolApproval(params: {
     const parentRequest = InAppAgentRunRequestSchema.safeParse(
       parentRun.request,
     );
+    const continuationNumber =
```
LysanderKie, line 281: As this is only needed for tracing, can we add a short comment explaining that (max 1 line)

packages/shared/src/in-app-agent/constants.ts
```diff
@@ -32,3 +32,15 @@ export const getInAppAgentInstrumentationObservationId = (runId: string) =>
 // level telemetry and feedback aggregate on the same per-turn trace.
 export const getInAppAgentInstrumentationTraceId = (runId: string) =>
   `${runId}-trace`;
+
+// Per-LLM-call generation ids stay deterministic from the run so retries and
+// late flushes can address the same observation.
+export const getInAppAgentLlmCallObservationId = (
```
LysanderKie, line 38: Is that still used and if so, by what? AS I think we skipped that no? We have a fixed/static name now which is the model name for each run

web/src/__tests__/server/in-app-agent-stream.servertest.ts
```diff
@@ -459,13 +485,139 @@ describe("IN_APP_AGENT_LANGFUSE_MCP_TOOL_POLICIES", () => {
...
     vi.clearAllMocks();
+    bedrockMocks.streamParts = [];
     promptMocks.getPrompt.mockResolvedValue({
       name: "in-app-agent-system-prompt",
       version: 2,
       compile: promptMocks.compile,
     });
   });
 
+  const initializeBasicTracedAgent = async (runId: string) => {
```
LysanderKie, line 496: Can we remove those tests as those test very bedrock specifics/internals and should not be needed? We should rather test on a higher interface to be independent from internal

LysanderKie: E.g. can we add one "integration test" which also covers that core instrumentation is triggered?

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -179,10 +168,15 @@ describe("InAppAgentInstrumentation", () => {
       }),
     );
     expect(
-      mocks.handler.langfuse.enqueue.mock.calls[0]?.[1].metadata,
+      mocks.handler.langfuse.enqueue.mock.calls.find(
+        ([type]) => type === "tool-create",
+      )?.[1].metadata,
     ).not.toHaveProperty("toolCallApproval");
-    expect(mocks.agentGeneration.update).toHaveBeenCalledWith(
+    expect(mocks.handler.langfuse.enqueue).toHaveBeenCalledWith(
```
LysanderKie, line 174: Did we change anything how we ingest in regards to what endpoints/flow we use? We should make sure we use the v4 flow and not any legacy/deprecated features/endpoints.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -723,20 +766,7 @@ describe("InAppAgentInstrumentation", () => {
...
-            current_url:
-              "https://cloud.langfuse.com/project/project-1/traces?filter=value",
-            browser_languages: ["en-US", "en", "de"],
-            current_trace: { id: "trace-1" },
-          },
-        },
-      }),
-    );
+    expect(mocks.trace.generation).not.toHaveBeenCalled();
     expect(mocks.handler.langfuse.trace).toHaveBeenCalledWith(
```
LysanderKie, line 769: Did we change anything how we ingest in regards to what endpoints/flow we use? We should make sure we use the v4 flow and not any legacy/deprecated features/endpoints.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -998,67 +1278,506 @@ describe("InAppAgentInstrumentation", () => {
...
+    ]);
+    instrumentation.end({});
+
+    expect(mocks.handler.langfuse.enqueue).not.toHaveBeenCalledWith(
+      "tool-create",
+      expect.objectContaining({ id: "tool-awaiting-approval" }),
+    );
+  });
+
+  it("closes an open LLM step as ERROR when the run is aborted", () => {
```
LysanderKie, line 1612: Lets challenge if that is really the best behavior? Aborted by whom? Worker dying, then this might be ok. If stopped/aborted by user, then no.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -998,67 +1278,506 @@ describe("InAppAgentInstrumentation", () => {
...
+          total: 1_150,
+        },
+        metadata: expect.objectContaining({
+          finish_reason: "tool-calls",
+        }),
+      }),
+    );
+  });
+
+  it("keeps a provider-finished generation successful when the run later fails", () => {
```
LysanderKie, line 1698: Can we remove this test or is it really required?

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -1069,8 +1788,87 @@ describe("InAppAgentInstrumentation", () => {
...
 
-    expect(mocks.agentGeneration.update).toHaveBeenCalledWith(
-      expect.not.objectContaining({ usageDetails: expect.anything() }),
+    expect(mocks.handler.langfuse.enqueue).not.toHaveBeenCalledWith(
+      "generation-create",
+      expect.anything(),
+    );
+  });
+
+  it("applies late step-finish usage to the last generation after end()", () => {
```
LysanderKie: Not critical test, can we remove it?

LysanderKie:
Not critical test, can we remove it?

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
```diff
@@ -1069,8 +1788,87 @@ describe("InAppAgentInstrumentation", () => {
...
+      }),
+    );
+    expect(
+      mocks.handler.langfuse.enqueue.mock.calls.filter(
+        ([type]) => type === "generation-create",
+      ),
+    ).toHaveLength(1);
+  });
+
+  it("clears a provisional generation error when a late step finish reports usage", () => {
```
LysanderKie: Can we remove this test or is it really required?

packages/shared/src/features/inAppAgent/types.ts
LysanderKie: No browser caller passes these tracing fields. decideToolApproval derives them server-side and stores them in the internal durable run request because a continuation can be picked up by a different worker. I kept that boundary and added the tracing-lineage comment in 7d5ce4f96.

packages/shared/src/features/inAppAgent/types.ts
LysanderKie: traceStartedAt preserves the original root timestamp when later workers emit continuation events with the same trace and root observation IDs. Without it, a continuation could shift the logical start of the complete turn. It is derived and propagated internally, not supplied by the browser.

packages/shared/src/in-app-agent/constants.ts
LysanderKie: This helper derives the deterministic observation ID for each model invocation, not its display name. Instrumentation uses it for generation creation; getInAppAgentLlmCallName separately uses the model ID as the visible name. Keeping it.

web/src/__tests__/server/in-app-agent-stream.servertest.ts
LysanderKie: Simplified in 7d5ce4f96: removed the separate provider happy-path test and retained one fail-open boundary test that proves finish interception cannot alter stream forwarding. The existing createAgUiStream lifecycle case already covers core instrumentation creation, event recording, step callbacks, end, and flush, so I did not add another heavily mocked integration test.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: This PR does not change the ingestion route. Internal tracing still exports local events through processEventBatch: it writes the legacy trace and observation tables and, when V4 write mode is not legacy, also propagates them to the events table. No deprecated HTTP endpoint is used. Moving internal tracing to events-only would be a separate migration.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: Same answer as the other ingestion thread: the event shape changed, not the ingestion route. processEventBatch keeps the existing legacy write plus events-table propagation under the V4 rollout mode; there is no deprecated endpoint introduced here.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: Agreed that the source matters. The runtime distinguishes cancelled, fenced, and worker_shutdown, but instrumentation currently receives only aborted. Per our discussion I am leaving this unchanged until we test user cancellation on the preview; keeping this thread open for that result.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: Removed in 7d5ce4f96 together with the production late-generation-update machinery. Provider finish is now the single fallback when onStepFinish is delayed or absent, and callbacks after instrumentation ends are ignored.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: Removed in 7d5ce4f96 together with the redundant late-update state and generation-update path. The provider finish fallback remains covered.

packages/shared/src/in-app-agent/server/runLifecycle.ts
LysanderKie: Added the requested one-line comment in 7d5ce4f96: Preserve tracing lineage across durable approval continuation runs.

web/src/__tests__/server/unit/in-app-agent-instrumentation.servertest.ts
LysanderKie: Keeping this focused regression. It proves that a model invocation already completed by the provider remains successful when the following tool or root run fails. That is distinct from testing that the root is marked as failed and protects the preview regression fixed in this PR.

PR #16045, opened 2026-08-12
feat(filters): search the filter sidebar by facet name

nkabardin:
🟡 **Live preview: [pr-16045.preview.langfuse.com](https://pr-16045.preview.langfuse.com)**

## TL;DR

The tracing filter sidebar lists **41 facets**. Finding one meant scrolling the whole
catalogue. This adds a name search above the facet list, and gives the "Add filter"
picker the same search. **Selected filters stay visible even when they don't match.**

Typing `sdk` takes the list from 41 facets to 2.

## Why

The v4 upgrade walkthrough kept needing the SDK / instrumentation facets — SDK Name,
SDK Version, Ingestion Source, API Key — which sit near the bottom of a 41-entry list.
Scrolling a long sidebar hunting for a filter *by name* is the whole problem.

The same component serves observations, sessions and scores, so this is generic rather
than traces-only. The active-only **"Add filter" picker** had it worse: ~30 entries in a
288px scroller with **no search at all**.

## What

- **A search row under the sidebar header**, on catalogues longer than a dozen facets —
  a 3-facet sidebar (eval logs, monitors) gains no chrome. It filters labels as you
  type, and matches the underlying column key too, so `userid` finds "User ID" (a
  label-only match misses it — the space defeats it).
- **Selections are never hidden.** An active or explicitly added facet stays listed
  whether or not it matches, so a query can't hide what is currently filtering.
- **An explicit dead end:** `No filters match "xyz"`, or `No other filters match "xyz"`
  when pinned selections are still listed above it — so the message never contradicts
  the facets on screen.
- **Escape clears the query**; once empty, a second Escape closes the sheet as before.
  Clearing restores the full list and never touches filter state.
- **The "Add filter" picker becomes a searchable command list**, the pattern already
  used by the metadata-key picker in the same sidebar. An empty query keeps its
  most-recently-used order.

## Result

Both long-list surfaces in the sidebar are now searchable by name, on desktop and in the
mobile Filters sheet. Verified live on the tracing table: 41 → 2 facets on `sdk`, a
selected facet surviving a non-matching query, Escape restoring all 41, and
type-`sdk`-then-Enter adding the SDK Name facet from the picker.

<details>
<summary>Mechanism, ordering invariant, a11y, analytics, verification</summary>

**One matching authority.** `facetNameRank` / `rankFacetsByName` join `rankFacetOptions`
in `features/filters/lib/facet-display.ts`, built on the search bar's `filterRank`, so
both surfaces rank identically: prefix matches before substring matches, best of
(label, column key). The facet **list** filters with the predicate and keeps its own
order; the **picker** ranks, because match quality is the only order that matters there.

**Search does not perturb facet order.** The sidebar deliberately freezes facet order
while someone works the list (`facetInteractionRef` + `advanceFacetOrder`) and re-settles
on an external change. Typing a query is not "working the list": the search input lives
*above* the scroll area, outside the facet list's `onKeyDownCapture`, so a filter applied
from the search bar mid-search still re-settles the order instead of teleporting or
freezing. Pinned by a covering test.

**Follow-scroll still resolves.** `scrollIntoView` finds facets by `data-facet-column`,
and filtering rows out of the DOM could make those lookups miss — except the facets those
two call sites target (a facet that just became active; a facet just added from the
picker) are exactly the ones the pin rule keeps mounted.

**Composition with active-only mode.** That mode has no catalogue left in the list, so
the list's search box steps aside and the picker's carries the search. Toggling the mode
clears the query rather than leaving a hidden one narrowing the list.

**A11y / keyboard.** A labelled input in normal tab order, a labelled clear button, and
Escape-to-clear that only stops propagating while there is something to clear. No new
global hotkey: `/` belongs to the query bar that is meant to become the primary filter
surface, not to the sidebar.

**Naming.** The sidebar already had a `placeholder="Search"` — a *value* input on string
facets — and a "Filter values" box inside categorical facets. The new one is
"Search filters", unambiguous against both.

**Analytics.** One new event, `filters:facet_search`, fired once per search session (the
first keystroke) per surface, carrying `tableName` / `surface` / `isV4` and **never the
query text**. Paired with the existing `filters:facet_added`, it gives a
search → add funnel. Pinned by a test that asserts the payload shape and the
once-per-session dedup.

**Verification.**
- `pnpm --filter web run test-client src/features/filters src/components/table` —
  `Test Files 27 passed (27)`, `Tests 279 passed (279)`, including 6 new sidebar cases and
  2 new pure-ranking cases.
- Both load-bearing new tests were confirmed to **fail** against the behaviour they pin
  (making typing count as a list interaction; dropping the keep-selected clause).
- `pnpm run lint` — `Tasks: 7 successful, 7 total`. `pnpm --filter web run typecheck` — clean.
- Live: desktop sidebar and the mobile Filters sheet, both flows end-to-end.
- Not run: a production build. This change adds no module-scope constants, dynamic
  imports or otherwise bundle-shaped code, and CI's client-bundle scan covers the class.

</details>

This PR adds facet-name search to long filter sidebars and the active-only “Add filter” picker while preserving selected facets and existing ordering.

- Adds shared label-and-column matching and ranking helpers.
- Introduces searchable sidebar and command-picker interfaces with empty-result and Escape handling.
- Adds query-free, once-per-session search analytics.
- Covers matching, selection visibility, ordering, short catalogs, picker search, and analytics with client tests.
</details>

The PR appears safe to merge with no concrete blocking or independently actionable non-blocking issues identified.

The changed search, picker, ordering, selection-preservation, and analytics paths remain consistent with existing filter identity and component lifecycle contracts.
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(filters): search the filter sidebar..."](https://github.com/langfuse/langfuse/commit/55a83866ba0feca4c8620f5182f350f35b1fed06) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52650243)</sub>


nkabardin:
@claude review

PR #16054, opened 2026-08-12
perf(trace): lay out large trace graphs off the main thread

nkabardin:
🟡 **Live preview: [pr-16054.preview.langfuse.com](https://pr-16054.preview.langfuse.com)**

## TL;DR

A large trace graph used to freeze the whole tab for as long as ELK took — a customer reported ~10s of dead UI — and the size budget that protected against it meant very large graphs got **no graph at all**. Graph layout now runs in a Web Worker: the tree, tabs and navigation stay live throughout (measured: 60fps, longest frame gap 46ms, tree click 87ms, navigation away 58ms *during* a 27s layout), the size ceiling rises 8× so previously-refused graphs render, and a layout that will never finish is bounded by a 60s deadline instead of a guess about counts.

## Why

`elkjs` is synchronous and uninterruptible, and it ran on the main thread. Two bad outcomes existed side by side:

1. **Under the budget** → the tab froze for the whole layout. Reproduced on a seeded graph: 250 nodes / 247 connections blocks for 27s in Chrome, 42s in Firefox.
2. **Over the budget** (250 connections / 500 nodes) → ELK was never started and the pane showed "too large to lay out". The budget had to be that tight *because* a slow layout froze the tab.

## What

- **Layout runs in a worker.** `web/src/workers/elk-layout.worker.ts` hosts elkjs's own worker build (`elk-worker.min.js`), and the main thread talks to it through `new ELK({ workerFactory })`.
- **The main thread keeps only cheap work**: dedupe the edges, reduce the observation map to per-node label widths, apply the count ceiling, build the ELK graph and map the result — all O(nodes + edges). `postMessage`'s structured clone is main-thread work too, so the payload stays the deduped graph, never the raw multigraph.
- **Cancellation is real.** Requests are id-tagged; switching trace, view mode or direction aborts, and a stale result can never land on a newer graph. Because ELK cannot be interrupted, cancelling a layout that has been running a while *terminates the worker* and re-issues anything queued behind it on a fresh one.
- **A 60s wall-clock deadline** replaces counts as the real budget, enforced the same way (terminate the worker) and surfaced as the existing "too large" notice with the graph's counts.
- **The count ceiling rises to 2,000 connections / 2,500 nodes** and is now a memory bound, not a time budget — cost tracks graph *shape*, not size: with the same counts, 60 nodes / 300 connections lay out in ~0.35s when connections concentrate on a few nodes and ~16s when they are spread evenly. No count can predict that; the deadline can.
- **The expanded ("as it ran") path goes through the same worker**, so it is bounded by the deadline too.
- Copy: the loading state now says, after four seconds, that the graph is large and the tree/timeline stay usable; a failed layout points at the tree/timeline instead of only offering a retry.
- New seeder scenario `agent-graph` — the shape that was missing locally: a trace that is large as a **graph** (many distinct node names and connections) rather than as a tree.

## Result

Measured against a production build, seeded graphs, Chrome and Firefox:

| Graph (aggregated) | Before | After |
| --- | --- | --- |
| 150 nodes / 147 connections | renders, tab frozen ~2s | renders in 1.5s, tab live |
| 40 nodes / 242 connections | renders, tab frozen ~2s | renders in 2.0s, tab live |
| 250 nodes / 247 connections | renders, **tab frozen 27s** (42s in Firefox) | renders, tab at 60fps throughout |
| 80 nodes / 452 connections | refused: "too large" | **renders in 30s**, tab at 60fps throughout |
| 500 nodes / 496 connections | refused: "too large" | attempted; ELK recursion overflows → notice + alternatives |
| 62 nodes / 1,352 connections | refused: "too large" | attempted for 60s (tab live), then the deadline → "too large (62 nodes, 1,352 connections)" |
| 120 nodes / 5,117 connections | refused: "too large" | refused: "too large" (over the ceiling) |

<details>
<summary>Mechanism, measurements and verification</summary>

**Why elkjs's own worker build, not a hand-rolled one.** The first cut imported `elk.bundled.js` inside a custom worker. That build detects the worker context (no `document`), installs *itself* as the worker's `onmessage` and then exports no in-process worker class — so `new ELK()` threw at module load, the worker never answered, and every layout sat until its deadline. Only a browser run caught it (a unit test with a stubbed worker cannot). `elk-worker.min.js` is exactly the artifact for this, and it already speaks `elk-api`'s message protocol.

**The split.** `layout/elkLayout.ts` stays free of worker plumbing and directly testable: `prepareGraphLayout` (dedupe → counter widths → ceiling) and `runGraphLayout(elk, request)` run unchanged on either thread. `layout/graphLayoutWorkerClient.ts` owns the worker singleton, request ids, cancellation, the deadline, and the fallback. A browser without Workers, or a worker script that fails to load (stale chunk after a deploy), lays out on the main thread — slow beats no graph.

**Main-thread responsiveness**, sampled with a rAF heartbeat during a 27s layout: 60fps sustained, longest frame gap 46ms (the React render, not ELK); a tree row click answered in 87ms and wrote `?observation=` to the URL; `about:blank` navigation completed in 58ms. Mid-layout view-mode switch: the expanded graph landed in 724ms, the worker log shows terminate + restart, and the superseded aggregated result never appeared over the next 15s.

**Common case.** A 5-node graph: longest frame gap 28-38ms with the worker vs 25-35ms without — no main-thread regression. Wall-clock to first paint of the graph is ~65ms higher because the worker parses elkjs itself; that time sits inside the loading state the pane already had, and it is time the main thread now does *not* spend parsing 1.6MB of elkjs.

**Recursion, in both browsers.** Very large aggregated graphs still exceed ELK's recursion depth — "Maximum call stack size exceeded" in Chrome, "too much recursion" in Firefox, both at ~500 nodes with deep layering. Worker stacks are smaller than the main thread's, so this changed shape rather than disappearing; it is caught across the worker boundary and rendered as a notice with alternatives, never an empty pane. (A follow-up worth measuring: `elk.layered.cycleBreaking.strategy` is `DEPTH_FIRST`, which is recursive; `GREEDY` may lift that ceiling. Out of scope here — it changes layout aesthetics for every graph.)

**Note on measuring this.** `next dev` is ~20× slower at ELK than a production build (a 242-connection graph: 45s in dev, 2.0s in prod). Every number above is from `pnpm --filter=web build` + `next start`.

**Test data.** `pnpm run seed -- agent-graph --v4` gives ~1,350 distinct connections from 350 observations; `--nodes 120 --steps 100 --parallel 8` crosses the ceiling, `--nodes 250 --steps 250 --parallel 1` is the 27s-layout case.

**Checks.** `turbo run lint`: 7 successful, 7 total. `tsc --noEmit`: clean. `vitest --project client src/features/trace-graph-view`: 21 passed (21). Production build: exit 0, `scan:client-bundle` clean. Browser passes in Chrome and Firefox against the production build.

</details>


nkabardin:
@claude review

PR #16067, opened 2026-08-13
fix(worker): fail-fast and auto-disable PostHog integration on customer-config faults

niklassemmler:
🟡 **Live preview: [pr-16067.preview.langfuse.com](https://pr-16067.preview.langfuse.com)**

## What does this PR do?

A deterministic customer-config fault in the PostHog integration — a bad or malicious PostHog hostname rejected by the outbound-URL/SSRF validator — was treated as an infra job failure. It logged at `error`, was rethrown, retried 5× by BullMQ, and tripped the PostHog processing-failures monitor on every attempt of every scheduled run, forever, because nothing recorded the fault or disabled the integration.

This classifies those faults as customer-config problems and gives them the treatment blob storage already has: persist the error, auto-disable the integration, notify the customer, and stop the alert noise.

**Fail-fast, not retry-then-disable.** The failures metric is emitted from BullMQ's `Worker.on("failed")` event, so it increments on *every* failed attempt — not only on terminal exhaustion. Any throw therefore trips the monitor, even one that disables on the final attempt. Since these faults are deterministic and cannot self-heal, the handler now disables and **resolves** the job on the first classified occurrence. Retries would add no signal.

What changed:

- **Classification.** The deterministic-fault classifier moved from `worker/src/features/blobstorage/isCustomerFaultError.ts` to a neutral `worker/src/features/integrations/customerFaultClassification.ts`; the blob module is now a thin re-export that keeps its own disabled-metric constant. Behavior is unchanged — this is a relocation, not a rewrite.
- **Allowlist as the transient filter.** Only deterministic outbound-URL codes disable: `blocked-hostname`, `blocked-ip`, `invalid-syntax`, `invalid-encoding`, `https-required`, `protocol-not-allowed`, `url-credentials-not-allowed`. `dns-lookup-failed` is deliberately excluded — resolvability depends on runtime resolver state, not config, so a transient DNS outage must never disable a working integration.
- **Cause preservation.** The hostname throw site rewrapped into a bare `Error`, discarding `.code`/`.name`. It now rethrows with `{ cause }` so the classifier can walk the chain.
- **Single classify-catch.** Both former throw sites (the hostname pre-check and the export catch) funnel into one catch that owns the disable-or-rethrow decision. Customer fault → persist `lastError`/`lastErrorAt`, atomically flip `enabled` true→false, record a disabled metric tagged by reason, notify once, return. Anything else → log at `error` and rethrow, so infra faults keep retrying and alerting exactly as before.
- **Notification.** New `posthog-export-failed` project-notification event with a dedicated email, mirroring the blob variant. The disable notification is sent exactly once, guaranteed by the atomic `enabled: true` claim.
- **Recovery.** A successful run clears the persisted error state; saving the integration as enabled resets the failure-notification cooldown.

Three nullable columns are added to `posthog_integrations` (`last_error`, `last_error_at`, `last_failure_notification_sent_at`). Additive, no backfill.

Mixpanel is out of scope — it takes no customer-supplied hostname, so this fault class does not apply. Aligning blob storage to the same fail-fast behavior is a tracked follow-up; blob keeps retry-then-disable here.

## Type of change

## Mandatory Tasks

## Checklist

- New and existing unit tests pass locally with my changes


niklassemmler:
* **#16067** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16067?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16067?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
Superseded by a three-PR stack. This PR had grown to ~1300 changed lines across refactorings, the behaviour fix, and new UI, which is more than one review can reasonably hold.

Split into:

- #16094 — `refactor(worker,shared)`: share integration helpers across export integrations. Relocates the customer-fault classifier to a neutral module, extracts the Prisma "record deleted mid-run" predicate, and parameterises the export-failure email so it exists once instead of once per integration. Behaviour-preserving.
- #16095 — `fix(worker)`: fail fast and auto-disable the PostHog export on config faults. The actual bug fix, stacked on the refactor.
- #16096 — `feat(web)`: show the persisted PostHog export fault in settings.

Two changes were made while splitting, rather than a straight re-cut:

- The PostHog failure email is no longer a copy of the blob storage one. Only 20 of 86 sender lines and 34 of 99 template lines actually differed, so the mechanics and template are now shared and parameterised — the email directory shrank by 95 lines and the PostHog-specific portion by 157. Blob's rendered output is unchanged: same subjects, visible body text, links, and log strings.
- The commit that made the project-notification event list exhaustive by construction was dropped. It was unrelated to this bug and belongs on its own.

Review feedback from this PR has been carried across; thanks to everyone who commented here.

PR #16068, opened 2026-08-13
fix(evals): align metadata contains-filter key-existence semantics

### LINKED ISSUE #16059: bug: mismatch between evaluator filter preview and actual executed traces
### Describe the bug

When configuring an LLM-as-a-Judge evaluator with a trace filter of metadata -> 'turn' -> contains -> '' (empty string), the setup preview behaves as a key-existence check: only traces that have a turn key in metadata are shown as matching.

However, once the evaluator is active, the worker enqueues and executes the evaluator on incoming traces that do not have the turn metadata key at all. The preview and the trigger path evaluate the same filter differently, so the filter that was validated visually in the wizard does not describe the population that actually gets judged (and billed).

# Expected behavior

Preview and worker should compile the filter identically. Whichever semantics is intended for contains "" on a metadata key (existence check vs. match-everything), both paths should agree — the preview is what users rely on to validate evaluator scope before spending eval tokens.

# Alternative

Please add a metadata key existence filter. 

### Steps to reproduce

1. Ingest traces via the OTel endpoint, some with metadata.turn = "<value>" set as a span/trace attribute, some without any turn key.
2. Create an evaluator (new traces) with filter: metadata → key turn → operator contains → value "" (empty string).
3. Observe the preview/matching sample in the wizard: only traces with the turn key appear. ✅
4. Activate the evaluator and ingest new traces without the turn key.
5. Check the evaluator log / job_executions: executions are created for the keyless traces as well. ❌

Self-hosted

4.10.0

_No response_

### Additional information

_No response_

Yes

wochinge:
🟡 **Live preview: [pr-16068.preview.langfuse.com](https://pr-16068.preview.langfuse.com)**

## What
Fixes a mismatch between the evaluator wizard's filter preview and the filter actually applied when triggering an evaluator: a metadata filter like `metadata -> key -> contains -> ""` was meant to behave as a key-existence check, but two of the three places that evaluate this filter didn't require the key to exist.

## Why
- The in-memory evaluator (used by the worker to decide whether to trigger an evaluator on a freshly ingested trace) coalesced a missing metadata key to an empty string, so `contains ""` matched every trace, including ones that never had the key.
- The ClickHouse SQL filter for the legacy `traces`/`observations` Map column had the same gap: accessing a missing key returns the Map's empty-string default, so the same comparison matched rows without the key too.
- Only the newer events-table SQL path (fixed in #13369) had a key-existence guard, which is why the wizard preview (which queries that path) looked correct while the worker's real trigger didn't.

## Fix
Both the in-memory evaluator (`InMemoryFilterService`) and the legacy-table SQL filter (`StringObjectFilter`) now require the key to exist before evaluating any operator (`=`, `contains`, `does not contain`, `starts with`, `ends with`), matching the events-table semantics.

## Tests
- Extended `inMemoryFilterService.test.ts` with a missing-key + empty-value case for `contains` and `=`.
- Extended `evalService.filtering.test.ts` with an integration test that exercises the real ClickHouse DB-fallback path and confirms only the trace with the key gets an eval job.

Fixes https://github.com/langfuse/langfuse/issues/16059

The PR aligns metadata key-existence semantics across in-memory evaluator matching and legacy ClickHouse Map filtering.
- Adds `mapContains` guards to legacy ClickHouse string-object predicates.
- Adds a missing-key guard to in-memory string-object evaluation.
- Extends unit and integration coverage for empty-value metadata filters.
</details>

The inherited-property mismatch should be fixed before merging because it can still cause evaluator triggers to disagree with database-backed previews.

The database path now checks true Map membership, while the in-memory path still accepts inherited properties as present metadata keys for unrestricted user-supplied filter names.

**Files Needing Attention:** packages/shared/src/server/services/InMemoryFilterService.ts
</details>

```mermaid
flowchart LR
  F[Metadata filter] --> P{Evaluation path}
  P -->|Cached trace| M[In-memory object lookup]
  P -->|DB fallback| C[ClickHouse Map filter]
  M --> K1[Require own metadata key]
  C --> K2[mapContains metadata key]
  K1 --> V[Evaluate string operator]
  K2 --> V
  V --> J{Create evaluator job}
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/services/InMemoryFilterService.ts:331-332
**Prototype properties bypass key guard**

When a metadata filter uses an inherited name such as `toString` or `constructor`, bracket lookup resolves the prototype property and passes the `undefined` guard, causing the in-memory worker to create evaluator jobs for traces that the ClickHouse key-membership checks reject.

```suggestion
    const recordValue = fieldValue as Record<string, unknown>;
    if (!Object.prototype.hasOwnProperty.call(recordValue, key)) {
      return false;
    }
    const objectValue = recordValue[key];
    if (objectValue === undefined) {
```

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(evals): make metadata contains-filte..."](https://github.com/langfuse/langfuse/commit/d996ead376741ccd83f8b7b8848147698fa99124) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52920306)</sub>

**Context used:**

- Knowledge Base — [Evaluation Pipeline (LLM-as-Judge)](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/evaluation-pipeline.md)


wochinge:
@claude review

PR #16076, opened 2026-08-13
fix(batch-export): route batch export ClickHouse reads to a read replica

niklassemmler:
🟡 **Live preview: [pr-16076.preview.langfuse.com](https://pr-16076.preview.langfuse.com)**

## What does this PR do?

Routes every ClickHouse read on the batch export path to a read replica instead of the primary writer cluster.

ClickHouse read-replica routing is opt-in per query: a read only reaches a replica if the caller passes `preferredClickhouseService`. The default is `"ReadWrite"`, which resolves to the primary. The events-table export paths already opted in; the older non-events paths did not — they forwarded only extended timeouts and tags, so they inherited the default. Batch exports are long-running, high-volume scans, which is exactly the workload that should never touch the writer. A trace export scanning the primary was enough to overload it.

Six reads are now routed:

| Export path | Read |
| -- | -- |
| Traces | main traces+scores stream query |
| All exports | `getDistinctScoreNames` — reached by every export stream |
| Observations (non-events) | main stream query |
| Scores (non-events) | `getScoresUiGeneric` |
| Sessions (non-events) | `getSessionsTableGeneric` |
| Dataset run items | `getDatasetRunItemsTableInternal` |

Dataset items and audit log exports are unaffected — they read Postgres only.

### How the routing is wired

The literal `"ReadOnly"` is introduced in **one** place: a `BATCH_EXPORT_CLICKHOUSE_SERVICE` constant in the batch export job, passed at two call sites. Everything below takes `preferredClickhouseService?: PreferredClickhouseService` with **no default**, so an absent value still resolves to the writer.

That distinction is load-bearing. Several of these functions are shared with the batch-action handler, which powers "add all to annotation queue" and "add all to dataset". Those enumerate a selection the user made against a list read from the primary, so moving their enumeration to a replica would silently create fewer items than the user selected whenever the replica lagged. Threading the option instead of pinning it keeps every non-export caller byte-identical.

The rule the diff follows: **literals only for worker-local export builders with a verified single production caller; parameters for anything in `packages/shared`** — a shared repository should not hard-code a worker-export policy. Each remaining literal carries a comment naming the invariant that licenses it, so the next person to add a caller sees why it was safe and notices when it stops being.

Where an export issues two reads whose results must agree, both go to the same service. The score-name list is a hard filter on exported score values — a score whose name is missing from the list is dropped from the output — so a name read lagging behind the value read would silently remove a column from the file. Both reads are pinned together on every path.

### Self-hosted deployments are unaffected

`getClickhouseUrl` falls back to the primary URL when the read-replica URL is unset, so a deployment without it gets identical behavior and reuses the same cached client.

### Tests

A new suite points the three ClickHouse URLs at recording pass-through proxies and asserts which cluster each statement actually reached, rather than asserting on the arguments passed. It observes the cluster, not the mechanism, so it survives a redesign of how routing is threaded.

Every routed call site was verified by mutation: reverting any one of them to the primary turns exactly one test red. That includes the two contract tests proving the shared entry points still default to the writer when no service is passed, which is what protects the batch-action paths. Control tests establish that each proxy sink can be populated, so a green result cannot mean "the harness observed nothing" — including one control that fails if the events-replica channel it proves ever breaks.

Worth knowing for future changes: outside this suite, routing is unobservable in the test environment, because the read-replica URL is unset there and every service collapses to one endpoint.

## Type of change

## Mandatory Tasks

## Checklist

- Verified with `pnpm turbo run lint` (`Tasks: 6 successful, 6 total`), `pnpm turbo run typecheck` (`Tasks: 7 successful, 7 total`), and the new suite (`Tests 19 passed (19)`).
- Production impact also depends on the read-replica URL being populated per region; that is a deployment precondition rather than a code change, and is worth confirming alongside this.

The PR routes ClickHouse-backed batch exports away from the writer while preserving primary routing for shared batch-action and UI paths.

- Adds optional ClickHouse service selection to shared score, session, and dataset-run-item queries.
- Routes trace and observation export queries consistently to their appropriate replicas.
- Adds integration-style regression coverage for export and non-export routing behavior.
</details>

The changed paths consistently route each export’s related ClickHouse queries to the intended service, while omitted service preferences retain existing primary routing for shared non-export callers.
</details>

```mermaid
flowchart TD
  Job[Batch export job] --> Table{Export table}
  Table -->|Traces| Trace[getTraceStream]
  Table -->|Observations, legacy| Observation[getObservationStream]
  Table -->|Events-backed data| Events[Events stream]
  Table -->|Scores, sessions, dataset run items| Generic[Paginated stream]
  Trace --> ReadOnly[(ReadOnly)]
  Observation --> ReadOnly
  Generic --> ReadOnly
  Events --> EventsReadOnly[(EventsReadOnly)]
  Actions[Batch actions / UI callers] -->|No preferred service| Primary[(ReadWrite)]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["test(batch-export): close routing covera..."](https://github.com/langfuse/langfuse/commit/e664b08d42179259b43b8d637d4d7a5c708b9b4f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52956108)</sub>

**Context used:**

- Knowledge Base — [Batch Actions and Data Exports](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/batch-actions-exports.md)

### DIFF
--- a/packages/shared/src/server/repositories/dataset-run-items.ts
+++ b/packages/shared/src/server/repositories/dataset-run-items.ts
@@ -544,7 +544,6 @@ type GetDatasetRunItemsTableOpts<IncludeIO extends boolean> =
   DatasetRunItemsTableQuery & {
     select: "count" | "rows";
     includeIO?: IncludeIO;
-    // Absent means "ReadWrite": only batch export opts into the read replica.
     preferredClickhouseService?: PreferredClickhouseService;
   };
 
@@ -915,13 +914,6 @@ const getDatasetRunItemsTableInternal = async <
 
 export const getDatasetRunItemsCh = async (
   opts: DatasetRunItemsTableQuery & {
-    // Reached only through the worker's paginated read stream, which serves
-    // both the batch export and the eval-create batch action. The export passes
-    // "ReadOnly"; the batch action passes nothing on purpose, so historic eval
-    // creation keeps enumerating from the primary. Do not pin a service here —
-    // that would hide unreplicated run items from eval creation. The four
-    // sibling wrappers cannot opt in at all, because this field is absent from
-    // DatasetRunItemsTableQuery.
     preferredClickhouseService?: PreferredClickhouseService;
   },
 ): Promise<DatasetRunItemDomain[]> => {
--- a/packages/shared/src/server/repositories/dataset-run-items.ts
+++ b/packages/shared/src/server/repositories/dataset-run-items.ts
@@ -915,11 +915,13 @@ const getDatasetRunItemsTableInternal = async <
 
 export const getDatasetRunItemsCh = async (
   opts: DatasetRunItemsTableQuery & {
-    // Export-only wrapper: batch export is its one production caller. The four
-    // sibling wrappers share the same internal query but cannot opt in, because
-    // this field is deliberately absent from DatasetRunItemsTableQuery. Moving
-    // it onto that shared type would put the UI and public API on the replica
-    // too.
+    // Reached only through the worker's paginated read stream, which serves
+    // both the batch export and the eval-create batch action. The export passes
+    // "ReadOnly"; the batch action passes nothing on purpose, so historic eval
+    // creation keeps enumerating from the primary. Do not pin a service here —
+    // that would hide unreplicated run items from eval creation. The four
+    // sibling wrappers cannot opt in at all, because this field is absent from
+    // DatasetRunItemsTableQuery.
     preferredClickhouseService?: PreferredClickhouseService;
   },
 ): Promise<DatasetRunItemDomain[]> => {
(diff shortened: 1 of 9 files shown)


niklassemmler:
* **#16076** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16076?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16076?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #16083, opened 2026-08-13
fix(ui): keep sr-only text from escaping its scroll container

nkabardin:
🟡 **Live preview: [pr-16083.preview.langfuse.com](https://pr-16083.preview.langfuse.com)**

## TL;DR

A screen-reader-only span could make the whole page scroll. `sr-only` is absolutely positioned with auto insets, so it lands at its static position but is only clipped by boxes that are its containing block — which the scroll container it lives in usually isn't. Pinning it to `top: 0; left: 0` keeps it inside whatever should clip it. Independent of #16080 / #16081; branched off `main`.

## Why

Root-causing extra scrollbars on the v4 upgrade panel ended at a **1px invisible span**: the V3/V4 toggle's `aria-describedby` text. It sat ~593px down the panel's own scroll flow, escaped both the panel's `overflow-y-auto` and the panel group's `overflow: hidden`, and extended the app shell's scrollable overflow — so the document scrolled and, with space-taking scrollbars, cascaded into four visible scrollbars.

The reason is a CSS rule that's easy to miss: **a scroll container clips only the descendants it is a containing block for.** An absolutely-positioned box skips every static ancestor and resolves against the nearest positioned one, so it is invisible to the clips in between while still contributing its geometry to that further-out ancestor's scroll area.

#16080 fixes the containing block on the drawer, which is the right structural fix. This closes the class: no `sr-only` anywhere can do it again, whatever container it ends up in.

## What

Override the `sr-only` utility in `globals.css` with Tailwind's own declarations plus `top: 0; left: 0`. Instead of sitting at its static position — potentially thousands of pixels down a scroll flow — the box sits at its containing block's origin, where every enclosing clip contains it. It stays absolute, out of flow, 1×1, clipped and invisible; nothing about layout or the accessibility tree changes.

## Result

With only this change and neither sibling PR applied, the reported bug is gone across the entire failing viewport band (540/580/620px tall × 1190/1440px wide, space-taking scrollbars, panel open): document overflow **0/0** everywhere, versus up to 48px vertical + 15px horizontal before.

<details>
<summary>Safety review of all 45 call sites, and why there is no new test</summary>

**Computed styles** confirm the override lands: every `.sr-only` element resolves to `position: absolute; top: 0px; left: 0px`, 1×1, `clip: rect(0,0,0,0)`, `overflow: hidden`. Page renders pixel-identically.

**All 45 usages** are non-focusable label/description spans plus `DialogTitle` / `DrawerTitle` / `SheetTitle` elements. Specifically checked:

- **No `not-sr-only`, no `focus:` variants** anywhere in the app, so the "skip link becomes visible on focus" pattern isn't in play. Were it added later, `not-sr-only` sets `position: static`, which makes both offsets inert — it still wins.
- **Three sites combine `sr-only` with `relative`** (`eval-templates-table`, `DatasetItemsTable`, `DatasetRunsTable`). Whichever `position` wins is unchanged by this PR; if `relative` wins, `top: 0; left: 0` offsets a relatively-positioned box by zero, which is a no-op.
- Nothing relies on an `sr-only` box's position — it is invisible by construction.

**No new test.** The escape needs real layout, so it can only be asserted in a browser; #16080 adds `app-shell-overflow.spec.ts`, which pins exactly this invariant (an absolutely-positioned box in a drawer must not scroll the document) and would catch a regression of this class. A test specific to this PR could only assert a CSS constant, which is not worth pinning. Verified by measurement instead, as above.

Checks: `pnpm run lint` → `Tasks: 7 successful, 7 total`; `prettier --check` clean. CSS-only change, so no type surface and no test suite touched.
</details>

This PR prevents visually hidden text from expanding page-level scroll areas by pinning the `sr-only` utility to its containing block’s origin.
- Re-declares Tailwind’s `sr-only` utility with its existing visually hidden properties.
- Adds zero `top` and `left` offsets so absolutely positioned accessibility text remains within enclosing clips.
</details>

The PR appears safe to merge, with no concrete regressions identified in the changed CSS utility.

The override preserves the existing visually hidden declarations and only pins absolutely positioned elements to zero offsets, while static or relative positioning continues to make those offsets inert or harmless.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ui): keep sr-only text from escaping..."](https://github.com/langfuse/langfuse/commit/63bb3d33f53b493ac07e4a9015429741ecbbcbcd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52977360)</sub>


nkabardin:
@claude review

PR #16091, opened 2026-08-13
fix(worker): pin analytics export egress to connect-time SSRF validation

niklassemmler:
🟡 **Live preview: [pr-16091.preview.langfuse.com](https://pr-16091.preview.langfuse.com)**

## What does this PR do?

The PostHog and Mixpanel export workers validated the configured destination host **once**, then sent the data through an unpinned `fetch` that re-resolved DNS independently. That is check-then-use: a host answering a public IP during validation and rebinding to a private or loopback address at connect time bypassed the check entirely, and the exporter would POST customer trace data to whatever the second resolution returned.

Both senders now go through the shared connect-time-pinned egress path already used for webhooks, blob storage and LLM calls, so every resolved IP is re-validated at socket connect, before any bytes leave the worker.

A connect-time block fails terminally, so the queue stops re-attempting a send that cannot succeed. A transient DNS resolution failure is deliberately excluded from that — a resolver blip is not a permanent misconfiguration and must stay retryable.

## Type of change

- [ ] Refactor
- [ ] Documentation update
- [ ] Chore

## Checklist

- [x] Tests added/updated
- [x] `pnpm run lint` and `pnpm run typecheck` pass
- [ ] Documentation updated — not required; no external contract, env var or user-facing surface changes

## Notes for review

- **Redirect budget.** The shared helper requires an explicit budget where the platform previously followed redirects implicitly, so this sets 10. A redirect-budget exhaustion stays retryable, matching current behaviour.
- **Errors are matched by `name`, not `instanceof`.** This is deliberate and worth not "fixing" back: several worker suites replace the shared server barrel with a full mock, so `instanceof` against those imports dereferences `undefined` and the classifier's own `TypeError` replaces the real error. That broke six pre-existing tests in an earlier revision, invisibly to lint and typecheck.
- **The terminal error carries no credentials, anywhere.** A rejected redirect target is reported with its URL embedded verbatim, and a `Location` header may carry userinfo — so a redirect to `http://user:pass@blocked-host/` could put a password into the worker log line and into the `failedReason` BullMQ persists. Redacting the composed message alone was not sufficient: the queue's generic `failed` handler passes the whole error to both the logger and `traceException`, and under `LANGFUSE_LOG_FORMAT=json` winston copies every enumerable field into the log line — including a nested `cause` and its raw `redirectUrl`. The retained cause is therefore a sanitized stand-in, with the URL stripped from its message *and* its stack, attached non-enumerably. Tests assert the whole serializable surface rather than just the message.
- **Both senders share one egress module.** The redirect budget, the validator, the whitelist source and the classify-and-rethrow step all live in `analyticsIntegrationEgress.ts`, so the two exporters cannot drift apart.

## Follow-up hardening, intentionally not here

Kept out to keep this reviewable. None is exploitable today.

- An IP-literal destination check for Mixpanel — the connect-time DNS hook does not fire for literals.
- Redirect fault reporting.
- A wider negative-test matrix.
- Credential redaction on the *shared* outbound-url path, which this PR only narrows rather than closes: `fetchWithSecureRedirects`'s own `logger.warn` and the `MaxRedirectsExceededError` / `CircularRedirectError` chain messages still carry raw URLs for the webhook, blob-storage and LLM callers.

Branch: `chore/analytics-egress-hardening-deferred`.

### DIFF
--- a/packages/shared/src/server/outbound-url/validation.ts
+++ b/packages/shared/src/server/outbound-url/validation.ts
@@ -28,6 +28,11 @@ export class OutboundUrlValidationError extends Error {
   }
 }
 
+// Prefix of the message thrown on DNS-resolution failure. Exported so consumers
+// can detect this failure by message when the typed `code` is lost — e.g. after
+// RedirectValidationError re-wraps the inner error by message text only.
+export const DNS_LOOKUP_FAILED_MESSAGE_PREFIX = "DNS lookup failed for";
+
 export interface OutboundUrlValidationWhitelist {
   hosts: string[];
   ips: string[];
@@ -78,7 +83,7 @@ export async function resolveHost(hostname: string): Promise<string[]> {
   if (!ips.size) {
     throw new OutboundUrlValidationError(
       "dns-lookup-failed",
-      `DNS lookup failed for ${hostname}`,
+      `${DNS_LOOKUP_FAILED_MESSAGE_PREFIX} ${hostname}`,
     );
   }
   return [...ips];
--- a/worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts
+++ b/worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts
@@ -260,12 +260,43 @@ describe("outbound validation classification", () => {
   });
 });
 
+/**
+ * Everything a structured logger or tracer can pull off a thrown error: its
+ * enumerable own fields (winston copies those into the JSON log line), its
+ * message, its stack, and the same again for each `cause` a chain-walking
+ * reporter follows. A credential must appear in none of them.
+ */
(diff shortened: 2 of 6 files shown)


niklassemmler:
* **#16091** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16091?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16091?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
@claude review

PR #16093, opened 2026-08-13
refactor(integrations): derive export-source availability from the v4 write mode

niklassemmler:
🟡 **Live preview: [pr-16093.preview.langfuse.com](https://pr-16093.preview.langfuse.com)**

## What does this PR do?

Export-source availability was decided in several places from several inputs. The PostHog and Mixpanel settings pages hardcoded `enrichedAvailable: true`, asserting the enriched events tables are always readable — so on a deployment that does not write them, the pages offered an export source the exporter cannot serve.

This derives availability from the one thing that actually determines it: the deployment's v4 write mode.

- `areEnrichedWritesActive` joins `areLegacyWritesActive` in the shared policy, so "is this source's data being written" has a single definition.
- Both analytics `get` procedures return the raw write mode instead of a pre-derived `legacyWritesActive` boolean, and every settings page builds its policy context through one `buildExportSourceContext`.
- The blob storage PUT endpoint, the worker's export guard, and the in-transaction race backstops all read from the same predicate rather than their own copies.
- The blocked-save alert for an unavailable enriched source pointed operators at the V4 preview opt-in, which no longer governs availability. It now names the setting that does.

**The per-user V4 beta gate is deliberately left in place here.** It is a policy decision with Cloud-visible consequences, so it is split into the stacked PR on top of this one and can be reviewed and accepted separately. This PR is the mechanical half.

### Also fixed: forms could mount before their inputs resolved

The write mode, the project and the integration row all arrive asynchronously, while `useForm` captures `defaultValues` once at first render. The pages fell back to a fabricated context (`?? true`, and an `isLoading`-only gate that lets a query *error* through with `data === undefined`), so a draft could be seeded from a value the deployment never had.

All three pages now render a skeleton until every input has resolved and mount the form keyed on entity identity, so background refetches cannot reset a draft in progress. This was latent before this PR and is fixed here because the following PR removes the synchronous session value that happened to mask it.

## Type of change

Behaviour changes only where the previous code was factually wrong: a deployment that does not write the enriched tables no longer offers enriched export sources.

## Mandatory Tasks

## Reviewing this

Roughly 60% of the diff is tests. Production code is a net reduction.

The served OpenAPI spec (`web/public/generated/api/openapi.yml`) is regenerated with `fern-api export`, not hand-edited. Note that nothing in CI runs that export, so the tracked spec can drift from its Fern source independently of this PR.

## Verification

- `pnpm turbo run typecheck lint` — 9 successful, 9 total
- `pnpm --filter web run test-client` — 3269 passed, 51 skipped
- `pnpm --filter @langfuse/shared run test` — 913 passed, 2 skipped

Server tests that assert REST behaviour under a specific write mode and region cannot be exercised locally; those are left to CI.

### DIFF
--- a/fern/apis/server/definition/blob-storage-integrations.yml
+++ b/fern/apis/server/definition/blob-storage-integrations.yml
@@ -80,7 +80,7 @@ types:
       - `OBSERVATIONS_V2`: same data model as the `/api/public/v2/observations` endpoint, plus scores. Columns are controlled by `exportFieldGroups`.
       - `LEGACY_TRACES_AND_ENRICHED_OBSERVATIONS`: both sets. Observation columns of both portions are controlled by `exportFieldGroups`.
 
-      **Note:** `OBSERVATIONS_V2` and the enriched-observations portion of `LEGACY_TRACES_AND_ENRICHED_OBSERVATIONS` rely on the enriched observations table (Langfuse Fast Preview / v4), which is currently available on Langfuse Cloud only. See https://langfuse.com/docs/v4.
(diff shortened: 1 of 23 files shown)


niklassemmler:
* **#16149** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16149?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16093** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16093?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #16094, opened 2026-08-13
refactor(worker,shared): share integration helpers across export integrations

niklassemmler:
🟡 **Live preview: [pr-16094.preview.langfuse.com](https://pr-16094.preview.langfuse.com)**

## What does this PR do?

Groundwork so a second analytics-export integration can reuse what blob storage already has, instead of copying it. Three behaviour-preserving extractions, no functional change:

- **Customer-fault classifier** — moved out of the blob-storage feature into a neutral `worker/src/features/integrations/` module. Blob keeps its own disable-metric constant and re-exports the classifier, so its imports and behaviour are untouched.
- **Prisma "record deleted mid-run" predicate** — the P2025 check was a private copy inside the blob handler; it now lives beside the classifier so a future broadening of the code set happens once.
- **Export-failure email** — the sender mechanics (SMTP guards, subject sanitisation, CRM bcc, error logging) and the React Email template existed once per integration. They now exist once in total and take the integration's labels and copy as arguments; blob storage becomes a thin caller supplying its own wording.

On the email refactor, blob's output is deliberately unchanged: same subject lines for both variants, same visible body text, same links, same log strings. The only byte-level difference is React's zero-width text-node separators, which appear where a literal label became an interpolated value — invisible to any recipient. This was verified by rendering all variants before and after and diffing.

Three separate label casings are passed rather than derived from one. Not every integration name lower-cases and title-cases predictably, so a transform would silently corrupt copy for names that don't follow the pattern.

The follow-up PR stacked on this one adds the second integration that consumes all three extractions.

## Type of change

- [x] Refactor (restructures existing code without changing behavior, e.g. simplify logic, split modules, reduce duplication)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I have read the [contributing guide](https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md)
- My code follows the style guidelines of this project
- I have commented my code, particularly in hard-to-understand areas
- New and existing unit tests pass locally with my changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary><h3>Greptile Summary</h3></summary>

The PR extracts reusable integration-export email delivery, customer-fault classification, and Prisma error helpers while preserving the existing blob-storage interfaces.

- Introduces a generic integration export-failure email template and sender with blob-storage-specific copy.
- Moves customer-fault classification into a neutral integrations module and retains blob-storage re-exports.
[...]

PR #16095, opened 2026-08-13
fix(worker): fail fast and auto-disable the PostHog export on config faults

niklassemmler:
🟡 **Live preview: [pr-16095.preview.langfuse.com](https://pr-16095.preview.langfuse.com)**

## What does this PR do?

> Now based directly on `main`. The persisted fault columns and the
> `posthog-export-failed` notification event this PR emits landed in #16135;
> the shared classifier, the Prisma predicate and the parameterised failure
> email it consumes landed earlier in #16094.

A bad or malicious PostHog hostname is a deterministic customer-config fault: it cannot succeed until the customer fixes it. The handler treated it as an infra failure — logged at `error`, rethrew, got retried five times by the queue, and tripped the processing-failures monitor on every attempt of every scheduled run. Forever, because nothing recorded the fault or stopped the schedule.

**Why fail-fast rather than retry-then-disable.** The failures metric is emitted per *failed attempt*, so any throw trips the monitor — even one that only disables on the final retry. Silencing it requires the job to complete without throwing. Since these faults cannot self-heal, retrying adds no signal, so the handler disables on the first classified occurrence and resolves the job.

On a classified fault the handler persists the error, atomically flips `enabled` true → false, records a disable metric tagged by reason, notifies the customer once, and returns. Anything else logs at `error` and rethrows, so infra and transient faults keep retrying and alerting exactly as before.

Notable details:

- **`dns-lookup-failed` is deliberately excluded** from the disable allowlist. Resolvability depends on runtime resolver state rather than configuration, so a DNS outage must not permanently disable a working integration. The allowlist is the transient filter, not the retry count.
- **The disable is scoped to the hostname the run validated**, so a job carrying stale config cannot disable a host the customer corrected mid-run.
- **The notification is awaited**, not fired and forgotten. The job resolves immediately afterwards and the scheduler only enqueues enabled integrations, so an interrupted dispatch would have no retry path and the customer would be silently disabled.
- **Recovery**: a successful run clears the persisted error.

### Second commit: closing the uncoded-rejection hole

Review flagged that a disallowed protocol or port would *still* retry forever, and it was right. `validateWebhookURL` threw a bare `new Error` for its protocol and port checks while every other outbound-URL rejection carried an `OutboundUrlValidationError` code — and the classifier matches on codes, never messages. So `ftp://` or `:8080` produced exactly the unfixable-by-retry failure this PR exists to stop, and classified as transient.

Both checks now throw coded errors. Protocol reuses the existing `protocol-not-allowed` code, which was already on the classifier's allowlist and already used by `validateBlobStorageEndpoint` for the identical check — `validateWebhookURL` was the only caller in the repo raising that condition uncoded. Ports get a new `port-not-allowed` code in the same deterministic bucket.

This is not a widening of what auto-disables: both conditions were always meant to be in the deterministic bucket, and a disallowed protocol or port is a property of the stored config rather than of runtime state. Error messages are byte-identical, so the callers that surface them and the tests that assert on them are unaffected — the only behavioural change is that these two rejections became classifiable. The new assertions check the *code* rather than the message, and fail against the previous bare-`Error` behaviour.

### Reading the handler diff

Hostname validation and sync-window computation move inside the existing `try` block so the single `catch` can classify the wrapped validation error off its `cause` chain. That reindents roughly 45 otherwise-unchanged lines — review with whitespace changes hidden.

Mixpanel is out of scope — it accepts no customer-supplied hostname, so this fault class doesn't apply. Aligning blob storage to the same fail-fast behaviour is a tracked follow-up; blob keeps retry-then-disable here.

## Type of change

## Mandatory Tasks

## Checklist

- New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/shared/src/server/outbound-url/validation.ts
+++ b/packages/shared/src/server/outbound-url/validation.ts
@@ -14,6 +14,7 @@ export type OutboundUrlValidationErrorCode =
   | "https-required"
   | "invalid-encoding"
   | "invalid-syntax"
+  | "port-not-allowed"
   | "protocol-not-allowed"
   | "url-credentials-not-allowed";
 
--- a/packages/shared/src/server/webhooks/validation.ts
+++ b/packages/shared/src/server/webhooks/validation.ts
@@ -1,5 +1,6 @@
 import { env } from "../../env";
 import {
+  OutboundUrlValidationError,
   type OutboundUrlValidationWhitelist,
   parseOutboundUrl,
   resolveHost,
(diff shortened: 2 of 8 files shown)


niklassemmler:
* **#16096** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16095** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16095?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16094** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16094?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #16097, opened 2026-08-13
fix(web): pin new Cloud analytics integrations to the enriched export source

niklassemmler:
🟡 **Live preview: [pr-16097.preview.langfuse.com](https://pr-16097.preview.langfuse.com)**

## What does this PR do?

Pins newly created Mixpanel and PostHog integrations on Langfuse Cloud to the
enriched events export source, and stops the export-source selector depending on
the V4 beta opt-in when running on Cloud.

Two problems existed:

1. **New integrations could still be created on the legacy export source.**
   Unlike blob storage, the analytics adapters never fed the integration row's
   `createdAt` into `ExportSourceContext`, so the integration-level ("exporter")
   cutoff never ran. A brand-new integration on an older Cloud project therefore
   defaulted to, and permitted, legacy `TRACES_OBSERVATIONS`.
2. **The selector was gated behind the V4 beta flag even on Cloud**, where
   dual-write has been live for a while and the enriched path is always
   available. That flag is a UI rollout gate and has nothing to do with data
   availability.

### Approach

The exporter cutoff is now **parameterized per integration family** rather than
hard-coded to blob storage's. `ExportSourceContext` gained an optional
`exporterCutoff`, defaulting to the existing blob constant so every blob call
site keeps its current behaviour unchanged; the analytics adapters pass their own
constant. `isLegacyBlobExporter` and the rejection message were both threaded
with the same value, so a rejection can never state a cutoff that was not the one
used for the decision.

On the server, both routers now pass the row's `createdAt` (and `null` for a
brand-new row, which follows new-customer rules) and default new Cloud rows to
the enriched source — the analytics equivalent of blob's `forceEventsOnCreate`.
On the client, both settings pages feed the same context in, which reuses the
existing hide-selector-and-pin path rather than introducing a second pinning
mechanism.

Grandfathered rows created before the cutoff keep their persisted source, keep a
visible selector, and are never silently rewritten.

### Behaviour worth a reviewer's attention

- **Create-pinning is deliberately date-independent.** A row with no prior
  existence is pinned on Cloud regardless of the wall-clock date; the cutoff
  governs *existing* rows only. A `now >= cutoff` reading would make the change
  inert until the cutoff date passes.
- **One self-hosted behaviour delta, contrary to a strict "unchanged" reading.**
  The upsert's CREATE payload now prefers the persisted source over the legacy
  new-row default. Under `LANGFUSE_MIGRATION_V4_WRITE_MODE=legacy` with a
  persisted enriched row, a save that omits the source and races a concurrent
  delete used to recreate the row on the legacy source; it now recreates it on
  the persisted enriched one. This is intentional — recreating a row with a value
  the caller never asked for silently rewrites their configuration — and it is
  only reachable under a concurrent delete. Called out explicitly rather than
  left implicit.
- **Grandfathered Cloud rows can now reach the selector without the beta flag.**
  For a row created before the cutoff, legacy sources remain selectable, so a
  Cloud project already on the enriched source can move back to a legacy one.
  The write was already permitted before this change (the cutoff was being
  skipped entirely); what changes is that the control is now visible. This is the
  intended grandfathering policy and is the one genuinely user-visible behaviour
  change here.
- The in-transaction race backstop now runs whenever the source is sent
  explicitly, not only when it is omitted. Because the settings form always
  submits the field, the previous condition meant the backstop never ran for any
  save made through the UI.

## Type of change

## Mandatory Tasks

## Checklist

- Blob storage behaviour is unchanged: every existing call site omits the new
  `exporterCutoff` and so keeps the blob cutoff. Verified by re-running the blob
  tRPC, public REST, and status suites (118 tests) plus the worker blob
  processing suite (33 tests).
- Self-hosted behaviour is unchanged in the `legacy`, `dual` + preview opt-in,
  and `events_only` write modes, with the single documented exception above.
  `(isLangfuseCloud || isBetaEnabled)` collapses to the previous expression when
  not on Cloud.
- Tests cover the new create-default and the write-time assert for new versus
  grandfathered Cloud rows, for both Mixpanel and PostHog; the raced-create
  backstop on both the omitted and explicit source paths; and the settings-page
  field-visibility and default derivations. Cutoff dates in tests are derived
  from the constants so the documented local-dev override cannot break them.
- Both settings pages were exercised in a browser against a backdated project
  with the beta flag off: a new integration shows no selector and persists the
  enriched source, and a grandfathered row shows the selector with legacy
  selectable.
- `pnpm run lint` and `pnpm run typecheck` both pass (7/7 tasks each).

This PR gives Blob Storage and PostHog/Mixpanel integrations separate exporter cutoff dates and centralizes analytics export-source validation across their routers and settings forms.

- Adds the analytics-specific cutoff environment override and shared policy parameter.
- Applies the analytics cutoff consistently to PostHog and Mixpanel creation/update paths.
- Expands policy, server, and client coverage for cutoff behavior and concurrent-delete handling.
</details>

The persisted-source rewrite should be fixed before merging because saving an existing post-cutoff analytics integration can silently change which data it exports.

The new integration-level cutoff reaches form initialization before the persisted value is considered, while the selector is hidden, so an unrelated settings save can persist EVENTS over an existing legacy source.

**Files Needing Attention:** web/src/features/analytics-integrations/exportSource.ts and the PostHog/Mixpanel settings pages
</details>

```mermaid
flowchart LR
  UI[PostHog or Mixpanel settings] --> Context[Build export-source context]
  Context --> Policy[Shared export-source policy]
  Policy -->|Analytics family| AnalyticsCutoff[Analytics exporter cutoff]
  Policy -->|Blob family| BlobCutoff[Blob exporter cutoff]
  Policy --> Validation[Router validation]
  Validation --> Upsert[Persist integration configuration]
  Upsert --> Worker[Worker exports selected streams]
```
</details>

`````markdown
### Issue 1
web/src/features/analytics-integrations/exportSource.ts:118
**Persisted export source gets replaced**

If an existing Cloud PostHog or Mixpanel row created on or after the analytics cutoff still stores a legacy export source, this branch initializes and resets the hidden field to `EVENTS` before consulting `persisted`. Saving any integration setting then silently changes which streams subsequent exports include, instead of preserving and surfacing the blocked persisted value.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test(web,shared): cover the client expor..."](https://github.com/langfuse/langfuse/commit/b19522ecb245b620f4137902225c0bae13f28999) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53036553)</sub>

**Context used:**

### DIFF
--- a/.env.dev.example
+++ b/.env.dev.example
@@ -70,6 +70,9 @@ NEXT_PUBLIC_LANGFUSE_CLOUD_REGION="DEV"
 # Same, for the integration-level cutoff applied to the blob storage
 # integration row's creation date instead of the project's.
 # NEXT_PUBLIC_LANGFUSE_BLOB_EXPORTER_CUTOFF=
+# Same, for the integration-level cutoff applied to the PostHog/Mixpanel
+# integration row's creation date.
+# NEXT_PUBLIC_LANGFUSE_ANALYTICS_EXPORTER_CUTOFF=
 
 # Langfuse experimental features
 LANGFUSE_ENABLE_EXPERIMENTAL_FEATURES="false"
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -37,6 +37,8 @@ const EnvSchema = z.object({
   NEXT_PUBLIC_LANGFUSE_BLOB_EXPORT_CUTOFF: z.iso.datetime().optional(),
   // Same, for the integration-level cutoff (BlobStorageIntegration.createdAt).
   NEXT_PUBLIC_LANGFUSE_BLOB_EXPORTER_CUTOFF: z.iso.datetime().optional(),
+  // Same, for the integration-level cutoff applied to PostHog/Mixpanel rows.
+  NEXT_PUBLIC_LANGFUSE_ANALYTICS_EXPORTER_CUTOFF: z.iso.datetime().optional(),
   NODE_ENV: z
     .enum(["development", "test", "production"])
     .default("development"),
--- a/packages/shared/src/features/analytics-integrations/export-source-policy.test.ts
+++ b/packages/shared/src/features/analytics-integrations/export-source-policy.test.ts
@@ -7,6 +7,7 @@ import {
   getAvailableExportSources,
   isEnrichedBlobExportSource,
   isLegacyBlobExportSource,
+  LEGACY_ANALYTICS_EXPORTER_CUTOFF,
   LEGACY_BLOB_EXPORT_CUTOFF,
   LEGACY_BLOB_EXPORTER_CUTOFF,
   validateExportSource,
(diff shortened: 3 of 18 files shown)


niklassemmler:
* **#16097** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16097?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16097?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #16124, opened 2026-08-14
feat(trace): pure layout core for the timeline

nkabardin:
🟡 **Live preview: [pr-16124.preview.langfuse.com](https://pr-16124.preview.langfuse.com)**

## TL;DR

The timeline's coordinates become **one pure function of the measured box**: `layout(tree, view, box, density)` in, absolutely-placed rows out. This is the foundation the next three PRs in the stack build on — #16131 rewires the production timeline onto it, #16133 adds the dense renderer, #16137 ships it behind a feature preview.

**Read the diff as production code.** Five of its six modules — `layout`, `viewTransform`, `timeCompression`, `textMeasurer`, `density` — move to `fns/timeline/` in #16131 and are what the production timeline positions every bar, label and tick through. Only `TimelineV2.tsx` is a throwaway Storybook harness, and it is the one file marked "not production code"; that marker does not cover its siblings. A defect in the core reaches users one PR later, so nits here are not spike nits.

What is true is that nothing *in this PR* has an app caller yet, so merging it alone cannot change what a user sees. Review it in Storybook: `pnpm --filter web run storybook` → **Playground → TimelineV2 → Findings**.

## Why

The scale basis was `SCALE_WIDTH = 900`, passed as a **default parameter** to four positioning helpers, with content width `Math.max(SCALE_WIDTH, maxEnd + 300)`. The measured box never reached the math, so at any lane narrower than 900px part of the trace was simply outside the viewport, and the only way to reach it was horizontal scroll — the thing that stops a trace being visible in one glance. Someone reported exactly that. There was also no zoom, and no degenerate-input contract (a collapsed trace divided by zero).

A scale basis that ignores the actual box cannot be right at more than one size, and the one size it was right at was 900px wide.

## What

- **`layout(tree, view, box, density) → PositionedNode[]`** — one pure function owns every coordinate. The box is **required**, with no default and no fallback constant; the function is total on degenerate input; and bars, labels, ticks and gap markers are all clamped inside the box, so no coordinate can ask for horizontal scroll.
- **A four-space view model** (trace / view / physical / container) as plain scalars: fit, clamp, zoom-about-anchor, pan, zoom-to-span. Default `view === trace space` is what makes "the whole trace at any size" true by construction rather than by luck.
- **Idle-gap time compression** and **canvas glyph text measurement**, both pure. The measurer is what lets the pure function decide whether a duration label fits inside its bar or must sit outside — a decision CSS flow currently makes, and makes wrong in a narrow lane.
- **Density as a resolved value**, from pointer modality and lines-per-row rather than a CSS constant.
- **A Storybook harness** — a throwaway renderer plus size / shape / compression / mobile stories, and play-function tests that assert the fits-the-box invariant in a real browser. The harness is how the model is reviewed; it is not shipped UI.

## Result

The model holds at every cell of the matrix, and three open questions now have evidence behind them:

- **React beats an imperative write path by ~3 orders of magnitude**, so this is a 146-line space model instead of a 2447-line imperative view manager.
- **Compression** should ship default-off and suggested in place, not as a user preference.
- **Mobile** wants a name-above-bar row below ~480px.

<details>
<summary>Measurements, review fixes, and the bugs this caught in itself</summary>

### Performance — the one expensive call

Measured on a 10k-span trace, 768×420 box, 486px lane:

| Measurement | Result |
| --- | --- |
| Mounted bars for 10,000 rows | **20** |
| `layout()`, 40-row window, per call | **0.0070 ms** (0.0046 ms compressed) |
| Frame time, 235 frames of continuous ctrl-wheel zoom | p50 **8.3 ms**, p95 9.3 ms, max 9.4 ms |
| Frames over 16.7 ms | **0** |
| Same run with no zoom driven (baseline) | p50 8.3 ms, max 9.4 ms |
| Zoom range actually swept | 0.02 → 486 px/ms |
| `prepareTimeline(10k)` — memoized on tree + collapse | 1.92 ms |
| `timeCompressionFor(10k)` — memoized on box width | 2.05 ms |

Frame time under continuous zoom is indistinguishable from idle. Measured against Storybook's dev build, so React is unminified and in development mode — production is faster, making these a pessimistic bound. If a future shape ever does drop frames, `layout()` is the seam an imperative write path plugs into without redesigning anything above it.

### Compression, and a fix to the reference implementation

Fitted honestly, a 36.02s trace whose real work is three 40ms spans gives 0.005 px/ms — three 4px slivers. Compressed: **0.372 px/ms, a 74× change.**

The reference implementation prices its duration-label buffer in *uncompressed* px/ms. In a 320px lane that reserves ~5.4 seconds around each span of a 36s trace, and compression becomes a no-op precisely where it is needed most. One fixed-point step — compress with no buffer, then re-compress with the buffer priced at the resulting scale — takes the long-tail case from ~15px of legible bar to ~41px.

Compression also needs **ticks at the collapsed-gap boundaries**: without them every nice-step tick falls inside a gap and the axis degrades to a lone `0ms`. With them it reads `0ms · 17.95s · 35.91s`, and the visible jump across a 28px band is what explains the band.

### Two bugs this caught in its own first render

Both are the exact failure class the architecture exists to remove:

1. **Deriving the lane width from a parent box is wrong twice** — by the frame border (2px, `box-sizing: border-box`) and again by a classic scrollbar (~15px, and 0 on macOS overlay scrollbars, so it passes locally and breaks elsewhere). The renderer measures the scroll viewport's own client box instead. #16131 measures the element that scrolls, not the one that contains it.
2. **A real click never selected a row.** Taking pointer capture on `pointerdown` retargets the derived `click` to the capture element, so it landed on the scroll container and no row handler ran. Capture is taken only after 4px of movement, which also makes "a drag does not select" fall out for free. `userEvent` dispatches synthetic events that ignore pointer capture and passed throughout, so the test asserts the handler's contract instead.

### Fixed after review

- **Drag-to-pan ran backwards.** The window start moved by `+delta`, so grabbing the surface and dragging right slid the bars left, away from the finger — the one thing a drag-to-pan must never do.
- **Lifting one finger of a pinch** left a live gesture whose start coordinates still pointed at the departed finger, so the survivor's next small move panned by the whole finger separation.
- **Double-clicking a collapse chevron** zoomed the view to that row: `stopPropagation` on click does not stop the native `dblclick` reaching the row.
- **A zoomed view outlived its coordinate space.** The window is in compressed ms, so switching trace or turning compression off left the numbers addressing somewhere else — and because the layout clamps them into the new space, the result was an unrelated window rather than an obviously broken one. It is dropped when the space is replaced.
- **The text measurer** only recognised a unit as the whole remainder of a string, so every label above a minute (`1m 35s`, `1h 30m`) measured its leading unit at the flat per-letter estimate.
- One `clamp` helper instead of the same three lines in two files.

Both of the last two were verified red before the fix.

### Other deviations from the reference

- **Clamping holds the zoom, not the pan.** At the right edge the reference holds `x` and shrinks the window; this holds the window and pulls `x` back — panning past the end should stop, not silently zoom out.
- **Plain scalar transform instead of `mat3`**: both spaces are one-dimensional in time, so matrix composition buys nothing.
- No `gl-matrix`, no scheduler event bus, no DOM-ref write path.

### Verification

- Client tests for the pure core, including a `Number.isFinite` sweep over all 10 trace shapes × 7 boxes × compression on and off.
- Storybook tests in real Chromium, including the fits-the-box invariant measured from `getBoundingClientRect`, drag direction, the pinch-release, and the coordinate-space reset.
- `Tasks: 7 successful, 7 total` — turbo lint. Typecheck clean. Reviewed by hand in both themes with real pointer input.

</details>

This PR adds a Storybook-only spike for a size-adaptive trace timeline with pure layout, zoom/pan, density resolution, text measurement, virtualization, and optional idle-time compression.
- Introduces coordinate transforms and a fit-to-box layout core with degenerate-input handling.
- Adds an interactive renderer and matrices covering trace shapes, widths, mobile compositions, and compression.
- Adds browser and client tests for layout bounds, scaling, virtualization, and view transforms.
</details>

The PR should not merge until the pinch-to-pan transition and stale-view behavior across coordinate-space changes are corrected.

The renderer can visibly jump when a pinch becomes a one-finger pan, and changing compression or trace inputs after zooming can reinterpret the persisted view as an unrelated time range.

**Files Needing Attention:** web/src/features/traces/components/TraceTimelineV2/TimelineV2.tsx
</details>

```mermaid
flowchart LR
  Tree[Trace tree] --> Prepare[prepareTimeline]
  Prepare --> Compression[timeCompressionFor]
  Box[Measured viewport] --> Compression
  Compression --> Space[Compressed trace space]
  View[Zoom and pan view] --> Layout[layout]
  Space --> Layout
  Box --> Layout
  Density[Resolved density] --> Layout
  Measure[Text measurer] --> Layout
  Layout --> Rows[Virtualized positioned rows]
  Layout --> Axis[Ticks and gap markers]
```
</details>

`````markdown
### Issue 1
web/src/features/traces/components/TraceTimelineV2/TimelineV2.tsx:332-336
**Pinch leaves stale pan state**

When one finger is lifted after a two-finger pinch, `startX` and `lastX` still refer to the previous gesture phase, so the remaining finger's next small movement pans by roughly the prior inter-finger distance and visibly jumps the timeline.

```suggestion
  const onPointerUp = (event: ReactPointerEvent<HTMLDivElement>) => {
    const state = gesture.current;
    state.pointers.delete(event.pointerId);
    state.pinchDistance = 0;
    state.dragging = false;

    const remainingX = state.pointers.values().next().value;
    if (remainingX !== undefined) {
      state.startX = remainingX;
      state.lastX = remainingX;
    }
  };
```

### Issue 2
web/src/features/traces/components/TraceTimelineV2/TimelineV2.tsx:150-153
**View outlives its coordinate space**

When `compress` or `roots` changes after zooming, the stored `view` remains expressed in the old compressed coordinate space while `compression` and `traceSpace` are replaced. `layout()` then interprets or clamps those stale coordinates in the new space, displaying an unrelated or empty time window.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): denser rows, decisive row fo..."](https://github.com/langfuse/langfuse/commit/9138b5f8609a0cbf0ff21cd4a046033812e1f731) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53288039)</sub>


nkabardin:
@claude review

PR #16131, opened 2026-08-14
feat(trace): fit the timeline to its measured lane

nkabardin:
🟡 **Live preview: [pr-16131.preview.langfuse.com](https://pr-16131.preview.langfuse.com)**

## TL;DR

The trace timeline stops being a fixed 900px canvas that the container scrolls, and becomes a renderer driven by its own **measured** lane. The whole trace now fits the width the user actually has, at any size, and the chart never scrolls horizontally.

**Stacked on #16124**, which is the pure core this builds on. Its diff is the model; this one is the production rewire. GitHub will retarget this to `main` when #16124 merges.

## Why

The scale basis was `SCALE_WIDTH = 900`, passed as a **default parameter** to four positioning helpers, with content width `Math.max(SCALE_WIDTH, maxEnd + 300)`. The measured box never reached the math, so in any lane narrower than 900px part of the trace was simply outside the viewport — and the only way to reach it was horizontal scroll, which is the thing that stops a trace being visible in one glance.

That is not a hypothetical narrow lane. On a 1440px screen with the detail panel open, the chart lane is **199px**; in the table's peek panel it is **137px**. A default parameter that ignores the box cannot be right at more than one size, and the one size it was right at was 900px.

## What

- **One pure function owns every coordinate.** Bars, duration labels, axis ticks and the playhead all come from `layout()` and are clamped inside the measured lane. `SCALE_WIDTH`, the four default-parameter helpers and the parallel metrics type are deleted.
- **The chart measures the element that scrolls.** Its border and its vertical scrollbar both take width an arithmetic guess would miss, and any such miss is a lane wider than the space available — which is the horizontal scroll being removed.
- **Which side of a bar its metric cluster sits on is a measured decision**, not a CSS-flow accident. Flow cannot see the lane and pushed the cluster off the edge in a narrow one.
- **Consequences of no horizontal scroll**, all simplifications: the scale strip needs no transform sync, the gutter reserves no scrollbar height, the playhead needs no follow-scroll, and selection reveal is vertical only.
- Trace shapes with no measurable duration (zero-duration, instantaneous, missing `endTime`) now produce finite coordinates instead of dividing by the scale span.

Kept working: the gutter/chart split and its resize, vertical virtualization, the playhead and seek, selection and scroll-into-view, and the playback row glow.

## Result

Verified in the browser against seeded traces — a 48-observation agent trace and a 1401-observation, 2h46m chain — in the trace panel and in the peek panel:

| Lane | `scrollWidth` vs `clientWidth` | Bars inside the lane | Rows mounted |
| --- | --- | --- | --- |
| 199px (trace panel) | 199 = 199 | 33/33, max right edge exactly 199 | 33 of 48 |
| 239px (after gutter resize) | 239 = 239 | rescaled, max right edge exactly 239 | 33 |
| 137px (peek panel) | 137 = 137 | 45/45 | 45 of 1401 |

This is what closes the community report about a trace not being visible in one glance.

<details>
<summary>Two defects browser verification caught, and the honest trade-off</summary>

### Both were about predicting a text width

1. **The text measurer assumed the wrong font.** It used `12px ui-sans-serif`; the app renders these labels at **11.2px** with a full system stack, and the canvas resolves that to a different face — 31.7px predicted vs 36.6px actual for `20.00s`, which is exactly enough to push an axis label past the lane edge. It now takes the computed font of a probe span carrying the label's own classes (36.6 vs 37.0 actual). The axis additionally clamps each label to the space remaining, because a prediction should not be the only thing keeping the axis inside the box.
2. **The metric cluster clipped mid-glyph** in a 137px lane: `layout()` measures the duration, but the subtree total and the cost ride along after it. The bar now fits its own optional metrics with the same measurer, highest priority first, and drops what does not fit instead of cutting a number in half — a duration beats a cost you cannot finish reading. Anything dropped stays reachable as the cluster's title.

Both were invisible to typecheck and to a glance; they were found by measuring `getBoundingClientRect` against the lane.

### The trade-off this makes, stated plainly

Fit-to-box is honest, and honesty costs bar width. In a 199px lane a 0.58s span of a 24.7s trace is 4.7px, so it renders at the 4px minimum marker — where the old 900px canvas gave it 21px and let you scroll to it. The trace is now visible in one glance; an individual short span is less legible than it was.

Two follow-ups already scoped for that, and they are why this PR stops here:

- **Zoom and pan as real view state**, so detail is reachable by narrowing the window instead of scrolling a canvas.
- **Idle-gap compression**, which on exactly this shape moves the scale from 0.005 to 0.372 px/ms — a 74× change — and is what makes fit-to-box legible rather than merely correct.

### Fixed after review

- **The metric cluster could still clip mid-content**, which is the one thing this PR exists to prevent. The comment icon and the score badges were each admitted against the *same* leftover budget, so a lane with room for either drew both and the overflow box cut the badges in half. They are charged to one running budget now, in the order they are drawn. Verified red before the fix: with 90px of room, the cluster rendered an icon, a duration and two score badges.
- **The layout measured one string and the bar drew another.** It picked the label's side by measuring `formatDurationMs` while the bar rendered `formatIntervalSeconds`, and the two disagree above an hour (`2h 46m` vs `2h 46m 17s`) — so the side chosen for the shorter string had no room for the longer one and the duration silently vanished into the tooltip. `layout()` takes the caller's formatter now, so it measures the exact string the bar draws.
- **An in-flight span showed no metrics at all.** It has no duration label, which the layout reports as placement `hidden`, and the bar read that as a verdict on the whole cluster — hiding its comment icon, cost and scores however much room the lane had.
- **The layout call is memoized.** It allocates a fresh positioned row per mounted row, and the row shells are `React.memo` precisely so a hover or a selection re-renders the two rows that changed; recomputing on every render handed every row a new object and re-rendered all of them.

### Verification

- `Tests 3277 passed | 51 skipped (3328)` — web client tests.
- `Tests 404 passed (404)` — Storybook tests in real Chromium, including the crowded-cluster and in-flight cases above.
- `Tasks: 7 successful, 7 total` — turbo lint. Typecheck clean.
- Playhead checked numerically: a click 60% across a 199px lane seeks to 14.82s of 24.70s, and playback advances the line and handle in step.

</details>

The PR rewires the production trace timeline to derive bars, ticks, labels, and playhead coordinates from the measured chart lane instead of a fixed-width canvas.

- Introduces the shared pure timeline layout pipeline and moves the spike utilities into the trace timeline function layer.
- Measures the chart viewport, removes horizontal scrolling, and fits virtualized rows and the playhead to that width.
- Adds finite handling for degenerate durations and measured placement for labels and metrics.
</details>

The PR should not merge until the metric cluster budgets comments and variable-width score badges together so enabled metadata is not silently clipped.

The measured layout itself is coherent, but the bar renderer independently admits optional elements against one unchanged budget and then hides overflow, causing reachable loss of score or comment content in constrained lanes.

**Files Needing Attention:** web/src/features/traces/components/TraceTimeline/TimelineBar.tsx
</details>

```mermaid
flowchart LR
  A[Trace tree and collapsed state] --> B[prepareTimeline]
  C[Measured chart client box] --> D[layout]
  E[Text measurer and density] --> D
  B --> D
  D --> F[Positioned virtual rows]
  D --> G[Axis ticks]
  D --> H[Playhead scale]
  F --> I[Gutter and timeline bars]
  G --> J[Measured lane]
  H --> J
  I --> J
```
</details>

`````markdown
### Issue 1
web/src/features/traces/components/TraceTimeline/TimelineBar.tsx:127-133
**Metric cluster overcommits width**

When a constrained row displays both comments and scores, both elements are admitted against the same unchanged `iconBudgetPx`, while score badges can render wider than the fixed 48px reservation. Their combined width and flex gaps exceed the cluster's `maxWidth`, causing `overflow-hidden` to clip comment or score content.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): measure labels in the font t..."](https://github.com/langfuse/langfuse/commit/b6e9aa96f38ff331a8104240979b2cf43eee0225) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53295394)</sub>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #16133, opened 2026-08-14
feat(trace): dense timeline renderer with a maps-style viewport

nkabardin:
🟡 **Live preview: [pr-16133.preview.langfuse.com](https://pr-16133.preview.langfuse.com)**

## TL;DR

Fit-to-box (#16131) fixed the time axis and left the other scrollbar alone. In a narrow, tall layout you still scrolled to see the trace, and at 26px a row the bars were slivers next to their own labels — a list of numbers, not a shape.

This is the renderer that inverts the priorities for that case: **nothing on identity, everything on shape**, then driven like a map. A 623-span trace becomes one continuous cascade you read at a glance, at **one device pixel per span**, with neither axis scrolling.

#16137 puts it in the trace panel behind a feature preview. Review it here in Storybook: `pnpm --filter web run storybook` → **Playground → TimelineDense**.

## What

- **Row height resolves DOWN** from the measured box and the row count, floored at **1px** — one device pixel per span. A 600px-tall box holds 600 rows, so most traces fit whole and the shape arrives all at once. A 4px floor still made a few thousand spans a scrolling job, which is the one thing this layout exists to avoid.
- **A 2D viewport, driven like a map.** Pinch and ⌘/ctrl + wheel zoom **both axes** about the cursor — the time window narrows and the rows grow together. Any plain wheel or two-finger scroll pans both axes; drag pans; **shift-drag draws a box** and flies to exactly that region; double-click flies to one element at a readable height. Zoom is exponential in a zoom level with per-frame accumulation, as mapping libraries do it.
- **A short trace stays readable.** The row window is allowed to over-hang the content, so three rows in a 600px box are three 26px rows and the rest stays empty — density never grows into a tall box's slack.
- **The left side follows one rule: if there is nothing to expand, it is not shown.** At bird's-eye density a row cannot hold a name and a colour rail beside colour-coded bars is repetition, so it takes no width at all. Once the rows can hold a name it appears if there is width to spare; where there is not, the colour rail is the affordance and hover or tap floats the names over the chart — as an overlay, so reading the names never moves the bars.
- **What a row shows follows from how tall it has become**: a type square around 10px, text around 20px. Zooming in grows the labels back rather than switching mode.
- **No text at hairline density.** Hover lights the row and a tooltip names it — type, name, duration, start offset — and the tooltip follows the content as you scroll under a stationary pointer.
- **Playback works here**: the line sweeps, the rows it is over glow, and the axis is the scrub track.

## Result

| Trace | Row height | Scrolls? |
| --- | --- | --- |
| 3 spans | 26px, labelled | no |
| 50 spans | 12px, compact | no |
| 600 spans | **1px, exactly at the floor** | no |
| 3000 spans | 1px | vertically; 600 rows in view, the rest panned to |

With type-coloured bars the model calls separate from the spans, so where the work happens is legible without reading anything.

<details>
<summary>Three things measurement changed, and the honest limits</summary>

### Integer row geometry is not a detail

Distributing the viewport across the rows gives fractional heights (604/150 = 4.027px). The browser rounds each row to device pixels; a hit-test computed from the unrounded model does not — so the highlighted row was the *neighbouring* row at roughly a third of cursor positions. Integral rows make it exact (**39/39 sampled positions**) and the bars crisp. The leftover few pixels stay empty at the bottom, which is the right trade.

### Zoom has to be told where the content is

The two axes are only ever *correlated*, never proportional: a chain of 600 spans can finish in the first third of a trace whose duration one long root sets. Zooming both about the same point therefore has a failure mode with nothing wrong with it and nothing in it — forty rows whose spans all lie outside the window, drawn as a column of edge carets, recoverable only by panning or by "Fit whole trace".

**This PR does not fix that; #16137 does**, with an `anchorTimeToRows` guard that slides the window onto the rows in view when a zoom lands on nothing at all. It is called out here because this is the PR that makes the failure reachable, and because the fix belongs to the diff that also has the app wiring to verify it against real traces.

### Wheel classification was a mistake

A wheel event with a large delta was read as a discrete mouse notch and zoomed. A fast two-finger flick on a Mac trackpad sends deltas far past any such threshold, so a quick scroll down zoomed *out* toward the fitted view — and swallowed the event either way, so there was no scrolling down at all. Every wheel pans now, whatever the device, with `deltaMode` normalized to pixels first; zoom keeps only the gestures that mean zoom.

### The lens needs a different substrate

The original idea was a fisheye: rows near the pointer expand, borrowing space from the rest so the total never changes. The maths hold — heights are weight-normalised, so the sum is exactly the height given, at any magnification, which is what keeps it scroll-free — and it is tested. But the hit-test cannot be made right on this substrate:

- Hit-test in **distorted** space and the lens feeds on its own output: the row moves out from under the cursor and a drag turns jumpy.
- Hit-test in **resting** space and the row that expands is reliably *not* the one under the cursor.

Correct fisheye hit-testing needs the transform to be invertible — natural in a scene graph, awkward across 600 absolutely positioned rows. It survives in `fns/timeline/focusLens.ts`, tested and deliberately unwired, as the starting point for its own spike (canvas or WebGL).

### Checks

- Client tests for the viewport and the pure core, including the no-scroll invariant under magnification and a finite sweep over degenerate input.
- Storybook tests in real Chromium: neither axis scrolls, scroll pans while pinch zooms (including a fast flick and line-mode deltas), hover peeks the names without moving a bar, hover opens a tooltip that follows the content, double-click focuses both axes, shift-drag zooms to its box, a zoom lands on content, playback sweeps and glows, and the type square stays inside its rail.
- `Tasks: 7 successful, 7 total` — turbo lint. Typecheck clean.

</details>

This PR adds an experimental high-density trace timeline renderer that shrinks rows to a four-pixel floor and prioritizes trace shape over labels.
- Adds dense row fitting, type-color rails, hover details, selection, zoom, and pan interactions.
- Adds optional focus-lens and time-compression experiments.
- Adds pure vertical-fit and lens utilities with client tests.
- Adds Storybook scenarios and browser assertions for overflow, tooltip, and no-scroll behavior.
</details>

The PR appears safe to merge with no concrete blocking or independently actionable non-blocking issue identified.

The dense renderer keeps its node and row indexes aligned on its non-virtualized layout path, clamps rendered time views to trace space, and explicitly confines the acknowledged inexact lens interaction to an experimental story.
</details>

```mermaid
flowchart TD
  Roots[Trace roots] --> Prepare[Prepare timeline rows]
  Box[Measured box] --> Fit[Resolve integral row height]
  Prepare --> Fit
  Prepare --> Layout[Compute horizontal time layout]
  Box --> Layout
  Fit --> Lens{Focus lens enabled?}
  Lens -->|No| Uniform[Uniform row geometry]
  Lens -->|Yes| Distorted[Weight-normalized row geometry]
  Layout --> Render[Render rail, bars, grid, and gaps]
  Uniform --> Render
  Distorted --> Render
  Pointer[Pointer and wheel gestures] --> View[Hover, select, pan, or zoom]
  View --> Layout
  View --> Lens
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(trace): extreme-density timeline sp..."](https://github.com/langfuse/langfuse/commit/9491d2fac262053d985a2e087d5dedcb4a60f618) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53301765)</sub>

### DIFF
--- a/web/src/features/traces/components/TraceTimelineDense/TimelineDense.stories.tsx
+++ b/web/src/features/traces/components/TraceTimelineDense/TimelineDense.stories.tsx
@@ -14,10 +14,11 @@ const meta = preview.meta({
 
 /**
  * A phone-shaped box: narrow and tall, which is the case this spike is for.
- * 640px leaves a 604px row viewport — exactly enough for 150 hairline rows, so
- * the "at the floor" story really is at the floor and not one row over it.
+ * 658px leaves exactly 600px of surface once the frame, toolbar, axis and
+ * readout are taken — room for 150 rows at the 4px floor, so the "at the floor"
+ * story really is at the floor.
  */
-const PHONE = { width: 360, height: 640 };
+const PHONE = { width: 360, height: 658 };
 /** The peek panel: narrow and short. */
 const PEEK = { width: 320, height: 300 };
 
@@ -26,7 +27,6 @@ export const FiftySpansOnAPhone = meta.story({
     roots: manySpans(50),
     box: PHONE,
     showNames: false,
-    lens: false,
     barColor: "neutral",
     compress: false,
     showReadout: true,
@@ -38,20 +38,18 @@ export const OneHundredFiftySpansExactlyAtTheFloor = meta.story({
     roots: manySpans(150),
     box: PHONE,
     showNames: false,
-    lens: false,
     barColor: "neutral",
     compress: false,
     showReadout: true,
   },
 });
 
-/** Past the floor: no-scroll breaks and the readout says so. */
-export const FiveHundredSpansOverflowing = meta.story({
+/** More rows than 4px each can show: the extra are panned to, maps-style. */
+export const FiveHundredSpansPannable = meta.story({
   args: {
     roots: manySpans(500),
     box: PHONE,
     showNames: false,
-    lens: false,
     barColor: "neutral",
     compress: false,
     showReadout: true,
@@ -63,7 +61,6 @@ export const TypeColouredBars = meta.story({
     roots: manySpans(150),
     box: PHONE,
     showNames: false,
-    lens: false,
     barColor: "type",
     compress: false,
     showReadout: true,
@@ -76,7 +73,6 @@ export const ShortTraceStaysReadable = meta.story({
     roots: threeSpans(),
     box: PHONE,
     showNames: false,
-    lens: false,
     barColor: "neutral",
     compress: false,
     showReadout: true,
@@ -88,7 +84,6 @@ export const ReporterShapeInAPeek = meta.story({
     roots: reporterTrace(),
     box: PEEK,
     showNames: false,
-    lens: false,
     barColor: "neutral",
     compress: false,
     showReadout: true,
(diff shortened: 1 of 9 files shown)


nkabardin:
@claude review

PR #16137, opened 2026-08-14
feat(trace): compact timeline behind a feature preview

nkabardin:
🟡 **Live preview: [pr-16137.preview.langfuse.com](https://pr-16137.preview.langfuse.com)**

## TL;DR

The dense timeline ships, behind the **Compact Timeline** feature preview — on by default for the Langfuse and ClickHouse team, opt-in for everyone else.

There is still one Timeline in the trace panel. What it *is* depends on the preview: the whole trace at once, every row down to a single device pixel, coloured by observation type — then zoomed and panned like a map, with a box you can drag to zoom into and a double-click that flies to one element at a readable height.

**Stacked on #16133.** GitHub retargets this to `main` when that merges.

## Why a preview and not a third mode

A third segment asks every user to understand and choose between two timelines. A preview asks nothing of anyone: the team gets it by default and can opt out per-preview, everyone else opts in, and GA is one line in `available-flags.ts` when it has earned it. The rollback is the same line.

`receivesFeaturePreviewsByDefault` already gives `@langfuse.com` / `@clickhouse.com` accounts previews by default, so the flag only had to join `featurePreviewFlags` — no new default-on machinery.

## What

- **The renderer stays presentational.** It takes a measured box, and now a playhead and an active-id set, and nothing implicit — so the entire Storybook size × shape matrix keeps reviewing it, and a story can drive a fake playhead by hand. A container measures its own box and wires the app around it.
- **Selection is controlled by `SelectionContext`**, not internal state, and selecting is one shared hook across all three views: record it, point the app at it, open the detail panel, and on mobile switch to the tab that shows it. Each view had grown its own copy.
- **Selection arriving from outside is revealed.** The tree, a search hit, a deep link and playback all select rows, and the layout does not even emit a node for a row outside the window — so the highlight was invisible exactly when it mattered. An external change pans both axes just far enough, never zooming, because "look at this one" is not a request to change how far in you are looking.
- **Playback works here.** The transport in the header drives the shared engine; this view now draws the two things you watch — a line that sweeps and the rows lighting up as it passes them — plus press-and-drag on the axis to place the playhead. The line moves imperatively off the position feed and the glow comes from the active-id set, so playback costs one render per boundary crossing rather than one per frame.
- **The view switcher stays Tree / Timeline.** The preview swaps which timeline renders, with `enableForAdmins: false` so a self-hosted admin does not silently get it.
- **The left side follows one rule: if there is nothing to expand, it is not shown.** At bird's-eye density a row cannot hold a name, and a colour rail beside colour-coded bars is repetition — so it takes no width and the whole box is timeline. Once rows can hold a name it simply appears if there is width to spare; where there is not, the rail becomes the affordance and a hover or tap floats the names over the chart. The toggle hides when it could do nothing.
- **A `timeline-shapes` seeder scenario** — twelve small hand-timed traces, one per timeline morphology (retry backoff with widening gaps, a wait on a human, an 8-way fan-out, one slow tool, in-flight spans, zero-duration checkpoints, a ten-level ladder, flat siblings, a three-turn agent loop), because what differs between real traces is their *time*, and that does not come out of scaling one shape up. Plus `agent-timeline --turn-gap-ms`, which makes a multi-hour run whose work is a few percent of its wall clock.

## Verified in the app

Against seeded traces in the trace panel, in both themes:

| Check | Result |
| --- | --- |
| A 623-span trace at rest | 623 rows at **1px**, bars from x=0, nothing scrolls |
| Distinct observation-type colours | 7 |
| Scrollable overflow, both axes | **0 / 0** |
| Click a row | `observation=…` in the URL, detail panel opens on it |
| Rail at 1px rows | **0px**, toggle hidden, no invisible DOM per row |
| Rail at 40px rows in a 398px panel | 10px affordance, toggle shown |
| Play | line sweeps, 24–32 rows glow, stop clears both |
| Axis click at 60% | playhead lands at 60% of the lane |
| Scroll under a stationary pointer | tooltip renames to the row now under it |

<details>
<summary>Four defects only the real app exposed, and the resolved open question</summary>

### Everything that predicted a size instead of measuring one

1. **The tooltip put scroll on a box that must not scroll.** It was positioned from a percentage guess at which side had room, and could poke past the surface — 5px of `scrollWidth` on a surface whose entire claim is that nothing scrolls. Clamped to the space actually left: **0px** across hover positions from 15% to 97%.
2. **A duration label drawn inside its bar was unreadable.** It has to contrast with the *bar*, not the page, and it cannot pick one colour that does — the type palette runs from pastels to 600s, so light text dies on the pastels and dark text on the 600s. It brings its own ground instead.
3. **The closed rail's type square was clipped by its own overflow box** at depth ≥ 3, so the one type indicator a closed rail has left showed as a sliver exactly on the deeply-nested spans. The rail width is computed from the deepest indent plus the square rather than guessed.
4. **A collapsed idle gap's tooltip was dead** — the rows are painted over that layer and cover its full width, so no pointer could ever reach it. Wide enough bands say how much time they swallowed in the band itself.

That makes five in this workstream, counting the text measurer's font and the frame border plus scrollbar from earlier PRs. The pattern is worth naming: in this renderer, anything that decides a coordinate must measure, not estimate.

### Two more from the same review

- **A double-click selected three times** — two clicks plus the focus handler — and selecting is not free: it captures an analytics event and reopens the detail panel.
- **Hover walked the tree per hovered row** instead of using the map the tree already builds, which matters when hover changes once per pixel.

### The open question, answered

Zooming the time axis while rows stay in tree order could leave visible rows empty, because the two axes are not both spatial — one is time, the other is tree order. Rather than filtering rows to the window (denser, but the vertical axis would jump as you zoom), the window now slides onto the rows in view when a zoom lands on nothing at all: it never resizes, and it does nothing whenever there is something to see, so it cannot move the ground under someone reading a span. Measured: four zoom steps into the middle of a 3-hour trace went from **0 bars visible to 40**.

### Checks

- `Tests 3277 passed | 51 skipped (3328)` — web client tests, including the timeline pure core (viewport zoom/pan/box-zoom/focus/reveal/clamp/interpolate, compression, layout, degenerate sweeps) and the flag defaults for team members, other users, and a per-preview opt-out.
- `Tests 404 passed (404)` — Storybook tests in real Chromium.
- `Tasks: 7 successful, 7 total` — turbo lint. Typecheck clean on web and shared.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #16148, opened 2026-08-14
fix(agent): keep old sandbox conversations writable

LysanderKie:
🟡 **Live preview: [pr-16148.preview.langfuse.com](https://pr-16148.preview.langfuse.com)**

## Summary

- Remove the age-based sandbox write lock so older sandbox conversations stay writable, including approval continuations and later user turns.
- When a Lambda MicroVM session is missing, already terminated, or lost in a resume race, replace it with a fresh sandbox instead of failing the run.
- Tell the agent when that happened. A replaced workspace loses every file the agent authored, so the run now carries a run-scoped instruction saying what is gone and what survived.
- Cap workspace lifetime at four hours, on both the existence and suspended-duration limits so the two agree.

## Why the reset has to be announced

Dropping the write lock is the right trade: losing a scratch workspace beats being unable to continue a conversation at all. But it turns a clear user-facing rejection into a silent partial loss.

`/workspace/tool_calls/` survives a replacement, because it is rebuilt from the event log and re-pushed on every sandbox request rather than living on the VM. What does **not** survive is anything the agent wrote itself: `read`/`write`/`edit`/`bash` are excluded from that archive by name, and nothing else captures workspace files. Installed packages and process state go too.

So the model resumes with its own earlier "I wrote this file" turns still in history, reads the path, gets an ENOENT, and improvises. The user sees a confused agent rather than an explained reset.

The notice is delivered as the run's own system message (Mastra's per-execution `system` option), not compiled into the managed prompt, since it describes one turn's environment rather than the agent's standing behaviour. It cannot be a transcript message: `@ag-ui/mastra` forwards only assistant, user and tool messages, so a `developer` message there would never reach the model.

Detecting it needs an explicit probe. `ensureSession` does discover a lost session, but only on the first sandbox tool call, which is after the model already has its input. `SandboxProvider` therefore gains an optional `probeSession` that reports *why* a persisted session is unusable without creating anything. The probe path reuses the existing `session_replaced` log and metric, so that telemetry stays continuous rather than going quiet for the most common case.

## On the four-hour lifetime

How long a workspace stays resumable is `min(maximumDurationInSeconds from creation, suspendedDurationSeconds from last activity)`, so the two limits have to agree or the larger one is unreachable config. On `main` the existence cap was one hour, which made the eight-hour suspended cap dead and workspace persistence largely notional.

Eight hours only mattered while the cap had to match the "8-hour limit" message this PR deletes. Four hours covers a working session and most approval turnarounds, at four times the idle tail rather than eight.

Two things worth measuring separately, neither blocking: whether a suspended MicroVM bills for its snapshot, and whether "exist" counts suspended time (the SDK wording implies wall-clock from creation).

## Also in here

- Forward the specific reason when a restored session's bridge never becomes ready, instead of labelling every such replacement a resume race.
- Remove `disablePendingToolApprovalActions`. This branch removed its only caller and the only story exercising a non-default value, leaving it permanently `false`.
- Correct the sandbox README. It claimed a data-retention cleanup tears down MicroVMs when a conversation is deleted; no such cleanup exists. Deleting a conversation clears `providerSessionId` *without* terminating, so that workspace lives until the AWS idle policy reclaims it.

## Known gaps, deliberately not in scope

- **Terminating a MicroVM on conversation delete.** The README now states the real behaviour instead of implying a guarantee we do not provide.
- **A durable statement across later turns.** The instruction fires on the turn where the reset is detected. On the next turn the probe finds the new VM healthy, so nothing fires, while history still shows the agent's earlier writes. A standing line in the managed prompt is the cheap fix and needs no deploy.
- **Timeouts on the sandbox `fetch` calls.** Every new replacement branch runs through `waitForBridge`, whose health probe is a bare `fetch` with no timeout. If it hangs instead of failing, none of the new branches is reached. Landing separately; this PR is not made worse by it, just less effective than it looks.

## Impacted packages

- `web`
- `worker`
- `@langfuse/shared`

## Verification

- `pnpm --filter web run test in-app-agent` — `Tests 138 passed (138)`
- `pnpm --filter worker run test executeInAppAgentRun` — `Tests 19 passed (19)`
- `pnpm --filter web run test-client InAppAgent` — `Tests 43 passed (43)`

Preview signoff still outstanding: the environment was torn down after two days idle and needs the `preview` label re-added. To exercise the reset without waiting out a real expiry, point a conversation's `providerSessionId` at a non-existent id and send another turn.


worker/src/features/in-app-agent/executeInAppAgentRun.test.ts
bezbac, line 1: I think it might make sense to assert that we insert a notice about the missing files from the sandbox in this test.

web/src/__tests__/server/unit/in-app-agent-lambda-microvm-provider.servertest.ts
bezbac, line 1: Isn't this effectively the same as the test added to `worker/src/features/in-app-agent/executeInAppAgentRun.test.ts`?

web/src/__tests__/server/in-app-agent-background-run.servertest.ts
bezbac, line 1: I also think these tests don't really add much :)

packages/shared/src/in-app-agent/server/sandbox/types.ts
```diff
@@ -7,6 +7,15 @@ export type InAppAgentSandboxProviderType =
   | "dangerous-docker"
   | "lambda-microvm";
 
+export const IN_APP_AGENT_SANDBOX_SESSION_REPLACEMENT_REASONS = [
+  "not_found",
+  "terminal_state",
+  "resume_race",
+] as const;
```
bezbac: This is exporting a non-type variable from a types.ts file. Might be a bit weird. What do you think about creating a `constants.ts` file instead?

packages/shared/src/in-app-agent/server/sandbox/service.ts
```diff
@@ -1,8 +1,16 @@
...
+  InAppAgentSandboxProviderType,
+  SandboxFile,
+  SandboxProvider,
+} from "./types";
+import { logger, recordIncrement } from "../../../server";
 
 export async function createInAppAgentSandbox(params: {
   conversationId: string;
   projectId: string;
+  runId?: string;
```
bezbac: I think it's a little awkward to pass the `runId` and `providerType` just for logging. I don't think the `createInAppAgentSandbox` should be aware of the `providerType` as it's effectively a method on a provider instance itself.

worker/src/features/in-app-agent/executeInAppAgentRun.test.ts
LysanderKie: Agreed, and done in 7a46b972 — though not in this test.

The notice is now a run-scoped instruction: `probeSession` reports up front that the persisted session is gone, `createInAppAgentSandbox` returns `workspaceWasReset`, and the worker passes it as `sandboxWorkspaceWasReset` so the run carries an extra system message saying what is gone and what survived.

It deliberately cannot be a transcript message. `agent.ts:874-878` spells out that `@ag-ui/mastra` forwards only assistant, user and tool messages, so a `developer` message would be dropped before the model ever sees it. I had it that way first and the test passed while the model saw nothing, which is exactly why the assertion sits at the model-facing boundary instead: `in-app-agent-stream.servertest.ts` → "adds a run instruction when the persisted workspace is gone", which runs `createAgUiStream` for real and reads the instruction off the Agent config. I checked it fails if the flag is not threaded through.

Asserting it here as well would only re-check a one-line option pass-through that TypeScript already covers.

web/src/__tests__/server/unit/in-app-agent-lambda-microvm-provider.servertest.ts
LysanderKie: They cover different layers, so no.

The provider test drives the AWS state machine: given a saved session reporting `TERMINATED`, no `ResumeMicrovm` is issued, a fresh `RunMicrovm` is, and the replacement is labelled `terminal_state`. Nothing else touches that.

The worker test asserts a run whose conversation is older than the old lock window and has a sandbox tool call in its history executes to `SUCCEEDED`.

That said, you have a point about the worker one specifically, just aimed slightly differently — see my reply on the background-run tests.

web/src/__tests__/server/in-app-agent-background-run.servertest.ts
LysanderKie: I would push back on these two and cut a different one instead.

"continues an old sandbox conversation after an approval decision" and "starts another user turn in an old sandbox conversation" are the inverse of the `rejects approval continuations for write-locked conversations` test this PR deletes, at the two places that actually threw the 412: `startBackgroundRun` and `decideBackgroundApproval`. They are the only coverage that the user-facing rejection is gone from both entry paths, and they would catch a reintroduced age lock. For the headline behaviour change of the PR that seems worth two cheap tests.

The weaker one is `executes an old conversation that already used sandbox tools` in `executeInAppAgentRun.test.ts`. The worker guard was defence-in-depth behind those two web checks, and with the lock deleted outright that test now mostly asserts "a run succeeds", which plenty of tests in that file already do. Happy to drop it if you want the count down — just say and I will.

packages/shared/src/in-app-agent/server/sandbox/types.ts
LysanderKie: Already gone as of 7a46b972, so no new file needed.

`IN_APP_AGENT_SANDBOX_SESSION_REPLACEMENT_REASONS` only ever existed to derive the union type and was exported with no consumer anywhere. ESLint flagged it as type-only once I stopped exporting it, which made the point: it is now just

```ts
export type InAppAgentSandboxSessionReplacementReason =
  | "not_found"
  | "terminal_state"
  | "resume_race";
```

`types.ts` is back to types only.

packages/shared/src/in-app-agent/server/sandbox/service.ts
LysanderKie: Fair on `providerType`, fixed in 989918ed. `SandboxProvider` now carries a `type`, so `createInAppAgentSandbox` reads it off the instance instead of being told, and the `?? "unknown"` fallback is gone. Cost was adding `type` to three provider test doubles, which seems a reasonable trade for the contract being self-describing.

On `runId` I would keep it. It is log context of exactly the same kind as the `conversationId` and `projectId` this function already took before this PR, and the service is genuinely operating within one run — the alternative is returning the reason and having the worker log it, which does not work for the lazy path where `ensureSession` replaces a session mid-tool-call and no caller is watching.

web/src/features/in-app-agent/components/ControlledInAppAgentWindow.tsx
```diff
@@ -197,17 +186,14 @@ export function ControlledInAppAgentWindow(
...
       isRunUnsettled={isRunUnsettled}
       isAwaitingApproval={isAwaitingApproval}
       isHeaderDragHandleEnabled={props.isHeaderDragHandleEnabled}
       isExpanded={props.isExpanded}
       isConversationInteractionDisabled={isConversationInteractionDisabled}
       isSelectedConversationHydrating={isSelectedConversationHydrating}
-      disablePendingToolApprovalActions={
-        selectedConversationIsWriteLocked || error?.type === "write_lock"
-      }
       messages={drawerMessages}
```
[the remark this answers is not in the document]
LysanderKie: Done in 7a46b972: the prop, its default, and the branch are gone, and `isDisabled` is now just `isRateLimited`.

One correction for the record, since it is the same file: greptile flagged the new provider test's dynamic `await import(...)` as an import-placement issue. That is wrong — it is the existing pattern in that file at lines 53, 113 and 172, so the new test matches its siblings.

packages/shared/src/in-app-agent/server/sandbox/service.ts
LysanderKie: Just saw that Claude got a bit too commited in directly answering your comment/feedback, will look into it and see that we can remove it 😄

web/src/__tests__/server/in-app-agent-background-run.servertest.ts
LysanderKie: Just saw that Claude got a bit "too commited" in directly answering your comment/feedback, will look into it and review that if we can get rid of them 😄

web/src/__tests__/server/unit/in-app-agent-lambda-microvm-provider.servertest.ts
LysanderKie: Just saw that Claude got a bit "too commited" in directly answering your comment/feedback, will look into it and review that if we can get rid of them 😄

packages/shared/src/in-app-agent/server/sandbox/service.ts
LysanderKie: Both gone now — you were right on `runId` too, I was defending it unnecessarily.

- `providerType` (989918ed): `SandboxProvider` carries a `type`, so the service reads it off the instance and the `?? "unknown"` fallback disappeared.
- `runId` (06591fcf): removed. The replacement log still carries `projectId`, `conversationId`, the provider type and the reason, which is enough to find the event.

LysanderKie: Follow-up: the `type` field I added to `SandboxProvider` is gone too (a8041be7).

Your argument applied to it as well — it only ever fed the same log line and metric tag, and in production `lambda-microvm` is the only provider that ships, so the tag was constant. The metric keeps `reason`, which is the part that actually distinguishes the paths.

Net effect on `createInAppAgentSandbox`: `providerType`, `runId` and the provider `type` field are all removed, and the contract no longer carries anything that exists purely for logging.

web/src/__tests__/server/in-app-agent-background-run.servertest.ts
LysanderKie: You were right and my earlier pushback was wrong — both are gone in 6474cdcd, along with the equivalent worker test.

I checked it rather than argued it. Each test backdated the conversation nine hours and inserted a sandbox `TOOL_CALL_START`, which only mattered while the write lock keyed on exactly that pair. Nothing reads `conversation.createdAt` for gating any more, so I stripped the setup and re-ran: all three still passed. At that point they are subsets of tests already in the same files — `commits a queued run with its user message and enqueues it by run id` and `decides an approval exactly once and reads the tool args server-side` here, and the first test in the worker suite.

I argued they were the only coverage that the 412 is gone from both entry paths. That was the wrong frame: the 412 is gone because the check is deleted, and you cannot meaningfully pin absent code this way — the old test did that in reverse, which is why deleting it was right.

web/src/__tests__/server/unit/in-app-agent-lambda-microvm-provider.servertest.ts
LysanderKie: Kept it as It's the only coverage of the replacement state machine: given AWS reports `TERMINATED`, no `ResumeMicrovmCommand` is issued, a fresh `RunMicrovm` is, and the reason is labelled.

The caveat: now that `probeSession` catches an expired session before the agent runs, the `not_found`/`terminal_state` branches inside `ensureSession` no longer fire on the common turn-start path. They only fire when a VM dies mid-turn, between the probe and a later tool call.

The three tests you flagged in the other threads are gone.

worker/src/features/in-app-agent/executeInAppAgentRun.test.ts
LysanderKie: Settled: the notice exists, and the test you were commenting on is gone in 6474cdcd for unrelated reasons (its setup no longer reached any code path).

The assertion lives in `in-app-agent-stream.servertest.ts` → "adds a run instruction when the persisted workspace is gone", which runs `createAgUiStream` for real and reads the instruction off the Agent config. That is the boundary where the model-facing behaviour actually happens; I verified it fails if the flag is not threaded through.

PR #16149, opened 2026-08-14
feat(web): offer export sources by deployment capability, not per-user beta

niklassemmler:
🟡 **Live preview: [pr-16149.preview.langfuse.com](https://pr-16149.preview.langfuse.com)**

## What does this PR do?

Stacked on the write-mode derivation PR below it. That one made availability a single deployment-level rule; this one removes the last input that contradicts it.

Selector visibility and the create default on PostHog and Mixpanel were gated on the per-user V4 beta flag. Consequences:

- Two members of the same project saw different options for one shared config row.
- Both pages diverged from blob storage, which was never gated this way.
- The flag is not an independent switch — `v4BetaEnabled` is itself derived from the write mode, so it was a second, less accurate copy of the rule this stack exists to unify.

All three pages now go through one `getExportSourceFieldState`, which decides purely from the deployment's write mode and the project's Cloud cutoff.

## ⚠️ This is Cloud-visible — please push back if you disagree

The gate is dropped **everywhere, including Cloud**. There is deliberately no `isCloud` escape hatch.

For a pre-cutoff Cloud project on `dual`, with a user who does not have the V4 beta enabled, on the PostHog or Mixpanel settings page:

| | Before | After |
|---|---|---|
| Export-source selector | not rendered | rendered, three selectable options |
| Create default | legacy traces + observations | events source |

The argument for accepting it: a pre-cutoff Cloud project on `dual` genuinely can export either source, so offering the choice is truthful rather than merely permissive. Blob storage already behaved this way, so this makes the three integrations consistent.

Post-cutoff Cloud projects are unaffected — the field stays hidden and the events source stays pinned.

The alternative is keeping the gate on the Cloud branch only. That preserves a second definition of availability in exactly the code path this stack unifies, and keeps the two-users-one-row inconsistency. Worth stating explicitly so it can be chosen deliberately.

## Tests

Adds the composition tests the extracted helpers never had. Visibility and the create default were only ever asserted apart, and it is their *agreement* that matters: a hidden selector whose default fails validation blocks every save with no field left to correct it. That combination is now asserted directly, including the invariant that a hidden selector always leaves a selectable default.

## Type of change

Non-breaking, but user-visible on Cloud as described above.

## Mandatory Tasks

## Verification

- `pnpm turbo run typecheck lint` — 9 successful, 9 total
- `pnpm --filter web run test-client` — 3280 passed, 51 skipped
- `pnpm --filter @langfuse/shared run test` — 913 passed, 2 skipped

The PR replaces the per-user beta gate for PostHog and Mixpanel export-source selection with deployment-capability-based behavior.

- Centralizes selector visibility, available options, and form defaults in `getExportSourceFieldState`.
- Applies the shared field state to both integration settings pages.
- Adds tests covering write modes, Cloud cutoffs, hidden-field defaults, and persisted unavailable sources.
</details>

The PR appears safe to merge with no concrete changed-code failure identified.

The centralized field state remains consistent with the client and server export-source validation contracts across supported write modes and Cloud cutoff behavior.
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(web): offer export sources by deplo..."](https://github.com/langfuse/langfuse/commit/cc62d719c4bf21f228a41039a94e63797c40cf54) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53361242)</sub>

**Context used:**

### DIFF
--- a/web/src/features/analytics-integrations/exportSource.clienttest.ts
+++ b/web/src/features/analytics-integrations/exportSource.clienttest.ts
@@ -403,10 +403,8 @@ describe("write mode drives the settings-page selector", () => {
   );
 });
 
-// The three settings pages consume only this composite, so the pieces agreeing
-// individually is not enough: visibility and the create default have to agree
-// with each other. A form mounted on a hidden selector whose default fails
-// validation blocks every save with no field to correct.
+// The pages consume only this composite, so the pieces being correct
+// individually is not enough — visibility and the default must agree.
 describe("getExportSourceFieldState", () => {
   const ctxFor = (
     writeMode: BlobExportWriteMode,
@@ -416,8 +414,6 @@ describe("getExportSourceFieldState", () => {
   ): ExportSourceContext =>
     buildExportSourceContext({ writeMode, isCloud: false, ...over });
 
-  // Whenever the selector is hidden the default must be saveable on its own,
-  // because there is no field left for the user to fix it with.
   it.each<BlobExportWriteMode>(["legacy", "dual", "events_only"])(
     "%s: a hidden selector always leaves a selectable default",
     (mode) => {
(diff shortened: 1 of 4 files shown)


niklassemmler:
* **#16149** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16149?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16149?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16093** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

PR #16168, opened 2026-08-14
fix(web): keep audio media in the strip when a message falls back to the JSON view

DavidTraina:
## What does this PR do?

Fixes a pre-existing bug in the trace IO preview's media-strip dedup: a message with `audio` whose content is not markdown-renderable (e.g. an over-limit string) falls back to a JSON view that renders neither the content's media nor the audio inline — yet `ChatMessageList` still marked the audio's media id as rendered, filtering it out of the shared media strip. The audio became unreachable while the markdown toggle was on (toggling markdown off recovered it). The dedup now marks media as rendered only when `canRenderContentAsMarkdown` says the markdown path actually renders the message, for content media and audio alike.

Behavior is identical on `main` today (the old `typeof content === "string"` short-circuit marked audio the same way), surfaced while reviewing #16166.

**Stacked on #16166** — this branch includes its commits because both change the same dedup block in `ChatMessageList.tsx`. Draft until #16166 merges; I'll rebase and mark ready then.

## Type of change

## Mandatory Tasks


DavidTraina:
Superseded: the fix and its regression test (`keeps referenced media and audio when over-limit content falls back to JSON`) landed on `main` as part of #16166 (bc4f4d80a). Rebasing onto `main` leaves no delta.

PR #16179, opened 2026-08-16
fix(web): include observation IO and metadata in legacy trace JSON download

### LINKED ISSUE #15337: bug: legacy trace JSON download omits observation input, output, and metadata
### Describe the bug

On self-hosted Langfuse v3.206.0 in legacy mode, observation input,
output, and metadata are displayed correctly in the observation detail UI.

However, using "Download trace as JSON" exports every observation with
`input: null`, `output: null`, and `metadata: "{}"`.

The data has been stored successfully because the observation detail UI can
retrieve and display it. The trace download is therefore incomplete.

Expected behavior: the downloaded JSON contains the same observation input,
output, and metadata shown in the UI.

Actual behavior: those fields are omitted or replaced with empty values.

### Steps to reproduce

1. Create an observation with non-empty input, output, and metadata. For
   example, use metadata `{"export_marker": "present"}`.
2. Open the trace in Langfuse.
3. Select the observation and verify that its input, output, and metadata
   appear in the observation detail panel.
4. Select "Download trace as JSON".
5. Inspect the observation in both `.observations[]` and
   `.trace.observations[]`.

The downloaded observation contains values resembling:

```json
{
  "input": null,
  "output": null,
  "metadata": "{}"
}
```

Self-hosted

3.206.0

- Langfuse Python SDK: 4.5.1
- Claude Agent SDK integration with manually recorded observation metadata

This does not appear to be an SDK ingestion issue because the stored metadata
is visible in the observation detail UI.

### Additional information

This appears to affect the legacy trace-download path rather than ingestion.

In v3.206.0, the trace router loads observations with `includeIO: false`:

https://github.com/langfuse/langfuse/blob/v3.206.0/web/src/server/api/routers/traces.ts#L385-L442

The legacy downloader then serializes that lightweight observation array
directly:

https://github.com/langfuse/langfuse/blob/v3.206.0/web/src/components/trace/lib/download-trace.ts#L72-L84

The observation detail UI uses a separate lookup that retrieves the complete
observation:

https://github.com/langfuse/langfuse/blob/v3.206.0/web/src/server/api/routers/observations.ts#L21-L45

The newer V4 download endpoint added in #13033 includes `input`, `output`,
and `metadata`, but that endpoint is gated behind V4/Fast Preview. Existing
self-hosted legacy deployments therefore remain affected.

Related PR:
https://github.com/langfuse/langfuse/pull/13033

I am happy to contribute a fix after confirming the preferred approach for
the legacy storage path.

Yes

zzyIyzz:
## What

Fixes the legacy (non-V4-beta) trace JSON download omitting observation input, output, and metadata — closes langfuse/langfuse#15337.

## Why

The legacy "Download trace as JSON" path serialized the observations already loaded for the page. Those observations are fetched with `includeIO: false` to keep page rendering light, so every downloaded observation lost its `input`, `output`, and `metadata` — even though the observation detail panel displays them (it fetches them separately).

This affects self-hosted deployments still on the legacy trace UI, where the downloaded JSON is silently incomplete.

## How

* Added a `traces.observationsWithInputOutput` tRPC procedure that returns the full observation set for a trace via `getObservationsForTrace({ includeIO: true })`, reusing the repository's existing payload-size guard.
* The legacy download now fetches this full set at download time, keeping the existing `{ trace, observations }` export shape.
* Traces at or above the omit-large-fields threshold keep the light set and surface the same warning the V4 download route shows, so both paths agree on large-trace behavior.

## Verification

* `tsc --noEmit` — clean.
* New regression test `traces.observationsWithInputOutput` asserts input/output/metadata round-trip and that access is scoped to the caller's project.
* Reverse verification: flipping `includeIO` back to `false` makes the new test fail with `expected null to deeply equal { question: ... }`, reproducing the exact missing-IO behavior from the issue.

This PR adds an authenticated tRPC query that retrieves full observation IO for legacy trace downloads and updates the legacy UI to fetch it on demand while retaining large-trace omission behavior.

* Adds project-scoped regression coverage for observation input, output, and metadata retrieval.
* Fetches full observations when downloading legacy traces below the large-field threshold.
* Adds a dedicated full-observation procedure backed by the existing ClickHouse repository query.

</details>

<details><summary>Confidence Score: 4/5</summary>

The PR should not merge until legacy downloads preserve observation metadata as structured JSON rather than an encoded string.

The new full-observation path fixes omitted IO, but its domain conversion changes metadata into a string and the pass-through download serializer writes that incompatible representation into every affected legacy export.

**Files Needing Attention:** web/src/server/api/routers/traces.ts, web/src/features/traces/components/TracePanelNavigationHeader/TracePanelNavigationHeader.tsx, web/src/**tests**/server/traces-trpc.servertest.ts

</details>

<details><summary>Sequence Diagram</summary>

sequenceDiagram participant U as Legacy trace UI participant T as traces.observationsWithInputOutput participant CH as ClickHouse repository participant D as JSON download helper U->>U: Check observation count alt Fewer than 350 observations U->>T: Fetch trace observations T->>CH: getObservationsForTrace(includeIO: true) CH-->>T: Full observations T-->>U: Observations with stringified metadata else 350 or more observations U->>U: Retain light observations and warn end U->>D: Serialize trace and observations D-->>U: Download JSON file

</details>

```markdown
### Issue 1
web/src/server/api/routers/traces.ts:484
**Metadata is double-encoded in exports**

When a legacy-UI user downloads a trace containing metadata and fewer than 350 observations, `toDomainWithStringifiedMetadata` converts each metadata object to a string before the pass-through download serializer writes it, causing parsed exports to contain escaped JSON strings instead of metadata objects.

---

```

</details>

Reviews (1): Last reviewed commit: ["fix(web): include observation IO and met..."](<https://github.com/langfuse/langfuse/commit/e52e267f4f36b3eb6f57bd50d6c9fe1f7de11f3d>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=53761192>)

**Context used:**

* Knowledge Base — [Web App (`web/`)](<https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/web-app.md>)

### DIFF
--- a/web/src/__tests__/server/traces-trpc.servertest.ts
+++ b/web/src/__tests__/server/traces-trpc.servertest.ts
@@ -692,6 +692,77 @@ describe("traces trpc", () => {
     });
   });
 
+  describe("traces.observationsWithInputOutput", () => {
+    // Regression for https://github.com/langfuse/langfuse/issues/15337: the
+    // legacy trace JSON download serialized the page's observations, which are
+    // fetched with includeIO: false — so every downloaded observation lost its
+    // input, output, and metadata even though the detail panel shows them.
+    it("returns observation input, output, and metadata for the legacy trace download", async () => {
+      const trace = createTrace({
+        project_id: projectId,
+      });
+      const observation = createObservation({
+        project_id: projectId,
+        trace_id: trace.id,
+        start_time: new Date(trace.timestamp).getTime(),
+        input: JSON.stringify({ question: "What is Langfuse?" }),
+        output: JSON.stringify({
+          answer: "An open-source LLM engineering platform",
+        }),
+        metadata: { export_marker: "present" },
+      });
+
+      await createTracesCh([trace]);
+      await createObservationsCh([observation]);
+
+      // ClickHouse insert visibility can lag.
+      let observationsRes: Array<{
+        id: string;
+        input: unknown;
+        output: unknown;
+        metadata: unknown;
+      }> = [];
+      await waitForExpect(async () => {
+        observationsRes = await caller.traces.observationsWithInputOutput({
+          projectId,
+          traceId: trace.id,
+          timestamp: new Date(trace.timestamp),
+        });
+        expect(observationsRes.some((o) => o.id === observation.id)).toBe(true);
+      });
+
+      const downloaded = observationsRes.find((o) => o.id === observation.id);
+      expect(downloaded?.input).toEqual({ question: "What is Langfuse?" });
+      expect(downloaded?.output).toEqual({
+        answer: "An open-source LLM engineering platform",
+      });
+      expect(downloaded?.metadata).toEqual(
+        JSON.stringify({ export_marker: "present" }),
+      );
+    });
+
+    it("rejects access to a trace of another project", async () => {
+      const differentProjectId = randomUUID();
+      const trace = createTrace({
+        project_id: differentProjectId,
+      });
+      const observation = createObservation({
+        project_id: differentProjectId,
+        trace_id: trace.id,
+      });
+
+      await createTracesCh([trace]);
+      await createObservationsCh([observation]);
+
+      await expect(
+        caller.traces.observationsWithInputOutput({
+          projectId,
+          traceId: trace.id,
+        }),
+      ).rejects.toMatchObject({ code: "NOT_FOUND" });
+    });
+  });
+
   describe("traces.filterOptions", () => {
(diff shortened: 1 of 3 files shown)


maxdeichmann:
@zzyIyzz we do not maintain the old legacy surfaces anymore.

zzyIyzz:
Understood, thanks for the quick look! Acknowledged — I won't pursue changes to the legacy surfaces.

PR #16200, opened 2026-08-17
chore(shared): log the node-level cause behind Redis connection errors

niklassemmler:
🟡 **Live preview: [pr-16200.preview.langfuse.com](https://pr-16200.preview.langfuse.com)**

## What does this PR do?

When ioredis cannot rebuild its cluster slot map it throws
`ClusterAllFailedError: Failed to refresh slots cache.` — the same message for
every underlying cause. The actionable detail lives on the error's
`lastNodeError` (the failure from the final node ioredis attempted), and the
default error serialisation drops it. A recent production burst produced ~9,200
log lines on a single service in ~3h, every one of them carrying an identical
message plus an ioredis-internal stack and nothing about the node failure.
Root-causing it meant inferring the failure mode from a stack frame and
correlating against ElastiCache metrics by hand.

This PR makes the log line answer the question directly:

- Extracts `lastNodeError` and logs its `message`, `name`, `code`, `errno`,
  `syscall`, `address` and `port`.
- Classifies the failure into an explicit `failureMode` — `timeout`,
  `connection-refused`, `connection-reset`, `connection-closed`,
  `host-unreachable`, `dns`, `tls`, `auth`, `cluster-down`, `readonly`,
  `unknown` — derived from the node-level cause rather than the wrapper. A
  slot-refresh timeout, a refused connection and a `CLUSTERDOWN` reply are three
  different incidents and now read as three different lines.
- Recovers the failing node's `host:port` from the matching `node error` event,
  which is the only place ioredis reports it. The address is only attached when
  the event's error is reference-identical to `lastNodeError`, so it can never
  be mismatched to a different node.
- Puts `failureMode` and the node cause into the message string too, so
  text-format logs (`LANGFUSE_LOG_FORMAT=text`) stay diagnosable, not just JSON.
- Applies the same treatment to the sentinel and single-node handlers.

Example of the emitted JSON, verified against the real winston config:

```json
{
  "level": "error",
  "message": "Redis cluster error [timeout]: Failed to refresh slots cache. (last node error: timeout from 10.0.1.5:6379)",
  "stack": "ClusterAllFailedError: Failed to refresh slots cache.\n    at …",
  "failureMode": "timeout",
  "error": { "name": "ClusterAllFailedError", "message": "Failed to refresh slots cache." },
  "lastNodeError": { "name": "Error", "message": "timeout", "node": "10.0.1.5:6379" }
}
```

`stack` stays at the **top level** rather than nested under `error`. Two
reasons: the text log format renders `info.stack` and nothing else, so nesting
it would silently drop Redis stacks from `LANGFUSE_LOG_FORMAT=text`
deployments; and the JSON payload already exposed a top-level `stack`. The new
payload is therefore a strict superset of the old one — `message`, `stack` and
`lastNodeError` keep their positions, and `lastNodeError` is now populated
instead of serialising as `{}` (its content is what the old logs dropped).

No stack is emitted for `lastNodeError`: it always points into ioredis
internals, and these lines arrive in the thousands during a burst. Absent
fields are omitted rather than logged as `null`, so line size is unchanged for
errors that carry no extra detail.

Behaviour is otherwise unchanged — this only affects what the existing
`error`-event handlers log.

## Type of change

- [x] Chore (tooling, dependencies, CI, workflows, repo upkeep, or other maintenance work)

## Mandatory Tasks

## Checklist

- [x] Tests cover the classification of each failure mode behind the identical
      wrapper message, the recovered node address, non-`Error` values, and the
      formatted message string (`packages/shared/src/server/redis/redis.test.ts`).
- [x] `pnpm turbo run lint --filter=@langfuse/shared --filter=web --filter=worker` — 6 successful, 6 total
- [x] `pnpm turbo run typecheck --filter=@langfuse/shared --filter=web --filter=worker` — 6 successful, 6 total
- [x] `pnpm --filter @langfuse/shared run test redis.test.ts` — 19 passed (19)

The PR enriches Redis connection-error logs with structured failure classifications and the node-level cause wrapped by ioredis.
- Adds helpers to extract, classify, structure, and format Redis errors.
- Applies the enriched logging to Cluster, Sentinel, and standalone Redis clients.
- Adds unit coverage for common Redis failure modes and formatted messages.
</details>

The PR appears safe to merge, with a non-blocking regression that removes Redis stack traces from text-format logs.

The new diagnostics preserve useful error messages and structured JSON context, but nesting the stack under `error.stack` prevents the existing text formatter from rendering it.

**Files Needing Attention:** packages/shared/src/server/redis/redis.ts
</details>

`````markdown
### Issue 1
packages/shared/src/server/redis/redis.ts:18
**Preserve stacks in text logs**

When `LANGFUSE_LOG_FORMAT=text`, the new plain context object stores the stack under `error.stack`, but the text formatter only renders top-level `info.stack`. Redis error stacks therefore disappear from text logs, reducing the diagnostic value of these messages.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore(shared): log the node-level cause ..."](https://github.com/langfuse/langfuse/commit/02383c60149b6ebc548421ed2c6c2cd876917952) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53958397)</sub>

**Context used:**

### DIFF
--- a/packages/shared/src/server/redis/redis.test.ts
+++ b/packages/shared/src/server/redis/redis.test.ts
@@ -11,6 +11,10 @@ vi.mock("../../env", () => ({
 
 import { env } from "../../env";
 import { safeMultiGet, scanKeys } from "./redis";
+import {
+  buildRedisErrorContext,
+  formatRedisErrorMessage,
+} from "./redisErrorContext";
 
 type ScanCall = [string, "MATCH", string, "COUNT", number];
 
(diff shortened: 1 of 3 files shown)


niklassemmler:
* **#16200** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #16202, opened 2026-08-17
fix(worker): pin analytics export egress to connect-time SSRF validation (v3 backport of #16091)

Steffen911:
v3 backport of #16091.

## Problem

The PostHog and Mixpanel analytics exporters validate the configured host once, up front, and then send over an unpinned `fetch`. A host that answers with a public IP during that pre-check can rebind to a private or loopback address before the socket actually connects (TOCTOU / DNS rebinding), so the export payload is delivered to an internal target.

On `v3` this is reachable on any deployment: the PostHog integration host is a normal project setting with no Cloud-only gating, `handlePostHogIntegrationProjectJob.ts` validates `posthogHostName` once via `validateWebhookURL`, and the SDK transport it injects then calls `globalThis.fetch` directly.

The same upstream commit also stops the rejected-redirect URL — which may carry userinfo in a `Location` header — from reaching the worker log line and the error message BullMQ persists as the job's `failedReason`.

## What this changes

- Both senders now go through `fetchWithSecureRedirects`, which re-resolves and re-validates every destination IP at connect time and re-validates redirect targets. That helper already existed on this branch (`packages/shared/src/server/outbound-url/fetch.ts`), used by `secureLlmFetch`, SSO config validation and the dataset router — no new shared infrastructure.
- A connect-time block is a permanent misconfiguration, so it surfaces as an `UnrecoverableError` and BullMQ skips the job's remaining attempts. A transient DNS-resolution failure stays retryable.
- Credentials are stripped from the thrown error's message, from the retained `cause`, and from that cause's stack; the cause is attached non-enumerably so a structured logger does not walk it.
- `DNS_LOOKUP_FAILED_MESSAGE_PREFIX` is exported from the shared validator so the retryable/terminal classifier can recognise a DNS failure by message text after `RedirectValidationError` re-wraps it. The thrown string is unchanged.

## Why this is an adapted backport, not a cherry-pick

Upstream, this fix sits on top of #16095, a fail-fast/auto-disable refactor of the PostHog job's error handling that was **not** backported. The upstream diff edits a `reason`-classification / `dispatchProjectNotification` block that does not exist on this branch, so a raw `git cherry-pick` conflicts or drops the security-relevant lines.

Ported here:

- `worker/src/errors/findOutboundUrlValidationError.ts` and `worker/src/features/analyticsIntegrationEgress.ts` are **byte-identical to upstream** — both turned out to be self-contained, and every symbol they need (`whitelistFromEnv`, `RedirectOptions`, `validateWebhookURL`, `UnrecoverableError`) already exists here with the same shape.
- The two call sites are adapted. The PostHog handler gets the terminal-classification call in its plain `catch`, without upstream's `reason === undefined` guard (there is no classifier here). The Mixpanel client has no timeout/abort plumbing on this branch, so only the fetch swap and the `catch` hook are ported.
- `LANGFUSE_MIXPANEL_TIMEOUT_MS` does not exist here, so the test's env mock omits it.

## Test adaptation, and a behaviour difference worth knowing

Upstream's suite drives the PostHog leg through the job handler. That does not work here: on this branch's `posthog-node` (`^5.32.0`, resolving to 5.38.4), `flush()` does **not** drain the queue — the send happens on the SDK's own background timer, roughly 10s later, after the handler has already returned. Measured directly with an injected fetch spy:

```
after capture(): 0 calls
after flush():   0 calls
after ~10s:      0 calls
after ~15s:      1 call   → http://localhost:PORT/batch/
after shutdown(): 1 call
```

So a handler-level assertion would pass whether or not the egress is pinned, and the handler's `on("error")`/`sendError` path cannot observe the send outcome. The PostHog test therefore targets `countingFetch`, the transport the handler injects — the seam every PostHog export byte passes through, and where the guarantee actually lives. The Mixpanel leg is a direct `fetch`, so it is still driven end-to-end through `client.flush()` and does assert the terminal `UnrecoverableError`.

Two consequences of that SDK behaviour, both **pre-existing on this branch and deliberately not addressed here**: PostHog export failures surface asynchronously after the job has already succeeded and advanced `lastSyncAt`, and the terminal-classification call in that handler's `catch` is therefore best-effort for the pinning case. It is kept for parity with upstream and still fires on synchronous failures.

## Verification

Run against this branch:

- `pnpm --filter worker run test analyticsIntegrationSsrfPinning` — `Tests 7 passed (7)`
- Neighbouring suites green: `posthogIntegrationProjectJob` `6 passed (6)`, `mixpanelIntegrationProjectJob` `8 passed (8)`, `posthogExportVolume` `1 passed (1)`, `mixpanelExportVolume` `1 passed (1)`, `outbound-connection-validation` `5 passed (5)`, `secureLlmFetch` `11 passed (11)`, `webhook-validation` `22 passed (22)`.
- `webhook-redirect` reports `7 failed | 4 passed (11)`, which is **identical on clean `v3`** (checked out detached at `f6c77b7084` and re-run) — pre-existing, environment-dependent, unrelated to this change.

### Revert-check

Production files restored to their pre-fix `v3` state with the new tests left in place, to confirm the tests catch a live vulnerability rather than a hypothetical one. Both sender legs fail, each on the request counter:

```
× blocks a validated host that resolves to a blocked IP before egress
× rejects a validated host that connects to a blocked IP, before egress and terminally
AssertionError: expected 2 to be 1
```

The count going from 1 to 2 is the export payload being delivered to the loopback address that the string pre-check had just accepted — the SSRF completing on unpatched `v3`. Each test also carries an in-test negative control (a bare `fetch` to the same URL, asserted to reach the server) so a count of 1 in the patched run cannot be an unreachable harness.

Working tree confirmed clean (`git status --porcelain` empty) after restoring.

This backport routes PostHog and Mixpanel analytics exports through connect-time destination validation, revalidates redirects, and safely classifies permanent SSRF blocks without persisting credential-bearing errors.

- Exports the shared DNS-failure message prefix for retry classification.
- Adds secure analytics redirect options and sanitized terminal-error handling.
- Replaces both analytics senders' plain fetch paths with secure, redirect-aware egress.
- Adds regression coverage for DNS rebinding, terminal classification, and credential redaction.
</details>

The PR appears safe to merge; the only identified issue is non-blocking import organization in the new test module.

The analytics senders now use the existing connect-time validation and redirect-validation path, and the terminal-error sanitizer preserves retry semantics without exposing redirect credentials; only the test file's import placement needs cleanup.

**Files Needing Attention:** worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts
</details>

```mermaid
sequenceDiagram
  participant Exporter as Analytics exporter
  participant SecureFetch as fetchWithSecureRedirects
  participant DNS as DNS resolver
  participant Validator as Outbound URL validator
  participant Vendor as PostHog / Mixpanel
  Exporter->>SecureFetch: Send analytics batch
  SecureFetch->>DNS: Resolve destination at connect time
  DNS-->>SecureFetch: Candidate addresses
  SecureFetch->>Validator: Validate every resolved address
  alt Address is allowed
    SecureFetch->>Vendor: Connect and send request
    Vendor-->>SecureFetch: Response or redirect
    opt Redirect
      SecureFetch->>Validator: Validate redirect target
      SecureFetch->>DNS: Resolve redirect at connect time
    end
    SecureFetch-->>Exporter: Final response
  else Address or redirect is blocked
    Validator-->>SecureFetch: Validation error
    SecureFetch-->>Exporter: Rejected request
    Exporter->>Exporter: Redact credentials and throw UnrecoverableError
  end
```
</details>

`````markdown
### Issue 1
worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts:77-84
**Keep static imports together**

These static imports appear after executable mock setup rather than at the top of the module, departing from the repository's required import organization and making the test's dependency setup harder to follow consistently.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(worker): pin analytics export egress..."](https://github.com/langfuse/langfuse/commit/e48f917fa217ceb20f855c864c186d52e4ba32b0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53968173)</sub>

**Learned From**
</details>

### DIFF
--- a/packages/shared/src/server/outbound-url/validation.ts
+++ b/packages/shared/src/server/outbound-url/validation.ts
@@ -27,6 +27,11 @@ export class OutboundUrlValidationError extends Error {
   }
 }
 
+// Prefix of the message thrown on DNS-resolution failure. Exported so consumers
+// can detect this failure by message when the typed `code` is lost — e.g. after
+// RedirectValidationError re-wraps the inner error by message text only.
+export const DNS_LOOKUP_FAILED_MESSAGE_PREFIX = "DNS lookup failed for";
+
 export interface OutboundUrlValidationWhitelist {
   hosts: string[];
   ips: string[];
@@ -77,7 +82,7 @@ export async function resolveHost(hostname: string): Promise<string[]> {
   if (!ips.size) {
     throw new OutboundUrlValidationError(
       "dns-lookup-failed",
-      `DNS lookup failed for ${hostname}`,
+      `${DNS_LOOKUP_FAILED_MESSAGE_PREFIX} ${hostname}`,
     );
   }
   return [...ips];
--- a/worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts
+++ b/worker/src/__tests__/analyticsIntegrationSsrfPinning.test.ts
@@ -1,0 +1,316 @@
+/**
+ * Connect-time SSRF pinning for the PostHog and Mixpanel analytics exporters.
+ *
+ * Both sender tests neutralise the string pre-check (that models "the host
+ * validated as public") and point the send at a `localhost` NAME, whose
+ * connect-time lookup re-resolves to 127.0.0.1 (that models "it rebinds at
+ * connect"). Pointing them at a statically-blocked host instead would pass even
+ * without connect-time pinning, which is why they are shaped this way.
+ *
+ * The PostHog leg is exercised through `countingFetch`, the transport the
+ * handler injects into the SDK, rather than through the handler itself: on this
+ * branch's posthog-node, `flush()` does not drain the queue — the send happens
+ * on the SDK's own background timer, after the handler has already returned — so
+ * the handler cannot observe the send outcome and a handler-level assertion
+ * would pass whether or not the egress is pinned. `countingFetch` is the seam
+ * every PostHog export byte goes through, so it is where the guarantee lives.
+ */
+import { createServer, type Server } from "node:http";
+import { AddressInfo } from "node:net";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { isUnrecoverableError } from "../errors/UnrecoverableError";
+import { findOutboundUrlValidationError } from "../errors/findOutboundUrlValidationError";
+
+// The senders read their allowlist from the environment, and vitest loads
+// ../.env. A developer who allowlists localhost for local webhook testing would
+// otherwise turn the block assertions below into mystery failures.
+vi.hoisted(() => {
+  delete process.env.LANGFUSE_WEBHOOK_WHITELISTED_HOST;
+  delete process.env.LANGFUSE_WEBHOOK_WHITELISTED_IPS;
+  delete process.env.LANGFUSE_WEBHOOK_WHITELISTED_IP_SEGMENTS;
+});
+
+const openServers: Server[] = [];
+
+async function startLoopbackServer(
+  onRequest: () => void,
+): Promise<{ nameUrl: string; port: number }> {
+  const server = createServer((_req, res) => {
+    onRequest();
+    res.end("{}");
+  });
+  openServers.push(server);
+  await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+  const port = (server.address() as AddressInfo).port;
+  return { nameUrl: `http://localhost:${port}`, port };
+}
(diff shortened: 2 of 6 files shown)


Steffen911:
No straightforward backport from v4 and therefore risky to apply. Will accept this for the time being with no intention to patch.

PR #16209, opened 2026-08-17
fix(api): scope the v3 sunset notices to Langfuse Cloud

niklassemmler:
🟡 **Live preview: [pr-16209.preview.langfuse.com](https://pr-16209.preview.langfuse.com)**

## What does this PR do?

The `November 16, 2026` sunset date on the legacy v3 endpoints is a Langfuse Cloud lifecycle commitment, but every deprecation message in the Fern definitions stated it unconditionally. A self-hosted operator reading the API reference was told their endpoints disappear on a fixed date, which is not true — self-hosted deployments choose when they move to v4.

This rewrites all 15 endpoint sunset messages to name Cloud as the scope and to say what actually happens self-hosted:

> **Before:** `Langfuse v3 is deprecated; this endpoint will be removed on November 16, 2026. Use GET /api/public/v3/scores instead.`
>
> **After:** `On Langfuse Cloud, Langfuse v3 is deprecated and this endpoint will be removed on November 16, 2026. Use GET /api/public/v3/scores instead. Self-hosted deployments are unaffected by this date; the endpoint becomes unavailable when they upgrade to Langfuse v4.`

The wording follows the pattern already established for the blob storage `exportSource` gates, which were the only correctly-scoped deprecation notices in the API surface.

Affected definitions: `sessions.yml` (2), `datasets.yml` (3), `trace.yml` (2), `scores.yml` (2), `dataset-run-items.yml` (2), `legacy/observations-v1.yml` (2), `ingestion.yml` (1), `legacy/metrics-v1.yml` (1). The generated `openapi.yml` mirror is restamped via `scripts/openapi/sync-deprecations.ts`, so only operation descriptions change there — no structural edits to the spec.

Deliberately **not** changed: field-level deprecations (`Model.inputPrice`/`prices`, `Observation.usage`, `calculated*Cost`, `parseIoAsJson`, SCIM `password`). Those are deployment-agnostic API-shape deprecations ("use `pricingTiers` instead"), not lifecycle sunsets, and are correct as unqualified statements.

A new case in `openapi-deprecations.servertest.ts` asserts every deprecated operation carries the Cloud scope, so an unqualified sunset message cannot land again. Verified it fails when the scope is removed from a message.

[...]

PR #16210, opened 2026-08-17
fix(api): serve the v3 deprecation signal on Cloud only

niklassemmler:
🟡 **Live preview: [pr-16210.preview.langfuse.com](https://pr-16210.preview.langfuse.com)**

## What does this PR do?

The `_deprecation` object on legacy v3 endpoint responses carries `sunsetAt: "2026-11-16"`, a Langfuse Cloud commitment. It was gated on `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN`:

```ts
const deprecation =
  env.LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN === "true"
    ? routeConfig.deprecation
    : undefined;
```

That flag has defaulted to `"true"` since the v4 env defaults were flipped, so every self-hosted deployment received the removal date unless the operator explicitly set the flag to `"false"`. The gate's own comment stated the intent — "self-hosted deployments still on v3 (opt-in unset/false) should not see them yet" — but `unset` resolves to `"true"` through the zod default, so the gate admitted exactly the deployments it meant to exclude. This is a correctness fix, not a tightening.

It now gates on `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION`, matching `handleBlobStorageIntegrationProjectJob.ts`, which resolves the same policy the same way ("Legacy-source deprecation is a Cloud policy … so the deprecation notice below is Cloud-only too"). The flag's default is deliberate and untouched here — this change routes around it rather than fighting it.

The integration suite keeps asserting the Cloud path (`.env.dev.example` sets the region to `DEV`). Its `maybeDescribe` guard is gone: half of it was dead — no endpoint under test reads `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN` (those routes gate on `LANGFUSE_MIGRATION_V4_WRITE_MODE` via `rejectInEventsOnlyMode`), and nothing in `.env*.example` or CI ever sets that flag to `false`. The guard only existed because `_deprecation` injection used to be gated on that flag, which is precisely what this PR changes. A plain `describe` now fails loudly instead of skipping silently.

Note the self-hosted case has no automated coverage: the suite drives real HTTP against a server on port 3000, so the gate runs in the server process and cannot be steered from the test process.

This also aligns the runtime notice text with the Cloud-scoped Fern/OpenAPI wording from #16209, now merged, so the response body and the published API reference state the same scope.

## Type of change

## Mandatory Tasks

## Checklist

- [x] `pnpm turbo run lint typecheck --filter=web` — `Tasks: 5 successful, 5 total`
- [x] `openapi-deprecations.servertest.ts`, `deprecation-signal.servertest.ts` — `Tests 24 passed (24)`, integration suite run against a local server with Postgres/ClickHouse

### DIFF
--- a/web/src/__tests__/server/deprecation-signal.servertest.ts
+++ b/web/src/__tests__/server/deprecation-signal.servertest.ts
@@ -22,21 +22,12 @@ import {
   V3_SUNSET_DATE,
 } from "@/src/features/public-api/server/deprecations";
 import { OBSERVATIONS_API_V2_DOCS_URL } from "@/src/features/public-api/server/rateLimitUpgradePaths";
-import { env } from "@/src/env.mjs";
 import { randomUUID } from "crypto";
 
-// `_deprecation` injection is gated on LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN,
-// which also enables the events read path (requires ClickHouse >= 25.12). The
-// -azure / -redis-cluster CI modes intentionally run older ClickHouse with the
-// flag off, so skip there — same pattern as the other events-read-path suites.
-const maybeDescribe =
-  env.LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN === "true"
-    ? describe
-    : describe.skip;
-
 // Legacy (v3-data-model) endpoints attach a top-level `_deprecation`
-// object so coding agents get a self-correcting migration signal.
-maybeDescribe("public API deprecation signal", () => {
+// object so coding agents get a self-correcting migration signal. Cloud-only:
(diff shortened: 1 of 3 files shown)


niklassemmler:
* **#16210** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16210?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16210?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16209** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #16216, opened 2026-08-17
feat(io-normalization): define normalized I/O contract and projections

marliessophie:
🟡 **Live preview: [pr-16216 app preview](https://pr-16216.preview.langfuse.com)**

## Summary

Draft of the canonical normalized observation I/O contract — please review types and
function signatures; implementations are WIP and nothing is wired into production.

One function normalizes I/O from any transport; projections derive consumer shapes
from the normalized result only:

    OTel span + context ────────┐
    ClickHouse event row ───────┼─▶ normalizeIO(source) ─▶ NormalizedIO ─▶ toToolColumns()
    { input, output, metadata } ┘                                       └▶ toEvalRecord()

## Interface

```ts
normalizeIO(source): NormalizedIO;
// source is a discriminated union:
//   { kind: "event-record", record }  // CH event row, metadata as name/value column arrays
//   { kind: "io", io }                // { input, output, metadata }
//   { kind: "otel", span, context }   // raw OTel span (extraction is a stub — see caveats)

toToolColumns(normalizedIO): ToolColumns;
toEvalRecord(normalizedIO, observation): NormalizedEvalRecord;
```

`NormalizedIO` contains:

- `messages` — one ordered stream (input messages, then output messages), each
  message keeping its input/output origin. Parts are a discriminated union:
  text, reasoning, tool call, tool result, file, data, custom.
- `toolDefinitions` — the tools available to the model, merged by name.
- `span` — the raw pre-normalization `{ input, output, metadata }` compiled from the source. Always returned for now.

## Todos

_To be done in seperate tickets_

- [ ] **JSON-parsing improvements**. Not optimised yet. 
- [ ] **OTel → SpanIO extraction**: the `{kind: "otel"}` source is a stub; extract the raw input/output/metadata discovery out of `OtelIngestionProcessor.extractInputAndOutput` / `extractMetadata`, with
   the processor calling in.
- [ ] **Rollout**: across product surfaces. Shadow ingestion first (env-gated
   via the sampled-projects pattern, divergence to metrics only), then an
   additive `normalizedIO` field on the trace-view endpoints with a
   raw-vs-normalized toggle, then evals. Include tests for `toEvalRecord` to reach full projection coverage. 
- [ ] **LangChain tool_call_chunks follow-up**: currently a documented ignore
   (streaming deltas, redundant with parsed `tool_calls`); later merge and
   parse them only when `tool_calls` is absent (mid-stream captures).
- [ ] **providerMetadata persistence**: transcripts, citations, media-token
   `source`, and other verbatim provider payloads ride on part-level
   `providerMetadata`; we need a story for saving and querying it. (The
   parser-computed semantics — refusal/reasoning/invalid — are typed part
   fields now, so the highest-value filters no longer depend on this.)

This PR introduces an experimental normalized observation I/O contract, provider-aware parsers, and projections for ClickHouse tool columns and evaluation records.
- Adds provider conventions and normalization for messages, parts, roles, finish reasons, tools, media, and metadata.
- Exposes client-safe and server-side normalized-I/O entry points.
- Adds broad fixture-based coverage for supported provider payloads.
- Adds projections from canonical normalized I/O into downstream consumer shapes.
</details>

The PR appears safe to merge as an unwired experimental contract, with two non-blocking normalization issues to address before consumers adopt it.

An unknown reserved part type can currently throw instead of using the lossless fallback, and the evaluation projection drops a tool-call field that normalization deliberately preserves.

**Files Needing Attention:** packages/shared/src/utils/normalized-io/core/normalize/part.ts; packages/shared/src/server/normalized-io/projections/eval-record.ts
</details>

The new client-safe parser can throw on a structured part with `type: "__proto__"` because the input-controlled type indexes a prototype-bearing handler registry. **How this was verified:** The lookup resolves `__proto__` to non-callable `Object.prototype`, which is then invoked unconditionally.
</details>

```mermaid
flowchart LR
  A[Event record] --> N[normalizeIO]
  B[Raw SpanIO] --> N
  C[OTel span and context] --> X[OTel extraction stub]
  X --> N
  N --> D[NormalizedIO]
  D --> M[Ordered normalized messages]
  D --> T[Tool definitions]
  D --> R[Raw SpanIO]
  D --> P1[toToolColumns]
  D --> P2[toEvalRecord]
```
</details>

`````markdown
### Issue 1
packages/shared/src/utils/normalized-io/core/normalize/part.ts:87-89
**Prototype-backed handler lookup**

When a structured part has `type: "__proto__"`, this lookup resolves to `Object.prototype` and attempts to call it, throwing a `TypeError` instead of preserving the unknown part through fallback handling.

**How this was verified:** The input-derived key indexes a plain object without an own-property check, so `__proto__` resolves to the non-callable prototype.

```suggestion
    const sharedHandler = Object.hasOwn(SHARED_TYPED_PART_HANDLERS, type)
      ? SHARED_TYPED_PART_HANDLERS[type]
      : undefined;
    if (sharedHandler) {
      const result = sharedHandler(value, partContext);
```

### Issue 2
packages/shared/src/server/normalized-io/projections/eval-record.ts:79
**Tool-call type is discarded**

For normalized calls with a `toolType`, this projection always emits an empty `type`, so evaluation consumers lose the call kind even though normalization retains it and the tool-column projection preserves it.

```suggestion
        type: part.toolType ?? "",
```

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor: format tool-call"](https://github.com/langfuse/langfuse/commit/204959a81a8f986d77dff69f64b7361ddaef042c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57891544)</sub>

### CODE REVIEW

File: packages/shared/src/server/ingestion/normalizedIO/parser.ts [hunk 1]
```diff
@@ -0,0 +1,2065 @@
...
+  );
+
+  return (
+    hasOpenAiShape ||
+    hasAiSdkShape ||
+    hasResponsesShape ||
+    hasAnthropicShape ||
+    hasFlatShape
+  );
+}
```
Steffen911: My idea around this would be a more provider focused flow that implements an interface. Then we'd parse the inputs through each of the Provider/Plugin implementations and join the results.

E.g. we'd have
```
class AnthropicParser implements LangfuseParser {
  overwrite function isToolCallLike(value: Record<string, unknown>): boolean {
    return Boolean(
      ["tool_use", "server_tool_use", "mcp_tool_use"].includes(
        String(value.type),
      ) &&
      value.name &&
      "input" in value,
    )
  }
}
```

And then we'd register the class with a central provider and it would do a 
```
   isToolCallLike = Array.some(registeredProviders.map((p) => p.isToolCallLike(input))
```
within a central component.

IMO this improves readability and testing as the logic for each provider would be in a very narrow file instead of scattered across 3000 lines of JavaScript. This would also allow for strong linking between fixture and implementation.

marliessophie: Ty for sharing review. 

In a perfect world, I'd love to implement your suggestion. Appreciate the underlying point on readability. Happy to improve here, but I'd have to push back on the provider-interface framing. 

The dialects aren't provider-disjoint. A LangChain-framed message carries OpenAI-shaped tool calls inside additional_kwargs; proxied traffic (OpenRouter/LiteLLM) blends conventions within a single record. And some shapes belong to no provider at all, but rely on catch all (eg hasFlatShape is the residue of proxies and older SDK serializations). A per-provider class implies each record has an owner; in practice telemetry data is more "messy" and conventions co-occur within one message, which is also why I want to avoid any "detect the provider, then parse" flow: it has to commit early on ambiguous evidence and partially parses mixed records. 

What I'd consider doing instead: split the shape predicates into convention modules (shapes/openai.ts, shapes/anthropic.ts, shapes/aiSdk.ts, shapes/generic.ts), and keep isToolCallLike as the one explicit OR-composition so precedence stays readable in one place. Deliberately calling them "shapes" rather than "parsers/providers" so nobody is tempted to add a detect() later.

Would that address your readability/testing concern?

Steffen911: @marliessophie My thinking was that all the implementations are "chained" and cumulative, i.e. just because something matches the Anthropic provider doesn't mean the the OpenAI provider is skipped. If a framework uses multiple conventions all of them would still apply. 
The conventions would have a more close "home" though.

If we close with a catchall provider that would also work IMO.

marliessophie: Ah I understand now. Ty, for clarification. 

In that case I think the ask is: pull the per-convention logic out of the 2k-line parser.ts into narrow per-provider files/provider interfaces; but keep matching cumulative as per current implementation. Consider moving fixtures next to per-provider parser files.

packages/shared/src/server/normalized-io/README.md
```diff
@@ -0,0 +1,201 @@
...
+  carriers, matching what the source actually provides.
+- **`SpanIO.metadata` may carry an `attributes` record with OTel keys**; the
+  parser mines known tool-definition attribute keys from it
+  (`ai.prompt.tools`, `gen_ai.tool.definitions`, `llm.tools.N.…`,
+  `model_request_parameters.function_tools`). Open tension: these are
+  OTel-format names inside the transport-independent core.
+
+### Intentional losses
+
+- Anthropic `cache_control` (request caching hint).
```
hassiebp: this should live in model params. where did this come up in input / output / metadata?

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Semantics the parser computes are typed fields on their parts (e.g.
+ * `refusal`, `reasoning`, `invalid`); `providerMetadata` carries verbatim
+ * provider payloads only (citations, transcripts, server labels, statuses).
+ * Part types encode what a consumer can do with a part; new provider
+ * concepts start as providerMetadata and get promoted to a typed field once
+ * they are cross-provider semantics we compute and consumers filter on.
+ */
+export type NormalizedPartBase = {
+  providerMetadata?: JsonObject;
```
hassiebp, line 30: This is basically an untyped catchall for provider specific data that is not mapped to the normalized messages?

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,197 @@
...
+  filename?: string;
+  /** File was produced as reasoning output (e.g. AI SDK reasoning-file). */
+  reasoning?: true;
+  content:
+    | { kind: "url"; url: string }
+    | { kind: "base64"; data: string }
+    | { kind: "reference"; id: string };
+};
+
+export type DataPart = NormalizedPartBase & {
```
hassiebp, line 103: What is usually in a `data` and `custom` part?

packages/shared/src/server/normalized-io/README.md
marliessophie: In the current PR, this came from the Anthropic schema rather than an observed production case. The fixture places it on an input text content block. Given that this only arose from schema coverage, I would be comfortable removing the explicit handling for now.

packages/shared/src/utils/normalized-io/types.ts
marliessophie: Yes. providerMetadata is the untyped catch-all for provider-specific fields that remain after a recognized message part or tool definition has been normalized.

packages/shared/src/utils/normalized-io/types.ts
marliessophie: Concrete custom examples include:
```
{
  type: "source",
  sourceType: "url",
  url: "https://example.com/rag-docs",
  title: "RAG integration docs",
}
```
which becomes custom { kind: "source" }, because it is a stream-positioned AI SDK source without a text anchor. Similar to the citations/annotations example, just as it's own part as it is not attached to a specific "source". 

Another example is an Anthropic document whose source contains inline text rather than a resolvable file:
```
{
  type: "custom",
  kind: "document",
  value: { type: "document", source: { type: "text", ... } },
}
```
An unknown future block such as { type: "chart", ... } would also become custom { kind: "chart" }

For data, 

arbitrary structured observation I/O that is not a chat message, for example a retrieval/function span:
```
input: {
  query: "What is the refund policy?",
  filters: { region: "EU" },
}
```

structured output such as `{ documents: [...], latencyMs: 42 }` becomes a data part too.

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,201 @@
...
+    | "error"
+    | "other"
+    | "unknown";
+  raw: string;
+};
+
+export type NormalizedMessage = {
+  id?: string;
+  role: NormalizedMessageRole;
+  name?: string;
```
nimarb: do we need this? what would it be eg

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,201 @@
...
+  | ReasoningPart
+  | ToolCallPart
+  | ToolResultPart
+  | FilePart
+  | DataPart
+  | CustomPart;
+
+export type NormalizedMessageRole =
+  | "system"
+  | "developer"
```
nimarb: consolidate w system?

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,201 @@
...
+  | "user"
+  | "assistant"
```
nimarb: should we name it llm ? something to be considered, no v strong opinion

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,201 @@
...
+  kind: string;
+  value: JsonValue;
+};
+
+export type NormalizedMessagePart =
+  | TextPart
+  | ReasoningPart
+  | ToolCallPart
+  | ToolResultPart
+  | FilePart
```
nimarb, line 120: should we differentiate here:

`MediaPart` -> something we can render within langfuse
`FilePart` / `DataPart` arbirtarty file which is always only donwloaded

packages/shared/src/utils/normalized-io/types.ts
```diff
@@ -0,0 +1,201 @@
...
+
+export type ToolCallPart = NormalizedPartBase & {
+  type: "tool-call";
+  toolCallId: string | null;
+  toolName: string;
+  input: JsonValue;
+  /** Raw source type used by compatibility projections (for example
+   * `function`, `custom`, or a provider-specific built-in kind). */
+  toolType?: string;
+  index?: number;
```
nimarb: drop? we know already the position of the call from the array

packages/shared/src/utils/normalized-io/types.ts
marliessophie: file part deliberately does not split into "renderable media" vs. "download-only file": renderability is a consumer capability decision that changes as the UI gains viewers. I would not want to look this in on a schema level, even if schema is currently only applied at read

packages/shared/src/utils/normalized-io/types.ts
marliessophie: no v strong opinion either, assistant felt closest to provider semantics. can rename to llm too, if we feel this is more feature proof

packages/shared/src/utils/normalized-io/types.ts
marliessophie: dropped. included senderName instead as some providers include user or agent name or similar with message.

PR #16223, opened 2026-08-18
fix(shared): register recurring cron jobs via atomic BullMQ job schedulers

Steffen911:
🟡 **Live preview: [pr-16223.preview.langfuse.com](https://pr-16223.preview.langfuse.com)**

## Problem

Every recurring cron job in Langfuse is registered via the deprecated `Queue.add(name, data, { repeat })` API. On that path, each booting container replaces the pending next iteration in **two separate Redis round trips**: a Lua script deletes the pending delayed job, then a separate `Job.create` call re-adds it. A producer that fails between the two — e.g. a container killed mid-boot during a fleet-wide deploy, or a Redis command lost in a cluster-wide connection storm — permanently ends the schedule: no error, no failed job, no stalled job. The queue just goes quiet until the next boot happens to re-register it.

This is not theoretical: it stalled event propagation in production for 85 minutes (delayed data visibility in the v4 UI), and the same registration pattern is shared by usage metering, data retention, blob-storage/PostHog/Mixpanel exports, DLQ retry, and the core-data S3 export. Upstream, the failure class is tracked in [taskforcesh/bullmq#3272](https://github.com/taskforcesh/bullmq/issues/3272); the atomic job-scheduler path was hardened there, but the legacy path never was, and bullmq v6 removes it entirely.

## Change

- New `scheduleRecurringJob` helper (`packages/shared/src/server/redis/scheduleRecurringJob.ts`):
  - registers the cron via `Queue.upsertJobScheduler` — delete + re-create runs as a **single atomic Lua script**, so concurrent boots and mid-boot crashes cannot lose the chain, and id collisions surface as thrown errors instead of silent no-ops;
  - removes the legacy md5-keyed schedules for the current pattern and any historical ones (`previousPatterns`), so the old chain cannot keep firing in parallel — this cleanup must stay while direct upgrades from legacy-scheduling versions are supported;
  - never rejects; all failures are logged (registration is fire-and-forget from queue singleton getters).
- Converted all 11 recurring-cron call sites to the helper. The scheduler id is the job name; job names, payloads, patterns, and default job options are unchanged.
- The two pre-existing `eslint-disable @typescript-eslint/no-deprecated` comments for `removeRepeatable` (blobStorage, v4LegacyApiUsage) are consolidated into one inside the helper (net 2 → 1); those two sites already carried the note "job scheduler migration should be handled separately" — this PR is that migration.
- Consumer side needs no changes: the worker already iterates scheduler-keyed jobs via the job-scheduler path in bullmq 5.76.3.

## Rollout / rollback notes

- **Rolling deploy**: safe. Old and new tasks run identical bullmq; only the boot-time registration differs. The legacy cleanup and the upsert cannot delete each other's entries (different key shapes), and any interleaving converges to exactly one scheduler.
- **Tick continuity**: the upsert schedules the next occurrence of the same pattern; no duplicate or missed tick expected at cutover.
- **Rollback**: a rolled-back container re-adds the legacy schedule while the named scheduler persists → duplicate ticks until cleanup. Handlers at highest duplicate sensitivity are already guarded (event propagation: Redis cursor + global concurrency 1; usage metering: DB lock + not-due check). To clean up after a rollback: `queue.removeJobScheduler("<job-name>")` per queue, or roll forward.

## Impacted packages

- `@langfuse/shared` (all changes; consumed by `worker` and `web` queue singletons)

## Verification

- `pnpm tc` → `Tasks: 7 successful, 7 total`
- `pnpm --filter @langfuse/shared run test scheduleRecurringJob` → `Tests 4 passed (4)`
- `pnpm --filter @langfuse/shared run test v4LegacyApiUsageQueue` → `Tests 1 passed (1)`

A deterministic failing-first test for the original race isn't feasible without fault injection between bullmq's two internal Redis calls; the new tests instead pin the invariants that close it (single atomic upsert, legacy cleanup ordered before the upsert, never-reject semantics).

This PR migrates recurring BullMQ jobs from deprecated repeat options to a shared job-scheduler helper.
- Adds centralized legacy schedule cleanup and `upsertJobScheduler` registration.
- Converts eleven recurring queue registrations while preserving names, payloads, and cron patterns.
- Adds tests for cleanup ordering, payload forwarding, and error handling.
</details>

The non-atomic legacy cleanup and scheduler upsert can still stop or duplicate recurring work and should be fixed before merging.

Worker boot removes legacy schedules before creating their replacements, but failures on either side are only logged and never reconciled, allowing zero or two active recurring chains.

**Files Needing Attention:** packages/shared/src/server/redis/scheduleRecurringJob.ts
</details>

```mermaid
sequenceDiagram
  participant W as Worker boot
  participant H as scheduleRecurringJob
  participant R as Redis/BullMQ
  W->>H: Fire-and-forget registration
  H->>R: removeRepeatable(legacy pattern)
  R-->>H: Legacy chain removed
  Note over H,R: Process or connection failure here leaves no chain
  H->>R: upsertJobScheduler(job name)
  R-->>H: Scheduler registered
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/redis/scheduleRecurringJob.ts:49
**Non-atomic schedule migration**

If a worker terminates or loses Redis after `removeRepeatable` succeeds but before `upsertJobScheduler` completes, this once-per-boot helper leaves no recurring chain until another worker boots. Conversely, when cleanup fails transiently and the subsequent upsert succeeds, the legacy and scheduler chains both remain active and enqueue duplicate ticks.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(shared): register recurring cron job..."](https://github.com/langfuse/langfuse/commit/7bda866fddabfa38c00b9204781f9e8c6763b253) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54280767)</sub>

**Context used:**

### DIFF
--- a/packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts
+++ b/packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts
@@ -2,6 +2,7 @@ import { Queue } from "bullmq";
 import { env } from "../../env";
 import { QueueName, QueueJobs } from "../queues";
 import { createBullMQQueueOptionsWithRedis } from "./redis";
+import { scheduleRecurringJob } from "./scheduleRecurringJob";
 import { logger } from "../logger";
 
 export class CloudFreeTierUsageThresholdQueue {
@@ -45,20 +46,16 @@ export class CloudFreeTierUsageThresholdQueue {
         "[CloudFreeTierUsageThresholdQueue] Scheduling recurring job",
         {
           pattern: "35 * * * *",
-          jobId: "free-tier-usage-threshold-hourly",
           description: "Every hour at minute 35",
           timestamp: new Date().toISOString(),
         },
       );
 
-      CloudFreeTierUsageThresholdQueue.instance.add(
-        QueueJobs.CloudFreeTierUsageThresholdJob,
-        { type: "recurring" },
-        {
-          repeat: { pattern: "35 * * * *" },
-          // jobId: "free-tier-usage-threshold-hourly", // CRITICAL: Unique ID prevents duplicates across containers
-        },
-      );
+      scheduleRecurringJob(CloudFreeTierUsageThresholdQueue.instance, {
+        jobName: QueueJobs.CloudFreeTierUsageThresholdJob,
+        pattern: "35 * * * *",
+        data: { type: "recurring" },
+      });
 
       // Optional: Bootstrap job for immediate execution on startup
       // This ensures usage thresholds are processed immediately when service starts
--- a/packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts
+++ b/packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts
@@ -51,10 +51,12 @@ export class CloudFreeTierUsageThresholdQueue {
         },
       );
 
+      // The legacy schedule carried data { type: "recurring" } to distinguish
+      // it from a bootstrap job that was never enabled; nothing reads it and
+      // job data has no semantics to BullMQ, so the payload is now empty.
       scheduleRecurringJob(CloudFreeTierUsageThresholdQueue.instance, {
         jobName: QueueJobs.CloudFreeTierUsageThresholdJob,
         pattern: "35 * * * *",
-        data: { type: "recurring" },
       });
 
       // Optional: Bootstrap job for immediate execution on startup
--- a/packages/shared/AGENTS.md
+++ b/packages/shared/AGENTS.md
@@ -154,6 +154,11 @@ the same PR.
 
 - Keep backward compatibility in queue payloads when possible during rolling
   deployments.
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: packages/shared/src/server/redis/cloudFreeTierUsageThresholdQueue.ts [hunk 1]
```diff
@@ -45,20 +46,16 @@ export class CloudFreeTierUsageThresholdQueue {
...
           pattern: "35 * * * *",
-          jobId: "free-tier-usage-threshold-hourly",
           description: "Every hour at minute 35",
           timestamp: new Date().toISOString(),
         },
       );
 
-      CloudFreeTierUsageThresholdQueue.instance.add(
-        QueueJobs.CloudFreeTierUsageThresholdJob,
-        { type: "recurring" },
```
Steffen911, line 56: What does the type do? Why don't we set it everywhere?

PR #16252, opened 2026-08-18
fix(shared): make legacy BullMQ cleanup safe for FIPS

### LINKED ISSUE #14611: bug: worker fails in Node FIPS mode because BullMQ repeatable jobs default to MD5 hashing

Leundai:
Fixes #14611.

BullMQ Job Schedulers do not use repeat-key hashes. This PR gives FIPS deployments a safe path away from legacy MD5 repeatables: clean Redis starts with cleanup disabled, while upgrades first remove legacy schedules in a non-FIPS deployment.

<details>
  <summary>What changed and testing</summary>

### What changed

- Add `LANGFUSE_BULLMQ_LEGACY_REPEATABLE_JOB_CLEANUP`, which defaults to the existing cleanup behavior.
- Discover and remove stored legacy schedules, including unlisted historical cron patterns and interval schedules. With cleanup disabled, inspect Redis without hashing and refuse to create a scheduler while any legacy entry remains.
- Document the fresh-install and upgrade paths in the environment examples.

### Testing plan

- Passed: focused shared test suite (9 tests), shared lint, typecheck, build, formatting, and repository lint checks.
- Web and worker typechecks are currently red on the rebased `main` because of unrelated missing shared exports in the newly merged integration and worker changes, plus existing web type errors.

</details>


packages/shared/src/server/redis/scheduleRecurringJob.ts
```diff
@@ -1,13 +1,39 @@
...
-// TRIGGERS FOREVER (bullmq v6 workers keep iterating stranded md5-keyed
-// chains; only manual Redis cleanup stops them).
-// HOLD THE BULLMQ v6 UPGRADE UNTIL LANGFUSE V5: v6 deletes
-// Queue.removeRepeatable(), which this cleanup requires.
+// Keep this cleanup through V4. Without it, a direct upgrade from the legacy
+// repeat API leaves an MD5-keyed schedule firing alongside the job scheduler.
+// BullMQ v6 removes Queue.removeRepeatable(), so the cleanup must be completed
+// before that upgrade. For FIPS, set cleanup to false from the first boot only
+// when Redis has no legacy schedules. Upgrades must run cleanup in a non-FIPS
+// deployment first.
```
Leundai, line 10: Don't think we should be pruning this comment or rewriting it..

PR #16253, opened 2026-08-18
refactor(integrations): share one export-source default across integration families

niklassemmler:
🟡 **Live preview: [pr-16253.preview.langfuse.com](https://pr-16253.preview.langfuse.com)**

## What does this PR do?

Export-source policy was duplicated across the three integration families that
share it (blob storage, PostHog, Mixpanel). This consolidates it and fixes one
behavioural divergence, plus a naming/wording cleanup in the shared module.

**One create default, and it prefers enriched.** The "which source should a new
integration get" rule existed twice with different answers: the settings pages
derived it from enriched availability, while the tRPC routers and the public
REST handler derived it from `isCloud`/legacy-write capability. They disagreed on
a self-hosted `dual` deployment — the page offered enriched observations while a
create landed the legacy source. There is now a single `defaultExportSource(ctx)`
in the shared policy that every write path calls, and it prefers the enriched
source wherever the deployment writes it.

**One post-upsert backstop.** Blob storage validated the landed row inline via
two `*OnCreate` booleans threaded into its service; PostHog/Mixpanel had their
own snapshot-diff variant that skipped the check entirely on some paths. Both now
use `assertPersistedExportSourceAllowed`, which runs unconditionally and judges a
row this transaction created by new-integration rules rather than by its
timestamp. That distinction matters while a family's exporter cutoff is still in
the future: a row created *now* is before the cutoff, so validating by date alone
would grandfather a brand-new integration and let it keep a legacy source.
`forceEventsOnCreate` and `refuseLegacyOnCreate` are gone — the resolved source
is now always concrete, so a CREATE can never fall through to the Prisma column
default.

**Customer-facing wording.** The enriched-unavailable validation message said
"Enriched blob export is not available on this deployment". PostHog and Mixpanel
reach that same string, so their users were told about blob export. It is now
worded integration-neutrally and names the write mode, matching its sibling
message.

**Naming.** Seven identifiers in the shared policy carried `Blob`/`BLOB` despite
applying to all three families — most notably `BlobExportWriteMode`, which just
mirrors `LANGFUSE_MIGRATION_V4_WRITE_MODE`. Renamed without `Blob`.
`LEGACY_BLOB_EXPORTER_CUTOFF` keeps its name: that one really is the blob
family's own cutoff. The `NEXT_PUBLIC_*` cutoff override env vars are unchanged.

Also renamed `server/analyticsExportSource.ts` to `server/exportSource.ts` (the
`analytics` prefix was redundant inside `analytics-integrations/` and misleading,
since blob storage is an analytics integration that did not use it), and removed
internal ticket references from the comments in the touched files.

### Behaviour change worth a reviewer's attention

On a **self-hosted `dual`** deployment, a create that omits `exportSource` — via
the settings pages, tRPC, or `PUT /api/public/integrations/blob-storage` — now
lands `EVENTS` instead of `TRACES_OBSERVATIONS`. That is the point of the change,
but it is the one externally visible difference. Cloud and the `legacy` /
`events_only` write modes are unaffected: the old and new rules already agreed
there. Existing rows are untouched; a persisted value is still never rewritten.

## Type of change

- [x] Refactor
- [ ] Documentation update
- [x] Chore (dependency upgrades, minor cleanups, etc.)

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (no user-facing docs describe the create default)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

### Verification

- shared `export-source-policy` — `Tests 33 passed (33)` (5 new `defaultExportSource` cases)
- web `exportSource.clienttest` — `Tests 49 passed (49)`
- web `assertExportSourceAllowed` — `Tests 16 passed (16)`
- web `posthog-integration` — `Tests 34 passed (34)`
- web `mixpanel-integration` — `Tests 33 passed (33)`
- web `blob-storage-integration-trpc` — `Tests 46 passed (46)` (1 new case: self-hosted `dual` create)
- web `blob-storage-integration-api` — `Tests 58 passed (58)`
- worker `blobExportAbortClassification` — `Tests 13 passed (13)`


niklassemmler:
* **#16253** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16253?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #16257, opened 2026-08-18
refactor(worker): name the blob export terminal-failure outcomes

niklassemmler:
🟡 **Live preview: [pr-16257.preview.langfuse.com](https://pr-16257.preview.langfuse.com)**

## What does this PR do?

Preparatory refactor of the terminal error path in the blob storage export
handler. No behavior change.

That `catch` block decides which failure email to send by recombining three
mutable flags (`persistedDisable`, `disableClaimRan`, `enabledAfterPersist`,
each with a fallback initializer) into negated conjunctions. Several earlier
changes each layered one more guard onto the same straight-line body, so the
result is correct but opaque: which flag combinations are actually reachable is
not apparent from reading it. That makes any edit to the path a re-derivation of
the races each guard was defending rather than a code move — which is the
expensive part of the follow-up that puts blob export on the same fail-fast
handling as the PostHog export.

Four commits, meant to be reviewed in order:

**1. Pin the unguarded cases** (test-only). Two situations the flag logic guards
against had no coverage: a concurrent run winning the customer-fault disable
claim, and the failure being unwritable because Postgres is unavailable. Either
guard could be deleted and the whole suite stayed green. Both new tests
discriminate rather than merely pass — the race one was verified by neutering the
guard, where it and only it fails — and both pass against the unmodified
handler, so they characterize existing behavior.

**2. Name the outcomes.** The database bookkeeping is extracted and now returns a
discriminated union describing what happened (row deleted, failure unwritable,
failure recorded plus the row's enabled state, disabled by this worker, disable
claim lost to a concurrent run). The caller switches over it, so every guard is
one named case.

Two incidental simplifications fall out: the disable reason tag no longer needs
its `"unknown"` fallback, because the type proves the branch is only reachable
with a defined reason; and the exhausted-retries condition moves onto the two
informational cases that actually need it, making it visible that a disable is
terminal by construction and never consults the attempt count.

**3. Make the switch actually exhaustiveness-checked.** Review correctly caught
that commit 2 claimed compiler-enforced totality it did not have: a plain switch
over a union compiles clean with cases missing, and no lint rule here catches it.
The `never` assignment in a `default` branch is what makes the check bite,
verified by adding a sixth variant and watching typecheck fail on it. Braces
scope the declaration, so unlike the two older exhaustiveness-checked switches in
this file it needs no `no-case-declarations` disable.

**4. Tighten the comments** down to what a reader cannot recover from the code.

All five outcomes are under test.

## Type of change

## Mandatory Tasks

## Checklist

Behavior preservation is the whole point, so it is what I verified: each commit
is green against the suite the first commit strengthened, including all five
terminal outcomes.

- `blobStorageIntegrationProcessing` — `Tests 35 passed (35)`, run 5x to rule
  out flakiness
- `blobExport*` — `Tests 19 passed (19)`; `blobStorageTuningWiring` —
  `Tests 3 passed (3)`

One note for the reviewer: on the first cold run of the suite I saw the file's
first two tests fail on a 5s timeout, which did not reproduce in 5 subsequent
runs. Those tests exercise no decision this PR changes, and the failure mode was
a timeout rather than an assertion mismatch, consistent with cold-start cost on
a network-bound test. Flagging it rather than burying it.

### DIFF
--- a/worker/src/__tests__/blobStorageIntegrationProcessing.test.ts
+++ b/worker/src/__tests__/blobStorageIntegrationProcessing.test.ts
@@ -404,6 +404,89 @@ describe("BlobStorageIntegrationProcessingJob", () => {
       expect(row.lastError).toMatch(/connection refused/i);
       expect(row.lastFailureNotificationSentAt).toBeNull();
     });
+
+    // A concurrent run (distinct jobId, so not queue-deduped) can win the
+    // disable claim first. The loser must stay silent: the winner already sent
+    // the terminal "disabled" email, and the loser's only fallback is the
+    // informational "will retry at the next scheduled export" variant, which is
+    // false for an integration that is now off.
+    it("does not notify when a concurrent run won the customer-fault disable claim", async () => {
+      const { projectId } = await createOrgProjectAndApiKey();
+      s3Prefix = projectId;
+      await createIntegration(projectId);
+      mockRecordIncrement.mockClear();
+      // Land the winner's disable between the handler's enabled check and the
+      // terminal catch, then raise the same fault the winner disabled on.
+      mockValidateBlobStorageEndpoint.mockImplementationOnce(async () => {
+        await prisma.blobStorageIntegration.updateMany({
+          where: { projectId },
+          data: { enabled: false },
+        });
+        throw accessDeniedError();
+      });
+
+      await expect(runAttempt(projectId, 4)).rejects.toThrow(/access denied/i);
+      await settleBackgroundTasks();
+
+      const row = await prisma.blobStorageIntegration.findUniqueOrThrow({
+        where: { projectId },
+      });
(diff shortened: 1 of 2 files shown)


niklassemmler:
* **#16261** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16257** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #16261, opened 2026-08-18
fix(worker): fail fast and auto-disable the blob export on config faults

niklassemmler:
🟡 **Live preview: [pr-16261.preview.langfuse.com](https://pr-16261.preview.langfuse.com)**

## What does this PR do?

Makes the blob storage export fail fast on deterministic customer-config faults,
matching the behavior the PostHog export already has.

A rejected bucket, credential, or endpoint cannot succeed until the customer
fixes it. The handler already classified these correctly, but still rethrew and
relied on retry-then-disable-on-the-final-attempt. The queue's failure counter is
emitted per *failed attempt*, so every scheduled run of a misconfigured export
lit the processing-failures monitor about five times before anything turned it
off — alert noise for a condition no retry can clear. Silencing it requires the
job to *complete*, not merely to stop retrying.

On a classified fault the handler now disables on first occurrence and resolves
the job. Unclassified and transient errors are untouched: they record the error,
notify once the retries run out, and rethrow. `dns-lookup-failed` stays off the
allowlist, so a DNS outage cannot disable a working integration.

Builds on #16257 (now merged), which named these outcomes and pinned them under
test; that is what reduces this change to the control-flow edit it should be.

Reviewer note: splitting the notification helper into an awaited function and a
fire-and-forget wrapper reindents about 80 lines of its body without changing
them. Review with whitespace changes hidden — the real edit is the two-line
wrapper plus the `await` at the call site.

Three consequences worth calling out:

- **The terminal notification is now awaited** rather than fired and forgotten.
  The job resolves immediately afterwards and the scheduler only enqueues enabled
  integrations, so a dispatch cut off mid-flight would leave the export off with
  nobody told and no retry path. The informational variant stays fire-and-forget,
  since its job is about to fail and be retried anyway.

  To be precise about what this buys: sequencing, not delivery. Dispatch failures
  are still logged and swallowed — as in the PostHog handler — because letting
  them fail the job would relight the monitor without redelivering the email (the
  integration is disabled by then, so a retry returns early and never notifies).
  The persisted `lastError`, surfaced in the integration settings, stays the
  durable signal. Raised by review; left open rather than papered over.

- **Losing the disable claim to a concurrent run also resolves.** The integration
  is already off and the winner already notified; throwing would light the
  monitor for a fault that has been handled.

- **An unwritable failure keeps retrying and no longer emails.** Nothing was
  persisted, so the run cannot claim the fault is handled — and the informational
  email's "will retry" text is the one thing the rethrow makes true.

The second commit fixes a customer-visible string this change invalidates: the
disabled-export email said the integration was turned off "after repeated
failures", which described a retry history the customer no longer has. It now
names the cause, matching the PostHog copy's "after a configuration fault".

## Type of change

## Mandatory Tasks

## Checklist

The tests that changed direction are the deliverable, so they are worth reading
as the spec: a classified fault now resolves and disables on the first attempt,
the race loser resolves silently, and an unwritable failure retries without
emailing. Each of those assertions was inverted from a test that pinned the old
behavior in #16257, so the change of intent is visible in the diff rather than
implied.

- `blobStorageIntegrationProcessing` — `Tests 35 passed (35)`
- `blobExport*` — `Tests 19 passed (19)`; `blobStorageTuningWiring` —
  `Tests 3 passed (3)`
- `dispatchProjectNotification` (shared, for the email copy) —
  `Tests 3 passed (3)`

Not done, flagging rather than assuming: the hosted docs describe the export
failure/disable behavior and live in a separate repo, so if they state the
retry-then-disable sequence they will need a matching edit.

### DIFF
--- a/packages/shared/src/server/notifications/types.ts
+++ b/packages/shared/src/server/notifications/types.ts
@@ -46,8 +46,8 @@ export const ProjectNotificationEventSchema = z.discriminatedUnion(
       eventType: z.literal(
         ProjectNotificationEventTypeSchema.enum["blob-export-failed"],
       ),
-      // true when the integration was auto-disabled after repeated failures
-      // (terminal event; selects the "disabled" email variant).
+      // true when the integration was auto-disabled after a configuration
+      // fault (terminal event; selects the "disabled" email variant).
       disabled: z.boolean().optional(),
     }),
     projectNotificationEventBaseSchema.extend({
--- a/packages/shared/src/server/services/email/blobStorageExportFailed/sendBlobStorageExportFailedEmail.tsx
+++ b/packages/shared/src/server/services/email/blobStorageExportFailed/sendBlobStorageExportFailedEmail.tsx
@@ -16,9 +16,9 @@ export const blobStorageExportFailedEmailCopy: IntegrationExportFailedEmailCopy 
     disabledBody: (projectName) => (
       <>
         The blob storage export for project &quot;{projectName}
-        &quot; has been disabled after repeated failures. This usually means the
-        destination configuration or credentials are no longer valid. Once you
-        have updated them, simply re-enable the export in the integration
+        &quot; has been disabled after a configuration fault. This usually means
+        the destination configuration or credentials are no longer valid. Once
+        you have updated them, simply re-enable the export in the integration
         settings and it will pick up right where it left off.
       </>
     ),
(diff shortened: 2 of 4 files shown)


niklassemmler:
* **#16261** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16257** <a href="https://app.graphite.com/github/pr/langfuse/langfuse/16257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

niklassemmler:
@claude review

Issue #16233, opened 2026-08-18
CI Runtime Report: 2026-08-12..08-18 trailing-7-day analysis (33 merge-group runs, no actionable regression)


wochinge:
Weekly CI runtime analysis of the `pipeline.yml` ("CI/CD") workflow.
Window: **trailing 7 calendar days, 2026-08-12 .. 2026-08-18 (UTC)**.
Timing population: **successful `merge_group` runs only** (33 runs across 6 days;
2026-08-15 (Sat) had zero successful merge-group runs and is omitted).

## Chart 1 — pipeline totals

🔵 overall incl. wait · 🟠 overall excl. wait · 🟣 runner wait

```mermaid
%%{init: {"themeVariables": {"xyChart": {"plotColorPalette": "#3987e5,#de5a20,#8875e0"}}}}%%
xychart-beta
    title "Daily merge-group medians: pipeline totals (seconds)"
    x-axis [08-12, 08-13, 08-14, 08-16, 08-17, 08-18]
    y-axis "seconds" 0 --> 700
    line [336, 676, 230, 194, 201, 195]
    line [306, 667, 207, 179, 186, 187]
    line [25, 9, 15, 15, 16, 10]
```

## Chart 2 — critical-path segments

🔵 run tests · 🟠 Build · 🟣 e2e-tests

```mermaid
%%{init: {"themeVariables": {"xyChart": {"plotColorPalette": "#3987e5,#de5a20,#8875e0"}}}}%%
xychart-beta
    title "Daily merge-group medians: segments (seconds)"
    x-axis [08-12, 08-13, 08-14, 08-16, 08-17, 08-18]
    y-axis "seconds" 0 --> 400
    line [96, 347, 82, 78, 77, 77]
    line [70, 149, 51, 47, 48, 48]
    line [169, 285, 197, 172, 149, 147]
```

| Day | runs | overall incl. wait | overall excl. wait | runner wait | run tests | Build | e2e-tests |
|---|---|---|---|---|---|---|---|
| 08-12 | 5 | 336 | 306 | 25 | 96 | 70 | 169 |
| 08-13 | 5 | 676 | 667 | 9 | 347 | 149 | 285 |
| 08-14 | 5 | 230 | 207 | 15 | 82 | 51 | 197 |
| 08-16 | 5 | 194 | 179 | 15 | 78 | 47 | 172 |
| 08-17 | 5 | 201 | 186 | 16 | 77 | 48 | 149 |
[...]

PR #16282, opened 2026-08-19
chore(ci): test pure client files without jsdom

wochinge:
🟡 **Live preview: [pr-16282.preview.langfuse.com](https://pr-16282.preview.langfuse.com)**

This is an experiment to measure the real CI impact of running pure client tests without jsdom.

## What changed

- 136 client test files now run in Vitest's Node environment.
- 135 files receive a mechanical `@vitest-environment node` annotation in this PR; one file already used Node on `main`.
- The regular client suite still passes: 283 files passed / 1 skipped, with 3,367 tests passed / 51 skipped.

## Reasoning

The [CI runtime report](https://github.com/langfuse/langfuse/issues/16233) shows substantial summed environment time on the client shard. A source-level scan found 140 candidates. The full suite exposed four files with transitive browser dependencies, so those remain on jsdom.

Local timings were too noisy to support a decision. Merge this only if repeated CI runs show a clear improvement.

Trade-off: this adds broad mechanical annotations across 135 files. The explicit annotations are intentional - they avoid a source-scanning heuristic that could silently select the wrong environment.

## Measurement

- [ ] Record the client shard `run tests` duration from the first CI run.
- [ ] Repeat the CI workflow at least twice on the same commit.
- [ ] Compare the median against recent `main` runs on the same runner class.
- [ ] Merge only with a clear, repeatable improvement; otherwise close the experiment.

### DIFF
--- a/web/src/__tests__/aggregateScores.clienttest.ts
+++ b/web/src/__tests__/aggregateScores.clienttest.ts
@@ -1,3 +1,5 @@
+// @vitest-environment node
+
 import {
   aggregateScores,
   type ScoreToAggregate,
--- a/web/src/__tests__/chart-loading-state-utils.clienttest.ts
+++ b/web/src/__tests__/chart-loading-state-utils.clienttest.ts
@@ -1,3 +1,5 @@
+// @vitest-environment node
+
 import { RESOURCE_LIMIT_ERROR_MESSAGE } from "@langfuse/shared";
 import {
   getChartLoadingProgress,
--- a/web/src/__tests__/eventsTablePaths.clienttest.ts
+++ b/web/src/__tests__/eventsTablePaths.clienttest.ts
@@ -1,3 +1,5 @@
+// @vitest-environment node
+
 import {
   buildEventsTablePathForObservationType,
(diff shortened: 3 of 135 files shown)


wochinge:
First CI measurements show a real improvement in the client-test job:

- Vitest runtime: `50.40s`, `50.88s`, `50.40s` -> median `50.40s`
- Nearest comparable jsdom run with the same 284 files: `56.24s`
- Vitest runtime: `-10.4%` / `-5.84s`
- GitHub Actions `run test-client` step: `59s` -> `53s`
- Full `tests-web-client` job: `86s` -> `81s`
- Summed environment time: `48.06s` -> `~26.2s` (`-45%`)

This job is not currently on the pipeline's critical path, so the change does not reduce the total CI completion time while the server / e2e jobs remain slower.

One initial run hit an unhandled Radix focus-scope timer during teardown after all 3,367 tests passed. The next two runs passed. That reliability signal is the remaining trade-off, so this stays a draft for now.

Decision: the client-test performance hypothesis is confirmed in real CI. Do not merge until we are comfortable that mixing Node / jsdom files does not increase teardown flakes.

PR #16295, opened 2026-08-19
fix(in-app-agent): surface step-limit truncation instead of silent success

LysanderKie:
🟡 **Live preview: [pr-16295.preview.langfuse.com](https://pr-16295.preview.langfuse.com)**

## Summary
- When the assistant hits the 20-step cap mid-loop, the run no longer looks like a clean success. Postgres records `SUCCEEDED` with `error_code = step_limit`, the activity header reads **Stopped after …**, and the chat shows a warning to send another message.
- Wrap-up guidance is injected on the last allowed model call via Mastra `feedback`, so the model tries to answer instead of calling another tool. Max steps live in shared tunables so wrap-up, detection, and Mastra `maxSteps` cannot drift.
- A follow-up that fails or cancels before the first assistant token no longer relabels an earlier **Worked for …** as **Failed after …** / **Stopped after …**. The notice band still explains that run.
- Visible window states (step-limit notice, previous turn still **Worked for …**, failed-after timeout, failed-before-first-token) are covered by Storybook plays. Client tests stay on the provider and session controller.

## Impacted packages
- `packages/shared` — step-limit tunable and `step_limit` error code
- `worker` — Mastra iteration counting, wrap-up feedback, persistence
- `web` — notice, activity copy, latest-turn activity stamping, Storybook coverage

## Test plan
- [ ] Start a long assistant turn that would previously stop mid-answer at 20 model calls; confirm a yellow notice appears, the activity header says **Stopped after …**, and the run row has `error_code = step_limit`.
- [ ] Confirm the last generation received wrap-up guidance (tag `<step_limit_wrap_up>`) rather than another tool call.
- [ ] Send a follow-up message and confirm the next turn can finish normally.
- [ ] Confirm a normal short turn still finishes with `error_code` null, **Worked for …**, and no notice.
- [ ] Confirm a timed-out run still shows **Failed after …**.
- [ ] After a completed turn, start a follow-up that fails before any assistant output; confirm the previous turn stays **Worked for …** and the notice band explains the failure.

## Verification
- `pnpm --filter web run test-client InAppAgentWindow.clienttest.tsx backgroundExecutionSession.clienttest.ts` — `Tests 39 passed (39)`
- `pnpm --filter worker run test agent.test.ts` — `Tests 33 passed (33)`
- Pre-commit `turbo run lint` — `Tasks: 7 successful, 7 total`
- `pnpm --filter web run test-storybook` skipped locally (Playwright Chromium not installed); CI should run the Step Limit, Failed, and Failed Before First Token plays


worker/src/features/in-app-agent/runtime/agent.ts
```diff
@@ -262,20 +311,29 @@ export async function createAgUiStream(params: {
...
+      stepLimitState.lastFinishReason = finishReason;
+      if (stepLimitState.modelCallCount === IN_APP_AGENT_MAX_STEPS - 1) {
+        stepLimitState.wrapUp = true;
       }
-    : undefined;
+    }
+    recordInstrumentation("recordModelStreamPart", (instrumentation) =>
+      instrumentation.recordModelStreamPart?.(part),
+    );
+  };
```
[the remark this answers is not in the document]
LysanderKie: Good catch — retries were inflating that counter past Mastra's `maxSteps`. Dropped the shadow `doStream` count and now use `defaultOptions.onIterationComplete` so wrap-up/truncation follow Mastra's logical steps. See 3d1eeb5.

LysanderKie:
@claude review

web/src/features/in-app-agent/components/InAppAgentBackgroundHint.tsx
```diff
@@ -16,43 +16,36 @@ export function InAppAgentBackgroundHint({
...
+              navigator.userAgent.includes("Mac")
+                ? "⌘"
+                : "Ctrl",
+              "I",
+            ]}
+          />
+        </Button>
+      }
+    >
+      I keep running in the background. Feel free to minimize, I&apos;ll notify
```
LysanderKie, line 47: We implicitly dropped the keyboard part which I like
<KeyboardShortcut
              className="bg-transparent shadow-none"
              keys={[
                typeof navigator !== "undefined" &&
                navigator.userAgent.includes("Mac")
                  ? "⌘"
                  : "Ctrl",
                "I",
              ]}

worker/src/features/in-app-agent/runtime/agent.stepLimit.test.ts
```diff
@@ -0,0 +1,89 @@
+import { Agent } from "@mastra/core/agent";
```
LysanderKie: This was more for local tests, lets remove that

web/src/features/in-app-agent/lib/backgroundExecutionSession.clienttest.ts
```diff
@@ -1364,6 +1365,35 @@ describe("InAppAgentBackgroundClient reconnect", () => {
   });
 });
 
+describe("getBackgroundRunNotice", () => {
```
LysanderKie: Same here, do we really need that test?

LysanderKie: Should be covered by storybook

web/src/features/in-app-agent/components/InAppAgentBackgroundHint.tsx
LysanderKie: Kept — it just moved into the `action` slot on `InAppAgentNotice`, so it still renders next to the minimize control with ⌘/Ctrl+I.

worker/src/features/in-app-agent/runtime/agent.stepLimit.test.ts
LysanderKie: Removed in 62eab7209. Wrap-up is still covered in `agent.test.ts` via `onIterationComplete` returning the `<step_limit_wrap_up>` feedback.

web/src/features/in-app-agent/lib/backgroundExecutionSession.clienttest.ts
LysanderKie: Agreed — dropped those client tests in b020ccd and moved the visible notice / Stopped after / Failed after coverage into the Step Limit and Failed stories.

web/src/features/in-app-agent/components/InAppAgentWindow.tsx
```diff
@@ -1018,6 +1045,9 @@ export function InAppAgentWindow(props: InAppAgentWindowProps) {
     () => buildConversationDisplayItems(messages, isRunUnsettled),
     [messages, isRunUnsettled],
   );
+  const lastSettledActivityIndex = displayItems.findLastIndex(
+    (item) => item.type === "activity" && !item.isInProgress,
+  );
```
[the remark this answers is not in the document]
LysanderKie: Good catch. `lastSettledActivityIndex` now only matches an activity group after the latest user message, so a follow-up that fails or cancels before the first assistant token keeps the previous turn as **Worked for …**. Covered by the Failed Before First Token story in 189ed51.

PR #16312, opened 2026-08-19
fix(rbac): let project members view the API keys page

### LINKED ISSUE #16285: bug: Project members with apiKeys:read cannot view API keys page
### Describe the bug

Users with the project MEMBER role suffers "Access Denied - You do not have permission to view API keys for this project" on the Project Settings → API Keys page, even though:
- The RBAC definition grants MEMBER the apiKeys:read scope ([projectAccessRights.ts#L210](https://github.com/langfuse/langfuse/blob/v4.13.0/packages/shared/src/features/rbac/projectAccessRights.ts#L210))
- Its LIST api, projectApiKeys.byProjectId only requires apiKeys:read ([projectApiKeyRouter.ts#L24](https://github.com/langfuse/langfuse/blob/v4.13.0/web/src/features/public-api/server/projectApiKeyRouter.ts#L24))

Seems like a frontend bug - ApiKeyList checks apiKeys:CUD ([ApiKeyList.tsx#L48-L51](https://github.com/langfuse/langfuse/blob/v4.13.0/web/src/features/public-api/components/ApiKeyList.tsx#L48-L51)]).

Expected behavior: members can view the key list, while create/delete/note-editing remain hidden behind apiKeys:CUD - which is already how those controls are individually gated.

### Steps to reproduce

1. In an organization, add a user to a project with project role MEMBER
 (org role MEMBER or NONE).
2. As that user, open Project Settings → API Keys.
3. "Access Denied" is shown and the list query is never issued, although
 calling projectApiKeys.byProjectId directly succeeds for the same user.

Users whose effective project role resolves to ADMIN/OWNER (e.g. via org role fallback) are unaffected, since those roles hold apiKeys:CUD.

Self-hosted

v3.172.1, but may also reproducible on the latest 4.13.0 and main

-

### Additional information

A one-line change gating the project list view on apiKeys:read (write actions untouched), plus a server test pinning that a MEMBER caller can list keys via projectApiKeys.byProjectId.

+) If it gets fixed, could it be possible to be applied to 3.2xx either?

Yes

Steffen911:
🟡 **Live preview: [pr-16312.preview.langfuse.com](<https://pr-16312.preview.langfuse.com>)**

Fixes langfuse/langfuse#16285

## Problem

A user with the project `MEMBER` role sees **Access Denied — You do not have permission to view API keys for this project** on Project Settings → API Keys, even though:

* the RBAC definition grants `MEMBER` the `apiKeys:read` scope (`packages/shared/src/features/rbac/projectAccessRights.ts`), and
* the list query `projectApiKeys.byProjectId` only requires `apiKeys:read`.

`ApiKeyList` resolved a single access flag from `apiKeys:CUD` and used it for both the page render and the list query's `enabled` flag, so members were walled off and the query never fired.

## Fix

Gate the project list view on `apiKeys:read`. Write actions are untouched — `CreateApiKeyButton`, `DeleteApiKeyButton`, and `ApiKeyNote` each already check `apiKeys:CUD` themselves, so a member gets a read-only table with no create button, no delete button, and a plain-text note.

Organization keys are unchanged: there is only one org scope (`organization:CRUD_apiKeys`), so that branch still uses it.

Roles unaffected: `OWNER`/`ADMIN` hold both scopes; `VIEWER` and `NONE` hold neither and still get Access Denied.

## Tests

* `web/src/features/public-api/components/ApiKeyList.clienttest.tsx` (new) renders the component against the real RBAC map for three project roles: `MEMBER` sees the list without write controls, `ADMIN` sees the create button, `VIEWER` still gets Access Denied. The MEMBER case fails on `main` with exactly the reported "Access Denied".
* `web/src/__tests__/server/project-api-key-trpc.servertest.ts` gains a pin that a `MEMBER` caller can list keys through `projectApiKeys.byProjectId` — the server contract the UI now relies on. This one passes on `main` too; it is there so a future scope tightening on the router cannot silently re-break the page.

## Verification

Local dev stack, seeded via `pnpm run db:seed`, with `member@langfuse.com`'s project role set to Member through the Members settings UI, then signed in as that user:

* Before (with only the source change reverted, same session and same user): the page shows the "Access Denied" alert.
* After: the page shows **Project API Keys** with the seeded key row (created date, note `seeded key`, public key, masked secret) and no create or delete controls.

The click path below reproduces both states on the preview.

Checks:

* `pnpm --filter web run test-client` → `Test Files  287 passed | 1 skipped (288)`, `Tests  3356 passed | 51 skipped (3407)`
* `pnpm --filter web run test src/__tests__/server/project-api-key-trpc.servertest.ts` → `Tests  6 passed (6)`
* `pnpm run lint` → clean; pre-commit `turbo run lint` → `Tasks: 7 successful, 7 total`
* `pnpm --filter web run typecheck` → clean

## Test steps on the preview

1. Open `pr-<N>.preview.langfuse.com`, create/pick a project, and invite a second user with **project role Member** (Project Settings → Members).
2. Sign in as that user and open Project Settings → API Keys.
3. The key list renders; there is no "Create new API keys" button, no delete icon, and the note is not editable.

## Follow-up

The reporter asks whether this can also land on a 3.2xx release. The fix is a one-file frontend change with no schema or migration surface, so a v3 backport is cheap if we want one — happy to open it as a separate PR.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

This PR aligns the project API-key page with the existing read permission while preserving separate write authorization.

* Gates project API-key listing and its query on `apiKeys:read`.
* Adds client coverage for MEMBER, ADMIN, and VIEWER behavior.
* Adds server coverage confirming MEMBER callers can list project API keys.

</details>

The PR appears safe to merge because it aligns the UI with the existing read authorization while retaining independent write controls.

The newly reachable MEMBER path returns only the already-authorized API-key listing, while create, delete, and note-edit actions remain gated by `apiKeys:CUD`.

</details>

Reviews (1): Last reviewed commit: ["fix(rbac): let project members view the ..."](<https://github.com/langfuse/langfuse/commit/4fb76d51939e32ed86bd581031ce38fc82db882f>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=54779956>)

### DIFF
--- a/web/src/__tests__/server/project-api-key-trpc.servertest.ts
+++ b/web/src/__tests__/server/project-api-key-trpc.servertest.ts
@@ -28,7 +28,7 @@ describe("project API keys trpc", () => {
   async function createProjectCaller(
     projectRole: "ADMIN" | "MEMBER" = "ADMIN",
   ) {
-    const { projectId, orgId } = await createOrgProjectAndApiKey();
+    const { projectId, orgId, publicKey } = await createOrgProjectAndApiKey();
 
     const session: Session = {
       expires: "1",
@@ -76,7 +76,7 @@ describe("project API keys trpc", () => {
     const ctx = createInnerTRPCContext({ session, headers: {} });
     const caller = appRouter.createCaller({ ...ctx, prisma });
 
-    return { caller, projectId };
+    return { caller, projectId, publicKey };
   }
 
   describe("projectApiKeys.byProjectId", () => {
@@ -97,6 +97,17 @@ describe("project API keys trpc", () => {
       expect(apiKeys.map((key) => key.note)).not.toContain(
         "In-app agent key hidden from project UI",
       );
+    });
+
+    // The settings page gates the list view on apiKeys:read, which MEMBERs
+    // hold without apiKeys:CUD. Pin that this read path stays open to them.
+    it("lists keys for MEMBER callers, who only hold apiKeys:read", async () => {
+      const { caller, projectId, publicKey } =
+        await createProjectCaller("MEMBER");
+
+      const apiKeys = await caller.projectApiKeys.byProjectId({ projectId });
+
+      expect(apiKeys.map((key) => key.publicKey)).toContain(publicKey);
     });
   });
 
--- a/web/src/features/public-api/components/ApiKeyList.clienttest.tsx
+++ b/web/src/features/public-api/components/ApiKeyList.clienttest.tsx
@@ -1,0 +1,116 @@
+import { render, screen } from "@testing-library/react";
+import { type Role } from "@langfuse/shared/src/db";
+
+const { projectApiKeys, mockSession } = vi.hoisted(() => ({
+  projectApiKeys: [
+    {
+      id: "key-1",
+      createdAt: new Date("2026-01-02T03:04:05.000Z"),
+      expiresAt: null,
+      lastUsedAt: null,
+      note: "Production key",
+      publicKey: "pk-lf-1234",
+      displaySecretKey: "sk-lf-...5678",
+      createdByUser: null,
+      createdByApiKey: null,
+    },
+  ],
+  mockSession: { data: null as unknown },
+}));
+
+vi.mock("next-auth/react", () => ({
+  useSession: () => mockSession,
+}));
+
+vi.mock("@/src/utils/api", () => {
+  const noopMutation = () => ({
+    mutate: vi.fn(),
+    mutateAsync: vi.fn(),
+    isPending: false,
+  });
+  return {
+    api: {
+      projectApiKeys: {
+        byProjectId: {
+          useQuery: (
+            _input: unknown,
+            options?: { enabled?: boolean },
+          ): { data?: typeof projectApiKeys } =>
+            options?.enabled ? { data: projectApiKeys } : {},
+        },
+        create: { useMutation: noopMutation },
+        updateNote: { useMutation: noopMutation },
+        delete: { useMutation: noopMutation },
+      },
+      organizationApiKeys: {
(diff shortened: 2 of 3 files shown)


Steffen911:
What I would doubt in this PR:

1. **Is a read-only key list actually the intended privacy posture for members?** The fix trusts the RBAC table — `MEMBER` has `apiKeys:read`, so the page should render. But that means every project member can now see the public keys, the masked secret suffixes, the notes, and who created each key. The list leaks no usable secret, and the server has always served this data to members, so this only aligns the UI with the API. Still, if the product intent was "members should not see keys at all", the correct fix is the opposite one: drop `apiKeys:read` from `MEMBER` and keep the CUD gate. Worth a deliberate call rather than inheriting mine.

2. **The server test does not fail on `main`.** Only the new client test reproduces the bug; the server pin passes before and after. I kept it because the page now depends on that router scope, but if you consider non-failing tests noise, it is the one thing to cut.

3. **The client test mocks `next-auth/react` and the tRPC client but uses the real `useHasProjectAccess`.** That is deliberate — it exercises the actual role→scope map, which is where the bug lived — but it means the test file carries a fair amount of mock surface for one component. If the house preference is thinner component tests, say so and I will trim it to the MEMBER case only.

4. **Only the project branch changed.** The organization branch still uses `organization:CRUD_apiKeys` for both view and write because there is no org-level read scope. If org viewers should be able to see org keys, that is a separate RBAC addition, not part of this fix.

onwn:
@Steffen911 Thank you for handling this!
BTW, could this bugfix be backported to further v3 releases either? 🙏

Steffen911:
@onwn Already on it!

PR #16316, opened 2026-08-19
feat(v4-migration): show the migration UI by default on dual write mode

Steffen911:
🟡 **Live preview: [pr-16316.preview.langfuse.com](https://pr-16316.preview.langfuse.com)**

## What does this PR do?

Shows the v4 migration UI to self-hosters on `dual` write mode instead of only to Langfuse/ClickHouse employees.

The migration surfaces — sidebar "Action required" pill, per-project chips and banner on the organization overview, the migration panel, and the `/v4-migration` status page — all hang off one flag, `featureFlags.v4UpgradeUi`. That flag defaulted on only for `@langfuse.com`/`@clickhouse.com` emails, and self-hosters could not reach it at all: `userAccount.setFeaturePreviewEnabled` rejected the enable unless the user was already on the v4 read path, so the Feature Preview modal rendered a "V4 Migration" tile that threw `PRECONDITION_FAILED` on click.

Availability now follows the deployment rather than the user:

| Deployment | Migration UI |
| --- | --- |
| Cloud (`dual`, `events_only`) | on |
| Self-hosted `dual` + `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true` | on |
| Self-hosted `dual` without the opt-in | off |
| Self-hosted `events_only` | off |
| Any `legacy` | off |

Self-hosted `dual` without the opt-in stays off because nobody on that deployment can move onto the v4 read path, so every call to action in the panel would be a dead end. `events_only` self-hosted stays off because the migration is already over there — the legacy public API routes 404, legacy evaluators are hidden, and the legacy analytics integrations are no-ops, so no readiness check has anything to report. `legacy` stays off because the events tables the surfaces read are not written.

No new env variable: write mode plus the existing preview opt-in already carry the signal, and a third knob would let operators produce incoherent combinations.

### How

- New pure predicate `isV4UpgradeUiAvailable` in `web/src/features/events/lib/v4Rollout.ts`, next to `canToggleV4`.
- Derived once in the `auth.ts` session callback alongside `v4BetaEnabled`/`canToggleV4`, then fed into `parseFlags` as availability context. `featureFlags.v4UpgradeUi` stays the single gate, so no surface and no nav filter changed.
- `userAccount.setFeaturePreviewEnabled` mirrors the same predicate, so an opted-out self-hoster can opt back in, and disabling a default-on preview now persists an opt-out marker instead of silently springing back on.

Two adjacent fixes came along because the change would otherwise ship a known-broken affordance:

- The Feature Preview modal now **explains** an ineligible deployment (disabled toggle + reason) instead of failing the mutation.
- Preview availability became a hard gate in `parseFlags`: a stale opt-in entry can no longer grant a preview whose read path the deployment does not support. This also closes the same hole for `modernSession`, which the modal already told users required the events-backed session view.
- Unrelated but blocking: the `it.each` cases in `userAccount.servertest.ts` hardcoded an email, so a second run against the same database failed on the `users.email` unique constraint. Now unique per run.

Per-project exceptions are unchanged — forced-v3 projects are still excluded downstream via `LANGFUSE_FORCE_V3_EXPERIENCE`, both in `useV4UpgradeUiEnabled` and in `getProjectMigrationReadiness` (`partner-managed`), so the vapi carve-out holds.

## Type of change

## Mandatory Tasks

## How to test

Preview: **pr-16316.preview.langfuse.com** — note previews build with `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION=DEV`, so they exercise the **Cloud** row of the table, not the self-hosted one. The preview confirms Cloud is on by default; the self-hosted path needs a local run.

Cloud row (on the preview):
1. Sign in, land on the organization overview at `/`.
2. Expect the v4 migration banner and per-project "Update" chips for projects with pending work, and the "Action required" pill in a project's sidebar.
3. Open the user menu -> Feature Previews. "V4 Migration" is **on** and togglable; turning it off hides every surface and restores the sidebar "V4 Preview" switch; turning it back on restores them.

Self-hosted row (locally):
1. Run with `LANGFUSE_MIGRATION_V4_WRITE_MODE=dual`, `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true`, and **no** `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION`.
2. Sign in as a non-`@langfuse.com`/`@clickhouse.com` user. Same expectations as above.
3. Drop `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN` and restart: every migration surface disappears and the Feature Previews tile explains why instead of erroring.

## Verification

- `pnpm --filter web run typecheck` — clean, 0 errors
- Client tests over `v4-migration`, `feature-flags`, `feature-previews`, `events/lib` — `Tests  187 passed (187)`
- `pnpm --filter web run test userAccount.servertest.ts` — `Tests  10 passed (10)`, and passes on a second consecutive run against the same database (the isolation fix above)
- The four new `parseFlags` behavior tests were confirmed red against the pre-change source before the fix landed.

Not done: no authenticated browser walkthrough. I did boot the app in self-hosted `dual` mode with the region unset and confirmed `/`, `/auth/sign-in` and `/v4-migration` all return 200 with no server error, but I did not create an account or enter a password to sign in, so the click-path above is unverified by me.

Linear Issue: [LFE-14647](https://linear.app/clickhouse/issue/LFE-14647/enable-the-v4-migration-ui-by-default-for-self-hoster)

The PR makes the v4 migration UI default to deployment-level availability and keeps preview persistence, session state, and mutation eligibility aligned.
- Adds a shared rollout predicate for Cloud and eligible self-hosted dual-write deployments.
- Hard-gates feature previews by deployment availability while preserving explicit opt-outs.
- Exposes availability through authenticated sessions and explains unavailable deployments in the preview modal.
- Adds rollout, flag-parsing, hook, and server mutation coverage.
</details>

The PR appears safe to merge, with deployment availability, session flags, persistence behavior, and migration actions remaining consistent across the supported rollout modes.

The changed predicate is applied consistently to session parsing and preview mutations, authenticated sessions receive the new availability field, and investigation found no reachable broken migration action or preview-state regression.
</details>

```mermaid
flowchart TD
  Config[Deployment write mode and preview opt-in] --> Predicate[isV4UpgradeUiAvailable]
  Predicate --> Auth[NextAuth session callback]
  Predicate --> Mutation[userAccount preview mutation]
  Auth --> Parse[parseFlags availability gate]
  Parse --> Flag[featureFlags.v4UpgradeUi]
  Auth --> Modal[Feature Preview modal availability]
  Flag --> Surfaces[Migration page, panel, banner, chips, and sidebar]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(v4-migration): show the migration U..."](https://github.com/langfuse/langfuse/commit/6104efa850522d4a45ebe233abb510f18597e7b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54794858)</sub>

- Knowledge Base — [Authentication, Multi-Tenancy, and RBAC](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/auth-rbac.md)
- Knowledge Base — [Enterprise Edition (EE), Entitlements, and Billing](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/ee-features.md)
</details>


Steffen911:
Things worth doubting here, in rough order of how much I'd want a second opinion:

**`events_only` self-hosted is off, which contradicts the issue description.** I followed the Aug-17 thread on the ticket (nothing actionable there: legacy API routes 404, legacy evaluators hidden, legacy integrations no-op) rather than the original matrix. The cost: an operator who flips straight to `events_only` before migrating gets 404s with no in-app explanation. If you'd rather cover that, the alternative is `isLangfuseCloud || v4WriteMode !== "legacy"` — in practice most readiness checks would read zero there anyway, so the visible difference is mostly the banner and status page.

**Cloud `events_only` is on, and I'm least sure about that line.** I kept it so the guidance does not vanish the moment a region flips. But the "nothing actionable" argument applies to Cloud just as much, so this may want to be off too.

**Load on the organization overview.** Cloud users all get the UI now, not just employees and opt-ins, so `useAccountV4MigrationData` fires four tRPC queries per org (plus `lastTraceByProject`) for everyone landing on `/`. Redis-backed with `staleTime` 5min and no refetch-on-focus, but it is a real step up in traffic on the busiest page.

**Two behavior changes that reach past the ticket.** Employees on self-hosted `dual` *without* `ALLOW_PREVIEW_OPT_IN` now lose the migration UI (they had it via the email default) — deliberate, but a regression for them. And making availability a hard gate in `parseFlags` also means a stale `modernSession` opt-in entry no longer grants that preview when the events read path is off. I think both are correct, but neither was asked for.

**Not verified by me:** no authenticated browser walkthrough. I booted the app in self-hosted `dual` with the region unset and confirmed the pages return 200 with no server error, but did not sign in, so the click-path in the description is untested. The `auth.ts` session-callback wiring is the one link with no automated coverage — the predicate and `parseFlags` on either side of it are both unit-tested.

Steffen911:
Review follow-ups, all in [`a49bb647bd`](https://github.com/langfuse/langfuse/commit/a49bb647bd0cfd2805505c94c3f2090ddf81569f).

**Fixed — `setFeaturePreviewEnabled` bypass.** Correct, and the reasoning was exact. On self-hosted `events_only`, `v4BetaEnabled` is forced on while the migration UI is deliberately off, so the generic bypass let an enable write a raw `v4UpgradeUi` entry that `parseFlags` then ignored forever. The migration UI is now gated solely on `v4UpgradeUiAvailable`; other previews keep the old precondition untouched. Added a servertest for the events_only reject that also asserts nothing was written — verified red against the previous precondition before the fix.

**Fixed — `warningReason` was wrong for `events_only`.** Also correct, and worse than a copy nit: following that advice would have resumed the legacy writes an already-migrated deployment stopped. The reason now branches on the write mode — `events_only` says the migration is complete, `legacy` names the write mode, and only the `dual` case mentions `ALLOW_PREVIEW_OPT_IN`.

**Acknowledged, not fixable — pre-marker opt-outs.** Agreed on both the mechanism and that there is nothing to key a backfill on: a plain removal is byte-identical to never having touched the toggle, and flipping the default for that empty state is the whole point of the change. Documented on `featurePreviewDefaultsToEnabled` so it reads as a deliberate tradeoff. Doubt this if you think the affected population is larger than "non-team users who enabled and then disabled the tile during the pre-PR window" — that is the assumption I am leaning on.

**Separately, from Steffen:** the migration UI hard-coded `November 16, 2026` in four places, and every one of them is now visible to self-hosters — where the statement is simply false. Nothing happens to a self-hosted deployment on that date; its legacy surfaces keep working until the operator moves the write mode off `dual`. All four now read the operator cutoff instead, behind one shared `useHasV4MigrationDeadline` (Cloud keeps the dated copy and the timeline link):

- the panel/status-page header note → "Some features may stop working without an upgrade once your administrator disables the legacy mode."
- FAQ "Do I have to do this?" → "…frozen and stop working once your administrator disables the legacy mode."
- FAQ "What if I do nothing?" → "Once your administrator disables the legacy mode, old SDKs stop sending data…"
- the switch-back-to-v3 warning → "…will be sunset once your administrator disables the legacy mode."

This follows the convention `deprecations.ts` already documents and `openapi-deprecations.servertest.ts` already enforces for the Fern messages: the date is a Cloud commitment, self-hosted is not bound by it.

**Two remaining date exposures I did not touch** (different surface, own verification requirements — happy to do them as a separate PR):

1. `web/src/features/public-api/server/deprecations.ts` — the runtime `_deprecation` key on legacy public API responses carries `sunsetAt: "2026-11-16"` and a message with the date. It is attached unconditionally, so self-hosters get it. `V3_NOTICE` says "On Langfuse Cloud, …" but, unlike the Fern copy, it omits the "Self-hosted deployments are unaffected by this date" clause that `openapi-deprecations.servertest.ts` enforces on the Fern side. Worth closing that gap plus a test.
2. `web/public/generated/api/openapi.yml` — served from the self-hoster's own instance, so their API reference shows the date. This one already carries the self-hosted clarifier, so it is correct as-is.

Verification: `pnpm run lint` `Tasks: 7 successful, 7 total`; `pnpm --filter web run typecheck` clean; client sweep over `v4-migration`/`feature-flags`/`feature-previews`/`events/lib` `Tests 190 passed (190)`; `userAccount.servertest.ts` `Tests 11 passed (11)`. Still no authenticated browser pass from me.

Steffen911:
Correction to my previous comment: **item 1 of the "remaining date exposures" is wrong — ignore it.** The `_deprecation` signal on legacy public API responses *is* Cloud-gated, in `createAuthedProjectAPIRoute.ts`:

```ts
// Cloud-only: the sunset date binds Cloud, not self-hosted deployments.
const deprecation = env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION
  ? routeConfig.deprecation
  : undefined;
```

`attachDeprecation` early-returns on a falsy `deprecation`, so a self-hosted deployment never stamps the key at all. I checked the two `attachDeprecation` call sites and concluded there was no gate without tracing where `deprecation` was bound. That also makes the "missing self-hosted clause in `V3_NOTICE`" a non-issue — self-hosters never receive that message — so there is no follow-up PR to do there.

Item 2 also holds up better than I described. `web/public/generated/api/openapi.yml` is served statically from a self-hoster's own instance, so the date is visible there — but every one of its 15 date occurrences is followed by "Self-hosted deployments are unaffected by this date; the endpoint becomes unavailable when they upgrade to Langfuse v4." (My earlier line-based count suggested 11 of 15; that was a YAML line-wrapping artifact in the grep, not a gap.)

So after a full re-audit of every `November 16, 2026` / `2026-11-16` occurrence in the repo, the complete picture is:

| Where | Self-hoster sees it? |
| --- | --- |
| 4 upgrade-UI sites | was yes — fixed in this PR |
| `web/public/generated/api/openapi.yml` (15) | yes, and each one says self-hosted is unaffected — correct as-is |
| `_deprecation` on API responses | no, Cloud-gated at the route |
| `fern/apis/**` definitions | no, source only |
| `*.clienttest.tsx` | no, test files |

Nothing outstanding. Thanks to Steffen for pushing back on the gating claim.

Steffen911:
@claude review

web/src/features/feature-flags/utils.ts
```diff
@@ -27,28 +27,55 @@ export const receivesFeaturePreviewsByDefault = (
...
+ */
+export const featurePreviewDefaultsToEnabled = (
+  flag: FeaturePreviewFlag,
+  email: string | null | undefined,
+): boolean =>
+  receivesFeaturePreviewsByDefault(email) ||
+  // The v4 migration UI is not a preview users have to discover: every
+  // deployment that can act on the migration gets it by default, and the
+  // Feature Preview modal only exists to opt back out.
+  flag === "v4UpgradeUi";
```
Steffen911: Can we remove the feature flag? I think by now we rolled it out across all cloud users and for self-hosted users it should apply instance-wide based on the dual write mode + opt in flag. I think we could simply remove logic for the feature flag?

web/src/features/v4-migration/V4MigrationContent.tsx
```diff
@@ -1053,6 +1070,17 @@ export function V4MigrationDocsLink() {
 }
 
 export function V4MigrationDeadlineNote() {
+  const hasDeadline = useHasV4MigrationDeadline();
+
+  if (!hasDeadline) {
+    return (
+      <p>
+        Some features may stop working without an upgrade{" "}
+        {V4_MIGRATION_SELF_HOSTED_CUTOFF}.
```
Steffen911: Why don't we inline the wording here? What's the advantage of keeping the string separate?

web/src/features/feature-flags/utils.ts
Steffen911: Done in [`b9f1a6ab0d`](https://github.com/langfuse/langfuse/commit/b9f1a6ab0d1ef6aed7c0a9c6d75520a58390afe1) — you were right, and it made the PR smaller rather than bigger.

`feature-flags/utils.ts`, its clienttest, and `userAccount.ts` are byte-identical to main again, so the `featurePreviewDefaultsToEnabled` helper, the `setFeaturePreviewEnabled` carve-out, and the hard availability gate in `parseFlags` are all gone. `session.user.v4UpgradeUiAvailable` is the only gate now — which is also closer to the issue's "set this via env variables" than what I had. Two things the bot review raised vanish with it: the pre-marker opt-out edge case and the per-write-mode warning copy in the modal.

Removed: `v4UpgradeUi` from `featurePreviewFlags` and `Flags`, the modal tile, the preview registry entry, and the story entry. The `/assets/v4-beta-intro.jpg` illustration stays — `V4IntroDialog` still uses it.

One thing worth a look: the sidebar entry can no longer use `featureFlag`, so it gates through `show({ v4UpgradeUiAvailable })` (one new field on the `show` contract, alongside the existing `isLangfuseCloud`/`v4WriteMode`). I did it that way rather than letting `V4MigrationNavItem` self-gate because dropping the route filter would mount the component on every project page and fire its `migrationActions` query even on ineligible deployments.

Net effect on the branch: 255 insertions / 114 deletions, down from 518 / 35.

Consequence you're choosing, just so it's on the record: there is no per-user off switch any more. On an eligible deployment the pill, chips, banner and status page are unconditional, and existing `feature-preview:v4UpgradeUi:disabled` markers go inert — so anyone who deliberately opted out gets it back for good. The operator levers all survive (`ALLOW_PREVIEW_OPT_IN`, write mode, `LANGFUSE_FORCE_V3_EXPERIENCE` per project), and the banner still self-dismisses for 7 days.

web/src/features/v4-migration/V4MigrationContent.tsx
Steffen911: Inlined at all four sites in [`b9f1a6ab0d`](https://github.com/langfuse/langfuse/commit/b9f1a6ab0d1ef6aed7c0a9c6d75520a58390afe1) — no advantage, and it was already leaking.

The constant was used at three of the four sites. The fourth had a hand-copied capitalized version, because that sentence starts with the phrase and a lowercase fragment does not fit:

```tsx
: "Once your administrator disables the legacy mode"}
```

So the single source of truth had already forked silently, which is the usual sign a bare sentence fragment cannot carry four different grammatical positions. The clienttests assert the full sentences verbatim either way, so a wording change is still caught by tests. Inlining also matches how `V4MigrationStatusPage.tsx` already works — every other FAQ string lives locally there rather than in the shared copy module.

Steffen911:
@claude review

nimarb (approved):

PR #16352, opened 2026-08-20
fix(trace): compact timeline, first round of review feedback

nkabardin:
🟡 **Live preview: [pr-16352.preview.langfuse.com](https://pr-16352.preview.langfuse.com)**

## TL;DR

First round of review feedback on the Compact Timeline preview, from someone who reads traces for a living. Nine items — the one that matters most is that **hover was repainting bars in the colour that means SPAN**, so the palette contradicted itself exactly when you were inspecting a row.

Each item has a story that fails without the fix.

## Why

The feedback split cleanly into three kinds, and the middle kind is the interesting one.

**Things that were wrong.** Hover recoloured the bar to the accent — the same blue as SPAN — so a hovered generation changed species under the cursor. That is the answer to "why is the generation span blue in some traces and pink in others": one screenshot was hovered and the other was not. Hue now carries type and nothing else; focus is the row's wash (already full-width, so it reads at any density) and selection is a ring.

**Things that were right and unreadable.** `@0ms` on a root is correct — it is the offset from the start of the trace, which for a root is zero. It is also noise. And where a row genuinely starts 11m 52s in and runs 11m 52s, two unlabelled durations side by side read as one number repeated. Labelling it was the first fix; testing it locally settled on the better one — the offset is gone, because the bar's position on the axis already says when a span began.

**Controls that lied.** "Fit whole trace" was always enabled and did nothing when the trace already fitted; its diagonal arrows read as "fullscreen", so a spent control looked broken. Expand/collapse-all in the header moved the tree's store and nothing happened here.

## What

- **Hue = type.** Accent no longer paints a bar, which also removes it from the set of colours whose label contrast has to be resolved — taking the "dark label on mid-violet" complaint with it.
- **The tooltip covers the whole row.** It was gated on the pointer being past the gutter, so hovering a name — the obvious way to ask what a span is — said nothing.
- **The tooltip says less, and says it plainly**: type beside its swatch, duration, and cost. No start offset — the axis carries that. Tokens came only on spans that *have* usage, so the tooltip changed shape row to row; cost means the same thing everywhere and the token split is one click away.
- **Collapse works.** The pure core has always accepted a collapsed set; this renderer never passed one on. At 1px rows, collapsing is how 3,000 rows become 40.
- **Drag draws the zoom box, no modifier.** Scroll already pans both axes, so drag was free, and a modifier nobody guesses is a footnote rather than a gesture. One-finger touch still pans — a finger has no scroll wheel.
- **Fit is disabled when the trace already fits**, and wears a viewfinder.
- **Two controls left the toolbar**: a colour legend (the tooltip names the type beside its swatch, which answers the same question on the row being looked at) and the names toggle (`auto` already decides, and a hover or tap on the rail floats the names when you want a look). The toolbar also stopped captioning the gestures — every one of them is a thing you would have tried, and the space now says where you are, only once you are somewhere other than the whole trace.

## Result

| Item | Story that fails without it |
| --- | --- |
| Hue = type | Hover Does Not Recolour The Bar |
| Tooltip over names | The Tooltip Covers The Whole Row |
| No offset in the tooltip | The Tooltip Does Not Repeat The Axis |
| Collapse honoured | Collapsed Rows Are Not Drawn |
| Drag zooms | Drag Zooms Without A Modifier |
| Touch still pans | A Finger Drag Still Pans |
| Fit disabled | Fit Is Disabled When It Would Do Nothing |
| Cursor matches the gesture | The Cursor Matches The Gesture |
| No caption, no legend, no toggle | The Toolbar Does Not Explain Itself |

<details>
<summary>What the tests had to be taught, and one thing left open</summary>

Two of the new stories passed against the bug before they were sharpened, both for the same reason — the fixture could not express the state under test:

- **Collapse** counted `[data-testid="timeline-dense-content"] > div`, which includes a child that is not a row. Rows carry their own testid now, so a row count is a row count.
- **Touch panning** used a trace that already fitted the box, so there was no window left to pan and "it moved" could never be false. It now pans 3,000 rows in a 600px box, and asserts the row *height* did not change — because a box zoom would have changed it, which is the failure mode the story exists to exclude.

Also worth recording: the accent leaving the bar palette is why the contrast complaint disappears rather than being retuned. The tone picker was not wrong about that violet's luminance; the bar simply should never have been that colour.

Left open deliberately: **a root that starts after its own child.** The `@11m 52s` that prompted the offset item is faithful — the origin is the earliest start of any node, `startOf` has no fallback, and the bar is drawn at the same offset — so that trace has a parent stamped later than its child. Rendering it deadpan is what made the tooltip look wrong; flagging the shape is a separate piece of work, probably a data-quality hint rather than timeline work.

Per-row expand chevrons are also not here: the header's all-or-nothing toggle is what was reported, and that now works.

### Checks

- `Tests 469 passed (469)` — Storybook tests in real Chromium.
- `Tests 488 passed | 48 skipped (536)` — web client tests for `src/features/traces`.
- `Tasks: 7 successful, 7 total` turbo lint; `Tasks: 8 successful, 8 total` turbo typecheck.
- Seven negative controls run one at a time: each fails its own story and nothing else.

</details>

The PR incorporates the first review round for the compact trace timeline, improving type-color consistency, tooltips, collapse handling, viewport controls, and pointer gestures.
- Adds a trace-specific color legend and preserves type hues during hover and selection.
- Passes shared collapsed-node state into timeline preparation.
- Changes mouse and pen dragging to marquee zoom while retaining one-finger touch panning.
- Simplifies tooltip facts and clarifies start offsets.
- Adds Storybook interaction coverage for the revised behavior.
</details>

The pointer-capture regression should be fixed before merging because ordinary mouse or pen clicks can stop reaching the row selection and focus handlers.

Non-touch pointerdown now captures the pointer at the surface before a drag is established, while selection and double-click focus remain exclusively on descendant row click handlers with no surface fallback.

**Files Needing Attention:** web/src/features/traces/components/TraceTimelineDense/TimelineDense.tsx
</details>

```mermaid
flowchart TD
  A[Pointer down on timeline row] --> B{Touch pointer?}
  B -->|Yes| C[Pan or pinch gesture]
  B -->|No| D[Surface captures pointer]
  D --> E[Create marquee]
  E --> F[Pointer up delivered to surface]
  F --> G[Commit marquee]
  F --> H[Release pointer capture]
  H --> I{Synthesized click reaches row?}
  I -->|Yes| J[Select or focus row]
  I -->|No| K[Selection or focus is lost]
```
</details>

`````markdown
### Issue 1
web/src/features/traces/components/TraceTimelineDense/TimelineDense.tsx:946-947
**Pointer capture bypasses row clicks**

In browsers that target synthesized clicks from the captured pointerdown and pointerup events, capturing every non-touch pointer immediately on the surface sends the click or double-click to the surface instead of the descendant row, causing row selection and double-click focus to stop working.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(trace): compact timeline, first roun..."](https://github.com/langfuse/langfuse/commit/e3d3e402c824c7f0b12017ec9177168508821b79) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55105992)</sub>


nkabardin:
@claude review

nkabardin:
@claude review

PR #16356, opened 2026-08-20
fix(automations): strip repository-dispatch token from the read path (v3 backport of #16300)

Steffen911:
v3 backport of #16300 (mainline commit 0d9bbf66d0), cherry-picked clean.

`convertActionToDomain` sanitized webhook configs and returned every other action config verbatim, so a `GITHUB_DISPATCH` action config reached clients with its stored `githubToken`. Every project role holds `automations:read`, so a project VIEWER could read the credential through `automations.getAutomations` and `automations.getAutomation`. The fix replaces the fail-open fallback with a switch that is exhaustive over `ActionTypes`: each action type sanitizes through its own allowlist converter, configs that fail to parse are projected onto the keys their safe schema declares instead of passed through, and the create/update responses share the same single sanitizer. See #16300 for the full write-up.

## Backport verification

- **Patch identity**: the diff of this branch against `origin/v3` is byte-identical to the mainline commit's diff, except a one-line hunk-offset drift in `.agents/skills/code-review/references/review-checklist.md`.
- **Revert check**: with the two production files reverted to `origin/v3` (tests kept), exactly the two new redaction tests fail — `should not expose the stored githubToken to a project VIEWER` and `should not expose the githubToken of a config that fails to parse` — i.e. the leak reproduces on v3 and this change closes it.
- `pnpm --filter web run test automations-trpc.servertest.ts` (v3 schema DB): `Tests  43 passed (43)`
- Worker consumers of the changed `getActionById` read path: `slack-processor.test.ts` `Tests  11 passed (11)`, `webhooks.test.ts` `Tests  22 passed (22)`
- `pnpm tc`: `Tasks:    7 successful, 7 total`
- `pnpm run lint`: `Tasks:    7 successful, 7 total`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary><h3>Greptile Summary</h3></summary>

The PR closes an automation credential disclosure by routing every client-facing action response through one exhaustive, allowlist-based sanitizer.

- Removes stored GitHub repository-dispatch tokens from list, detail, create, and update responses.
- Preserves raw-secret access through the explicitly named execution accessor.
- Fails closed for malformed stored configs by projecting only safe-schema fields.
- Adds VIEWER-level regression coverage for valid and malformed GitHub dispatch configurations.
- Documents the repository’s secret read-path review requirements.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

[...]

PR #16357, opened 2026-08-20
fix(automations): preserve request headers when regenerating a webhook secret (v3 backport of #16315)

Steffen911:
v3 backport of #16315 (mainline commit 59ef87adc5), cherry-picked clean.

**Stacked on #16356** (same file) — merge that one first; GitHub retargets this PR to `v3` automatically when the base branch is deleted. After #16356 squash-merges, the files view here will transiently show both changes until merge; the merge result is still just this commit's change.

`regenerateWebhookSecret` read the action through `getActionById`, which returns the sanitized webhook config, then wrote that object straight back to the database. The sanitizer drops `headers`/`requestHeaders`, so rotating a signing secret permanently deleted the automation's custom request headers from the stored config while `displayHeaders` kept listing them in the UI. The fix reads the raw row and replaces only `secretKey`/`displaySecretKey`, so stored header values carry through still encrypted. See #16315 for the full write-up, including how to identify rows already affected.

## Backport verification

- **Patch identity**: the diff of this branch against the backport of #16300 is byte-identical to the mainline commit's diff, including hunk offsets.
- **Revert check**: with `web/src/features/automations/server/router.ts` reverted to the #16300-only state (tests kept), exactly the new test `should preserve custom request headers when regenerating the secret` fails — the header loss reproduces on v3 and this change closes it.
- `pnpm --filter web run test automations-trpc.servertest.ts` at the stack tip (v3 schema DB): `Tests  44 passed (44)`
- `pnpm tc` at the stack tip: `Tasks:    7 successful, 7 total`
- `pnpm run lint` at the stack tip: `Tasks:    7 successful, 7 total`

This backport fixes webhook secret regeneration so it preserves encrypted custom request headers instead of writing a sanitized configuration back to storage.
- Reads the project-scoped raw webhook action configuration before rotating its signing secret.
- Replaces only `secretKey` and `displaySecretKey`, retaining all other persisted fields.
- Adds coverage for legacy headers, structured request headers, and encrypted-at-rest secret header values.
</details>

The PR appears safe to merge with no actionable regressions identified.

The raw lookup retains existing project scoping and validation while preventing sanitized configuration data from deleting stored webhook headers, and the regression test covers both legacy and structured encrypted header formats.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(automations): preserve request heade..."](https://github.com/langfuse/langfuse/commit/c55d0abac21b058b9743501299a3ab6f587b7c24) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55137385)</sub>

**Context used:**

### DIFF
--- a/.agents/skills/code-review/references/review-checklist.md
+++ b/.agents/skills/code-review/references/review-checklist.md
@@ -59,6 +59,12 @@ This is the canonical shared review checklist for Langfuse.
 - For changes that add a new integration, secret-bearing field, redirect
   follower, or RBAC scope, run the rest of the
   [`security-review/references/checklist.md`](../../security-review/references/checklist.md).
+- For changes that shape what a read route returns — a repository or domain
+  converter, a `Safe*` type, a response projection — a catch-all branch that
+  returns the stored value for unhandled types, or an `as Safe*` assertion
+  standing in place of a sanitizer, is a finding: feature read scopes are held
+  by VIEWER. See
+  [`security-review/references/secret-read-paths.md`](../../security-review/references/secret-read-paths.md).
 
 ## JavaScript / TypeScript Style
 
--- a/.agents/skills/security-review/SKILL.md
+++ b/.agents/skills/security-review/SKILL.md
@@ -38,6 +38,7 @@ correct validation surfaces land in the plan, not in a follow-up CVE.
 | Topic | Open when | File |
 | --- | --- | --- |
 | SSRF and outbound URL validation | The change accepts or fetches a user-supplied URL, host, or endpoint | [references/outbound-url-validation.md](references/outbound-url-validation.md) |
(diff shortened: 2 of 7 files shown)


Steffen911:
What a reviewer should doubt here:

- **The stricter guard.** `isSafeWebhookActionConfig` → `isWebhookActionConfig` means a webhook row **missing** `secretKey` now errors on regenerate instead of silently receiving a fresh secret. Mainline decided surfacing the broken row is right (the worker can't execute such a row either), but v3 has an older row population — if such rows exist there, regenerate turns into a visible error for them.
- **Stacked diff.** Base is the #16356 branch; after that one squash-merges, this PR's files view will transiently show both commits until GitHub retargets/merges. The merge result is only this commit's change.

PR #16363, opened 2026-08-20
fix(docker): bound self-hosted Compose logs

### LINKED ISSUE #16339: bug: unbounded ClickHouse Docker log can exhaust self-hosted server after ENOSPC
### Describe the bug

The default self-hosted Docker Compose configuration does not bound the ClickHouse container logs. With Docker's default `json-file` logging driver, stdout/stderr can grow without rotation.

On a self-hosted instance, the host filesystem reached `ENOSPC`. ClickHouse could no longer write or rotate its internal files under `/var/log/clickhouse-server` and repeatedly emitted the following failure to stderr:

```text
Cannot log message in OwnAsyncSplitChannel channel:
Poco::Exception. Code: 1000, File access error:
/var/log/clickhouse-server/clickhouse-server.log
Poco::RotateBySizeStrategy::mustRotate(...)
```

Docker captured this output in the container's unbounded `*-json.log`. That file reached **89.1 GiB**, the 244 GiB root filesystem reached **100% / 0 bytes available**, and even `docker exec` / `docker run` began failing with `no space left on device`.

Expected: the bundled self-hosting path should either bound container logs or prominently require host-level Docker log rotation, so a ClickHouse logging failure cannot consume the entire host filesystem.

This is related to, but distinct from, #13123:

- #13123 concerns ClickHouse system tables stored inside `/var/lib/clickhouse`.
- This report concerns the outer Docker `json-file` log under `/var/lib/docker/containers` and the feedback loop after the filesystem fills.

Both occurred on this instance: ClickHouse `system` tables occupied **22.87 GiB**, while the Langfuse `default` database occupied **487.20 MiB**.

### Steps to reproduce

1. Deploy Langfuse using the default self-hosted `docker-compose.yml` and Docker's default `json-file` logging driver.
2. Do not configure daemon-level or per-service log rotation.
3. On a disposable test host, exhaust the filesystem until ClickHouse cannot write or rotate `/var/log/clickhouse-server/clickhouse-server.log`.
4. Observe repeated `OwnAsyncSplitChannel` / `RotateBySizeStrategy::mustRotate` errors on ClickHouse stderr.
5. Observe the ClickHouse container's `*-json.log` growing without a configured limit.

The mitigation validated on the affected deployment was:

```yaml
clickhouse:
  logging:
    driver: json-file
    options:
      max-size: "10m"
      max-file: "3"
```

After stopping and recreating only the ClickHouse container while retaining both named volumes:

- 81 GiB became available on the root filesystem;
- ClickHouse returned healthy with the existing 24 GiB data volume;
- the repeated stderr failure disappeared;
- Docker reported `max-size=10m` and `max-file=3` for the recreated container;
- the new JSON log remained at 859 bytes during a 15-second post-start check.

Self-hosted

Langfuse 3.177.1 (`langfuse/langfuse:3`, image revision `9bba37054c4ca5d41d345da0fd203451a7b466a3`)

- ClickHouse 26.5.1.882
- Docker Engine 29.1.3, API 1.52
- Ubuntu 22.04.5 LTS, Linux 5.15, x86_64
- No SDK is involved in the failure path

### Additional information

Current `main` still has no per-service logging limit in `docker-compose.yml`:
https://github.com/langfuse/langfuse/blob/main/docker-compose.yml

Docker documents that `json-file` is unbounded by default and recommends the `local` driver or explicit rotation to prevent disk exhaustion:
https://docs.docker.com/engine/logging/configure/

Possible fixes, depending on the maintainers' preferred ownership boundary:

1. Add bounded logging to the ClickHouse service (or all bundled services) in the example Compose file.
2. Add an explicit Docker log-rotation prerequisite to the self-hosting documentation.

I can contribute the preferred change after scope is confirmed.

Yes

I3eka:
## What does this PR do?

Adds an explicit opt-in Compose override for bounded local JSON logs across all six bundled services, addressing #16339 without changing the base file's Docker daemon logging policy.

- The base `docker-compose.yml` inherits the daemon's logging driver and options.
- `docker-compose.logging.yml` explicitly selects `json-file` with `max-size: "10m"` and `max-file: "3"` for all six services.
- README documents the opt-in command, custom-driver tradeoff, container recreation, and the fact that these bounds do not cover data volumes or ClickHouse's internal logs.

This is a proposal for the opt-in approach, not a claim that a maintainer has approved the logging-policy choice. Deployments using only the base Compose file must configure retention at the host/backend level; this does not silently enable rotation for existing installations.

## Type of change

- [x] Self-hosting configuration and documentation fix, with logging-driver changes only when explicitly opting in.

## Verification

- [x] `node scripts/compose-logging.test.mjs`: six services retain daemon defaults; the opt-in override applies all bounds without altering other configuration; custom `journald`, `syslog`, and `fluentd` overrides retain no JSON-specific rotation options; default Compose discovery does not load the opt-in file.
- [x] Regression failed against the prior base Compose configuration and passes after the fix.
- [x] Both base and opt-in `docker compose ... config --quiet` checks.
- [x] `node --check scripts/compose-logging.test.mjs` and `git diff --check`.

No containers were started or recreated, and no database was accessed. Application lint/typecheck/integration suites were not run for this configuration-only change. These checks validate Compose rendering, not live log rotation.

The PR adds a shared Docker logging configuration with three-file, 10 MB rotation and applies it to all six services. However, explicitly selecting `json-file` overrides operators’ daemon-level logging-driver choices rather than merely bounding existing JSON logs.
</details>

The PR should not merge until it avoids silently replacing self-hosters’ configured Docker logging backend.

All six services now explicitly use `json-file`, so deployments relying on a non-JSON daemon default will stop forwarding their service logs to the configured backend after container recreation.

**Files Needing Attention:** docker-compose.yml
</details>

`````markdown
### Issue 1
docker-compose.yml:7
**Hard-coded driver bypasses daemon logging**

If a self-hosted Docker daemon uses `journald`, `syslog`, `fluentd`, or another centralized logging driver, this explicit `json-file` setting overrides that daemon default for all six services, causing their logs to disappear from the operator's configured logging backend after the containers are recreated.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: [1f6e089](https://github.com/langfuse/langfuse/commit/1f6e0894e27d5e11b1025d2314d831db7736eccc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55189675)</sub>

### CODE REVIEW

**Comment by I3eka** [comment]
Reviewer focus: please double-check whether `10m` and `3` are the desired project-wide defaults. The implementation intentionally keeps `json-file` and applies the same bounds to all six services.

**Comment by I3eka** [comment]
@claude review

**Comment by I3eka** [comment]
@Steffen911 Before changing the PR, could you confirm the preferred ownership boundary for log rotation?

Greptile raised a compatibility concern: forcing `json-file` in Compose overrides daemon-level `journald`, `syslog`, `fluentd`, or other centralized drivers. A static Compose mapping cannot both inherit an arbitrary daemon driver and apply the `json-file`-specific `max-size` / `max-file` options.

The viable approaches I see are:

1. Keep the current PR: force `json-file` with `10m` / `3` for all six services. This is bounded by default, but overrides a custom daemon driver.
2. Use the Docker-recommended `local` driver in Compose. It rotates automatically, but still overrides a custom daemon driver unless we add and document an explicit driver variable.
3. Keep the base Compose driver-agnostic and add an opt-in Compose override containing bounded `json-file` settings for all six services. This preserves centralized logging, but adds another file and startup command.
4. Keep Compose driver-agnostic and document host-level daemon rotation as a self-hosting prerequisite (`local`, or bounded `json-file`). This preserves existing centralized drivers, applies only to newly created containers after a daemon restart, and may belong in the docs repository.

Making only the driver configurable is not sufficient because `max-size` and `max-file` are invalid for some other drivers.

Which approach would you prefer? If rotation should remain Compose-managed, should we keep `10m` / `3` or use `local` defaults? If daemon-managed, should this PR move to documentation here or to `langfuse-docs`?

**Comment by Steffen911** [comment]
@claude review

**Comment by I3eka** [comment]
@Steffen911, following up on the Claude review for `8fcf26b`: it found no bug in the change itself and correctly identified the remaining logging-driver choice as a maintainer decision. One correction to the review: `I3eka` is the PR author, not a human reviewer; my earlier comment was asking you for project direction. Could you confirm whether you prefer bounded `json-file` in the base Compose file, an opt-in Compose override, or daemon-level rotation documentation? I will update the PR accordingly.

**Comment by I3eka** [comment]
@Steffen911, gentle follow-up on the remaining maintainer decision for log rotation. Could you choose between bounded `json-file` in the base Compose file, an opt-in Compose override, or daemon-level rotation documentation? I will update the PR to the chosen ownership boundary immediately.

**Comment by I3eka** [comment]
@Steffen911, the branch is updated to the current `main` (`7637df1e`), with head `2b53bd6e`. The PR still changes only `docker-compose.yml` (+12 lines); `docker compose config --quiet` and `git diff --check` pass, and all six services resolve to the intended logging options. CLA is green.

Could you approve the fork workflow runs for this head? [CI/CD](https://github.com/langfuse/langfuse/actions/runs/34032795919) is waiting with `action_required` and has not started any jobs. The logging-driver design question is still open; this branch refresh does not change the proposed behavior.

**Comment by I3eka** [comment]
@Steffen911, refreshed this branch with main at 387fcf5c; head is 41212924, with no conflicts and no commits behind the checked base. The original +12-line Compose diff is unchanged, and Compose validation plus all six resolved logging configurations pass.

I corrected the PR description: this proposal explicitly overrides custom daemon logging drivers; it does not preserve them. That concern remains open. Could you choose whether the fix should live in the base Compose file, an opt-in override, or daemon-level rotation documentation? Please also approve the fork [CI/CD run](https://github.com/langfuse/langfuse/actions/runs/34316077706), which is waiting at action_required with no jobs executed.

File: docker-compose.yml [hunk 1]
```diff
@@ -3,10 +3,17 @@
 # In addition, we recommend to restrict inbound traffic on the host to langfuse-web (port 3000) and minio (port 9090) only.
 # All other components are bound to localhost (127.0.0.1) to only accept connections from the local machine.
 # External connections from other machines will not be able to reach these services directly.
+x-logging: &default-logging
+  driver: json-file
+  options:
+    max-size: "10m"
+    max-file: "3"
```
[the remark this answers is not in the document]
I3eka: Confirmed. I corrected the PR description and the non-breaking claim to state that explicit json-file overrides a custom daemon logging driver. The behavior itself has not been changed or approved: the base Compose versus opt-in override versus daemon-level documentation choice is still awaiting a maintainer decision. Leaving this thread open; merely removing driver while retaining json-file-specific options would also need compatibility validation.

I3eka:
@Steffen911, following up on the outstanding logging-policy decision. I refreshed the branch to `main` at `173cdea0`; head is `57bcdfb3` and the original +12-line Compose patch is unchanged. Compose validation, all six resolved logging configurations, and `git diff --check` pass; CLA is green.

The proposal still overrides non-default daemon logging drivers. Could you confirm whether you prefer bounded logging in the base Compose file, an opt-in override, or daemon-level rotation documentation? Please also approve the fork [CI/CD run](https://github.com/langfuse/langfuse/actions/runs/34835014385): it is `action_required`, with no jobs executed. The design threads remain open.

I3eka:
@Steffen911, I've pushed an opt-in version in 27528dcf so there is a concrete proposal to review: the base Compose file keeps the daemon's logging policy, and `docker-compose.logging.yml` enables bounded JSON logs for all six services, as you requested in #16339. The README spells out the command, centralized-logging tradeoff, and recreation requirements.

The regression failed with the previous forced driver and passes with this change: `Compose logging: 6 services preserve daemon defaults; opt-in bounds and 3 custom drivers passed.` Both Compose validation commands, the JavaScript syntax check, and `git diff --check` also pass. These are configuration checks; I haven't started containers or claimed live rotation or full application CI coverage.

Could you review the opt-in approach and the proposed `10m` / `3` limits? This removes the silent logging-backend change, but deliberately means deployments using only the base file still need host/backend retention configured. I updated the PR description and resolved the two compatibility findings; the policy choice is still yours to approve.

README.md
```diff
@@ -122,6 +122,20 @@ Run Langfuse on your own infrastructure:
 
 See [self-hosting documentation](https://langfuse.com/self-hosting) to learn more about architecture and configuration options.
 
+#### Docker log rotation
+
+The base `docker-compose.yml` inherits your Docker daemon's logging driver and options. Docker's default `json-file` driver has no rotation limit, so configure [host-level log rotation](https://docs.docker.com/engine/logging/configure/) or explicitly opt in to bounded local JSON logs:
```
Steffen911: @I3eka Thank you for this change and the design! My feeling is that we should just document that host level log rotation should be used and leave it at that? Or we document how to set the log configuration instead of using an overwrite?

To me, the entire thing feels like an edge-case that should be documented without too much noise around it.

What do you think?

I3eka: Thanks @Steffen911, agreed: documenting the daemon setting is enough here. In 9015eceb I removed the optional override and its dedicated test, and restored the base Compose file unchanged. The complete PR diff is now just six README lines linking Docker's logging configuration, noting that existing containers need recreation, and distinguishing stdout/stderr retention from data volumes and ClickHouse's internal logs.

I checked those details against Docker's documentation and verified that the diff contains only README.md, with no references left to the removed files. No daemon settings or running containers were changed. Application tests were skipped for this documentation-only revision. Could you take another look at the reduced scope?

Steffen911 (approved):

PR #16392, opened 2026-08-21
feat(structure): deny new files that would add a structure violation

nkabardin:
🟡 **Live preview: [pr-16392.preview.langfuse.com](https://pr-16392.preview.langfuse.com)**

> Stacked on #16391 (the rules skill every deny message links to). Merge #16390 → #16391 → this.

## TL;DR

An agent can no longer *create* a file whose path adds a project-structure violation. `structure:check-path` answers "would this path add a violation?" from the path alone — one directory read, ~70 ms as a process — and a pre-write hook wired into Claude Code, Cursor and Codex denies the write with the rule, the correct path, and a link to the rule file. Creation only, fails open, one env var turns it off.

## Why

The rules that keep growing are all decided at the moment a file is created:

| rule | what decides it | drift since the last baseline |
| --- | --- | --- |
| 3 | a filename | +15 |
| 4 | a filename in `fns/` | +14 |
| 5 | a folder name | — |

Nothing checked at that moment, so the sensor counted them afterwards and the number went up. Counting is not preventing.

A full `structure:stats` run cannot do this job: it needs a dependency-cruiser graph and a TS-parse census, and most of the wall time through `pnpm` is package-manager startup. So the check had to be a different, much smaller thing.

## What

**`web/scripts/structure/check-path.mjs`** — the placement question, path-only.

- Decides **rules 1, 3, 4, 5, 9, 13** (deny) and surfaces **rule 18** (ask).
- Never has an opinion about a graph rule — rule 8 cannot be known before the file's imports exist.
- Rule 5 calls `detectors.rule5` directly, so the check and the dashboard cannot disagree about what a kind folder is. Rules 1 and 3 mirror only the *naming* half of their detectors, because "named after the export" needs the file's text.
- Usable by hand: `pnpm --filter web run structure:check-path <path>`.

**`scripts/agents/hooks/structure-path-guard.mjs`** — one implementation, three response shapes behind `--tool`.

- **Creation only.** Fires only when the target path does not exist. An edit to a badly-named file is never blocked — its name is not the bug you were asked to fix.
- **Fails open, always.** Unreadable stdin, unparseable payload, a throw, a missing check, a 1.5 s stdin timeout: every one allows the write.
- **Escape hatch, named in the message.** `LANGFUSE_SKIP_STRUCTURE_HOOK=1`, following `LANGFUSE_PRE_COMMIT_SKIP_LINT` in `.husky/pre-commit`. A gate with no exit gets ripped out.
- Never calls `pnpm`.

**One declaration, three shims.** A new tool-neutral `hooks` key in `.agents/config.json`; `sync-agent-shims.mjs` maps it to each tool's event name and tool-name matcher and emits `.claude/settings.json`, `.cursor/hooks.json` and `.codex/hooks.json`. The two `hooks.json` files are committed, like `.cursor/environment.json`, because those tools read hooks before `pnpm install` can generate anything.

## Result

Every deny message carries the rule number, the rule in one line, the correct path, the rule file, and the escape hatch — because an agent told "no" without being told where the file goes just tries a different wrong place.

```
Project structure: this path would add 1 new violation.

web/src/features/traces/utils/formatCost.ts
  rule 5 — kind folders are a closed list: components, hooks, contexts, stores, fns, server, constants, types
    'utils' is not a kind folder (components|hooks|contexts|stores|fns|server|constants|types)
    correct path: web/src/features/traces/fns/formatCost.ts
    rule file: .agents/skills/project-structure/rules/05-kind-folders-are-a-closed-list.md

Existing files are grandfathered; this fires only on creating a new one.
Escape hatch: set LANGFUSE_SKIP_STRUCTURE_HOOK=1 to disable this hook.
```

<details>
<summary>Calibration — is it right?</summary>

The verdict is a prediction of what the sensor would count, so it was checked against the sensor over **every existing file** in `web/src`:

```
2712 existing files checked in 26 ms — 700 would be flagged
disagreements with the sensor: 1
rule 1  src/features/dashboard/components/home-preset-registry.tsx
```

That one is a kebab-case `.tsx` that exports no component. No rule counts it today, but the name is wrong under either naming rule, so it stays denied.

Getting there required one refinement worth knowing about: a path-only rule 1 initially flagged five files, four of them camelCase `.tsx` modules that return JSX (`showErrorToast.tsx`) rather than components. A camelCase stem is now taken at its word — it is asserting "I am a module", which rule 3 governs — and only a stem that is neither PascalCase nor camelCase is denied.

</details>

<details>
<summary>Latency — the measured number, not the estimated one</summary>

```
hook wall time over 10 runs: min 63 ms, median 69 ms, max 80 ms
bare `node -e ''` on the same machine: 60, 62, 68 ms
```

So the check itself is single-digit milliseconds and **the entire cost is node process startup** (in-process, 2,712 paths take 26 ms — 0.01 ms per path). The design estimate was ~20 ms; the honest number is ~70 ms, because node takes 60 ms to boot here. Still 50× cheaper than a scoped `structure:stats` and imperceptible inside a tool call. If you ever see this above ~150 ms, something is wrong.

</details>

<details>
<summary>Which tools were exercised live, and which by contract</summary>

**Claude Code — live, in a real session.** `.claude/settings.json` is picked up by the file watcher, so this was verified end-to-end:

| attempt | result |
| --- | --- |
| create `web/src/features/traces/utils/hookProbe.ts` | denied, rule 5, with the correct path |
| create `web/src/features/traces/hooks/use-hook-probe.ts` | denied, rule 3, `→ useHookProbe.ts` |
| create `web/src/features/traces/fns/hookProbe.ts` | allowed |
| **edit** `web/src/components/ui/button.tsx` (frozen folder, kebab name) | allowed — creation-only holds |

**Cursor and Codex — contract only. I did not launch either, and I am not claiming they work.** What was verified: the generated config matches each product's documented schema (Cursor `{"version": 1, "hooks": {"preToolUse": [{command, matcher, timeout}]}}`; Codex `{"hooks": {"PreToolUse": [{matcher, hooks: [{type, command, timeout}]}]}}`), and the hook was fed each tool's documented stdin payload and asserted to return its documented response shape, including Codex's `apply_patch` envelope with the target inside the patch text.

Two places where a tool has no native equivalent, so the hook approximates and says so in a comment:

- **Cursor's `preToolUse` takes `allow` or `deny` only** — no `ask`. Rule 18's ask becomes an allow carrying `agent_message`, so the note reaches the agent without blocking.
- **Codex has no `ask` for `PreToolUse` either.** The ask is returned as `additionalContext`, leaving Codex's own approval flow alone. If a Codex version ignores that field, the write proceeds — which is the fail-open direction.

</details>

<details>

- `pnpm --filter web run test src/__tests__/server/unit/structure-path-guard.servertest.ts` → `Tests  3 passed (3)`
- `pnpm run typecheck` → `Tasks: 8 successful, 8 total` (`web/tsconfig.json` includes `**/*.mjs`, so `check-path.mjs` is type-checked through its JSDoc)
- `pnpm run agents:check` → exit 0, shims in sync
- `prettier --check "**/*.{js,jsx,ts,tsx,css}"` → `All matched files use Prettier code style!`

The test covers the check's verdicts as a table (18 paths, including the legitimate ones it must stay silent about), each tool's response shape, and the fail-open paths: an existing file, an out-of-scope path, a payload with no path at all, unparseable stdin, empty stdin, and the escape hatch.

</details>

1. **Rule 5 fires even when the bad folder already exists.** Adding a file to an existing `features/x/utils/` does not raise the sensor's count (rule 5 counts directories), but it is denied anyway — otherwise the blind spot grows, since a file in `utils/` also escapes rule 3's naming check. This is the one place the hook is deliberately stricter than the count.
2. **Rule 18 is `ask`, not `deny`.** At pre-write time "no adjacent subject" and "test written first" are the same observation, and this repo's own guidance says to write the failing test first. The message names both fixes. It also degrades to a note on Cursor and Codex.
3. **`Bash` is not hooked.** `cat > web/src/features/x/utils/foo.ts` bypasses everything. Matching `Bash` would mean parsing shell redirections; the CI ratchet is the real backstop for that.
4. **The tool-name matchers are the fragile part.** `Write` for Claude and Cursor, `apply_patch|Edit|Write` for Codex, from each product's current docs. A renamed tool means the hook silently stops firing — quiet, not broken, and it fails in the allowing direction.

## Next in this series

Not in this PR, deliberately: the `SessionStart` briefing from a cached snapshot, the CI `structure` job with a touched-files ratchet, and the pre-commit gate.

The PR adds a creation-only, fail-open structure guard for agent file writes and a fast path-based checker for selected project-structure rules.
- Declares one tool-neutral pre-write hook and generates Claude, Cursor, and Codex configurations.
- Adds path normalization, rule findings, corrective guidance, and an environment-variable escape hatch.
- Adds process-level tests for rule decisions, tool response shapes, and fail-open behavior.
</details>

The PR appears safe to merge, with one non-blocking import-placement cleanup in the hook implementation.

The structure guard and checker have no accepted behavioral or security failures, but the hook dynamically imports its checker from inside main instead of keeping module imports at the top level.

**Files Needing Attention:** scripts/agents/hooks/structure-path-guard.mjs
</details>

```mermaid
flowchart TD
  A[Agent file-write request] --> B[Tool-specific hook shim]
  B --> C[Extract candidate paths]
  C --> D{Existing file or outside web/src?}
  D -->|Yes| E[Allow write]
  D -->|No| F[Run structure path checks]
  F --> G{Finding verdict}
  G -->|No finding| E
  G -->|Ask| H[Return tool-specific guidance]
  G -->|Deny| I[Block creation with corrective path]
  F -->|Error or invalid input| E
```
</details>

`````markdown
### Issue 1
scripts/agents/hooks/structure-path-guard.mjs:181-183
**Function-local checker import**

The structure checker is dynamically imported inside `main`, contrary to the repository convention that module dependencies remain at the top level; this makes initialization and dependency analysis less explicit.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(structure): refuse new files that w..."](https://github.com/langfuse/langfuse/commit/0364fb57188c56d09afb75b7c2f7ad6476747653) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55510571)</sub>

**Context used:**

**Learned From**

### DIFF
--- a/.agents/README.md
+++ b/.agents/README.md
@@ -19,10 +19,11 @@ or `.vscode/`.
 
 ## `config.json`
 
-`.agents/config.json` contains four kinds of data:
+`.agents/config.json` contains five kinds of data:
 
 - `shared`: defaults used across tools
 - `mcpServers`: project MCP servers and how to connect to them
+- `hooks`: tool-neutral agent hooks, one declaration per hook
 - `claude`: Claude-specific generated settings inputs
 - `codex`: Codex-specific generated settings inputs
 - `cursor`: Cursor-specific generated settings inputs
@@ -59,6 +60,16 @@ Current shape:
       "transport": "http",
       "url": "https://mcp.linear.app/mcp"
     }
+  },
+  "hooks": {
+    "preFileWrite": [
+      {
+        "name": "structure-path-guard",
+        "description": "...",
+        "script": "scripts/agents/hooks/structure-path-guard.mjs",
+        "timeoutSeconds": 10
+      }
+    ]
   },
   "claude": {
     "settings": {
@@ -118,9 +129,15 @@ Generated local artifacts:
 - `.codex/config.toml`
 - `.codex/environments/environment.toml`
 
+Generated and committed:
+
+- `.cursor/environment.json`
+- `.cursor/hooks.json`
+- `.codex/hooks.json`
+
 Cursor must read its environment contract before it can run the install script,
-so `.cursor/environment.json` is the one generated configuration file committed
-to the repository. Generate it from `.agents/config.json`; never edit it by
+and it reads its hooks before `pnpm install` can generate anything, so those
+files are committed. Generate it from `.agents/config.json`; never edit it by
 hand. `.cursor/Dockerfile` is also committed because it is an intentionally
 Cursor-specific runtime definition.
 
@@ -217,6 +234,38 @@ Update values in `shared`:
 - `setupScript`
 - `devCommand`
 - `devTerminalDescription`
+
+### Add a hook
+
+Declare it once under `hooks`, with a tool-neutral event name, and let the
+generator emit each tool's shape. Do not hand-write `.claude/settings.json`,
+`.cursor/hooks.json` or `.codex/hooks.json`.
+
+```json
+{
+  "hooks": {
+    "preFileWrite": [
+      {
+        "name": "example",
+        "description": "What it refuses, and how to get past it",
+        "script": "scripts/agents/hooks/example.mjs",
+        "timeoutSeconds": 10
+      }
+    ]
+  }
+}
+```
+
+`HOOK_EVENTS` in `scripts/agents/sync-agent-shims.mjs` maps a neutral event to
+each tool's event name and tool-name matcher; a tool with no mapping for an
+event simply gets no hook. One implementation per hook lives in
+`scripts/agents/hooks/`, receives the tool's payload on stdin, and writes that
+tool's response shape based on `--tool`.
+
+Two rules for any hook that can block: **fail open** — an unreadable payload,
+a throw or a timeout must allow the action — and **name an escape hatch** in
+the message it prints, following `LANGFUSE_PRE_COMMIT_SKIP_LINT` in
+`.husky/pre-commit`. A gate with no exit gets deleted.
 
 ### Add tool-specific generated inputs
 
--- a/.agents/README.md
+++ b/.agents/README.md
@@ -19,7 +19,7 @@ or `.vscode/`.
 
 ## `config.json`
 
-`.agents/config.json` contains five kinds of data:
+`.agents/config.json` contains six kinds of data:
 
 - `shared`: defaults used across tools
 - `mcpServers`: project MCP servers and how to connect to them
--- a/.agents/config.json
+++ b/.agents/config.json
@@ -28,6 +28,16 @@
       "url": "https://mcp.linear.app/mcp"
     }
   },
+  "hooks": {
+    "preFileWrite": [
+      {
+        "name": "structure-path-guard",
+        "description": "Refuse a new file under web/src whose path would add a project-structure violation. Creation only, fails open, disabled by LANGFUSE_SKIP_STRUCTURE_HOOK.",
+        "script": "scripts/agents/hooks/structure-path-guard.mjs",
+        "timeoutSeconds": 10
+      }
+    ]
+  },
   "claude": {
     "settings": {
       "permissions": {
--- a/.agents/skills/project-structure/SKILL.md
+++ b/.agents/skills/project-structure/SKILL.md
@@ -121,8 +121,18 @@ Two things about `--next` that read like bugs and are not:
 
 ## Enforcement today
 
-Measured, not gated. The tooling counts violations, points at them and proposes
-the rework; nothing blocks a merge on the total. Enforcement arrives one rule at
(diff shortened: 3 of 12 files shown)


nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

nkabardin:
@claude review

PR #16404, opened 2026-08-21
fix(in-app-agent): mark loop errors on the APM span after ACK

LysanderKie:
🟡 **Live preview: [pr-16404.preview.langfuse.com](https://pr-16404.preview.langfuse.com)**

## Summary
- Call `traceException` from the loop `onError` path so the BullMQ span is `status:error` even when the stream closes cleanly and the job is ACK'd.
- Log enumerable `name` / `message` / `stack` on `Error in agent execution` (Winston was serializing `Error` as `{}`).
- Does not add a second failed-run info log.

## Test plan
- `pnpm --filter worker run test executeInAppAgentRun.test.ts agent.test.ts` — `Tests 63 passed (63)` (includes the microsoft-agent file vitest also matched)
- After deploy, fail a background run the same way as this morning's 11:01 case and confirm the APM table row has `@error.message`.

Impacted packages: `worker`

### DIFF
--- a/.agents/skills/react-component-guidelines/SKILL.md
+++ b/.agents/skills/react-component-guidelines/SKILL.md
@@ -14,6 +14,7 @@ Components are useful because they act as an encapsulated unit and therefore pro
 
 ## Composition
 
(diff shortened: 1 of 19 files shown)


LysanderKie:
What to doubt: `onError` and the outer catch can both call `traceException` on the failStream path (double recordException on the same span). I left that in so a race where catch wins still marks the span. Also, `approval_expired` / `queue_timeout` still will not appear in this APM table — they never error the job span.

PR #16408, opened 2026-08-21
feat(filters): add the search bar to the sessions table

nkabardin:
🟡 **Live preview: [pr-16408.preview.langfuse.com](https://pr-16408.preview.langfuse.com)**

## TL;DR

The query bar now works on the **sessions table**, behind a `Sessions Search Bar`
Feature Preview. Sessions is the *second* surface for the bar, so the real
deliverable is the reusable pattern — the remaining ~17 filterable views copy
whatever this PR establishes.

> [!NOTE]
> **Rebased onto `main` now that #15906 has merged.** That PR squash-merged, so its
> commits are not in `main` — merging `main` in would have replayed all 85 of them and
> conflicted throughout. This branch is rebased instead, so it carries only the
> sessions work.

**What the merge changed here.** The feature-preview system was reworked in the same
window: per-preview titles moved into a shared `featurePreviewLabels` map, the retired
`searchBar`/`v4UpgradeUi` entries and their illustrations were deleted, and two new
surfaces appeared — a per-user previews popover and an organization-level defaults
page. Both enumerate the registry, so this preview joins them by registering its
label, and the illustration is restored under a name tied to this preview so the next
cleanup of that one cannot take it again. Two of their new tests asserted a hardcoded
count of preview switches, which any third preview breaks; both now derive the count
from the registry. The seam this builds on — `fieldRegistryFromColumns` and the
injected `FieldRegistry` — landed unchanged, so nothing in the registry design needed
revisiting.

## Why

The bar's own spec says the front half — registry, grammar, value validation — is
the only per-view fork, and that the recipe should be validated "against a real
consumer, not ahead of it". Sessions is that consumer, and it exercised parts of
the recipe the events table never could: two config generations, keyed score
columns, no full-text lane at all.

Sessions also has the product's **2nd-heaviest saved-view surface**, so breaking
round-trips was the worst outcome available. That constraint drove the main design
choice below.

## What

**The registry derives from the sidebar's FACET list, not the raw `ColumnDefinition[]`.**
A view's column list also carries things the sidebar deliberately never offers: a
duplicate (`usage` = `totalTokens`), a column owned by another control (`createdAt`,
the time-range picker), a retired one (`bookmarked`). Deriving from facets makes
"the bar is a strict subset of the sidebar" **structural** rather than aspirational —
the bar cannot author a filter the sidebar is unable to display or clear, and
anything outside the set resolves to null, so an older saved view's filter on it is
*preserved* in `skippedFilters` instead of becoming an unparseable token. Adding a
facet gives the bar the field for free.

**A bare word searches session ids.** Sessions has no full-text lane — `sessions.all*`
takes no `searchQuery` and its toolbar never had a search field — so a bare word had
nothing to bind to. But `id contains` is by far the most-applied sessions filter
(~13.6k applications / ~1.6k users per quarter, and *every one* of them uses
`contains`, i.e. "typed a fragment of an id"); the only other text column, comment
content, sees ~35. So a bare word lowers onto a declared `defaultTextField` through
the normal field path. It is offered before Enter (it joins the matching-filter
suggestions) and visible after (the bar re-renders `id:refund`, which commits to the
identical filter), so it is a canonicalization rather than a hidden rewrite.

**Three seam gaps sessions was the first surface to hit** — fixed in the shared helper
rather than worked around locally:

1. Keyed score columns were derived as plain fields. `score_categories` is
   `categoryOptions`, not `*Object`, so the existing filter let through a bogus
   keyless `score_categories:` field that lowers to a filter the backend cannot answer.
2. The trace-score namespace was tied to the observation one. Sessions aggregates
   scores at session level and has no `trace_scores_*` columns, so `traceScores.` is
   now separately closable.
3. **Ask AI would have used the events prompt.** `buildFilterSystemPrompt`
   branches on the registry id and falls through to the events prompt for
   anything unrecognised, so sessions got a correct 14-field catalog wrapped in
   events prose and worked examples. The registry now declares whether a view
   has its own prompt; the affordance is gated on that as well as the org
   entitlement, and the endpoint refuses a registry without one. **Ask AI is
   therefore off on sessions in this PR** — it comes back when someone writes
   the sessions prompt.
4. Placeholder/`has:` example copy was hardcoded to the events table and shipped
   verbatim everywhere — advertising `latency:>2` and `has:endTime` on views with
   neither. (This also fixes the evaluation-rules bar, which had the same wrong
   placeholder.)

<details>
<summary>Mechanism, decisions, and verification</summary>

### v3 vs v4: one registry, not two

`sessionsViewCols` and `sessionsEventsViewCols` differ by **exactly one column** —
v4 adds `metadata`. So one derivation parameterised by the existing `metadata` flag
covers both, and no future v3/v4 split needs a second registry either.

### `createFieldRegistry` stays unexported

Both derived views (rules, sessions) are fully expressed by
`fieldRegistryFromColumns` + overlay. Nothing has needed to hand-assemble a
registry, so it stays private — the concrete answer to the "do we need to export
it?" question, rather than exporting it speculatively.

### The registry prop is a silent footgun

`registry` defaults to `EVENTS_FIELD_REGISTRY` on **both** `useEventsSearchBar` and
`EventsSearchBarRow`. Passing it to the hook only leaves a view that *works* —
commits validate against the right fields — while the autocomplete offers the
events field list. Nothing throws. I hit this myself; it is now documented, and a
new view has to check the field dropdown, not just that Enter applies.

### Verified in the browser

- `chat` → suggestion `id:chat`, Enter → `?filter=id;string;;contains;chat`, bar
  re-renders `id:chat`, table narrows to the matching session.
- `test 123` → **one** filter `id contains "test 123"` (not two AND-ed per-word
  filters — that was a real bug caught during review of the running build).
- `scores.session-quality:>0.5 duration:>30 -env:dev countTraces:8` → all four
  lower correctly (aliases and negation included) and return exactly the right rows.
- `tags:(chat AND agent) metadata.region:eu commentContent:*bug*` → all-of arrays,
  keyed metadata, and glob contains all round-trip unchanged.
- Sidebar shows `Filters 4` with each facet populated — bar and sidebar reflect each
  other with no explicit sync.
- **Saved views**: created one from bar-authored filters, cold-loaded `?viewId=…`,
  and all four filters re-derived into the bar with the same rows.
- **No silent drops**: a URL carrying a single-value `tags all of` (a shape the
  grammar cannot render) shows only the renderable filters, and committing from the
  bar leaves the all-of filter intact in the URL.
- `latency:>2` and `traceScores.nps:>3` are rejected with the URL unchanged — events
  fields and the closed namespace do not leak. Console clean.

### Checks

`Tasks: 7 successful, 7 total` (turbo lint) · `Tasks: 8 successful, 8 total` (turbo
typecheck) · `Tests 262 passed (262)` (search-bar + feature-flag client tests) ·
`Tests 62 passed (62)` (AI filter-prompt server test).

The property harness (`searchBarInvariants`) runs the full field × operator × value ×
score-context matrix against the sessions registry, so commit-gate parity, no
silent drop/rewrite, and serialize↔parse symmetry are all covered per-registry.

</details>

## Result

Sessions filtering by keyboard, with the sidebar unchanged and saved views
round-tripping. More importantly: the extension recipe in the bar's README is now
written from a real second consumer rather than from intent, so surface #3 is data
and config, not code.


nkabardin:
@claude review

PR #16424, opened 2026-08-21
fix(in-app-agent): rewrite step-limit wrap-up so Bedrock never sees assistant prefill

LysanderKie:
🟡 **Live preview: [pr-16424.preview.langfuse.com](https://pr-16424.preview.langfuse.com)**

## Summary
- Opus on Bedrock rejects a trailing assistant message (`This model does not support assistant message prefill`). Mastra injects the step-limit wrap-up as `feedback`, which becomes that trailing assistant turn.
- Rewrite a trailing assistant prompt message to `user` immediately before `doStream`/`doGenerate`. This is last-mile on the provider payload only, so wrap-up does not appear as a chat user bubble.
- `instructions()` cannot carry wrap-up: Mastra snapshots them at stream start. Emitting wrap-up as an AG-UI user message would show in the assistant window.

## Impacted packages
- `worker`

## Test plan
- [ ] `CI=true pnpm --filter worker run test agent.test.ts`
- [ ] Confirm a step-limit wrap-up run succeeds on Opus/Bedrock and the last generation input ends with `role: user` containing `<step_limit_wrap_up>`
- [ ] Confirm the assistant transcript still shows only the real user prompt plus the assistant answer (no wrap-up user bubble)

This PR prevents Bedrock from receiving the step-limit wrap-up as an unsupported trailing assistant prefill.
- Rewrites the final assistant-role provider prompt to a user role immediately before streaming or generation.
- Applies the rewrite without changing the persisted AG-UI transcript.
- Adds coverage for streaming, generation, and unchanged tool-ending prompts.
- Adds structured logging when the step-limit wrap-up is injected.
</details>

The PR appears safe to merge, with no concrete blocking or independently actionable non-blocking issue identified.

The changed provider wrapper converts the problematic trailing assistant turn before both generation paths while leaving non-assistant-ending prompts unchanged, and the added tests exercise those behaviors.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(in-app-agent): rewrite step-limit wr..."](https://github.com/langfuse/langfuse/commit/042e3b2ad657bedfd2cef3c67ba94548c71e1524) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55573554)</sub>


LysanderKie:
Reviewer: doubt the last-mile rewrite more than the wrap-up copy.

Mastra still injects wrap-up as assistant `feedback`; we only flip the last prompt role to `user` at the Bedrock boundary. That is broader than wrap-up (any trailing assistant message gets rewritten), and `doGenerate` is rewritten without going through the tracing `onStart` path. If a real assistant prefill ever needs to stay assistant, this would silently convert it.

Also doubt that this is the right layer forever: `instructions()` cannot deliver wrap-up because Mastra snapshots them at stream start, and emitting a synthetic AG-UI user message would show in the chat. If Mastra later lets us append feedback as user/system, prefer that over rewriting every trailing assistant turn.

LysanderKie:
@claude review

LysanderKie:
@claude review

PR #16434, opened 2026-08-21
fix: batch trace/observation lookups in dataset run comparison grid

longduongbao29:
## Summary
- The dataset run comparison grid fired one `traces.byId` / `observations.byId` request per cell, so a page of up to 50 items x N selected runs sent hundreds of concurrent requests the instant the grid rendered.
- Add batched `traces.byIds` / `observations.byIds` tRPC procedures scoped to the caller's project.
- `DatasetCompareRunsTable` now collects the trace/observation ids the current page needs and fetches each kind once, passing lookup maps down to `DatasetAggregateTableCell` / `DatasetAggregateCellContent` instead of each cell querying independently.

## Test plan
- [ ] Open a dataset with 2+ runs compared on a preview deployment (`pr-<N>.preview.langfuse.com`), seed data if needed, open the browser Network tab, and confirm the compare grid issues one `traces.byIds` / `observations.byIds` request per page load instead of one request per cell.
- [ ] Change the selected runs / page and confirm exactly one new batched request fires per change, not one per cell.
- [ ] Added server tests: `traces.byIds` (batches multiple ids, empty input, omits unresolved ids) in `web/src/__tests__/server/traces-trpc.servertest.ts`; `observations.byIds` in the new `web/src/__tests__/server/observations-router-trpc.servertest.ts`.

The PR replaces per-cell trace and observation requests in the dataset run comparison grid with project-scoped batch procedures and lookup maps.

- Adds `traces.byIds` and `observations.byIds` procedures with server coverage.
- Collects IDs for the current page and issues one query per resource kind.
- Threads batch results and loading state through aggregate-column and cell components.
</details>

The PR should not merge until batch request failures stop being presented as if every affected trace or observation was deleted.

A transient or server-side failure of either new batch query empties that resource lookup and causes all corresponding cells to render false Not Found states while disabling their actions.

**Files Needing Attention:** web/src/features/datasets/components/DatasetAggregateTableCell.tsx, web/src/features/datasets/components/DatasetCompareRunsTable.tsx
</details>

```mermaid
sequenceDiagram
  participant Grid as DatasetCompareRunsTable
  participant TraceAPI as traces.byIds
  participant ObservationAPI as observations.byIds
  participant Cell as DatasetAggregateTableCell
  Grid->>Grid: Collect current-page IDs
  par Fetch traces
    Grid->>TraceAPI: traceIds + shared fromTimestamp
    TraceAPI-->>Grid: traces[]
  and Fetch observations
    Grid->>ObservationAPI: observationIds
    ObservationAPI-->>Grid: observations[]
  end
  Grid->>Grid: Build ID lookup maps
  Grid->>Cell: Maps and loading flags
  Cell->>Cell: Render output or missing state
```
</details>

`````markdown
### Issue 1
web/src/features/datasets/components/DatasetAggregateTableCell.tsx:81-83
**Batch failures become not-found state**

When either batched lookup fails, its data map is empty and its pending flag becomes false, so every affected cell is classified as not found and loses its Annotate and peek actions even though the resources may still exist. Propagate the query error separately so only a successful response that omits an ID produces `NotFoundCard`.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: batch trace/observation lookups in ..."](https://github.com/langfuse/langfuse/commit/22f87c132abbf0993d7a49536b0fc7635ef8b3fb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55685778)</sub>

**Context used:**

- Knowledge Base — [Datasets, Dataset Runs, and Experiments](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/datasets-experiments.md)

### DIFF
--- a/web/src/__tests__/server/observations-router-trpc.servertest.ts
+++ b/web/src/__tests__/server/observations-router-trpc.servertest.ts
@@ -1,0 +1,121 @@
+import type { Session } from "next-auth";
+import { prisma } from "@langfuse/shared/src/db";
+import { appRouter } from "@/src/server/api/root";
+import { createInnerTRPCContext } from "@/src/server/api/trpc";
+import {
+  createTrace,
+  createTracesCh,
+  createObservation,
+  createObservationsCh,
+} from "@langfuse/shared/src/server";
+import { randomUUID } from "crypto";
+
+// Note: `observations-trpc.servertest.ts` already exists and covers the
+// `generations` router (a legacy alias), not the `observations` router
+// tested here — kept as a separate file to avoid colliding names.
+describe("observations router trpc", () => {
+  const projectId = "7a88fb47-b4e2-43b8-a06c-a5ce950dc53a";
+
+  const session: Session = {
+    expires: "1",
+    user: {
+      id: "user-1",
+      canCreateOrganizations: true,
+      name: "Demo User",
+      organizations: [
+        {
+          id: "seed-org-id",
+          name: "Test Organization",
+          role: "OWNER",
+          plan: "cloud:hobby",
+          cloudConfig: undefined,
+          metadata: {},
+          aiFeaturesEnabled: false,
+          aiTelemetryEnabled: false,
+          projects: [
+            {
+              id: projectId,
+              role: "ADMIN",
+              retentionDays: 30,
+              deletedAt: null,
+              name: "Test Project",
+              hasTraces: true,
+              metadata: {},
+              createdAt: new Date().toISOString(),
+            },
+          ],
+        },
+      ],
+      featureFlags: {
+        excludeClickhouseRead: false,
+        templateFlag: true,
+        searchBar: false,
+        v4BetaToggleVisible: false,
+        observationEvals: false,
+        experimentsV4Enabled: false,
+      },
+      admin: true,
+    },
+    environment: {} as any,
+  };
+
+  const ctx = createInnerTRPCContext({ session, headers: {} });
+  const caller = appRouter.createCaller({ ...ctx, prisma });
+
+  describe("observations.byIds", () => {
+    // Regression coverage for the dataset run comparison grid, which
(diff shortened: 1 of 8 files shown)


longduongbao29:
Things worth doubting here:

- **`traces.byIds` uses the legacy `traces` table only** (`getTracesByIds`), not the V4 events-table routing that `traces.byId`/`enforceTraceAccess` does. In dual-write mode a trace that only exists in the events tables (e.g. some internally produced traces) would be silently omitted from the batch and render as "not found" in a compare cell, even though `traces.byId` would resolve it via the events-table fallback. I judged this acceptable since dataset-run trace items are normal ingested traces, but it's a real behavior gap versus the per-cell endpoint it replaces.
- **Batching key/shape**: one `traces.byIds` + one `observations.byIds` call per page load, keyed off the currently rendered page's `datasetItemsWithRunData` rows (not virtualized — `DataTable` still mounts the full page). `fromTimestamp` for the batch is the *minimum* `createdAt - 1 day` across the page (previously each cell had its own bound), which is a slightly wider window than before per-item but should still be tight enough to matter for CH performance.
- **`byIds` procedures use `protectedProjectProcedure`** (project-membership check only) rather than `protectedGetTraceProcedure`'s per-trace public/session visibility check that `byId` has. That's intentional — the whole compare page already requires project membership — but it does mean these new endpoints are project-scoped only, not usable for cross-project public-trace lookups the way `byId` is.
- **Loading/not-found state per cell**: a cell now shows a skeleton until its *entire* batch (all ids of its kind on the page) resolves, and "not found" if the batch completed but the id wasn't returned — slightly coarser than before, where each cell had independent loading/error state.

Test plan is unticked — I couldn't exercise the actual browser flow (no local Docker/DB stack in this sandbox), so the click-path verification is left for a reviewer/preview.

PR #14650, opened 2026-06-30
feat(redis): opt-in short-lived credentials via Azure Managed Identity

Oscmage:
## What does this PR do?

Adds optional short lived Redis credentials via Azure Managed Identity, building on the credential core merged in #14325. This is the Redis half of the plan in #14278; Postgres follows in a later PR.

An `azure_managed_identity` provider mints a Microsoft Entra token through `@azure/identity` (scope `https://redis.azure.com/.default`, username is the identity object id). On each refresh the Redis binding updates the connection password and issues a fresh AUTH on the open connection. New env vars: `REDIS_AUTH_METHOD` (defaults to `static`), `REDIS_AZURE_CLIENT_ID`, and `REDIS_AZURE_SCOPE`. The `static` default keeps today's username and password path unchanged, so the change is fully backward compatible and off by default.

## Testing

`credentials.test.ts` mocks `@azure/identity` and uses fake timers: provider token scope and username (system assigned and user assigned), the Redis bind (initial token, fresh AUTH on refresh, no op when already connected), and the env factory including the static fallback. No cloud resources required.

cc @Steffen911 @jannikmaierhoefer

<details open><summary><h3>Greptile Summary</h3></summary>

This PR adds opt-in short-lived Redis credentials via Azure Managed Identity, building on the `RefreshingTokenManager` from #14325. The `static` default leaves existing auth behaviour unchanged; opting in via `REDIS_AUTH_METHOD=azure_managed_identity` mints a Microsoft Entra token and re-AUTHs the live connection before each expiry.

- Adds `AzureManagedIdentityCredentialProvider` (wraps `@azure/identity`), `bindManagedCredentialToRedis` (applies token to ioredis options and issues live `AUTH` on refresh), and `getRedisManagedCredentialProviderFromEnv` (env-driven factory).
- Integrates with `createNewRedisInstance` using `lazyConnect: true` so the credential bootstrap can set `client.options.password` before the first connection attempt, but the bootstrap is launched fire-and-forget, creating a window where callers (notably BullMQ) can connect before credentials are applied.
- Cluster and sentinel code paths silently ignore the managed-credential setting with no log warning.
</details>

<details><summary><h3>Confidence Score: 3/5</h3></summary>

The static/default path is unchanged and safe to merge as-is; the managed-credential path has a real connection timing problem that would surface as NOAUTH errors in production for BullMQ queue connections when Azure Managed Identity is enabled.

[...]

PR #16444, opened 2026-08-22
feat(redis): opt-in short-lived credentials via Azure Managed Identity

### LINKED ISSUE #16185: feat(redis): support Microsoft Entra ID authentication
## Problem

Langfuse can only authenticate to Redis with a static password (`REDIS_AUTH`) or a connection string. On Azure this blocks two common deployments:

- **Azure Managed Redis caches with access keys disabled**, where Microsoft Entra ID is the only accepted authentication method. Entra is enabled by default on newly created AMR caches, so this affects new deployments out of the box.
- **Environments with a no-static-secrets policy**, where a long-lived cache password in the environment is not permitted.

The hard part is not acquiring a token — it is that Entra credentials are short-lived. The server drops connections when a token expires, so the client must re-`AUTH` on live connections before expiry. ioredis has no credential-provider hook (redis/ioredis#2022 is still open) and Microsoft's `@redis/entraid` only supports node-redis, so this has to be handled in Langfuse.

## Proposal

Introduce a credential-provider seam in `packages/shared/src/server/redis/` rather than special-casing Azure.

`createNewRedisInstance` asks a provider for credentials instead of reading `REDIS_AUTH` directly:

- The **default provider is static** and behaviourally identical to today — no timers, no connection registry, no re-authentication.
- A **rotating provider** handles acquisition, renewal and re-`AUTH` on live connections.

Everything auth-specific reduces to one factory returning an `@azure/identity` `TokenCredential`, which covers managed identity, workload identity, service principal (secret or certificate) and the local Azure CLI credential from a single code path. The same machinery would later cover AWS ElastiCache IAM auth, which may be the more widely requested case.

Gated behind `REDIS_AUTH_MODE`, defaulting to `password`, so existing deployments are untouched.

## Questions for maintainers

1. **New dependency.** This needs `@azure/identity`. Langfuse currently authenticates Azure Blob with `StorageSharedKeyCredential` (`StorageService.ts:375`), so there is no existing managed-identity precedent. I import it dynamically so it is never loaded unless `REDIS_AUTH_MODE=azure-entra`. The same credential would later enable managed identity for Blob storage.
2. **Abstraction shape.** Is a general credential provider the right call, or would you prefer something narrower and Azure-specific?
3. **PR shape.** I can split this into a behaviour-preserving `refactor(redis): extract credential provider` followed by the feature, if that reviews more easily.

## Implementation notes

Three things that were not obvious while building this:

- **BullMQ duplicates the connection.** `Worker` builds its blocking connection via `duplicate()` on the client it is given, and ioredis shallow-copies options on duplicate. Those connections need explicit registration, or workers silently stop consuming at the first token expiry while producers keep enqueuing successfully.
- **Renewal must be token-driven, not a fixed interval.** Managed identity tokens last ~24h; service principal and workload identity tokens ~60-90min. Renewal takes the earliest of `refreshAfterTimestamp`, a fraction of the lifetime, and a hard margin before expiry (Azure requires the replacement at least three minutes early).
- **Token acquisition must be deduplicated process-wide.** Langfuse opens enough connections that per-connection token requests would hit instance-metadata-endpoint throttling, turning a routine refresh into an outage at the next expiry.

Entra mode is rejected at startup without TLS, with `REDIS_AUTH` set, or with `REDIS_CONNECTION_STRING`, which cannot carry a rotating credential.

## Status

I have a working implementation that typechecks clean across `shared` and `worker`, and am happy to open a PR. Raising this first per CONTRIBUTING.

Reference: [Use Microsoft Entra for cache authentication with Azure Managed Redis](https://learn.microsoft.com/en-us/azure/redis/entra-for-authentication)

Karan9bhat:
## What does this PR do?

Adds opt-in short-lived Redis credentials via Azure Managed Identity — the Redis half of the plan in #14278, and the first consumer of the credential core merged in #14325.

This **replaces #14650**, as discussed in #16185. The first commit is @Oscmage's original work, re-applied onto current `main` with authorship preserved. The rest address the open review comment and fix defects found by testing it against a real Redis, a real Entra token, and two rounds of security review.

An `azure_managed_identity` provider mints a Microsoft Entra token through `@azure/identity` (scope `https://redis.azure.com/.default`, username is the identity's object id). The token is written to the connection's `options.password` — what ioredis builds its auth handshake from — and live sockets are re-`AUTH`ed in place on refresh, so a rotation doesn't drop the connection or its in-flight work.

New env vars: `REDIS_AUTH_METHOD` (defaults to `static`), `REDIS_AZURE_CLIENT_ID`, `REDIS_AZURE_SCOPE`. **The `static` default leaves the existing username/password path completely unchanged**, and every new rule is inert unless the managed path is enabled.

---

### 1. The `connect()` wrapper is removed — it broke every command-triggered connect

`bindManagedCredentialToRedis` monkey-patched `connect()` to await a token first. But ioredis assigns `this.condition` synchronously *inside* `connect()` and reads it back in the same tick when queueing a command:

```js
// ioredis 5.10.1, Redis.js sendCommand
if (this.status === "wait") this.connect().catch(noop);
...
this.offlineQueue.push({ command, stream, select: this.condition.select });  // <- undefined
```

Deferring the real connect left `condition` undefined, so any command issued before an explicit `connect()` threw `TypeError: Cannot read properties of undefined (reading 'select')` — *synchronously*, so callers could not catch it. BullMQ takes exactly that path.

| case | result |
|---|---|
| bound client, command before connect | `TypeError … reading 'select'` |
| bound client, explicit `connect()` first | PONG |
| **unbound control**, command before connect | PONG |

The token is written to `options.password` instead. Connections are registered for their lifetime and never unregistered, so a rotation also refreshes the options of a *disconnected* client — meaning whenever BullMQ revives one through `RedisConnection.reconnect()`, its first handshake already carries the current token.

### 2. Acquiring the first credential is required, and wired into startup

ioredis does not retry a rejected AUTH handshake — `retryStrategy` covers socket failures, not protocol-level rejection — so a connection that outruns the first token is closed for good. `initializeRedisManagedCredentials()` is awaited in the worker bootstrap before `app.ts` registers any queue, and in the web instrumentation hook before the first request. Clients keep `lazyConnect`, so nothing opens a socket before that runs.

Verified: without it, a real BullMQ workload processes **zero** jobs.

### 3. Two defects that made it unusable at Langfuse's connection count

- `getRedisManagedCredentialProviderFromEnv()` built a fresh provider — hence a fresh `TokenCredential` — and each bind a fresh `RefreshingTokenManager`, **per connection**. With a producer and consumer per queue plus the cache singleton and rate limiters, that's dozens of refresh loops against a metadata endpoint that throttles. Both are now process-wide singletons.

- BullMQ's `Worker` doesn't use the connection it's handed; it duplicates it, and ioredis implements `duplicate()` as `new Redis({ ...this.options })`. The copy inherited no tracking and a frozen password, so it authenticated once and was never renewed: at first expiry the server drops it, the reconnect replays a stale password, and **job consumption stops silently while producers keep enqueuing successfully**. `duplicate()` is now wrapped so copies join the same manager.

### 4. Security hardening

Two review rounds found several ways the feature could leak or quietly weaken the credential it exists to protect:

- **The token could be written to the logs.** ioredis attaches the command it was running to reply errors, and for AUTH the arguments are the credential. winston is configured with `errors({stack:true})` + `json()`, which serialise an Error's own enumerable properties. That shape — `logger.error("... error", err)` on a connection or queue `error` event — appears **57 times across 46 files** in the queue layer, and BullMQ's queue-base forwards the ioredis error verbatim. Redacted at the logger, which covers every one of those sites and any added later, mirroring the allow-list already applied to ioredis spans in `server/instrumentation`. Only the arguments are replaced; command name, message and stack are kept.

- **Nothing required TLS, and requiring it didn't require verifying the peer.** A static password is scoped to one cache; an Entra token is a bearer credential the whole audience accepts, replayable until expiry. The managed path now requires `REDIS_TLS_ENABLED`, and rejects `REDIS_TLS_REJECT_UNAUTHORIZED=false` and `REDIS_TLS_CHECK_SERVER_IDENTITY=false` — otherwise an unauthenticated endpoint could still be handed the token.

- **`REDIS_USERNAME` is now required.** Without it ioredis sends no AUTH at all, so a server permitting anonymous access would be connected to unauthenticated rather than failing closed. Azure requires the object id as the username, so any working configuration already sets it.

- **`REDIS_CONNECTION_STRING` is rejected.** Only the host/port branch cleared the static password, so an operator who "moved to managed identity" but left credentials in the URL kept using the long-lived secret — and a failed token fetch was invisible.

- **`REDIS_AZURE_SCOPE` is validated** against the Redis resource (sovereign clouds and the application-id form included). It was a bare string, so any audience could be minted and sent to the Redis host as a password.

- **Cluster and sentinel are rejected**, rather than accepted and then silently downgraded — `createNewRedisInstance` routes them to builders that never bind the credential.

`RefreshingTokenManager`'s contract is unchanged; it gained only the shared error-rendering helper.

Fixes #16185. Supersedes #14650.

---

## How has this been tested?

**Unit** — 52 tests across 7 suites. `credentials.test.ts` covers the provider, token distribution, live re-`AUTH`, the no-AUTH-when-not-connected rule, retry after a failed first fetch, cross-connection deduplication, `duplicate()` and nested duplicates, credential currency for a disconnected client's revival, the production guard on the test seam, and that a failed re-`AUTH` never reaches the logger with a payload. `logger.test.ts` pins the redaction. `@azure/identity` is mocked and timers faked. The pre-existing `RefreshingTokenManager.test.ts` passes unchanged.

**Real Redis + real BullMQ** — because the original bug was invisible to stubbed clients. A real Redis 7.2 server with an ACL user whose password rotates on a short TTL and expires on schedule, and `CLIENT KILL` simulating the server dropping connections at expiry:

| run | result |
|---|---|
| static-password control | 220/220 jobs, 0 lost |
| rotating credential | 220/220 jobs, one token per refresh |
| **`duplicate()` patch removed** | **73/219 jobs** — `WRONGPASS`, consumption dead, producers unaffected |
| startup acquisition skipped | **0/219 jobs** |
| revival after rotation, retries disabled | authenticates on the **first** attempt |

The failing rows matter: the regression tests fail in exactly the production failure mode.

**Real Entra token** — `az login`, then the shipping code end to end against a local Redis (which only string-compares the password, so a real JWT works):

- token audience `https://redis.azure.com`; `oid` claim matches `az ad signed-in-user show` exactly
- driven through the real path — env config → `getRedisManagedCredentialProviderFromEnv()` → `initializeRedisManagedCredentials()` → `createNewRedisInstance()` → BullMQ — credential applied at construction, `lazyConnect` forced, **50/50 jobs**, and **4 connections authenticated as the identity** (cache, queue, worker, and BullMQ's duplicated blocking client)
- a wrong object id is rejected, so the passing runs aren't passing by accident
- log-leak checks against real failed AUTHs with a real token, through the repo's own logger: the credential is absent from both the credentials module and a real BullMQ `error` handler

**Env validation matrix — 15/15**, including two regression rows proving the static path is untouched (`static + rejectUnauthorized=false` and `static + cluster` still pass).

Not covered, and not coverable without an Azure resource: that Azure accepts the token, and the data-access-policy assignment on a real cache.

## Type of change

## Considered, not addressed here

- **`DefaultAzureCredential` breadth** — with `REDIS_AZURE_CLIENT_ID` unset (the documented system-assigned path) the chain also includes environment and developer credentials, so a misconfigured host can authenticate as an unintended identity rather than failing. A `REDIS_AZURE_CREDENTIAL_TYPE` pin would close it if you'd prefer.
- **Refresh backoff** — `RefreshingTokenManager` retries a failed refresh on a fixed 5s interval with no backoff, jitter or cap; across many pods that's a self-inflicted throttle. Left alone deliberately, since that file is upstream-owned.
- **`DOCKER_BUILD=1` bypasses the whole env schema**, including these guards. Pre-existing, and the shipped Dockerfiles set it back to `0` in the runtime stage.
- **`AUTH` is invalid in ioredis subscriber mode**, so such a connection can't be re-authenticated in place. Not reachable today (nothing calls `subscribe`), and it recovers on the server-initiated drop at expiry.
- **Cluster and sentinel support** — now rejected at startup rather than silently downgraded, but still unimplemented.
- **Shutdown** — the manager's timer is `unref()`'d so the process exits, but nothing calls `.stop()`.
- Docs for the new env vars in `langfuse/langfuse-docs`.

## Mandatory Tasks

Adds opt-in Azure Managed Identity credentials for single-node Redis, including environment validation, process-wide token refresh, live connection reauthentication, duplicate-client tracking, credential log redaction, and web/worker startup prefetching.
- Introduces an Azure credential provider and shared Redis token manager.
- Enforces TLS, verified peers, username, scope, and topology constraints for managed authentication.
- Updates Redis construction and BullMQ duplicates to consume rotating credentials.
- Adds credential-redaction and managed-auth unit coverage.
</details>

The PR should not merge until web startup always prefetches managed credentials and failed live Redis reauthentication is retried.

Valid web configurations can skip the only awaited credential initialization and race their first Redis connection, while transient AUTH failures during later rotations are logged without recovering the affected live socket.

**Files Needing Attention:** web/src/instrumentation.ts; packages/shared/src/server/auth/credentials/redisCredentials.ts
</details>

```mermaid
sequenceDiagram
  participant Boot as Web/Worker startup
  participant Manager as Token manager
  participant Azure as Azure Identity
  participant RedisClient as ioredis/BullMQ client
  participant Redis as Redis
  Boot->>Manager: initialize credentials
  Manager->>Azure: getToken(scope)
  Azure-->>Manager: token and expiry
  Manager->>RedisClient: set options.password
  RedisClient->>Redis: Connect and AUTH
  loop Before token expiry
    Manager->>Azure: refresh token
    Azure-->>Manager: replacement token
    Manager->>RedisClient: update options.password
    RedisClient->>Redis: AUTH replacement token
  end
```
</details>

`````markdown
### Issue 1
web/src/instrumentation.ts:10-18
**Managed credential prefetch is skipped**

If a web deployment uses `REDIS_AUTH_METHOD=azure_managed_identity` with `NEXT_PUBLIC_LANGFUSE_RUN_NEXT_INIT=false`, this branch skips the only awaited credential prefetch while Redis remains enabled. The first command can open its lazy connection before the asynchronous token fetch populates `options.password`, causing Redis-backed rate limiting, caching, or queue production to fail from first use.

### Issue 2
packages/shared/src/server/auth/credentials/redisCredentials.ts:66-72
**Failed reauthentication is not retried**

When a ready connection encounters a transient error during the refresh `AUTH`, this detached promise only logs the rejection and the manager schedules the next ordinary token refresh. The socket remains authenticated with the old token until expiry, causing queue consumption or cache operations to be interrupted even though the process already holds a valid replacement token.

### Issue 3
web/src/instrumentation.ts:14-15
**Import is inside startup function**

The new shared-server import is placed inside `register()` instead of at module scope, violating the repository's required import structure and adding an unnecessary runtime import boundary to startup initialization.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/redis-mana..."](https://github.com/langfuse/langfuse/commit/7d9861e3d0dafc0fac5e3839f8eed3a1c788fab4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55891262)</sub>

**Context used:**

**Learned From**


Karan9bhat:
@Steffen911 following up on #16185 — this is the replacement for #14650 we discussed, with @Oscmage's original commit preserved.

PR #16446, opened 2026-08-23
fix(evals): keep corrupted-filter evaluation rules patchable and deletable

### LINKED ISSUE #16445: Corrupted evaluation-rule filter prevents disabling or deleting an active rule via public API
### Describe the bug

A migrated/v2 evaluation rule with two LLM-as-a-judge assignments is actively producing scores, but every public API operation on that rule fails with HTTP 500:

```json
{"message":"Evaluation rule filter is corrupted","code":"internal_error"}
```

This includes:

- `GET /api/public/unstable/evaluation-rules/{id}`
- `PATCH ...` with only `{"enabled": false}`
- `PATCH ...` with a complete valid replacement observation filter plus `enabled: false`
- `DELETE ...`
- list, because it includes the same rule

A separate rule in the same project with a normal `isRootObservation = true` filter can be read and disabled successfully, so auth/project access is not the issue.

### Impact

There is no public-API escape hatch to stop a live rule once its stored filter no longer passes the new public adapter. The rule continues invoking judges and incurring cost. Even DELETE is blocked.

### Likely cause

On current `main`, `updatePublicEvaluationRule` calls `toApiWritableV2EvaluationRule(existing)` before applying the replacement input, and `deletePublicEvaluationRule` does the same before deleting. That serialisation reaches `toApiFilters()` and throws before either mutation can run:

- `web/src/features/evals/server/unstable-public-api/evaluation-rule-service.ts`
- `web/src/features/evals/server/unstable-public-api/adapters.ts`

### Expected behavior

At minimum, disabling and deleting should remain possible for a malformed legacy/migrated rule. A full filter replacement should also be able to repair it without first parsing the old filter. Possible approaches:

1. let `PATCH {enabled:false}` and DELETE mutate from the stored record without public serialisation first;
2. when a complete replacement filter is provided, do not derive `nextFilter` from `existingPublic`;
3. return a degraded read representation plus a corruption warning instead of making the resource unmanageable.

### Environment

- Langfuse Cloud
- Public API/CLI schema 4.10.0
- Observed 2026-08-23

chelsealong:
## What does this PR do?

Fixes #16445.

A migrated evaluation rule whose stored filter no longer parses under the
current public schema made **every** public API call on that rule fail with
HTTP 500 `"Evaluation rule filter is corrupted"` — `GET`, `PATCH`, `DELETE`,
and `list`. There was no way to disable or delete a live, misbehaving rule
through the public API.

Root cause: `toApiV2EvaluationRule` (used by every read path, and by
`updatePublicEvaluationRule`/`deletePublicEvaluationRule` to build the
pre-mutation "existing" representation) threw as soon as it tried to
re-serialize the stored filter into the public shape — before any mutation
could even be attempted.

### Fix

- `toApiV2EvaluationRule` now degrades an unparseable stored filter to `[]`
  in the public read shape instead of throwing (the parse failure is still
  logged via the existing `logger.error` call in `toApiFilters`). This alone
  unblocks `GET`, `list`, and lets the pre-mutation read succeed for `PATCH`
  and `DELETE`.
- A `PATCH` that only flips `enabled`/`sampling` (no `name`/`target`/`filter`/
  evaluator changes) is now routed through `RuleService.setEnabled` instead
  of `RuleService.update`. `setEnabled` never re-validates or rewrites the
  stored filter, so it can disable a rule even when that filter is corrupted
  — `update` unconditionally re-validates whatever filter it's given
  (inherited or not), which would still reject a corrupted rule. This also
  fixes a related edge case: without this, an enabled/sampling-only PATCH on
  a corrupted rule would have silently overwritten the real stored filter
  with the degraded `[]` value.
- `DELETE` needed no separate change: it was already filter-independent once
  the pre-mutation read stopped throwing.

This is a read/write-path fix only; it does not touch how the rule engine
itself evaluates the stored filter at runtime.

### Test plan

Added `web/src/__tests__/server/unstable-evals-api.servertest.ts` >
`"still allows disabling and deleting an observation rule with a corrupted
stored filter"`. It inserts a rule directly via Prisma with a filter shape
that fails the current public schema (mirroring a migrated/legacy rule),
then verifies:
- `PATCH { enabled: false }` returns 200 (was 500 before the fix).
- The stored filter is left byte-for-byte untouched afterward (no silent
  data loss).
- `DELETE` then succeeds (200).

Confirmed the test fails against the pre-fix code (`git checkout HEAD~1` on
the two source files, i.e. reverting just the fix while keeping the test):

```
AssertionError: expected 500 to be 200
```

Full verification, run locally against Postgres/ClickHouse via
`docker compose -f docker-compose.dev.yml`:

- `pnpm --filter web run test src/__tests__/server/unstable-evals-api.servertest.ts`
  → `Tests  16 passed (16)`
- `pnpm --filter web run test src/__tests__/server/unit/unstable-evals-service.servertest.ts src/__tests__/server/unit/unstable-evals-contract.servertest.ts`
  → `Tests  42 passed (42)`
- `pnpm --filter web run lint` → clean
- `pnpm --filter web run typecheck` → clean

## Type of change

## Mandatory Tasks

## Checklist

- I have checked if my PR needs changes to the documentation (no — internal
  behavior fix, no contract change)

## What to doubt in review

- The `isEnabledOrSamplingOnlyPatch` check in `updatePublicEvaluationRule`
  only bypasses `RuleService.update` when the request touches none of
  `name`/`target`/`filter`/`evaluator(s)`. A PATCH that changes anything else
  (e.g. just renaming) on a corrupted-filter rule still 400s inside
  `RuleService.update`'s own filter re-validation — that's unchanged by this
  PR and out of scope per the issue's "at minimum, disable/delete" ask, but
  worth confirming that's an acceptable residual limitation.
- The degraded filter (`[]`) is only ever used for the public API's *read*
  representation and for the (skippable) preflight check when re-enabling; it
  is never written back to storage — writes either omit the filter (routed
  through `setEnabled`) or come from a fresh, valid `input.filter`. Worth a
  second look that no other code path derives a write value from the
  degraded read shape.
- This PR does not add a "corruption" flag to the public response (one of
  the three options the issue proposed). Consumers reading a corrupted rule
  will see `filter: []` with no signal that it differs from a genuinely
  empty filter, beyond the server-side log line already emitted by
  `toApiFilters`.

## AI disclosure

This PR was authored by an autonomous coding agent (Claude) operating on
this repository, with the changes reviewed against the repo's own tests,
lint, and typecheck before being pushed.

This PR makes evaluation rules with corrupted stored filters readable and routes enabled/sampling-only updates through the filter-independent status mutation path.
- Degrades corrupted filters to an empty public representation.
- Preserves stored filter bytes for enabled/sampling-only PATCH requests.
- Adds an integration test covering disable and delete recovery.
</details>

The PR needs a fix before merging because some filter-omitting PATCH requests can silently erase a corrupted rule's stored filter.

The new degraded read value is correctly isolated for enabled/sampling-only updates, but name and assignment updates still derive writable fields from that value and persist an empty filter.

**Files Needing Attention:** web/src/features/evals/server/unstable-public-api/adapters.ts, web/src/features/evals/server/unstable-public-api/evaluation-rule-service.ts
</details>

```mermaid
flowchart TD
  A[Load rule with corrupted stored filter] --> B[toApiV2EvaluationRule]
  B --> C[Public filter becomes empty array]
  C --> D{PATCH fields}
  D -->|enabled or sampling only| E[Preserve stored filter and call setEnabled]
  D -->|name or assignment fields| F[Use empty public filter in ruleFields]
  F --> G[RuleService.update]
  G --> H[Stored filter overwritten with empty array]
```
</details>

`````markdown
### Issue 1
web/src/features/evals/server/unstable-public-api/adapters.ts:630
**Degraded filter leaks into writes**

When a corrupted-filter rule receives a PATCH that omits `filter` but changes its name or assignments, this degraded `[]` value becomes the default writable filter and `RuleService.update` persists it, silently erasing the stored filter and potentially activating the rule without filtering.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(evals): keep corrupted-filter evalua..."](https://github.com/langfuse/langfuse/commit/64b25fcb48d375cc426617935bda156217c2c2f8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55971896)</sub>

**Context used:**

- Knowledge Base — [Evaluation Pipeline (LLM-as-Judge)](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/evaluation-pipeline.md)

### DIFF
--- a/web/src/__tests__/server/unstable-evals-api.servertest.ts
+++ b/web/src/__tests__/server/unstable-evals-api.servertest.ts
@@ -783,6 +783,62 @@ describe("/api/public/unstable evaluators API", () => {
     expect(deleted.status).toBe(200);
   });
 
+  it("does not overwrite a corrupted stored filter when a PATCH renames the rule", async () => {
+    const evaluator = await makeZodVerifiedAPICall(
+      PostUnstableEvaluatorResponse,
+      "POST",
+      "/api/public/unstable/evaluators",
+      {
+        name: "Corrupted filter rename evaluator",
+        prompt: "Judge {{input}}",
+        outputDefinition: numericOutputDefinition,
+        mapping: [{ variable: "input", source: "input" }],
+      },
+      auth,
+    );
+
+    const corruptedFilter = [{ type: "legacy_unsupported", column: "input" }];
+    const nativeRule = await prisma.evaluationRule.create({
+      data: {
+        projectId,
+        name: "corrupted_filter_rename_rule",
+        targetObject: "event",
+        status: "ACTIVE",
+        sampling: 1,
+        delay: 0,
+        timeScope: ["NEW"],
+        filter: corruptedFilter,
+        assignments: {
+          create: {
+            projectId,
+            evaluatorId: evaluator.body.id,
+          },
+        },
+      },
+    });
+
+    const patched = await makeAPICall(
+      "PATCH",
+      `/api/public/unstable/evaluation-rules/${nativeRule.id}`,
+      { name: "renamed_corrupted_filter_rule" },
+      auth,
+    );
+    // The rule's own filter still can't be re-validated, so this patch is
+    // rejected rather than applied — but critically, it must fail loudly
+    // instead of silently persisting the degraded (empty) public read
+    // representation over the real stored filter.
+    expect(patched.status).toBe(400);
+
+    await expect(
+      prisma.evaluationRule.findUniqueOrThrow({
+        where: { id: nativeRule.id },
+      }),
+    ).resolves.toMatchObject({
+      name: "corrupted_filter_rename_rule",
+      filter: corruptedFilter,
+    });
+  });
+
   it("rejects create bodies that combine evaluators with the deprecated mapping alias", async () => {
     const evaluator = await makeZodVerifiedAPICall(
       PostUnstableEvaluatorResponse,
--- a/web/src/__tests__/server/unstable-evals-api.servertest.ts
+++ b/web/src/__tests__/server/unstable-evals-api.servertest.ts
@@ -719,6 +719,70 @@ describe("/api/public/unstable evaluators API", () => {
     expect(deleted.status).toBe(200);
   });
 
+  it("still allows disabling and deleting an observation rule with a corrupted stored filter", async () => {
+    const evaluator = await makeZodVerifiedAPICall(
+      PostUnstableEvaluatorResponse,
+      "POST",
+      "/api/public/unstable/evaluators",
+      {
+        name: "Corrupted filter rule evaluator",
+        prompt: "Judge {{input}}",
+        outputDefinition: numericOutputDefinition,
+        mapping: [{ variable: "input", source: "input" }],
+      },
+      auth,
+    );
+
+    // Mirrors a migrated rule whose stored filter no longer parses under the
+    // current public schema (e.g. an unrecognized filter `type`).
+    const corruptedFilter = [{ type: "legacy_unsupported", column: "input" }];
+    const nativeRule = await prisma.evaluationRule.create({
+      data: {
+        projectId,
+        name: "corrupted_filter_rule",
+        targetObject: "event",
+        status: "ACTIVE",
+        sampling: 1,
+        delay: 0,
+        timeScope: ["NEW"],
+        filter: corruptedFilter,
+        assignments: {
+          create: {
+            projectId,
+            evaluatorId: evaluator.body.id,
+          },
+        },
+      },
+    });
+
+    const patched = await makeAPICall(
+      "PATCH",
+      `/api/public/unstable/evaluation-rules/${nativeRule.id}`,
+      { enabled: false },
+      auth,
+    );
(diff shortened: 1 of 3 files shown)


chelsealong:
Fixed the P1 from greptile: the degraded-filter preservation in \`updatePublicEvaluationRule\` was only applied on enabled/sampling-only patches. A rename (or evaluator-only) PATCH on a corrupted-filter rule skipped that guard, so the degraded `[]` public read shape became the writable filter and `RuleService.update` persisted it, silently erasing the real stored filter.

Generalized the guard to cover any patch that leaves `target`/`filter` untouched, regardless of what else it changes. Such a patch now keeps the raw stored filter bytes and — since that filter still can't pass `RuleService.update`'s re-validation — the request 400s instead of silently corrupting the rule. Added a test asserting the stored filter survives untouched after a rename-only PATCH on a corrupted rule (`web/src/__tests__/server/unstable-evals-api.servertest.ts`).

PR #16449, opened 2026-08-23
fix(api): keep migrated evaluation rules manageable

### LINKED ISSUE #16445: Corrupted evaluation-rule filter prevents disabling or deleting an active rule via public API

wochinge:
🟡 **Live preview: [pr-16449.preview.langfuse.com](https://pr-16449.preview.langfuse.com)**

## Summary

- Allow evaluation-rule reads to represent persisted filters from the public filter union even when a migrated filter no longer matches the rule target.
- Preserve omitted stored filters during patches, while continuing to validate newly supplied filters against their target so rules can be repaired, deactivated, or deleted.
- Add web regression coverage for listing, ordinary patches, valid replacement filters, deactivation, and deletion.
- Impacted package: `web`. Tested with `pnpm --filter web run test unstable-evals-service unstable-evals-queries evaluation-rule-v2-service` and `pnpm run lint`.
- fix https://github.com/langfuse/langfuse/issues/16445
- 
## Major Decisions

- Keep the API schemas unchanged. Read compatibility and strict write validation are separated in server logic so malformed persisted state remains manageable without permitting new invalid state.

## Review Focus

- Check the boundary between target-independent read adaptation and target-specific validation for create and replacement-filter updates.
- Check that omitted patch fields remain omitted when passed into the rule service.

This PR keeps migrated evaluation rules manageable when their stored filters no longer match current target-specific schemas.

- Serializes legacy stored filters through the public read API using the broader public filter schema.
- Preserves unsupported stored filters during unrelated partial updates.
- Allows migrated rules to be repaired, disabled, or deleted.
- Adds unit coverage for listing and managing migrated rules.
</details>

The PR appears safe to merge with no actionable correctness or security defects identified.

The changed read and update paths preserve migrated filters for lifecycle operations while retaining validation when callers explicitly replace the filter or target.
</details>

```mermaid
flowchart TD
  A[Migrated evaluation rule] --> B{Public API operation}
  B -->|List or read| C[Map legacy filter to public representation]
  B -->|Unrelated patch| D[Preserve stored filter without revalidating it]
  B -->|Replace filter or target| E[Validate and persist supplied configuration]
  B -->|Disable or delete| F[Allow lifecycle operation]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(api): keep migrated evaluation rules..."](https://github.com/langfuse/langfuse/commit/d0fd27b93196a0970365b2e32b6fa73c3ab967c8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56007318)</sub>

### DIFF
--- a/web/src/__tests__/server/unit/unstable-evals-service.servertest.ts
+++ b/web/src/__tests__/server/unit/unstable-evals-service.servertest.ts
@@ -84,7 +84,10 @@ vi.mock("@langfuse/shared/src/db", async () => {
       findFirst: vi.fn(),
       create: vi.fn(),
       delete: vi.fn(),
+      deleteMany: vi.fn(),
+      update: vi.fn(),
     },
+    jobExecution: { deleteMany: vi.fn() },
     $transaction: vi.fn(),
     evaluationRuleEvaluatorAssignment: { update: vi.fn(), create: vi.fn() },
   };
@@ -144,10 +147,16 @@ import {
   CodeEvalJobConfigError,
   assertCodeEvalRuleCanRun,
 } from "@/src/features/evals/server/codeEvalJobConfigValidation";
-import { createPublicEvaluationRule } from "@/src/features/evals/server/unstable-public-api/evaluation-rule-service";
+import {
+  createPublicEvaluationRule,
+  deletePublicEvaluationRule,
+  listPublicEvaluationRules,
+  updatePublicEvaluationRule,
+} from "@/src/features/evals/server/unstable-public-api/evaluation-rule-service";
 import {
   findPublicV2EvaluationRule,
   findPublicV2EvaluatorInFamilyOrThrow,
+  listPublicEvaluationRulePage,
 } from "@/src/features/evals/server/unstable-public-api/queries";
 import { MAX_ACTIVE_EVALUATION_RULES } from "@/src/features/evals/v2/server/rules/ruleErrors";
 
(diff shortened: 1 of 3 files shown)


wochinge:
What to doubt: the read adapter deliberately accepts the target-independent public filter union, while create and supplied PATCH filters still use target-specific validation. Please scrutinize that boundary, plus the use of `undefined` to preserve omitted patch fields.

PR #16469, opened 2026-08-24
fix(chatml): render reasoning block when completion is empty string

### LINKED ISSUE #16464: bug(chatml): legacy OpenAI outputs with empty `completion` lose the `reasoning` value during normalization
### Problem

For a generation whose output object is `{"completion": "", "reasoning": "..."}` (a legitimate shape for reasoning models that produce reasoning but deliberately emit no assistant text), the observation panel's pretty view omits the reasoning block. The data itself is intact: fetching the same trace via the public API returns the full reasoning string, so this is purely a rendering issue in the chatml normalization pipeline.

Both prior PRs (#15370 `fix/reasoning-block-empty-completion` and #16158 `agent/fix-empty-completion-reasoning`) attempted this fix and were closed without merging. The diagnosis they shared is correct, but the fix needs to live in a different place.

### Root cause

`packages/shared/src/utils/IORepresentation/chatML/types.ts` defines `BaseChatMlMessageSchema` with the following fields:

```ts
{
  role: z.string().optional(),
  name: z.string().optional(),
  content: z.union([...]).nullish(),
  audio: OpenAIOutputAudioSchema.optional(),
  additional_kwargs: z.record(z.string(), z.any()).optional(),
  tools: z.array(ToolDefinitionSchema).optional(),
  tool_calls: z.array(ToolCallSchema).optional(),
  tool_call_id: z.string().optional(),
  // Thinking/reasoning content from models
  thinking: z.array(ThinkingContentPartSchema).optional(),
  redacted_thinking: z.array(RedactedThinkingContentPartSchema).optional(),
}
```

It has `thinking` and `redacted_thinking` for the OpenAI Responses API reasoning content, but **no `reasoning` field for the legacy OpenAI Chat Completions `{"completion", "reasoning"}` shape**.

Then `ChatMlMessageSchema` is `BaseChatMlMessageSchema.refine(...).transform(({ additional_kwargs, ...other }) => ({ ...other, ...additional_kwargs })).transform(({ role, name, content, audio, type, tools, tool_calls, tool_call_id, thinking, redacted_thinking, ...other }) => ({ role, name, content, audio, type, tools, tool_calls, tool_call_id, thinking, redacted_thinking, ...(Object.keys(other).length === 0 ? {} : { json: other }) }))`.

The second `.transform` destructures the known fields and wraps everything else in `json: { ...other }`. So `{role, content: "", completion: "", reasoning: "..."}` becomes `{role, content: "", json: {completion: "", reasoning: "..."}}`. The reasoning value is still there — but it's now nested under the `json` key.

The downstream rendering then has to display the `json` field as an expandable row. The inner `completion: ""` row gets filtered by `shouldShowValue` (empty string is filtered), and the inner `reasoning: "..."` row should be shown — but in practice the user reports seeing only the completion row, suggesting the rendering path that should expose the inner `reasoning` field is not finding it.

The actual user-visible fix needs to add `reasoning: z.string().nullish()` (or equivalent) to `BaseChatMlMessageSchema` so the OpenAI adapter's `normalizeMessage` output for `{"completion": "", "reasoning": "..."}` becomes a first-class message with `content: ""` and `reasoning: "..."` (not nested in `json`). This matches the pattern of how `thinking` and `redacted_thinking` are handled for the Responses API.

### Proposed solution

In `packages/shared/src/utils/IORepresentation/chatML/types.ts`:

1. Add `reasoning: z.string().nullish()` to `BaseChatMlMessageSchema` next to `thinking` and `redacted_thinking`.
2. Extend `ChatMlMessageSchema`'s second `.transform` to pass through the new `reasoning` field (don't include it in the `...other` group that gets wrapped in `json`).
3. Update the OpenAI Chat Completions adapter (`packages/shared/src/utils/chatml/adapters/openai.ts`) to construct the `reasoning` field when normalizing an output of shape `{completion: "", reasoning: "..."}` (and similar). The expected normalized message is `{role: "assistant", content: "", reasoning: "..."}`.

### Alternatives considered

- **Move `reasoning` to `json` and fix the renderer** — addresses only the rendering, not the data model. A future refactor that re-parses the message will hit the same shape.
- **Add a runtime fallback in `hasRenderableContent` for empty `content` plus a `json` field that has a `reasoning` key** — fragile, hides the data model problem.
- **Add `reasoning` only to the legacy-output handling, not to `BaseChatMlMessageSchema`** — duplicates the field, increases drift surface.

### Backward compatibility

Purely additive schema change. Existing messages (no `reasoning` field) continue to validate identically. New messages that have `reasoning` (either directly or via the legacy OpenAI path) will surface the field to the renderer.

### Acceptance criteria

- A generation output of `{"completion": "", "reasoning": "..."}` renders the reasoning block in the observation panel's pretty view, regardless of whether `completion` is empty or non-empty.
- A generation output of `{"completion": "abc", "reasoning": "..."}` continues to render both blocks (no regression).
- A generation output with no `reasoning` field continues to render only the `completion` block (no regression).
- `pnpm --filter @langfuse/shared exec tsc --noEmit` is clean.
- A unit test in `worker/src/__tests__/chatml/adapters/openai.test.ts` (or the equivalent) covers: input `{completion: "", reasoning: "..."}` produces a normalized message with both `content: ""` and a non-empty `reasoning` field, and the message is `shouldRenderMessage` true in the chat-message-utils.

Harsh23Kashyap:
Fixes #16464.

## Problem

Legacy OpenAI Chat Completions outputs of shape `{completion: "", reasoning: "..."}` (an increasingly common shape for reasoning models that produce reasoning but emit no assistant text) lost the `reasoning` value during normalization.

The OpenAI adapter's `normalizeMessage` and `preprocessData` both fell through this shape unchanged. The schema parser then routed the `reasoning` value into the `json` overflow bag because `BaseChatMlMessageSchema` did not declare a `reasoning` field. The chatml observation panel's pretty view reads only first-class fields, so the reasoning block was hidden in the JSON tree rather than rendered.

## Prior attempts

Two previous PRs closed without merging attempted this fix:
- #15370 `fix/reasoning-block-empty-completion` (closed 2026-08-22)
- #16158 `agent/fix-empty-completion-reasoning` (closed 2026-08-16)

Both had the correct diagnosis. They were scoped to the renderer / JSON overflow path; this PR addresses the same problem at the data model so the field is first-class on the message rather than hidden in `json`.

## Fix

- **`packages/shared/src/utils/IORepresentation/chatML/types.ts`** — add `reasoning: z.string().nullish()` to `BaseChatMlMessageSchema`; update the second `.transform` in `ChatMlMessageSchema` to pass `reasoning` through alongside `thinking` and `redacted_thinking`.
- **`packages/shared/src/utils/chatml/adapters/openai.ts`** — handle the legacy `{completion, reasoning}` shape in both `normalizeMessage` (produces `{role, content, reasoning}` when both are strings) and `preprocessData` (routes the legacy shape into `normalizeMessage`).
- **`worker/src/__tests__/chatml/adapters/openai.test.ts`** — three regression tests: reasoning surfaced when completion is empty, non-empty completion preserved alongside reasoning, and completion-only output continues to render without reasoning.

## Testing

- `pnpm --filter worker exec vitest run src/__tests__/chatml/adapters/openai.test.ts` → 30/30 pass
- `pnpm --filter worker exec vitest run src/__tests__/chatml/` → 135/135 pass
- prettier --check on all 3 changed files: clean
- eslint on changed files: no errors

## Compatibility

- `reasoning: z.string().nullish()` accepts `string | null | undefined`, matching the existing `content: z.union([...]).nullish()` pattern in the same schema.
- `normalizeMessage` only sets `reasoning` when `reasoning` is a non-empty string — pre-existing outputs without the field continue to render as before.
- No new public API; this only widens the chatml internal data model.

This PR promotes reasoning from legacy `{completion, reasoning}` OpenAI-compatible outputs into the typed ChatML message model.
- Adds a nullable first-class `reasoning` field to the shared ChatML schema.
- Normalizes legacy completion outputs into assistant messages carrying content and reasoning.
- Adds worker regression coverage for empty, non-empty, and completion-only variants.
- The emitted field is not connected to the renderer’s existing thinking representation, so reasoning-only messages still do not render as intended.
</details>

This PR should not merge until the normalized reasoning value is connected to the renderer’s recognized reasoning representation.

The new field survives normalization, but current rendering and filtering only recognize thinking arrays, causing the target reasoning-only output to appear as generic JSON or disappear in conversation mode.

**Files Needing Attention:** packages/shared/src/utils/chatml/adapters/openai.ts, packages/shared/src/utils/IORepresentation/chatML/types.ts
</details>

```mermaid
flowchart LR
  A[Legacy completion and reasoning output] --> B[OpenAI normalizeMessage]
  B --> C[Top-level reasoning string]
  C --> D[Shared ChatML schema]
  D --> E{Display mode}
  E -->|All| F[Generic JSON table]
  E -->|Conversation| G[Message filtered out]
  C -. Expected .-> H[Dedicated reasoning block]
```
</details>

`````markdown
### Issue 1
packages/shared/src/utils/chatml/adapters/openai.ts:240-242
**Reasoning bypasses rendering path**

When a legacy output has an empty `completion` and a non-empty `reasoning`, this branch emits a top-level reasoning string while the renderer and conversation filter recognize only `thinking` and `redacted_thinking` arrays, causing the reasoning to appear as generic JSON in the all-content view and disappear entirely in conversation mode.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(chatml): render reasoning block when..."](https://github.com/langfuse/langfuse/commit/7c8550762e1ae88b1e1f7067035bff11cee3c47d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56195854)</sub>

### DIFF
--- a/packages/shared/src/utils/IORepresentation/chatML/types.ts
+++ b/packages/shared/src/utils/IORepresentation/chatML/types.ts
@@ -269,6 +269,11 @@ export const BaseChatMlMessageSchema = z
     // Thinking/reasoning content from models
     thinking: z.array(ThinkingContentPartSchema).optional(),
     redacted_thinking: z.array(RedactedThinkingContentPartSchema).optional(),
+    // Legacy OpenAI Chat Completions `{"completion": "", "reasoning": "..."}`
+    // shape (no Responses API reasoning item wrapper). The OpenAI adapter
+    // normalizes this into a top-level `reasoning` field on the message
+    // rather than nesting it under `json`.
+    reasoning: z.string().nullish(),
   })
   .loose();
 
@@ -337,6 +342,7 @@ export const ChatMlMessageSchema = BaseChatMlMessageSchema.refine(
       tool_call_id,
       thinking,
       redacted_thinking,
+      reasoning,
       ...other
     }) => ({
       role,
@@ -349,6 +355,7 @@ export const ChatMlMessageSchema = BaseChatMlMessageSchema.refine(
       tool_call_id,
       thinking,
       redacted_thinking,
+      reasoning,
       ...(Object.keys(other).length === 0 ? {} : { json: other }),
     }),
   );
--- a/packages/shared/src/utils/chatml/adapters/openai.ts
+++ b/packages/shared/src/utils/chatml/adapters/openai.ts
@@ -228,6 +228,21 @@ function normalizeMessage(msg: unknown): Record<string, unknown> {
     };
   }
 
+  // Handle legacy Chat Completions shape {completion, reasoning}.
+  // Reasoning models (e.g., via OpenRouter) may emit a reasoning trace
+  // alongside an empty completion; surface `reasoning` as a top-level
+  // field so the renderer shows the reasoning block rather than nesting
+  // it under `json` (#15343).
+  if (typeof working.completion === "string") {
+    return {
+      role: working.role ?? "assistant",
+      content: working.completion,
+      ...(typeof working.reasoning === "string" && working.reasoning.length > 0
+        ? { reasoning: working.reasoning }
+        : {}),
+    };
+  }
+
   // Convert direct function call message to tool_calls array format
   // Format: { type: "function_call", name: "...", arguments: {...}, call_id: "..." }
   // Convert to: { role: "assistant", tool_calls: [{ id, name, arguments, type }] }
@@ -494,6 +509,18 @@ function preprocessData(data: unknown): unknown {
 
   // Single message
   if (typeof data === "object" && "role" in data) {
+    return normalizeMessage(data);
+  }
+
+  // Legacy {completion, reasoning} output (OpenRouter broadcast and
+  // similar). Promote the reasoning value to a top-level message field so
+  // the chatml renderer can show the reasoning block rather than nesting
+  // it under `json` (#15343).
+  if (
+    typeof data === "object" &&
+    !Array.isArray(data) &&
+    typeof (data as Record<string, unknown>).completion === "string"
+  ) {
     return normalizeMessage(data);
   }
 
--- a/worker/src/__tests__/chatml/adapters/openai.test.ts
+++ b/worker/src/__tests__/chatml/adapters/openai.test.ts
@@ -710,4 +710,46 @@ describe("OpenAI Adapter", () => {
       );
     });
   });
+
+  describe("legacy {completion, reasoning} output (#15343)", () => {
+    it("should surface reasoning as a top-level field when completion is empty", () => {
+      // Regression test for #15343: reasoning models via OpenRouter may emit
+      // a reasoning trace alongside an empty completion. The reasoning value
+      // must be a top-level field on the normalized message so the renderer
+      // can show the reasoning block rather than nesting it under `json`.
+      const output = {
+        completion: "",
+        reasoning: "thought process",
+      };
+
+      const result = normalizeOutput(output, { framework: "openai" });
+      expect(result.success).toBe(true);
+      expect(result.data?.[0].content).toBe("");
+      expect(result.data?.[0].reasoning).toBe("thought process");
+      // The reasoning value should not be wrapped in `json` — it must reach
+      // the renderer as a first-class field.
+      expect(result.data?.[0].json).toBeUndefined();
+    });
+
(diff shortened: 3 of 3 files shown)


Steffen911:
slop

PR #16471, opened 2026-08-24
fix(in-app-agent): persist signed Bedrock reasoning so follow-up turns can cache-read

LysanderKie:
🟡 **Live preview: [pr-16471.preview.langfuse.com](https://pr-16471.preview.langfuse.com)**

## Summary
- Follow-up of the merged Bedrock prompt-cache work: in-loop cache hits already work, but the next user turn often rewrites conversation history instead of reading it.
- Persist Bedrock thinking signatures and send those signed blocks back on the next model call, so the conversation prefix can stay byte-identical across turns.
- Unsigned reasoning (older turns without a signature) stays display-only so existing threads do not 400.

This should cut **cache-write tokens on continued chats**. After the prefix is stable, a follow-up writes the new user message instead of rewriting prior tool calls and results at 1.25×. It does **not** yet fix the other prefix mismatches (tool results flipping `json` → `text`, `proposeRedirect` being stripped).

## Test plan
- [ ] Restart worker and send a two-turn assistant chat on the same page
- [ ] Confirm turn 2's first `invoke-model` includes prior `reasoning` parts with `providerOptions.bedrock.signature`
- [ ] Confirm `cache_read_input_tokens` on that call covers more than tools+system when the previous turn is still within 5 minutes
- [ ] Confirm an older conversation without signatures still loads and does not 400

## Impacted packages
- `packages/shared`
- `worker`
- `web` (architecture note only)

## Verification
- `pnpm --filter @langfuse/shared run test src/in-app-agent/server/persistence.test.ts` — `Tests 4 passed (4)`
- `pnpm --filter worker run test src/features/in-app-agent/runtime/replayReasoning.test.ts src/features/in-app-agent/runtime/agent.test.ts` — `Tests 26 passed (26)`
- `pnpm --filter web run test in-app-agent-persistence.servertest.ts` — `Tests 25 passed (25)`
- Pre-commit `turbo run lint` — `Tasks: 7 successful, 7 total`


LysanderKie:
Worth doubting: whether Mastra still drops `role: reasoning` after a library bump (the processor is compensating for that), and whether a missing or late `reasoning-signature` part leaves us replaying unsigned text. Unsigned history is still stripped on purpose — if that filter is wrong, follow-up turns 400 instead of just missing cache.

PR #16479, opened 2026-08-24
fix(automations): require fresh tokens for dispatch URL changes

hassiebp:
🟡 **Live preview: [pr-16479.preview.langfuse.com](https://pr-16479.preview.langfuse.com)**

## Summary

Prevent GitHub repository-dispatch automations from reusing a stored personal access token when an existing dispatch URL changes.

The server now requires a fresh non-empty token for actual destination changes while preserving the existing token for updates that retain the same endpoint. The edit form mirrors that rule so users get an inline requirement before submission.

## Changes

- Enforce the token-destination invariant in the shared GitHub dispatch configuration helper before stored-token fallback.
- Compare canonical URL forms so hostname case and default HTTPS-port formatting do not count as destination changes.
- Require and explain the replacement token in the edit form when the endpoint changes.
- Cover rejection, token rotation with a changed URL, equivalent endpoint preservation, and client-side validation.

## Verification

- `pnpm --filter web run test 'src/__tests__/server/automations-trpc.servertest.ts'`
  - `Tests 44 passed (44)`
- `pnpm --filter web run test-client 'src/features/automations/components/actions/GitHubDispatchActionHandler.clienttest.ts'`
  - `Tests 2 passed (2)`
- Targeted ESLint passed with `--max-warnings 0`.
- Targeted Prettier check passed.
- `git diff --check` passed.
- Browser-exercised the GitHub Dispatch create/edit flow in the PR preview. The final follow-up state could not be re-verified because the public endpoint continued serving the prior pod revision after the new image published; the local Kubernetes context was unavailable for rollout inspection.

The full web typecheck and repository lint were not used as sign-off gates because the checkout has unrelated baseline failures: stale shared-package exports and a missing `@repo/storybook-play-requires-test-name` ESLint rule.

## What to doubt

The security-sensitive edge is the endpoint identity comparison: verify that WHATWG URL canonicalization matches the supported GitHub.com and GitHub Enterprise URL forms.

### DIFF
--- a/web/src/__tests__/server/automations-trpc.servertest.ts
+++ b/web/src/__tests__/server/automations-trpc.servertest.ts
@@ -2832,7 +2832,7 @@ describe("automations trpc", () => {
           type: "GITHUB_DISPATCH",
           config: {
             type: "GITHUB_DISPATCH",
-            url: "https://api.github.com/repos/owner/repo/dispatches",
+            url: "https://github.com/api/v3/repos/owner/repo/dispatches",
             eventType: "original-event-type",
             githubToken: encryptedToken,
             displayGitHubToken: "ghp_...xyz",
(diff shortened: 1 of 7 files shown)


hassiebp:
Reviewer doubt: the guard compares the persisted URL representation exactly. Please verify that this matches how endpoint identity is defined for GitHub.com and Enterprise URLs, including any normalization or formatting cases.

PR #16486, opened 2026-08-24
feat(evals): add dedicated evaluator execution columns

wochinge:
🟡 **Live preview: [pr-16486 app preview](https://pr-16486.preview.langfuse.com)**

## Summary

The new eval UX no longer
- uses job execution postgres table to show execution logs -> Uses the `events` table instead
- directly links to the scores created by the particular evaluator (based on evaluator ID)

Currently this works based on inspecting the metadata -> this is expensive to read.

This PR
- adds `evaluator_id`, `evaluation_rule_id` and `evaluator_test` to the `events_full`, `events_core`, `scores` table
  - `evaluator_test` is needed because we want to exclude test calls from cost calculations
- adds bloom indexes on `evaluator_id` / `evaluation_rule_id` to `events_core`, `scores` (not needed on `events_full`)
- does not run a backfill -> instead we use `default` to read the data from the `metadata` if needed

Important to know:
- The `evaluator` / `rules` table in the UI only queries the last 7 days of data per evaluator / rule (so historic data is not super important) -> Only if people inspect the full history with a long timespan the `default` would kick i

**Decisions**
- no index on `events_full` for now as we don't query that table
- no proper backfill of the attributes as I'd expect us to mostly look at the last 7 days. Besides `evaluation_rule_id` (which is equivalent to historic `job_configuration_id`s) none of these have historic values anyway

The PR promotes evaluator execution metadata into dedicated ClickHouse columns and updates ingestion, query, filtering, and cost paths to use them.

- Adds logically backfilled evaluator, rule, and test-execution columns plus Bloom indexes.
- Keeps `events_full`, `events_core`, and their materialized view aligned in clustered and unclustered migrations.
- Derives dedicated identifiers during event and score ingestion while preserving the legacy rule-ID fallback.
- Separates evaluator and rule dimensions across score and event query mappings.
</details>

The PR appears safe to merge with no concrete correctness, migration, or compatibility failures identified.

The schema, materialized-view ordering, ingestion derivation, legacy rule fallback, evaluator-test exclusion, and query mappings remain internally aligned across the changed paths.
</details>

```mermaid
flowchart LR
  A[Evaluation event metadata] --> B[Ingestion normalization]
  B --> C[evaluator_id]
  B --> D[evaluation_rule_id]
  B --> E[evaluator_execution_is_test]
  C --> F[events_full]
  D --> F
  E --> F
  F --> G[events_core materialized view]
  B --> H[scores]
  G --> I[Event filters and cost queries]
  H --> J[Score filters and analytics]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(evals): add dedicated evaluator exe..."](https://github.com/langfuse/langfuse/commit/a00d2372cdda3491316b4ce38d41040dbf384859) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56294233)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Observability data platform](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/observability-data-platform.md)
</details>


wochinge:
Reviewer doubt: please scrutinize the migration semantics for historical parts—the new columns use read-through DEFAULT expressions and materialize only skip indexes, not column data. The other deliberate tradeoff is keeping evaluator/rule Bloom indexes on events_core but not events_full, matching the experiment ID pattern.

wochinge:
@claude review

packages/shared/clickhouse/migrations/clustered/0047_add_eval_execution_columns.up.sql
```diff
@@ -0,0 +1,119 @@
...
+  ADD COLUMN IF NOT EXISTS evaluator_execution_is_test Bool DEFAULT has(metadata_names, 'evaluator_test') AND arrayElement(metadata_values, indexOf(metadata_names, 'evaluator_test')) = 'true' AFTER evaluation_rule_id
+  SETTINGS enable_full_text_index = 1, alter_sync = 2;
+ALTER TABLE events_core ON CLUSTER default
+  ADD INDEX IF NOT EXISTS idx_evaluator_id evaluator_id TYPE bloom_filter(0.01) GRANULARITY 1
+  SETTINGS enable_full_text_index = 1, alter_sync = 2;
+ALTER TABLE events_core ON CLUSTER default
+  ADD INDEX IF NOT EXISTS idx_evaluation_rule_id evaluation_rule_id TYPE bloom_filter(0.01) GRANULARITY 1
+  SETTINGS enable_full_text_index = 1, alter_sync = 2;
+ALTER TABLE events_core ON CLUSTER default MATERIALIZE INDEX IF EXISTS idx_evaluator_id SETTINGS enable_full_text_index = 1, mutations_sync = 2;
+ALTER TABLE events_core ON CLUSTER default MATERIALIZE INDEX IF EXISTS idx_evaluation_rule_id SETTINGS enable_full_text_index = 1, mutations_sync = 2;
```
Steffen911: I'm concerned about deriving a default value from metadata _and_ then creating an materializing an index on top of that new column. I'm concerned that this will re-read or re-evaluate all parts, stall the migration, and then we'll see issues on self-hosters if the migration takes minutes or hours.

worker/src/services/IngestionService/index.ts
```diff
@@ -337,7 +339,7 @@ export class IngestionService {
             projectId: eventData.projectId,
             pricingMatchAttributeValues: {
               modelParameters,
-              metadata: eventData.metadata,
+              metadata,
```
Steffen911, line 342: The result would be that we have all the eval stuff in metadata _and_ in the dedicated columns, right? Is that desirable?

And do those records take the regular ingestion flow or are they taking some shortcut? Otherwise, doing things in the Otel Processor might be preferable to make use of @marliessophie parsing logic eventually.

wochinge: Hassieb and I figured that this is better than having to read from the potentially heavy metadata column, no?

packages/shared/clickhouse/migrations/clustered/0047_add_eval_execution_columns.up.sql
wochinge: I had Codex check this:
- the index doesn't materialize the evaluator ID column
- but yes: it'll re-read all old values (not so much an issue for evaluator ID, but rather for `evaluation_rule_id` as that goes way back (as it's the equivalent of `job_configuration_id`)

Simplest would probably be to drop the materialization so that only new columns get it. After 1-2 weeks we should be fine.

worker/src/services/IngestionService/index.ts
Steffen911: @wochinge My question is why we still store it in metadata at all? We could remove the respective keys from the metadata object (unless we already do that and I missed it)?

And then the other question is where we do the parsing. And for me the IngestionService seems rather "late".

packages/shared/clickhouse/migrations/clustered/0047_add_eval_execution_columns.up.sql
Steffen911: @wochinge Thank you for the check. I'll let @sumerman confirm whether he sees any other issues, but dropping the materialization seems preferable to me.

worker/src/services/IngestionService/index.ts
wochinge: > My question is why we still store it in metadata at all? We could remove the respective keys from the metadata object (unless we already do that and I missed it)?

Yeah, of course - will drop

> And then the other question is where we do the parsing. And for me the IngestionService seems rather "late".

Will adjust - let's do it in the same place as the experiment fields

wochinge: pushed. Just one note: for experiments we're currently storing it duplicate (metadata + separate columns): https://cloud.langfuse.com/project/clkpwwm0m000gmm094odg11gi/experiments/results?baseline=cmt8f7dp900bdad0fgp87d9vg&peek=2fa1750f-6545-44bc-a0e2-99b9c36c02b6&traceId=598d7de658bd9f980162b70abc9f2490&timestamp=2026-08-25T08%3A46%3A05.987Z&observation=9ec2486035ff052a

packages/shared/clickhouse/migrations/clustered/0047_add_eval_execution_columns.up.sql
wochinge: dropped the materialization

wochinge:
@claude review

packages/shared/clickhouse/migrations/clustered/0047_add_eval_execution_columns.down.sql
wochinge, line 1: This would require your review @sumerman

worker/src/services/IngestionService/index.ts
wochinge, line 1: and this would require your review @sumerman - see also the discussion with Steffen below

packages/shared/src/server/otel/OtelIngestionProcessor.ts
wochinge, line 1: and this @sumerman

wochinge (commented):
@sumerman Once you're back and recovered - Tagged the exact places that would require review (ignore the rest pls)

packages/shared/src/features/query/dataModel.ts
```diff
@@ -852,11 +851,19 @@ const createScoreSpecificDimensions = (
     description: "Name of the score (e.g., accuracy, toxicity).",
   },
   evaluatorId: {
-    sql: `coalesce(nullIf(${tableAlias}.metadata['${EvalExecutionMetadataKey.EVALUATOR_ID}'], ''), ${tableAlias}.metadata['${EvalExecutionMetadataKey.JOB_CONFIGURATION_ID}'])`,
+    sql: `${tableAlias}.evaluator_id`,
```
sumerman: this may potentially break charts for customers if we aren't backfilling. could probably slap another `coalesce` on top

worker/src/services/IngestionService/index.ts
```diff
@@ -355,15 +366,16 @@ export class IngestionService {
...
 
+    const metadataToPersist = eventData.evaluationContext
+      ? removePromotedEvalMetadata(metadata)
+      : metadata;
+
     // Flatten raw metadata first (before stringification destroys nested structure)
-    const flattened = eventData.metadata
-      ? flattenJsonToPathArrays(eventData.metadata)
-      : { names: [], values: [] };
+    const flattened = flattenJsonToPathArrays(metadataToPersist);
```
sumerman: why no default now?

worker/src/services/IngestionService/index.ts
sumerman: >Will adjust - let's do it in the same place as the experiment fields

I still see evaluator_* manipulation in this file although from your earlier discussion it looked like you were going to move them. Is that intended? (I don't have an opinion here, just double checking).

sumerman: Also here and in OtelProcessor. I am not 100% sure removing attributes is entirely safe. Depending on when migration executes vs the new worker first starts executing we might end up with rows that already don't have metadata but MV didn't propagate columns properly either.

packages/shared/src/server/repositories/events.ts
```diff
@@ -377,6 +373,13 @@ const PUBLIC_API_TRACES_COLUMN_MAPPING = createPublicApiTracesColumnMapping(
 // not joined from a separate observations table (with 'o' prefix). We need to remap these.
 const TRACES_FROM_EVENTS_UI_COLUMN_DEFINITIONS =
   tracesTableUiColumnDefinitions.map((col) => {
+    if (col.uiTableId === "evaluatorId") {
```
sumerman: looks like a weird hack. is it v3 backcompat?

sumerman (commented):
Overall looks reasonable to me! Nice approach with DEFAULT! For attribute removal I didn't see tests asserting that nothing but necessary attributes are removed. Plus see my other comments.

worker/src/services/IngestionService/index.ts
wochinge: this is an issue with the diff - we add the default in line 316 now

worker/src/services/IngestionService/index.ts
wochinge: yeah, fair. I'll drop the removal - that's the same behavior as we do with experiments then.

packages/shared/src/server/repositories/events.ts
wochinge: Yep, was v3 compat but actuallly not needed as any legacy stuff uses the old UI anyway. Will drop

wochinge:
@claude review

PR #16500, opened 2026-08-24
fix(automations): surface prompt-version webhook enqueue failures instead of swallowing them

### LINKED ISSUE #16494: bug(automations): Execution logs get stuck on PENDING when jobs fail.
### Describe the bug

Prompt-version automations (webhook / Slack actions) can be **permanently lost** with a stuck `PENDING` automation-execution record, because:

1. The `automationExecutions` row is created **before** the webhook job is enqueued, and
2. any error inside the per-trigger loop — including the enqueue failure itself — is caught, logged, and **swallowed**, so the entityChange job completes successfully and BullMQ never retries.

```ts
// worker/src/features/entityChange/promptVersionProcessor.ts (~L107–112)
} catch (error) {
  logger.error(`Error processing trigger ${trigger.id} ... ${error}`);
  // Continue processing other triggers instead of failing the entire operation
}
```

```ts
// same file, enqueueAutomationAction() (~L156–199)
await prisma.automationExecution.create({
  data: { id: executionId, ..., status: ActionExecutionStatus.PENDING, ... },   // ← first
});

logger.debug(`Created automation execution ${executionId} ...`);

await WebhookQueue.getInstance()?.add(QueueName.WebhookQueue, { ... });         // ← second
```

Failure scenario: `automationExecution.create` succeeds, then `WebhookQueue.getInstance()?.add(...)` throws (Redis blip / connection drop). The throw propagates into the per-trigger catch, gets logged, and is swallowed — the outer processor completes normally, so the configured webhook/Slack action for that prompt version **never fires**: no retry, no DLQ entry. The `automationExecution` row remains `PENDING` forever, permanently skewing the executions history users see in the Automations UI. The same swallow also hides `InternalServerError`s thrown by `getAutomations()` (which runs *inside* `enqueueAutomationAction`).

### Steps to reproduce

1. Create an automation with a Webhook action triggered on prompt version changes.
2. Point Redis at a proxy that accepts the initial connection but drops `XADD`s (or temporarily stop Redis after the Postgres write), then publish a new prompt version that matches the trigger.
3. Observe: `automation_executions` gains a row with `status = 'PENDING'`; no webhook delivery is ever attempted; the entityChange job logs "Error processing trigger …" and completes successfully; no retry happens after Redis recovers.
4. The Automation's execution list permanently shows a pending execution that never progresses.

Note the asymmetry: the outer processor *does* rethrow top-level errors ("Re-throw to trigger retry mechanism"), so only failures below the per-trigger catch are affected — which is exactly where transient Redis failures land.

Self-hosted

Current `main` @ `3c3ca18eed76b164b418776d8d93cc1590e1d65b`

N/A (server-side worker behavior)

### Additional information

**Suggested fix** (any of):

* Enqueue before creating the row, and mark the row `PENDING` via a follow-up update once enqueued succeeds; or
* Wrap create+enqueue in a try/catch that deletes (or marks FAILED) the just-created execution when `WebhookQueue.add` throws, then **rethrow** so the entityChange job's retry budget applies; or
* Collect per-trigger failures and throw an aggregate after the loop instead of swallowing, keeping the loop's continue-on-error semantics for genuinely per-trigger problems but letting infra failures bubble.

Optionally add a sweeper that fails automations stuck in `PENDING` beyond a threshold, mirroring other status reconciliation in the codebase.

**Non-duplication:** searched open/closed issues and PRs for "automationExecution stuck pending", "webhook queue automation lost", "prompt automation webhook not firing" — no existing report.

chelsealong:
## What does this PR do?

Prompt-version automations (webhook / Slack actions) could permanently lose a
webhook delivery. `enqueueAutomationAction` in
`worker/src/features/entityChange/promptVersionProcessor.ts` creates an
`automationExecution` row as `PENDING` and then enqueues the webhook job. If
the enqueue call threw (e.g. a Redis blip), the error propagated into the
per-trigger `catch` block in `promptVersionProcessor`, which only logged it
and continued — so the entityChange job completed "successfully", BullMQ
never retried, and the execution row was left stuck `PENDING` forever with no
webhook ever delivered.

This PR:
- Wraps the webhook enqueue call in a `try`/`catch`. On failure it marks the
  already-created execution row `ERROR` (with the error message and
  `finishedAt`) instead of leaving it stuck `PENDING`, then rethrows.
- Changes the per-trigger loop in `promptVersionProcessor` to collect errors
  instead of only logging them, and throws an aggregate error after
  processing all triggers. This preserves the existing "one bad trigger
  doesn't block the others" behavior while letting real failures propagate
  so BullMQ's retry budget applies (the outer `catch` already rethrows to
  trigger retries).

Fixes #16494

## Type of change

## Mandatory Tasks

## Checklist

- I have checked that new and existing unit tests pass locally with my changes

## Test plan

Added a new test to the existing suite,
`worker/src/__tests__/promptVersionProcessor.test.ts`, that mocks
`WebhookQueue.getInstance().add` to reject, then asserts:
1. `promptVersionProcessor` rejects (so BullMQ retries).
2. The `automationExecution` row ends up `ERROR`, not stuck `PENDING`.

Confirmed the new test fails against the pre-fix code (`git checkout HEAD~1 --
worker/src/features/entityChange/promptVersionProcessor.ts`, keeping the new
test):

```
 × marks the execution as ERROR and rethrows when enqueueing the webhook job fails 54ms (retry x3)
AssertionError: promise resolved "undefined" instead of rejecting
```

With the fix restored, the full suite passes:

```
$ pnpm --filter worker run test src/__tests__/promptVersionProcessor.test.ts
 ✓ src/__tests__/promptVersionProcessor.test.ts (10 tests) 287ms
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

Also ran, both clean:

```
$ pnpm --filter worker run lint
$ pnpm --filter worker run typecheck
```

## AI disclosure

This change (diagnosis, fix, and test) was drafted by an AI coding agent and
reviewed before opening this PR.

This PR marks prompt-version automation executions as failed when webhook enqueueing throws and propagates per-trigger failures so BullMQ can retry the entity-change job.

- Adds aggregate failure propagation after all matching triggers are processed.
- Updates failed enqueue executions from `PENDING` to `ERROR` before rethrowing.
- Adds a regression test for a rejected webhook queue add.
</details>

The PR should not merge until unavailable queues are surfaced and retries are made idempotent for triggers that already enqueued successfully.

A null queue still silently loses deliveries and leaves PENDING executions, while aggregate retries can create duplicate executions and external deliveries after partial success.

**Files Needing Attention:** worker/src/features/entityChange/promptVersionProcessor.ts
</details>

```mermaid
flowchart TD
  E[Prompt-version entity-change job] --> T[Process matching triggers]
  T --> S[Successful trigger enqueues delivery]
  T --> F[Another trigger fails]
  F --> A[Throw AggregateError]
  A --> R[BullMQ retries whole entity-change job]
  R --> T
  S --> D[Duplicate execution and delivery on retry]
```
</details>

`````markdown
### Issue 1
worker/src/features/entityChange/promptVersionProcessor.ts:193
**Unavailable queue bypasses failure handling**

When `WebhookQueue.getInstance()` returns `null` because Redis is unavailable or unconfigured, optional chaining skips `add` without throwing, so the execution remains `PENDING`, the delivery is lost, and the entity-change job completes without retrying.

```suggestion
    const webhookQueue = WebhookQueue.getInstance();
    if (!webhookQueue) throw new Error("Webhook queue is unavailable");
    await webhookQueue.add(QueueName.WebhookQueue, {
```

### Issue 2
worker/src/features/entityChange/promptVersionProcessor.ts:120-124
**Partial success creates duplicate deliveries**

When one matching trigger enqueues successfully and another trigger fails, the aggregate error retries the entire entity-change job without an idempotency guard. Each retry creates fresh execution and webhook job IDs for the already-successful trigger, causing duplicate execution records and duplicate webhook or Slack deliveries.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(automations): surface prompt-version..."](https://github.com/langfuse/langfuse/commit/79b3231de8f08c91455c7ab151b19d5a773f1e15) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56353746)</sub>

**Context used:**

- Knowledge Base — [Asynchronous processing](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/asynchronous-processing.md)
- Knowledge Base — [Integration delivery and notifications](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/integration-delivery-and-notifications.md)

### DIFF
--- a/worker/src/__tests__/promptVersionProcessor.test.ts
+++ b/worker/src/__tests__/promptVersionProcessor.test.ts
@@ -445,4 +445,84 @@ describe("promptVersionChangeWorker", () => {
     expect(executions).toHaveLength(1);
     expect(webhookQueueAddMock).toHaveBeenCalledTimes(1);
   });
+
+  it("does not skip a legitimate execution for a different event action on the same prompt id", async () => {
+    const promptId = v4();
+    const actionId = v4();
+    await prisma.action.create({
+      data: {
+        id: actionId,
+        projectId,
+        type: ActionType.WEBHOOK,
+        config: {
+          type: "WEBHOOK",
+          url: "https://webhook.example.com/test",
+          headers: {},
+          method: "POST",
+        },
+      },
+    });
+
+    const triggerId = v4();
+    await prisma.trigger.create({
+      data: {
+        id: triggerId,
+        projectId,
+        eventSource: TriggerEventSource.Prompt,
+        eventActions: [],
+        status: JobConfigState.ACTIVE,
+        filter: [],
+      },
+    });
+
+    const automationId = v4();
+    await prisma.automation.create({
+      data: {
+        id: automationId,
+        name: `automation-${v4()}`,
+        projectId,
+        triggerId,
+        actionId,
+      },
+    });
+
+    const baseEvent = {
+      entityType: "prompt-version" as const,
+      projectId,
+      promptId,
+      prompt: {
+        id: promptId,
+        projectId,
+        name: "test-prompt",
+        version: 1,
+        prompt: { messages: [{ role: "user", content: "Hello" }] },
+        config: null,
+        tags: [],
+        labels: [],
+        type: PromptType.Chat,
+        isActive: true,
+        createdBy: "test-user",
+        createdAt: new Date(),
+        updatedAt: new Date(),
+        commitMessage: null,
+      },
+      user: {
+        id: "test-user",
+        name: "Test User",
+        email: "test@example.com",
+      },
+    };
+
+    // Same prompt id, two distinct lifecycle events for that same trigger.
+    // The second event must not be treated as a duplicate of the first.
+    await promptVersionProcessor({ ...baseEvent, action: "created" });
+    await promptVersionProcessor({ ...baseEvent, action: "deleted" });
+
+    const executions = await prisma.automationExecution.findMany({
+      where: { projectId, automationId },
+    });
+
+    expect(executions).toHaveLength(2);
+    expect(webhookQueueAddMock).toHaveBeenCalledTimes(2);
+  });
 });
--- a/worker/src/__tests__/promptVersionProcessor.test.ts
+++ b/worker/src/__tests__/promptVersionProcessor.test.ts
@@ -14,6 +14,7 @@ import { ActionType, prisma } from "@langfuse/shared/src/db";
 import { promptVersionProcessor } from "../features/entityChange/promptVersionProcessor";
 
 const webhookQueueAddMock = vi.fn();
+const webhookQueueGetInstanceMock = vi.fn();
 
 vi.mock("@langfuse/shared/src/server", async (importOriginal) => {
   const actual =
@@ -21,7 +22,7 @@ vi.mock("@langfuse/shared/src/server", async (importOriginal) => {
   return {
     ...actual,
     WebhookQueue: {
-      getInstance: () => ({ add: webhookQueueAddMock }),
+      getInstance: () => webhookQueueGetInstanceMock(),
     },
   };
 });
@@ -34,6 +35,8 @@ describe("promptVersionChangeWorker", () => {
     projectId = result.projectId;
     webhookQueueAddMock.mockReset();
     webhookQueueAddMock.mockResolvedValue(undefined);
+    webhookQueueGetInstanceMock.mockReset();
+    webhookQueueGetInstanceMock.mockReturnValue({ add: webhookQueueAddMock });
   });
 
   it.each([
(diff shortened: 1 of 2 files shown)


chelsealong:
Addressed both P1 review comments:

- `WebhookQueue.getInstance()` returning `null` (e.g. Redis unavailable) is no longer silently skipped by optional chaining — it now throws explicitly, which the existing `catch` marks `ERROR` and rethrows for retry.
- `enqueueAutomationAction` now checks for an existing non-`ERROR` execution (by `projectId`/`triggerId`/`actionId`/`sourceId`) before creating a new one, so a BullMQ retry after a partial failure no longer re-enqueues and duplicate-delivers triggers that already succeeded.

Added two tests covering both cases in `worker/src/__tests__/promptVersionProcessor.test.ts`; full suite (12 tests) passes, plus `pnpm run lint` and `pnpm --filter worker run typecheck` clean.

marksalpeter:
@claude review

chelsealong:
Addressed the idempotency-guard key-mismatch found in the latest review: the guard keyed only on `projectId`/`triggerId`/`actionId`/`sourceId` (prompt id), so a legitimate second event for the same prompt (e.g. `deleted` after `created`) was silently treated as a duplicate and its webhook was dropped. The guard now also matches on the event `action` (stored in `input.action` and filtered via a JSON-path check), so retries still dedupe correctly while distinct lifecycle events for the same prompt id are no longer skipped.

Added a regression test (`does not skip a legitimate execution for a different event action on the same prompt id`) confirming a `created` then `deleted` event on the same prompt id now produce two executions/webhook enqueues instead of one. Full targeted suite (13 tests), `pnpm --filter worker run lint`, and `pnpm --filter worker run typecheck` all pass.

marksalpeter:
@claude review

marksalpeter (approved):

marksalpeter (changes requested):
lets fix this please

chelsealong:
Fixed the crash-recovery gap flagged in the last review: if a worker died between creating the PENDING execution row and enqueueing/erroring it, the idempotency guard treated that row as an in-flight duplicate forever, permanently dropping the delivery — reproducing the exact stuck-PENDING bug this PR fixes. `enqueueAutomationAction` now only treats a PENDING row as in-flight while it's recent (< 5 min, well beyond the job's worst-case retry window); a stale one falls through to a fresh attempt. Added a regression test seeding a backdated PENDING row and asserting a retry still creates and enqueues a new execution. `pnpm --filter worker run test src/__tests__/promptVersionProcessor.test.ts` (14 passed), `lint`, and `typecheck` all pass.

chelsealong:
@marksalpeter — this is ready for another look when you have a moment.

All three rounds of review feedback are in, and the `CHANGES_REQUESTED` from
2026-08-26 09:06 predates the last two of them:

- **Null-queue handling** — `WebhookQueue.getInstance()` returning `null` is no
  longer swallowed by optional chaining; it throws, the existing `catch` marks
  the execution `ERROR` and rethrows for retry.
- **Idempotency key mismatch** (08-25) — the guard keyed only on
  `projectId`/`triggerId`/`actionId`/`sourceId`, so a legitimate second event
  for the same prompt (`deleted` after `created`) was treated as a duplicate
  and its webhook dropped.
- **Crash-recovery gap** (08-30, latest push `a…`) — if a worker died between
  creating the `PENDING` execution row and enqueueing it, the guard treated
  that row as an in-flight duplicate forever and the delivery was permanently
  lost. That is the same stuck-`PENDING` class this PR set out to fix, so it
  seemed wrong to leave it.

The branch is green and the diff is unchanged in shape since your approval —
the additions are the three fixes above. Happy to split any of them out if that
makes review easier, or to rebase if it has drifted.

marksalpeter:
Thanks @chelsealong! @claude review

marksalpeter:
@claude review

chelsealong:
Looked at both new findings from today's review. Leaving both open rather than shipping a fix I can't actually verify closes the gap:

- **ERROR rows not covered by the dedup guard**: real, but the suggested fix (a deterministic BullMQ job id so a duplicate `add()` no-ops) wouldn't close it — `WebhookQueue` sets `removeOnComplete: true` (`packages/shared/src/server/redis/webhookQueue.ts:25`), so a successfully-processed job is gone from BullMQ well before a retry ~5s-75s later could dedupe against it. Closing this for real needs an idempotency check at the delivery layer (dedupe by execution id before sending the HTTP request), which is a bigger change than this PR's blast radius.
- **Consecutive-ERROR auto-disable counting enqueue failures**: also real, but fixing it needs a way to distinguish "enqueue failed" from "delivery failed" (new status/flag + migration), and touches `getConsecutiveAutomationFailures` in `packages/shared/src/server/repositories/automation-repository.ts:329`, outside this bug's scope.

Happy to take either as a follow-up if you'd like them split out — just didn't want to bolt on a partial fix to this PR that looks like it solves the tail case but doesn't.

PR #16507, opened 2026-08-24
fix(ingestion): preserve explicit usage.total of 0 in legacy events

### LINKED ISSUE #16503: bug: Legacy ingestion overwrites explicit usage.total: 0 via || truthiness bug
## What

In the legacy ingestion path, when `usage.total` is explicitly provided as `0`, the `||` operator treats it as falsy and overwrites it with a computed total derived from `input + output`. This silently corrupts the stored usage data.

## Affected file

`worker/src/services/IngestionService/index.ts`, lines 1924-1932:

```typescript
const newTotalCount =
  ("usage" in obs.body ? obs.body.usage?.total : undefined) ||
  (Object.keys(
    "usageDetails" in obs.body ? (obs.body.usageDetails ?? {}) : {},
  ).length === 0
    ? newInputCount && newOutputCount
      ? newInputCount + newOutputCount
      : (newInputCount ?? newOutputCount)
    : undefined);
```

## Reproduction

Send an ingestion event with:

```json
{
  "type": "generation-create",
  "id": "...",
  "traceId": "...",
  "usage": { "input": 100, "output": 50, "total": 0 }
}
```

**Expected:** `provided_usage_details.total` = `0` (the caller explicitly stated 0 tokens)
**Actual:** `provided_usage_details.total` = `150` (computed from `input + output` because `0 || ...` is falsy)

## Why this happens

JavaScript's `||` operator short-circuits on any falsy value, including `0`, `false`, and `""`. Since `usage.total` can legitimately be `0` (e.g., cached responses where no tokens are counted, or when the caller intentionally reports 0), the `||` should be replaced with `??` (nullish coalescing) which only short-circuits on `null` or `undefined`.

## Impact

- **Data corruption:** Stored `provided_usage_details.total` and `usage_details.total` will be wrong for any SDK that sends `total: 0`
- **Dashboard inaccuracy:** "Total Usage" metrics, cost calculations, and export data will be inflated
- **Cascading:** Cost details derived from usage details will also be incorrect since they depend on the stored total

Note: the `newInputCount && newOutputCount` expression on line 1929 has a similar issue -- if `input: 0` is sent without `output`, the expression evaluates to `0` (falsy) and falls through to `newInputCount ?? newOutputCount`. However, this branch only executes when `total` is not provided AND `usageDetails` is empty, so it is a less likely scenario. The primary bug is the `usage.total` overwriting.

## Suggested fix

Replace `||` with `??` on line 1925:

```typescript
const newTotalCount =
  ("usage" in obs.body ? obs.body.usage?.total : undefined) ??
  (Object.keys(
    "usageDetails" in obs.body ? (obs.body.usageDetails ?? {}) : {},
  ).length === 0
    ? newInputCount && newOutputCount
      ? newInputCount + newOutputCount
      : (newInputCount ?? newOutputCount)
    : undefined);
```

This preserves explicit `0` values while still falling through to the computed total when `total` is `undefined` or `null`.

- Whether this is the only place in the ingestion pipeline where `||` is used on numeric usage fields -- check if `newInputCount` or `newOutputCount` have the same truthiness issue elsewhere in the merge step
- Whether the `newInputCount && newOutputCount` fallback (line 1929) should also be fixed -- `0 && ...` evaluates to `0` which is falsy, so it falls through to `??`. This is technically correct for the fallback path but may warrant explicit `!= null` checks for clarity
- Whether any SDKs currently send `total: 0` -- if none do, this is a latent bug waiting to happen rather than an actively exploited one

chelsealong:
## What does this PR do?

Fixes #16503

In the legacy ingestion path, `newTotalCount` was computed with
`("usage" in obs.body ? obs.body.usage?.total : undefined) || (...)`.
Since `||` treats `0` as falsy, an explicitly provided `usage.total: 0`
was silently overwritten with a computed `input + output` sum. This
corrupted `provided_usage_details.total` and `usage_details.total` for
any event that legitimately reports zero total tokens (e.g. cached
responses), which cascades into "Total Usage" dashboard metrics and
cost calculations.

The fix replaces `||` with `??` on that line
(`worker/src/services/IngestionService/index.ts`), so only
`null`/`undefined` fall through to the computed fallback, while an
explicit `0` is preserved.

## Type of change

## Mandatory Tasks

## Test plan

Added a unit test in
`worker/src/services/IngestionService/tests/IngestionService.unit.test.ts`:
"preserves an explicit usage.total of 0 instead of overwriting it with
input + output". It sends a `generation-create` legacy event with
`usage: { input: 100, output: 50, total: 0 }` and asserts the queued
observation record's `provided_usage_details.total` and
`usage_details.total` are `0`.

Confirmed the test fails against the pre-fix code
(`git checkout HEAD~1 -- worker/src/services/IngestionService/index.ts`,
then reran):

```
FAIL  ... IngestionService unit tests > preserves an explicit usage.total of 0 ...
- Expected
+ Received
  {
    "provided_usage_details": { "input": 100, "output": 50, "total": 0 },
-   "total": 0,
+   "total": 150,
  }
 Test Files  1 failed (1)
      Tests  1 failed | 12 passed (13)
```

After restoring the fix, `pnpm --filter worker run test IngestionService.unit.test.ts`:

```
 Test Files  1 passed (1)
      Tests  13 passed (13)
```

Also ran, both clean:

```
pnpm --filter worker run lint      # no errors
pnpm --filter worker run typecheck # no errors
```

`worker/src/services/IngestionService/tests/calculateTokenCost.unit.test.ts`
was not run to completion in this environment because it requires a
live Postgres/Redis connection that isn't available here (unrelated to
this change).

## Checklist

- [x] My code follows the style guidelines of this project (`pnpm run format` / lint clean)
- [x] New and existing unit tests for the touched file pass locally with my changes

---

This fix was drafted with AI assistance (Claude Code) and reviewed by a human before submission.

This PR fixes legacy observation ingestion so an explicitly supplied `usage.total` of zero is preserved instead of replaced by the input/output sum.

- Replaces the truthy fallback with a nullish fallback when calculating total usage.
- Adds a unit test covering persistence of zero in both provided and normalized usage details.
</details>

The PR appears safe to merge with no actionable defects identified.

The fallback change matches the validated usage contract, preserves the intended explicit zero, and is covered by a focused regression test.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(ingestion): preserve explicit usage...."](https://github.com/langfuse/langfuse/commit/028f869f6a75f8d6965e27241b385ae1510c3be9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56367594)</sub>

### DIFF
--- a/worker/src/services/IngestionService/index.ts
+++ b/worker/src/services/IngestionService/index.ts
@@ -1922,7 +1922,7 @@ export class IngestionService {
         "usage" in obs.body ? obs.body.usage?.output : undefined;
 
       const newTotalCount =
-        ("usage" in obs.body ? obs.body.usage?.total : undefined) ||
+        ("usage" in obs.body ? obs.body.usage?.total : undefined) ??
         (Object.keys(
           "usageDetails" in obs.body ? (obs.body.usageDetails ?? {}) : {},
         ).length === 0
--- a/worker/src/services/IngestionService/tests/IngestionService.unit.test.ts
+++ b/worker/src/services/IngestionService/tests/IngestionService.unit.test.ts
@@ -365,6 +365,64 @@ describe("IngestionService unit tests", () => {
     }
   });
 
+  it("preserves an explicit usage.total of 0 instead of overwriting it with input + output", async () => {
+    const addToQueue = vi.fn();
+    const ingestionService = new IngestionService(
+      {} as any,
+      {} as any,
+      { addToQueue } as any,
+      {} as any,
+    );
+    const timestamp = "2026-07-22T00:00:00.000Z";
+    const observationEventList: ObservationEvent[] = [
+      {
+        id: "event-id",
+        timestamp,
+        type: "generation-create",
+        body: {
+          id: "observation-id",
+          traceId: "trace-id",
+          startTime: timestamp,
+          usage: {
+            input: 100,
+            output: 50,
+            total: 0,
+          },
+          environment: "default",
+        },
+      },
+    ];
+
+    vi.spyOn(ingestionService as any, "getClickhouseRecord").mockResolvedValue(
+      null,
+    );
+    vi.spyOn(ingestionService as any, "getPrompt").mockResolvedValue(null);
+    vi.spyOn(ingestionService as any, "getGenerationUsage").mockResolvedValue(
+      {},
+    );
+
+    await (ingestionService as any).processObservationEventList({
+      projectId: "project-id",
+      entityId: "observation-id",
+      createdAtTimestamp: new Date(timestamp),
+      observationEventList,
+      writeToStagingTables: false,
+      attribution: {
+        ingestionApiKey: "api-key",
+        ingestionSdkName: "sdk",
+        ingestionSdkVersion: "1.0.0",
+      },
+    });
+
+    const queuedRecord = addToQueue.mock.calls.find(
+      ([table]) => table === TableName.Observations,
+    )?.[1];
+    expect(queuedRecord).toMatchObject({
+      provided_usage_details: { input: 100, output: 50, total: 0 },
+      usage_details: { input: 100, output: 50, total: 0 },
+    });
+  });
+
   it("correctly sorts events in ascending order by timestamp", async () => {
     const firstTrace = { timestamp: 1, type: "observation-create" };
     const secondTrace = { timestamp: 1, type: "observation-update" };
--- a/worker/src/services/IngestionService/tests/IngestionService.unit.test.ts
+++ b/worker/src/services/IngestionService/tests/IngestionService.unit.test.ts
@@ -406,7 +406,7 @@ describe("IngestionService unit tests", () => {
       entityId: "observation-id",
       createdAtTimestamp: new Date(timestamp),
       observationEventList,
-      writeToStagingTables: false,
+      writeToStagingTables: true,
       attribution: {
         ingestionApiKey: "api-key",
         ingestionSdkName: "sdk",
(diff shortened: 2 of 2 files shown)


aniruddhaadak80:
﻿Independent verification while triaging this area (we filed #16503 for the same root cause): the fallback at worker/src/services/IngestionService/index.ts (~line 1925, mapObservationEventsToRecords) is indeed the overwrite site — with `||`, an explicit `usage.total: 0` falls through to the derived `input + output` value, so a zero-cost generation silently reports a nonzero total.

Two notes reviewers may find useful:

1. The regression lock needs both record copies asserted: `provided_usage_details.total` AND the staged `usage_details.total` (Observations + ObservationsBatchStaging), since they are assembled separately.
2. Adjacent pre-existing quirk, possibly out of scope here: when only `input: 0` is provided with `output > 0` and no explicit total, the inner `newInputCount && newOutputCount ? input + output : (input ?? output)` chain treats `0` as falsy and returns `0` instead of summing — so derived totals can be understated even after the `??` fix.

Happy to share the unit-test case we wrote for the zero-total preservation if useful (fork branch fix/ingestion-explicit-zero-total).

chelsealong:
Thanks for the review. Addressed note 1: extended the regression test to run with `writeToStagingTables: true` and assert `provided_usage_details.total`/`usage_details.total` on both the `Observations` and `ObservationsBatchStaging` queue writes, not just `Observations`. (In the current code both come from the same `finalObservationRecord` object, so this locks that in rather than fixing a live bug.) Left note 2's input/output falsy-zero quirk alone since you flagged it as out of scope for this PR. `pnpm --filter worker run test IngestionService.unit.test.ts` (13 passed), lint and typecheck clean.

PR #16512, opened 2026-08-25
chore(deps): bump @ferrucc-io/emoji-picker to 0.1.2

### LINKED ISSUE #16472: perf(comments): bump @ferrucc-io/emoji-picker to fix a multi-second stall on Windows (Chrome/Edge)
### Summary
 
`web/src/features/comments/ReactionPicker.tsx` imports `@ferrucc-io/emoji-picker`, which until recently ran a synchronous canvas-based emoji support detection at module evaluation time. **On Windows with Chrome or Edge this blocks the main thread for roughly 5 seconds on page load** — even though the picker itself sits behind a `Popover` and is never rendered until the user clicks the trigger.
 
I tracked this down, submitted a fix upstream, and it has been merged and released:
 
- ferrucc-io/emoji-picker#114 (merged)
- Published as **0.1.2** on 2026-08-17

Langfuse pins `"@ferrucc-io/emoji-picker": "^0.0.47"` in `web/package.json`. Because `^0.0.x` is constrained to the `0.0.x` range under semver, this release will not be picked up automatically. **Bumping the pin to `^0.1.2` should be all that is needed.**
 
### Who is affected
 
We tested across our team. The pattern is Chromium-based browsers on Windows:
 
| Environment | Result |
|---|---|
| **Windows + Chrome** | Reproducible multi-second stall, clearly noticeable |
| **Windows + Edge** | Same, comparable to Chrome |
| Windows + Firefox | No perceptible delay |
| macOS (all our members) | No perceptible delay |
 
The same work runs everywhere — the detection performs an identical number of canvas operations regardless of environment. What differs is the per-operation cost, and the spread is roughly two orders of magnitude. Most environments land below the threshold of human perception; Windows with a Chromium-based browser lands well above it.
 
Since Edge is the default browser on Windows and Chrome is the most common alternative, this covers a substantial share of Windows users, which is why I think it is worth picking up rather than leaving for the next routine dependency bump.
 
### What the problem was
 
`filterSupportedEmojis()` ran at module scope. For each emoji it drew the glyph twice on a 2×2 canvas (blue, then red) and read the pixels back to decide whether the platform renders it as a colour emoji.
 
With `unicode-emoji-json@0.8.0` there are 1,906 emoji, of which 1,870 pass the width pre-check — so **3,740 synchronous `getImageData` calls** ran before the importing module finished evaluating. There were three module-scope call sites: two components and a Jotai atom.
 
The dominant cost is not the pixel readback; it is the colour-emoji glyph rasterisation that each `fillText` forces. That is why the impact varies so much by browser and platform.
 
### Measurements
 
Measured with a standalone harness that wraps `CanvasRenderingContext2D.prototype.getImageData` before the library is loaded and then dynamically imports the package — the library in isolation, without Langfuse or Next.js involved.
 
| | `getImageData` calls | Cumulative time |
|---|---|---|
| Before — Windows 11 / Chrome 150 / Segoe UI Emoji | 3,740 | **~5,200 ms** |
| Before — 0.0.47, Linux / software rendering / headless / Noto Color Emoji | 3,740 | ~400 ms (333–519) |
| **After — 0.1.2, same Linux harness** | **0** | **0.0 ms** |
 
The last two rows are a controlled pair: 0.0.47 and 0.1.2 were each installed from npm into identical throwaway Vite projects and measured with the same harness, three runs each. The control reproduced 3,740 calls in all three runs before 0.1.2 produced zero, so the zero is not an artefact of a broken harness.
 
The call count is identical across environments; only the per-call cost differs. Detection still happens in 0.1.2 — it just runs at the picker's first render instead of at import, deduplicated by the library's existing cache. Rendering output is unchanged.
 
### One thing worth flagging, to save anyone the detour
 
This looks like a GPU-accelerated canvas readback problem, and it is not. I spent a while on that hypothesis before ruling it out:
 
- The detection canvas is created with `willReadFrequently: true`, so it is CPU-backed from the start.
- Running Chrome with `--disable-accelerated-2d-canvas` (verified in `chrome://gpu`) changed nothing: 5,191 ms with acceleration vs 5,158 ms without.
- `getImageData` in isolation runs at ~0.004 ms per call.
### Compatibility: 0.0.47 → 0.1.2
 
I diffed the published packages. No breaking changes:
 
| | 0.0.47 | 0.1.2 |
|---|---|---|
| `index.d.ts` exports | `EmojiPicker` | `EmojiPicker` + type exports + `isCustomEmoji` (additive) |
| `EmojiPickerProps` | `children` / `className` / `onEmojiSelect` / `emojisPerRow` / `emojiSize` / `maxUnicodeVersion` | same six, plus `customSections` / `frequentlyUsedEmojis` / `renderHeader`, all optional |
| Compound components | `Header` / `Input` / `Group` / `List` / `Preview` / `Content` / `SkinTone` | same set |
| peerDependencies | react `^18.2.0 \|\| ^19.0.0` | react `^18.0.0 \|\| ^19.0.0` (relaxed) |
| `tailwind-merge` | `^2.6.0` | `^3.4.0` — consistent with the Tailwind v4 migration in #12590 |
 
`ReactionPicker.tsx` only uses `EmojiPicker` with `onEmojiSelect`, `emojiSize`, `emojisPerRow` and the compound children, all of which are unchanged.
 
### Optional: lazy-loading `ReactionPicker`
 
Separately from the version bump, `ReactionPicker` could be loaded lazily via `next/dynamic`. The picker is behind a `Popover` and is not needed until the user opens it, so this would also keep the emoji dataset out of the initial bundle. I ran this against my self-hosted instance while the upstream fix was pending and it worked, but it is an independent optimisation rather than something the fix requires — happy to leave it out if you would rather keep the change minimal.
 
### Environment
 
- Langfuse v3.222.0, self-hosted (Docker Compose on AWS EC2)
- Affected: Chrome 150 and Edge on Windows 11 25H2
- Not reproducible on Windows + Firefox, or on macOS
- Affected file: `web/src/features/comments/ReactionPicker.tsx`

Following CONTRIBUTING.md I am raising this as an issue first — happy to open the PR for the bump if that is useful.

ktachibana-ca:
Fixes #16472

## What

Bumps `@ferrucc-io/emoji-picker` in `web` from `^0.0.47` to `^0.1.2` (one-line manifest change + regenerated `pnpm-lock.yaml`).

## Why

`@ferrucc-io/emoji-picker` ≤ 0.0.x ran a synchronous canvas-based emoji-support probe at import time, which blocked the main thread for multiple seconds on Windows + Chrome/Edge (see #16472 for the full investigation). The fix landed upstream in ferrucc-io/emoji-picker#114 and shipped as 0.1.2. `^0.0.47` does not pick up 0.1.x under semver, so a manifest bump is required.

## Compatibility

- Public API is backwards compatible. `ReactionPicker.tsx` only uses `EmojiPicker` with `onEmojiSelect` / `emojiSize` / `emojisPerRow` and the compound children; none of these changed between 0.0.47 and 0.1.2. No source changes were needed.
- Followed `.agents/skills/pnpm-upgrade-package`: release-age helper reports 0.1.2 (published 2026-08-17) is installable under `minimumReleaseAge: 7200` without any `minimumReleaseAgeExclude` entry.
- `pnpm dedupe` was run once. Besides the emoji-picker bump itself (and its transitive deps `jotai`, `@tanstack/virtual-core`), dedupe collapsed `web`'s existing `@tanstack/react-virtual` (3.13.12 → 3.14.10) and `tailwind-merge` (3.5.0 → 3.6.0) onto the versions 0.1.2 brought in, so each is installed once. Both stay inside the ranges already declared in `web/package.json`.
- `pnpm why -r @ferrucc-io/emoji-picker` → only `0.1.2` (via `web`).

## Verification

- `pnpm --filter web typecheck` — passed
- `pnpm --filter web lint` — passed (`--max-warnings 0`)
- `pnpm --filter web build` — `✓ Compiled successfully`, 83/83 static pages generated

The PR upgrades `@ferrucc-io/emoji-picker` from 0.0.47 to 0.1.2 to avoid the synchronous import-time emoji-support probe.
- Updates the web package manifest and regenerated lockfile.
- Deduplicates compatible `@tanstack/react-virtual`, `@tanstack/virtual-core`, `jotai`, and `tailwind-merge` versions.
- Preserves the existing React, React DOM, Tailwind, and emoji-picker usage contracts.
</details>

The PR appears safe to merge with no concrete correctness, security, installation, or compatibility issue identified.

The manifest and lockfile are consistent, the upgraded package retains the API used by the web application, and the incidental deduplicated versions remain within existing declared ranges.
</details>

<sub>Reviews (1): Last reviewed commit: ["chore(deps): bump @ferrucc-io/emoji-pick..."](https://github.com/langfuse/langfuse/commit/6d5cd72f0427564c4f9ef1f25f1df03ebac760bf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56537248)</sub>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -520,8 +520,8 @@ importers:
         specifier: ^3.2.2
         version: 3.2.2(react@19.2.4)
       '@ferrucc-io/emoji-picker':
-        specifier: ^0.0.47
-        version: 0.0.47(@babel/core@7.29.6(supports-color@8.1.1))(@babel/template@7.28.6)(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(tailwindcss@4.2.2)
+        specifier: ^0.1.2
+        version: 0.1.2(@babel/core@7.29.6(supports-color@8.1.1))(@babel/template@7.28.6)(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(tailwindcss@4.2.2)
       '@hookform/resolvers':
         specifier: ^5.2.2
         version: 5.2.2(react-hook-form@7.72.0(react@19.2.4))
@@ -656,7 +656,7 @@ importers:
         version: 8.20.5(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       '@tanstack/react-virtual':
         specifier: ^3.13.12
-        version: 3.13.12(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+        version: 3.14.10(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       '@trpc/client':
         specifier: ^11.13.4
         version: 11.13.4(@trpc/server@11.13.4(typescript@7.0.2))(typescript@7.0.2)
@@ -851,7 +851,7 @@ importers:
         version: 2.2.2
       tailwind-merge:
         specifier: ^3.5.0
-        version: 3.5.0
+        version: 3.6.0
       tailwindcss-animate:
         specifier: ^1.0.7
         version: 1.0.7(tailwindcss@4.2.2)
@@ -2146,12 +2146,12 @@ packages:
     resolution: {integrity: sha512-+CNAzxglkrpNf/kKywqQfk74QjtceuOE7Qm+AF8miRvPF/wmmK5+OJOgVh3AVTT3RP2mH3+FOaxlE5v72owk0A==}
     engines: {node: ^20.19.0 || ^22.13.0 || >=24}
 
-  '@ferrucc-io/emoji-picker@0.0.47':
-    resolution: {integrity: sha512-w7lRzT3bvf7NdNl+cfJT3+KFCrdWXy8H0LPtDamFxHoOb6SBbWvRnMBX0Ot24NIvNtfFkXXeo5z9/SMo2e9DzQ==}
+  '@ferrucc-io/emoji-picker@0.1.2':
+    resolution: {integrity: sha512-P15x8qOOz3n12lkCWrnKbniy2otHTg0JLyK2LNsQzrj2UJwltSDcitqj10+FRf4dMEFuGKHPjiwqP/8Xz2e+YA==}
     engines: {node: '>=18'}
     peerDependencies:
-      react: ^18.2.0 || ^19.0.0
-      react-dom: ^18.2.0 || ^19.0.0
+      react: ^18.0.0 || ^19.0.0
+      react-dom: ^18.0.0 || ^19.0.0
       tailwindcss: '>=3.0.0'
 
   '@floating-ui/core@1.7.3':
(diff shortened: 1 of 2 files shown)


ktachibana-ca:
Reviewer notes — things worth doubting:

- `pnpm dedupe` moved two of `web`'s own direct deps (`@tanstack/react-virtual` 3.13.12→3.14.10, `tailwind-merge` 3.5.0→3.6.0) onto the versions emoji-picker 0.1.2 pulls in. Both are inside the declared ranges and it avoids shipping two copies, but if you'd rather keep this PR strictly to emoji-picker I can drop the dedupe step.
- I only ran `web` typecheck/lint/build locally, not the full monorepo build or the vitest suites — the change is manifest + lockfile only, but CI is the real check here.
- Runtime fix (no main-thread stall) is only verifiable on Windows Chrome/Edge; I have not re-measured it on the preview.

PR #16543, opened 2026-08-25
fix(events): per-trace tool-call count for events list 'Tool Calls' column

### LINKED ISSUE #16520: bug: trace list "Tool Calls" column is always 0
### Describe the bug

> Env: self-hosted Langfuse (v4.1.0), ClickHouse 26.7.1.1315, Langfuse `schema_migrations` latest version 46. Model: `deepseek-v4-flash` (via OpenRouter, OTLP ingestion)
> Conclusion: this is a Langfuse UI/API aggregation issue, **not an instrumentation-side problem**. Tool-call data is complete and accurately recorded on the GENERATION spans in `events_full`, in conformance with the OpenTelemetry GenAI semantic conventions. The "Tool Calls" column on the trace list reads the wrong level of the span tree, so it always shows 0.

## 1. Symptom 

On the Langfuse trace list page, filtered by `type=chain`:

- Every trace shows **"Tool Calls" = 0** in the list.
- But expanding any trace id reveals **multiple rounds of tool calls (gen_ai LLM turns + tool executions)**, all present and correct.

That is: the list-level aggregate=0, while the detail-level expansion = many tool calls. The two disagree.

## 2. Root cause

`tool_calls` / `tool_call_names` are **GENERATION-level** OTel GenAI attributes ("which tool calls this LLM call declared"). CHAIN is the **top-level aggregate span** of the whole agentic loop, and structurally never carries `tool_calls` - declarations live on its child GENERATION spans, executions on sibling TOOL spans. This is spec-compliant, not missing data.

The events-list "Tool Calls" column is defined as `length(e.tool_calls)` **per event row, with no trace-level aggregation** (`eventsTable.ts:462-465`; `eventsService.ts` `getEventList` paginates per-event, no `groupBy`). With `type=chain`, every row is a CHAIN event whose `tool_calls` is empty -> the column shows 0 on every row. The correct aggregation would be the sum of `tool_calls` across the trace's GENERATION spans (equivalenty, the count of its `type=TOOL` spans).

Confirmed end-to-end against the Langfuse main-app source:

- **Type classification** - `ObservationTypeMapper.ts` maps `openinference.span.kind=CHAIN`->CHAIN, `gen_ai.operation.name=chat` / `LLM`->GENERATION, `execute_tool` / `TOOL` -> TOOL.
- **Write is type-independent** - `OtelIngestionProcessor.ts` writes `toolCalls`/`toolCallNames` to any event type whenever non-empty (not gated by type).
- **Source is LLM raw output (GENERATION only)** - `extractToolsBackend.ts` `extractToolCallsFromRawOutput` parses `choices[].message.tool_calls` / `additional_kwargs` from the **span's own** `output`. A CHAIN span's output is the agent-loop result, not an LLM response body ->nothing to parse ->empty. This is why only GENERATION carries `tool_calls`. 

## 3. Expected behavior

The "Tool Calls" column on the events list page should aggregate **the sum of `tool_calls` across all GENERATION spans of the trace** (equivalently, the count of `type='TOOL'` spans), rather than the per-event `tool_calls` field of the row (empty for a CHAIN row). Then the column matches what the expanded trace shows.

### UI
1. Open Langfuse, go to the **Events** page (not the lagacy Traces page - `tracesTable.ts` has no "Tool Calls" column at all).
2. Add the filter `type = chain`.
3. Observe the **"Tool Calls" column shows 0 on every row**.
4. Click any trace id to expand it. Observe multiple `GENERATION` spans carrying `tool_calls` and `TOOL` spans executing them - the data is clearly there.

Self-hosted

v4.1.0

_No response_

## Re-checked against Langfuse 4.17.0 - **still present, not fixed**

`normalizeToolsForObservation(input, output, metadata)` is still called with the **span's own**  `output` (not a trace aggregate), so CHAIN stays empty.

Yes

VieDica:
## What does this PR do?

Fixes #16520

The Events list page "Tool Calls" column shows `0` on every row when filtered to `type=chain`. Each CHAIN is a trace's top-level span and structurally never carries `tool_calls` (a GENERATION-level GenAI attribute). The column reads the per-event `tool_calls` field of the row, which is empty for CHAIN -> `0`, even though expanding the trace shows many tool calls. Data is complete and spec-correct; only the column's aggregation scope is wrong.

This PR computes a **per-trace tool-call count** at query time instead of reading the (empty for CHAIN) per-event array.

### What changed

1. `event-query-builder.ts`
    - Add trace-level `tool_calls_count: "countIf(type = 'TOOL') AS tool_calls_count"` to `EVENTS_AGGREGATION_FIELDS` (valid under `GROUP BY trace_id`, auto-joins `"all"`).
    - Add per-row scalar `toolCallsCount` to `EVENTS_FIELDS` (correlated subquery counting TOOL spans in the row's trace) and to `FIELD_SETS.base` / `baseWithoutTools`.

2. `event.ts`
    - Extend `EventsObservationQueryResult` with `tool_calls_count?: number`.
    - In `enrichObservationsWithTraceFields`, prefer the DB scalar over `observation.toolCalls.length`, falling back to the array length when absent.

3. `event-query-builder.test.ts` - assert `selectFieldSet("all")` SQL contain the new expressions.

The original `toolCalls` / `toolCallNames` / `toolDefinitions` arrays pass through unchanged.

### Why `countIf(type = 'TOOL')`

The `tool_calls` array lives on GENERATION spans (LLM declarations); TOOL/CHAIN/SPAN carry an empty array. Counting TOOL spans measures actual executions and reads no array. Verified on a real trace: sum of GENERATION `tool_calls` declarations equals the TOOL-span count.

### Why `eventsTable.ts` `toolCalls.internal` is left unchanged

`column.internal` is consumed only by `getEventsNumericStatsByFilterColumn`, which wraps it as `min/max/avg/count(...)` over `GROUP BY tuple()`. Substituting an aggregate there yields "aggregate inside aggregate" - a ClickHouse error that breaks the events-page numeric filter (live caller `getEventFilterNumericRange` -> `getEventsNumericStatsByFilterColumn`). Filtering/ordering use `mapEventsTable.ts` `clickhouseSelect`, not `internal`, so changing `internal` would only add a regression without fixing the symptom. The per-row scalar fixes the actual display path without touching the numeric-stats path.

## Type of change

## Mandatory Tasks

## Checklist

- I have checked if new and existing unit test pass locally with my changes (`pnpm run type check`, `pnpm test event-query-builder`)

This PR adds ClickHouse expressions for counting TOOL events per trace and exposes the result during event-record enrichment.
- Adds scalar and aggregation-level tool-call counts to the event query builder.
- Prefers the database count when enriching event observations.
- Adds an SQL-generation assertion for the aggregation expression.
</details>

<details><summary><h3>Confidence Score: 2/5</h3></summary>

This PR is not safe to merge because the affected Events column still ignores the new count, the count can include obsolete event versions, and the new test fails on an alias mismatch.

The server computes a property that the Events table does not render, scans raw replacing-table rows without current-record semantics, and emits a singular aggregation alias where tests and consumers require the plural name.

**Files Needing Attention:** packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts, packages/shared/src/server/repositories/events.ts
</details>

```mermaid
flowchart LR
  EC[(events_core)] --> Q[Per-trace TOOL count]
  Q --> E[Enriched toolCallsCount]
  E -. currently ignored .-> T[Events table Tool Calls column]
  A[Per-event toolCalls array] --> T
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/repositories/events.ts:354-359
**Trace count bypasses table mapping**

When a CHAIN row has TOOL spans elsewhere in its trace, enrichment stores the database result in `toolCallsCount`, but the Events table still builds the displayed value from `observation.toolCalls.length`, causing the “Tool Calls” column to remain zero.

### Issue 2
packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts:167-168
**Count includes obsolete event versions**

When a TOOL span has been updated or deleted while multiple `ReplacingMergeTree` rows remain present, this raw correlated subquery counts every physical row without version deduplication or deletion filtering, causing inflated per-trace tool-call counts.

### Issue 3
packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts:573
**Aggregation alias is singular**

The field is declared and consumed as `tool_calls_count`, but this expression emits `tool_call_count`, causing the new query-builder assertion to fail and preventing aggregation consumers from reading the result under its declared name.

```suggestion
  tool_calls_count: "countIf(type = 'TOOL') AS tool_calls_count",
```

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(events): per-trace tool-call count f..."](https://github.com/langfuse/langfuse/commit/dbb4a3812c3d0846ceb9715cfffe762ab6befeb8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56636688)</sub>

**Context used:**

- Knowledge Base — [Observability data platform](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/observability-data-platform.md)

### CODE REVIEW

File: packages/shared/src/server/queries/clickhouse-sql/event-query-builder.ts [hunk 1]
```diff
@@ -164,6 +164,8 @@ const EVENTS_FIELDS = {
   toolDefinitions: 'e.tool_definitions as "tool_definitions"',
   toolCalls: 'e.tool_calls as "tool_calls"',
   toolCallNames: 'e.tool_call_names as "tool_call_names"',
+  toolCallsCount:
+    "(SELECT countIf(type = 'TOOL') FROM events_core e2 WHERE e2.trace_id = e.trace_id AND e2.project_id = e.project_id) as \"tool_calls_count\"",
```
Steffen911, line 168: This will be expensive to execute at scale. We're looking into other ways on how to rollup metrics on a trace/session/user level from within the maintainer group. Until this is in, unfortunately we cannot accept such hot-fixes.

PR #16546, opened 2026-08-25
feat(web): allow serving build output from a dedicated asset host

Steffen911:
🟡 **Live preview: [pr-16546.preview.langfuse.com](https://pr-16546.preview.langfuse.com)**

**TL;DR** — Adds `NEXT_PUBLIC_ASSET_PREFIX`, so a build can serve its `/_next/static/*` output from a dedicated hostname (a CDN) instead of only the app origin. Unset — which is every environment and every OSS image until someone sets it — nothing changes.

A browser tab that outlives a deploy keeps requesting the chunk filenames of the build it was served. A container only ever holds the current build, so those requests 404 and the tab dies with a stale-module error. Holding the assets at a CDN edge lets them keep resolving. Companion infrastructure PR: langfuse/infrastructure#1283 (private).

### Why a build argument and not a runtime variable

Two independent reasons, both verified:

- The bundler writes the prefix into the client runtime's chunk base path at build time.
- The container runs `node web/server.js`, which reads the config frozen into `.next/required-server-files.json` — `next.config.mjs` is never re-evaluated at runtime.

`_deploy_ecs_service.yml` already builds one image per environment from environment-scoped `vars`, which is how `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` works, so this follows that path and is set per environment there.

### What changes when it is set

- **CSP** is widened to the asset origin for `default-src`, `script-src`, `style-src`, `font-src` and `connect-src`. `style-src` is the load-bearing one — CSS chunks turn cross-origin and a blocked stylesheet renders the app unstyled. `img-src` and `media-src` already allow `https:`, and `worker-src` is deliberately untouched (see below). A host under `langfuse.com` would already match the existing wildcards, but the policy should not silently depend on where the asset host lives.
- **`crossOrigin: "anonymous"`**, so an exception thrown by a cross-origin script keeps its stack instead of arriving at `window.onerror` as a bare `"Script error"` — which would blind Sentry to exactly the failures this setup exists to observe. This makes `Access-Control-Allow-Origin` on the asset host load-bearing: scripts fail closed rather than quietly losing their stacks.
- **The three bundled web workers are skipped**, because they cannot work. This is the one real regression and it is worth reading the next section.

### The web workers cannot follow the assets (verified against build output)

Turbopack's worker factory compiles to:

```js
let c = new URL(e.h(bootstrapChunk, e.b), location.origin);
return new n(c, i ? { ...i, type: void 0 } : void 0);
```

`e.b` is `TURBOPACK_CHUNK_BASE_PATH`, which the build bakes to the asset prefix, so `c` is absolute and cross-origin — and `type: void 0` means Turbopack **discards** `{ type: "module" }` and always creates a *classic* worker. Browsers reject a cross-origin classic worker script outright with a `SecurityError`, whatever CSP allows.

There is no app-level fix. The base path is baked into the Turbopack runtime chunk, Next exposes no per-worker override, and webpack's `output.workerPublicPath` is unreachable because `next build` on Next 16 takes the Turbopack path (this config defines both a `turbopack` and a `webpack` key, so the webpack hook is ignored).

So `canUseBundledWorker()` checks the asset origin up front. All three call sites already fall back to the main thread on failure; the guard makes that a deliberate, single-place-to-delete state instead of a `SecurityError` on every attempt. Consequence on staging once the prefix is on: large trace payloads parse on the main thread, and the trace-graph ELK layout is unavailable. That is a genuine cost and an input into the prod decision, not something to paper over.

### Verification

- `pnpm run typecheck --filter=web` — `Tasks: 4 successful, 4 total`
- `pnpm run lint --filter=web` — `Tasks: 4 successful, 4 total` (plus the pre-commit hook's full-repo `turbo run lint`: `Tasks: 7 successful, 7 total`)
- `pnpm --filter web run test-client web-workers` — `Tests 4 passed (4)`
- `pnpm run build:check` with the prefix **set** — succeeds, and the emitted HTML carries the prefix on both scripts and stylesheets:
  - `<script src="https://static-staging.langfuse.com/_next/static/chunks/169ey_uhvsdfu.js" async="" crossorigin="">`
  - `<link rel="stylesheet" href="https://static-staging.langfuse.com/_next/static/chunks/1d7f5jysnbjgy.css" crossorigin="anonymous" data-n-g=""/>`
  - guard in the client bundle: `function d(){if(!window.Worker)return!1;let e=u.env.NEXT_PUBLIC_ASSET_PREFIX;return!e||new URL(e).origin===window.location.origin}`
- `node scripts/scan-client-bundle.mjs web/.next-check/static` — `clean: every referenced identifier is declared or a known global`
- The rendered CSP with the variable unset is **identical** to `main`, including whitespace; with it set, `assetPrefix` and `crossOrigin` flip and the five directives gain exactly one source each.

### Not testable on the PR preview

`pr-<N>.preview.langfuse.com` builds with the variable empty, so the preview exercises the unset path only — which is the path that must stay unchanged, so it is still worth a look. The prefixed path can only be exercised once the CDN hostname exists; the staging test plan lives on the ticket.

The PR adds build-time support for serving Next.js static assets from a dedicated host while preserving the existing behavior when no host is configured.
- Passes the asset prefix through the ECS workflow and Docker build.
- Configures Next.js asset URLs, cross-origin loading, CSP sources, and same-origin worker assets.
- Adds environment validation and self-hosting documentation.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

```mermaid
flowchart LR
  Config[NEXT_PUBLIC_ASSET_PREFIX] --> Build[Docker and Next.js build]
  Build --> HTML[Emitted page asset URLs]
  HTML --> CDN[Dedicated asset host]
  Build --> CSP[CSP allows asset origin]
  Build --> Workers[Worker assets remain on app origin]
  CDN --> Browser[Browser loads JS, CSS, and fonts]
  Workers --> Browser
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(web): keep bundled workers same-orig..."](https://github.com/langfuse/langfuse/commit/2973de6e282010761a307051fe5fb479dd4238b2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56640675)</sub>


Steffen911:
Where I would push back on this:

- **The worker trade-off is the whole decision, not a detail.** Losing main-thread-offloaded JSON parsing and the ELK graph layout is a real cost. If it is unacceptable, the honest alternative is to drop `assetPrefix` entirely and instead put CloudFront in front of the *app domain* with a `/_next/static/*` behavior — no cross-origin, no CORS, no CSP change, workers keep working — at the price of a CDN in the path of all app traffic, which is what we moved away from with Cloudflare. I picked the dedicated hostname because that was the plan on the ticket; the worker finding is new information that arguably reopens it.
- **`connect-src` is defensive, not demonstrated.** Nothing in the app `fetch()`es the asset origin today — chunks arrive via `<script>`/`<link>`, which `script-src`/`style-src` already cover. I added it to avoid a silent blocked-prefetch class of bug. Reasonable to cut.
- **I did not touch `turbo.json`.** Turborepo's framework inference already hashes `NEXT_PUBLIC_*` for Next packages — observed: the unset build rebuilt, then the prefixed build cache-hit its earlier twin. That is inferred behavior rather than a declared contract, and `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` etc. are equally undeclared today, so I left it alone. Say the word if you want it explicit in the `build` task's `env`.
- **`crossOrigin` is gated on the prefix** so prod stays byte-identical. Arguably it should just be on always — it is good hygiene for same-origin scripts too — at the cost of one more difference between the two paths.
- **The `.refine()` rejecting a path in the prefix is mine, not Next's.** Next accepts `https://cdn.example.com/foo`; I reject it because the CSP source is computed from the origin and nothing verifies a path is mirrored by whatever fronts the host. If someone needs a path prefix, that guard has to go with eyes open.
- **`next/image` grows a remote pattern for free.** Next auto-appends the asset prefix host to `images.remotePatterns` because statically imported images get prefixed, while `images.path` stays on the app origin — so the optimizer starts making server-side HTTPS calls out to the CDN through the NAT. I deliberately did nothing about it; flag if that is unwelcome.

PR #16558, opened 2026-08-25
chore(in-app-agent)!: LANGFUSE_AI_* replaces LANGFUSE_AWS_BEDROCK_* for the Langfuse AI model

LysanderKie:
🟡 **Live preview: [pr-16558.preview.langfuse.com](https://pr-16558.preview.langfuse.com)**

## Summary

`LANGFUSE_AI_MODEL` and `LANGFUSE_AI_SMALL_MODEL` were Anthropic-only, so Bedrock deployments had to use `LANGFUSE_AWS_BEDROCK_MODEL` / `_SMALL_MODEL` for the same concept. Self-hosting docs for the Assistant are about to be published and should not introduce two names for one thing.

`LANGFUSE_AI_MODEL`, `LANGFUSE_AI_SMALL_MODEL` and `LANGFUSE_AI_AWS_BEDROCK_REGION` now apply to both providers. **The three `LANGFUSE_AWS_BEDROCK_*` variables are removed, not deprecated.**

## Breaking change

| Removed | Use instead |
| --- | --- |
| `LANGFUSE_AWS_BEDROCK_MODEL` | `LANGFUSE_AI_MODEL` |
| `LANGFUSE_AWS_BEDROCK_SMALL_MODEL` | `LANGFUSE_AI_SMALL_MODEL` |
| `LANGFUSE_AWS_BEDROCK_REGION` | `LANGFUSE_AI_AWS_BEDROCK_REGION` |

This needs a release note. Rationale for removing rather than deprecating:

- These variables did nothing on self-hosted until the Assistant and Ask AI shipped to OSS, so the exposure window is short.
- They have never appeared in a docs page or in `.env.prod.example`. The only way a self-hoster could have found them is by reading `docker-compose.yml` or the source.
- The setup docs in langfuse-docs#3619 name only `LANGFUSE_AI_*`. Removing before those publish means we never ship a documented name we intend to retire.

## Cloud is unaffected

`terraform/environments/services/locals.tf` and `worker.tf` set `LANGFUSE_AI_MODEL = var.langfuse_aws_bedrock_model`, `LANGFUSE_AI_SMALL_MODEL = var.langfuse_aws_bedrock_small_model` and `LANGFUSE_AI_AWS_BEDROCK_REGION = var.region`, alongside every legacy name, from the same terraform variables. That covers web, web-iso and web-ingestion via `shared_env_web`, the worker via its own block, and previews via `k8s/preview/chart/values.yaml`. `worker-cpu` gets neither set and runs no agents.

The env schemas are not strict, so infrastructure can keep setting the old names until its own cleanup lands. No coordinated deploy needed.

## Also in scope

- `web/src/features/ai-features/server/availability.ts` resolved the model by hand, a second and Bedrock-only source of truth for provider resolution. It now calls `getInAppAgentModelConfig()`.
- `.env.prod.example` had no AI, Assistant, Bedrock, sandbox or queue-consumer variables at all, so a self-hoster reading the production example had no path to enabling the Assistant. Added.
- The precondition message a self-hoster sees now names the surviving variables.

## Packages

- `@langfuse/shared` (provider resolution, env schema, Bedrock region helper)
- `web` (env schema, AI features availability)

## Test plan

- [x] `pnpm --filter @langfuse/shared run test modelProvider` — `Tests 6 passed (6)`
- [x] `pnpm --filter web run test in-app-agent` — `Tests 96 passed (96)`
- [x] `pnpm --filter web run test evaluator-v2-service` — `Tests 22 passed (22)`
- [x] `pnpm --filter web run test ai-filter-access` — `Tests 3 passed (3)`
- [x] `pnpm --filter worker run test in-app-agent` — `Tests 129 passed (129)`
- [x] `pnpm run lint` — `Tasks: 7 successful, 7 total`
- [x] `pnpm tc` — `Tasks: 8 successful, 8 total`
- [ ] Preview: send an Assistant message with `LANGFUSE_AI_MODEL` set, and confirm the precondition error names the new variables when it is unset.

## Follow-up

An infrastructure PR drops the now-unread `LANGFUSE_AWS_BEDROCK_*` assignments from terraform and the preview values.

This breaking configuration cleanup consolidates the Langfuse-operated AI model settings under `LANGFUSE_AI_*` for both Bedrock and Anthropic.
- Removes the legacy Bedrock-specific variables from environment schemas and Compose forwarding.
- Centralizes web AI-feature availability on the shared provider resolver.
- Documents the Assistant, worker consumer, model, concurrency, and sandbox settings in the production environment example.
- Updates provider and integration tests to use the consolidated variables.
</details>

The PR appears safe to merge because no blocking failure remains within the eligible follow-up review scope.

No blocking failure remains.
</details>

```mermaid
flowchart LR
  E[LANGFUSE_AI environment variables] --> R[getInAppAgentModelConfig]
  R --> P{Configured provider}
  P -->|Bedrock| B[Bedrock model and optional region]
  P -->|Anthropic| A[Anthropic model and API key]
  B --> W[Web availability checks]
  A --> W
  B --> X[Worker agent execution]
  A --> X
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(in-app-agent): treat an invalid Bedr..."](https://github.com/langfuse/langfuse/commit/23277cf2c0427579937cc010d50eea1de26e5d4d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56702807)</sub>


LysanderKie:
What to doubt here:

1. **The fallback ordering.** Bedrock now does `params.modelId ?? LANGFUSE_AI_MODEL ?? LANGFUSE_AWS_BEDROCK_MODEL`. If any deployment sets `LANGFUSE_AI_MODEL` to an Anthropic-shaped ID while running Bedrock, it silently starts winning over the Bedrock ID that used to be used. Cloud sets both to the same value so it is safe there, but it is the one behavior change a self-hoster could hit.

2. **`availability.ts` now reads the shared env, not the web env.** That is the point, since it was a duplicate resolver, but it moved which env object drives evaluator name and description suggestions. Both schemas read the same `process.env`, so production is unaffected. Three servertests needed retargeting, which is the visible cost.

3. **`titleModelId` no longer mixes families.** It was `LANGFUSE_AWS_BEDROCK_SMALL_MODEL ?? modelId`; it is now `LANGFUSE_AI_SMALL_MODEL ?? LANGFUSE_AWS_BEDROCK_SMALL_MODEL ?? modelId`. Worth confirming that is the precedence you want rather than pairing the small model with whichever family supplied the main model.

4. **I pinned `LANGFUSE_AI_PROVIDER` in the background-run servertest.** It was inheriting the provider from the local `.env`, so on a machine with `LANGFUSE_AI_PROVIDER=anthropic` its Bedrock-named cases were exercising the Anthropic branch. That was pre-existing, and the pin is arguably a separate fix from this PR.

5. **`.env.prod.example` grew a large commented block.** I judged the omission a real gap for self-hosters, but it is the least related change in here and easy to split out.

Not verified: the preview run with each variable set in isolation. Everything else in the test plan has a quoted result.

LysanderKie:
Updated: this now **removes** the `LANGFUSE_AWS_BEDROCK_*` variables rather than keeping them as fallbacks, per review direction, and the docs PR no longer mentions them either.

What to doubt, revised:

1. **The claim that Cloud is safe rests on terraform code, not on applied state.** I verified that every place setting a legacy name also sets the new one from the same variable, in `locals.tf`, `worker.tf` and `k8s/preview/chart/values.yaml`. I cannot verify from here that those task definitions are actually deployed in all five environments. If any environment is running an older task definition, the Assistant there loses its model on the next app deploy. Worth confirming before merge.

2. **`worker-cpu` gets neither variable set.** That is intentional since it runs no agents, but it means the ingestion worker would report the model as unconfigured if anything ever asked it. Nothing does today.

3. **Removing rather than deprecating is a judgment call about exposure.** My reasoning is that these names were never documented for self-hosting and only became functional there about a week ago. If you think anyone is running a self-hosted Assistant configured from `docker-compose.yml`, this wants a deprecation cycle instead and I should put the fallbacks back.

4. **Test churn is mostly mechanical renaming**, but two spots were not: the `evaluator-v2-service` helper and the `background-run` scaffolding both had paired old and new assignments that collapsed into duplicate keys. I rewrote both. Worth a look that they still assert what their names claim.

5. **`.env.prod.example` growing a large commented block** is the least related change here and is easy to split out.

LysanderKie:
@claude review

LysanderKie:
@greptileai review

PR #16561, opened 2026-08-25
feat(trace): ready the compact timeline to be the only one

nkabardin:
🟡 **Live preview: [pr-16561.preview.langfuse.com](https://pr-16561.preview.langfuse.com)**

## TL;DR

Makes the compact renderer ready to be the *only* Timeline (LFE-15489), and it is deliberately less than it started as: four features from the wide timeline were ported, loaded onto one trace, looked at, and **three were taken back out**. What survives is a first-token mark, a duration and a cost that cannot clip and no longer move when you zoom — plus two measurement bugs fixed and the pricing engine relocated out of the file GA will delete.

GA and the deletion follow as separate PRs, so a rollback stays small.

## Why it shrank

The four were score badges, comment indicators, heat-map colouring and the streaming first-token mark. A new seeder scenario (`timeline-annotated`) puts every one of them on one trace, next to plain rows for contrast. Reading the rendered clusters in the browser settled it:

- **Heat-map colouring put a second meaning on colour.** A label inside a bar already takes white or black from that bar's own luminance; the heat map overrode it, so a 19.8s root read as dark red on mid blue. `GUARDRAIL` is red in the type palette too, so red meant a type *and* a severity. This is the rule that removed hover's hue swap in #16352, broken from the other side.
- **Scores could not be carried consistently.** Eleven of thirteen rows showed duration and cost only. A row with two scores was refused them **by nine pixels while 250px of lane sat empty**; a row with four was confined inside its own bar with ~200px free to its right. Identical rows showed different things depending on where their bar happened to sit — the same inconsistency the tooltip had when it changed shape row to row.
- **Comments** are the borderline case: they fitted on two of the three rows that had them, and their only real fault was a count badge reading as a number next to the duration (`12 900ms`). Out for now by decision, to be revisited with somewhere better to put them.

Annotations belong somewhere that does not move — the tree, the detail panel, or a fixed column of their own — not in the space a bar is competing for.

## What is in it

- **The pricing engine moved** out of `TimelineBar.tsx` into `fns/timeline/metricCluster.ts`, unchanged. The wide timeline's own eighteen stories still pass against it, which is what makes the move trustworthy — and it has to leave that file before GA can delete it.
- **`TimelineRowMetrics`** places the duration and the cost, priced with the measurer so the lane can never overflow, and **picks its side by which one has room**.
- **The first-token mark**: `layout()` has always emitted `firstTokenX`; this renderer never drew it. A divider rather than the wide timeline's second shade, because a shade needs a bar tall enough to read one and these are one pixel at the floor.
- `showDuration` and `showCostTokens` honoured. `showScores` / `showComments` now mean what their storage keys always said: the observation tree.
- **`timeline-annotated`**, the scenario that turned this from an argument into a measurement.

## The zoom bug this fixes

The cluster used to inherit the side `layout()` chose by measuring **the duration alone**. So zooming grew a bar until the duration fitted inside it, at which point the budget stopped being "the rest of the lane" and became "this bar" — content that fitted a moment ago was dropped, and reappeared only once the bar itself was wide enough. More room for the bar bought less room for the label. Inside is now the last resort, and the control fails with `expected 'inside' to be 'after'`.

<details>
<summary>Two measurement bugs, and what the tests had to be taught</summary>

### The probe measured a size nothing rendered

The measurer is seeded from an invisible probe span — the trick that stops it pricing text against a font the DOM is not using. The probe was pinned at `10px` while the metrics render at `density.labelFontPx`, so every string was priced ~10% short. Both read the same number now.

### A canvas can refuse a font and say nothing

Setting `context.font` to a string the canvas rejects leaves it at its own default — `10px sans-serif` on a fresh context — and every measurement afterwards is short by the size ratio, silently, in the direction that clips. `createTextMeasurerFrom` reads the size back and scales for the difference, with a test whose fake context refuses every font it is given.

That is also why the new component's story renders in `sans-serif` and measures in `sans-serif`: a canvas resolves `ui-sans-serif` to a face ~19% narrower than the DOM does, so a story seeded with the app's stack prices text nobody renders.

### With nothing fitting, the tooltip is the fallback

Review asked what discloses the metrics when none fit. Nothing is drawn, and deliberately no title — a title on a box of zero width is unhoverable. The tooltip states the same duration and cost for every row at every density; 600 rows at 1px draw no metrics and hovering one still names both. Silencing the tooltip's cost fails three stories.

### Checks

- `Tests 731 passed (731)` — Storybook tests in real Chromium, including the wide timeline's eighteen against the moved pricing engine.
- `Tasks: 7 successful, 7 total` turbo lint. Typecheck clean.

</details>


nkabardin:
@claude review

nkabardin:
@claude review

PR #16562, opened 2026-08-25
fix(evals): make trace eval job dedup atomic

### LINKED ISSUE #16492: bug(evals): non-atomic eval job dedup creates duplicate executions/scores and can strand PENDING executions forever
### Describe the bug

`createEvalJobs` in `worker/src/features/evaluation/evalService.ts` deduplicates eval executions with a **non-atomic** find-then-create, and creates the `job_executions` row **before** enqueueing the execution job. Two failure modes result:

**1. Duplicate eval executions → duplicate visible scores under concurrency.**
Three independent producers call this concurrently for the same trace: trace-upsert shard workers, dataset-run-item-upsert workers, and `CreateEvalQueue` workers. Two workers can both pass the existence check before either inserts:

```ts
// worker/src/features/evaluation/evalService.ts (~L457–472)
const allExistingJobs = await prisma.jobExecution.findMany({
  where: { projectId: event.projectId, jobInputTraceId: event.traceId,
           jobConfigurationId: { in: configIds } },
});
...
// (~L758–778)
if (existingJob.length > 0) { ...continue; }        // ← non-atomic check
await prisma.jobExecution.create({ data: { id: jobExecutionId, ..., status: "PENDING" } });
```

There is no DB unique constraint to stop the double insert — `job_executions` only has `@@index([projectId, jobConfigurationId, jobInputTraceId])` (`packages/shared/prisma/schema.prisma`, model `JobExecution`). Both rows then execute LLM-as-judge evaluations and produce **two distinct score rows**, because evaluation score ids are derived from the random `jobExecutionId` (`packages/shared/src/server/evals/evalScoreIds.ts` — `uuidv5(jobExecutionId, …)`). Users see duplicated scores; LLM spend doubles.

**2. Crash between create and enqueue strands the execution as PENDING forever.**

```ts
await prisma.jobExecution.create({ data: { ..., status: "PENDING" } });   // ← first
const shardingKey = `${event.projectId}-${jobExecutionId}`;
await EvalExecutionQueue.getInstance({ shardingKey })?.add(...);          // ← second
```

If the worker dies or Redis errors between these two calls, BullMQ redelivers the original job — but the retry now finds the existing row via `findMatchingJob` and hits `continue`, never re-enqueuing. There is no reconciler for stale `PENDING` `job_executions`, so that evaluation is never executed and the row sits PENDING indefinitely.

### Steps to reproduce

*Duplicate scores:*
1. Configure an active LLM-as-judge evaluation rule.
2. Ingest a burst of events for the same trace such that two producer queues deliver near-simultaneously (e.g. a trace update and its dataset-run-item creation landing on different shard workers within the dedup window).
3. Observe two `job_executions` rows with the same `(projectId, jobConfigurationId, jobInputTraceId)` but different ids, both reaching COMPLETED, and **two score rows** for one trace + rule.

*Stuck PENDING:*
1. Insert a fault (SIGKILL the worker / drop Redis) between the `jobExecution.create` and `EvalExecutionQueue.add` calls.
2. After restart, the entity-change job retries, finds the existing PENDING row, and skips re-enqueueing.
3. The `job_executions` row stays PENDING forever; no evaluator run occurs.

Self-hosted

Current `main` @ `3c3ca18eed76b164b418776d8d93cc1590e1d65b`

N/A (server-side worker behavior)

### Additional information

**Suggested fix**

- Add a partial unique constraint matching the dedup key, e.g. on `(projectId, jobConfigurationId, jobInputTraceId, jobInputDatasetItemId, jobInputObservationId)` (or at minimum `(projectId, jobConfigurationId, jobInputTraceId)` where dataset-item inputs are absent), plus a Prisma migration.
- Switch creation to insert-on-conflict-do-nothing / upsert and enqueue **only when this worker actually inserted** the row (this also fixes the stranded-PENDING case, since a redelivered job that loses the race still finds no enqueued work — pair it by moving the enqueue before the create with a compensating delete, or by reconciling stale PENDING rows).
- Alternative without schema change: wrap check+create in an interactive transaction with `SELECT … FOR UPDATE` on the config row to serialize per-config.

**Non-duplication:** searched open/closed issues and PRs for "jobExecution duplicate eval pending", "createEvalJobs race", "duplicate scores evaluator" — no existing report. (#16213 touched monitor scheduling only.)

wochinge:
🟡 **Live preview: [pr-16562.preview.langfuse.com](https://pr-16562.preview.langfuse.com)**

## Problem

`createEvalJobs` deduplicates trace/dataset eval executions with a read-then-write check: it queries `job_executions` for an existing row, and inserts one if it finds none. Three producers run this concurrently for the same trace — trace-upsert shard workers, dataset-run-item-upsert workers, and `CreateEvalQueue` workers — so two of them can both pass the check before either inserts. There is no unique constraint to stop the second insert; `job_executions` only has a plain index on `(project_id, job_configuration_id, job_input_trace_id)`.

Both rows then run the LLM-as-judge evaluation. Because score ids are derived from the *random* `jobExecutionId` (`createDeterministicEvalScoreId`), the two runs write to two different score ids, so users see duplicated scores and pay for the evaluation twice.

Second, smaller failure mode: the row is created *before* the execution job is enqueued. If enqueueing fails, the BullMQ redelivery finds the existing row, hits the dedup `continue`, and never enqueues — the execution stays `PENDING` forever with nothing to pick it up.

## Fix

1. **Derive the job execution id from the dedup key** instead of randomising it, so racing producers compute the same primary key.
2. **Insert with `createMany({ skipDuplicates: true })`** — an `INSERT ... ON CONFLICT DO NOTHING` on the primary key. The loser gets `count: 0` and skips enqueueing. No migration needed: the existing primary key does the work.
3. **Delete the row if enqueueing throws**, so the retry recreates and enqueues it rather than stranding it at `PENDING`.

Rows written before this change still carry random ids. They keep deduplicating correctly, because the existence check matches on the key columns, not on the id.

This mirrors what the observation-eval scheduler already does (`scheduleObservationEvals.ts`: deterministic `createW3CTraceId` id + `upsert`), which is why that path is not affected by this bug.

## Tests

Two new cases in `worker/src/__tests__/evalService.test.ts`, both verified to fail against the previous behaviour:

- `creates a single eval job when two producers race for the same trace` — previously produced 2 rows, now 1.
- `does not leave a PENDING eval job behind when enqueueing fails` — previously left an orphaned `PENDING` row that the retry skipped; now the row is gone and the retry recreates it.

## Not covered

A `SIGKILL` landing exactly between the insert and the enqueue still strands the row — the compensating delete cannot run if the process is gone. Closing that needs a reconciler for stale `PENDING` executions, which I left out of this PR.

## Verification

- `pnpm --filter worker run test evalService observationEval` — `Tests 292 passed (292)`
- `pnpm run typecheck` — `Tasks: 8 successful, 8 total`

Closes #16492

### DIFF
--- a/worker/src/__tests__/evalService.test.ts
+++ b/worker/src/__tests__/evalService.test.ts
@@ -799,62 +799,7 @@ Respond with JSON: {"score": <number>, "reasoning": "<explanation>"}`;
 
       const jobs = await prisma.jobExecution.findMany({ where: { projectId } });
 
-      // One row, so one score - the primary key rejected the second insert.
       expect(jobs.length).toBe(1);
-    }, 10_000);
-
-    test("still enqueues the eval when a concurrent producer won the insert", async () => {
-      const { projectId } = await createOrgProjectAndApiKey();
-      const traceId = randomUUID();
-
-      await upsertTrace({
-        id: traceId,
-        project_id: projectId,
-        timestamp: convertDateToClickhouseDateTime(new Date()),
-        created_at: convertDateToClickhouseDateTime(new Date()),
-        updated_at: convertDateToClickhouseDateTime(new Date()),
-      });
-
-      await createMigratedEvalConfig({
-        data: {
-          id: randomUUID(),
-          projectId,
-          filter: JSON.parse("[]"),
-          jobType: "EVAL",
-          delay: 0,
-          sampling: new Decimal("1"),
-          targetObject: EvalTargetObject.TRACE,
-          scoreName: "score",
-          variableMapping: JSON.parse("[]"),
-        },
-      });
-
-      // Simulate losing the insert race: the row is already there, so the
-      // ON CONFLICT DO NOTHING insert reports zero rows written. Racing on the
-      // real interleaving would make this assertion flaky.
-      const createManySpy = vi
-        .spyOn(prisma.jobExecution, "createMany")
-        .mockResolvedValue({ count: 0 });
-
-      const add = vi.fn().mockResolvedValue(undefined);
-      const getInstanceSpy = vi
-        .spyOn(EvalExecutionQueue, "getInstance")
-        .mockReturnValue({ add } as never);
-
-      try {
-        await createEvalJobs({
-          sourceEventType: "trace-upsert",
-          event: { projectId, traceId },
-          jobTimestamp,
-        });
-      } finally {
-        createManySpy.mockRestore();
-        getInstanceSpy.mockRestore();
-      }
-
-      // The loser must not stay silent. A duplicate execution is idempotent,
-      // whereas skipping the enqueue risks an execution nobody ever runs.
-      expect(add).toHaveBeenCalledTimes(1);
     }, 10_000);
 
     test("does not leave a PENDING eval job behind when enqueueing fails", async () => {
--- a/worker/src/__tests__/evalService.test.ts
+++ b/worker/src/__tests__/evalService.test.ts
@@ -43,7 +43,10 @@ vi.mock("@langfuse/shared/src/server", async () => {
 });
 
 // Import the mocked function
-import { generateLLMText } from "@langfuse/shared/src/server";
+import {
+  EvalExecutionQueue,
+  generateLLMText,
+} from "@langfuse/shared/src/server";
 import { UnrecoverableError } from "../errors/UnrecoverableError";
 
 let OPENAI_API_KEY = process.env.OPENAI_API_KEY;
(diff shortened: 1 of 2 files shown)


wochinge:
Things worth doubting here:

- **The deterministic id changes score identity.** Score ids are `uuidv5` over the job execution id, so any job execution that gets deleted and recreated for the same (config, trace, dataset item, observation) now writes to the *same* score id and overwrites the old score instead of adding a second one. I think that is the behaviour we want, and it matches the observation-eval path — but it is a semantic change beyond "stop the race", and it is the part I would look at hardest.

- **The compensating delete is scoped to `status: PENDING`.** If something flipped the row out of PENDING between the insert and the failed enqueue (I don't think anything can, since nothing else knows the id yet), the delete silently no-ops and we are back to a stranded row. Cheap insurance, but it is an assumption.

- **`skipDuplicates` now swallows a class of error.** A genuine primary-key collision from a bug elsewhere would previously throw; now it logs at debug and continues. I judged that acceptable because the id is fully determined by the dedup key, so a collision *is* a duplicate — but it is less loud than before.

- **The race test relies on `Promise.all` interleaving at the awaits** rather than real parallel workers. It fails reliably against the old code (2 rows), so it does prove the bug, but it is not a stress test.

Not fixed: a `SIGKILL` between insert and enqueue still strands the row. That needs a reconciler for stale PENDING executions; happy to do it separately if we think it is worth the moving parts.

PR #16587, opened 2026-08-26
feat(telemetry): report AI feature and Assistant enablement from self-hosted

LysanderKie:
🟡 **Live preview: [pr-16587.preview.langfuse.com](https://pr-16587.preview.langfuse.com)**

## Summary

The self-hosted telemetry event carries nothing about AI features today, so once the Assistant setup docs publish (langfuse-docs#3619) we would have no way to tell whether anyone enabled it.

Six new properties on the existing `telemetry` event:

| Property | Answers |
| --- | --- |
| `aiFeaturesEnabledOrganizations` | how many organizations enabled **AI Features** in settings |
| `totalOrganizations` | the denominator, so the count above is readable |
| `langfuseAiModelConfigured` | whether a usable instance-wide model is configured |
| `langfuseAiProvider` | which provider resolved, or null when no usable model exists |
| `assistantInstanceEnabled` | whether `LANGFUSE_IN_APP_AGENT_ENABLED` is on |
| `assistantRuns` | Assistant runs started in the existing reporting window |

Since `langfuseAiProvider` is null unless a usable config exists, a non-null provider is by itself the "properly configured" filter.

## Design notes

**Config fields resolve through `getInAppAgentModelConfig()`, not by re-reading env vars.** That function owns provider defaulting (unset means Bedrock), the "incomplete provider config counts as unconfigured" rule, and the invalid-Bedrock-region rule. Reading the env vars directly would let telemetry claim an instance is configured when the product refuses to run on it. The single test in here exists to hold that line.

**No model ID is sent.** With OpenAI-compatible endpoints now supported, it can be a private gateway's internal deployment name, which is infrastructure detail rather than configuration. The provider already answers the useful part.

**Runs are counted unconditionally**, not only when the feature is enabled. A zero on an instance that never switched it on is exactly the signal we want, and a conditionally-absent property is harder to chart than a zero.

All new queries sit inside the existing `try`, preserving the job's "catch all errors so telemetry does not break the application" contract.

## Scope

Deliberately excluded for now: run error codes and status histograms, sandbox posture, Ask AI volume, and any per-project or per-org breakdown. Sandbox posture is not observable from here at all, because telemetry runs in web while `LANGFUSE_IN_APP_AGENT_SANDBOX_*` is declared only in `worker/src/env.ts`; reading `process.env` in web would report "unset" for a correctly configured deployment.

## Packages

- `web` (telemetry job, new config resolver, one unit test)

Docs: the field table in `content/self-hosting/security/telemetry.mdx` enumerates every property we send under a promise that no raw content leaves the instance, so it gets the six new rows in a matching langfuse-docs change. All new fields are booleans, counts, and one enum value, so the promise holds.

## Test plan

- [x] `pnpm --filter web run test aiFeatureTelemetry` — `Tests 3 passed (3)`
- [x] Drift check: injecting the exact failure the test guards against (reading `env.LANGFUSE_AI_MODEL` and `env.LANGFUSE_AI_PROVIDER` directly instead of going through the resolver) gives `Tests 1 failed | 2 passed (3)`; restoring the resolver returns `Tests 3 passed (3)`
- [x] `pnpm run lint` — `Tasks: 7 successful, 7 total`
- [x] `pnpm tc` — `Tasks: 8 successful, 8 total`
- [ ] Live check against a throwaway PostHog project: clear `cron_jobs.last_run` for the `telemetry` row, hit `/api/public/health`, confirm one event arrives with the six properties. Needs `NODE_ENV=production` and `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` unset or the job returns early.
- [ ] Confirm the inverse: with `TELEMETRY_ENABLED=false` and no PostHog key, nothing is sent.

The PR extends self-hosted telemetry with aggregate AI-feature adoption, resolved model configuration, Assistant enablement, and Assistant-run activity.
- Counts organizations and AI-feature-enabled organizations.
- Reports whether a usable Assistant model configuration resolves and which provider it uses.
- Reports instance-level Assistant enablement and runs created during the telemetry window.
</details>

The PR appears safe to merge with no identified blocking or non-blocking issues.

The new telemetry fields use aggregate database counts and established server-side configuration helpers, while preserving the existing reporting window and error-isolation behavior.
</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(telemetry): inline AI feature c..."](https://github.com/langfuse/langfuse/commit/9ecf055de5a7e94cc2f65664ef4a000d9a28499a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57402699)</sub>


LysanderKie:
What to doubt here:

1. **This is the first product-feature telemetry the event has ever carried.** I traced every commit to `web/src/features/telemetry/index.ts`: the payload has been core-object volume plus instance identity since its original design, and `eeLicenseKey` is the only field ever added afterwards, for license compliance rather than product insight. Evals, prompt management, playground, code evaluators and monitors never added anything. So this is a convention change worth a conscious yes, not an incremental addition. On sensitivity it is a non-issue (booleans and counts are tamer than `userDomains`, which we have sent since 2024), but on precedent it opens a door.

2. **Nobody consumes these fields yet.** Without a PostHog insight pointed at them they are write-only and will rot. That is the thing that decides whether this was worth merging, and it is not in this PR.

3. **`totalOrganizations` is a new dimension, not just a denominator.** It happens to be useful on its own for sizing self-hosted deployments, which is arguably scope creep past what the ticket asks for. Easy to drop if you would rather keep the surface minimal.

4. **The reporting-window semantics are inherited, not chosen.** `assistantRuns` uses the same `startTimeframe`/`endTimeframe` as the existing counts, which means the very first run on an instance reports all-time rather than a 12-hour window, because `startTimeframe` is null then. Consistent with every other count in the event, but worth knowing before charting it.

5. **Adjacent, not fixed here.** `content/self-hosting/security/telemetry.mdx` states that enterprise telemetry cannot be disabled, but `ServerPosthog` returns a null client whenever `TELEMETRY_ENABLED === "false"`, so an EE instance sends nothing while still running every aggregate and then dropping the event. This PR adds three more queries to that wasted work. Either the code should enforce the claim or the doc should stop making it; I left it alone because it is a commercial call.

Not verified: the live PostHog round trip. Everything else in the test plan has a quoted result.

web/src/features/telemetry/index.ts
```diff
@@ -237,6 +238,23 @@ async function posthogTelemetry({
...
 
+    // AI features. Organizations opt in individually, so the enabled count
+    // needs the total as its denominator. Runs are counted unconditionally:
+    // zero on an instance that never enabled the Assistant is itself the
+    // answer we are looking for.
+    const totalOrganizations = await prisma.organization.count();
+    const aiFeaturesEnabledOrganizations = await prisma.organization.count({
+      where: { aiFeaturesEnabled: true },
+    });
+    const countAssistantRuns = await prisma.inAppAgentRun.count({
```
LysanderKie, line 252: TBD: Should we indicate only "SUCCEDED" runs

web/src/features/telemetry/aiFeatures.ts
nimarb, line 1: lets not add a new file to telemetry, inline what we need in index.ts

web/src/__tests__/server/unit/aiFeatureTelemetry.servertest.ts
nimarb, line 1: do u think a test is really required here? i agree setup is a bit convoluted; but we dont have any telemetry tests

PR #16596, opened 2026-08-26
security(api-keys): gate organization API key creation on the admin-api entitlement

Steffen911:
🟡 **Live preview: [pr-16596.preview.langfuse.com](https://pr-16596.preview.langfuse.com)**

## TL;DR

The `organizationApiKeys.create` tRPC mutation checked OWNER RBAC but never checked the `admin-api` entitlement, so an owner on a plan without that entitlement could mint organization-scoped API keys. This adds the missing `throwIfNoEntitlement` call.

## What was wrong

`create` enforced only `organization:CRUD_apiKeys`:

```ts
throwIfNoOrganizationAccess({ session, organizationId, scope: "organization:CRUD_apiKeys" });
// ...straight to createAndAddApiKeysToDb
```

The organization settings UI already hides the whole API keys page behind `useHasEntitlement("admin-api")`, so this was only reachable by calling tRPC directly. Every consuming org-scoped endpoint (`/api/public/organizations/*`, `/api/admin/organizations/*`) checks `admin-api` on its own, so a key created this way was inert until the organization upgraded. The impact is a commercial restriction bypass — no cross-tenant access.

## Fix

`throwIfNoEntitlement({ entitlement: "admin-api", ... })` after the RBAC check, the same pattern the admin API routes use around the same key-creation handler.

**Only creation is gated.** Listing, renaming and deleting stay open on purpose: an organization that downgrades must still be able to see and revoke keys issued while it was entitled. Gating those would strand live credentials, which would be a worse outcome than the bug being fixed. There is a test pinning that.

## Test evidence

Extended the existing suite in `web/src/__tests__/server/organizations-api-key-trpc.servertest.ts`. The existing `ownerSession` fixture is `admin: true`, and instance admins short-circuit `hasEntitlement`, so it could never have caught this — the new tests use non-admin OWNER sessions on `cloud:hobby` (no `admin-api`) and `cloud:team` (has it).

Three tests: unentitled owner is rejected with FORBIDDEN **and no key row is written**; entitled owner still succeeds; unentitled owner can still list and revoke a pre-existing key.

Revert-check — with the router change reverted, exactly the new gate test fails:

```
FAIL src/__tests__/server/organizations-api-key-trpc.servertest.ts > organizationApiKeys.create > owner on a plan without admin-api cannot create organization API keys
AssertionError: promise resolved "{ …(6) }" instead of rejecting
+   "secretKey": "sk-lf-...",
Tests  1 failed | 20 passed (21)
```

With the fix in place:

| Check | Result |
| --- | --- |
| `pnpm --filter web run test organizations-api-key-trpc` | `Tests  21 passed (21)` |
| `pnpm --filter web run test-client organization-settings-pages` | `Tests  4 passed (4)` |
| `pnpm run lint` | `Tasks: 7 successful, 7 total` |
| `pnpm --filter web run typecheck` | clean |

## Manual test steps

No UI change — the settings page was already hidden on unentitled plans, and the fix is server-side. To exercise it on `pr-<N>.preview.langfuse.com`, call `organizationApiKeys.create` via tRPC as an org owner whose plan lacks `admin-api`; it now returns FORBIDDEN instead of a key.

## Scope check

`organizationApiKeys.create` was the only ungated org-key creation path. I audited every `createAndAddApiKeysToDb` caller: the two admin API entry points already check `admin-api`, `initialize.ts` is PROJECT-scoped self-host provisioning, and the worker's in-app-agent key is a separate entitlement.

The PR closes a server-side commercial entitlement bypass by requiring the `admin-api` entitlement before organization API-key creation, while intentionally preserving listing and revocation after a downgrade.
- Adds the entitlement check after organization RBAC and before key persistence.
- Adds coverage for denied unentitled creation, successful entitled creation, and post-downgrade key listing and revocation.
</details>

The PR appears safe to merge and correctly narrows organization API-key creation without preventing credential cleanup after downgrades.

The requested organization is authorized before the entitlement check, current plan information is reconstructed for each request, and key persistence occurs only after both guards pass; the tests cover both plan outcomes and intentional downgrade behavior.
</details>

```mermaid
flowchart LR
    A[Organization owner calls create] --> B[Authenticate session]
    B --> C[Check organization membership and API-key RBAC]
    C --> D{Has admin-api entitlement?}
    D -- No --> E[Return FORBIDDEN]
    D -- Yes --> F[Create organization API key]
    F --> G[Persist key and write audit log]
    H[List or revoke existing key] --> I[Check organization RBAC]
    I --> J[Allow downgrade cleanup without entitlement gate]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["security(api-keys): gate organization AP..."](https://github.com/langfuse/langfuse/commit/970e6245b8f212a0fdd316c1b9eee20174fdab85) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57025757)</sub>

### DIFF
--- a/web/src/__tests__/server/organizations-api-key-trpc.servertest.ts
+++ b/web/src/__tests__/server/organizations-api-key-trpc.servertest.ts
@@ -10,6 +10,7 @@ import { appRouter } from "@/src/server/api/root";
 import { createInnerTRPCContext } from "@/src/server/api/trpc";
 import { TRPCError } from "@trpc/server";
 import { createAndAddApiKeysToDb } from "@langfuse/shared/src/server";
+import { randomUUID } from "crypto";
 
 describe("organization API keys trpc", () => {
   const organizationId = "seed-org-id";
@@ -31,6 +32,9 @@ describe("organization API keys trpc", () => {
     });
   });
 
+  // Instance admins bypass both RBAC and entitlement checks, so this session
+  // cannot cover plan gating. `entitledOwnerCaller` / `unentitledOwnerCaller`
+  // below are the non-admin owners used for that.
   const ownerSession: Session = {
     expires: "1",
     user: {
@@ -129,6 +133,48 @@ describe("organization API keys trpc", () => {
 
   const unAuthedCtx = createInnerTRPCContext({ session: null, headers: {} });
   const unAuthedCaller = appRouter.createCaller({ ...unAuthedCtx, prisma });
+
+  // Organization-scoped API keys are a paid feature (`admin-api` entitlement).
+  // These owners are deliberately not instance admins so the plan gate applies.
+  const ownerSessionOnPlan = (plan: SessionOrg["plan"]): Session => ({
+    expires: "1",
+    user: {
+      id: "user-1",
+      canCreateOrganizations: true,
+      name: "Demo User",
+      organizations: [
+        {
+          id: organizationId,
+          name: "Test Organization",
+          role: "OWNER",
+          plan,
+          cloudConfig: undefined,
+          metadata: {},
+          projects: [] as SessionOrg["projects"],
+        } as SessionOrg,
+      ],
+      featureFlags: {
+        excludeClickhouseRead: false,
+        templateFlag: true,
+      } as SessionFeatureFlags,
+      admin: false,
+    },
+    environment: {} as any,
+  });
+
+  const callerForSession = (session: Session) =>
+    appRouter.createCaller({
+      ...createInnerTRPCContext({ session, headers: {} }),
+      prisma,
+    });
+
+  // cloud:hobby lacks `admin-api`; cloud:team carries it.
+  const unentitledOwnerCaller = callerForSession(
+    ownerSessionOnPlan("cloud:hobby"),
+  );
+  const entitledOwnerCaller = callerForSession(
+    ownerSessionOnPlan("cloud:team"),
+  );
 
   describe("organizationApiKeys.byOrganizationId", () => {
     it("owner can fetch organization API keys", async () => {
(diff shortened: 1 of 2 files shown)


Steffen911:
Things worth doubting here:

**The self-hosted plans.** `oss` and `self-hosted:pro` do not carry `admin-api`, so on those deployments org owners can no longer create organization API keys through the UI path. I believe that is already the status quo rather than a regression — the settings page is hidden on those plans, and the consuming `/api/public/organizations/*` endpoints reject the key anyway — so the change only stops handing out keys that never worked. But if any self-hoster is creating org keys via tRPC and using them somewhere I did not find, this turns a silent no-op into a hard FORBIDDEN. Worth a second opinion from whoever knows the self-host story best.

**Gating only `create` is a judgment call.** `byOrganizationId`, `updateNote` and `delete` stay ungated so a downgraded org can still revoke keys it was issued while entitled. The alternative reading of the ticket is "gate the whole router", which is tidier but would strand live credentials on downgrade. I pinned my reading with a test, so if you disagree the test is the thing to argue with.

**`throwIfNoEntitlement` trusts the session.** The org plan comes from `ctx.session.user.organizations`, not a fresh DB read, so the gate is only as fresh as the JWT. Same as every other caller of this helper, and an upgrade/downgrade lag of one session refresh is not interesting for a P4 commercial gate — but it does mean this is not a hard authorization boundary, and it should not be treated as one.

**The test fixture footgun this exposed.** The existing `ownerSession` in that suite is `admin: true`, and instance admins short-circuit `hasEntitlement`, so the pre-existing "owner can create" tests would have passed no matter what entitlement I required. There are probably other suites with the same admin-flagged "owner" fixture giving false confidence about entitlement gates. Out of scope here, but someone may want to look.

PR #16599, opened 2026-08-26
fix(automations): require fresh tokens for dispatch URL changes (v3 backport of #16479)

Steffen911:
Backport of #16479 to `v3`.

## TL;DR

Changing the dispatch URL on an existing GitHub-dispatch automation re-used the stored PAT against the new URL. A fresh token is now required whenever the URL changes, server-side and in the form.

## Why it matters on v3

`processGitHubDispatchActionConfig` fell back to `existingActionConfig.githubToken` whenever the request omitted a token, without ever comparing the incoming URL to the stored one. Anyone able to edit an automation could therefore repoint an existing action at a URL they control and have Langfuse replay the org's stored GitHub PAT to it — the token is never returned to the client, so the URL swap is the whole exploit.

## Cherry-pick

Clean, no adaptation. `git diff origin/v3..HEAD` is line-for-line identical to `git show bbc391cf57` — 205 content lines across 7 files, only `@@` hunk offsets differ.

No new dependencies.

## Verification

```
pnpm --filter web run test automations-trpc     → Tests  44 passed (44)
pnpm --filter web run test-client GitHubDispatchActionHandler
                                               → Tests  2 passed (2)
```

With `githubDispatchHelpers.ts` reverted to `origin/v3`, tests untouched:

```
Tests  1 failed | 43 passed (44)
```

The failing case is the URL-change guard: v3 returned the *updated* action instead of rejecting.

With `GitHubDispatchActionHandler.tsx` reverted to `origin/v3`:

```
Tests  1 failed | 1 passed (2)
```

v3's handler returned `{ isValid: true, errors: undefined }` where the fix returns `GitHub token is required when changing the dispatch URL`.

Both restored afterwards; suites re-run green and `git status --porcelain` is empty.

This backport prevents an existing GitHub-dispatch PAT from being reused after its dispatch URL changes.

- Adds a shared URL-equivalence helper used by client and server validation.
- Tracks the original URL in edit forms and prompts for a fresh token when the destination changes.
- Rejects unsafe updates before persistence while preserving the token for equivalent or unchanged URLs.
- Adds client and server regression coverage for changed and canonically equivalent URLs.
</details>

The PR appears safe to merge and closes the stored-PAT destination-change vulnerability without introducing an actionable regression.

The server performs destination comparison before persistence, requires a nonblank fresh token for changed URLs, preserves credentials only for equivalent destinations, and remains authoritative when client state differs.
</details>

```mermaid
flowchart TD
  A[Edit GitHub-dispatch automation] --> B{Submitted URL equivalent to stored URL?}
  B -->|Yes| C{New PAT supplied?}
  C -->|Yes| D[Encrypt and store new PAT]
  C -->|No| E[Preserve stored encrypted PAT]
  B -->|No| F{Fresh PAT supplied?}
  F -->|No| G[Reject update before transaction]
  F -->|Yes| D
  D --> H[Persist updated action]
  E --> H
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(automations): require fresh tokens f..."](https://github.com/langfuse/langfuse/commit/7a35a46c0f0a8a98c0c8c983ec59718fd96a6015) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57026807)</sub>

**Context used:**

- Knowledge Base — [Integration delivery and notifications](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/integration-delivery-and-notifications.md)


Steffen911:
Worth doubting: `areGitHubDispatchUrlsEquivalent` compares `new URL(a).href === new URL(b).href`, so a trailing-slash or default-port change counts as "same URL" and will not force a fresh token, while a case change in the path will. That is upstream behaviour verbatim, but it is the line where a reviewer should decide whether normalising is the right call for a security check — a permissive comparison is the failure direction here.

Also: the guard only fires when `existingActionConfig !== undefined`, so it constrains edits, not creates. Correct as written, but it means the protection depends entirely on the update path passing the stored config in.

PR #16487, opened 2026-08-24
fix: enforce OTLP ingestion request body limits

sumerman:
🟡 **Live preview: [pr-16487.preview.langfuse.com](https://pr-16487.preview.langfuse.com)**

## Summary

- Add a configurable limit for encoded and decompressed OTLP request bodies.
- Reject oversized requests with HTTP 413 responses and close encoded-body connections safely.
- Add gzip decompression limits and focused request-body tests.

<details open><summary><h3>Greptile Summary</h3></summary>

The PR adds configurable encoded and decompressed size limits to the public OTLP traces endpoint.

- Reads request streams through a bounded raw-body helper.
- Bounds gzip output and maps both overflow stages to HTTP 413 responses.
- Adds environment configuration, dependency metadata, telemetry, and focused request-body tests.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge with no concrete blocking or independently actionable issues identified.

The changed path consistently bounds encoded and decompressed OTLP bodies, maps verified overflow conditions to HTTP 413, and handles encoded overflow connections explicitly.
</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  A[OTLP request] --> B{Encoded body within limit?}
  B -->|No| C[Drain request and close connection]
  C --> D[Return HTTP 413]
  B -->|Yes| E{Content-Encoding includes gzip?}
  E -->|No| H[Decode and validate OTLP payload]
  E -->|Yes| F{Decompressed body within limit?}
  F -->|No| D
  F -->|Yes| H
  H --> I[Continue ingestion]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Track OTEL request body limit exceedance..."](https://github.com/langfuse/langfuse/commit/4c9046b548fbb262d0084c325de39fc7a54235c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56310205)</sub>

**Context used:**

- Knowledge Base — [Trace and observation ingestion](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/trace-observation-ingestion.md)
- Knowledge Base — [API and SDK contracts](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/api-and-sdk-contracts.md)

PR #16600, opened 2026-08-26
fix: enforce OTLP ingestion request body limits (v3 backport of #16487)

Steffen911:
Backport of #16487 to `v3`.

## TL;DR

The OTLP traces endpoint buffered request bodies with no size ceiling, and gunzipped them with no ceiling either. Both reads are now bounded by `LANGFUSE_OTEL_INGESTION_MAX_BODY_BYTES` and answer oversized posts with a 413.

## Why it matters on v3

`/api/public/otel/v1/traces` read the stream into an unbounded `Buffer.concat`, then optionally gunzipped it — so a single large or highly compressible post is bounded only by container memory. This is the mechanism behind the `web-ingestion` OOM + ALB-5xx pattern we already saw in production against multi-hundred-MB OTLP posts; v3 has the same unbounded read.

## Cherry-pick

Text merge was clean. The one apparent divergence from the upstream patch is cosmetic: git attributed an identical `}` line to context on main and to an addition on v3. Verified equivalent by comparing resulting file content rather than patch text — the delta between this branch's `traces/index.ts` and upstream's post-image is the same 123-line `validateOtelSpanIds` block that already differed *before* the pick (main-only work that was deliberately not backported), so the pick introduced exactly the upstream change and nothing else.

### Lockfile

This commit adds a direct dependency (`raw-body@3.0.2`). Rather than trust the merged `pnpm-lock.yaml`, I re-ran `pnpm install` and diffed:

```
NO CHANGE — merged lockfile was already what pnpm resolves
✓ Lockfile passes supply-chain policies (2271 entries in 5.1s)
```

`raw-body@3.0.2` was already in the lockfile as a transitive dep, so the merge only needed the `web` importer entry.

## Verification

```
pnpm --filter web run test otel-request-body    → Tests  7 passed (7)
```

With `traces/index.ts` and `env.mjs` reverted to `origin/v3` (the new `otelRequestBody.ts` helper kept, so its 5 unit tests still stand), tests untouched:

```
Tests  2 failed | 5 passed (7)
```

Both route-level cases fail, and they fail in the telling way — v3 swallowed the oversized body and fell through to `Failed to parse OTel JSON Trace` instead of `OTel request body exceeds the 64 bytes limit after decompression`. Suite duration also went from 0.36s to 30.3s, which is the unbounded read itself showing up on the clock.

The PR backports bounded OTLP trace-body handling to v3, limiting both encoded request reads and decompressed gzip output.

- Adds a configurable 512 MiB default limit with development and production environment documentation.
- Returns a structured 413 response and records telemetry when either limit is exceeded.
- Adds focused tests for declared-length, chunked, compressed, and route-level overflow cases.
- Adds the directly pinned `raw-body@3.0.2` dependency and corresponding lockfile importer entry.
</details>

The PR appears safe to merge, with no actionable correctness or security defects identified in the changed body-limit path.

Encoded and decompressed OTLP payloads are independently bounded, oversized requests preserve the intended 413 response through the public-route wrapper, and the configuration, dependency lock, and focused tests align with the implementation.
</details>

```mermaid
flowchart TD
  A[Authenticated OTLP trace request] --> B{Encoded body exceeds limit?}
  B -->|Yes| C[Drain request and return 413]
  B -->|No| D[Read bounded body]
  D --> E{Content-Encoding includes gzip?}
  E -->|No| H[Parse JSON or protobuf]
  E -->|Yes| F[Gunzip with bounded output]
  F --> G{Decompressed body exceeds limit?}
  G -->|Yes| I[Return 413]
  G -->|No| H
  H --> J[Validate and queue ingestion]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: enforce OTLP ingestion request body..."](https://github.com/langfuse/langfuse/commit/7c9441497771202c1b94696d627646a5634523a4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57026853)</sub>

**Context used:**

- Knowledge Base — [API and SDK contracts](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/api-and-sdk-contracts.md)


Steffen911:
Worth doubting: the `}`-attribution wobble I describe in the body. The text merge was clean, but the upstream patch and this branch disagree on which identical `}` is context and which is an addition. I concluded they are equivalent by diffing resulting *file content* against upstream's post-image rather than comparing patch text — the residual delta is the same 123-line `validateOtelSpanIds` block that already differed before the pick. If you want to double-check my reasoning, that comparison is the thing to re-run, not the patch diff.

Second: the two route-level tests pin exact error strings (`OTel request body exceeds the 64 bytes limit after decompression`). That is what made the revert-check legible, but it is also brittle if the message is ever reworded.

PR #16601, opened 2026-08-26
fix: handle malformed Unicode escapes in ClickHouse JSON across v3 and v4 (v3 backport of #16559)

Steffen911:
Backport of #16559 to `v3`.

## TL;DR

Malformed Unicode escapes in ClickHouse JSON output are now handled explicitly: the client picks `no_throw` or `sanitize` based on the server version, with `LANGFUSE_JSON_BAD_UNICODE_ESCAPE` as the override.

## Why it matters on v3

`input_format_json_throw_on_bad_escape_sequence` only exists from ClickHouse 24.4, so v3 deployments on older servers had no way to keep a single lone surrogate from failing a read. The backport gives v3 the same version-aware resolution main has, plus the `stringifyJsonWithSanitizedSurrogates` write-side helper.

## Cherry-pick

Clean, no adaptation. `git diff origin/v3..HEAD` is line-for-line identical to `git show 036732d4d9` — 212 content lines across 10 files, only `@@` hunk offsets differ.

No new dependencies.

## Verification

```
pnpm --filter @langfuse/shared run test src/server/clickhouse/
                                               → Test Files  4 passed (4)
                                                  Tests  28 passed (28)
```

All three suites are fully mocked (`vi.mock` on `../../env` and `@clickhouse/client`), so no live ClickHouse is involved.

With `client.ts`, `compatibility.ts` and `env.ts` reverted to `origin/v3` (the new `json.ts` kept so its own unit test still stands), tests untouched:

```
Tests  9 failed | 19 passed (28)
```

`resolveClickHouseJsonBadUnicodeEscapeMode is not a function` for the 6 compatibility cases, plus the client-side caching case that expects a new cached client when the resolved mode changes.

The PR adds version-aware handling for malformed Unicode escapes, selecting either a ClickHouse parser setting or client-side surrogate sanitization and exposing an environment override.

- Adds the `LANGFUSE_JSON_BAD_UNICODE_ESCAPE` configuration to environment validation and Docker Compose.
- Resolves behavior from the detected ClickHouse version and includes it in client construction and caching.
- Adds a JSON serializer that replaces lone UTF-16 surrogates and tests the compatibility modes and cache behavior.
</details>

The PR should not merge until JSON compatibility is resolved per configured ClickHouse service, because an older read replica can reject the newly applied setting.

A version detected exclusively from the read-write endpoint controls clients for separately configured read-only endpoints, allowing a ClickHouse 24.4+ primary to make requests against a pre-24.4 replica fail with an unsupported setting.

**Files Needing Attention:** packages/shared/src/server/clickhouse/compatibility.ts, packages/shared/src/server/clickhouse/client.ts
</details>

```mermaid
flowchart LR
  Detect[Detect version from CLICKHOUSE_URL] --> Resolve{Version at least 24.4?}
  Resolve -->|Yes| NoThrow[Select no_throw]
  Resolve -->|No or unknown| Sanitize[Select sanitize]
  NoThrow --> RW[ReadWrite client]
  NoThrow --> RO[ReadOnly client]
  NoThrow --> ERO[EventsReadOnly client]
  RO --> Risk[Older independent replica may reject setting]
  ERO --> Risk
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/clickhouse/compatibility.ts:144-148
**Global version misconfigures replicas**

If a read-only ClickHouse endpoint runs a pre-24.4 version while `CLICKHOUSE_URL` runs 24.4 or newer, the globally detected primary version selects `no_throw` for the read-only client too, causing requests to that replica to fail because it does not support `input_format_json_throw_on_bad_escape_sequence`.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: handle malformed Unicode escapes in..."](https://github.com/langfuse/langfuse/commit/0912c991631fd2f345c794c823b83dba3e8e8ca4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57026869)</sub>


Steffen911:
Worth doubting: `resolveClickHouseJsonBadUnicodeEscapeMode` takes an `applicationVersion` and the upstream test table asserts different defaults for `v3.225.4` vs `v4.17.0`. Those cases pass here, but the v3-vs-v4 branching inside a v3-only build is the part I would want a second read on — it is deliberately cross-version logic living on a maintenance branch, and it is easy to convince yourself the v3 arm is right just because the test says so.

Also note all 28 tests are fully mocked — `vi.mock` on `../../env` and `@clickhouse/client`. Nothing here was exercised against a real ClickHouse of either version, so the version-band boundaries are asserted, not observed.

PR #6143, opened 2025-03-21
feat: add blob storage integration

Steffen911:
> [!IMPORTANT]
> Add blob storage integration with support for AWS S3, S3-compatible, and Azure Blob Storage, including scheduled exports and UI configuration.
> 
>   - **Behavior**:
>     - Adds blob storage integration with support for AWS S3, S3-compatible, and Azure Blob Storage.
>     - Scheduled exports of trace data to configured storage.
>     - Supports hourly, daily, and weekly export frequencies.
>   - **Database**:
>     - Adds `BlobStorageIntegration` model to `schema.prisma`.
>     - Creates migration `20250324110557_add_blobstorage_integration_table` for new table.
>   - **API**:
>     - Adds `blobStorageIntegrationRouter` to handle CRUD operations for blob storage configurations.
>     - Updates `appRouter` in `root.ts` to include `blobStorageIntegrationRouter`.
>   - **UI**:
>     - Adds blob storage integration settings page in `blobstorage.tsx`.
>     - Updates project settings to include blob storage integration in `index.tsx`.
>   - **Queues**:
>     - Adds `BlobStorageIntegrationQueue` and `BlobStorageIntegrationProcessingQueue` in `queues.ts`.
>     - Implements processors for handling blob storage integration jobs in `blobStorageIntegrationQueue.ts`.
>   - **Tests**:
>     - Adds tests for blob storage integration processing in `blobStorageIntegrationProcessing.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse&utm_source=github&utm_medium=referral)<sup> for f5fcaab7778232ef6e8af7c5f61eb22361e08980. It will automatically update as commits are pushed.</sup>

PR #16613, opened 2026-08-26
fix(blob-export): half-open export time window to stop chunk-boundary double-count

niklassemmler:
🟡 **Live preview: [pr-16613.preview.langfuse.com](https://pr-16613.preview.langfuse.com)**

## What does this PR do?

The four blob-storage export ClickHouse query builders filtered their time window **inclusive on both ends** (`timestamp >= min AND timestamp <= max`). Every run/chunk handoff sets the next window's inclusive `minTimestamp` to the previous window's `maxTimestamp` (`lastSyncAt = maxTimestamp`), so a row whose timestamp landed **exactly** on a window boundary was exported in **both** adjacent windows — a silent duplicate for consumers doing exactly-once/idempotent processing across windows.

Fix: make the upper bound **exclusive** (`timestamp < max`) in all four builders, matching the sibling analytics-integration exporters (PostHog/Mixpanel), which already use the half-open convention. Because the next window's inclusive `minTimestamp` equals this window's exclusive `maxTimestamp`, a boundary row moves to the next window and is still exported exactly once — no row is dropped at the seam.

Affected builders (Parquet and raw variants share these, so all output formats are covered):
- `buildTracesForBlobStorageExportQuery`
- `buildObservationsForBlobStorageExportQuery`
- `buildScoresForBlobStorageExportQuery`
- `buildEventsForBlobStorageExportQuery`

Adds a construction regression test (`blobStorageExportWindow.test.ts`) asserting each builder emits an inclusive lower bound and an exclusive upper bound, plus an explicit two-window seam assertion. The test fails against the previous inclusive-inclusive code and passes after the fix.

## Type of change

## Checklist

- [x] `pnpm run lint` / typecheck pass (shared: 4/4 tasks)
- [x] Targeted tests pass: shared `blobStorageExportWindow` (5/5), worker `blobStorageTuningWiring.unit` (3/3)
- [x] Confirmed the regression test is red against the old behavior and green after the fix

This PR changes all four blob-storage export queries to use half-open time windows, preventing records exactly on a chunk boundary from being emitted by both adjacent exports.
- Replaces inclusive upper bounds with exclusive upper bounds for trace, observation, score, and event exports.
- Adds regression coverage for the generated lower and upper time-window predicates.
- Confirms the adjacent-window convention in which the previous maximum becomes the next inclusive minimum.
</details>

The PR appears safe to merge because the exclusive upper bounds match the export worker’s adjacent-window handoff and preserve each boundary record exactly once.

The worker advances a successful export’s position to the exact prior maximum, which becomes the next window’s inclusive minimum, so the changed predicates remove boundary duplicates without creating a gap.
</details>

```mermaid
flowchart LR
  A["Window N: min ≤ timestamp < max"] --> B["Persist lastSyncAt = max"]
  B --> C["Window N+1: max ≤ timestamp < nextMax"]
  D["Row timestamp = max"] --> C
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(blob-export): make export time windo..."](https://github.com/langfuse/langfuse/commit/3812e3cb99037aa8d29b4ada4ec59088f5a25419) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57045067)</sub>

**Context used:**

- Knowledge Base — [Media and blob processing](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/media-and-blob-processing.md)

### DIFF
--- a/packages/shared/src/server/repositories/blobStorageExportWindow.test.ts
+++ b/packages/shared/src/server/repositories/blobStorageExportWindow.test.ts
@@ -1,0 +1,113 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+// The blob-export builders must filter their time window half-open:
+// `timestamp >= minTimestamp AND timestamp < maxTimestamp`. Every run/chunk
+// handoff sets the next window's inclusive `minTimestamp` to the previous
+// window's `maxTimestamp` (lastSyncAt = maxTimestamp). With an inclusive upper
+// bound a row landing exactly on the boundary was exported in BOTH adjacent
+// windows (duplicate); with a half-open upper bound it moves to the next window
+// and is exported exactly once. See the analytics-integration exporters, which
+// already use `< maxTimestamp`.
+
+const mockQueryClickhouseStream = vi.hoisted(() =>
+  vi.fn(() => (async function* () {})()),
+);
+
+vi.mock("./clickhouse", () => ({
+  queryClickhouse: vi.fn(),
+  queryClickhouseStream: mockQueryClickhouseStream,
+  queryClickhouseStreamRawText: mockQueryClickhouseStream,
+  queryClickhouseExecRaw: vi.fn(),
+  commandClickhouse: vi.fn(),
+  upsertClickhouse: vi.fn(),
+  parseClickhouseUTCDateTimeFormat: vi.fn(),
+  clickhouseCompliantRandomCharacters: vi.fn(() => "x"),
+  BLOB_EXPORT_PARQUET_CLICKHOUSE_SETTINGS: {},
+}));
+
+// Break the query-options -> server-barrel import cycle and avoid a real
+// ClickHouse round-trip for the observations FINAL decision.
+vi.mock("../queries/clickhouse-sql/query-options", () => ({
+  shouldSkipObservationsFinal: vi.fn().mockResolvedValue(false),
+}));
+
+import { getTracesForBlobStorageExport } from "./traces";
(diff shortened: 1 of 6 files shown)


niklassemmler:
> 🤖 Generated by Claude, posted on behalf of the author.

What a reviewer should doubt here:

- **The regression test guards the SQL convention, not end-to-end de-dup.** `blobStorageExportWindow.test.ts` asserts each builder emits `< {maxTimestamp` and never `<= {maxTimestamp` (string-level), because no integration test hits ClickHouse for these builders. It proves the operator flipped and would catch a revert; it does *not* prove a real boundary row lands in exactly one output object at runtime. If you want that guarantee, it'd need a ClickHouse-backed test inserting a row at exactly `maxTimestamp` across two windows.
- **The exactly-once seam relies on the handoff invariant `nextMin = lastSyncAt = prevMax`.** The fix is only correct because the next window's *inclusive* `minTimestamp` equals this window's *exclusive* `maxTimestamp` (see `getMinTimestampForExport`). If any caller ever advances `lastSyncAt` to `maxTimestamp + epsilon`, a boundary row would be dropped instead. Worth confirming that invariant holds on every handoff path (catch-up chunks + steady-state).
- **Original intent.** Inclusive-inclusive was founding code (langfuse/langfuse#6143). The sibling analytics exporters already use half-open, which is the evidence it was an oversight rather than a deliberate choice — but the original author is the authority on that.

Steffen911 (approved):

PR #16617, opened 2026-08-26
fix(in-app-agent): teach traces vs observations in the system prompt

LysanderKie:
🟡 **Live preview: [pr-16617.preview.langfuse.com](https://pr-16617.preview.langfuse.com)**

## What does this PR do?

Users still think in traces. The in-app agent has no traces tools; traces are observations that share a `traceId` / `traceName`.

This adds a `<data_model>` block to the system prompt:
- list or count traces as observations with `isRootObservation` true
- a named trace ID lists all observations for that `traceId`
- cost, tokens, and generation latency stay on child observations; do not filter those metrics to roots
- replies still say "traces" unless the user asks about the data model

MCP tool descriptions are unchanged. Cloud production does not pick this up until `in-app-agent-system-prompt` is synced with `scripts/in-app-agent/sync-prompt.sh` (one run, confirm each region).

## Type of change

## Mandatory Tasks

- [x] Self-reviewed

## Checklist

- [x] Lint passed on commit (`Tasks: 7 successful, 7 total`)

## Packages

- `packages/shared` — system prompt template

## Verification

Skipped tests, typecheck, and browser: this is static prompt copy.

## Follow-up

Sync the managed `in-app-agent-system-prompt` so Cloud prompt management matches the repo template.

[Slack Thread](https://clickhouse-inc.slack.com/archives/D0BFFER27A6/p1787652355941609?thread_ts=1787652355.941609&cid=D0BFFER27A6)


LysanderKie:
What to doubt in review
- Cloud still serves the old managed prompt until someone runs `sync-prompt.sh`. Local/dev is the only path that gets this immediately.
- “How many traces?” is now root-observation count, not `uniq(traceId)`. They matched on Cloud this week, but they can diverge if a trace has multiple roots.
- The prompt tells the model not to filter cost/tokens/latency to roots. If that line is too easy to miss, the agent can still undercount those metrics.

PR #16622, opened 2026-08-26
feat(shared): add toolCallInvocations measure for per-tool invocation counts

nmtrang29:
🟡 **Live preview: [pr-16622.preview.langfuse.com](https://pr-16622.preview.langfuse.com)**

## What changed

Adds a per-tool invocation measure without introducing a second array-expansion model.

- `calledToolNames` now explodes `arrayDistinct(tool_call_names)`. The query grain is one row per observation and distinct called tool, so repeated names do not duplicate sibling metrics.
- `toolCallInvocations` uses `countEqual(@@AGG1@@(tool_call_names), calledToolNames)` with `agg1: any` and `requiresDimension: "calledToolNames"` in both observation view versions.
- `requiresDimension` now carries either `explodeArray` or `pairExpand` machinery when it auto-includes a dimension. The existing `costByType` example remains in the builder comment, with the tool invocation case added alongside it.
- Query declaration comments now document the semantic difference between independent `explodeArray` expansion and positional `pairExpand` expansion.

## Query mechanics

In the two-level query, `calledToolNames` is a grouping key and `any(tool_call_names)` returns the original array, which is constant inside each observation/tool group. `countEqual` recovers the number of repeated calls to that tool, and the outer aggregation combines those per-observation counts.

In the single-level events query, the template wrapper is stripped and the requested aggregation is applied directly to `countEqual(tool_call_names, calledToolNames)`. This keeps the optimized events path single-level while returning the same counts.

## Compatibility

Existing widgets using `calledToolNames` keep the same dimension values, but repeated occurrences of one tool inside an observation are consistently treated as one observation/tool row. Invocation totals should use `sum(toolCallInvocations)`; observation counts should continue to use `count(count)`.


sumerman:
@claude review

PR #16661, opened 2026-08-27
feat(auth): gate public-API project routes through the policy seam in all migration modes

marksalpeter:
🟡 **Live preview: [pr-16661 app preview](https://pr-16661.preview.langfuse.com)**

## What does this PR do?

> PR title must follow Conventional Commits, for example `feat(web): add trace filters` or `fix: handle empty dataset names`.

Fixes # (issue)

## Type of change

## Mandatory Tasks

## Checklist


marksalpeter:
VERDICT: FAIL (spec incomplete). Type checks, lint, and pure-core unit tests pass; the code that exists is sound. But required acceptance criteria are unmet.

Checks run (all green):
- pnpm --filter web run typecheck: exit 0
- pnpm --filter worker run typecheck: exit 0
- pnpm --filter web run lint: exit 0
- policy unit tests (src/__tests__/server/unit/policy/*): Tests 49 passed (49)
NOTE: The HTTP servertest suite (Basic/Bearer/shadow/enforce) could NOT be run here — needs a Postgres/ClickHouse/Redis stack + a web server on the test DB. Left unverified.

Spec gaps vs LFE-15678 acceptance criteria (decisive):
1. Shadow servertests missing. Criterion "every project-route response is byte-identical to legacy; a legacy-blocked request stays blocked; expected parity cell + coverage counter recorded on the span" — no such HTTP servertest was added. (Author's own commit body admits: "The shadow byte-identical and enforce 200/403 HTTP servertests are not yet added.")
2. Enforce servertests missing. Criterion "correct 200/403 per scope across project routes" — not added.
3. Fern regeneration not done. fern/apis/server/definition/api.yml adds the Bearer auth-scheme, but generated/** was NOT regenerated (diffstat touches no generated files). Criterion "Fern regeneration for the Bearer scheme pass" and repo Verification rules (public API contract changes must update+regenerate Fern) are unmet — the source now diverges from the committed generated clients.
   Also web/src/__tests__/server/policy-auth.servertest.ts is written but, per the author, unrun.

Standards concern (non-decisive but real):
- web/src/features/auth/policy/shadow.projects.ts verifyAuth() calls enforceProjectAuth() UNCONDITIONALLY, before the env-mode branches. In legacy mode (the default, and the mode self-host stays on) the full new pipeline — authenticate() Verifier redis/Postgres lookups, resolveContext, authorize, plus the Verifier's fast-hash backfill Postgres write — runs on every project-route request and the result is discarded. This doubles auth DB/redis load for every default/self-host deployment, contrary to the spec's intent that self-host is unchanged. Gate the new pipeline on env !== "legacy".

The pure core (authorize deny-overrides/deny-by-default/kind-gating/admin wildcard, resolveContext keying+invalidate, verifier scheme dispatch, diffResults) is well covered by the 49 unit tests and matches the spec design. The failure is the missing HTTP-level shadow/enforce/Bearer-Basic servertests and the un-regenerated Fern clients — the criteria that prove the seam is actually byte-identical and correct in prod.

packages/shared/src/features/rbac/projectAccessRights.ts
```diff
@@ -83,6 +83,18 @@ export const projectScopes = [
 
   "alerts:read",
   "alerts:CUD",
+
+  // Public-API action tokens; not granted to any UI role.
```
marksalpeter: we should grant these to their respective user roles now. we don't need user role gating to be part of this pr, but we should assign them where we think they should go.

web/public/generated/api/openapi.yml
```diff
@@ -16,6 +16,10 @@ info:
     - password: Langfuse Secret Key
 
 
+    A private (secret) key may also authenticate with `Authorization: Bearer
```
marksalpeter: this was implemented too early. in shadow mode, this is not true. implementation is also backwards compatible. updating the docs comes after release.

fern/apis/server/definition/api.yml
```diff
@@ -7,13 +7,25 @@ docs: |
...
 
+  A private (secret) key may also authenticate with `Authorization: Bearer <secret_key>`.
+
   ## Exports
 
   - OpenAPI spec: https://cloud.langfuse.com/generated/api/openapi.yml
 
 error-discrimination:
   strategy: status-code
-auth: basic
```
marksalpeter, line 16: updating docs comes last, well after shadowing and release phases. rollback.

.env.prod.example
```diff
@@ -243,6 +243,9 @@ LANGFUSE_ENABLE_BLOB_STORAGE_FILE_LOG=true
 # Cache configuration
 # LANGFUSE_CACHE_API_KEY_ENABLED=
 # LANGFUSE_CACHE_API_KEY_TTL_SECONDS=
+
+# Public-API authorization migration: legacy (default) | shadow | enforce
+# PUBLIC_API_AUTHZ_MIGRATION=
```
marksalpeter: there should be a single env variable that moves all auth, not just the public api. it can start here with juts the public api as `API_AUTH_MIGRATION` (indicates that we're migrating authn and authz, which is true).

web/src/__tests__/server/in-app-agent-api-route-auth.servertest.ts
```diff
@@ -54,6 +54,7 @@ describe("in-app agent public API route auth", () => {
   async function callRoute(params: { allowInAppAgentKey?: boolean }) {
     const handler = createAuthedProjectAPIRoute({
       name: "Test Route",
+      action: null,
```
marksalpeter: lets not force null actions to be passed. params can have `action?`

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,195 @@
...
+export type ContextCache = {
+  read: (key: string) => Promise<AuthorizationContext | undefined>;
+  write: (key: string, context: AuthorizationContext) => Promise<void>;
+  invalidate: (target: InvalidateParams) => Promise<void>;
+};
+
+/** InvalidateParams selects the entries to evict: one api key's contexts, or every context under an org. */
+export type InvalidateParams = { apiKeyId: string } | { orgId: string };
+
+/** OrgEnrichment is the org's PrincipalOrganization the resolver fetches at materialization. */
```
marksalpeter: loose the alias.

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,195 @@
...
+  type ErrorResult,
+  type Policy,
+  type Principal,
+  type PrincipalOrganization,
+  type Resource,
+  type SystemPolicy,
+  type Success,
+} from "./types";
+
+/** notWiredOrgRepo is the default org enricher until the Verifier slice injects the prisma-backed one. */
```
marksalpeter: If an org repo is not wired by default we can just omit it for others to inject into the structure

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,195 @@
...
+function adminContext(): AuthorizationContext {
+  const principal: Principal = { kind: "admin", userId: null };
+  return {
+    principal,
+    policies: apiKeyAccessRights.ADMIN.map((policy) => bind(policy, principal)),
+  };
+}
+
+/** createContextCache is the in-memory ContextCache with org and apiKey reverse indices for invalidation. */
+function createContextCache(): ContextCache {
```
marksalpeter: redis cache must land in this branch.

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,195 @@
...
+  | { authorization: "publicKey" | "privateKey"; apiKey: ApiKey }
+  | { authorization: "adminKey" };
+
+/** ResolveContextResult is the resolved context or a typed failure; resolveContext returns, never throws. */
+export type ResolveContextResult =
+  | (Success & { context: AuthorizationContext })
+  | ErrorResult<InternalServerError>;
+
+/** OrgRepo resolves an org's enrichment — the prisma-backed collaborator. */
+export type OrgRepo = {
```
marksalpeter: I  expect something like `type OrganizationRepository { getByOrgId(id:string): Organization; }` it should probably return raw prisma models or an error extending from BaseError (eg GetOrgByIdResult = Success & { organization: Organization } | ErrorResult<InternalImplementationError|NotFoundError>

web/src/features/auth/policy/verifier.ts
```diff
@@ -0,0 +1,123 @@
...
+  constructor(
+    private readonly store: KeyStore,
+    private readonly salt: string = env.SALT,
+    private readonly adminApiKey: string | undefined = env.ADMIN_API_KEY,
+    private readonly isCloud = Boolean(env.NEXT_PUBLIC_LANGFUSE_CLOUD_REGION),
+  ) {}
+
+  /** verify resolves the Authorization header to a credential presentation, or a 401. */
+  async verify(authHeader: string | undefined): Promise<VerifyResult> {
+    if (authHeader?.startsWith("Basic ")) {
```
marksalpeter: Lets make these prefixes constants.

web/src/features/auth/policy/verifier.ts
```diff
@@ -0,0 +1,123 @@
...
+  | (Success & ResolveContextParams)
+  | ErrorResult<UnauthorizedError>;
+
+/** KeyStore reads `ApiKey` rows by the two disjoint unique indexes and backfills the fast hash. */
+export type KeyStore = {
+  findByFastHash: (hash: string) => Promise<ApiKey | null>;
+  findByPublicKey: (publicKey: string) => Promise<ApiKey | null>;
+  verifySlow: (secretKey: string, apiKey: ApiKey) => Promise<boolean>;
+  backfillFastHash: (apiKey: ApiKey, hash: string) => Promise<void>;
+};
```
marksalpeter: Lets call this the ApiKeyRepository. Methods should return errors similar to VerifyResult.

web/src/features/auth/policy/types.ts
```diff
@@ -0,0 +1,108 @@
...
+  | {
+      kind: "apiKey";
+      apiKeyId: string;
+      userId: string | null;
+      organizations: PrincipalOrganization[];
+      boundResource?: Resource;
+    };
+
+/** Source describes where a policy came from: a role or an explicit grant. */
+export type Source = { kind: "role"; id: string } | { kind: "grant" };
```
marksalpeter: is source being used for anything? if not lets pull it from spec. if it is, what it is used for?

web/src/features/auth/policy/verifier.ts
```diff
@@ -0,0 +1,123 @@
...
+import { env } from "@/src/env.mjs";
+import { type ErrorResult, type Success } from "./types";
+import { type ResolveContextParams } from "./resolveContext";
+
+/** invalidCredentials is today's single 401 body for any unknown or malformed token. */
+const invalidCredentials =
+  "Invalid credentials. Confirm that you've configured the correct host.";
+
+/** Verifier authenticates a request credential into a resolvable presentation, dispatching Basic vs Bearer and never throwing. */
+export class Verifier {
```
marksalpeter: verifier is missing its redis backed cache with considerations for cache invalidation. must implement.

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,195 @@
+import { type ApiKey } from "@langfuse/shared/src/db";
```
marksalpeter: Lets call this file policyService and rename ContextResolver to PolicyService, resolveContext to getPolicies(GetPolicyParams): GetPolicyResults, etc.  This should go in `features/rbac/policyService.ts`

web/src/features/auth/policy/shadow.projects.ts
```diff
@@ -0,0 +1,86 @@
+import { type NextApiRequest } from "next";
```
marksalpeter: the verifyAuth code should be added directly into `createAuthedProjectAPIRoute`

web/src/features/auth/policy/enforceProjectAuth.ts
```diff
@@ -0,0 +1,114 @@
+import { type IncomingHttpHeaders } from "http";
```
marksalpeter, line 1: move to  `web/features/apiKey/enforceProjectAuth`.ts

web/src/features/auth/policy/enforce.ts
```diff
@@ -0,0 +1,4 @@
+/** headerValue normalizes a possibly-repeated header to its first value. */
```
marksalpeter: this could be moved to `enforceProjectAuth` for now and not exported if no one else uses it.

web/src/features/auth/policy/shadow.ts
```diff
@@ -0,0 +1,113 @@
+import { getCurrentSpan, recordIncrement } from "@langfuse/shared/src/server";
```
marksalpeter, line 1: Move to `feature/apiKey/shadow.ts`. We should start this with a comment including key datadog queries and what we could conclude from each. if theres just one, then thats fine as well. a ts query grouped by our decision matrix over time would also be good to include.

web/src/features/rbac/constants/organizationAccessRights.ts
```diff
@@ -12,6 +12,9 @@ export const organizationScopes = [
   "organizationMembers:CUD",
   "langfuseCloudBilling:CRUD",
   "auditLogs:read",
+
+  // Public-API action token; not granted to any UI role.
+  "projects:read",
```
marksalpeter: Change to "project:read" for consistency with the project scopes.

web/src/features/auth/policy/types.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/** BasePolicy carries the origin and effect every policy shares. */
+export type BasePolicy = {
+  source: Source;
+  effect: "allow" | "deny";
+};
+
+/** OrganizationSystemPolicy is a resource-less org-level policy. */
+export type OrganizationSystemPolicy = BasePolicy & {
+  kind: "organization";
```
marksalpeter, line 84: The following model will allow us to ship more features faster. This will need a grilling session to work out the implications of, which are numerous. But ultimately this will set us up for the next few iterations much better.

```typescript
type SystemRoles = "ADMIN" | "VIEWER" | ....;

type SystemPolicy = { 
  roleId: SystemRoles;
  effect: "ALLOW" | "DENY";
  actions: Action[] | Wildcard;
}

type OrganizationResourceId = `organization/${string}`;

type ProjectResourceId = `project/${string}`

type ResourceId = OrganizationResourceId | ProjectResourceId;

type Policy = SystemPolicy & {
  resources: ResourceId[] | Wildcard;
}
```

At resolution time, we should  expand org scoped api keys into the set of resources that include all of their projects, eg `["organization/o", "project/a", …, "project/z"]`. Project scoped keys recieve only their project id as a resource `["project/p"]`. All actions, wether they are org or project scoped, apply to that policy. Wether the caller has a project id or an org id, they will resolve the correct set of permissions.

Later, when we adopt this model for users, each role assignment will translate to a single policy with the resources populated based on a tenancy relationship to an organization or project.

Things that may have to change: actions that are keyed the same but legitimately guard different functionality (eg org audits vs project audits - unless the project audit is a subset of the org audit, it may be desirable to differentiate these actions).

web/src/features/rbac/constants/organizationAccessRights.ts
```diff
@@ -12,6 +12,9 @@ export const organizationScopes = [
   "organizationMembers:CUD",
   "langfuseCloudBilling:CRUD",
   "auditLogs:read",
```
marksalpeter: Should having the ability to read organization access logs grant permission for the user to read project level access logs? Open question.

web/src/features/auth/policy/types.ts
```diff
@@ -0,0 +1,108 @@
+import { type z } from "zod";
```
marksalpeter, line 1: These types should move to rbac/policy.ts

web/src/features/auth/policy/verifier.ts
```diff
@@ -0,0 +1,123 @@
+import crypto from "node:crypto";
```
marksalpeter: This should probably go in \`web/features/apiKey/verifier.ts\` and be called ApiKeyVerifier.verify(credentials:string):ApiKeyVerifierResults

web/src/features/auth/policy/authorize.ts
```diff
@@ -0,0 +1,55 @@
+import { ForbiddenError } from "@langfuse/shared";
```
marksalpeter, line 1: this should probably go in `features/rbac/authorize.ts`

web/src/features/auth/policy/identity.ts
```diff
@@ -0,0 +1,205 @@
+import { type IncomingHttpHeaders } from "http";
```
marksalpeter: lets move this to `features/apiKey/auth.ts`. Main authenticate method should be on class  ApiKeyAuth.auth(ApiKeyAuthParams):ApiKeyAuthResults.

web/src/features/rbac/constants/apiKeyAccessRights.ts
```diff
@@ -0,0 +1,71 @@
+import {
```
marksalpeter, line 1: this could be renamed to systemRoles.ts

web/src/features/auth/policy/types.ts
marksalpeter: Cut for time. We'll refactor in the next pass

web/src/features/auth/policy/types.ts
marksalpeter: refactor in a second pass.

web/src/pages/api/public/unstable/dashboard-widgets/[widgetId].ts
```diff
@@ -18,6 +18,7 @@ import {
 export default withUnstablePublicApiMiddlewares({
   GET: createUnstablePublicApiRoute({
     name: "Get Unstable Dashboard Widget",
+    action: null,
```
Comment (marksalpeter) [inline web/src/pages/api/public/unstable/dashboard-widgets/[widgetId].ts]: All endpoints (including the unstable endpoints) must be guarded with a corresponding action. If a suitable action doesn't exist, add one.

We should make action mandatory here.

web/src/features/auth/policy/resolveContext.ts
Steffen911: @marksalpeter We could also keep the cache in memory per container? Depends on how expensive the lookups are.

web/src/__tests__/server/in-app-agent-api-route-auth.servertest.ts
Steffen911: Why? What cases do we have for a non-defined action?

web/src/features/auth/policy/authorize.ts
```diff
@@ -0,0 +1,55 @@
...
+  resource: Resource,
+): Decision {
+  const matches = ctx.policies
+    .filter(hasResourceKind(resource))
+    .filter(hasAction(action))
+    .filter(hasResourceId(resource));
+
+  if (matches.some(hasEffect("deny"))) {
+    return forbidden();
+  }
```
Steffen911, line 26: Didn't we out-scope deny policies initially and worked under the assumption that we focus on `Allow` only?

web/src/features/auth/policy/verifier.ts
```diff
@@ -0,0 +1,123 @@
...
+    const byPublicKey = await this.store.findByPublicKey(token);
+    if (byPublicKey) {
+      return { success: true, authorization: "publicKey", apiKey: byPublicKey };
+    }
+    return unauthorized();
+  }
+
+  /** matchesAdminKey timing-safe compares the token to the self-host admin key. */
+  private matchesAdminKey(token: string): boolean {
+    if (this.isCloud || !this.adminApiKey) return false;
```
Steffen911: Why do we need the cloud check here? Aren't admin keys available on cloud and self-hosted?

web/src/features/rbac/constants/apiKeyAccessRights.ts
```diff
@@ -0,0 +1,71 @@
...
+  type SystemPolicy,
+} from "@/src/features/auth/policy/types";
+
+/** ApiKeyRole is the policy layer's role vocabulary for an API key, decoupled from the DB `ApiKeyScope` enum. */
+export type ApiKeyRole =
+  | "PROJECT"
+  | "ORGANIZATION"
+  | "SCORES_INGEST"
+  | "ADMIN"
+  | "VIEWER";
```
Steffen911: Are those the "new" roles we target or is this an attempt to map existing behaviours? What would viewer represent?

web/src/pages/api/public/feedback.ts
```diff
@@ -9,6 +9,7 @@ import {
 export default withMiddlewares({
   POST: createAuthedProjectAPIRoute({
     name: "Submit Feedback",
+    action: null,
```
Steffen911: What does this do?

web/src/pages/api/public/unstable/dashboards/[dashboardId].ts
```diff
@@ -18,6 +18,7 @@ import {
 export default withUnstablePublicApiMiddlewares({
   GET: createUnstablePublicApiRoute({
     name: "Get Unstable Dashboard",
+    action: null,
```
Comment (Steffen911) [inline web/src/pages/api/public/unstable/dashboards/[dashboardId].ts]: Can we set the type to non-null to force a type error here?

Steffen911 (commented):
This prepares bearer api keys with a new format, but doesn't implement them, right? So right now, the only bearer case remains the public key score ingest?

web/src/features/rbac/constants/apiKeyAccessRights.ts
Steffen911: If those are the new "presets" I would also like to add ingest only to the list.

web/public/generated/api/openapi.yml
marksalpeter: Reverted in 7be0df96b — the Bearer advertisement is removed from the public API spec. In shadow mode this isn't the live contract, and the implementation is backward-compatible, so the docs land as a post-release step.

fern/apis/server/definition/api.yml
marksalpeter: Reverted in 7be0df96b — the Fern source auth-schemes/Bearer change is rolled back and openapi.yml regenerated. Docs come after shadow + release.

.env.prod.example
marksalpeter: Done in 61c10d86a — renamed PUBLIC_API_AUTHZ_MIGRATION to API_AUTH_MIGRATION (one flag for all auth migration, authn + authz). Enum legacy|shadow|enforce and the legacy default are kept; all .env*.example are synced.

web/src/__tests__/server/in-app-agent-api-route-auth.servertest.ts
marksalpeter: Resolved in b95ec9d3b — action is now mandatory and non-nullable on the route, so the null path is gone entirely; this test passes a real action (project:read) instead of null. See the related decision on r3879971211.

web/src/features/auth/policy/resolveContext.ts
marksalpeter: Removed in d0da9fcb0 — the OrgEnrichment alias is gone; the code uses PrincipalOrganization directly. (Falls out of the pure-resolver refactor, which no longer holds the enrichment type.)

web/src/features/auth/policy/resolveContext.ts
marksalpeter: Done in d0da9fcb0 — the resolver no longer has a default org repo at all. Org fetch + enrichment moved up into the Authenticator, so the resolver is pure (receives a materialized PrincipalOrganization) and the throwing not-wired stub is deleted.

web/src/features/auth/policy/resolveContext.ts
marksalpeter: Landed in d6f74fe44 — a consolidated redis context cache in its own authz:context: namespace (distinct from the legacy api-key: cache so shadow mode can run both). See r3872125478 for the invalidation.

web/src/features/auth/policy/resolveContext.ts
marksalpeter: Done in d0da9fcb0 / 11125ba2b — OrganizationRepository.getByOrgId returns the raw Organization (with its projects relation) as Success | ErrorResult<LangfuseNotFoundError | InternalServerError>. Note the 404 class is LangfuseNotFoundError; there is no InternalImplementationError, so unexpected failures map to InternalServerError. "Org missing for a valid key" is treated as an invariant break (500), not a client 404.

web/src/features/auth/policy/verifier.ts
marksalpeter: Done in 11125ba2b — extracted basicPrefix and bearerPrefix module constants, used by both the startsWith check and the slice.

web/src/features/auth/policy/verifier.ts
marksalpeter: Done in d0da9fcb0 / 11125ba2b — renamed to ApiKeyRepository. Reads return { success: true; apiKey: ApiKey | null } | ErrorResult<InternalServerError> — a lookup miss stays apiKey: null (normal control flow; the verifier emits 401 only after all lookups miss), and only infra failures are errors. verifySlow returns { valid } | ErrorResult; backfillFastHash is best-effort.

web/src/features/auth/policy/verifier.ts
marksalpeter: Implemented in d6f74fe44 — one consolidated cache keyed by a hash of the presented credential (sha(secret+salt) / sha(publicKey+salt)), value is the materialized AuthorizationContext, namespace authz:context:. Invalidation reuses the existing subsystem: the 4 invalidateApiKeys functions now also delete both new-namespace keys per row, and invalidateCachedOrgApiKeys is inlined at the 3 project-create sites (onboarding after commit). Negative caching + fail-open included.

web/src/features/auth/policy/shadow.projects.ts
marksalpeter: Done in c2b5a5a73 — the redundant verifyAuth wrapper is collapsed so createAuthedProjectAPIRoute calls the seam directly. The shadow/diff/coverage machinery stays in its own module (shadow.ts), not inlined into the route wrapper.

web/src/features/auth/policy/enforce.ts
marksalpeter: Done in c2b5a5a73 — headerValue moved into enforcement.projects.ts as a private (unexported) helper and enforce.ts deleted. One caveat: it turned out to have two consumers (the Authenticator in features/apiKey/auth.ts too), so it lives as two byte-identical private copies rather than one shared module — a 2-line helper across two feature areas.

web/src/features/rbac/constants/organizationAccessRights.ts
marksalpeter: Resolved in c48604e46, but via a redesign rather than a literal rename. Projects are a single dataset, so there is one project:read action; resource binding decides which projects a principal sees (an org key binds to all its org's projects, a project key to the one). The org-list projects:read token is deleted and org keys are granted a bounded project-kind project:read. A literal projects:read -> project:read rename would have collided with the existing project-scope project:read.

web/src/features/rbac/constants/organizationAccessRights.ts
marksalpeter: Answered in c48604e46 — two actions, not a cascade. Org-only audit rows genuinely exist (AuditLog.orgId is required, projectId is nullable; org membership/billing/SSO events carry no project), so org and project audit logs are distinct datasets: orgAuditLogs:read and projectAuditLogs:read. Org admins end up holding both; reading org audit logs does not implicitly grant project audit-log read.

web/src/features/auth/policy/identity.ts
marksalpeter: Done in 11125ba2b — moved to web/src/features/apiKey/auth.ts. Named class Authenticator with auth(params): ApiKeyAuthResults rather than ApiKeyAuth, because it also handles the admin key and produces the authz context, so the name reflects more than API-key authn. features/auth/policy/ stays the pure PDP; features/apiKey/ holds the Authenticator, verifier, and repos.

web/src/pages/api/public/unstable/dashboard-widgets/[widgetId].ts
Comment (marksalpeter) [reply web/src/pages/api/public/unstable/dashboard-widgets/[widgetId].ts]: Done in b95ec9d3b — action is mandatory and non-nullable on every project route, including the unstable ones (dashboards/dashboard-widgets/evaluators/evaluation-rules all now carry read/CUD actions). Where no existing action fit, added feedback:create (public-API token, no UI role). Goal is the hard non-nullable action for this PR; we can relax later with cause.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+  constructor(
+    private readonly verifier: Verifier = buildVerifier(),
+    private readonly orgs: OrganizationRepository = buildOrgRepo(),
+    private readonly cache: ContextCache = buildContextCache(),
+  ) {}
+
+  /** auth runs the full pipeline read-through the consolidated context cache, returning a typed failure rather than throwing. */
+  async auth(params: ApiKeyAuthParams): Promise<ApiKeyAuthResults> {
+    const authHeader = headerValue(params.headers.authorization);
+    const cacheKey = this.verifier.cacheKey(authHeader);
```
marksalpeter: 1. Cache key generation should be owned by the cache, not by the verifier. 
2. How can the cache invalidate properly if projects are added and removed from the organization or if api keys are invalidated.
3. If the cache key generation fails, then the header is not valid and we should early return instead of continuing.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+
+    if (cacheKey) {
+      const cached = await this.cache.read(cacheKey);
+      if (cached) return cached;
+    }
+
+    const verified = await this.verifier.verify(authHeader);
+    if (!verified.success) {
+      if (cacheKey && verified.error instanceof UnauthorizedError) {
+        await this.cache.writeUnauthorized(cacheKey, verified.error);
```
marksalpeter: Why does the negative cache require a separate method. Just cache the ResolveContextResult

web/src/features/auth/policy/resolveContext.ts
```diff
@@ -0,0 +1,106 @@
...
+  type Policy,
+  type Principal,
+  type PrincipalOrganization,
+  type Resource,
+  type Success,
+  type SystemPolicy,
+} from "./types";
+
+/** resolveContext materializes an authenticated credential into its `AuthorizationContext`; pure — the caller pre-enriches the organization. */
+export function resolveContext(input: ResolveContextParams): Resolved {
```
marksalpeter: Its the resolvers responsibility to load the organization material required to materialize the context. Repo should come in on the context. ResolveContextResult should include ErrorResult to account for potential repo failures.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+      }
+      return verified;
+    }
+
+    const gate = gateKeyKind(verified, params);
```
marksalpeter: This key gating functionality needs to be reworked to apply to the final cached payload. This may mean caching an api key partial.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+    if (gate) return gate;
+
+    if (verified.authorization === "adminKey") {
+      return resolveContext({ authorization: "adminKey" });
+    }
```
marksalpeter: Remove. This logic belongs in the resolver.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+
+    const enriched = await this.enrichOrg(verified.apiKey);
```
marksalpeter: Remove, this logic belongs in the resolver. enrichOrg is a very bad name. `OrganizationRepository.getOrganization(orgId:string):Promise<Organization>` returning the prisma organization and its projects is the appropriate seam. Transformation into a princiapl organization is the responsibility of the resolver.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+    if (!enriched.success) return enriched;
+
+    const resolved = resolveContext({
+      authorization: verified.authorization,
+      apiKey: verified.apiKey,
+      organization: enriched.organization,
```
marksalpeter: Again, terrible design and no.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+    });
+
+    // In-app-agent keys ride a route-specific gate; caching their context would
+    // let a hit skip the gate on a route that disallows them.
+    if (cacheKey && !verified.apiKey.isInAppAgentKey) {
+      await this.cache.writeContext(cacheKey, resolved.context);
```
marksalpeter: The gating should be applied after the context is written.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+        });
+      } catch (error) {
+        logger.error("authz api key fast-hash backfill failed", error);
+      }
+    },
+  };
+}
+
+/** prismaOrganizationRepository reads an org with its live projects, returning a miss or infra failure as a value. */
+function prismaOrganizationRepository(
```
marksalpeter: This must be a class. Dependencies are injected via the constructor. It lives in its own file.

web/src/features/apiKey/auth.ts
```diff
@@ -0,0 +1,362 @@
...
+  };
+}
+
+/** cacheEnabled is the shared on/off switch for the context cache, reusing the legacy api-key cache flag. */
+function cacheEnabled(redis: Redis | Cluster | null): redis is Redis | Cluster {
+  return Boolean(redis) && env.LANGFUSE_CACHE_API_KEY_ENABLED === "true";
+}
+
+/** redisContextCache stores the materialized `AuthorizationContext` (and negative 401s) under the `authz:context:` namespace, failing open on any redis error. */
+export function redisContextCache(redis: Redis | Cluster | null): ContextCache {
```
marksalpeter: This must live in its own file, its dependencies must be injected via a constructor.

web/src/features/auth/policy/enforcement.projects.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import {
+  type ForbiddenError,
+  type InternalServerError,
+  InvalidRequestError,
+  type UnauthorizedError,
+} from "@langfuse/shared";
+
+import { authorize } from "./authorize";
+import { defaultAuthenticator } from "@/src/features/apiKey/auth";
```
marksalpeter: lets just call this the authenticator, lower case.

web/src/features/rbac/constants/apiKeyAccessRights.ts
marksalpeter: Fair point. Lets add the ingest only for the complete set (\["traces:create", "scores:create"\]).

web/src/pages/api/public/feedback.ts
marksalpeter: AI slop. will be removed

web/src/pages/api/public/unstable/dashboards/[dashboardId].ts
Comment (marksalpeter) [reply web/src/pages/api/public/unstable/dashboards/[dashboardId].ts]: Agree

web/src/features/auth/policy/verifier.ts
marksalpeter: Agree

web/src/features/rbac/constants/apiKeyAccessRights.ts
marksalpeter: Shipping only the reachable roles (PROJECT/ORGANIZATION/SCORES_INGEST/ADMIN) in 4708266fb; VIEWER was unreachable dead code and is removed. Ingest-only + the preset menu are deferred to the role-assignment pass — shadow mode can only diff what a real key actually resolves to. Supersedes my earlier note about adding ingest-only now.

marksalpeter: ^ ai. decided to pull viewer from this pr. well add it back along with scores injest when expanding the roles

marksalpeter:
@claude review

web/src/features/apiKey/helpers/parseAuthorizationHeader.ts
```diff
@@ -0,0 +1,35 @@
...
+    return { kind: "bearer", token: header.slice(bearerPrefix.length) };
+  }
+  return { kind: "malformed" };
+}
+
+/** decodeBasic decodes a base64 `public:secret` payload, or undefined if malformed. */
+function decodeBasic(
+  encoded: string,
+): { publicKey: string; secretKey: string } | undefined {
+  const [publicKey, secretKey] = atob(encoded).split(":");
```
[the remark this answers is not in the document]
marksalpeter: Fixed in 83b1a1f6930b311248c7a72fe7cec2c92c64893f: decodeBasic now wraps atob in try/catch and returns malformed for any base64 decode failure, not just a missing colon, so a malformed Basic header no longer throws out of parseAuthorizationHeader.

web/src/features/auth/policy/enforceProjectAuth.ts
```diff
@@ -0,0 +1,113 @@
...
+  const header = headerValue(headers[projectIdHeader]);
+  if (header && boundProjectId && header !== boundProjectId) {
+    return {
+      success: false,
+      error: new InvalidRequestError(
+        `${projectIdHeader} disagrees with the API key's project`,
+      ),
+    };
+  }
+  const projectId = header ?? boundProjectId;
```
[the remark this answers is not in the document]
marksalpeter: Fixed in 39a31aa63ee676ea26c902222e47491feb491ed8: normalize a blank `x-langfuse-project-id` header to `undefined` at the read site so an empty value falls back to the bound project instead of 400ing.

web/src/features/auth/policy/authorize.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/** hasResourceKind matches a policy of the kind that governs the checked resource. */
+const hasResourceKind = (resource: Resource) => (p: Policy) =>
+  p.kind === ("projectId" in resource ? "project" : "organization");
+
+/** hasResourceId matches a policy whose resources cover the checked resource's id, by wildcard or listing. */
+const hasResourceId = (resource: Resource) => (p: Policy) =>
+  p.resources === wildcard ||
+  p.resources.includes(
+    "projectId" in resource ? resource.projectId : resource.orgId,
```
maxdeichmann, line 41: what do you think about this ternary operator? Else orgid may be dangerous long term.

marksalpeter: Fair point. If we ever add a resource that is not a project or an organization it could bite us. Could be nicer if the policy resource ids carried the kind. I will change this in the next pass.

PR #16670, opened 2026-08-27
feat(v4): attribute deprecated API callers

hassiebp:
🟡 **Live preview: [pr-16670.preview.langfuse.com](https://pr-16670.preview.langfuse.com)**

## Summary

- attribute deprecated public API usage to bounded User-Agent and canonical Langfuse SDK name/version data
- preserve endpoint totals while grouping callers and showing exact Python/JavaScript migration methods, including resource, dataset, experiment, and time filters
- add canonical SDK name/version attributes to all incoming request spans and retain ingestion dashboard aliases
- bound caller cardinality inside ClickHouse and defer the 20-caller display cap until the complete trailing window is merged

## Privacy and compatibility

- does not add API-key identifiers to ClickHouse tags, Redis, tRPC, UI, or telemetry
- sanitizes and bounds caller metadata without splitting Unicode code points, with overflow retained as one counted Other callers row
- keeps pre-change Redis blobs readable as unknown callers

## Verification

- shared attribution/query-tag/rollup tests: 36 passed
- worker query-log pipeline tests: 10 passed
- migration client tests: 52 passed
- v4 transition router tests: 36 passed
- Storybook migration-section tests: 13 passed in Chromium
- repository lint: 7 successful, 7 total
- repository typecheck: 8 successful, 8 total
- visually reviewed Python SDK, JavaScript SDK, coding-agent, and unknown-caller states in Storybook

## Reviewer focus

Please doubt the ClickHouse topK(100) caller-candidate bound and the endpoint-specific SDK replacement strings first; those are the deliberate accuracy/cardinality tradeoffs.


hassiebp:
Reviewer doubt list: please scrutinize the system.query_log grouping/cardinality assumptions and the generated-client method matrix, especially the deliberate distinction between legacy Scores v1 REST calls and SDK-backed Scores v2 calls. Also check whether retaining the two ingestion dashboard aliases while removing arbitrary x-langfuse span attributes preserves the dashboards you rely on.

packages/shared/src/server/ingestion/ingestionAttribution.ts
```diff
@@ -44,6 +44,15 @@ export type IngestionSdkAttributionStatus =
...
+export type PublicApiCallerAttribution = {
+  sdkName?: IngestionSdkCanonicalName;
+  sdkVersion?: string;
+  userAgent?: string;
+};
+
+export const PUBLIC_API_SDK_VERSION_MAX_LENGTH = 64;
+export const PUBLIC_API_USER_AGENT_MAX_LENGTH = 256;
+
 const getHeaderValue = (
```
hassiebp, line 56: can we make getHeaderValue robust agains keys that have some uppercase chars? e.g. X-Langfuse-Sdk-Version should be matched as well

hassiebp: Good catch — mixed-case header maps could previously miss SDK attribution. Fixed in cfcb318f4 by keeping the exact-key fast path and adding a case-insensitive fallback that also normalizes hyphens and underscores. Covered by the shared attribution regression test (`Tests 25 passed (25)` across attribution and bootstrap SDK suites).

PR #16678, opened 2026-08-27
fix(blobstorage): suppress sub-second remainder chunk to prevent silent key collision (v3 backport of #16592)

Steffen911:
**TL;DR** — v3 backport of langfuse/langfuse#16592. A caught-up scheduled blob export could emit a full-interval chunk followed by a sub-second remainder chunk that landed on the *same* second-precision object key, silently overwriting the full window's data files and manifest. Cherry-pick applied to v3 with no conflicts and no adaptation.

Source PR: langfuse/langfuse#16592 (`9b955df9c289eb63601109128eb62c9889187067`)

## The bug, and why v3 is affected identically

Object keys and manifest keys are built from `formatBlobExportTimestamp`, which on v3 is:

```ts
// worker/src/features/blobstorage/manifest.ts:39
export const formatBlobExportTimestamp = (date: Date): string =>
  date.toISOString().replace(/:/g, "-").substring(0, 19);
```

`substring(0, 19)` truncates to whole seconds — no milliseconds. So when a catch-up run's interval-capped `maxTimestamp` ends a few hundred milliseconds short of the frontier, the run re-enqueues immediately, and the tiny remainder chunk's key collides with the chunk that just preceded it. On buckets without versioning that is silent, permanent data loss (~7–8 windows/day on an hourly integration).

The fix treats a remainder below `BLOB_STORAGE_REMAINDER_COALESCE_MS` (1000 ms — the key granularity) as caught up, so the remainder is never emitted. The deferred sub-second tail is picked up by the next scheduled run, whose `minTimestamp = lastSyncAt` already covers it. No change to the object-key format or the queue payload contract.

## Verification

Run against a v3 worktree with a **fresh** ClickHouse volume (v3's own 37 migrations, not a main-contaminated one) and a dedicated `langfuse_v3_test` Postgres at 424/424 applied migrations.

- **Patch identity** — `git diff origin/v3..HEAD` is line-for-line identical to `git show 9b955df9c2` (hunk headers stripped). The surrounding file has diverged substantially between v3 and main (462 lines), but every line this commit touches is byte-identical, and the `caughtUp` / `uncappedMaxTimestamp` / `nextSyncAt` logic it modifies is unchanged on v3.
- **Full suite on the branch** — `pnpm --filter worker run test blobStorageIntegrationProcessing` → `Tests  32 passed (32)`. Worth noting explicitly: the change alters `caughtUp` semantics, and none of the 31 pre-existing tests (including `should schedule normally when caught up` and the other chunked-historic-export cases) regress.
- **Revert-check — the test genuinely catches v3's bug.** Reverting just the `caughtUp` logic to v3's original two lines, while leaving the new constant exported so the test compiles unchanged, fails the new test decisively:

  ```
  FAIL  should coalesce a sub-second remainder instead of emitting a colliding chunk
  AssertionError: expected 71 to be greater than 1800000
   ❯ expect(updatedIntegration.nextSyncAt.getTime() - now.getTime())
       .toBeGreaterThan(frequencyIntervalMs / 2);
  ```

  On v3's logic `nextSyncAt` is set **71 ms** past `now` — the immediate catch-up re-enqueue that emits the colliding chunk. With the fix it is one full hour out. The prod file was restored afterwards and `git status --porcelain` is empty.
- `pnpm run format:check` → `All matched files use Prettier code style!`
- `pnpm run typecheck` → `Tasks: 7 successful, 7 total`

This backport prevents sub-second catch-up remainder chunks from overwriting full export windows that share the same second-precision object key.

- Adds a 1000 ms coalescing threshold aligned with blob-export key granularity.
- Treats smaller remainders as caught up while retaining the capped timestamp as the next export position.
- Adds regression coverage for scheduling and retained-boundary behavior.
</details>

The PR appears safe to merge; the deferred tail remains reachable through the next run’s inclusive lower boundary and receives a non-colliding export key.

The new threshold exactly matches the object-key timestamp granularity, all supported export frequencies are substantially larger, and persisted synchronization state ensures deferred records are exported by the next scheduled run.
</details>

```mermaid
flowchart TD
  A[Calculate capped export max timestamp] --> B[Export window and persist manifest]
  B --> C{Remainder below 1000 ms?}
  C -->|No| D[Immediately enqueue catch-up run]
  C -->|Yes| E[Persist capped max as lastSyncAt]
  E --> F[Schedule next regular run]
  F --> G[Next run starts inclusively at lastSyncAt]
  G --> H[Export deferred sub-second tail under a later key]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(blobstorage): suppress sub-second re..."](https://github.com/langfuse/langfuse/commit/4d63d75a787cf7d622bd434861bbdaf21be24072) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57464602)</sub>

**Context used:**

- Knowledge Base — [Media and blob processing](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/media-and-blob-processing.md)

### DIFF
--- a/worker/src/__tests__/blobStorageIntegrationProcessing.test.ts
+++ b/worker/src/__tests__/blobStorageIntegrationProcessing.test.ts
@@ -60,6 +60,7 @@ import { Job } from "bullmq";
 import {
   handleBlobStorageIntegrationProjectJob,
   BLOB_STORAGE_LAG_BUFFER_MS,
+  BLOB_STORAGE_REMAINDER_COALESCE_MS,
 } from "../features/blobstorage/handleBlobStorageIntegrationProjectJob";
 import {
   BlobStorageIntegrationType,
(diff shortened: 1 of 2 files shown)


Steffen911:
What I'd doubt here:

1. **The test proves scheduling changed, not that the collision is gone.** It asserts on `nextSyncAt`/`lastSyncAt` in Postgres — that a near-frontier run is treated as caught up instead of re-enqueuing. That the two chunks *would* have shared an object key is argued from `formatBlobExportTimestamp`'s `substring(0, 19)`, not asserted against actual S3 keys. A test that listed bucket keys across two runs would be the direct proof; this is one inference removed.
2. **Nothing proves the deferred tail is exported rather than dropped.** The fix's safety rests on "the next scheduled run's `minTimestamp = lastSyncAt` already covers it." I read that path and it holds, but I did not write a two-run test showing the sub-second tail actually lands in a later export. That's the residual risk worth the most scrutiny: if that reasoning is wrong, the change converts a silent overwrite into a silent gap.
3. **The 1000 ms constant sits exactly on the boundary.** `remainderMs < BLOB_STORAGE_REMAINDER_COALESCE_MS` means a remainder of exactly 1000 ms is *not* coalesced — correct, since chunks 1000 ms apart truncate to distinct second-precision keys — but the off-by-one is reasoned, not exercised by a boundary test.
4. **MinIO was a shared dev volume.** ClickHouse I rebuilt from a fresh volume with v3's own 37 migrations, but the S3 layer reuses a dev bucket other worktrees have written to. Irrelevant for these assertions (none read objects back), but I'd rather say so than imply a pristine environment.

PR #16687, opened 2026-08-27
feat(query-ast): clean Kysely base for the ClickHouse query AST

niklassemmler:
🟡 **Live preview: [pr-16687 app preview](https://pr-16687.preview.langfuse.com)**

## What does this PR do?

Establishes a clean continuation base for the Kysely-backed ClickHouse
query-AST spike, combining the best of two prior attempts and closing the
security gap flagged across all of the earlier PR's automated reviews.

It supersedes the earlier Kysely-arm PR (now closed): the library
evaluation settled on Kysely, so the competing hand-rolled and alternative
library arms are dropped. From the hand-rolled arm, only its one genuinely
valuable idea — a typed semantic-model registry — is carried forward.

**Changes on top of the Kysely arm:**

- **Tenancy soundness fix.** The injection pass treated any `project_id = …`
  predicate as already scoping a relation without checking the bound value
  or the qualified table — a query carrying `project_id = <otherProject>`
  compiled with no corrective scope, and one unqualified `project_id` was
  treated as covering every tenanted table in a JOIN (sibling-relation
  leak). `predicateCovers` now requires the value to equal the execution
  context's project, and an unqualified column only counts when a single
  tenanted relation is in scope. Adds two regression tests.
- **One table registry as the source of truth.** A single `defineTable`
  declaration per relation now derives the Kysely row types, the runtime
  column-type map the aggregate type-check pass consults, and the tenanted
  table set — replacing three hand-maintained tables that drift. This ports
  the typed-schema idea from the hand-rolled arm onto Kysely.
- Documents the compiler's coupling to Kysely internals (`ArrayIndexNode`
  handling) as a known fragility, and removes a no-op tenancy guard branch.

Physical tuning metadata (partition / sort / dedup keys) is deliberately
out of scope here and will land when dedup lowering consumes it.

This is an early, non-merge continuation base for ongoing work, not a
finished feature.

## Type of change

- [x] Refactor (restructures existing code without changing behavior)

## Mandatory Tasks

## Checklist

- Added tests that prove the tenancy fix is effective (`kysely/nodes.test.ts`) and lock the registry contract (`kysely/schema.test.ts`).
- `pnpm turbo run lint` — 7 successful; `typecheck` (shared/web/worker) — 6 successful; `src/server/query-ast` tests — 43 passed. `environments.ts` golden parity unchanged.

This PR introduces a Kysely-backed ClickHouse query AST, migrates environment reads to it, and adds schema-derived runtime metadata and mandatory tenancy injection.
- Adds a compile-only ClickHouse dialect with custom ARRAY JOIN, LIMIT BY, metadata access, and virtual-view nodes.
- Derives Kysely row types, aggregate validation metadata, and tenanted-table membership from one table registry.
- Replaces handwritten environment SQL with compiled AST queries and adds parity and tenancy tests.
- The tenancy coverage check still has an alias-collision case that can leave a joined physical relation unscoped.
</details>

The PR should not be merged until the tenancy coverage check distinguishes aliases from physical table names and reliably scopes every joined relation.

A qualifier can simultaneously match one relation's alias and another relation's physical name, causing the compiler to omit the second relation's project constraint and admit cross-project rows into a join.

**Files Needing Attention:** packages/shared/src/server/query-ast/kysely/tenancy.ts
</details>

The tenancy pass can incorrectly treat one qualified project predicate as covering two relations when one relation's alias equals another relation's physical name. This can omit the joined relation's project constraint and allow cross-project rows into the join. **How this was verified:** Tracing the colliding qualifier through `isProjectIdColumn` shows that it matches the first relation's alias and the second relation's physical name, causing the join-injection guard to skip the second scope.
</details>

```mermaid
flowchart TD
  Q[Kysely query builder] --> V[Virtual-view and extension plugins]
  V --> T[Tenancy injection transformer]
  T --> R[Discover physical FROM and JOIN relations]
  R --> P{Predicate uniquely scopes each relation?}
  P -->|No| I[Inject project_id predicate]
  P -->|Yes| S[Skip injection]
  I --> C[Type validation and ClickHouse compilation]
  S --> C
  C --> E[queryClickhouse execution]
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/query-ast/kysely/tenancy.ts:290
**Alias collision bypasses tenancy**

When one tenanted relation's alias equals another relation's physical table name, `predicateCovers` counts the first relation's qualified `project_id` predicate as covering both. The joined relation then receives no project constraint, allowing cross-project rows to participate in the join.

**How this was verified:** A colliding qualifier matches the first relation's alias and the second relation's physical name, causing the join-injection guard to skip the second scope.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(query-ast): re-export formatSql afte..."](https://github.com/langfuse/langfuse/commit/933e7e43c895cc9e6be490001abece48f7145932) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57499935)</sub>

**Context used:**

- Knowledge Base — [Identity, access, and tenancy](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/identity-access-and-tenancy.md)

### CODE REVIEW

File: packages/shared/src/server/query-ast/kysely/compiler.ts [hunk 1]
```diff
@@ -0,0 +1,293 @@
...
+  if (typeof value === "bigint") return "Int64";
+  return "String";
+}
+
+/**
+ * ClickHouse SQL compiler. Identifier quoting is omitted so `clickhouse format`
+ * matches the golden snapshots (unquoted identifiers). Parameters are named
+ * `{pN:Type}` binds, interned by (type, value) so UNION branches that reuse
+ * the same project_id collapse to one placeholder — matching today's named
+ * `{projectId:String}` style after the golden harness's param normalizer.
```
niklassemmler: please provide examples to ease understanding

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * ARRAY JOIN and LIMIT BY are read off extra fields on the select node; they
+ * are real {@link ArrayJoinNode}/{@link LimitByNode}s, not raw SQL strings.
```
niklassemmler: why does this need to be stated?

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+  namedParameters: Record<string, unknown> = {};
+  private intern = new Map<string, string>();
+
+  constructor() {
+    super();
+    // FRAGILE: `ArrayIndexNode` is not one of Kysely's closed `OperationNode`
+    // kinds, so the default visitor dispatch cannot reach it. We wrap the
+    // private `visitNode` and push/pop the private `nodeStack` to route it to
+    // `visitArrayIndex`. This couples to Kysely internals and must be
+    // re-verified on any Kysely upgrade (pinned at 0.28.17).
```
niklassemmler: What happens if we upgrade without noticing this? After all a comment in code is not a great place to identify restriction to the upgrade path. Will we fail silently (bad) or produce incorrect SQL (very bad).

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+    const parentVisit = this.visitNode.bind(this);
+    (
+      this as unknown as { visitNode: (node: OperationNode) => void }
```
niklassemmler: AFAIK this is an anti pattern. At least leave a comment to justify it.

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+    ).visitNode = (node: OperationNode) => {
+      if (ArrayIndexNode.is(node)) {
+        (this as unknown as { nodeStack: OperationNode[] }).nodeStack.push(
```
niklassemmler: see above

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+    return super.compileQuery(node, queryId);
+  }
+
+  protected override getLeftIdentifierWrapper(): string {
+    return "";
+  }
+
+  protected override getRightIdentifierWrapper(): string {
+    return "";
+  }
```
niklassemmler, line 117: leave a comment on why we override this

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  protected override appendValue(parameter: unknown): void {
+    this.append(this.bindValue(parameter));
+  }
+
+  /**
+   * ClickHouse `IN` takes a single `Array(T)` bind, not a parenthesized value
+   * list. Emitting `({p:Array(String)})` is what the golden scores query uses.
```
niklassemmler: The reference to the "golden score query" looks obscure here

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+    if (values.every((v) => v !== undefined)) {
+      this.append("(");
+      this.append(this.bindValue(values, inferClickHouseType(values)));
+      this.append(")");
+      return;
+    }
+    super.visitValueList(node);
+  }
+
+  protected override visitSelectQuery(node: SelectQueryNode): void {
```
niklassemmler, line 162: AFAIK this is mostly a copy paste of the parent libraries method. Can you please explain state the changes as a comment so that a reader doesn't have to search?

packages/shared/src/server/query-ast/kysely/compiler.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  private visitLimitBy(node: LimitByNode): void {
+    this.append("limit ");
+    this.visitNode(node.count);
+    this.append(" by ");
+    this.compileList(node.columns);
+  }
+
+  private bindValue(value: unknown, type?: string): string {
```
niklassemmler, line 305: I worry that by inferring we don't exactly match the clickhouse types. In work on another database, we often had types of different length. Could that become an issue here?

packages/shared/src/server/query-ast/kysely/composition.test.ts
```diff
@@ -0,0 +1,120 @@
...
+          .select("environment")
+          .$call(costArrayJoin()),
+      )
+      .selectFrom("aj")
+      .select("environment");
+
+    const { sql } = compileClickhouseQuery(qb, ctx);
+    // The outer query has no ARRAY JOIN, so its presence proves the clause
+    // survived being embedded as a CTE body rather than being dropped.
+    expect(sql).toMatch(/array join/i);
```
niklassemmler: shouldn't we check that the array join is only in the with not the outer query?

packages/shared/src/server/query-ast/kysely/extensions.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      .selectFrom("events_core as e")
+      .select((_eb) => [metadataValue("e", "a").as("a")])
+      .where((eb) => eb(metadataValue("e", "a"), ">", 2));
+
+    const node = qb
+      .withPlugin(new TenancyInjectionPlugin(ctx))
+      .toOperationNode();
+    const json = JSON.stringify(node);
+    expect(json).toContain('"kind":"ArrayIndexNode"');
+    expect(json).toContain('"func":"indexOf"');
```
niklassemmler, line 55: it would help here if you could show more of the plan. It's difficult for me to judge this based on the excerpts. Same for the other tests in this file. At least justify as response to this comment why this is necessary.

packages/shared/src/server/query-ast/kysely/extensions.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    expect(json).not.toMatch(/"kind":"RawNode"/);
+
+    const selections = (
+      node as { selections?: { selection?: { kind: string } }[] }
+    ).selections;
+    expect(selections?.[0]?.selection?.kind).toBe("AliasNode");
```
niklassemmler, line 63: really difficult to understand what this does

packages/shared/src/server/query-ast/kysely/extensions.ts
```diff
@@ -0,0 +1,199 @@
+import {
```
niklassemmler, line 17: This file would benefit from a comment on what it does

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a worked example to the class JSDoc: a UNION whose two branches both filter `project_id = 'p'` collapses to a single `{p1:String}` reused in both branches, instead of `{p1:String}`/`{p2:String}` bound to the same value. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Reworded to say *why*: ARRAY JOIN / LIMIT BY have no Kysely-native node, so a reader might reasonably assume they're spliced-in raw SQL. Stating they're real `ArrayJoinNode`/`LimitByNode` operation nodes makes clear they go through the normal visitor — identifier-escaped and parameter-bound, i.e. injection-safe — not string concatenation. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Good question — it fails loud, not silent. `kysely` is pinned to an exact version (`0.28.17`, no caret), so a routine `pnpm install` can't move it. On a deliberate bump that changed these internals, the array-index path is exercised by SQL-asserting compile tests — `extensions.test.ts` asserts `metadata_values[indexof(...)]` and `nodes.test.ts` asserts the `ArrayIndexNode` SQL — plus the golden-SQL harness. So broken routing changes the emitted SQL and breaks CI rather than shipping wrong SQL. I recorded this in the comment.

If you'd prefer a hard belt-and-suspenders, I can add a constructor-time assertion that the private `visitNode` / `nodeStack` members still exist and throw on a version mismatch. Happy to add it if you want that guarantee independent of test coverage.

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Agreed it's an anti-pattern. Expanded the comment to justify it: Kysely exposes no public hook to register a new node kind, so the only seam is patching the private `visitNode` + `nodeStack`; the `as unknown as` casts exist solely to reach those private members, and the alternative is forking the compiler. It's now called out explicitly as a deliberate, contained trade. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Same fix as the comment above (line 77): the expanded justification comment now covers both the `visitNode` patch and the `nodeStack` push/pop, explaining why the private-member casts are necessary and why forking is the only alternative. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a comment at the override: identifiers are emitted unquoted so raw compiler output stays byte-comparable with `clickhouse format` golden snapshots (which normalize to unquoted names). Safe here because column/table names in this codebase are plain identifiers. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Reworded to drop the obscure reference — it now explains the semantics directly: ClickHouse `IN` takes a single `Array(T)` bind, not a parenthesized list of scalar binds, so `col IN (1, 2, 3)` compiles to `col IN ({p:Array(Int64)})`. Noted that this is also the shape the existing production scores list query relies on. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a comment stating the delta vs the parent: it's Kysely's `DefaultQueryCompiler.visitSelectQuery` (kept in sync with the pinned 0.28.17) copied verbatim, with exactly two ClickHouse-only insertions — the ARRAY JOIN block (after JOINs, before WHERE) and the LIMIT BY block (before LIMIT), both reading extra fields off `ClickHouseSelectQueryNode`. Overridden wholesale rather than via `super` because the parent emits clauses in a fixed order with no hook to inject between JOIN and WHERE. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Real concern, and yes it can matter for one case. The inferred type only types the bind *placeholder*; ClickHouse coerces it to the compared column's type, and we infer the widest canonical types (Int64 / Float64 / String / DateTime64(3)), so the widening cases we emit are safe — no truncation like the FixedString-length mismatches you're thinking of.

The one genuine risk is precision: an inferred `Float64` compared against a `Decimal` column can round. That's exactly what the explicit `type` parameter on `bindValue` is for — callers that know the column type pass it and bypass inference. I added a comment noting this.

We don't currently bind against Decimal columns on this path, but if you'd like I can open a follow-up to audit Decimal columns specifically before we lean on inference more widely.

packages/shared/src/server/query-ast/kysely/composition.test.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Good catch — the old assertion only checked that ARRAY JOIN was present, so it would have passed even if the clause leaked into the outer query. Strengthened it: it now splits on the outer query (the `) select ... from aj` that closes the CTE) and asserts ARRAY JOIN appears in the CTE body and is absent from the outer query. (`c768c6c`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: Please remove this phrase. This is too much advance thinking for a comment. Just state what is done.

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: Can you improve the format of the example and make it a separate paragraph?

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: test coverage is fine

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: That's a horrible way to do it. Nobody will read the comment that is just attached to the long paragraph comment. Add an explicit comment to the first occurrence of this anti pattern

packages/shared/src/server/query-ast/kysely/extensions.test.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Fair — added a comment block above this test explaining the strategy, and here's the justification. These `metadata[key]` tests are deliberately two-layer:

- **This test asserts the AST shape** (via the serialized tree): `metadata[key]` must *lower* to real operation nodes — an `ArrayIndexNode` subscript wrapping an `indexOf(...)` `FunctionNode` — and must **not** contain a `RawNode`. That's the property that matters for safety: real nodes are identifier-escaped and the key is a bound parameter, whereas a `RawNode` would be spliced-in SQL. A SQL-only assertion can't distinguish "built from real nodes" from "built by string concatenation that happens to render the same today", so this guards the lowering itself.
- **The sibling tests assert the emitted SQL** (`metadata_values[indexof(e.metadata_names...`, bound key param, no inline `'a'` literal) — i.e. that the compiler renders that tree correctly, in SELECT / WHERE / HAVING positions.

Splitting them keeps "builder produced the right tree" independent from "compiler rendered it right". (`3c1e9de`)

packages/shared/src/server/query-ast/kysely/extensions.test.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added an inline comment. This line navigates the built AST: because the selection is `metadataValue("e", "a").as("a")`, the `.as("a")` wraps the expression in an `AliasNode`. So the first selection's inner `.selection` node must be an `AliasNode`, not the bare `ArrayIndexNode` — the assertion proves the alias wrapping survived the lowering rather than being dropped or flattened. (`3c1e9de`)

packages/shared/src/server/query-ast/kysely/extensions.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a file header. It explains the module supplies ClickHouse constructs Kysely has no native node for, in three forms: node-attaching plugins (`ArrayJoinPlugin` / `LimitByPlugin`), type-extending `$call` helpers (`arrayJoin` / `limitBy`) that also widen the builder's output row type, and parameter-bound expression helpers (`mapKeys` / `mapValues` / `metadataValue`) that build map/array calls and `metadata[key]` subscripts as real nodes rather than raw SQL. (`3c1e9de`)

packages/shared/src/server/query-ast/kysely/extensions.ts
```diff
@@ -0,0 +1,199 @@
...
+function columnRef(name: string): OperationNode {
+  const parts = name.split(".");
+  if (parts.length === 2) {
+    return ReferenceNode.create(
+      ColumnNode.create(parts[1]),
+      TableNode.create(parts[0]),
+    );
+  }
+  return ColumnNode.create(name);
+}
```
niklassemmler, line 140: I am curious. Does Kysely not already have this?

niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Partly. Kysely *does* parse `"table.column"` — but only at the **expression-builder** layer, e.g. `eb.ref("t.c")` or a string reference passed to `.select()/.where()`. These plugins and helpers operate one level down, building raw `OperationNode`s inside `transformQuery` / expression wrappers where there's no `ExpressionBuilder` in scope. Kysely does not export its internal string-reference parser (I checked: `parseStringReference` isn't on the package's public surface; only the node primitives `ReferenceNode` / `ColumnNode` / `TableNode` are). So `columnRef` splits the dotted name and assembles those primitives itself.

It's a small, contained helper rather than a reimplementation of anything public. Added a comment saying exactly this. (`9cc4ce0`)

packages/shared/src/server/query-ast/kysely/extensions.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Lower `metadata[key]` to a traced array-subscript + `indexOf` node.
+ * The key is a bound `ValueNode`, not a SQL literal.
+ */
+export function metadataValue(
+  tableAlias: string,
+  key: string,
+): ExpressionWrapper<ClickHouseDatabase, "events_core", string | number> {
+  return new ExpressionWrapper(
+    ArrayIndexNode.create(
+      columnRef(`${tableAlias}.metadata_values`),
```
niklassemmler: Can we not create Table and Column nodes directly? The step via a string looks arbitrary.

packages/shared/src/server/query-ast/kysely/extensions.ts
```diff
@@ -0,0 +1,199 @@
...
+      FunctionNode.create("indexOf", [
+        columnRef(`${tableAlias}.metadata_names`),
+        ValueNode.create(key),
+      ]),
+    ) as unknown as OperationNode,
+  );
+}
+
+type ElementOf<E> =
```
niklassemmler, line 184: A comment to explain this would be nice

packages/shared/src/server/query-ast/kysely/extensions.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

You're right — here it was arbitrary. `metadataValue` already had the table alias and the column names as separate values, then concatenated them into `"alias.metadata_values"` just so `columnRef` could split them apart again. Removed the round-trip: added `qualifiedColumn(table, column)` that builds the `ReferenceNode` directly from the parts, and `metadataValue` now calls it straight.

`columnRef` stays only for the genuinely-string case — the `limitBy` column list, where a possibly-dotted column arrives as a single string from the caller — and it now delegates to `qualifiedColumn` for the dotted branch. (`3fc4059`)

packages/shared/src/server/query-ast/kysely/extensions.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a comment. `ElementOf<E>` extracts the element type `T` from an array-typed `Expression<T[]>`. `arrayJoin` uses it so each produced alias carries the array's *element* type in the builder's output row — ARRAY JOIN yields one row per element, so a `mapValues(...)` of type `V[]` should surface the alias as `V`, not `V[]`. It resolves to `never` for a non-array expression. (`a1c0fa6`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Done — the example is now its own paragraph in the class JSDoc, set off as an indented code block showing the two UNION branches sharing a single `{p1:String}` bind. (`f76d5d9`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Removed. The comment now just states that ARRAY JOIN and LIMIT BY ride as extra fields on the select node as real `ArrayJoinNode` / `LimitByNode` operation nodes — no reader-rationale prose. (`f76d5d9`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Understood — leaving it at test coverage, no runtime version guard added. I did trim the comment block to just note the exact version pin and that the array-index compile tests catch a breaking upgrade. (`f76d5d9`)

packages/shared/src/server/query-ast/kysely/compiler.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Fair. Added a short, explicit anti-pattern note right at the first cast — the line that overwrites Kysely's private `visitNode` — instead of relying on the paragraph above. Also shortened that paragraph so the point-of-use comment stands on its own. (`f76d5d9`)

packages/shared/src/server/query-ast/kysely/README.md
```diff
@@ -0,0 +1,105 @@
+# kysely/ — compile-only ClickHouse dialect: patterns & gotchas
+
+This folder adapts Kysely to emit ClickHouse SQL. Several patterns here are
+deliberately unusual — this guide is for anyone editing the folder. For the
+custom-node mechanics (ARRAY JOIN / LIMIT BY / metadata nodes) and the tenancy
+choke point, see [`../README.md`](../README.md); this file covers the rest.
```
niklassemmler: I think we should consolidate it. Let's keep everything that is relevant for anyone developing the query builder in here and everything that is relevant for using the query builder in the README one level up. Try to extract details from the comments here to make the comments shorter.

packages/shared/src/server/query-ast/kysely/schema.ts
```diff
@@ -0,0 +1,139 @@
+/**
```
niklassemmler, line 1: This file has no kysely imports can we pull it out of the kysely folder?

packages/shared/src/server/query-ast/kysely/README.md
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Consolidated along that split (`39c69e5c`):

- **`kysely/README.md` (developing)** now owns the internals: it absorbs the "how ClickHouse clauses land (no fork)" table + the node-vs-raw-SQL note, and the full tenancy-injection mechanics (the numbered choke-point steps, incl. the alias/table-name coverage rule). Its intro now points *down* to the folder rather than deferring *up*.
- **`../README.md` (using)** is now a usage guide: compile only through `compileClickhouseQuery(query, ctx)`, never filter `project_id` yourself, ClickHouse-only clauses via `$call(helper())` — then a pointer to `kysely/README.md` for internals. Dropped the duplicated extension table and mechanics.
- **Comments:** moved the visitNode-patch upgrade rationale to the README's "upgrade hazard" section and trimmed the code comment to a pointer. I intentionally kept the few code comments you'd explicitly asked for earlier (the `visitSelectQuery` diff, the anti-pattern note at the cast) rather than re-removing them — tell me if you'd rather those move into the README too.

packages/shared/src/server/query-ast/kysely/schema.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

You're right that it has no `kysely` imports — it's pure type/registry declarations. But before moving it: all five of its importers live inside `kysely/` (`tenancy.ts`, `extensions.ts`, `dialect.ts`, `typecheck.ts`, `schema.test.ts`), and its whole job is to type the relations the Kysely compiler targets (`ClickHouseDatabase` = the Kysely row types). So moving it up would put a file *outside* the folder that exclusively consumes it, keyed by "no import" rather than "no coupling".

My recommendation is to leave it in `kysely/` — the coupling is real even though the import isn't. But it's your call; if you'd still like it at the `query-ast/` level I'll move it and update the five `./schema` imports to `../schema`. Want me to?

packages/shared/src/server/query-ast/README.md
```diff
@@ -9,19 +9,42 @@ sugar, semantic field metadata, `FilterState` embedding) will land under
...
+## Using it
+
+- **Compile only through `compileClickhouseQuery(query, ctx)`.** It is the one
+  supported path from a builder to `{ sql, params }`; the repository layer hands
+  that to `queryClickhouse`. `ctx` (an `ExecutionContext` carrying `projectId`)
+  is required — omitting it is a compile-time type error and an empty one throws.
+- **Never filter `project_id` yourself.** The compile step injects
+  `project_id = {projectId}` into every tenanted relation, so call sites pass
+  only `{ projectId }` (see `repositories/environments.ts`).
+- **ClickHouse-only clauses use `$call(helper())`** — e.g.
```
niklassemmler: Can you please improve this so someone looking to learn "How do I write a limit by / array join / metadata select statement" can easily identify it?

packages/shared/src/server/query-ast/kysely/schema.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import {
+  condition7TypeAssertions,
+  extensionTypeAssertions,
+} from "./types.assert";
+
+describe("table registry derivation", () => {
+  it("derives the tenanted table set from the registry", () => {
+    // Only relations modeled with `tenant: true` are scoped. `events` /
+    // `events_full` are intentionally absent: they are not selectable (not in
+    // the row type) and raw sources are rejected, so they were unreachable.
```
niklassemmler: I don't follow. What is the difference between `events_core` and `events_full` from this angle? Shouldn't they behave just the same?

packages/shared/src/server/query-ast/kysely/schema.ts
```diff
@@ -0,0 +1,139 @@
...
+  Float: "number",
+  DateTime: "date",
+  "Array(String)": "array",
+  "Map(String, Float)": "map",
+};
+
+function defineTable<const Cols extends Record<string, ChColumnType>>(spec: {
+  columns: Cols;
+  /** Whether the tenancy pass must inject a `project_id` scope. */
+  tenant: boolean;
```
niklassemmler: This should default to true. I think we only have tables that are project gated.

packages/shared/src/server/query-ast/kysely/tenancy.ts
```diff
@@ -0,0 +1,357 @@
...
+ * tenanted physical relation that does not already have that predicate, then
+ * stamps the tree so {@link ClickHouseQueryCompiler} will compile it.
+ *
+ * Any `RawNode` whose SQL fragments introduce a relation (`SELECT` / `FROM` /
+ * `JOIN`) is rejected — not only FROM/JOIN table sources. Kysely's own
+ * keyword fragments (`asc` / `desc`) are not relations and are allowed.
+ */
+export class TenancyInjectionPlugin implements KyselyPlugin {
+  constructor(private readonly ctx: ExecutionContext) {
+    if (!ctx?.projectId) {
```
niklassemmler, line 57: Are we double-checking this? I remember we also had a check in the compiler.

packages/shared/src/server/query-ast/kysely/tenancy.ts
```diff
@@ -0,0 +1,357 @@
...
+function isProjectIdColumn(
+  node: OperationNode,
+  table: Extract<Relation, { kind: "table" }>,
+  requireQualified: boolean,
+): boolean {
+  if (ColumnNode.is(node)) {
+    // Bare column with no table qualifier: accept only when unambiguous.
+    return !requireQualified && node.column.name === PROJECT_ID_COLUMN;
+  }
+  if (ReferenceNode.is(node) && ColumnNode.is(node.column)) {
```
niklassemmler, line 281: In Kysely, can a reference be nested? Reference that connects to a reference ... connects to a ColumnNode?

packages/shared/src/server/query-ast/kysely/transformer.ts
```diff
@@ -0,0 +1,79 @@
...
+ * so without this override, ARRAY JOIN / LIMIT BY / the tenancy stamp would
+ * vanish the moment any plugin walked the tree.
+ */
+export class ClickHouseOperationNodeTransformer extends OperationNodeTransformer {
+  protected override transformNodeImpl<T extends OperationNode>(
+    node: T,
+    queryId?: QueryId,
+  ): T {
+    if (ArrayIndexNode.is(node)) {
+      return this.transformArrayIndex(node) as unknown as T;
```
niklassemmler, line 34: leave another comment explaining the anti-pattern

packages/shared/src/server/query-ast/README.md
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Added a **Recipes** section to the README with a clearly-labeled, copy-pasteable example for each: "Write an ARRAY JOIN", "Write a LIMIT BY", and "Select a metadata value" — each showing the `$call(helper())` form and the SQL it emits, so "how do I write X" is findable by heading. (`17751d2`)

packages/shared/src/server/query-ast/kysely/schema.test.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

You're right — from the tenancy angle they behave identically. Both `events_core` and `events_full` are project-gated physical tables (I checked the migrations: `0039_create_events_full`, `0040_create_events_core`, and `events_core` is populated by the `events_core_mv` materialized view). `events_full` would be `tenant: true` too if modeled.

The only reason it's not in the registry is that the builder targets `events_core` (the MV projection) and never selects `events_full` — so it isn't in the row type, raw sources are rejected, and it's unreachable, hence no entry needed. It is **not** exempt from tenancy. I reworded the test comment to say exactly this, and (see the schema.ts thread) `tenant` now defaults to true, so the day a query selects `events_full` it's scoped even if someone forgets the flag. (`17751d2`)

packages/shared/src/server/query-ast/kysely/schema.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Agreed. `defineTable`'s `tenant` is now optional and defaults to `true`, and the default is fail-closed — a relation added without the flag is still scoped, which is the safe direction. Set `tenant: false` only for a genuinely global table. All four current tables are project-gated, so I dropped the now-redundant explicit `tenant: true` lines — the default carries them and the derived `TENANTED_TABLES` set is unchanged (test still green).

packages/shared/src/server/query-ast/kysely/tenancy.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Good memory — there are two checks, but they guard different things and different entry paths, so it isn't redundant double-work:

1. **`projectId` presence** — checked here in the plugin constructor, and also by `requireExecutionContext` inside `compileClickhouseQuery`. The reason both exist: the compile path always goes through `requireExecutionContext`, but the plugin is *also* constructed directly (the tests do `new TenancyInjectionPlugin(ctx)`, and any future direct caller could too). The constructor check keeps the plugin safe on its own rather than relying on every caller having pre-validated.
2. **The compiler check is different** — it verifies the identity **stamp** (`assertTenancyStamped`/`hasTenancyStamp`), i.e. that the injection pass actually ran on this exact tree. That's a distinct guarantee from "projectId is present", and it's what makes `qb.compile()` without the plugin fail.

So: the `projectId` check is duplicated across two entry points (compile vs direct construction) by design; the compiler isn't re-checking projectId, it's checking the stamp. I ran a simplification pass over `tenancy.ts` and it flagged the same thing and concluded both should stay. Happy to consolidate if you'd prefer a single guard — but it would mean the plugin trusts its caller, which the direct-construction tests currently don't.

packages/shared/src/server/query-ast/kysely/types.assert.ts
```diff
@@ -0,0 +1,87 @@
+import { type InferResult } from "kysely";
+
+import { getClickhouseKysely } from "./dialect";
+import { arrayJoin, limitBy, mapKeys, mapValues } from "./extensions";
+import { defineView, fromView } from "./views";
+
+/**
+ * Compile-time assertions for condition 7 / 8. Never called at runtime;
```
niklassemmler: Nobdoy knows what "Compile-time assertions for condition 7 / 8." means.

packages/shared/src/server/repositories/environments.ts
```diff
@@ -11,54 +14,69 @@ export const getEnvironmentsForProject = async (
   props: EnvironmentFilterProps,
 ): Promise<{ environment: string }[]> => {
   const { projectId, fromTimestamp } = props;
+  const writeMode = env.LANGFUSE_MIGRATION_V4_WRITE_MODE;
+  const db = getClickhouseKysely();
+  // project_id is intentionally not filtered in the query bodies below: the
+  // compile step's mandatory tenancy pass injects `project_id = {projectId}`
+  // into every tenanted relation from this context, and refuses to compile a
+  // query that has no scope.
```
niklassemmler: Not relevant outside of this PR. Drop or restructure that only comments that explain behavior to future readers is kept.

packages/shared/src/server/query-ast/kysely/tenancy.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

No — a Kysely reference can't nest into another reference. `ReferenceNode` is typed `{ column: ColumnNode | SelectAllNode; table?: TableNode }`, so `.column` is never itself a `ReferenceNode`. The deepest shape is `ReferenceNode → ColumnNode → IdentifierNode`, plus an optional `TableNode` qualifier. `isProjectIdColumn` therefore covers every case: the bare `ColumnNode` branch, the qualified `ReferenceNode`-with-`ColumnNode` branch, and `SelectAllNode` (`t.*`) correctly falls through to `false` since it isn't a `project_id` column.

packages/shared/src/server/query-ast/kysely/transformer.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Done in `012fafed6`. Added a comment at the cast: `ArrayIndexNode` is a custom node kind outside Kysely's closed `OperationNode` union, so `transformArrayIndex`'s return type can't be proven assignable to the generic `T` the base method is parameterized over. The `.is` guard above establishes at runtime that `T` is `ArrayIndexNode` here, which is what makes the `as unknown as T` cast sound.

packages/shared/src/server/query-ast/kysely/types.assert.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Agreed — the "condition 7 / 8" numbering was internal shorthand that means nothing here. Fixed in `012fafed6`: renamed `condition7TypeAssertions` → `schemaTypeAssertions` (and updated the two test references), and rewrote the doc comment to state what it actually asserts — that the typed schema and virtual views enforce their guarantees at the type layer, with `tsc` as the test and every `@ts-expect-error` required to stay a real error.

packages/shared/src/server/repositories/environments.ts
niklassemmler: 🤖 _Replying as a Claude Code agent, on Niklas's behalf._

Done in `012fafed6`. Trimmed the comment to behavior-only prose that a future reader needs: tenancy is applied at compile time — `compileClickhouseQuery` injects `project_id = {projectId}` into every tenanted relation and refuses to compile an unscoped query, so the query bodies below carry no `project_id` filter of their own. Dropped the PR-narrative framing.

packages/shared/src/server/query-ast/kysely/tenancy.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (node.column.column.name !== PROJECT_ID_COLUMN) return false;
+    if (!node.table) return !requireQualified;
+    const referenced = tableNameOf(node.table);
+    return referenced === table.tableName || referenced === table.alias;
```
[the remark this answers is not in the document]
niklassemmler: should be fixed

PR #16706, opened 2026-08-27
feat(billing): trace ClickHouse Billing operations in APM and align plan codes

Steffen911:
🟡 **Live preview: [pr-16706.preview.langfuse.com](https://pr-16706.preview.langfuse.com)**

## What does this PR do?

Two things, both found by chasing one failed checkout on staging:

1. Adds APM spans around the ClickHouse Billing (CHB) REST client and its Auth0 token source, so a failing billing operation is diagnosable from a trace instead of only from application logs.
2. Fixes the plan codes we put on the wire — CHB validates `planCode` against its own enum and 400s ours.

**Motivation.** A checkout on staging returned a 500, and the trace carried an `http.server` span, three prisma spans, and *nothing* for the outbound calls to CHB or Auth0. The whole failure chain had to be reconstructed from logs: `POST checkout-sessions` came back `401 {"error":"Invalid or expired token"}`, the client invalidated its token, minted a fresh one, replayed, and got `401` again. APM could not show which attempt failed, whether a token was minted in between, or which issuer and audience the token carried — and that last point is what actually distinguishes the possible causes, since the 401 body is identical for all of them.

**What is instrumented.**

- `ChbApiClient` opens one span per CHB operation, named after the operation (`chb.checkout_session.create`, `chb.bundle.get`, `chb.bundle.scheduled.set` / `.clear`, `chb.invoices.list`, `chb.portal_session.create`) rather than after the route — a path carrying a bundle id would otherwise fan one operation out into an unbounded set of APM resources. Tags: `http.method`, `http.route`, `http.status_code`, `chb.ch_organization_id`, `chb.idempotency_key`, `chb.token_replayed`.
- Each HTTP attempt additionally gets its own `chb.api.request` span (`chb.attempt`, `http.status_code`, `peer.hostname`, `chb.operation`), so a token replay reads as two attempts rather than one opaque slow call.
- `ChbAccessTokenProvider` opens `chb.auth.token.mint` around the Auth0 client-credentials grant, tagging the tenant, audience and client id it asked for (`chb.auth.domain` / `.audience` / `.client_id`), the grant result (`http.status_code`, `chb.auth.expires_in_seconds`), and the token's shape: `chb.auth.token_format` (`jwt` / `opaque`) plus its `chb.auth.token_issuer` and `chb.auth.token_audience` claims.

Today's failure would now render as `chb.checkout_session.create` → `chb.api.request` (attempt 1, 401) → `chb.auth.token.mint` → `chb.api.request` (attempt 2, 401), with the issuer and audience visible on the mint span.

## Reviewer notes — the parts worth doubting

**Reading JWT claims without verifying the signature.** `tagTokenClaims` in `chbAccessToken.ts` base64url-decodes the payload to tag `iss` and `aud`. Those two claims are exactly what CHB's verifier checks, so tagging them turns an opaque `Invalid or expired token` into an attributable tenant-or-audience mismatch. This is a diagnostic on a token we received from Auth0 over TLS moments earlier, never an authorization decision — nothing branches on the values. Worth a look regardless: the token itself is never tagged or logged — that one is enforced by review, not by a test — and an undecodable payload leaves `token_format` as the signal rather than throwing, because a diagnostic must not be the reason a billing request fails.

**`token_format`.** Auth0 hands back an opaque token when the requested audience is not a registered API, and CHB's JWT verifier then rejects it. A 3-segment check is a cheap proxy that costs nothing and names that case directly.

**Span naming and cardinality.** Operation-named spans mean the span name is a fixed set; the id-bearing path stays on `http.route`, which is a tag. Reasonable to challenge if the house convention prefers route-named spans — the sibling `stripeBillingService` is operation-named (`stripe.subscription.retrieve`), which is what I matched.

**Layering.** Two spans per CHB call (operation + per-attempt) is deliberate: the operation span is the unit you alert on, the attempt spans are where retry behaviour becomes visible. I did **not** wrap the twelve `ChbBillingService` methods — each is already bracketed by the tRPC span above and the operation span below, so a third layer would be near-duplicate. Say so if you want it anyway.

**A non-obvious nesting detail.** The mint span nests *inside* the attempt span that triggered it, because `getToken()` is called while building the request headers. That is intentional and the code comment says so, but it is the kind of thing worth confirming reads well in a real flamegraph.

**Error status on attempt spans.** A 4xx/5xx attempt span records `http.status_code` but is not marked as an error; the operation span carries the failure. That keeps a 401-then-200 replay from inflating APM error rates while leaving the 401 queryable. Deliberate, but a judgement call.

## Plan codes

With the 401 resolved on the CHB side, the next staging attempt got a 400:

```
Request failed endpoint validation for checkoutSessionsCreate:
✖ Invalid option: expected one of "LANGFUSE_HOBBY"|"LANGFUSE_CORE"|"LANGFUSE_PRO"|"LANGFUSE_PRO_TEAMS"|"LANGFUSE_ENTERPRISE"
  → at planCode
```

We were sending `core` — the lowercased tier name from the original BIL-5791 spec. `chbPlanCodes` now mirrors CHB's `LANGFUSE_FIXED_PLAN_NAME_TO_PLAN_CODE` verbatim, since CHB owns that vocabulary and validates against it. Two entries are worth a second look:

- **`LANGFUSE_PRO_TEAMS` → `cloud:team`.** Confirmed with the CHB side; it is not `LANGFUSE_TEAM`.
- **`LANGFUSE_HOBBY` is listed even though a free Langfuse org carries no CHB bundle.** It is there so a Hobby bundle reported *by* CHB resolves to `cloud:hobby` instead of being dropped to null. We still never send it: the outbound code is only ever produced by the Stripe-product lookup, and no product maps to `cloud:hobby` — pinned by a test. Reasonable to argue this should have stayed out of the enum entirely.

CHB's `LANGFUSE_USAGE` is deliberately absent — it is the metered component, not a tier, so it resolves to no Langfuse plan. **This is the open question I would most want checked:** if a bundle response or webhook ever reports `LANGFUSE_USAGE` as `plan.planCode`, plan resolution nulls it silently rather than erroring, because `cloudConfigSchema` uses `.catch(null)` by design.

Nothing deployed writes `cloudConfig.clickhouse.planCode` yet (the writer is the webhook handler, still unmerged), so there is no stored code to migrate. Ordering is unaffected: CHB orders HOBBY < CORE < PRO < PRO_TEAMS < ENTERPRISE, matching the Stripe order keys `isChbUpgrade` reads. The plan code is part of the checkout and plan-switch idempotency keys, so those keys change with it — harmless here, since CHB rejected the old attempts at validation and never stored their keys.

**Not fixed here.** The original staging 401 was a CHB-side token-verification config problem and was resolved there, not in this repo. The spans are what named it, and what will name the next one.

## Type of change

## Mandatory Tasks

## Verification

- `pnpm --filter web run test chb getPlan` → `Tests  187 passed (187)`, `Test Files  11 passed (11)`
- `pnpm run lint` → `Tasks:    7 successful, 7 total`
- `pnpm run typecheck` → `Tasks:    8 successful, 8 total`

The two DB-backed CHB suites needed local Postgres/Redis, which this worktree did not have; brought up `docker-compose.dev.yml` and migrated `langfuse_test` before the run above.

**On test coverage.** An earlier revision asserted span names and tags in both CHB suites. Those assertions restated the instrumentation rather than protecting behaviour, and the span-recorder mock they needed to read attributes back was a large surface for that, so they are gone. One case survives because it is not about spans: `chbAccessToken.servertest.ts` feeds the provider a token whose claims cannot be decoded and asserts the grant still returns it — the diagnostic must never be the reason a billing request fails. It runs against the real tracer with no mocking.


Steffen911:
Dropped the span assertions from both CHB suites in 83c86b4c14 — they restated the instrumentation instead of protecting behaviour, and the span-recorder mock needed to read attributes back was a big surface for that. Net diff is now 3 files, +155/−24.

One case survives, and it is worth doubting whether I drew the line in the right place: `chbAccessToken.servertest.ts` feeds the provider a token whose claims cannot be decoded and asserts the grant still returns it. That one is not about spans — it pins the invariant that reading `iss`/`aud` for a tag can never fail a billing request. It needs no mocking and runs against the real tracer.

What that revert gives up: there is no longer a test asserting the access token itself never lands on a span. That property now rests on review of `tagTokenClaims`, which tags only `iss`, `aud` and a segment-count-derived format. Happy to put that single assertion back if you would rather have it enforced than reviewed.

Steffen911:
Applied the 409 finding in 301782fedf and resolved the thread. It was right, and it contradicted a choice I had already made one level down: the attempt span deliberately records `http.status_code` *without* being marked an error so a 401-then-200 token replay stays off the error rate, and then the operation span turned every routine "add a payment method" 409 into an APM error. `ChbPaymentRequiredError` is now carried out of the span as a value and rethrown outside it, tagged `chb.payment_required`. The caller contract is unchanged — the existing "maps 409 to ChbPaymentRequiredError" test still passes, which is the part worth protecting.

On the ruled-out note about concurrent callers not getting their own `chb.auth.token.mint` span: I looked and I am deliberately leaving it. Single-flighting means the Auth0 call genuinely happens once, and giving each waiter its own mint span would represent one HTTP request as N. The waiters are correctly shown as spending that time inside `chb.api.request`.

Also pushed 0bc2f156f1, which is a different bug the spans surfaced: with the 401 resolved CHB started returning a 400, because it validates `planCode` against `LANGFUSE_HOBBY|LANGFUSE_CORE|LANGFUSE_PRO|LANGFUSE_PRO_TEAMS|LANGFUSE_ENTERPRISE` and we were sending `core`. See the new "Plan codes" section in the description — the `LANGFUSE_HOBBY` inclusion and the absent `LANGFUSE_USAGE` are the two calls I would most like doubted.

Steffen911:
@claude review

Steffen911:
@claude review

PR #16711, opened 2026-08-27
feat(auth): lfe 15680 mcp two level authorization seam all modes

marksalpeter:
🟡 **Live preview: [pr-16711 app preview](https://pr-16711.preview.langfuse.com)**

Closes LFE-15680


web/src/features/auth/policy/shadow.mcp.ts
```diff
@@ -0,0 +1,55 @@
...
+    allowInAppAgentKey: true,
+  });
+
+  if (!result.success) {
+    if (env.API_AUTH_MIGRATION === "enforce") throw result.error;
+    return {};
+  }
+  return { authz: result.ctx };
+}
+
```
marksalpeter: lets remove all of this and have the mpc route call

```
  const result = await shadowAuth({
    req: params.req,
    allowedAccessLevels: ["project"],
    allowInAppAgentKey: true,
  });
```

directly

web/src/features/mcp/core/define-authorized-tool.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/** DefineAuthorizedToolOptions is a tool definition plus the project action the credential must hold to call it. */
+export interface DefineAuthorizedToolOptions<
+  TInput,
+  TName extends string = string,
+> extends Omit<DefineToolOptions<TInput, TName>, "preValidate"> {
+  /** Project action the credential must hold to call this tool. */
+  action: ProjectAction;
+}
```
marksalpeter: again, I think we remove all of this an just extend the existing defineTool implementation to `shadowAuthorize(ctx, action, projectId)` when the context is available.

web/src/features/mcp/core/define-tool.ts
```diff
@@ -118,6 +121,7 @@ export function defineTool<TInput, const TName extends string>(
     baseSchema,
     inputSchema,
     handler,
+    preValidate,
```
marksalpeter: unclear why this is needed

web/src/features/mcp/types.ts
```diff
@@ -63,6 +65,9 @@ export interface ServerContext {
 
   /** In-app-agent-specific MCP authorization state. */
   inAppAgent?: InAppAgentContext;
+
+  /** Resolved policy-core context, present only once the new authz path resolves the connection (shadow/enforce). */
+  authz?: AuthorizationContext;
```
marksalpeter: lets call this auth

web/src/features/auth/policy/shadow.mcp.ts
```diff
@@ -0,0 +1,55 @@
...
+/** assertMcpToolAccess authorizes a tool's own action against the resolved context, fail-closed; absent context means legacy or shadow, where the tool is not gated. */
+export function assertMcpToolAccess(params: AssertMcpToolAccessParams): void {
```
marksalpeter: this can just be directly in the tool call or inline as a helper function neat the tool definition

web/src/features/mcp/core/define-authorized-tool.ts
```diff
@@ -0,0 +1,33 @@
+import { assertMcpToolAccess } from "@/src/features/auth/policy/shadow.mcp";
```
marksalpeter: again this whole file can probably be removed.

web/src/pages/api/public/media/[mediaId].ts
Comment (bezbac) [inline web/src/pages/api/public/media/[mediaId].ts:1]: This looks unrelated from the MCP server

web/src/features/mcp/core/define-tool.ts
```diff
@@ -52,6 +56,7 @@ export interface DefineToolOptions<TInput, TName extends string = string> {
 export interface ToolDefinition<TName extends string = string> {
   name: TName;
   description: string;
+  action: ProjectAction | null;
```
bezbac: Should this be nullable?

bezbac: Which tools currently don't have an associated action?

web/src/pages/api/public/media/[mediaId].ts
Comment (marksalpeter) [reply web/src/pages/api/public/media/[mediaId].ts:1]: Sorry, this diff was created by a regression in an earlier branch. Feel free to ignore.

web/src/features/mcp/core/define-tool.ts
marksalpeter: Only the health check. I'm concerned that gating this would defeat the purpose of a health check. 

An alternative would be to create a `"public"` action that always passes authorization. This could potentially be worse from a security standpoint. I would have to think about how it impacts the policy engine.

bezbac: Makes total sense for the health check. I'm just a little worried about forgetting to set the correct scope. Maybe we could use a typescript discriminated union like:

```tsx
{
  action: ProjectAction
} | {
  __dangerouslyMakePublic: true
}
```

That way we make it super unergonomic and obvious.

marksalpeter: I like it 👍

marksalpeter: I decided to thread this concept through the api authorization layer instead of passing null or undefined actions around. The syntax now reads `action: __dangerouslySkipAuthz`. Then this can be consistant across the new auth seams (eg `enforceAuth(req, __dangerouslySkipAuthz, ...)  // tool definitions handle authorizaton`.

Fine by you?

bezbac: Love this. Thanks

bezbac (approved):

PR #16725, opened 2026-08-27
feat(filters): surface has:/-has: while typing a nullable field name

nkabardin:
🟡 **Live preview: [pr-16725 app preview](https://pr-16725.preview.langfuse.com)**

## TL;DR

Typing a nullable field name now offers `has:<field>` and `-has:<field>`.
Presence filters were previously reachable only if you already knew the `has:`
pseudo-field existed.

## Why

`has:` is the only way to filter on a column being set or missing, and nothing
led you to it: typing `userId` offered the `userId` field and nothing else. The
one hint that mentioned it was buried in the generic field list as
`has` — "field has a value, e.g. has:endTime".

## What

A **Presence** section in the field stage, on every view the bar serves. It
appears while you type, not only on an exact match, so `userI` already shows it.
A term that already carries a `-` gets only the positive form, since a second
one would splice `--has:`. Non-nullable fields and terms under two characters
stay quiet.

<details>
<summary>The second fix, and why it is in this PR</summary>

The `has:` hint's example field was derived from "the first nullable field in
the registry". That is the same derive-copy-from-ids mistake the placeholder
examples were changed to avoid, and it had already bitten: on the
generally-available events bar the hint silently changed from `has:endTime` to
`has:name`, because `name` happens to be the first nullable entry in `FIELDS`.
Reordering that list rewrote user-visible copy.

The example is now written down per view — events keeps `endTime`, sessions
uses `userIds` — with the derived first-nullable left only as a fallback. Both
halves of this PR are the same subject: `has:` being discoverable and honestly
advertised.

### Verified in the browser

- Sessions: typing `userId` shows `has:userIds` / `-has:userIds`; picking the
  negative applies `?filter=userIds;null;;is null;` and the bar re-renders
  `-has:userIds`.
- Events (GA): typing `userId` shows `has:userId` / `-has:userId`, and the
  generic hint reads `has:endTime` again.

`Tasks: 8 successful, 8 total` (turbo typecheck) · `Tasks: 7 successful, 7 total`
(turbo lint) · `Tests 265 passed (265)` (search-bar client tests).

</details>

## Result

The filter the grammar has always supported is now findable by typing the
column name, and its example copy can no longer drift when a field list is
reordered.

Typing a nullable field name now surfaces presence-filter completions, while each view can pin the field used by its generic `has:` example.

- Adds `has:<field>` and `-has:<field>` suggestions for nullable field-name matches.
- Preserves an existing leading negation when applying a completion.
- Adds per-registry `hasExample` configuration with a nullable-field fallback.
- Covers session and event registry behavior with client tests.
</details>

The pull request appears safe to merge with no actionable correctness or security issues identified.

Presence completions resolve nullable registry fields consistently, preserve an existing leading negation during replacement, and lower to the intended null or non-null filter semantics.
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(filters): surface has:/-has: while ..."](https://github.com/langfuse/langfuse/commit/5d8896f944447cc1640b4aa02b3f20732f3e0235) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57626703)</sub>


nkabardin:
@claude review

PR #16731, opened 2026-08-28
fix(clickhouse): allow explicitly empty migration passwords

### LINKED ISSUE #16730: bug: ClickHouse migration rejects empty CLICKHOUSE_PASSWORD
### Describe the bug

The ClickHouse migration scripts reject an empty `CLICKHOUSE_PASSWORD`, even though an empty password is valid for the built-in `default` user and the Langfuse app already accepts it.

[`packages/shared/clickhouse/scripts/up.sh#L27-L32`](https://github.com/langfuse/langfuse/blob/95933f73a6685884ae946961cb3a77249d68520d/packages/shared/clickhouse/scripts/up.sh#L27-L32):

In `up.sh` and `dev-tables.sh`:

~~~sh
if [ -z "${CLICKHOUSE_PASSWORD}" ]; then
  echo "Error: CLICKHOUSE_PASSWORD is not set."
  exit 1
fi
~~~

`-z` treats unset and empty variables the same, so `CLICKHOUSE_PASSWORD=""` causes the migration to fail.

The app defines `CLICKHOUSE_PASSWORD` as `z.string()`, so an empty value is valid. `down.sh` and `drop.sh` also use the same variable without this check.

### Steps to reproduce

Configure an external ClickHouse with the default user and an empty password:

~~~sh
CLICKHOUSE_PASSWORD=""
~~~

The web container then fails during ClickHouse migrations with:

~~~text
Applying clickhouse migrations failed.
~~~

## Proposed fix

Check whether the variable is set rather than whether it is non-empty:

~~~sh
if [ -z "${CLICKHOUSE_PASSWORD+x}" ]; then
  echo "Error: CLICKHOUSE_PASSWORD is not set."
  exit 1
fi
~~~

This keeps the current behavior for an unset variable while allowing an explicitly empty password.

| CLICKHOUSE_PASSWORD | Current | Proposed |
|---|---|---|
| unset | fail | fail |
| `""` | fail | pass |
| `somevalue` | pass | pass |

Self-hosted

v3

_No response_

### Additional information

Affected files:

- `packages/shared/clickhouse/scripts/up.sh`
- `packages/shared/clickhouse/scripts/dev-tables.sh`

Yes

mikemikimike:
Fixes a ClickHouse migration configuration bug: the password checks rejected an explicitly empty password even though it is valid for the default ClickHouse user. The checks now distinguish an unset variable from a set-but-empty value while preserving the existing failure for missing variables.

Added a dependency-free Bash regression test covering up.sh and dev-tables.sh with unset, empty, and non-empty passwords using stubbed commands.

Tests:
- bash -n on the changed scripts and regression test
- password_presence_test.sh
- git diff --check

Related issue: #16730

Fixes #16730


mikemikimike:
I rechecked this PR. The description targets the ClickHouse migration configuration bug, and there is no human changes-requested review or failing check reported for the current head. Could a maintainer please review whether the change resolves the issue and approve it, or advise on any missing validation?

PR #16756, opened 2026-08-28
feat(experiments): add the search bar to the experiments table

nkabardin:
🟡 **Live preview: [pr-16756 app preview](https://pr-16756.preview.langfuse.com)**

## TL;DR

Adds the query bar to the **experiments** table — the third surface, following
the extension recipe in `features/search-bar/README.md`. Deliberately narrow:
it covers the experiment columns the backend can actually filter, and nothing
else.

## Why narrow

Two exclusions, both because the alternative is a filter that renders a pill
and silently changes nothing.

**Score dot-paths stay closed.** The adapter lowers `scores.<name>` onto the
canonical `scores_avg` / `score_categories` / `score_booleans` columns
(`SCORE_COLUMNS`, applied unconditionally). Experiments names its score columns
`obs_*` and `trace_*`, so an open `scores.` namespace here would emit filters on
columns this view does not have. Renaming those columns is exactly what the
score-unification work does, so `scores.` arrives with it rather than shipping
broken. Pinned in a test so it cannot quietly open.

**The run metrics stay out.** `itemCount`, `errorCount`, `totalCost` and
`latencyAvg` have `ColumnDefinition`s — they are selected, displayed and
sortable — but **no `UiColumnMapping`**. The repository partitions filters by
allow-list:

```ts
const preAggFilterState  = filter.filter(f => experimentPreAggCols.some(c => c.uiTableId === f.column));
const scoreAggFilterState = filter.filter(f => experimentScoreAggCols.some(c => c.uiTableId === f.column));
```

Anything in neither group is **discarded with no error**. I had added sidebar
facets for all four, and they looked right until I checked the row count:
`errorCount:>0` wrote `?filter=errorCount;number;;>;0`, rendered the pill, and
returned all seven runs. Both the facets and the bar fields are reverted. These
are aggregates over items, so filtering them needs a `HAVING` path that does not
exist yet — see the follow-up note below.

So the bar offers `name`, `experimentDatasetId` (+ aliases `dataset`,
`datasetId`), and `metadata.<key>`. A bare word searches the experiment name,
since that is the one text column and what people look a run up by.

<details>
<summary>Verification and the wider finding</summary>

### Verified against a seeded project

- `sonnet` → commits as `name:sonnet`, URL `?filter=name;string;;contains;sonnet`,
  narrows 7 runs → the 3 sonnet runs.
- `metadata.model:*opus*` → `?filter=metadata;stringObject;model;contains;opus`,
  narrows to the 3 opus runs.
- `errorCount:>0` → correctly rejected as an unknown field (this is the case that
  used to look like it worked).

`Tasks: 8 successful, 8 total` (turbo typecheck) · `Tasks: 7 successful, 7 total`
(turbo lint) · `Tests 268 passed (268)`, including a new experiments block in the
invariants harness: all three invariants over the new registry, the exact field
list, score paths closed, the bare-word rewrite, and the run metrics staying
unresolvable.

### The silent-drop partitioning is worth a look on its own

That allow-list partition means **any** filter on an unmapped experiments column
vanishes without complaint — a typo'd column in a hand-edited URL, or a saved
view whose column was renamed, returns unfiltered results rather than an error.
Out of scope here, but it is the reason this PR is narrow rather than wrong.

</details>

## Result

Experiments filtering by keyboard for the columns that genuinely filter, with
the two gaps documented and pinned rather than papered over. Score filtering
stays in the sidebar until the unification lands.

Adds the shared grammar search bar to the experiments table, exposing supported experiment-name, dataset-ID, and metadata filters while intentionally excluding unsupported score and aggregate fields.

- Derives an experiments field registry from filterable sidebar facets.
- Integrates the search composer with the table’s existing filter state and sticky toolbar.
- Extends registry typing and server input validation for the experiments identifier.
- Adds invariant tests covering field parity, serialization, bare-word rewriting, and excluded fields.
</details>

The PR should not merge until the dataset example is changed to use a valid identifier or the field gains name-to-ID resolution; the unsupported server registry mapping should also be tightened.

The main filtering integration preserves existing filters and excludes unsupported columns, but its user-facing dataset example commits a display name against an ID column and therefore produces incorrect empty results.

**Files Needing Attention:** web/src/features/experiments/constants/experimentsSearchRegistry.ts, web/src/features/search-bar/server/router.ts
</details>

```mermaid
flowchart LR
  A[Experiment search input] --> B[Experiments field registry]
  B --> C[Validate and lower grammar]
  C --> D[Shared FilterState]
  D --> E[Experiment backend filter mappings]
  E --> F[Filtered experiments table]
```
</details>

`````markdown
### Issue 1
web/src/features/experiments/constants/experimentsSearchRegistry.ts:84
**Dataset example uses name**

When a user follows `dataset:my-evals`, the `dataset` alias commits `my-evals` to the exact-option `experimentDatasetId` field, but dataset options use `d.id` as the filter value and `d.name` only for display. The backend therefore compares a dataset name with stored dataset IDs, returning no matching experiments even when that named dataset exists.

### Issue 2
web/src/features/search-bar/server/router.ts:65
**Experiments selects events registry**

An authenticated `generateFilter` request with `registryId: "experiments"` passes this schema but falls through the registry selection to `EVENTS_FIELD_REGISTRY`. Because that registry enables AI filtering, the availability guard also passes and generates event-only filters instead of rejecting AI generation for the experiments view.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(experiments): add the search bar to..."](https://github.com/langfuse/langfuse/commit/fac0cb14cd7ba519aa84ef014f87426cf62d4e5c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57850543)</sub>


nkabardin:
@claude review

PR #16757, opened 2026-08-28
fix(billing): report Stripe webhook org lookup misses at the right severity

Steffen911:
🟡 **Live preview: [pr-16757 app preview](https://pr-16757.preview.langfuse.com)**

**TL;DR** — the Stripe webhook endpoint reports a handled, expected outcome as an APM error, so 0.2–0.3% of webhook requests per Cloud region are permanently "failing" on HTTP 200. This makes the real failures invisible and burns the billing availability error budget. The lookup now classifies *why* it could not find the organization, and each caller logs at a severity that matches whether anyone has to act.

## Why the misses happen

Stripe delivers every subscription event to **all** Cloud regions, but only one region owns the organization behind a subscription. `getOrgForSubscriptionWithFallbacks` tries four fallbacks (active subscription id → Stripe customer id → checkout session client reference → `metadata.orgId`) and, until now, ended every miss with `logger.error` + `traceException`. In the regions that do not own the subscription — the majority — that is the expected outcome, not a fault.

## What changed

The lookup returns a typed result instead of `null`, and no longer logs:

| miss | meaning | severity |
| --- | --- | --- |
| `other-region` | a region marker (subscription metadata or checkout session client reference) names a different region | `info` |
| `unattributable` | no region marker at all, so no region can own the event | `warn` |
| `org-row-missing` | a marker names this region, but the organization row is gone | `error` + traced exception |

Two details worth spelling out:

- **`unattributable` is deliberately not an error span.** Every region reports the same miss for the same subscription, so it is a fleet-wide fact about the Stripe object rather than one region's fault. Four error spans for one condition is not a useful signal; detecting it properly belongs in a billing reconciliation sweep that can read Stripe once.
- **`org-row-missing` on `customer.subscription.deleted` is expected.** Organization deletion cancels the subscription and then deletes the row, so Stripe delivers the deletion event after the row is already gone. There is nothing left to write and nothing to fix, so the caller marks it expected and it logs at `info`. The same miss on `created`/`updated` stays an error — that case means a subscription is live in Stripe with no organization behind it.

The log message now carries the miss reason, which makes a log monitor on the remaining error class precise.

## Tests

`web/src/__tests__/server/stripe-webhook-handler.servertest.ts` gains three cases that assert on `traceException`, since that call is what marks the webhook span as an error. All three fail against the previous behaviour (1, 2 and 2 calls respectively):

- no region marker on the subscription → no error span
- org row already gone on `subscription.deleted` → no error span
- org row gone on `subscription.updated` → exactly one error span (not two: previously the inner and outer layer each raised one)

The file mocks modules, so vitest's content-based classification moves it from the shared-context `server` project to `server-isolated` automatically.

## Verification

```
pnpm --filter web run test stripe-webhook-handler   # Tests  10 passed (10)
pnpm --filter web run test createDefaultSpendAlerts # Tests  5 passed (5)
pnpm run lint                                       # Tasks: 7 successful, 7 total
pnpm run typecheck                                  # Tasks: 8 successful, 8 total
```

No behaviour change for resolved organizations: the fallback order, the Stripe calls, and every write path are untouched. Only the severity of a miss changes.

This PR classifies failed Stripe subscription-to-organization lookups by region ownership and moves logging decisions to callers so expected cross-region misses no longer mark webhook spans as errors.

- Returns a typed organization lookup result with `other-region`, `unattributable`, and `org-row-missing` outcomes.
- Logs missing organizations according to whether the caller considers the outcome actionable.
- Treats a missing organization during subscription deletion as expected.
- Adds isolated tests for traced-exception behavior across the principal miss cases.
</details>

The PR appears safe to merge with no actionable correctness, security, or test-isolation issues identified.

Successful organization resolution remains unchanged, while unresolved lookups are consistently classified and tested so only actionable owning-region misses mark webhook spans as errors.
</details>

```mermaid
flowchart TD
  A[Stripe subscription event] --> B[Lookup by subscription ID]
  B -->|Found| Z[Process organization update]
  B -->|Missing| C[Lookup by customer ID]
  C -->|Found| Z
  C -->|Missing| D[Lookup checkout session]
  D -->|Found organization| Z
  D -->|Missing| E[Lookup metadata orgId]
  E -->|Found| Z
  E -->|Missing| F{Region marker}
  F -->|Other region| G[Info log]
  F -->|No marker| H[Warning log]
  F -->|Own region| I{Deletion event?}
  I -->|Yes| G
  I -->|No| J[Error log and traceException]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(billing): report Stripe webhook org ..."](https://github.com/langfuse/langfuse/commit/1a80fa92016e32fcff77bc0d2531663aac01487c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57856382)</sub>


Steffen911:
What I'd doubt in this change:

1. **`unattributable` → `warn` makes a real gap quieter.** Subscriptions with no region marker anywhere are genuinely unhandled by every region — plan changes on them are dropped today. I argue the fix is a reconciliation sweep that reads Stripe once rather than four error spans per event, but if that sweep never lands, this change trades a loud signal for a quiet one. Push back if you'd rather keep one region erroring until the sweep exists.

2. **Subscription metadata outranks the checkout-session marker.** If `metadata.cloudRegion` is ever wrong — stale after a region migration, or set by hand — a real miss in the owning region gets classified `other-region` and logged at `info`, i.e. silently dropped. Metadata is written once and never revised, so I think this is safe, but it is the one place where a wrong marker now costs visibility instead of noise.

3. **`expected: action === "deleted"` assumes a missing row means a deleted org.** It does in every case I traced, but a transient read failure (replica lag, failover) that returns no row on a deletion event would now log at `info` too. The blast radius is one log line, since the handler had nothing to write either way.

4. **The tests assert on `traceException` being called**, which is the mechanism that marks the span as an error rather than the outcome itself. If span error marking ever moves elsewhere, these tests would keep passing while the error rate regresses.

Also note the log message wording changed (the reason is now appended, and the old double-space `in Environment  US` phrasing is gone). No monitor referenced the old text, but saved log views might.

Steffen911:
Correction to doubt 1 above: it is resolved, not open. The subscriptions that hit the `unattributable` path are **self-hosted plans tracked in Stripe with no Cloud organization behind them** — confirmed on the payloads we log (product `prod_RCpfTSjkm7G9tt`, price *"Langfuse Pro (self-hosted), self-serve, USD100/seat"*, `collection_method: send_invoice`, no subscription metadata). No region can or should own those events, so there is no dropped-plan-change gap and no reconciliation sweep needed.

I'm still keeping that class at `warn` rather than `info`: a Cloud subscription created by hand in the Stripe dashboard without metadata would be indistinguishable from a self-hosted one, and `warn` keeps that case findable at zero cost to the error rate. Product id would be the discriminator if we ever want it silent (self-hosted products are not in `stripeProducts`, so `mapStripeProductIdToPlan` returns null for them) — but that would also silence the botched-Cloud-subscription case, which is why I did not do it here.

Doubts 2–4 stand.

PR #16762, opened 2026-08-28
feat(experiments): filter datasets by name instead of id

nkabardin:
🟡 **Live preview: [pr-16762 app preview](https://pr-16762.preview.langfuse.com)**

## TL;DR

Experiments filter by **dataset name**, not dataset id. The sidebar already
showed names; the URL, saved views and search bar carried cuids. Now everything
says `legal-answer-quality`.

> [!NOTE]
> **Second in a stack.** Base is `lfe-15470-move-experiments-to-the-filter-bar-…`
> (#16756), which adds the bar to this table. Score-facet unification follows on
> top of this one.

## Why

The two surfaces disagreed. `useExperimentFilterOptions` built dataset options as
`{ value: d.id, displayValue: d.name }`, so the sidebar's picker rendered the
name while the emitted filter carried the id. That was invisible until the search
bar arrived and had to render the filter as *text*: it offered
`dataset:cmtc1zmkv0007phmyo7xum5lr`, and typing `dataset:legal` matched nothing.

Making the bar display names while the text held ids would mean the derived
committed text depends on async-loaded filter options — a first paint showing
ids that flips to names when options land, and ids whenever they fail. That text
deriving purely from URL state is the invariant every bar surface relies on.

So instead of teaching two surfaces to translate, the **filter value itself is
now the name**.

## What

- The dataset facet is `experimentDatasetName`, whose options are plain names —
  no `displayValue` indirection.
- `fns/datasetNameFilter.ts` swaps name → id at one boundary, where the query
  payload is built. The name is not a ClickHouse column; it reaches the client
  from Postgres with the filter options, so this is the only place it can happen.
  Names are unique per project, so the mapping is lossless.
- An unknown name resolves to no id rather than being dropped — dropping it
  would widen the query to every dataset, and a deleted dataset must return
  nothing.
- `experimentDatasetId` stays filterable but is no longer offered, so existing
  links and saved views keep resolving.

<details>

Checked against a seeded project with two datasets:

- `dataset:` offers `groundedness-judge-calibration` and `legal-answer-quality`
  — names, where it previously listed cuids.
- Picking one commits as `dataset:legal-answer-quality`, writes
  `?filter=experimentDatasetName;stringOptions;;any of;legal-answer-quality`,
  and narrows 7 runs → the 5 in that dataset.
- **Backwards compat**: an old id-based permalink
  (`experimentDatasetId;…;cmtc1zmkv…`) still filters correctly — same 5 rows,
  `Filters 1` in the sidebar. The bar leaves it as a preserved skipped filter
  rather than dropping it, which is the documented no-silent-drop behaviour and
  matches how #14692 retired `trace_scores_avg`.
- The Dataset column still renders the name (it now reads an id → name map
  rather than the options' `displayValue`).

`Tasks: 7 successful, 7 total` (turbo lint) · typecheck clean ·
`Tests 306 passed (306)`, including four cases on the translation: the swap,
untouched non-dataset and pre-existing id filters, the unknown-name case, and
multi-value selections.

</details>

## Result

One readable, unique value for datasets across the sidebar, the bar, the URL and
saved views — and one boundary where it becomes an id.

The PR makes dataset names the canonical experiment-filter value and translates them back to dataset IDs at the query boundary.

- Adds a name-based dataset facet and search-bar field while preserving legacy ID filters.
- Builds bidirectional dataset lookup maps for filtering and table rendering.
- Adds tests for name resolution and experiment search-field invariants.
</details>

The PR should not merge until valid persisted dataset-name filters can resolve independently of sampled and capped facet options.

Building the translation map from restricted filter-option results can convert a valid dataset name into a nonmatching dataset ID filter and incorrectly return an empty experiment table.

**Files Needing Attention:** web/src/features/experiments/hooks/useExperimentFilterOptions.ts; web/src/features/experiments/fns/datasetNameFilter.ts
</details>

```mermaid
flowchart LR
  A[URL, saved view, or picker: dataset name] --> B[datasetIdByName lookup]
  C[Sampled and capped filter options] --> B
  B -->|Found| D[experimentDatasetId filter]
  B -->|Missing| E[Name retained as ID value]
  D --> F[Experiments query]
  E --> F
  F --> G[Filtered table results]
```
</details>

`````markdown
### Issue 1
web/src/features/experiments/hooks/useExperimentFilterOptions.ts:56-58
**Lookup omits valid datasets**

When a valid dataset is omitted from the sampled, top-1000 filter-option response, this map lacks its name, so `withDatasetNamesResolved` submits the name as an `experimentDatasetId` and the experiment table incorrectly returns no matching rows. Build the translation map from the complete project dataset list rather than the restricted facet options.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(experiments): filter datasets by na..."](https://github.com/langfuse/langfuse/commit/3ca1dd340db5325b30cf74adc2ca50f0140c5ee5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57866259)</sub>

**Context used:**

### DIFF
--- a/web/src/features/experiments/components/table/ExperimentsTable.tsx
+++ b/web/src/features/experiments/components/table/ExperimentsTable.tsx
@@ -280,15 +280,6 @@ export default function ExperimentsTable({
   showControlsInPageHeader = false,
 }: ExperimentsTableProps) {
   const router = useRouter();
-  const filterConfig = useMemo(
-    () =>
-      getExperimentsFilterConfig(
-        fixedFilter
-          .map((filter) => filter.column)
-          .filter(isExperimentsOmittableFilterColumn),
-      ),
-    [fixedFilter],
-  );
 
   const { setDetailPageList } = useDetailPageLists();
   // Selection lives in a per-mount vanilla zustand store (not useState) so a
@@ -352,6 +343,19 @@ export default function ExperimentsTable({
     projectId,
     oldFilterState,
   });
+
+  // Built after the dataset map, which its filter-state migration needs to
+  // translate a legacy dataset id into the name the facet is keyed by.
+  const filterConfig = useMemo(
+    () =>
+      getExperimentsFilterConfig(
+        fixedFilter
+          .map((filter) => filter.column)
+          .filter(isExperimentsOmittableFilterColumn),
+        datasetNameById,
+      ),
+    [fixedFilter, datasetNameById],
+  );
 
   const queryFilter = useSidebarFilterState(filterConfig, filterOptions, {
     loading: isFilterOptionsPending,
--- a/web/src/features/experiments/components/table/ExperimentsTable.tsx
+++ b/web/src/features/experiments/components/table/ExperimentsTable.tsx
@@ -11,6 +11,7 @@ import { useQueryFilterState } from "@/src/features/filters/hooks/useFilterState
 import { usePaginationState } from "@/src/hooks/usePaginationState";
 import { useSidebarFilterState } from "@/src/features/filters/hooks/useSidebarFilterState";
 import { EXPERIMENTS_FIELD_REGISTRY } from "@/src/features/experiments/constants/experimentsSearchRegistry";
+import { withDatasetNamesResolved } from "@/src/features/experiments/fns/datasetNameFilter";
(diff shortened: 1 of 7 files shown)


nkabardin:
@claude review

PR #16764, opened 2026-08-28
feat(experiments): unify experiment score filters across levels

nkabardin:
🟡 **Live preview: [pr-16764 app preview](https://pr-16764.preview.langfuse.com)**

## TL;DR

The experiments runs table had **six** score filter facets — three for observation-level scores, three for trace-level. Now it has **three**, and each offered score name is tagged with the level(s) it exists at. Filtering matches at either level; excluding means "at neither level". Old `obs_*` URLs and saved views keep working.

## Why

Whether a score lands on an observation or on the trace is an instrumentation detail — it's decided by where the SDK call sits, not by what the user wants to ask. But the filter UI made the user pick that detail up front, in a pair of identically-named facets:

> Scores (numeric) · Scores (categorical) · Scores (boolean) · Trace Scores (numeric) · Trace Scores (categorical) · Trace Scores (boolean)

Pick the wrong half and the filter matches nothing, with no hint why. Worse, both halves were offered the same project-wide score-name list, so a trace-level-only score appeared under the observation facet — where it could never match.

## What

Three facets by data type — **Numeric / Categorical / Boolean** — with the level marked per option (`Observation` / `Trace`), the way the tracing tables already mark them. One score name, one row, tagged.

Matching is level-agnostic. Rather than unioning two joins, the score aggregation keeps the level inside its `GROUP BY`, so the array holds one entry per (name, level). Because score filters compile to array-existence checks, a positive filter then matches at either level, and De Morgan gives the exclusion — "at neither level" — for free, with no special casing.

## Result

Verified against a seeded demo project (7 runs, scores at both levels):

| Filter | Level of the score | Rows | Matches ground truth |
| --- | --- | --- | --- |
| `groundedness > 0.5` | trace only | 2 of 7 | the 2 runs averaging 0.633 / 0.524 |
| `judge_verdict any of [fully-grounded]` | trace only | 2 of 7 | the 2 runs carrying the score |
| `judge_verdict none of [fully-grounded]` | trace only | 5 of 7 | exactly complementary |
| `obs_scores_avg` legacy URL, `supports > 0.65` | observation | 2 of 7 | same rows, URL rewritten to the canonical column |

Before this change the first three could not match at all: the offered facet compiled against the observation-level array only.

<details>
<summary>Mechanism, compatibility, and verification</summary>

**Backwards compatibility.** `obs_scores_avg` / `obs_score_categories` / `obs_score_booleans` are aliases of the canonical columns, so existing saved views and bookmarked URLs resolve and are rewritten to the canonical id on load. `trace_*` stays filterable but is no longer offered. This matters more than usual here: the repository partitions filters by an allow-list and **silently discards** anything it cannot resolve — a lost alias would not error, it would widen the result set. A unit test pins every alias, and a servertest asserts a legacy filter still bites (with a second unscored run present, so a dropped filter would show up as an extra row).

**Discovery is scoped to what can match.** The agnostic filter only matches a score that rolls up into a trace, so discovery is scoped the same way (`trace_id IS NOT NULL`); session- and dataset-run scores are deliberately not offered, since offering them would advertise a filter that cannot fire. Levels are then discovered per level to build the tag map. Offered-set == matchable-set.

**The level map is split per data type** (`score_name_levels_{numeric,categorical,boolean}`) because a name can be reused across types at different levels — a numeric observation-level `accuracy` beside an unrelated categorical trace-level one. A name-only map would mislabel both.

**The per-level shape stays.** The run charts genuinely need it: a merged metric identity cannot say which level's series to plot. So the per-level lists remain the source and the agnostic set is a projection on top, added alongside rather than replacing — chart consumers are untouched.

**Not included: the experiment *items* table.** Its score filters run through a different path — two separate aggregation CTEs joined on different keys (`observation_id` vs `trace_id`) — so unifying it is a query change of its own rather than a reuse of this one, and that table is being substantially rewritten in parallel work. Its facets still read per-level.

**Checks.** `pnpm run lint` → `Tasks: 7 successful, 7 total`. `pnpm run typecheck` → `Tasks: 8 successful, 8 total`. `pnpm --filter @langfuse/shared run test experiments.test` → `Tests 10 passed (10)`. `pnpm --filter web run test experiment-score-levels` → `Tests 4 passed (4)`.

</details>

---

Stacked on #16762 (which is stacked on #16756).

This PR consolidates experiment-run score filtering into numeric, categorical, and boolean facets that operate across observation and trace levels while retaining legacy filter compatibility.
- Adds level-agnostic ClickHouse score aggregation while preserving per-level aggregation for charts and legacy trace filters.
- Merges score-option discovery across observation and trace levels and supplies per-type level tags.
- Canonicalizes observation score filter identifiers while retaining `obs_*` aliases for existing URLs and saved views.
- Updates the experiments UI, server router, and tests for the unified filtering behavior.
</details>

The PR appears safe to merge; no concrete changed-code defect or security issue remains after reviewing the unified aggregation, option discovery, and alias compatibility paths.

The new canonical filters consistently map to level-aware score arrays, legacy observation identifiers are normalized at both frontend and backend boundaries, and the investigated edge cases were either unreachable from current consumers or contradicted by the implemented predicates and tests.
</details>

```mermaid
flowchart LR
  O[Observation-level scores] --> A[Level-aware aggregation]
  T[Trace-level scores] --> A
  A --> N[Numeric score facet]
  A --> C[Categorical score facet]
  A --> B[Boolean score facet]
  L[Legacy obs_* filters] --> X[Alias normalization]
  X --> N
  X --> C
  X --> B
  N --> Q[Experiment query]
  C --> Q
  B --> Q
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(experiments): unify experiment scor..."](https://github.com/langfuse/langfuse/commit/cb005b29d8ee7945c7276bf6d29f6a0d0e859222) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57873933)</sub>

</details>


nkabardin:
@claude review

nkabardin:
@claude review

PR #16781, opened 2026-08-28
feat(experiments): unify item score filters across levels

nkabardin:
🟡 **Live preview: [pr-16781 app preview](https://pr-16781.preview.langfuse.com)**

## TL;DR

The items view *inside* an experiment kept the six split score facets the runs table just shed. Now it has three — Numeric / Categorical / Boolean — with each offered score name tagged `Observation` or `Trace`, matching at either level. This is the last surface in the app that split score filters by level.

## Why

#16764 unified the experiments **runs** table. Open one experiment and the same six facets were still there:

> Scores (numeric) · Scores (categorical) · Scores (boolean) · Trace Scores (numeric) · Trace Scores (categorical) · Trace Scores (boolean)

An LLM-as-judge writing to the trace and a metric written to the item's span are the same question to the person filtering — which items scored well — and the level is decided by where the SDK call sits, not by what they're asking.

## What

Items needed a query change, not a config change. Its filters compile against **two** aggregate CTEs joined on **different keys** (`observation_id = the item's root span`, `trace_id = its trace`), and a flat filter can't say "either level" across two aliases.

So there is now one per-item aggregate, keyed by the **root span id** — unique per (experiment, item), so a join on it scopes to one run without needing an experiment predicate of its own — that unions both levels with the level kept in the inner `GROUP BY`. One array entry per (name, level) means a positive filter matches at either level, and its negation means "at neither", with no special casing. Observation-level scores keep counting only when they sit on the item's root span, which is the pre-existing scope.

## Result

Verified against a seeded experiment, each number checked against a ClickHouse query first:

| Filter | Score's level | Items |
| --- | --- | --- |
| `judge_verdict any of [fully-grounded]` | trace only | **3 of 20** |
| legacy `obs_score_categories` URL, `verified_article_status any of [all-in-force]` | observation | **8 of 15**, URL rewritten to the canonical column |

The first was impossible before: the offered facet compiled against the observation-level aggregate only, so a trace-level score could never match it.

<details>
<summary>Mechanism, compatibility, and verification</summary>

**Compatibility.** `obs_scores_avg` / `obs_score_categories` / `obs_score_booleans` are aliases of the canonical columns, so saved views and bookmarked URLs resolve and rewrite to the canonical id on load — and start matching trace-level scores too. `trace_*` stays filterable but is no longer offered. The items table's score *columns* are unchanged and still per-level; this is about filtering.

[...]

PR #16786, opened 2026-08-28
feat(experiments): an item's scores are its whole run's scores

nkabardin:
🟡 **Live preview: [pr-16786 app preview](https://pr-16786.preview.langfuse.com)**

## TL;DR

If your evaluator scores the **model call inside** each experiment item rather than the item's own top-level span, those scores did not exist as far as the experiment views were concerned: empty score column, name missing from the score picker, impossible to filter on. This makes an item's scores mean *the scores of that item's whole run*.

## Why

You point an evaluator at an experiment and ask the obvious question afterwards — **which items did badly?**

Where a score physically lands is an instrumentation detail. An evaluator on traces writes to the trace. One configured on generations writes to the individual LLM call inside the item. One that scores the item as a whole writes to its top-level span. All three are normal, and none of them is a statement about what you want to see in the table.

The experiment views only read observation-level scores from the item's **top-level span**. Trace-level scores were always read. So whether a score appeared at all depended on which of those three an evaluator happened to use.

How much this hides is not marginal. In our seeded demo project, two scores are written by the evaluator to the model call inside each item:

| Score | On item top-level spans | On the model call inside |
| --- | --- | --- |
| `supports` (numeric) | **0** | 210 |
| `in_force` (boolean) | **0** | 210 |

Every one of those 420 scores was invisible in the experiment items view. They existed, and you could see them by opening the individual trace — but the column was blank, the name was not in the score picker, and no filter could reach them. Meanwhile `groundedness`, written at trace level, showed up fine. Same experiment, same run, same question, different answer purely because of where the SDK call sat.

## What

An observation-level score now counts when it sits on **any** observation the item produced. Concretely, on the items view inside an experiment:

- the score **column shows a value** per item, averaged over that item's run
- the score **name appears in the picker**, tagged `Observation`
- **filtering on it works** — `supports > 0.65` narrows the item list

Unchanged: trace-level scores behave as before, and experiment-level (whole-run) scores stay their own separate thing rather than being folded in.

Display, filtering and the offered name list all move together on purpose. Widening only the filter would let you match scores you cannot see; widening only the display would show scores you cannot filter on.

## Result

The question "which items scored badly" now gets the same answer regardless of how the evaluator was wired.

**One thing to weigh before merging:** an item's score cell can now aggregate more scores than before — `Ø` over the run's observations, where it used to be the single value on the top-level span. That is the intended meaning, and it matches the rule the runs table already applied to its own score columns, but it does mean a number someone has been reading can change.

<details>
<summary>Mechanism, what to doubt, and verification</summary>

**Mechanism.** Three paths, all backend:

- **The per-item score aggregate** joins scores to every span of the item, keyed by the item's root span id — unique per (experiment, item), so the join scopes to one run without needing an experiment predicate of its own. The trace-level branch stays pinned to the root row so a trace score is not counted once per span. The level stays in the inner `GROUP BY`, so the arrays hold one entry per (name, level).
- **Score-name discovery** scans the same spans, at both experiment surfaces. This also fixes an existing inconsistency on the runs surface: its filters already matched any observation-level score in a run's traces while discovery only offered the ones on root spans, so it under-offered names you could filter on.
- **The items query** reads observation scores for the whole subtree via a new `getObservationScoresForExperimentItems`, tagging each score with the item's root span so the existing row grouping is unchanged.

**What a reviewer should doubt**, in the order I would want it doubted:

1. **The `OR` inside the join condition.** `item_spans` is one row per observation, so the join input grows with subtree size. The scan is scoped to the selected experiments, but this was **not benchmarked on a large project** — that is the open question.
2. **The aggregation semantics, not just the SQL.** Two observations carrying the same numeric score name collapse through `avg()`. Intended, and the reason the cell can change, but worth confirming it reads correctly for a run with many same-named scores.
3. **`scores FINAL` + `LIMIT 1 BY s.id, s.project_id, isp.root_span_id`** in the new reader. The shape is copied from the existing item-score reader, but a score can now be reached through several spans of the same item, so the dedup key deserves a second look.

**Verification.** `pnpm run lint` → `Tasks: 7 successful, 7 total`. `pnpm run typecheck` → `Tasks: 8 successful, 8 total`. `pnpm --filter web run test experiment-score-levels` → `Tests 10 passed (10)`, two of them new: an item scored *only* on a nested generation qualifies, and does **not** qualify above the value (so a silently dropped filter cannot fake a pass); and its score name is offered with the right level tag. Both fail against the previous root-span-only scope — verified by reverting the scope, **rebuilding `@langfuse/shared`**, and re-running: `Tests 2 failed | 8 passed`. That rebuild matters: this suite resolves `@langfuse/shared` to the built package, so a source-only revert proves nothing.

`pnpm --filter web run test experiment-repository` → `Tests 25 passed (25)` (the suite that owns the score-option shapes).

</details>

---

Stacked on #16781 → #16764 → #16762 → #16756, all merged.

This PR consistently expands experiment-item observation scores from the root span to every observation belonging to the item’s run.

- Updates item filtering and score-name discovery to include nested observation scores.
- Adds a project- and experiment-scoped score reader that tags each score with its item root.
- Uses the root tag to aggregate nested scores into existing experiment-item response rows.
- Adds server tests covering nested-score filtering and discovery.
</details>

The PR appears safe to merge, with the changed score retrieval, discovery, filtering, and grouping paths aligned around the same experiment-item subtree semantics.

The new queries remain scoped by project and selected experiments, join observation scores to explicitly tagged item spans, preserve root-span grouping, and include focused tests for nested-score discovery and threshold filtering.
</details>

```mermaid
flowchart LR
  E[Experiment item events] --> S[All item spans]
  S --> J[Join observation scores by project, trace, and span]
  J --> R[Tag scores with item root span]
  R --> A[Aggregate scores per experiment item]
  A --> V[Display, discovery, and filtering]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(experiments): an item&#39;s scores are ..."](https://github.com/langfuse/langfuse/commit/47fa8297d90e4b0a77164307f7b189708c243f3a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57933548)</sub>

**Context used:**


nkabardin:
@claude review

marliessophie:
Hey @nkabardin! Could you share some context on why you made this change? Specifically, could you explain the context that lead to this sentiment: "evaluators often score the nested generation/model call instead of the experiment item’s root span"? All scores created via langfuse evaluation feature or the experiment runner or prompt experiments should live on the experiment item root span. Thank you! 

Context: We deliberately chose to include only the scores that sit on the experiment item root observation as in the context of experiments, this is recommended and documented behaviour.

nkabardin:
thanks @marliessophie this is exactly why I wanted you to look at it. I was in doubt if this is even necessary at all, but it was agent trying to apply practices from tracing to experiments (it was prompted to plan for feature parity)

marliessophie:
I see. In that case I would vote to close this PR. The change comes at performance cost that I am unsure outweighs the gain of merging.

nkabardin:
Yes, absolutely. Thanks for confirming the doubt! Now it's clear

marliessophie:
Thank you!

nkabardin:
Closing per @marliessophie's ruling above: in experiments, scores are meant to sit on the experiment item's root observation, and every supported writer (evaluations, the experiment runner, prompt experiments) puts them there — so the subtree broadening buys nothing and costs a wider join. The search-bar follow-up that was stacked on this branch is being made independent and stays open.

PR #16788, opened 2026-08-28
feat(experiments): show score filters in the search bar

nkabardin:
🟡 **Live preview: [pr-16788 app preview](https://pr-16788.preview.langfuse.com)**

## TL;DR

A score filter added in the experiments sidebar left the search bar empty, so the bar read as though no score filter were applied. The bar now renders score filters and offers every score name when you type `scores.` — the same as the tracing tables.

## Why

The sidebar and the bar edit **one** filter state, and the bar is meant to be a view of it. A filter it cannot express is a filter it silently omits, and the user sees an empty bar next to an active filter.

The `scores.` namespace was closed on this view deliberately, and the reason is in the registry's own comment: `scores.<name>` lowers onto the canonical `scores_avg` / `score_categories` / `score_booleans` columns, and experiments used to name its score columns `obs_*`. Unifying the facets moved this view onto exactly those columns, so the blocker is gone.

## What

One flag: `scores: true` on the experiments registry.

Everything the bar needs was already in place. The score-type context — which routes a typed name to numeric, categorical or boolean — is derived from the same observed options the sidebar uses, keyed by the canonical column names the unification introduced.

`traceScores.` stays closed. The canonical columns already match at trace level, so a second namespace would be another way to say the same thing, against `trace_*` columns the sidebar no longer offers.

## Result

| | |
| --- | --- |
| Sidebar sets a score filter | bar shows `scores.citation_usable_rate:4` |
| Type `scores.` | 18 names, each tagged `Observation`/`Trace` and typed (`numeric` / `categorical` / `boolean` score) |
| Type `traceScores.` | not offered |

The example in the placeholder is now `scores.groundedness:>0.8`, and the stale `dataset:` note about ids is gone — that view filters datasets by name since #16762.

<details>

`pnpm run lint` → `Tasks: 7 successful, 7 total`. `pnpm run typecheck` → `Tasks: 8 successful, 8 total`. `pnpm --filter web run test-client "features/search-bar"` → `Tests 269 passed (269)`.

The invariant test that pinned the closed namespace now pins the open one, plus a case for the reported bug: a `scores_avg` filter must derive to `scores.<name>:<value>` with nothing skipped. INV-4 still holds — the derived text parses for this registry — and the experiments sidebar cases now expect text rather than a skip.

Verified live on a seeded project: the reported URL renders the token, and typing `scores.` lists all 18 names across the three data types.

</details>

---

Stacked on #16786 → #16781 → #16764 → #16762 → #16756.

The PR enables the `scores.` namespace in the experiments search registry so sidebar score filters render in the search bar and typed score filters become available.

- Opens `scores.` while retaining the closed `traceScores.` namespace.
- Updates search examples and invariant tests for score resolution and serialization.
- The enabled path can misroute score filters committed before observed score types are available.
</details>

The score-filter activation should be fixed before merging because filters committed before score metadata loads can silently target the wrong score column and return incorrect experiment results.

The new experiments namespace accepts score expressions even with an empty observed-type context, and the adapter's fallback persists an incorrectly typed filter that the backend applies without later correction.

**Files Needing Attention:** web/src/features/experiments/constants/experimentsSearchRegistry.ts; web/src/features/search-bar/lib/searchBarInvariants.clienttest.ts
</details>

`````markdown
### Issue 1
web/src/features/experiments/constants/experimentsSearchRegistry.ts:80
**Score types route incorrectly**

When a user commits a boolean score such as `scores.in_force:true`, or a numeric-labeled categorical score, before the observed score types are available, the enabled namespace routes it to `score_categories` or `scores_avg` respectively. The experiment backend accepts the mismatched filter but searches an array that does not contain that score, causing empty or incorrect results; loading the options later does not correct the committed filter.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(experiments): show score filters in..."](https://github.com/langfuse/langfuse/commit/252d3c243f67e7fcc3b36751d733b4f906636bab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57938084)</sub>


nkabardin:
Superseded by #17209 — same change rebuilt as a single commit on `main`. This PR was stacked on #16786's branch; #16786 was closed and deleting its branch closed this one, and GitHub will not retarget a stacked PR's base.

PR #16803, opened 2026-08-29
fix(in-app-agent): surface output-token truncation instead of a blank turn

DavidTraina:
## Problem

When the model's last step ends with finish reason `length`, the Assistant renders an empty turn with no explanation.

Concretely, with an OpenAI Responses model the provider returns HTTP 200 with `status=incomplete` and `incomplete_details.reason=max_output_tokens`. `@ai-sdk/openai` maps that to `finishReason.unified = "length"`, Mastra treats `length` as terminal (pending tool calls are discarded), and `onIterationComplete` fires with `finishReason: "length"`. The worker's `complete()` only knows how to detect step-limit truncation (`iteration >= IN_APP_AGENT_MAX_STEPS && lastFinishReason !== "stop"`), so a `length` finish on an early step persists the run as `SUCCEEDED` with no `errorCode`. `getBackgroundRunNotice` therefore returns `null`, and the user sees a blank or seemingly stuck assistant turn.

The in-app agent never sets `maxOutputTokens`, so this depends entirely on the provider's default output cap and on how much the model wants to write (reasoning models are especially prone to it).

## Fix

Mirror the existing `STEP_LIMIT` path for output-token truncation:

- **shared**: new `InAppAgentRunErrorCode.OUTPUT_LIMIT = "output_limit"`.
- **worker `agent.ts`**: `InAppAgentCompleteOutcome.truncatedByOutputLimit` (`lastFinishReason === "length"`), passed to `onComplete`; the trace-level `result` metadata now carries both truncation flags and the finish reason whenever either truncation happened.
- **worker `executeInAppAgentRun.ts`**: a small `resolveCompletedRunFinish(outcome)` maps `truncatedByOutputLimit` -> `SUCCEEDED` + `errorCode: output_limit` + `errorMessage: "The response hit the model's output-token limit before a final answer"`. It takes precedence over `step_limit` because the length finish is the proximate cause. Adds the `langfuse.in_app_agent.output_limit_reached` counter next to `step_limit_reached`.
- **web `backgroundExecutionSession.ts`**: `SUCCEEDED + output_limit` shows a warning notice ("The assistant had to stop before finishing this answer. The response hit the model's output limit. Send another message to continue.") and the settled activity reads as "Stopped". The two truncation codes are now looked up from one small map rather than two hand-written conditions.

Successful runs are unchanged: a `stop` finish still yields `SUCCEEDED` with no error code and no notice, and `step_limit` behaves exactly as before.

## Why a new error code

Per `concept-reformulate-before-extending` (`.agents/skills/backend-dev-guidelines/references/new-concepts.md`), the existing vocabulary was considered first:

- **Reuse `STEP_LIMIT`**: wrong. It is documented as "loop hit maxSteps without a `stop` finish", and its notice says "Too many steps in one turn". A `length` finish on step 1 never reached the step cap. The user remedy differs (ask something narrower vs. let a multi-step investigation continue) and the operator remedy differs (set an output-token budget or pick a different model vs. tune `IN_APP_AGENT_MAX_STEPS`/wrap-up).
- **`FAILED` + `AGENT_ERROR`**: wrong. The run did not fail; partial text streamed and was persisted. The failure notices tell users to "try again", which would send the same over-long prompt back into the same cap, and it would inflate failure metrics.
- The enum's docblock requires persisted run errors to stay readable, so overloading an existing code would silently change the meaning of historical rows. `InAppAgentRunErrorCode` is the documented extension point (writers are constrained, readers tolerate unknown strings), and all three consumers handle the new member symmetrically with `step_limit`.

## Tests

- `worker/src/features/in-app-agent/runtime/agent.test.ts`: the mocked adapter gained a `beforeComplete` hook so a test can drive `defaultOptions.onIterationComplete` before the stream completes. An `it.each` over (iteration, finishReason) covers: `stop` on step 1 (no truncation, `end({})`), `length` on step 1 (output limit only), `tool-calls` at `IN_APP_AGENT_MAX_STEPS` (step limit only), and `length` at `IN_APP_AGENT_MAX_STEPS` (both flags), asserting both the `onComplete` outcome and the instrumentation `result`. The `length` rows failed before the production change; the `tool-calls` row fails if the output-limit predicate is widened to `!== "stop"`.
- `worker/src/features/in-app-agent/executeInAppAgentRun.test.ts`: `output_limit` is persisted on `SUCCEEDED`, and wins over `step_limit` when both flags are set (Postgres-backed; runs in CI).
- `web/src/features/in-app-agent/lib/backgroundExecutionSession.clienttest.ts`: `getBackgroundRunNotice` returns the warning and `getSettledActivityOutcome` returns `"stopped"` for both `step_limit` and `output_limit`; a plain `SUCCEEDED` run with no error code returns no notice and `"worked"`. The `output_limit` case failed before the lib change.

Checks: `pnpm run lint` (`Tasks: 7 successful, 7 total`), `pnpm run typecheck` (`Tasks: 8 successful, 8 total`), `agent.test.ts` (`Tests 32 passed (32)`), `backgroundExecutionSession.clienttest.ts` (`Tests 37 passed (37)`).

## How to verify locally

1. Point the in-app agent at an OpenAI Responses model (`LANGFUSE_AI_*` env) and start the stack.
2. Trigger a run that overruns the provider's output cap. The simplest lever is a one-off local edit adding `modelSettings: { maxOutputTokens: 64 }` to the Agent `defaultOptions` in `worker/src/features/in-app-agent/runtime/agent.ts`; alternatively, ask for a very long answer (e.g. "list every trace attribute with a paragraph each") with a reasoning model.
3. Open the Assistant, send the message, and let the run settle.
4. Expected: the partial answer is shown, the run status area shows the warning "The assistant had to stop before finishing this answer. The response hit the model's output limit. Send another message to continue.", the activity summary reads "Stopped after …", and the composer stays enabled. In Postgres, `in_app_agent_runs.error_code = 'output_limit'` with `status = 'SUCCEEDED'`.
5. Send a normal short message: no notice, status `SUCCEEDED`, `error_code` null.

The PR surfaces model output-token truncation as a distinct, persisted successful-run outcome rather than leaving users with an unexplained blank turn.

- Adds the shared `output_limit` run error code.
- Detects terminal `length` finish reasons and propagates output- and step-limit metadata through worker completion.
- Persists output truncation with precedence over step-limit truncation and records a dedicated metric.
- Displays a warning and `Stopped` activity outcome for truncated successful runs.
- Adds worker and web coverage for normal, step-limited, output-limited, and combined-limit completions.
</details>

The PR appears safe to merge, with truncation detection, durable persistence, and user-facing handling aligned across the changed paths.

The new output-limit outcome is propagated consistently from the terminal model finish through the worker’s atomic terminal write to the web notice, while ordinary success and existing step-limit behavior remain covered.
</details>

```mermaid
flowchart LR
  A[Model iteration completes] --> B{Finish reason is length?}
  B -- Yes --> C[Mark output truncation]
  B -- No --> D{Step limit truncation?}
  C --> E[Persist SUCCEEDED with output_limit]
  D -- Yes --> F[Persist SUCCEEDED with step_limit]
  D -- No --> G[Persist plain SUCCEEDED]
  E --> H[Web warning and Stopped outcome]
  F --> H
  G --> I[No warning and Worked outcome]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(in-app-agent): surface output-token ..."](https://github.com/langfuse/langfuse/commit/b46048b061837539eae8acf4b0fa92b0345f3c75) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58709192)</sub>

**Context used:**

- Knowledge Base — [In-app agent execution](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/in-app-agent-execution.md)


DavidTraina:
Context: the blank turns that motivated this were traced to a rejected tool set, not a genuine output cut — OpenAI reports it as `incomplete` / `max_output_tokens` with zero usage. #16809 fixes that cause; this PR remains the visibility fix for any `length` finish (for the zero-usage case the notice text is approximate, since the provider's own label is).

PR #16807, opened 2026-08-29
fix(otel): detect third-party spans via attributes and warn on unmasked PII

### LINKED ISSUE #15372: bug(sdk-python): mask is silently skipped for third-party OTel instrumentation spans - PII can reach the API unmasked
## Summary

When a `Langfuse` client is configured with the `mask` hook, spans emitted by third-party OpenTelemetry instrumentation (OpenLIT, OpenInference instrumentors, etc.) are exported **without any masking applied**, while native Langfuse observations are masked correctly. There is no runtime warning, and the framework integration docs that recommend wiring instrumentors into the Langfuse tracer do not mention this, so users can reasonably believe their masking covers everything when it silently does not.

I understand this is by design: the masking docs state that `mask` transforms data at SDK attribute-creation time and does not inspect raw OTel span attributes, and `mask_otel_spans` (export-stage) is the documented recommendation for that case. The problem is that the failure mode is **silent** and the discoverability is poor: nothing in the `mask` path or in the integration guides (e.g. Agno via `openlit.init(tracer=langfuse._otel_tracer, ...)`) signals that `mask` will not cover those spans. For applications handling regulated data (in our case, privileged legal documents under Brazilian LGPD), this is a PII-leak footgun.

## Expected behavior

One of:
- `mask` also applies to third-party spans flowing through the Langfuse span processor, or
- a clear runtime warning when `mask` is configured and non-SDK spans are exported without `mask_otel_spans`, or
- at minimum, an explicit callout in the masking docs and in each third-party integration guide.

## Actual behavior

Native observation content is masked; third-party span attributes (`input.value`, `llm.input_messages.0.message.content`, ...) reach the exporter with their original content. No warning is emitted.

## Reproduction

Fully offline (fake keys + in-memory exporter), only dependency is `langfuse`:

```python
Steps in the "Steps to reproduce session"
```

Output on langfuse 4.14.1:

```
captured spans: ['native-observation', 'third_party.llm.call']
native observation canary in export:   ABSENT (masked) - OK
third-party span canary in export:     [('third_party.llm.call', 'input.value'), ('third_party.llm.call', 'llm.input_messages.0.message.content')]

RESULT: `mask` applied to the native observation but NOT to the third-party instrumentation span. PII reaches the exporter unmasked.
```

## Versions

- langfuse **4.14.1** (also confirmed on 4.14.0)
- Python **3.12.10** (Windows)
- Repro needs no other packages. Originally observed in production instrumentation of **Agno 2.4.7** via **openinference-instrumentation-agno 0.1.27**, wired to the Langfuse tracer provider as the integration docs describe.

## Workaround we use

`mask_otel_spans` at export stage with a **fail-closed allowlist**: every attribute key is redacted unless explicitly allowlisted as structural/metric (model, token counts, cost, level, span kind, ...). Fail-closed matters here because instrumentors emit attribute names that vary by library and version; a denylist silently leaks anything new. Sketch:

```python
def mask_otel_spans(params):
    patches = {}
    for identifier, span in params.spans.items():
        redacted = {
            key: "[REDACTED]"
            for key, value in span.attributes.items()
            if not is_allowlisted_structural(key, value)
        }
        patches[identifier] = OtelSpanPatch(set_attributes=redacted)
    return MaskOtelSpansResult(span_patches=patches)
```

This works well. The issue is purely about the silent gap between `mask` and third-party spans.

## Proposal

1. Emit a one-time warning when `mask` is configured, third-party (non-SDK) spans pass through the Langfuse span processor, and `mask_otel_spans` is not set.
2. Add a prominent callout to the masking docs and to each third-party integration guide (OpenLIT, OpenInference/Agno, etc.): "`mask` does not apply to spans emitted by this instrumentation; use `mask_otel_spans`."

I'm happy to open a docs PR for (2) if that helps.

### Steps to reproduce

"""Minimal repro: `mask` does not apply to third-party OTel instrumentation spans.

Runs fully offline: fake keys + in-memory span exporter (nothing leaves the machine).
Requires only: pip install langfuse
"""
import re

from langfuse import Langfuse
from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult

CANARY_NATIVE = "PII-CANARY-NATIVE-123"
CANARY_THIRD_PARTY = "PII-CANARY-THIRDPARTY-456"

captured = []

class InMemoryExporter(SpanExporter):
    """Captures the spans exactly as they would be sent to the Langfuse API."""

    def export(self, spans):
        captured.extend(spans)
        return SpanExportResult.SUCCESS

    def shutdown(self):
        pass

def masking_function(*, data, **kwargs):
    if isinstance(data, str):
        return re.sub(r"PII-CANARY-\S+", "[MASKED]", data)
    return data

client = Langfuse(
    public_key="pk-lf-0000",
    secret_key="sk-lf-0000",
    mask=masking_function,          # the masking hook most users configure
    span_exporter=InMemoryExporter(),
)

### 1) Native Langfuse observation: mask is expected to apply.
with client.start_as_current_observation(
    name="native-observation",
    input=CANARY_NATIVE,
    output=CANARY_NATIVE,
):
    pass

### 2) Third-party OTel instrumentation span. This is exactly what framework instrumentors (OpenLIT, OpenInference AgnoInstrumentor, etc.) emit when wired to the Langfuse tracer as the integration docs describe (e.g. openlit.init(tracer=langfuse._otel_tracer)).
tracer = client._otel_tracer
with tracer.start_as_current_span("third_party.llm.call") as span:
    span.set_attribute("input.value", CANARY_THIRD_PARTY)
    span.set_attribute("llm.input_messages.0.message.content", CANARY_THIRD_PARTY)

client.flush()
client.shutdown()

### Inspect what would have been exported to the Langfuse API.
native_leaks, third_party_leaks = [], []
for s in captured:
    for key, value in (s.attributes or {}).items():
        if CANARY_NATIVE in str(value):
            native_leaks.append((s.name, key))
        if CANARY_THIRD_PARTY in str(value):
            third_party_leaks.append((s.name, key))

print(f"captured spans: {[s.name for s in captured]}")
print(f"native observation canary in export:   {native_leaks or 'ABSENT (masked) - OK'}")
print(f"third-party span canary in export:     {third_party_leaks or 'ABSENT (masked) - OK'}")

if third_party_leaks and not native_leaks:
    print("\nRESULT: `mask` applied to the native observation but NOT to the "
          "third-party instrumentation span. PII reaches the exporter unmasked.")

Langfuse Cloud

_No response_

_No response_

### Additional information

Langfuse Cloud (US region), but the repro is fully offline (SDK-only, in-memory exporter), so the behavior is independent of the deployment type.

Yes

lorenzozanee:
## What does this PR do?

Fixes #15372

When a third-party instrumentor is wired through `langfuse._otel_tracer` (e.g. `openlit.init(tracer=langfuse._otel_tracer, ...)`), its spans inherit the `langfuse-sdk` instrumentation scope but lack `langfuse.observation.*` attributes. The previous scope-only check misclassified those spans as native SDK spans, dropping `filteredAttributes` from `metadata.attributes` and silencing any masking gap, so content the user expected to be masked could reach ingestion unmasked with no signal.

This change keeps the scope name as a fast path and uses the Langfuse IO attributes (`langfuse.observation.input/output`, `langfuse.trace.input/output`) as ground truth. A `langfuse-sdk` scope without those keys is treated as third-party, `filteredAttributes` are preserved, and a bounded warning plus metric (`langfuse.ingestion.otel.third_party_unmasked_span`, capped at 10 per processor) points to `mask_otel_spans` / ingestion masking docs. Empty native observations without PII-like keys do not warn.

## Type of change

## Mandatory Tasks

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works

This PR refines OpenTelemetry span classification so third-party spans using the Langfuse tracer retain their attributes, and adds bounded telemetry for potentially unmasked content.

- Uses Langfuse IO attributes to distinguish native spans from third-party spans sharing the SDK scope.
- Preserves filtered third-party attributes in normalized metadata.
- Adds a warning and metric capped at ten emissions per processor.
- Adds tests for classification, warning suppression, distinct scopes, and the cap.
</details>

The PR should not merge until the warning distinguishes successfully masked data and deduplicates spans processed by both conversion pipelines.

The new signal reports successfully redacted spans as unmasked and counts each qualifying worker span twice, producing misleading telemetry and prematurely suppressing later warnings.

**Files Needing Attention:** packages/shared/src/server/otel/OtelIngestionProcessor.ts
</details>

```mermaid
flowchart LR
  A[Masked or raw OTLP spans] --> P[Shared processor instance]
  P --> I[processToIngestionEvents]
  P --> E[processToEvent]
  I --> W1[Warning and metric]
  E --> W2[Warning and metric]
  W1 --> C[Shared cap]
  W2 --> C
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/otel/OtelIngestionProcessor.ts:380-407
**Masked spans trigger unmasked alerts**

When ingestion masking successfully redacts a third-party span while preserving keys such as `input.value` or `gen_ai.*`, this check still emits the unmasked-span metric and warning because the processor receives no masking-status flag and examines only attribute names. Operators are therefore told to configure masking even though it is already active, making the new alert unreliable.

### Issue 2
packages/shared/src/server/otel/OtelIngestionProcessor.ts:1078-1083
**Span warnings are double-counted**

The worker sends the same spans through both `processToIngestionEvents` and `processToEvent` on one processor, and each path invokes this warning without per-span deduplication. Every affected span is therefore counted twice, so the ten-warning cap is exhausted after five unique spans and later signals in the job are suppressed.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(otel): detect third-party spans via ..."](https://github.com/langfuse/langfuse/commit/3e0b56198b234cbf3645864f332afb5dd73dafab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58165334)</sub>

**Context used:**

### DIFF
--- a/packages/shared/src/server/otel/OtelIngestionProcessor.ts
+++ b/packages/shared/src/server/otel/OtelIngestionProcessor.ts
@@ -272,6 +272,26 @@ export class OtelIngestionProcessor {
 
   private static readonly METADATA_DROP_WARN_CAP = 10;
   private static readonly ARRAY_ATTRIBUTE_DROP_WARN_CAP = 10;
+  private static readonly THIRD_PARTY_UNMASKED_WARN_CAP = 10;
+
+  private static readonly LANGFUSE_IO_ATTRIBUTES = new Set([
+    LangfuseOtelSpanAttributes.OBSERVATION_INPUT,
+    LangfuseOtelSpanAttributes.OBSERVATION_OUTPUT,
+    LangfuseOtelSpanAttributes.TRACE_INPUT,
+    LangfuseOtelSpanAttributes.TRACE_OUTPUT,
+  ]);
+
+  // PII-prone attribute keys commonly emitted by third-party instrumentors.
+  // Used for bounded warning when a third-party span carries such keys without
+  // server-side masking. This is observability only, not redaction.
+  private static readonly PII_LIKE_ATTRIBUTE_PREFIXES = [
+    "input.value",
+    "output.value",
+    "llm.",
+    "gen_ai.",
+    "traceloop.",
+    "openinference.",
+  ];
 
   private seenTraces: Set<string> = new Set();
   private reportedMetadataDrops = new WeakMap<object, Set<string>>();
@@ -280,6 +300,7 @@ export class OtelIngestionProcessor {
     ReconstructedAttributeDropReason,
     number
   >();
+  private thirdPartyUnmaskedWarnCount = 0;
   private isInitialized = false;
   private traceEventCounts = {
     shallow: 0,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/shared/src/server/otel/OtelIngestionProcessor.ts [hunk 1]
```diff
@@ -312,6 +333,79 @@ export class OtelIngestionProcessor {
...
+    scopeName: string | undefined,
+  ): void {
+    if (!this.hasPiiLikeAttributes(attributes)) return;
+    if (!this.isThirdPartySpan(attributes, scopeName)) return;
+    if (
+      this.thirdPartyUnmaskedWarnCount >=
+      OtelIngestionProcessor.THIRD_PARTY_UNMASKED_WARN_CAP
+    )
+      return;
+    this.thirdPartyUnmaskedWarnCount++;
```
RichardoMrMu, line 391: [P2] Cap resets for every ingestion job

`thirdPartyUnmaskedWarnCount` is scoped to this processor instance, but the worker constructs a new `OtelIngestionProcessor` for each S3 queue job (`worker/src/queues/otelIngestionQueue.ts:545`). A project sending two uploads with ten matching spans therefore emits twenty warnings/metrics, and continued uploads can repeat indefinitely; the current test covers only repeated batches on one processor. That does not provide the one-time/bounded operator signal proposed in #15372. Please move warning suppression to a project-scoped/durable rate-limit boundary (or explicitly define a per-job contract) and cover separate processor/job instances.

PR #16849, opened 2026-08-31
feat(public-api): add PATCH /api/public/models/{modelId}

### LINKED ISSUE #16847: feat(public-api): add PATCH /api/public/models/{modelId} endpoint
### Problem

The public `/api/public/models` API supports `GET`, `POST`, and `DELETE`, but there is no way to update an existing model definition. Customers who need to fix a `matchPattern` typo, swap a tokenizer, or adjust the start date have to `DELETE` and re-`POST`, which loses the model `id`, breaks any references to it, and produces a noisy audit log of "delete + create" instead of one "update" entry.

The internal tRPC `modelsRouter` already supports this in-UI via an upsert (matching on the model `id`); the public API just doesn't expose the same surface.

### Proposed solution

Add `PATCH /api/public/models/{modelId}` for the updateable fields:

- `matchPattern` (string, validated as Postgres regex)
- `tokenizerId` (`"openai" | "claude" | null`)
- `tokenizerConfig` (object | null)
- `startDate` (ISO datetime | null)
- `unit` (`TOKENS | CHARACTERS | MILLISECONDS | SECONDS | REQUESTS | IMAGES`)

Out of scope for this endpoint (intentional, to keep the contract small):

- `modelName` rename — the `(projectId, modelName)` uniqueness constraint means renaming a model is conceptually closer to "create a new model and migrate references" than a property update.
- `pricingTiers` / flat prices — the existing tRPC path replaces the entire pricing tier set, which is a heavier operation; if needed, a follow-up PR can add a dedicated `PUT /api/public/models/{modelId}/pricing-tiers` or extend PATCH to take `pricingTiers` as a replace-array.

### Behavior

- 404 if the model does not exist in the caller's project.
- 400 if `matchPattern` is not a valid Postgres regex.
- 200 with the updated `APIModelDefinition` shape on success.
- Audit log: `action: "update"`, `resourceType: "model"`, with `before` and `after`.
- `clearModelCacheForProject` after a successful update so the new matchPattern / tokenizer takes effect on the next observation ingestion.

### Why this belongs

Mirrors the existing `comments` and `annotation-queues` public API surfaces, which all expose full CRUD. The infrastructure (`prismaToApiModelDefinition`, `auditLog`, `clearModelCacheForProject`, `createAuthedProjectAPIRoute`) is already in place — this is a thin additive change.

### Acceptance criteria

- `PATCH /api/public/models/{modelId}` with `{ "matchPattern": "..." }` updates the row, returns 200 + the updated model, and writes an `"update"` audit log entry.
- `PATCH` with an invalid `matchPattern` returns 400 with a clear error.
- `PATCH` on a non-existent or out-of-project model returns 404.
- `modelName` and pricing fields are rejected at the schema level (`.strict()` on the request body).
- Server test coverage: `model-definitions.servertest.ts` gets a new `describe("PATCH /models/{modelId}")` block covering happy path, 404, 400 on bad regex, and rejection of `modelName`/pricing fields.
- Fern API spec at `fern/apis/server/definition/models.yml` gets the new endpoint and `UpdateModelRequest` type.

### Testing requirements

- Unit / server test: the four cases above.
- Existing tests still pass.

### Backward compatibility

Additive only. No existing route or schema changes.

### Risks

Minimal. The new route is a thin wrapper around an existing pattern; cache invalidation and audit logging are already exercised by the create/delete routes.

Harsh23Kashyap:
Fixes #16847.

## Problem

The public `/api/public/models` API exposes `GET`, `POST`, `PUT` (upsert), and `DELETE`, but there is no narrow partial update. Customers who need to fix a `matchPattern` typo, swap a tokenizer, or adjust a `startDate` currently have two options:

- Use `PUT /models/{id}` (the existing upsert), which requires sending a full `PostModelsV1Body` with all pricing fields and is documented as "create or replace".
- `DELETE` then re-`POST`, which loses the model `id`, breaks references, and produces a noisy "delete + create" audit pair.

Neither matches the natural shape of a partial update.

## Proposed solution

Add `PATCH /api/public/models/{modelId}` for the mutable fields:

- `matchPattern` (string, validated as Postgres regex)
- `tokenizerId` (`"openai" | "claude" | null`)
- `tokenizerConfig` (object | null)
- `startDate` (ISO datetime | null)
- `unit` (`TOKENS | CHARACTERS | MILLISECONDS | SECONDS | REQUESTS | IMAGES`)

Out of scope (intentional, to keep the contract small):

- `modelName` rename — the `(projectId, modelName)` uniqueness constraint means renaming is conceptually closer to "create a new model and migrate references". Use the existing `PUT` upsert for renames.
- `pricingTiers` / flat prices — replacing the pricing set is a heavier operation; use `PUT` (upsert) or `DELETE` + `POST` for now. A future PR can add a dedicated pricing endpoint if needed.

## Behavior

- 404 when the model does not exist in the caller's project (or is built-in / cross-project — IDOR-safe).
- 400 when `matchPattern` is not a valid Postgres regex.
- 400 when the body is empty (no fields provided) or contains disallowed keys (e.g. `modelName`, `inputPrice`) — Zod `.strict()` + `.superRefine()`.
- 200 with the updated `APIModelDefinition` shape on success.
- Audit log: `action: "update"`, `resourceType: "model"`, with `before` and `after`.
- `clearModelCacheForProject` runs after the transaction commits, so the new `matchPattern` / tokenizer takes effect on the next observation ingestion.

## Files

- `web/src/features/public-api/types/models.ts` — new `PatchModelV1Query`, `PatchModelV1Body`, `PatchModelV1Response` Zod schemas. Body is `.strict()` with a `superRefine` that requires at least one field.
- `web/src/features/models/server/publicApiModelService.ts` — new `updateModelForApi` mirroring the create/delete/upsert functions in the same file. Validates `matchPattern` via `isValidPostgresRegex`, finds the model scoped to `(id, projectId)` (404 on miss), runs the prisma `update` inside a ``, audits `before`/`after`, then clears the per-project model cache.
- `web/src/pages/api/public/models/[modelId]/index.ts` — wires `PATCH` into the existing `createAuthedProjectAPIRoute` handler, alongside `GET`, `PUT`, and `DELETE`.
- `fern/apis/server/definition/models.yml` — new `update:` endpoint and `UpdateModelRequest` type.

## Tests

8 new regression tests in `web/src/__tests__/server/model-definitions.servertest.ts`:

1. Happy path — updates `matchPattern`, `tokenizerId`, `tokenizerConfig`, `unit`; `modelName` is preserved; audit log records the update.
2. 404 when the model does not exist.
3. 400 when `matchPattern` is not a valid regex.
4. 400 when the body contains `modelName` or `inputPrice` (Zod strict).
5. 400 when the body is empty (no fields).
6. 404 when patching a built-in model (`projectId = null`).
7. 404 when patching a model owned by a different project (IDOR).
8. Partial update — only the provided field changes; `matchPattern`, `inputPrice`, `outputPrice` are preserved when only `unit` is patched.

## Compatibility

Additive only. No existing route or schema is changed; the new `PATCH` route is gated by the same auth middleware as the other verbs on the path.

## Risk

Minimal. The new route is a thin wrapper around the established `createModelForApi` / `deleteModelForApi` / `upsertModelForApi` pattern in the same file. Cache invalidation and audit logging are already exercised by those.

Adds a project-scoped public endpoint for partially updating mutable model-definition fields, with strict request validation, PostgreSQL-regex validation, audit logging, and cache invalidation.
- Defines the PATCH route and request/response schemas.
- Implements transactional model updates and project ownership checks.
- Adds Fern API definitions and regression coverage for validation, tenancy, auditing, and partial-update behavior.
</details>

The model-conflict path should be fixed before merging because a valid PATCH can currently surface a database uniqueness collision as an internal server error.

Updating startDate or unit can collide with an existing model's composite unique key, and the new service neither detects that state nor translates Prisma's constraint error; the Fern request also omits its runtime non-empty constraint.

**Files Needing Attention:** web/src/features/models/server/publicApiModelService.ts, fern/apis/server/definition/models.yml
</details>

```mermaid
sequenceDiagram
    participant Client
    participant Route as PATCH /models/{modelId}
    participant Service as updateModelForApi
    participant DB as PostgreSQL
    participant Audit
    participant Cache
    Client->>Route: PATCH mutable fields
    Route->>Route: Authenticate and validate body
    Route->>Service: projectId, modelId, input
    Service->>DB: Find project-owned model
    Service->>DB: Update model
    Service->>Audit: Record before and after
    DB-->>Service: Updated model with pricing tiers
    Service->>Cache: Clear project model cache
    Service-->>Client: 200 APIModelDefinition
```
</details>

`````markdown
### Issue 1
web/src/features/models/server/publicApiModelService.ts:304-307
**Unique collisions return 500**

When a PATCH changes `startDate` or `unit` to a combination already used by another same-named model in the project, `tx.model.update` raises an untranslated Prisma uniqueness error, causing a 500 response instead of a validation or conflict response.

### Issue 2
fern/apis/server/definition/models.yml:121-140
**Nonempty constraint is undocumented**

Every Fern request property is optional, so generated types permit an empty update, while the route rejects that request with HTTP 400. Document the requirement that at least one mutable field must be supplied so SDK and API-reference consumers see the runtime constraint.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(public-api): add PATCH /api/public/..."](https://github.com/langfuse/langfuse/commit/5e49676c7b5e48a90924b6543cd8254e9e15048b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58634254)</sub>

**Context used:**

- Knowledge Base — [API and SDK contracts](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse/-/docs/api-and-sdk-contracts.md)

### DIFF
--- a/fern/apis/server/definition/models.yml
+++ b/fern/apis/server/definition/models.yml
@@ -41,6 +41,18 @@ service:
       path-parameters:
         id: string
       response: commons.Model
+    update:
+      method: PATCH
+      docs: |
+        Partially update a model definition. Only the mutable fields are accepted:
+        `matchPattern`, `tokenizerId`, `tokenizerConfig`, `startDate`, `unit`.
+        `modelName` and pricing are not patchable; use `PUT /models/{id}` (upsert)
+        or `DELETE` + `POST /models` for those.
+      path: /models/{id}
+      path-parameters:
+        id: string
+      request: UpdateModelRequest
+      response: commons.Model
     delete:
       method: DELETE
       docs: Delete a model. Cannot delete models managed by Langfuse. You can create your own definition with the same modelName to override the definition though.
@@ -103,6 +115,23 @@ types:
           If omitted, you must provide flat prices instead (inputPrice/outputPrice/totalPrice),
           which will automatically create a single default tier named "Standard".
         type: optional<nullable<list<commons.PricingTierInput>>>
+      tokenizerId:
+        docs: Optional. Tokenizer to be applied to observations which match to this model. See docs for more details.
+        type: optional<nullable<ModelTokenizerId>>
+      tokenizerConfig:
+        docs: Optional. Configuration for the selected tokenizer. Needs to be JSON. See docs for more details.
+        type: optional<nullable<unknown>>
+  UpdateModelRequest:
+    properties:
+      matchPattern:
+        docs: "Regex pattern which matches this model definition to generation.model. Useful in case of fine-tuned models. If you want to exact match, use `(?i)^modelname$`"
+        type: optional<string>
+      startDate:
+        docs: Apply only to generations which are newer than this ISO date.
+        type: optional<nullable<datetime>>
+      unit:
+        docs: Unit used by this model.
+        type: optional<commons.ModelUsageUnit>
       tokenizerId:
         docs: Optional. Tokenizer to be applied to observations which match to this model. See docs for more details.
         type: optional<nullable<ModelTokenizerId>>
--- a/web/src/__tests__/server/model-definitions.servertest.ts
+++ b/web/src/__tests__/server/model-definitions.servertest.ts
@@ -7,6 +7,7 @@ import {
   DeleteModelV1Response,
   GetModelV1Response,
   GetModelsV1Response,
+  PatchModelV1Response,
   PostModelsV1Response,
   PutModelV1Response,
 } from "@/src/features/public-api/types/models";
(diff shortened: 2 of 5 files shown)


Steffen911:
slop

PR #16869, opened 2026-09-01
feat(web): persist environment-filter selection in the URL

### LINKED ISSUE #16868: feat(web): persist environment-filter selection in the URL
### Problem

`useEnvironmentFilter` (used by the project home, dashboards, and users pages) stores the per-user "visible environments" map in `localStorage` only. The selection is therefore tied to a single browser: a user who filters down to two environments, copies the URL, and shares it with a colleague sees those two environments — the colleague opens the same URL and sees the project's default set (everything except `langfuse-*`).

The TODO at `web/src/hooks/useEnvironmentFilter.tsx:8` calls this out:

```ts
// TODO: Add the environment selection to the query params to persist the environment filter when sharing links.
// See useDashboardDateRange for an example of this.
```

### Proposed solution

Add a URL query parameter that mirrors the selection when present, falling back to `localStorage` (and the existing default-visibility rules) when absent. Pattern after `useDashboardDateRange` / `useGlobalDateRange`, which is the established way URL-state is plumbed through the rest of the app.

- New param: `environments=foo,bar` (comma-separated, sorted for stable URLs).
- `?environments` is the source of truth when present and non-empty: exactly the listed envs are visible, even if the localStorage map disagrees.
- `?environments=` (empty) or absent: fall back to the current localStorage + default-visibility behaviour, so the existing per-browser experience is preserved.
- `handleSetVisibilityMap` writes both: the URL (for sharing) and the localStorage map (so the next visit without the URL still shows the last pick).
- When the user clears the selection (empty list), the URL param is removed and the localStorage map records every env as hidden — the existing semantics. The existing `convertSelectedEnvironmentsToFilter` already maps `[]` to "no filter" for the table, which is the legacy behaviour and out of scope to change here.

### Why this matters

The environment filter is the first knob most users reach for when they have multiple staging/prod environments and need to focus. Losing the selection on a tab share is a small but real papercut; the existing date-range hook already established the pattern, and the other filter surfaces in the app follow it. Closing the loop here removes a long-standing TODO and brings the env filter in line with the rest of the page.

### Acceptance criteria

- `useEnvironmentFilter(availableEnv, projectId)` reads `?environments=a,b,c` on mount and treats that as the visible set when the param is present and non-empty.
- When the param is absent or empty, the hook falls back to the localStorage map + the existing `langfuse-` default-visibility rule.
- `setSelectedEnvironments([...])` writes the list to the URL (comma-separated) AND to the localStorage map. Writing an empty list removes the URL param.
- The filter is filtered against `availableEnvironments` so a stale URL that references a deleted environment does not surface it.
- All three call sites (`/[projectId]`, `/[projectId]/dashboards/[dashboardId]`, `/[projectId]/users`) keep working with no caller-side changes — the hook signature is unchanged.
- New `useEnvironmentFilter.clienttest.tsx` covers: URL-authoritative read, localStorage fallback, dual write on set, empty-list removes the URL param, stale URL entries filtered out.

### Backward compatibility

- Existing localStorage state continues to work for users who never set the URL param.
- No changes to any call site — the hook is the abstraction.
- The TODO comment is removed.

### Risks

Minimal. The hook is the only writer and the only reader of the URL param on these three pages. The worst case — a stale URL referencing a deleted env — is handled by filtering against `availableEnvironments`, so a user will see one fewer environment than the URL lists, not a crash.

Harsh23Kashyap:
Fixes #16868.

## Problem

`useEnvironmentFilter` (used by the project home, dashboards, and users pages) stored the per-user "visible environments" map only in `localStorage`. A user who filters down to two environments, copies the URL, and shares it with a colleague sees those two environments — the colleague opens the same URL and sees the project's default set (everything except `langfuse-*`).

The TODO at `web/src/hooks/useEnvironmentFilter.tsx:8` called this out:

```ts
// TODO: Add the environment selection to the query params to persist the environment filter when sharing links.
// See useDashboardDateRange for an example of this.
```

## Proposed solution

Add a `?environments=foo,bar` URL param that mirrors the per-user selection when present; the localStorage map remains the per-browser fallback. Pattern follows the established `useDashboardDateRange` / `useGlobalDateRange` URL-state hooks already in the app.

## Behavior

- `?environments=foo,bar` is the source of truth when present and non-empty. The listed envs are the visible set; the localStorage map is ignored.
- `?environments` empty, absent, or the router pre-hydration: fall back to the localStorage map and the existing `langfuse-*` hidden-by-default rule. The pre-hydration fallback avoids a one-frame flash on shared links.
- `setSelectedEnvironments([...])` writes the URL (comma-separated, each name URL-encoded) and the localStorage map. Writing `[]` removes the URL param. The write is a no-op when the selection has not changed.
- The setter bails on the localStorage write when the available-envs list has not yet resolved (otherwise the previous per-env map would be wiped against an empty list).
- The "initialize new env with default visibility" useEffect no longer runs when the URL is authoritative, so a newly-discovered env does not silently become visible.

## Files

- `web/src/hooks/useEnvironmentFilter.tsx` — read `?environments` via `useQueryParam`; treat as source of truth when present and non-empty; fall back to localStorage + the existing `langfuse-*` hidden-by-default rule. The setter short-circuits when the selection is unchanged and bails on the localStorage write when the available-envs list has not yet resolved. Each env name is URL-encoded.
- `web/src/hooks/useEnvironmentFilter.clienttest.tsx` — 13 regression cases: URL-authoritative read, localStorage fallback, hydration gap, dual write, empty-list removes the URL param, comma-in-name encoding, no-op short-circuit, empty-available guard, deleted-env URL entries filtered, and newly-discovered env not auto-selected when URL is present.

## Compatibility

- The hook signature is unchanged. The three call sites (`/[projectId]`, `/[projectId]/dashboards/[dashboardId]`, `/[projectId]/users`) need no caller-side changes.
- `convertSelectedEnvironmentsToFilter` is unchanged: an empty selection still maps to "no filter / show all" for the table, the pre-existing behavior.

## Risk

Minimal. The hook is the only writer and reader of the URL param on these three pages. The worst case — a stale URL referencing a deleted env — is handled by intersecting with `availableEnvironments`, so a user sees one fewer environment than the URL lists, not a crash.

This PR makes environment-filter selections shareable by mirroring them into an `environments` URL parameter while retaining localStorage as a fallback.
- Adds URL parsing, serialization, hydration handling, and synchronization to `useEnvironmentFilter`.
- Adds client tests covering URL precedence, localStorage fallback, encoding, empty selections, unresolved options, and newly discovered environments.
</details>

The explicit empty-selection state must be represented distinctly in the URL before merging so shared links preserve the sender’s effective environment filter.

Clearing every environment removes the new query parameter, and a recipient consequently falls back to browser-local or default visibility instead of reproducing the sender’s show-all state.

**Files Needing Attention:** web/src/hooks/useEnvironmentFilter.tsx
</details>

```mermaid
flowchart TD
  A[Read environments query parameter] --> B{Router ready and value non-empty?}
  B -->|Yes| C[Decode and intersect with available environments]
  B -->|No| D[Use localStorage and default visibility]
  E[User changes selection] --> F{Selection empty?}
  F -->|No| G[Write explicit URL value]
  F -->|Yes| H[Remove URL parameter]
  G --> I[Update localStorage visibility map]
  H --> I
```
</details>

`````markdown
### Issue 1
web/src/hooks/useEnvironmentFilter.tsx:110-112
**Empty selection loses URL state**

When a user clears all environments, this branch removes the `environments` parameter, so a shared or freshly loaded URL falls back to localStorage or default visibility instead of preserving the sender's explicit show-all selection.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(web): persist environment-filter se..."](https://github.com/langfuse/langfuse/commit/bb537558477f74ff06b9cf345f93734a251c9d69) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58952060)</sub>

### DIFF
--- a/web/src/hooks/useEnvironmentFilter.clienttest.tsx
+++ b/web/src/hooks/useEnvironmentFilter.clienttest.tsx
@@ -1,0 +1,266 @@
+import { act, renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import React from "react";
+
+// Mock the use-query-params library. The mock provides a controllable
+// queryParamStore map that the hook reads from / writes to, mirroring the
+// pattern used by useSidebarFilterState.clienttest.tsx for the same
+// library.
+const queryParamStore = new Map<string, string | undefined>();
+
+vi.mock("use-query-params", () => ({
+  useQueryParam: (key: string) => {
+    const initial = queryParamStore.has(key)
+      ? queryParamStore.get(key)
+      : undefined;
+    const [value, setValue] = React.useState<string | undefined>(initial);
+    const setQueryValue = React.useCallback(
+      (next: string | undefined) => {
+        setValue(next);
+        if (next === undefined || next === "") {
+          queryParamStore.delete(key);
+        } else {
+          queryParamStore.set(key, next);
+        }
+      },
+      [key],
+    );
+    return [value, setQueryValue];
+  },
+  StringParam: {},
+}));
+
+// Mock useRouter. isReady defaults to true so the hook treats the URL as
+// authoritative in tests; individual tests can flip it via setRouterReady.
+let routerIsReady = true;
+vi.mock("next/router", () => ({
+  useRouter: () => ({ isReady: routerIsReady }),
+}));
+
+import { useEnvironmentFilter } from "@/src/hooks/useEnvironmentFilter";
+
+const PROJECT = "test-project";
+const PROJECT_KEY = `langfuse-environment-visibility-${PROJECT}`;
+
+describe("useEnvironmentFilter", () => {
+  beforeEach(() => {
+    queryParamStore.clear();
+    localStorage.clear();
+    routerIsReady = true;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("falls back to localStorage + default visibility when URL has no envs", () => {
+    localStorage.setItem(
+      PROJECT_KEY,
+      JSON.stringify({ default: true, langfuse: false, prod: true }),
+    );
+
+    const { result } = renderHook(() =>
+      useEnvironmentFilter(["default", "langfuse", "prod"], PROJECT),
+    );
+
+    // "langfuse-*" is hidden by default; "default" and "prod" explicitly true.
+    expect(result.current.selectedEnvironments.sort()).toEqual([
+      "default",
+      "prod",
+    ]);
+  });
+
+  it("URL is authoritative when present: hides a localStorage-visible env", () => {
+    localStorage.setItem(
+      PROJECT_KEY,
+      JSON.stringify({ prod: true, staging: true }),
+    );
+    queryParamStore.set("environments", "prod");
+
+    const { result } = renderHook(() =>
+      useEnvironmentFilter(["prod", "staging"], PROJECT),
+    );
+
+    // URL says only "prod" — localStorage's "staging" is overridden.
+    expect(result.current.selectedEnvironments).toEqual(["prod"]);
+  });
+
+  it("URL is authoritative when present: shows a localStorage-hidden env", () => {
+    localStorage.setItem(
+      PROJECT_KEY,
+      JSON.stringify({ prod: true, staging: false }),
+    );
+    queryParamStore.set("environments", "prod,staging");
+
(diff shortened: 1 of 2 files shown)


Steffen911:
slop

PR #16891, opened 2026-09-01
docs(ai-gateway): add V1 RFC

hassiebp:
🟡 **Live preview: [pr-16891 app preview](https://pr-16891.preview.langfuse.com)**

## Summary

- define the concise V1 plan for the Langfuse AI Gateway
- use native OpenAI and Anthropic API namespaces with one organization connection per namespace
- introduce a controlled provider registry and upstream-authoritative model discovery
- define Gateway virtual keys, project attribution, cached resolution, telemetry modes, and ingestion contexts
- scope Cloud and self-hosted deployment around the same Rust service
- explicitly defer translation, arbitrary origins, Bedrock/Vertex transports, routing, and policy enforcement

## Impacted packages

- specification only: `specs/llm-gateway-rfc.md`

## Verification

- `git diff --check origin/main...HEAD`
- verified `origin/main...HEAD` contains exactly one Markdown file

### DIFF
--- a/specs/llm-gateway-rfc.md
+++ b/specs/llm-gateway-rfc.md
@@ -8,14 +8,14 @@
 
 Langfuse will ship a BYOK AI Gateway that lets Claude Code, Codex, and general OpenAI or Anthropic applications use Langfuse as their LLM API base URL. The gateway executes requests with organization-managed LLM credentials, streams the result back to the client, and records usage, cost, latency, and optionally request/response content in Langfuse without client-side hooks.
 
-V1 launches on Langfuse Cloud. The same container and contracts ship for self-hosting as a fast follow.
+V1 launches on Langfuse Cloud and self-hosted at the same time. Both deployment models use the same container and contracts.
 
 V1 includes:
 
-- Anthropic Messages, OpenAI Chat Completions, OpenAI Responses, and their native auxiliary endpoints used by Claude Code and Codex.
-- OpenAI and Anthropic BYOK connections.
+- Anthropic Messages, OpenAI Chat Completions, OpenAI Responses, Claude Code's native Bedrock and Vertex protocols, and their required auxiliary endpoints.
+- OpenAI, Anthropic, Bedrock, and Vertex BYOK connections.
 - Native protocol passthrough when the client and upstream protocols match. Incompatible client/upstream protocol pairs are rejected before provider dispatch.
-- Gateway virtual keys, deterministic model-to-connection resolution, and automatic Langfuse telemetry.
+- Gateway virtual keys, deterministic protocol-to-connection resolution, and automatic Langfuse telemetry.
 
 V1 does not include cross-provider protocol translation, inference resale, multiple routing candidates, fallbacks, gateway retries, rate limits, spend limits, or guardrails.
 
@@ -23,57 +23,51 @@ V1 does not include cross-provider protocol translation, inference resale, multi
 
 Langfuse currently observes coding agents through client-side hooks. This works for individuals but is difficult to deploy and govern across organizations with hundreds of developers. A gateway gives platform teams one centrally managed integration point for tracing, cost attribution, credentials, and future runtime policy.
 
-Cross-provider translation would strengthen that value proposition by letting Codex use Anthropic models or Claude Code use OpenAI models without changing their wire protocol. However, OpenAI Responses and Anthropic Messages are not equivalent contracts. Translation introduces state, tool-lifecycle, streaming, media, and provider-owned artifact semantics whose failures can be silent and agent-breaking. V1 therefore prioritizes reliable native execution and treats translation as a separate interoperability product with its own compatibility contract and release criteria.
+Cross-provider translation would strengthen that value proposition by letting Codex use Anthropic models or Claude Code use OpenAI models without changing their wire protocol. However, OpenAI Responses and Anthropic Messages are not equivalent contracts. Translation introduces state, tool-lifecycle, streaming, media, and provider-owned artifact semantics whose failures can be silent and agent-breaking. V1 therefore prioritizes reliable native execution. Translation will only be reconsidered if strong measured demand justifies a separate compatibility product and its permanent maintenance burden.
 
 ## Product Scope
 
-### Public endpoints
+### Public protocol namespaces
 
-| Endpoint | V1 support |
+Each native protocol has an explicit namespace. The namespace determines the upstream wire contract and response format without inspecting headers or negotiating a schema.
+
+| Namespace and endpoint family | V1 support |
 | --- | --- |
-| `POST /v1/messages` | Required: Anthropic Messages and Claude Code |
-| `POST /v1/messages/count_tokens` | Native Anthropic pass-through; optional to Claude Code |
-| `POST /v1/chat/completions` | Required: native OpenAI |
-| `POST /v1/responses` | Required: native OpenAI |
-| `POST /v1/responses/compact` | Required for native OpenAI/Codex |
-| `GET /v1/models` | Required: authenticated gateway-owned catalog |
+| `POST /anthropic/v1/messages` | Native Anthropic Messages and Claude Code |
+| `POST /anthropic/v1/messages/count_tokens` | Native Anthropic token counting |
+| `GET /anthropic/v1/models` | Native Anthropic model discovery through the selected connection |
+| `POST /openai/v1/chat/completions` | Native OpenAI Chat Completions |
+| `POST /openai/v1/responses` | Native OpenAI Responses |
+| `POST /openai/v1/responses/compact` | Native OpenAI response compaction for Codex |
+| `GET /openai/v1/models` | Native OpenAI model discovery through the selected connection |
+| `/bedrock/...` | Claude Code's native Bedrock InvokeModel, streaming, and token-counting paths |
+| `/vertex/...` | Claude Code's native Vertex `rawPredict`, streaming, and token-counting paths |
+
+Clients configure the namespace as part of their base URL. The tradeoff is that Langfuse exposes provider-specific base URLs instead of one universal URL; this is acceptable because V1 does not route or translate between protocols, and the namespace makes response/error/model-discovery formats unambiguous. A future unified root API may be added as an alias or separate routing product without breaking these native endpoints.
 
 ### Protocol matrix
 
-The requested model deterministically selects one upstream adapter and LLM Connection. V1 only executes matching client and upstream protocols.
+The protocol namespace deterministically selects one upstream adapter and LLM Connection. V1 only executes matching client and upstream protocols.
 
-| Client protocol | OpenAI upstream | Anthropic upstream |
+| Client protocol | Upstream adapter | Behavior |
 | --- | --- | --- |
-| Anthropic Messages | Not supported | Native passthrough |
-| OpenAI Chat Completions | Native passthrough | Not supported |
-| OpenAI Responses | Native passthrough | Not supported |
+| Anthropic Messages | Anthropic | Native HTTP/SSE passthrough |
+| OpenAI Chat Completions | OpenAI | Native HTTP/SSE passthrough |
+| OpenAI Responses | OpenAI | Native HTTP/semantic-event passthrough |
+| Claude Code Bedrock | Bedrock | Native InvokeModel HTTP/AWS event-stream passthrough |
+| Claude Code Vertex | Vertex | Native `rawPredict` HTTP/SSE passthrough |
 
-An incompatible model/protocol pair returns a clear client error before provider dispatch. The gateway never silently translates, drops, or approximates provider-specific semantics in V1.
+An unavailable or incompatible protocol namespace returns a clear client error before provider dispatch. The gateway never silently translates, drops, or approximates provider-specific semantics in V1. Bedrock and Vertex support means accepting Claude Code's native provider protocols; it does not mean translating `/anthropic/v1/messages` into those protocols.
 
 ### Model discovery
 
-`GET /v1/models` always returns one OpenAI-compatible list:
+Model discovery is namespaced and upstream-authoritative:
 
-```json
-{
-  "object": "list",
-  "data": [
-    {
-      "id": "claude-sonnet-4-5",
-      "object": "model",
-      "created": 0,
-      "owned_by": "anthropic",
-      "display_name": "Claude Sonnet 4.5"
-    }
-  ]
-}
-```
+- `/anthropic/v1/models` forwards to the selected Anthropic connection and returns its native response unchanged.
+- `/openai/v1/models` forwards to the selected OpenAI connection and returns its native response unchanged.
+- Claude Code does not use model discovery for its native Bedrock or Vertex protocols.
 
-The [Claude Code gateway contract](https://code.claude.com/docs/en/llm-gateway-protocol#request-and-response) reads `data[].id` and optional `data[].display_name`, so this shape supports both OpenAI clients and Claude Code without switching response schemas based on a header. It is not full Anthropic Models API compatibility. If that becomes necessary, add an explicit `/anthropic` protocol namespace rather than changing the representation of the root endpoint.
-
-The endpoint requires a Gateway virtual key and returns only models enabled for that key's organization and resolved project policy. The list is generated from Langfuse configuration and never fetched synchronously from an upstream provider. A listed model maps to exactly one upstream adapter and connection, and unlisted models are rejected. This is deterministic destination selection, not routing: there is no candidate set, policy evaluation, or fallback.
-
-Claude Code only discovers IDs containing `claude` or `anthropic`. Because V1 does not translate Messages to OpenAI, the catalog must not make an OpenAI destination appear Messages-compatible by assigning it a Claude-like alias.
+Langfuse does not maintain, merge, normalize, or enforce a gateway model catalog in V1. The selected connection determines which models are available, and the provider remains responsible for accepting or rejecting the request's native model identifier. The discovery path is authenticated and uses the same connection resolution as inference. It is not cached initially; a short response cache may be added later if upstream latency or rate limits require it.
 
 ## Architecture
 
@@ -82,8 +76,8 @@ flowchart LR
     C["Claude Code, Codex, or application"] -->|"native API request"| G["Rust AI Gateway"]
     G -->|"cache miss: resolve"| W["Langfuse Web control plane"]
     W --> P[("Postgres and existing auth/config")]
-    G -->|"native request"| O["OpenAI or Anthropic"]
-    O -->|"JSON or SSE"| G
+    G -->|"native request"| O["OpenAI, Anthropic, Bedrock, or Vertex"]
+    O -->|"native response stream"| G
     G -->|"native response"| C
     G -.->|"batched telemetry and media"| W
     W -.-> I["Existing Langfuse ingestion"]
@@ -108,19 +102,18 @@ Clients cannot submit or override a project ID. Changing the organization defaul
 
 ### LLM Connections and destinations
 
-LLM Connections become organization-level resources. For V1, an administrator explicitly selects one default connection for each upstream adapter, initially OpenAI and Anthropic. The design must later allow multiple candidates and fallbacks without changing the connection resource, but V1 always resolves one destination.
+LLM Connections become organization-level resources. For V1, an administrator explicitly selects one default connection for each enabled upstream adapter: OpenAI, Anthropic, Bedrock, and Vertex. The selection contract may later return multiple candidates for routing or fallback, but V1 always resolves exactly one connection.
 
-The model catalog records which adapter owns each model. A request resolves:
+A request resolves without consulting a model catalog:
 
 ```text
-Gateway virtual key + ingress protocol + model
+Gateway virtual key + protocol namespace
   -> organization and project
-  -> upstream adapter
   -> one default LLM Connection
-  -> native execution
+  -> native execution with the request's model unchanged
 ```
 
-Official OpenAI and Anthropic origins are supported first. Custom base URLs require save-time, use-time, connection-time, DNS, and redirect validation plus strict credential stripping. They only enter V1 if that security work is completed for both Cloud and self-hosted deployments.
+Official provider origins and custom LLM Connection base URLs are supported. Custom origins require the Rust equivalent of Langfuse's secure LLM fetch boundary: scheme and URL checks when saved and used, DNS/IP validation when connecting, validation of every redirect, and strict credential/header stripping. Cloud blocks private, loopback, link-local, and metadata destinations. Self-hosters may explicitly allow private destination ranges needed for their own inference infrastructure.
 
 ### Telemetry policy
 
@@ -129,14 +122,14 @@ An organization chooses one gateway telemetry mode:
 | Mode | Customer telemetry |
 | --- | --- |
 | `full` | Input, output, tools, media references, metadata, usage, and cost |
-| `metadata` | Model, status, timing, attribution, usage, and cost; no content |
+| `usage` | Model, status, timing, attribution, usage, and cost; no content |
 | `none` | No customer request trace or cost record |
 
-`full` is the default. `none` does not disable content-free service health, security, and capacity telemetry. Rate or spend enforcement may later require at least metadata mode.
+`full` is the default. `none` does not disable content-free service health, security, and capacity telemetry. Rate or spend enforcement may later require at least `usage` mode.
 
-## Cached Resolution and Internal Authentication
+## Cached Resolution and Gateway-to-Web Authentication
 
-The gateway must not call Web for every inference request. It caches a resolved execution configuration in memory by a fingerprint of the Gateway virtual key, ingress protocol, and model.
+The gateway must not call Web for every inference request. It caches a resolved execution configuration in memory by a keyed fingerprint of the Gateway virtual key and protocol namespace.
 
 ```mermaid
 sequenceDiagram
@@ -151,7 +144,7 @@ sequenceDiagram
     else Cache miss or hard expiry
         Gateway->>Web: Resolve service token + virtual key
         Web-->>Gateway: Destination, telemetry policy, sealed context
-        Gateway->>Gateway: Cache for at most 5 minutes
+        Gateway->>Gateway: Cache for at most 60 seconds
     end
     Gateway->>Provider: Execute once
     Provider-->>Gateway: Response stream
@@ -159,7 +152,7 @@ sequenceDiagram
     Gateway-->>Web: Batched telemetry + sealed context
 ```
 
-### Internal authentication envelope
+### Service authentication envelope
 
 Gateway-to-Web requests use the standard `Authorization` header so proxies and observability systems are most likely to redact it. The credentials use one deliberately simple, versioned encoding:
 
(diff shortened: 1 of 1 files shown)


specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+- Status: Draft
+- Linear: [LFE-15204](https://linear.app/clickhouse/issue/LFE-15204/ai-gateway-v1-scope)
+- Last updated: 2026-09-01
+
+## Decision
+
+Langfuse will ship a BYOK AI Gateway that lets Claude Code, Codex, and general OpenAI or Anthropic applications use Langfuse as their LLM API base URL. The gateway executes requests with organization-managed LLM credentials, streams the result back to the client, and records usage, cost, latency, and optionally request/response content in Langfuse without client-side hooks.
+
+V1 launches on Langfuse Cloud. The same container and contracts ship for self-hosting as a fast follow.
```
hassiebp: it launches on self hosted at the same time

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+      "object": "model",
+      "created": 0,
+      "owned_by": "anthropic",
+      "display_name": "Claude Sonnet 4.5"
+    }
+  ]
+}
+```
+
+The [Claude Code gateway contract](https://code.claude.com/docs/en/llm-gateway-protocol#request-and-response) reads `data[].id` and optional `data[].display_name`, so this shape supports both OpenAI clients and Claude Code without switching response schemas based on a header. It is not full Anthropic Models API compatibility. If that becomes necessary, add an explicit `/anthropic` protocol namespace rather than changing the representation of the root endpoint.
```
hassiebp: do you think it makes sense to add an explicit /openai or /anthropic protocol namespace from the get go? this makes it very clear on what the response format should be. examine tradeoffs

hassiebp: i think cloudflare had it this way and changed to root level later

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+The endpoint requires a Gateway virtual key and returns only models enabled for that key's organization and resolved project policy. The list is generated from Langfuse configuration and never fetched synchronously from an upstream provider. A listed model maps to exactly one upstream adapter and connection, and unlisted models are rejected. This is deterministic destination selection, not routing: there is no candidate set, policy evaluation, or fallback.
```
hassiebp: the model list should be always fetched upstream. the connection decides what models are available

hassiebp: i want to avoid managing model lists in langfuse

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+```text
+Gateway virtual key + ingress protocol + model
+  -> organization and project
+  -> upstream adapter
+  -> one default LLM Connection
+  -> native execution
+```
+
+Official OpenAI and Anthropic origins are supported first. Custom base URLs require save-time, use-time, connection-time, DNS, and redirect validation plus strict credential stripping. They only enter V1 if that security work is completed for both Cloud and self-hosted deployments.
```
hassiebp: Is that really true? we could go and build a similar thing as the secureLlmFetch for rust . if DNS record changes between save-time and later, that's up to the user, no?

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Telemetry policy
+
+An organization chooses one gateway telemetry mode:
+
+| Mode | Customer telemetry |
+| --- | --- |
+| `full` | Input, output, tools, media references, metadata, usage, and cost |
+| `metadata` | Model, status, timing, attribution, usage, and cost; no content |
```
hassiebp: do we have a better name than `metadata` ? this term is already taken for e.g. observation metadata

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+| `none` | No customer request trace or cost record |
+
+`full` is the default. `none` does not disable content-free service health, security, and capacity telemetry. Rate or spend enforcement may later require at least metadata mode.
+
+## Cached Resolution and Internal Authentication
+
+The gateway must not call Web for every inference request. It caches a resolved execution configuration in memory by a fingerprint of the Gateway virtual key, ingress protocol, and model.
```
hassiebp: can't we keep the fingerprint even on the level of virtual key + ingress protocol?

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+    participant Web
+    participant Provider
+
+    Client->>Gateway: Request and Gateway virtual key
+    alt Valid execution config in memory
+        Gateway->>Gateway: Reuse configuration
+    else Cache miss or hard expiry
+        Gateway->>Web: Resolve service token + virtual key
+        Web-->>Gateway: Destination, telemetry policy, sealed context
+        Gateway->>Gateway: Cache for at most 5 minutes
```
hassiebp: let's start with 60 seconds

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+Telemetry and media submissions replace `key` with the sealed context:
+
+```http
+Authorization: Langfuse-Gateway <base64url({"v":1,"service":"...","context":"..."})>
+```
+
+The envelope avoids custom authorization-parameter parsing while keeping both credentials in one redacted header. Web validates the shared gateway service token before authenticating the customer key or opening the context. The header must not be logged and is stripped before upstream execution.
+
+`Langfuse-Gateway` distinguishes this private service exchange from ordinary customer `Bearer` authentication. Parsing is one scheme split, one bounded base64url decode, and strict JSON-schema validation. Base64url is encoding, not protection; HTTPS and the two credentials provide the security. Implementations reject oversized envelopes, unsupported versions, unknown or duplicate fields, and invalid combinations. Web, Gateway, and the self-hosting guide must explicitly suppress the full `Authorization` field from logs because arbitrary proxies cannot be assumed to do so.
```
hassiebp: drop this line, too much detail

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+    Gateway-->>Client: Forward stream
+    Gateway-->>Web: Batched telemetry + sealed context
+```
+
+### Internal authentication envelope
+
+Gateway-to-Web requests use the standard `Authorization` header so proxies and observability systems are most likely to redact it. The credentials use one deliberately simple, versioned encoding:
+
+```http
+Authorization: Langfuse-Gateway <base64url({"v":1,"service":"...","key":"..."})>
```
hassiebp, line 158: also note what is what 
service -> gateway token
key -> provided virtual key

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+Telemetry and media submissions replace `key` with the sealed context:
+
+```http
+Authorization: Langfuse-Gateway <base64url({"v":1,"service":"...","context":"..."})>
```
hassiebp: be clearer on what the sealed context contains

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+The envelope avoids custom authorization-parameter parsing while keeping both credentials in one redacted header. Web validates the shared gateway service token before authenticating the customer key or opening the context. The header must not be logged and is stripped before upstream execution.
+
+`Langfuse-Gateway` distinguishes this private service exchange from ordinary customer `Bearer` authentication. Parsing is one scheme split, one bounded base64url decode, and strict JSON-schema validation. Base64url is encoding, not protection; HTTPS and the two credentials provide the security. Implementations reject oversized envelopes, unsupported versions, unknown or duplicate fields, and invalid combinations. Web, Gateway, and the self-hosting guide must explicitly suppress the full `Authorization` field from logs because arbitrary proxies cannot be assumed to do so.
+
+The shared service token supports current/next rotation and is distributed through Secrets Manager on Cloud or one shared Kubernetes Secret for self-hosting. The internal endpoints are also private at the network layer.
```
hassiebp: we don't add any network layer restrictions for v1, would be too much config for self hosters

hassiebp: also there are no internal endpoints, the resolve endpoint is public and the ingestion / media endpoints are also publis

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Execution configuration and sealed context
+
+On a cache miss, Web returns:
+
+- Organization, project, key, and attribution identifiers.
+- Ingress and upstream protocols plus the resolved native model.
```
hassiebp: what do you mean by resolved native model? we will simply forward the model unchanged to the upstream

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+- One connection ID, official origin, safe headers, and provider credential.
+- Telemetry mode.
+- A cache hard-expiry timestamp.
+- An opaque sealed execution context.
+
+The provider credential exists only in gateway memory and is held in a secret-safe type. The cache key uses a keyed cryptographic digest of the Gateway virtual key and is never logged. The default hard cache TTL is five minutes, matching Langfuse's current API-key cache window. An expired entry is never used; resolution failure without a valid entry fails closed. This makes five minutes the explicit maximum propagation window for revocation or configuration changes in V1.
```
hassiebp: cache period should be at most 60 seconds. control plane cache period is longer but has invalidation logic, gateway doesn't

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+Web seals the context with a Web-only authenticated-encryption key. The gateway treats it as opaque. A longer context lifetime, initially 24 hours, authorizes only delayed telemetry/media delivery; it can never authorize a new inference request or extend the five-minute execution-config cache. There is no grant table, Redis record, refresh flow, or JWKS subsystem.
+
+## Data Plane
+
+### Request lifecycle
+
+1. **Classify and bound:** Select the ingress protocol, normalize the client credential, reject an invalid method/content type or declared oversize body, and read the body once under a hard byte limit.
+2. **Authenticate and resolve:** Extract the model, load or refresh the cached execution configuration, and use the catalog's single adapter/connection mapping. No client-controlled origin, connection, or project is accepted.
+3. **Protect:** Apply the native protocol's minimal boundary checks plus header and media policy. Reject an incompatible client/upstream protocol pair.
```
hassiebp: is that necessary? we can let upstream handle that

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+4. **Prepare:** Preserve the native body and apply only the resolved origin, model placement, provider authentication, and safe header adjustments required by the upstream transport.
```
hassiebp: why model placement? no need for that

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+V1 therefore supports Claude Code with Anthropic destinations and Codex/OpenAI clients with OpenAI-compatible destinations. Codex-to-Claude and Claude-Code-to-OpenAI remain valuable fast-follow use cases, but they require a separate RFC defining one named compatibility profile, explicit rejections, conversation/provider pinning, direct pairwise adapters, translator version telemetry, and real-client multi-turn conformance tests.
+
+This decision is about sequencing, not rejecting interoperability. The likely first post-V1 profile is OpenAI Responses to Anthropic Messages for Codex, limited initially to complete stateless history, text/images, custom function tools/results, and streaming/non-streaming execution. Stateful Responses, hosted tools, provider file IDs, encrypted reasoning translation, and cross-provider fallback after a conversation starts would be rejected.
```
hassiebp: remove that line and reframe it to we will ship this only if there is massive user need that justifies the maintenance overhead and scope explosion

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Telemetry and Media
+
+The gateway builds one Langfuse generation per provider execution, including project/key attribution, protocol pair, model and connection, timing and time-to-first-token, provider request ID, terminal outcome, usage/cache usage, and cost inputs. Claude Code session and agent headers may be recorded for grouping but are never treated as user identity. Content inclusion follows the resolved telemetry mode and explicit byte bounds.
+
+Telemetry is accumulated into timeout-or-size bounded batches and submitted to a private gateway ingestion endpoint. Web validates the sealed context, derives the project, and reuses the existing S3/queue/OTLP ingestion path. Media uses a corresponding private upload endpoint. The gateway does not receive storage or queue credentials.
```
hassiebp: we will reuse the regular otel ingestion endpoint, not a private gateway ingestion endpoint

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+V1 does not add a separate ClickHouse gateway-logs table. Langfuse traces/generations remain the customer-facing request record; Datadog metrics and sanitized service logs cover operational health. A separate durable execution ledger should only be introduced when spend enforcement, settlement, or audit guarantees require semantics that best-effort observability cannot provide.
+
+## Deployment and Reliability
+
+### Langfuse Cloud
+
+Deploy `ai-gateway` as an independently scalable Rust ECS/Fargate service in every existing regional environment. Reuse the regional VPC, private subnets, NAT egress, ALB/WAF, Secrets Manager, deployment pipeline, and monitoring foundations while giving the gateway dedicated tasks, target group, security group, IAM roles, and autoscaling.
+
+The public edge uses a gateway hostname and streaming-safe idle/drain settings. Gateway-to-Web resolution, telemetry, and media endpoints are private. Scale primarily on active streams, with CPU, memory, bytes in flight, connection rate, and telemetry-queue pressure as guardrails. Deployments require readiness, graceful draining, a maximum stream duration, and progressive regional rollout.
```
hassiebp: gateway to web resolution is currently public, or can we keep this in the private subnet easily?

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Self-hosted fast follow
```
hassiebp: self hosted will be shipped at same time as cloud

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+## Failure Semantics
+
+- Authentication, expired-cache resolution, destination selection, protocol matching, and boundary validation fail closed before provider dispatch.
+- Inference POSTs are never retried by the gateway.
+- Provider statuses and errors are passed through unchanged.
+- After provider dispatch, telemetry and media failures fail open.
+- Capture, batches, and queues are byte- and count-bounded.
+- Saturation sheds new work or telemetry according to explicit limits; it never permits unbounded memory growth.
+
+## Delivery Milestones
```
hassiebp, line 208: adjust milestones more on control plane / data plane / cloud / self host readiness

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,320 @@
...
+- Golden native request/response/error fixtures and exact SSE pass-through behavior.
+- Strict negative tests for incompatible client protocol and destination combinations.
+- Credential/header leak, tenant isolation, revocation-window, SSRF, redirect, and custom-origin tests.
+- Long-lived stream, slow-client, disconnect, deployment-drain, Web outage, provider failure, and telemetry saturation tests.
+- Measured gateway overhead and capacity on production-like multi-core containers.
+
+## Open Questions
+
+- Are custom public HTTPS LLM Connection base URLs included in V1 after the required transport-security spike, or deferred?
+- Are near-native Bedrock Messages and Vertex Messages transport adapters included in V1 after provider conformance, or deferred?
```
hassiebp: yeah we should ship with bedrock messages and vertex messages support

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
+# RFC: Langfuse AI Gateway V1
+
+- Status: Draft
+- Linear: [LFE-15204](https://linear.app/clickhouse/issue/LFE-15204/ai-gateway-v1-scope)
```
hassiebp: Please remove the internal ticket reference from this public RFC and from the PR body. Repository guidance requires published content to be self-contained and excludes internal ticket IDs/URLs; the authorization dependency can be described directly without the private link.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Decision
+
+Langfuse will ship a BYOK AI Gateway that lets Claude Code, Codex, and general OpenAI or Anthropic applications use Langfuse as their LLM API base URL. The gateway executes requests with organization-managed LLM credentials, streams the result back to the client, and records usage, cost, latency, and optionally request/response content in Langfuse without client-side hooks.
+
+V1 launches on Langfuse Cloud and self-hosted at the same time. Both deployment models use the same container and contracts.
+
+V1 includes:
+
+- Anthropic Messages, OpenAI Chat Completions, OpenAI Responses, Claude Code's native Bedrock and Vertex protocols, and their required auxiliary endpoints.
```
hassiebp: Let us narrow the V1 adapter scope here: keep Anthropic Messages/token counting/models and OpenAI Chat Completions/Responses/compact/models, but defer Bedrock and Vertex. Those two add distinct cloud authentication, path/error handling, and AWS binary event-stream work. Please update the protocol matrix, architecture diagram, milestones, and launch gates consistently. Model discovery should remain upstream-authoritative and should not create a customer generation.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+A **Gateway virtual key** is a Langfuse-issued credential accepted only by the gateway. It is named, reveal-once, attributable, expirable, rotatable, revocable, and authorized with an explicit `gateway:invoke` permission preset.
+
+Only organization administrators create Gateway virtual keys in V1. The implementation extends the authorization and key lifecycle planned in the [granular API permissions RFC](https://linear.app/clickhouse/document/rfc-granular-api-permissions-a-unified-authorization-core-for-langfuse-04e83c1c436b); the gateway must not create a parallel permissions system.
+
+Each execution maps to exactly one Langfuse project. Project selection is:
+
+1. A project override fixed when the key is created, if the organization permits it.
+2. Otherwise, the organization's current default gateway project.
+
+Clients cannot submit or override a project ID. Changing the organization default affects executions after their cached configuration expires.
```
hassiebp: Organization-level connections make sense for sharing one provider credential across projects. Project attribution should still be immutable per Gateway key: the organization default can prefill the choice at key creation, but resolution should not read a mutable default on every cache miss. Otherwise changing the default silently moves an existing key/session to a different telemetry project after at most 60 seconds. V1 can use the organization default connection for each adapter; project-level connection overrides can follow later.

hassiebp (commented):
Focused V1 design changes: keep Chat Completions, model discovery, and organization-level shared connections; defer Bedrock and Vertex; make project attribution immutable; simplify the ingestion grant; default to usage telemetry; make custom origins a fast follow; and make the deployment/security gates explicit.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+A request resolves without consulting a model catalog:
+
+```text
+Gateway virtual key + protocol namespace
+  -> organization and project
+  -> one default LLM Connection
+  -> native execution with the request's model unchanged
+```
+
+Official provider origins and custom LLM Connection base URLs are supported. Custom origins require the Rust equivalent of Langfuse's secure LLM fetch boundary: scheme and URL checks when saved and used, DNS/IP validation when connecting, validation of every redirect, and strict credential/header stripping. Cloud blocks private, loopback, link-local, and metadata destinations. Self-hosters may explicitly allow private destination ranges needed for their own inference infrastructure.
```
hassiebp: Let us ship V1 with official Anthropic/OpenAI origins only and make custom base URLs the first fast follow. Please keep an upstream-target abstraction so this is additive. The follow-up must implement save-time, use/connect-time, and redirect-time validation, cross-origin credential stripping, self-hosted allowlists, and negative SSRF tests before the surface is enabled.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+An organization chooses one gateway telemetry mode:
+
+| Mode | Customer telemetry |
+| --- | --- |
+| `full` | Input, output, tools, media references, metadata, usage, and cost |
+| `usage` | Model, status, timing, attribution, usage, and cost; no content |
+| `none` | No customer request trace or cost record |
+
+`full` is the default. `none` does not disable content-free service health, security, and capacity telemetry. Rate or spend enforcement may later require at least `usage` mode.
```
hassiebp: Please switch the default to usage. Full content capture should require explicit organization opt-in because coding-agent traffic can contain source code, tool arguments, and secrets. It would also be useful to record the selected mode on each observation so usage-only records are distinguishable from incomplete capture.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+- `service` is the shared Gateway-to-Web service token.
+- `key` is the caller-provided Gateway virtual key used to resolve an execution configuration.
+- `context` is Web's sealed execution and ingestion context, used instead of `key` for telemetry and media.
+
+Web validates the shared gateway service token before authenticating the customer key or opening the context. The header must not be logged and is stripped before upstream execution.
+
+`Langfuse-Gateway` distinguishes this service exchange from ordinary customer `Bearer` authentication. Base64url is encoding, not protection; HTTPS and the two credentials provide the security. Web, Gateway, and the self-hosting guide must suppress the full `Authorization` field from logs.
+
+The shared service token supports current/next rotation and is distributed through Secrets Manager on Cloud or one shared Kubernetes Secret for self-hosting. Resolution, telemetry ingestion, and media upload are public authenticated Langfuse endpoints in V1; no private-network connectivity is required.
```
hassiebp: Use current/previous rather than current/next. Verifiers accept both; gateways emit only current. Rotation order should be: publish new current plus old previous, deploy Web and web-ingestion verifiers, deploy gateways, wait for the maximum cache/grant window, then remove previous. That makes the zero-downtime behavior explicit.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+The sealed context is reusable across all requests using that cached execution configuration. It contains a version, organization ID, project ID, Gateway virtual-key ID and attribution, selected connection ID, ingress/upstream protocols and adapter, telemetry mode, issued time, and expiry. It intentionally does **not** contain a request ID, model, or provider credential. Each inference request generates its own `gateway_request_id`, returned as `x-langfuse-gateway-request-id` and included in telemetry.
+
+Web seals the context with a Web-only authenticated-encryption key. The gateway treats it as opaque. A longer context lifetime, initially 24 hours, authorizes only delayed telemetry/media delivery; it can never authorize a new inference request or extend the 60-second execution-config cache. There is no grant table, Redis record, refresh flow, or JWKS subsystem.
```
hassiebp: Twenty-four hours is too long for this authority. Make this a project-scoped ingestion grant with an explicit OTLP/media write audience and expiry tied to the bounded retry horizon, roughly 5-10 minutes, and batch by project. We do not need new policy machinery: grant-authenticated requests should continue through the existing ingestion middleware so current suspension and limits still apply. For V1, a bounded hard cache with no stale serving and normal request timeouts is enough; add singleflight or negative caching only if load tests show resolver amplification.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Deployment and Reliability
+
+### Langfuse Cloud
+
+Deploy `ai-gateway` as an independently scalable Rust ECS/Fargate service in every existing regional environment. Reuse the regional VPC, private subnets, NAT egress, ALB/WAF, Secrets Manager, deployment pipeline, and monitoring foundations while giving the gateway dedicated tasks, target group, security group, IAM roles, and autoscaling.
+
+The public edge uses a gateway hostname and streaming-safe idle/drain settings. Gateway-to-Web resolution, telemetry, and media use public authenticated Web endpoints; private routing may be added later as transparent defense-in-depth. Scale primarily on active streams, with CPU, memory, bytes in flight, connection rate, and telemetry-queue pressure as guardrails. Deployments require readiness, graceful draining, a maximum stream duration, and progressive regional rollout.
```
hassiebp: Keep a stable custom gateway hostname from V1, routed through the existing ALB to a dedicated gateway target group. The hostname and load balancer are separate decisions: this gives clients the final URL now while initially sharing the ALB's 102-second idle timeout. Keep DNS/routing movable so the same hostname can later point at a dedicated ALB if silent-stream tests, WAF policy, or connection isolation require it.

specs/llm-gateway-rfc.md
```diff
@@ -0,0 +1,310 @@
...
+
+These tracks can proceed mostly independently once the resolution response, protocol identifiers, sealed-context contents, terminal outcomes, and telemetry facts are fixed. V1 launches only when all four tracks are ready.
+
+## Launch Gates
+
+- Real Claude Code and Codex sessions, including tools, media, cancellation, long sessions, and applicable model discovery.
+- OpenAI and Anthropic TypeScript/Python SDK conformance for native paths, plus Claude Code conformance for Anthropic, Bedrock, and Vertex.
+- Golden native request/response/error fixtures and exact SSE pass-through behavior.
+- Strict negative tests for incompatible protocol namespaces and connection combinations.
+- Credential/header leak, tenant isolation, revocation-window, SSRF, redirect, and custom-origin tests.
```
hassiebp: Please make credential-header protection an explicit launch gate. Enumerate every accepted customer/provider header, at least Authorization and x-api-key, and verify protection in WAF logs and samples, ALB/proxy logs, application logs, traces, errors, and support diagnostics. Also verify that the virtual key is stripped before upstream dispatch and provider credentials are never returned or recorded.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. V1 now contains only the native OpenAI and Anthropic namespaces; Bedrock and Vertex are explicit follow-ups, and the matrix, architecture, milestones, and launch gates are consistent.

specs/llm-gateway-rfc.md
hassiebp: Updated in 6430c21cb with the final control-plane decision. Connections are organization-owned and selected once per API namespace. Each execution maps to one project: key override first, otherwise the current organization default. Default changes affect future resolutions after cache expiry and are warned and audited; existing ingestion contexts retain their project.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. V1 now uses a controlled provider registry with official, provider-owned origins and no arbitrary base URLs. Custom origins require a separate secure-fetch boundary and remain out of scope.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. Usage-only telemetry is the default, full content capture requires explicit organization opt-in, and the selected mode is recorded on each observation.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. Rotation now uses current and previous: Web accepts both during rollout and gateways emit only current.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. The ingestion context is short-lived, project-scoped, and limited to OTLP and media writes for the bounded retry window. Existing public ingestion endpoints and controls are reused; there is no grant table or separate ingestion path.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. V1 uses a stable gateway hostname routed through the existing ALB to a dedicated target group, while DNS remains movable if isolation later requires a separate load balancer.

hassiebp: Addressed in 6430c21cb. V1 uses a stable gateway hostname routed through the existing ALB to a dedicated target group, while DNS remains movable if isolation later requires a separate load balancer.

specs/llm-gateway-rfc.md
hassiebp: Addressed in 6430c21cb. Credential redaction and boundary checks are now explicit launch gates covering Authorization, x-api-key, virtual keys, ingestion contexts, and provider credentials across WAF, proxy, application, trace, error, and support output.

hassiebp:
Consolidated the control-plane and namespace decisions in `6430c21cb`:

- Gateway connections are organization-owned and separate from existing project connections.
- One compatible connection powers the complete OpenAI or Anthropic API namespace, including model discovery; endpoint-level bindings are intentionally excluded.
- Connections come from a controlled provider registry. OpenAI, Anthropic, and OpenRouter are the initial types; arbitrary origins, Bedrock, and Vertex are deferred.
- Users with effective `gateway:invoke` permission can create user-owned virtual keys; organization administrators can create service keys.
- Each execution resolves to one telemetry project via key override first, then the current organization default.
- Usage-only telemetry is the default; full capture is opt-in, and telemetry/media reuse the existing public ingestion path with a short-lived ingestion context.

The RFC remains deliberately high-level and is now shorter despite covering the full V1 control-plane and data-plane shape.

PR #16892, opened 2026-09-01
fix(clickhouse): stringify DateTime64 timestamps on insert

### LINKED ISSUE #16858: bug: numeric DateTime64 inserts are misparsed on ClickHouse >= 26.8, multiplying every timestamp by 1000
### Describe the bug

On our self-hosted deployment (Langfuse **v3.225.5** OSS, web + worker, ClickHouse **26.8.1.2041**), a subset of traces is written to ClickHouse with `timestamp` and `created_at` set to **`9999-12-31 23:59:59.000`** — the maximum representable value for `DateTime64` — instead of the real event time.

**The trace content itself is perfectly intact**: `input`, `output`, `metadata`, `totalCost`, `latency` and `observations` are all correct and complete. Only the two date columns are wrong.

### Impact

Because `9999-12-31` sorts after every real date, these rows permanently occupy the top of any "most recent first" query — which is the default ordering of the Tracing UI. In our main production project this made the UI look **completely frozen**, as if ingestion had stopped days earlier, while in reality thousands of valid, correctly-timestamped traces were being ingested and were simply buried behind the corrupted rows.

This is not a display/read-path issue: `GET /api/public/traces?limit=5` also returns `"timestamp": "9999-12-31T23:59:59.000Z"` and `"createdAt": "9999-12-31T23:59:59.000Z"` for these rows, so the corruption happens at **write time**.

### Scope — this is org-wide, not integration-specific

It affects **every project in our organization that received traffic recently**, across different SDKs and integrations, in all environments (`producao`, `homologacao`, `desenvolvimento`, `default`). Corruption rate varies per project from **0.59% to 21.43%** of writes, but is never zero for any actively-used project.

| project (anonymized) | environment | total | invalid | % invalid | last valid timestamp |
|---|---|---:|---:|---:|---|
| project-A | producao | 240568 | 3081 | 1.28 | 2026-08-31 12:08:55 |
| project-B | default | 401 | 73 | 18.20 | 2026-08-31 12:05:35 |
| project-C | default | 427 | 64 | 14.99 | 2026-08-31 12:03:59 |
| project-A | desenvolvimento | 56 | 12 | 21.43 | 2026-08-31 12:00:55 |
| project-D | default | 505 | 3 | 0.59 | 2026-08-31 11:33:58 |
| project-E | default | 394 | 4 | 1.02 | 2026-08-27 18:38:54 |
| project-F | producao | 115 | **0** | 0.00 | 2026-08-27 00:00:52 |
| project-G | homologacao | 1203 | **0** | 0.00 | 2026-08-14 15:41:58 |
| project-H | desenvolvimento | 744 | **0** | 0.00 | 2026-08-13 11:45:58 |

Note the pattern in the last three rows: **every** project showing 0% corruption is also a project whose last activity predates ~2026-08-27. Every project with activity after that point shows corruption. This strongly suggests an ongoing regression introduced around that date, rather than a historical one-off.

### Expected behavior

`timestamp` and `created_at` should always reflect the actual event time derived from the OTel span, and should never silently fall back to a sentinel/overflow value under normal operation. If the timestamp cannot be resolved, failing loudly (or falling back to ingestion time) would be far preferable to a value that silently poisons every default-sorted query.

### Related

Possibly related to #12576 ("Langfuse v3: Traces written to ClickHouse but /api/public/traces returns empty"), which was closed as not planned. This report adds concrete, reproducible SQL evidence and a much wider blast radius.

### Environment

- **Langfuse self-hosted:** `v3.225.5` OSS — `docker.io/langfuse/langfuse:3` and `docker.io/langfuse/langfuse-worker:3` (floating `:3` tag, Docker Swarm)
- **ClickHouse:** `26.8.1.2041` (official build)
- **Client SDK:** Python `langfuse==3.15.0`, `opentelemetry-sdk==1.43.0`
- **Postgres:** external instance
- **Other projects affected** use different integrations, so this is not specific to one SDK path.

### Steps

1. Deploy self-hosted Langfuse `v3.225.5` (web + worker) against ClickHouse `26.8.1.2041`.
2. Instrument an application with Python SDK `langfuse==3.15.0`, using `from langfuse.openai import OpenAI` as a drop-in wrapper plus `langfuse.propagate_attributes(metadata=..., session_id=..., tags=...)` for trace-level attributes.
3. Send a sustained volume of LLM generation calls over several days.
4. Query ClickHouse directly and group by project/environment:

```sql
SELECT
    project_id,
    environment,
    count() AS total,
    countIf(toYear(timestamp) >= 9000) AS total_invalid,
    round(100 * countIf(toYear(timestamp) >= 9000) / count(), 2) AS pct_invalid,
    max(timestamp) AS max_timestamp_raw,
    maxIf(timestamp, toYear(timestamp) < 9000) AS max_timestamp_valid
FROM traces
GROUP BY project_id, environment
ORDER BY total DESC;
```

**Observed:** a non-zero `total_invalid` for every project/environment with recent traffic, and `max_timestamp_raw` is always exactly `9999-12-31 23:59:59.000`.

5. Inspect an individual corrupted row — note that both date columns are affected while everything else is fine:

```sql
SELECT id, name, timestamp, created_at, environment
FROM traces
WHERE toYear(timestamp) >= 9000
LIMIT 3
FORMAT Vertical;
```

```
id:          a12a657e2c10a6764ca48d74c2f592c0
name:        llm_chat_completion
timestamp:   9999-12-31 23:59:59.000
created_at:  9999-12-31 23:59:59.000
environment: producao
```

6. Confirm the same values are served by the public API (i.e. it is not a UI-only problem):

```
GET /api/public/traces?limit=5
```

```json
{
  "id": "6745c9757433a7c6a8c3444a1fa5634b",
  "name": "llm_chat_completion",
  "timestamp": "9999-12-31T23:59:59.000Z",
  "createdAt": "9999-12-31T23:59:59.000Z",
  "updatedAt": "9999-12-31T23:59:59.000Z",
  "environment": "producao",
  "totalCost": 0.00032595,
  "latency": 2356,
  "metadata": {
    "resourceAttributes": {
      "telemetry.sdk.language": "python",
      "telemetry.sdk.name": "opentelemetry",
      "telemetry.sdk.version": "1.43.0",
      "langfuse.environment": "producao",
      "service.name": "unknown_service"
    },
    "scope": { "name": "langfuse-sdk", "version": "3.15.0" }
  }
}
```

(`input` / `output` omitted here — they were fully populated and correct.)

### Additional context

- We use the floating `:3` image tag, so every `docker stack deploy` re-resolves it. The exact patch version in use has therefore drifted over time without us explicitly changing it, which makes it hard for us to bisect the regression on our end. Happy to run any diagnostics maintainers suggest.
- Around the same period our node ran out of disk space (affecting Redis RDB saves and ClickHouse background merges), resolved on 2026-08-31. Our analysis puts the start of the timestamp corruption at roughly 2026-08-27, which appears to **predate** the disk pressure — so we are not asserting a causal link, but flagging it in case the worker has a write-retry or fallback path that could emit a sentinel value under transient I/O errors.
- `system.errors` on ClickHouse is empty and the worker logs show no ingestion errors — the corrupted writes are completely silent.

Self-hosted

v3.225.5 (OSS) — self-hosted on Docker Swarm  Images (floating `:3` tag, which currently resolves to 3.225.5): - docker.io/langfuse/langfuse:3 - docker.io/langfuse/langfuse-worker:3

**Client application (the one producing the corrupted traces):**

```
langfuse==3.15.0            # Python SDK (OpenTelemetry-based)
opentelemetry-sdk==1.43.0
```

Integration used: `from langfuse.openai import OpenAI` (drop-in OpenAI wrapper) together with `langfuse.propagate_attributes(metadata=..., user_id=..., session_id=..., tags=...)` for trace-level attributes. Runtime: Python / FastAPI, containerized.

**Self-hosted stack:**

```
langfuse (web)          3.225.5   (docker.io/langfuse/langfuse:3)
langfuse-worker         3.225.5   (docker.io/langfuse/langfuse-worker:3)
clickhouse-server       26.8.1.2041 (official build)
redis                   7.4.11
postgres                external instance
```

**Important:** other projects in the same organization are affected by the identical bug while using *different* integrations and SDKs, so this does not appear to be specific to the Python SDK or to the OpenAI wrapper — it looks like a server-side (worker) issue.

### Silent failure

There is no error signal anywhere. `system.errors` in ClickHouse returns **0 rows**, the `langfuse-worker` logs show no ingestion errors, and the API returns HTTP 200 for the affected traces. The only way we found the problem was by querying ClickHouse directly. This makes the bug particularly costly to diagnose — from the outside it simply looks like "ingestion stopped working".

### Floating image tag

We pin both images to the floating `:3` tag, so every `docker stack deploy` re-resolves it and the exact patch version in use has drifted over time without us explicitly changing it. This unfortunately prevents us from bisecting the regression on our own. We are happy to run any diagnostics that maintainers suggest, or to test a specific candidate version if you can point us to one known to be unaffected.

### Possible (unconfirmed) correlation with disk pressure

Around the same period, the node hosting the stack ran out of disk space, which affected Redis RDB background saves (`Failed opening the temp RDB file ... No space left on device`) and ClickHouse background merges. This was resolved on 2026-08-31, and the disk currently has ~14 GB free with ClickHouse reporting ~87 GB of free space on its data disk.

However, our analysis places the start of the timestamp corruption at roughly **2026-08-27**, which appears to **predate** the disk pressure. We are therefore *not* asserting a causal link — flagging it only in case the worker has a write-retry or fallback path that could emit a sentinel timestamp under transient I/O errors.

### Corruption appears to be worsening

Projects created more recently show substantially higher corruption rates (up to ~21%) than long-running ones (~1%), which may indicate the rate is increasing over time rather than being stable.

### Current workaround

Applying an explicit date range filter in the Tracing UI (instead of relying on the default "most recent" ordering) hides the `9999-12-31` rows and makes genuinely recent traces visible again. This restores day-to-day usability but obviously does not address the underlying data corruption, and the affected rows remain effectively unqueryable by time.

No

Steffen911:
🟡 **Live preview: [pr-16892 app preview](https://pr-16892.preview.langfuse.com)**

Fixes https://github.com/langfuse/langfuse/issues/16858

Fixes https://github.com/langfuse/langfuse/issues/16858

## What does this PR do?

ClickHouse 26.8 changed the default of `input_format_read_datetime_number_as_raw_value` from `1` to `0`. An unquoted JSON number for a `DateTime64` column is now a Unix timestamp in **seconds**, then scaled by the column precision. Millisecond ticks overflow to `9999-12-31 23:59:59.000`.

Producers now write DateTime64 columns as ClickHouse datetime strings (`YYYY-MM-DD HH:MM:SS.sss`) via `toClickhouseDateTime`, and insert types require `string`. ClickHouse parses those strings the same way on every supported version, so no parser-setting workaround is needed.

Trace upsert eval jobs parse those insert strings back with `parseClickhouseUTCDateTimeFormat` so a non-UTC worker timezone cannot shift `exactTimestamp` and silently skip evaluations.

**Impacted packages:** `@langfuse/shared`, `worker`, `web` (tests/fixtures)

## Type of change

## Mandatory Tasks

## Checklist

- DateTime64 insert fields are `z.string()` and produced with `toClickhouseDateTime`
- Removed the table→column quoting helper, year-range guards, and `input_format_read_datetime_number_as_raw_value` compatibility setting
- Live insert test still proves quoted strings store the real event time on ClickHouse 26.8, and that unquoted millisecond ticks overflow
- Pinned `docker-compose.dev-redis-cluster.yml` to ClickHouse 26.8 so that CI job runs the affected server version
- Worker fixtures that insert via `clickhouseClient().insert()` now stringify DateTime64 fields (blob-log and retention helpers)
- Test factories keep explicit `null` on optional DateTime64 columns instead of writing now
- Deleted-mask cleaner integration drains leftover lightweight-delete patch parts. Required on ClickHouse 26.8 once sibling tests insert real months; reverting it failed `tests-worker (mode-redis-cluster)` again. Drain now lists every leftover month on `events_full` instead of the cleaner work query (`LIMIT 1 BY table`).
- Eval `exactTimestamp` is parsed as UTC from the DateTime64 insert string

## Verification

Against local ClickHouse `26.8.2.7`:

- `pnpm --filter @langfuse/shared run test client.datetime`: `Tests  6 passed (6)` including UTC round-trip of insert strings
- `pnpm --filter worker run test deletedMaskCleaner -t "DeletedMaskCleaner helpers"`: `Tests  7 passed | 1 skipped (8)`
- `pnpm --filter @langfuse/shared run test client.datetime compatibility`: `Tests  30 passed (30)` (earlier SHA)
- `pnpm --filter @langfuse/shared run test tracing-factory`: `Tests  3 passed (3)` including millisecond vs microsecond factory ticks
- `pnpm --filter worker run test datetime64Insert`: quoted `toClickhouseDateTime` string stores `2024-11-06 20:37:00.123`; unquoted millisecond ticks overflow to year `>= 9000`
- `pnpm --filter worker run test ClickhouseWriter.unit IngestionService.unit scoreValidationDroppedMetric`: `Tests  61 passed (61)` across those files
- Pre-commit turbo lint: `Tasks: 7 successful, 7 total`

Full web/worker servertest suites were not re-run locally. `tests-worker (mode-redis-cluster)` is the ClickHouse 26.8 cell. The previous SHA failed that job on leftover patch isolation (`201010` / `201102` / `201506`), not DateTime64 (`datetime64Insert` passed). The integration test itself was not re-run here after Redis hooks hung on this VM.

Linear Issue: [LFE-15835](https://linear.app/clickhouse/issue/LFE-15835/bug-numeric-datetime64-inserts-are-misparsed-on-clickhouse-268)

<div><a href="https://cursor.com/agents/bc-6486d8ae-f809-41b8-9a8f-e875f32421da?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6486d8ae-f809-41b8-9a8f-e875f32421da&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

This PR serializes numeric DateTime64 ticks as ClickHouse datetime strings and adds a ClickHouse 26.8+ compatibility setting for remaining numeric inserts.
- Adds shared millisecond and microsecond DateTime64 serialization helpers.
- Applies serialization to worker writes, repository upserts, deletion tombstones, backfills, and test helpers.
- Adds environment configuration and tests for ClickHouse compatibility behavior.
</details>

The negative microsecond timestamp conversion should be corrected before merging because it persists a different instant for reachable pre-epoch event timestamps.

The new precision-6 serializer combines truncation toward zero with an absolute remainder, corrupting negative microsecond ticks even though ingestion permits them to reach the events writer.

**Files Needing Attention:** packages/shared/src/server/clickhouse/datetime.ts
</details>

`````markdown
### Issue 1
packages/shared/src/server/clickhouse/datetime.ts:93-94
**Negative microsecond ticks are corrupted**

When an `events_full` timestamp is before the Unix epoch, truncating toward zero and taking the absolute remainder produces a datetime string for a different instant, shifting the persisted timestamp by up to 999 microseconds.

```suggestion
  const milliseconds = Math.floor(ticks / 1000);
  const microsecondRemainder = ((ticks % 1000) + 1000) % 1000;
```

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(clickhouse): quote DateTime64 ticks ..."](https://github.com/langfuse/langfuse/commit/15740e1fbb76e5f49c3fe0b4d3ecdee7c3505a19) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59056973)</sub>

### CODE REVIEW

File: packages/shared/src/server/clickhouse/datetime.ts [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+      "created_at",
+      "updated_at",
+      "event_ts",
+      "experiment_item_version",
+    ],
+  },
+  traces_null: {
+    precision: 3,
+    fields: ["start_time", "end_time", "created_at", "updated_at", "event_ts"],
+  },
```
Steffen911: Does this even exist at this time? Where is it being used?

packages/shared/src/server/clickhouse/datetime.ts
```diff
@@ -0,0 +1,171 @@
...
+} as const satisfies Record<string, DateTime64InsertTableSpec>;
+
+export type ClickHouseDateTime64InsertTable =
+  keyof typeof CLICKHOUSE_DATETIME64_INSERT_FIELDS;
+
+// ClickHouse DateTime64 calendar range is 1925-01-01 through 2283-11-11.
+// ISO-8601 expanded years (`+058638-...`) are not valid DateTime64 input.
+const CLICKHOUSE_DATETIME64_MIN_YEAR = 1925;
+const CLICKHOUSE_DATETIME64_MAX_YEAR = 2283;
```
Steffen911: Why do we need those? Apparently clickhouse happily inserted the year 9999 according to the ticket. I think we just need to stringify the right columns? Can't we do that in the IngestionService implementation when we write?

packages/shared/src/server/clickhouse/compatibility.ts
```diff
@@ -87,6 +88,15 @@ const CLICKHOUSE_COMPATIBILITY_RULES: ClickHouseCompatibilityRule[] = [
...
   },
+  {
+    id: "read-datetime-number-as-raw-value",
+    setting: "input_format_read_datetime_number_as_raw_value",
+    value: 1,
+    reason:
+      "ClickHouse 26.8+ reads an unquoted JSON DateTime64 number as Unix seconds and scales it by the column precision. Langfuse still inserts millisecond/microsecond ticks on some paths; restore the pre-26.8 raw-tick parse for any remaining unquoted numbers.",
+    versionBands: [{ minInclusive: "26.8.0.0" }],
+    overrideEnvKey: "CLICKHOUSE_READ_DATETIME_NUMBER_AS_RAW_VALUE",
+  },
```
Steffen911: I don't think we need this setting, right? Can't we just stringify numbers on write? They should be parsed correctly with either permutation of this flag?

Steffen911:
@claude review

packages/shared/src/server/clickhouse/client.ts
```diff
@@ -268,3 +268,27 @@ export const convertDateToClickhouseDateTime = (date: Date): string => {
   // 2024-11-06T20:37:00.123Z -> 2024-11-06 20:37:00.123
   return date.toISOString().replace("T", " ").replace("Z", "");
 };
+
+const CLICKHOUSE_DATETIME_STRING = /^\d{4}-\d{2}-\d{2} /;
```
maxdeichmann, line 272: Is this only required for inserts or also for selects (e.g. filter conditions)

PR #16898, opened 2026-09-01
fix(evals): simplify default environment filters

wochinge:
🟡 **Live preview: [pr-16898 app preview](https://pr-16898.preview.langfuse.com)**

## Summary

- Remove internal environment exclusions from the new evaluator and rule defaults.
- Keep the legacy evaluator experience and its trace / observation defaults unchanged.
- Default new rules to root spans and keep the useful `GENERATION` / semantic-root defaults.
- Hide Langfuse evaluation environments from new sample rows, counts, and environment suggestions under the hood.
- Rely on the existing worker scheduling and execution safeguards.

## Rationale

Loop prevention already lives in the worker, not in editable filters. This keeps the new setup UI simpler without allowing evaluations to recursively evaluate their own traces.

The legacy UI stays unchanged to avoid altering established behavior. Existing saved rule filters remain visible and unchanged. `sdk-experiment` remains visible and evaluable in the new experience. Langfuse-owned evaluation environments and the public `llm-as-a-judge` alias stay hidden from samples.

## Before / after

Before: new rules exposed root-span, environment, SDK experiment, and experiment-id exclusions.

After: new rules only expose `isRootObservation: true`. Internal environment exclusions are applied only to sample queries. The legacy evaluator defaults remain unchanged.

## Tests

- `Tests  21 passed (21)` - legacy defaults, new evaluator/rule defaults, and hidden sample filters
- `Tests  4 passed (4)` - final environment gate
- `Tests  7 passed (7)` - observation scheduling gate
- `Tests  7 passed | 101 skipped (108)` - trace scheduling coverage
- `Tasks: 7 successful, 7 total` - lint

### DIFF
--- a/web/src/features/evals/components/inner-evaluator-form.tsx
+++ b/web/src/features/evals/components/inner-evaluator-form.tsx
@@ -453,7 +453,8 @@ export const InnerEvaluatorForm = (props: {
       filter: props.existingEvaluator?.filter
         ? z.array(singleFilter).parse(props.existingEvaluator.filter)
         : defaultTarget === EvalTargetObject.TRACE
-          ? DEFAULT_TRACE_FILTER
+          ? // For new trace evaluators, exclude internal environments by default
+            DEFAULT_TRACE_FILTER
           : defaultTarget === EvalTargetObject.EVENT
             ? // For new observation evaluators, default to GENERATION type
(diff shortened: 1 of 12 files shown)


wochinge:
@claude review

PR #16913, opened 2026-09-01
feat(experiments): show the score columns, and make a score cell readable

nkabardin:
🟡 **Live preview: [pr-16913 app preview](https://pr-16913.preview.langfuse.com)**

**TL;DR** — Scores are the point of an experiment, and every score column shipped hidden. The experiments list carried **48 score columns, 0 visible**, and **24 of them were empty for every row**. After: **24 columns, all visible, all populated**.

**2 of 8** in the LFE-15711 experiments-UI stack. Base: #16912.

## Why

Read from `table:column_visibility_changed` over 90 days: on the experiments list, **449 users opened the column picker and 440 of them (98%) ended with at least one score column visible**. On the results page, 697 users made 5,218 column changes, **76% ending with a score column on**. Essentially everyone who opens the column picker is there to turn scores on.

Turning them all on is only useful if the empty ones stay out. On the list, `main` ships 48 score columns of which the 24 populated (trace-level) ones are off by default, while the 24 that are visible belong to another dataset and read empty on every row.

Two more things made a score cell hard to trust: a boolean score rendered as `0`/`1`/`True`, indistinguishable from a numeric score that happens to hold 0 and 1 — a user read her own data wrong because of it — and a categorical cell listed its values in event-timestamp order, so the same two values appeared as "true, false" in one row and "false, true" in the next.

## What

Score columns default to visible, with a one-time localStorage migration for returning users; columns are created only for the score keys the query actually returned; the score level moves out of the column id into the header alone, so a level can be renamed without orphaning a persisted layout. Boolean scores render as `true`/`false`, categorical values sort. An unrecorded cost or latency says "not recorded" instead of `0` or `-`. `Description` and `Metadata` come off the list's default columns to make room, the empty `Expected Output` column disappears rather than rendering `""`, and the picker lists real column names instead of `totalCost`.

## Result

Experiments list score columns: **48 total / 0 visible → 24 total / 24 visible**, all populated. Column picker `35/59` → `33/35`. Horizontal scroll extent 4112px → 3712px.

<details>
<summary>Mechanism, files and verification</summary>

**Visible by default + the migration.** `defaultHidden` alone only ever reaches new users: a returning user has `false` persisted for every score column. `revealScoreColumns` (`features/scores/lib/scoreColumns.ts`) opts them in once, and only while that stored state is still the old default — as soon as one score column is enabled the user has made their own choice and their layout is untouched. It returns `null` while no score column is known yet (they arrive with the score-column query) so the migration is retried rather than consumed against an empty set. The migration helpers move from `useColumnOrder` into a shared `features/column-visibility/lib/one-time-migration.ts`, since visibility now needs what order already had.

**Only where they have values.** `collectPresentScoreKeys` / `withPresentScoreKeys` narrow the column set to keys present in the fetched rows. `undefined` while the data loads, so columns do not disappear and come back on each fetch.

**The header/id split.** `createScoreColumns` takes an options object and a separate `headerPrefix`, so the group headers can read Trace / Observation / Experiment Scores while the persisted column ids keep their original prefix.

**Boolean and ordering.** `toBooleanScoreValue` in `aggregateScores.ts`, with the cell colour map extended to the lower-case forms; the categorical branch sorts by value.

**Not recorded.** `createNumberTableColumn` grows an `emptyCell` so a column can supply a node rather than a string; cost treats `0` as missing, because a run whose calls reported no usage or pricing sums to zero and reads as free.

**Data type.** `getScoreDataTypeExplanation` spells out what each type means for how its values read, shown in the score hover card in the side-by-side grid, which has no column headers to carry it.

**Verification:** typecheck clean, `turbo lint` clean across 7 packages, 239 client tests green (including the score-column, aggregate-score and column-picker suites). Numbers measured at 1512×800 against the reconciled reference branch: [`lfe-15711-experiments-ui-rebuild`](https://github.com/langfuse/langfuse/tree/lfe-15711-experiments-ui-rebuild).

</details>

The PR makes experiment score columns visible by default, limits them to scores present in fetched results, and improves score and missing-metric presentation.
- Adds reusable one-time migrations for persisted table visibility and ordering.
- Filters score columns using score keys present in experiment rows and separates display headers from persisted IDs.
- Normalizes boolean score labels, sorts categorical values, conditionally shows expected output, and adds explicit “not recorded” metric cells.
</details>

The visibility migration should be fixed before merging because returning users can still have valid score columns remain hidden after navigating between result pages or filters.

The migration records a project-wide completion flag after processing a page-scoped subset of score IDs, while persisted false values for score IDs absent from that page survive reconciliation.

**Files Needing Attention:** web/src/features/experiments/components/table/ExperimentsTable.tsx and web/src/features/experiments/components/table/ExperimentItemsTable.tsx
</details>

`````markdown
### Issue 1
web/src/features/experiments/components/table/ExperimentsTable.tsx:752-753
**Migration consumes partial score state**

When a returning user first opens a page containing only a subset of the project's score keys, this migration reveals that subset and records the project-wide migration as complete. Score columns first encountered on later pages retain their persisted `false` visibility, causing them to remain hidden despite the new visible-by-default behavior; the same pattern is present in `ExperimentItemsTable.tsx`.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(experiments): name the results colum..."](https://github.com/langfuse/langfuse/commit/91b61a2f1f313660317857751090e10bbf0a3c46) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59175073)</sub>

### CODE REVIEW

File: web/src/components/design-system/table/columns/createNumberTableColumn.tsx [hunk 1]
```diff
@@ -13,11 +14,14 @@ export function createNumberTableColumn<
...
 >({
   emptyValue,
+  emptyCell,
   formatter,
   getValue,
   ...options
 }: TableColumnOptions<TData, TValue> & {
   emptyValue?: string;
+  /** Rendered when there is no value; takes precedence over `emptyValue`. */
+  emptyCell?: ReactNode;
```
bezbac: Do you really need this? We already have `emptyValue`

nkabardin: @bezbac yeah nicely spotted. original idea was that they could look different

web/src/components/design-system/table/columns/createNumberTableColumn.tsx
```diff
@@ -14,14 +14,12 @@ export function createNumberTableColumn<
...
-  emptyCell,
   formatter,
   getValue,
   ...options
 }: TableColumnOptions<TData, TValue> & {
-  emptyValue?: string;
-  /** Rendered when there is no value; takes precedence over `emptyValue`. */
-  emptyCell?: ReactNode;
+  /** Rendered when there is no value. */
+  emptyValue?: ReactNode;
```
bezbac: Do we really need `ReactNode`? I would like to avoid generic slots. Maybe just migrate the table to not use `web/src/components/design-system/table/columns/createNumberTableColumn.tsx` anymore instead of widening the `createNumberTableColumn` for all callers.

bezbac (changes requested):

web/src/components/design-system/table/columns/createNumberTableColumn.tsx
nkabardin: sure, no opinion on that

bezbac: Thanks!

bezbac (approved):

PR #16930, opened 2026-09-02
feat(evals): add observation filter builder

wochinge:
🟡 **Live preview: [pr-16930 app preview](https://pr-16930.preview.langfuse.com)**

## Summary

- add a persisted Query/Builder toggle to observation filters in evaluator and rule configuration
- reuse the existing inline filter builder with the same filter set and ordering as evaluation rules
- use compact, consistently spaced builder columns, a subtle add action, a responsive layout that shrinks columns before wrapping rows, and a maximum filter-section width
- keep query-only filters out of new builder rows and clear invisible free-text constraints when entering Builder mode
- keep selected multi-select values visible when they are absent from the latest option response
- record the selected filter experience on evaluator create and update events
- add wide, narrow, and compact Storybook stories for the builder

## Testing

- `pnpm --filter web run test-client ObservationFilterBuilder.clienttest multi-select.clienttest single-select.clienttest`
- `pnpm --filter web run test-storybook ObservationFilterBuilder`
- `pnpm run lint`
- `pnpm --filter web run structure:stats`
- `git diff --check`

Browser verification was intentionally skipped.


wochinge:
@claude review

PR #16966, opened 2026-09-02
fix(worker): exit cleanly on SIGTERM instead of hanging until the container is killed

Steffen911:
🟡 **Live preview: [pr-16966 app preview](https://pr-16966.preview.langfuse.com)**

## What does this PR do?

**TL;DR:** the worker never exits on its own after a graceful shutdown, so every SIGTERM (autoscaling scale-in, deploy) ends with the orchestrator SIGKILLing the container at its stop timeout — exit code 137, which reads like an OOM kill. Separately, closing the queue workers can wait forever, and the ClickHouse writer flush runs *after* it, so that kill drops buffered rows of jobs that were already acked. This PR makes the worker exit on its own terms and bounds the wait.

### Why the process does not exit

The drain finishes in milliseconds and logs `Shutdown complete, exiting process...`, then Node sits there until it is killed.

Nothing in the drain releases the dedicated Redis client that every queue owns. `createBullMQQueueOptionsWithRedis` and `createBullMQWorkerOptionsWithRedis` hand each queue and each worker its own client, and BullMQ treats a caller-supplied ioredis instance as `shared`, so `queue.close()` and `worker.close()` deliberately leave it connected — a `Worker` only closes the blocking connection it duplicated itself. Those sockets keep the event loop alive, and because `redisQueueRetryOptions.retryStrategy` retries forever they keep reconnecting too.

#17096 reports exactly this from a self-hosted deployment, with a handle dump taken five seconds after `Shutdown complete`: 60 undestroyed sockets to `:6379` and nothing else but stdout/stderr. #17097 releases those clients at the source and is the better fix for that leak. This PR is the backstop underneath it: exit explicitly once the drain is done, so a leftover handle — today's Redis clients, tomorrow's something else — costs a clean exit instead of the whole stop timeout.

What this PR originally opened for — the tokenization `worker_threads` pool respawning threads that `terminate()` had just killed, and holding the loop open with them — landed separately in #17300, so it is no longer part of this diff.

### Why the ClickHouse flush can be skipped

Measured on our production ECS fleet over 7 days:

- Per-task Fargate metrics keep reporting until SIGTERM + 111–120 s, i.e. the ECS `stopTimeout = 120` SIGKILL. Sidecars exit 0, the app container exits 137.
- In one region ~28% of shutdowns never got past `WorkerManager.closeWorkers()` — bimodal, 0.2 s median when it works, otherwise silence until the kill. Those tasks were mid-way through very large ingestion payloads.
- `ClickhouseWriter.shutdown()` runs after `closeWorkers()`, so those tasks never flushed. The jobs were already acked, so their buffered rows are simply gone.

### Changes

- `worker/src/queues/workerManager.ts`: `closeWorkers(timeoutMs)` races the BullMQ `close()` calls against a timeout, logs which queues still had active jobs, and returns `false` instead of waiting forever. Abandoned jobs are recovered by BullMQ's stalled-job handling, exactly as they already were after the SIGKILL — but now the flush and the exit still happen.
- `worker/src/utils/shutdown.ts`: ignore a second signal while a shutdown is running, track the phase the drain is on, arm an unref'd hard deadline that logs that phase and exits 1, and `process.exit(0)` once the drain returns. The exit lives in `onShutdown` rather than in `drainAndClose`, because the fatal-error path (#16996) owns its own exit code.
- `worker/src/env.ts`: two optional knobs, both of which must stay below the orchestrator's stop timeout:
  - `LANGFUSE_SHUTDOWN_QUEUE_CLOSE_TIMEOUT_MS` (default `90000`) — how long to wait for active queue jobs.
  - `LANGFUSE_SHUTDOWN_TIMEOUT_MS` (default `110000`) — hard deadline for the whole shutdown.
- Tests: a `closeWorkers` block in `workerManager.test.ts`.

### How to verify after deploy

- ECS/K8s report exit code 0 for the app container on scale-in; time from SIGTERM to task stop drops from ~120 s to a few seconds.
- Tasks that were previously killed silently now log `Queue workers did not close within 90000ms, continuing shutdown with active jobs abandoned in: <queues>` followed by the ClickHouse writer flush.

### Verification

- `pnpm exec turbo run lint --force` → `Tasks: 10 successful, 10 total`, `Cached: 0 cached, 10 total`
- `pnpm run typecheck` → `Tasks: 10 successful, 10 total` (`Cached: 2 cached, 10 total` — the two build tasks)
- `pnpm --filter worker run test workerManager.test.ts` → `Tests 14 passed (14)`
- Against `main`'s unbounded `closeWorkers`, the three new tests fail: `Tests 3 failed | 11 skipped (14)` — one of them by hanging past its timeout, which is the bug.

### What a reviewer should doubt

- **Exit code 0 even when queue jobs were abandoned.** The shutdown itself succeeded and the warning log is the signal; a non-zero code would make dashboards noisier without changing recovery. Happy to flip it.
- **Defaults vs. Kubernetes.** 90 s / 110 s fit ECS's 120 s stop timeout. A Kubernetes deployment on the default 30 s `terminationGracePeriodSeconds` still gets SIGKILLed unless it lowers both knobs or raises the grace period. Worth a line in the self-hosting docs.
- **`process.exit(0)` masks handle leaks.** That is the point of a backstop, but it does mean a future leak will not announce itself. #17097 is what actually keeps the handle count honest; this PR should not be read as a replacement for it.
- **The unref'd hard deadline.** If a step hangs with no live handles at all, Node exits 0 before the deadline fires instead of logging the phase. That case still exits, which is the goal.
- **The second-signal guard is now partly redundant** with `drainAndClose`'s memoization; it still prevents a second SIGTERM from arming a second deadline.

## Type of change

## Mandatory Tasks


Steffen911:
@claude review

PR #17011, opened 2026-09-03
refactor(experiments): one Table settings popover on the results view

nkabardin:
🟡 **Live preview: [pr-17011 app preview](https://pr-17011.preview.langfuse.com)**

## TL;DR

The results view now has the same single **Table settings** popover as the experiments list, with the cell format joining Columns and row height inside it. Layout, diff mode and item visibility stay in **Display** — and the PR argues why: those three live in the URL and travel with a shared link, the other three are per-user preferences that a link does not carry.

Closes [LFE-15943](https://linear.app/clickhouse/issue/LFE-15943). **Position in the stack:** PR 13, based on `lfe-15711-12-arrows-one-line` (#17010) — one commit each.

## Why

The reviewer: *"why not unify the row & display options under `Table settings` to be consistent with /experiments table"*.

Two surfaces of the same feature disagreed. The list merged its column picker and row-height switch into one popover; the results view kept **three** controls that shape its table, in **two** places — Columns and Row height in the toolbar, and a Display menu stranded up in the page header. That is the confusion [LFE-8420](https://linear.app/clickhouse/issue/LFE-8420) originally filed against the old compare view, reproduced.

## What

**One commit, two parts.**

1. **The shared seam.** `settingsSections` on the toolbar forwards a node into the merged popover, under the shared two controls. Optional, and only read when the merged popover is opted into.

2. **The results view adopts it.** One `Table settings` button; **Format** (JSON / Formatted) moves out of `Display` and into the popover beside row height; `Display` keeps Layout, Diff and Item visibility.

## Result

| | before | after |
| --- | --- | --- |
| /experiments | `Table settings` | unchanged — Columns + Row height, no Format |
| /results toolbar | `Columns 19/22` + `Row height` | `Table settings` — Columns + Row height + Format |
| /results page header | `Display` — Layout, Diff, Item visibility, Format | `Display` — Layout, Diff, Item visibility |
| every other table | `Columns` + `Row height` | unchanged |

<details>
<summary>The judgement call, and the proof the other tables are untouched</summary>

### Why `Display` stayed separate

The ticket left this open, and warned that one popover mixing "which columns" with "which layout" may be worse than two clear ones. It is — and there is a sharper line available than "view mode vs setting":

**What is in the URL, and what is not.** Layout, diff mode and item visibility are `?layout=`, `?diff=` and `?itemVisibility=`. They are *the view*: they travel with a shared link, and they belong beside the baseline and comparison pickers in the page header, which is where the rest of that same URL state already lives. Columns, row height and cell format are `localStorage`: they are *your* preferences, they do not travel, and they belong on the table's own toolbar.

That line also settles Format, which the old grouping got wrong: it was in `Display` with the URL state, but it is a local preference of the same kind as row height. Moving it is what makes both containers coherent rather than just fewer.

Two further reasons not to merge everything: the layout switch is the most-used control on this page (the reviewer's own notes are organised by layout), and putting it under a settings button next to a list of column checkboxes would cost real discoverability; and the page header is where a reader already looks to see *what* they are comparing.

**And it is cheap to reverse.** The sections are passed in as a node, so if the reviewer wants one button after all, moving Display's sections into the same popover is a one-line change at the call site — no new mechanism.

### Every other table

- **Static:** `mergeSettingsIntoPopover` has exactly two call sites (the experiments list, and now the results view) out of 34 toolbars; `settingsSections` has one. Both are optional and default off.
- **Test:** the existing `data-table-toolbar.clienttest.tsx` case asserts a default toolbar renders `Columns` and row height as separate controls and *no* `Table settings` button. A new case asserts a passed section is absent until the popover opens and then appears inside it.
- **Browser:** the traces table still shows `Columns 16/34` + `Row height` and no `Table settings`; the dataset items table likewise; the experiments list's popover still holds only Columns and Row height, with no Format leaking in.

### Analytics

Unchanged, and none renamed. `table:column_visibility_changed` and `table:row_height_switch_select` fire from inside the popover — the same code path the experiments list has been using — and the old standalone row-height switch is no longer rendered on this surface, so there is one source, not two. `experiment:layout_changed` and `experiment:diff_mode_changed` stay on the page's own handlers, which this PR does not touch. (The dev server has no PostHog key, so these were verified by code path and by the controls behaving, not by captured events.)

### Per-branch checks

`shared` rebuilt with `db:generate` + `build` first.

- `pnpm exec knip` — clean
- `pnpm --filter=web run typecheck` — clean
- `turbo lint --force` after `rm -rf web/.next/cache/eslint` — 7 successful, **`Cached: 0`**
- `vitest run --project client` — 412 files, 4,070 tests passed

**Browser** (dark + light, 1512×800 and 1280×720): each control driven from inside the popover — row height Small → Medium (persisted, popover stays open), Format JSON ↔ Formatted (persisted, cells switch renderer), a column toggled off and back on through the drawer that opens from inside the popover — and `Display` still writing `?layout=` and `?diff=` from the page header. Checked in both the diff and side-by-side layouts.

</details>

## Known limitation, and why it stays

Picking a format **closes the popover**, unlike Columns and Row height. Both reviewers flagged it; it is deliberate, and the naive fix was tried and reverted:

`results.tsx` keys `ExperimentItemsTable` on `ioRenderMode`, so a format change remounts the subtree holding the popover. Dropping that key does keep the popover open — and then the format switch stops working entirely until a reload. The shared table's `MemoizedTableBody` has a hand-written `React.memo` comparator (`web/src/components/table/data-table.tsx`) that compares data, row height, selection, column visibility and column order, but **not the column definitions** — and `singleLine` lives on the IO column defs, so a format change alone never invalidates the body. Verified in the browser: with the key gone, picking *Formatted* flips the buttons and persists the preference while every Input cell keeps rendering the JSON tree; after a reload the same state renders one raw line.

A popover that closes is worse than nothing changing at all, so the remount stays. The real fix belongs to the shared table's comparator, not to this page, and is filed as [LFE-16094](https://linear.app/clickhouse/issue/LFE-16094).

https://claude.ai/code/session_01Bxa4EbY1xw1VkJwxABKgm3

This PR consolidates the experiment-results table controls into a single Table settings popover while retaining URL-backed comparison controls under Display.
- Adds an extension section to the shared table-settings popover.
- Moves experiment cell formatting into that section.
- Enables the merged settings popover for ExperimentItemsTable and updates its tests and prop contracts.
</details>

The PR appears safe to merge, with a non-blocking interaction issue where choosing a format immediately closes Table settings.

The consolidated controls otherwise remain reachable, but changing ioRenderMode remounts the keyed table subtree and resets the popover that now contains the format control.

**Files Needing Attention:** web/src/pages/project/[projectId]/experiments/results.tsx
</details>

`````markdown
### Issue 1
web/src/pages/project/[projectId]/experiments/results.tsx:177-183
**Format selection closes settings**

Selecting JSON or Formatted changes the key of the enclosing `ExperimentItemsTable`, remounting the subtree and resetting the open popover. Table settings therefore closes after every format selection, forcing users to reopen it before adjusting another option or switching back.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(experiments): one Table setting..."](https://github.com/langfuse/langfuse/commit/de82ba276e62e0e1f3497531da0d65c33970d7ab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59921533)</sub>

### DIFF
--- a/web/src/components/table/data-table-settings-popover.tsx
+++ b/web/src/components/table/data-table-settings-popover.tsx
@@ -20,7 +20,8 @@ import { usePostHogClientCapture } from "@/src/features/posthog-analytics/usePos
 
 /**
  * One "Table settings" surface for the controls that shape the same table —
- * columns and row height — instead of a button per control. Filed against the
+ * columns and row height, plus whatever presentation settings the surface adds
+ * through `extraSections` — instead of a button per control. Filed against the
  * old compare view as "fields vs. columns confused me as a new user"
  * and reproduced by the new experiments UI with one more control.
  *
@@ -39,6 +40,7 @@ export function DataTableSettingsPopover<TData, TValue>({
   tableName,
   isV4,
   onColumnGroupToggle,
+  extraSections,
 }: {
   columns: LangfuseColumnDef<TData, TValue>[];
   columnVisibility: VisibilityState;
@@ -52,6 +54,10 @@ export function DataTableSettingsPopover<TData, TValue>({
   tableName: string;
   isV4: boolean;
   onColumnGroupToggle?: (payload: ColumnGroupTogglePayload) => void;
+  /** A table's own presentation settings, rendered under the shared two. Each
+   *  should be a labelled block; a separator is drawn between them. Keeps
+   *  surface-specific vocabulary out of this component. */
+  extraSections?: React.ReactNode;
 }) {
   const capture = usePostHogClientCapture();
 
@@ -104,6 +110,12 @@ export function DataTableSettingsPopover<TData, TValue>({
               ))}
             </div>
           </div>
+          {extraSections && (
+            <>
+              <Separator />
+              {extraSections}
+            </>
+          )}
         </div>
       )}
     >
--- a/web/src/components/table/data-table-toolbar.clienttest.tsx
+++ b/web/src/components/table/data-table-toolbar.clienttest.tsx
@@ -221,4 +221,22 @@ describe("DataTableToolbar merged table settings", () => {
       screen.queryByRole("button", { name: /^Columns/ }),
     ).not.toBeInTheDocument();
   });
+
+  // A surface can add its own presentation settings to the popover (the
+  // experiment results view adds the cell format). They belong inside it, not
+  // as another button beside it — which is the shape this whole prop exists to
+  // avoid.
+  it("puts a surface's own settings section inside the popover", () => {
+    render(
+      <DataTableToolbar
+        {...settingsProps}
+        mergeSettingsIntoPopover
+        settingsSections={<p>Format</p>}
+      />,
+    );
+
+    expect(screen.queryByText("Format")).not.toBeInTheDocument();
+    fireEvent.click(screen.getByRole("button", { name: "Table settings" }));
+    expect(screen.getByText("Format")).toBeInTheDocument();
+  });
 });
--- a/web/src/components/table/data-table-toolbar.tsx
+++ b/web/src/components/table/data-table-toolbar.tsx
@@ -170,6 +170,10 @@ interface DataTableToolbarProps<TData, TValue> {
    *  shape is validated on the experiments list; needs both controls, so a
    *  table that passes only one keeps its single button. */
   mergeSettingsIntoPopover?: boolean;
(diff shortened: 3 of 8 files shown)


nkabardin:
@claude review

PR #17017, opened 2026-09-03
docs(agents): let agents write to the tracker, and give the practice a home

nkabardin:
🟡 **Live preview: [pr-17017 app preview](https://pr-17017.preview.langfuse.com)**

**2/3 of a stack.** Based on #17013; #17036 is based on this. Base retargets to `main` when #17013 lands.

## TL;DR

Agents may now write to the Langfuse issue tracker instead of stopping to ask, and the practice around it — read a feature's history before you change it, plan big work as subtickets, hand your reasoning over when you finish — becomes three skills in this repo. Marking replaces the approval gate: three permitted write shapes, each labelled and each marked as agent-written in its own text. Docs only, no code touched.

**Stacked on #17013**, which adds `pr-stack-workflow` and the `AGENTS.md` sections these skills point at. Base retargets to `main` when that lands.

## Why

The old rule was that an agent must present a findings table and wait for a human to pick rows before any tracker write. `skill-creator` made that a *mandatory* convention for every new skill, and four skills implemented it. It cost more than it protected: the expensive part of an agent session — the decisions, the reversals, the traps — evaporated when the session ended, because the one place it could have been written down was closed.

The replacement is not "no boundary". It is **marking plus a bounded set of shapes**, with exactly one gate left, on the one write whose cost falls on somebody else:

| Shape | Gate | Label |
| --- | --- | --- |
| Comment, when a human must be told something now | none | `AI commented` |
| Append a separated block to a description — where durable context belongs | none | `AI edited` |
| Create a subticket of an existing ticket | none | `AI created` |
| Create a ticket with **no parent** | **ask once, then file it yourself** | `AI created` |

State changes — assigning, moving, closing, estimating, re-prioritising — stay with a human throughout.

Marking is what replaces the gate, and it is load-bearing for a specific reason: the tracker's API authenticates as the **human who configured it**, so an unmarked agent write reads as that person's own words to everyone who sees it. The label is for filtering; the in-text marking is what prevents misattribution.

## What

**Three new skills**, next to `pr-stack-workflow`:

- **`linear-agent-writes`** — the policy above, and the single authority for it. Every other file points here and restates nothing.
- **`linear-context-handover`** — reconstruct a feature's history (commits → PRs → head branch → work item → prior agent blocks) before designing anything; write the reasoning onto the work item before asking for review. Ships `scripts/lf-context.sh`, which walks the git half automatically and hands off the tracker half.
- **`linear-planning`** — turn a feature into subtickets that map one-to-one onto an intended PR stack, each carrying enough context to be implemented from the ticket alone. Where to *cut* the stack stays in `pr-stack-workflow`; this skill does not restate it.

**Seven existing skills lose their gate** and gain a pointer plus their own domain consequence — `skill-creator` (the convention itself), `linear-bug-triage`, `incident-alert-tickets`, `infra-scaling`, `housekeeping`, `weekly-production-review`, `cursor-agents-workflow`. The four that review a whole queue now file their findings after one go-ahead for the set, and report `filed <key>` instead of leaving a dozen titles for somebody to paste.

**One rule narrowed.** `.agents/skills/**` is carved out of the internal-identifier ban, for identifiers only: maintainer guidance uses them as provenance an engineer can follow, and the rule already had thirteen violations of exactly that shape. Tracker URLs stay banned everywhere — a fork cannot open them and they carry the workspace name — including one that was already in the tree.

## Result

The reasoning behind a change has somewhere to live that outlasts the session that produced it, and an agent finishing work in this repo is told to put it there.

<details>
<summary>Why these skills are here rather than in the private plugin, and what review changed</summary>

They were built in a private plugin marketplace first, on the argument that a plugin installs once and reaches every repo while a skill here only loads here. Two things undid it.

The stronger half was that publishing here meant shipping a skill contradicting `skill-creator`'s mandatory approval gate — which this PR amends, so nothing conflicts. The reach half turned out weaker than it looked: every Langfuse repo that runs agents has this same `.agents/skills` machinery, so per-repo skills are already the answer to per-repo guidance; a skill here needs no install because `pnpm install` links it; and the plugin was not actually installed anywhere, which made the authority every other file cites the one file nobody loaded.

The duplication cost is also measured rather than hypothetical. That marketplace already held two skills copied out of repos and **both had drifted** — one by 197 diff lines against its source, the other pointing at a file that does not exist in the repo it describes. An anti-drift argument is not well made by the tree where drift has already happened twice. So: one home, cross-references instead of copies, and two greps that make the boundary testable — `pr-stack-workflow` contains no tracker vocabulary, and the tracker skills contain none of its build mechanics.

Secrecy was never the reason. `linear-bug-triage`, `housekeeping`, `incident-alert-tickets` and `cursor-agents-workflow` already ship publicly naming this tracker, its label names and its state names.

**Two consequences applied rather than invented.** Appending a cause section to a ticket that already exists is a description edit, so it needs nobody's permission. And a review that sweeps a queue asks once for the whole set, or for named rows — twelve separate questions would be worse than the pasting they replace.

**Verified.** `pnpm run agents:check` exits 0 and the shim sync links all three skills (they loaded in the session that wrote this). `skill-creator`'s own validator reports *Skill is valid!* for each. `prettier --check` and `turbo run lint` clean — `Tasks: 7 successful, 7 total`. Spellcheck clean on the changed files. No dead relative links introduced; the seven that remain in the tree are pre-existing illustrative paths in `skill-creator`.

**Not verified:** nothing was exercised as a Cursor or Codex agent, and `lf-context.sh`'s tracker-API path was tested only as far as its authentication error — the request is well-formed and the failure degrades to the documented hand-off, but the success-path field selection has not been run against a real token. Its git-to-PR walk was run end to end.

</details>

### DIFF
--- a/.agents/skills/linear-context-handover/references/handover-template.md
+++ b/.agents/skills/linear-context-handover/references/handover-template.md
@@ -1,0 +1,73 @@
+# Handover block template
+
+Append this to the ticket's **description**, below the human's prose, separated by
+a rule. Never edit the text above it. Label the ticket `AI edited`.
+
+Keep the headings you have content for and drop the rest — an empty section is
+worse than no section. Quote the human verbatim where a decision hinged on a
+phrase; a paraphrase loses the force of it.
+
+```markdown
+---
+
+## 🤖 AI post-context
+
+*Written by <agent> on <date>. Everything above this line is human-authored.*
+
+### What shipped
+
+One paragraph, then the PR numbers. Include the thing that was **not** obvious
+from the diff — the surface nobody tested, the path that changed by accident.
+
+### Decisions, and why — reversals included
+
+1. **<Decision>.** The reasoning, and the evidence if there was any. If it was
+   reversed later, say so here rather than leaving the first version to be found
+   and trusted.
+2. **<Thing that was built and then deliberately removed>.** What replaced it, or
+   why nothing did. Name where the deleted code can be recovered from.
+
+### How the human steered
+
+- What they were reacting to, and the redirect, quoted.
+- What they refuse outright (half-baked mechanisms, a particular register of
+  writing, a class of solution).
+- How they prefer to decide — from renders, from numbers, from a live click-path.
+
+### Tools and recipes
+
+The commands and fixtures that produced the work: seeder scenario, test command,
+analytics query, browser trick. Enough that the next agent reproduces the state
+instead of rebuilding it.
+
+### Traps
+
+The failures that cost time and would cost it again. One line each, cause first.
+
+### Left open
+
+- <Ticket id> — <what remains>
+- <Unticketed follow-up>, and why it was not done
+```
+
+## Pointer block, for leaf tickets
+
+```markdown
+---
+
+## 🤖 AI post-context
+
+*Written by <agent> on <date>.*
+
+<One or two decisions specific to THIS ticket.>
+
+Full handover for the effort: <TICKET-ID>.
+```
+
+## What not to put in it
+
+- A changelog of the diff. The PR already has that.
+- Status updates ("in progress", "merged"). The state field has that.
+- The same essay copied onto every ticket in the project.
+- Anything confidential in a place that syncs outward: no customer names, no
+  support-ticket ids, no vendor references that the public PR rules already ban.
--- a/.agents/skills/linear-context-handover/references/handover-template.md
+++ b/.agents/skills/linear-context-handover/references/handover-template.md
@@ -1,7 +1,9 @@
 # Handover block template
 
-Append this to the ticket's **description**, below the human's prose, separated by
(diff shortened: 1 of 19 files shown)


.agents/skills/linear-context-handover/references/handover-template.md
```diff
@@ -0,0 +1,73 @@
...
+Keep the headings you have content for and drop the rest — an empty section is
+worse than no section. Quote the human verbatim where a decision hinged on a
+phrase; a paraphrase loses the force of it.
+
+```markdown
+---
+
+## 🤖 AI post-context
+
+*Written by <agent> on <date>. Everything above this line is human-authored.*
```
nkabardin: " Everything above this line is human-authored.*" could be a lie, as this could be ai created ticket. we shoudl remove this note, don't make assumptions

PR #17022, opened 2026-09-03
feat(experiments): resolve the experiment behind a trace-attached score

nkabardin:
🟡 **Live preview: [pr-17022 app preview](https://pr-17022.preview.langfuse.com)**

## TL;DR

A score attached to a **trace** could not be resolved back to the experiment it
belongs to, so it was invisible to charts and to the experiment metric list —
and trace level is exactly where an LLM-as-judge on a dataset run writes. This
adds the two dimensions that resolve it, plus the linkage rule that decides
which trace-level scores are actually attributable to a single run.

**Backend-only, extracted deliberately so a backend engineer can review it on
its own.** 5 files, +183/−6, of which 118 lines are tests. No Prisma schema, no
ClickHouse migration, no worker, no public API, no queue contract.

It is the backend half of an 11-PR frontend stack (#16914 … #17011) and lands
first; the five files here are byte-identical to their state in #16914, so that
PR absorbs this merge as a no-op.

## Why

`scoresV2` resolves a score's experiment by joining `events_observations`. A
trace-attached score has no `observation_id`, so that join never matches and
`experimentName` / `experimentId` read null next to it. The score exists, is
aggregated correctly everywhere else, and simply cannot be grouped by run.

## What

**1. Two trace-scoped dimensions** (`dataModel.ts`) — `traceExperimentName` and
`traceExperimentId`, reading `events_traces` instead of `events_observations`.
Both `uiHidden`, both on a relation the model already declares, so they only
add a join when something selects them.

**2. A root-span fragment and an opt-in linkage**
(`query-fragments.ts`, `experiments.ts`) — a trace-level score name is only
offered as a metric when the trace's own parentless row carries the experiment.
That is the linkage the `scores → events_traces` relation joins on, so a name
offered under it can actually be plotted; offering a metric that draws nothing
is worse than leaving it out. The repository parameter defaults to the existing
per-item linkage, which is what the item tables filter on.

## What I would like a second opinion on

1. **Two notions of "root" now coexist.** This fragment uses the narrow
   `parent_span_id = ''`, matching the join it mirrors
   (`dataModel.ts`, the `events_traces` relation). But `eventsTable.ts` uses the
   broader `(parent_span_id = '' OR is_app_root = true)`. Consistent with what
   it mirrors — but if app-root traces should also count, then both places are
   wrong today, not just this one.
2. **Cost of the two dimensions.** They are `uiHidden` and per-dimension
   `relationTable`, so I read that as "join only when selected". Worth
   confirming against how the builder assembles relations.
3. **The narrowing's blast radius.** It changes the trace-level and agnostic
   fields returned by `getExperimentScoreOptions`. On `main` the only consumer
   of that procedure is the experiments charts grid, which reads the
   observation- and run-level fields and never the trace-level or agnostic ones
   — so nothing on `main` observes the change. The **item** table's facets and
   score columns come from `getExperimentItemScoreOptionsByLevel`, which this
   PR does not touch. Please sanity-check that reading.

<details>

- `parent_span_id` is a non-`Nullable` `String` in `events_full` and
  `events_core`, so `= ''` is a sound root test — a `Nullable` column would have
  made the predicate silently match nothing.
- `turbo run typecheck` — `Tasks: 8 successful, 8 total`
- `turbo lint --force` — `Tasks: 7 successful, 7 total`, `Cached: 0 cached`
- `pnpm exec knip` — clean
- `experiment-score-levels.servertest.ts` — `Tests 9 passed (9)`
- `experiment-repository.servertest.ts` — `Tests 26 passed (26)`
- Exercised end-to-end in the browser on the stack, against a project whose
  trace-level scores are the whole story: the metric list offers them, the
  score matrix aggregates them, and the per-run series plot.

</details>

This PR enables experiment attribution for trace-attached scores while ensuring score discovery follows the same linkage used by experiment charts.

- Adds hidden trace-scoped experiment name and ID dimensions to the v2 score model.
- Adds a parentless trace-root experiment query fragment.
- Uses trace-root linkage for experiment chart score options while preserving item-root linkage for item tables.
- Adds server tests for trace-level options and nested-item behavior.
</details>

The PR appears safe to merge with no concrete blocking or independently actionable non-blocking issues identified.

The new dimensions and score-option query use the same parentless trace linkage, nested-item behavior is deliberately separated from item-table semantics, and tests cover the affected trace-level paths.
</details>

```mermaid
flowchart LR
  S[Trace-attached score] --> T{Consumer}
  T -->|Experiment charts| R[Parentless trace-root linkage]
  R --> D[traceExperimentName / traceExperimentId]
  D --> C[Experiment metric series]
  T -->|Experiment item table| I[Item-root linkage]
  I --> F[Item score options and filters]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(experiments): only offer trace-score..."](https://github.com/langfuse/langfuse/commit/82f34f092af5ec22611e8ae7d7507c4cd838f864) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59993478)</sub>

**Context used:**

### DIFF
--- a/packages/shared/src/features/query/dataModel.ts
+++ b/packages/shared/src/features/query/dataModel.ts
@@ -882,6 +882,29 @@ const scoresV2BaseDimensions: DimensionsDeclarationType = {
     highCardinality: true,
     uiHidden: true,
   },
+  // The same two fields for trace-attached scores. A trace-level score has no
+  // `observation_id`, so the `events_observations` join above never matches and
+  // the dimensions next to it read null — which is where an LLM-as-judge on a
+  // dataset run writes. These read the experiment off the trace's root event
+  // instead.
+  traceExperimentName: {
+    sql: "nullIf(events_traces.experiment_name, '')",
+    alias: "traceExperimentName",
+    type: "string",
+    relationTable: "events_traces",
+    description: "Name of the experiment the scored trace belongs to.",
+    highCardinality: true,
+    uiHidden: true,
+  },
+  traceExperimentId: {
+    sql: "nullIf(events_traces.experiment_id, '')",
+    alias: "traceExperimentId",
+    type: "string",
+    relationTable: "events_traces",
+    description: "ID of the experiment the scored trace belongs to.",
+    highCardinality: true,
+    uiHidden: true,
+  },
 };
 
 // Factory for shared score-specific dimensions (both numeric and categorical)
--- a/packages/shared/src/features/query/dataModel.ts
+++ b/packages/shared/src/features/query/dataModel.ts
@@ -9,6 +9,7 @@ import {
 import { InvalidRequestError } from "../../errors";
 import {
   eventsTableIsRootObservationSqlForAlias,
+  eventsTableScoreTraceRootSqlForAlias,
   eventsTableTraceNameAggregationSqlForAlias,
   eventsTableTraceNameSqlForAlias,
 } from "../../eventsTable";
@@ -1039,8 +1040,9 @@ const createScoreTableRelations = (
       // its app roots come from newer SDKs and are expected to carry trace_name.
       // A semantic-root join can match both rows in dual-write data and
       // multiply scores for little fallback value.
-      joinConditionSql:
-        "ON scores.trace_id = events_traces.trace_id AND scores.project_id = events_traces.project_id AND events_traces.parent_span_id = ''",
+      joinConditionSql: `ON scores.trace_id = events_traces.trace_id AND scores.project_id = events_traces.project_id AND ${eventsTableScoreTraceRootSqlForAlias(
+        "events_traces",
+      )}`,
       timeDimension: "start_time",
       useFinal: false,
     },
--- a/packages/shared/src/server/queries/clickhouse-sql/query-fragments.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/query-fragments.ts
@@ -440,6 +440,23 @@ export const eventsExperimentsRootSpans = (params: {
   );
 
 /**
+ * Narrows to the trace's parentless row, and only when that row itself carries
+ * the experiment. `experiment_id` is stamped on an item's root span and its
+ * subtree, so a trace whose item is nested inside a wider trace has a
+ * parentless row without one.
+ *
+ * This is the linkage the scores -> events_traces relation joins on, so a
+ * trace-level score reachable through this fragment is one a chart over that
+ * relation can actually plot. Use `eventsExperimentsRootSpans` instead for the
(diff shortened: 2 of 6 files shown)


packages/shared/src/server/queries/clickhouse-sql/query-fragments.ts
```diff
@@ -439,6 +439,23 @@ export const eventsExperimentsRootSpans = (params: {
...
+ * This is the linkage the scores -> events_traces relation joins on, so a
+ * trace-level score reachable through this fragment is one a chart over that
+ * relation can actually plot. Use `eventsExperimentsRootSpans` instead for the
+ * per-item semantics the item tables need.
+ */
+export const eventsExperimentsTraceRootSpans = (params: {
+  projectId: string;
+  experimentIds?: string[];
+}): EventsQueryBuilder =>
+  eventsExperiments(params).whereRaw("e.parent_span_id = ''");
```
sumerman: no longer an accurate way to filter for root spans, `is_app_root` should be also consulted.

packages/shared/src/features/query/dataModel.ts
```diff
@@ -882,6 +882,29 @@ const scoresV2BaseDimensions: DimensionsDeclarationType = {
...
+  // the dimensions next to it read null — which is where an LLM-as-judge on a
+  // dataset run writes. These read the experiment off the trace's root event
+  // instead.
+  traceExperimentName: {
+    sql: "nullIf(events_traces.experiment_name, '')",
+    alias: "traceExperimentName",
+    type: "string",
+    relationTable: "events_traces",
+    description: "Name of the experiment the scored trace belongs to.",
+    highCardinality: true,
```
sumerman, line 897: if both dimenstions are marked as uiHidden what are they for?

sumerman (changes requested):

packages/shared/src/features/query/dataModel.ts
```diff
@@ -1016,8 +1040,9 @@ const createScoreTableRelations = (
       // its app roots come from newer SDKs and are expected to carry trace_name.
       // A semantic-root join can match both rows in dual-write data and
       // multiply scores for little fallback value.
-      joinConditionSql:
-        "ON scores.trace_id = events_traces.trace_id AND scores.project_id = events_traces.project_id AND events_traces.parent_span_id = ''",
+      joinConditionSql: `ON scores.trace_id = events_traces.trace_id AND scores.project_id = events_traces.project_id AND ${eventsTableScoreTraceRootSqlForAlias(
```
sumerman: that is incorrect to do in a JOIN because can lead to duplicates

PR #17044, opened 2026-09-04
feat(in-app-agent): improve tool payload and evaluator code rendering

LysanderKie:
🟡 **Live preview: [pr-17044 app preview](https://pr-17044.preview.langfuse.com)**

## Summary

Render assistant tool arguments and results as an expandable JSON tree. Configured evaluator tools open the first source field as highlighted code, with its exact JSON path visible. “View JSON” returns to the full structure, where each source field has its own “View as code” action.

Only one code field is shown at a time. Copying stays inside the code block and JSON tree; there are no duplicate header controls. Unconfigured tools retain the standard JSON tree.

## Implementation

- `tool-payload-renderers.ts` owns tool/argument/result registrations, evaluator source discovery, bounded nested-JSON preparation, and exact-source copying.
- The payload component owns selection and presentation. Field identity and paths distinguish repeated source strings without changing literal escapes.
- Shared JSON-view customization is opt-in. The existing Markdown provider in client tests is needed by the shared JSON viewer.
- Impacted package: `web`. Temporary prototypes and local seeder changes are excluded.

## Verification

- `pnpm run lint`: 7 successful, 7 total.
- `pnpm --filter web run typecheck`: passed.
- Targeted `InAppAgentToolCallCard.stories.tsx`: 13 passed.
- Targeted assistant execution/window client tests: 29 passed.
- Browser checked in the running local assistant: default code/path, JSON navigation, and return to the selected source.

## Review steps

In Storybook, open **Features / In-App Agent / InAppAgentToolCallCard**:

1. **Approval required with code** shows the complete evaluator source before approval. **Evaluator tool call** shows a completed call with arguments and results; choose **View JSON** in the result, then either version's **View as code**. Verify the selected path and copied code.
2. Both main stories exercise full JSON and exact source copying. **Nested evaluator payload** covers nested parsing; **Unconfigured code payload** stays JSON-only. The existing **Error** story covers malformed JSON.
3. In the app, open an evaluator tool call in Assistant and try the same navigation. Code and JSON each retain their native copy control.


web/src/components/design-system/Codeblock/Codeblock.tsx
```diff
@@ -15,10 +15,19 @@ interface Props {
   showLanguage?: boolean;
   /** Match immutable form fields instead of using the recessed code surface. */
   variant?: "default" | "read-only";
+  /** Constrain long read-only code to its own scroll area. */
+  maxHeight?: string | number;
```
bezbac: Can you limit this with a surrounding element instead? Generally I would like to avoid props like this and just make the components fill available space, then place constraints around them in the layout. This makes components more reusable.

bezbac (changes requested):

web/src/components/design-system/Codeblock/Codeblock.tsx
LysanderKie: Done in 2d137c624 — the height constraint now lives in the payload layout, and CodeBlock’s API is unchanged.

bezbac: Thanks

bezbac (approved):
Design system changes now look good, have not looked at the rest :)

LysanderKie:
@greptileai review

PR #17063, opened 2026-09-04
ci: drop apt-get from the pipeline hot path and retry flaky setup steps

nimarb:
🟡 **Live preview: [pr-17063 app preview](https://pr-17063.preview.langfuse.com)**

## TL;DR

Removes every `apt-get` from `pipeline.yml`'s hot path and stops a failed Docker Hub login from failing the job. On 2026-09-04 Canonical's Ubuntu mirrors, Docker Hub and the npm registry were all degraded between 06:00 and 10:20 UTC, and 37 of that day's 40 failed CI runs died in exactly these steps. Over the last 14 days this change would have prevented 78 of the 95 infrastructure-caused job failures. Green-run wall time is unchanged; this is a reliability PR.

## What changed

- **Playwright install without `--with-deps`** (`tests-storybook`, `e2e-tests`) via the new `scripts/ci/install-playwright-chromium.sh`. The runner image already ships every Chromium runtime library; `--with-deps` only added CJK fonts plus apt upgrades, and its `apt-get update` against Canonical's mirrors was the top flake. The script downloads the headless shell (3 bounded attempts), runs a launch probe, and only falls back to `playwright install-deps` (apt) when the probe fails, with a warning annotation so the fallback is visible. The previous retry loop never retried effectively: `timeout` killed pnpm but left the spawned `apt-get` holding the dpkg lock, so attempts 2 and 3 failed instantly.
- **`e2e-tests`**: background install uses the same script, the wait drops from 600 s to 420 s, and the job gets `timeout-minutes: 15`. 10 of the 13 recent e2e browser-install failures sat at exactly the 600 s wait.
- **`tests-shared` ClickHouse binary**: static tarball from `packages.clickhouse.com` instead of adding an apt repository and running `apt-get update`, which refreshed Canonical's sources as well.
- **`tests-in-app-agent-sandbox-runtime`**: Blacksmith-backed builder (`cache-key: langfuse/sandbox-runtime-e2e`) so the alpine/apk/npm layers of the e2e image persist across runs. That builder uses buildx's `remote` driver, so the `test:e2e` script now builds with `docker buildx build --load` to land the image in the local daemon for dockerode.
- **Docker Hub login no longer fails the job** (`continue-on-error: true` on the 7 read-only logins in test jobs). Login only avoids anonymous pull rate limits, and Blacksmith's image cache serves most pulls without contacting Docker Hub; its intermittent 5xx (7 failures in 14 days) should not fail a job. Release logins are untouched.

## Impacted packages

`.github/workflows/pipeline.yml`, new `scripts/ci/install-playwright-chromium.sh`, `packages/in-app-agent-sandbox-runtime/package.json` (`test:e2e` script only) and its README.

## Verification

- `actionlint .github/workflows/pipeline.yml`: exit 0 once the pre-existing unknown Blacksmith runner labels are ignored.
- `zizmor --persona=regular` on the workflow and both composite actions: 0 medium/high. The new low `self-repository` hints are the same class as the existing `./.github/actions/notify-slack-failure` references; actionlint rejects the `$/` syntax zizmor suggests.
- `pnpm run agents:check`: exit 0. `pnpm run format:check`: all files formatted.
- Install script run locally end to end (macOS, browsers cached): exit 0, launch probe passes in about 2 s.
- ClickHouse tarball layout verified: `clickhouse-common-static-26.4.5.143/usr/bin/clickhouse`; the pinned SHA-512 matches the published `.sha512` and the actual download, and the CI step logs `OK`.
- The decisive check is this PR's own CI run: every changed job executes here.

## What to doubt

- Not retried on purpose: `pnpm/setup`'s pnpm binary download (8 failures during one npm-registry 504 window on 09-04). A retry needs a duplicated `uses:` step in all 13 jobs or a composite action; judged not worth the indirection, a rerun fixes it.
- `docker buildx build --load` on Blacksmith's remote builder exports the image through the daemon. If the sandbox e2e job gets slower rather than faster, that export is the suspect.
- The ClickHouse step keeps `continue-on-error: true`, so a tarball download failure still silently skips the SQL-equivalence tests. Kept deliberately for the first weeks.


.agents/AGENTS.md
```diff
@@ -40,6 +40,11 @@ evaluating, and debugging AI applications.
 - Always quote file paths in shell commands, or use `noglob` for path-heavy
   commands, to avoid zsh glob expansion issues with dynamic Next.js routes.
 - Never invoke Node-installed binaries through `./node_modules/.bin/*`. Always run them through `pnpm`.
+- CI jobs in `.github/workflows/pipeline.yml` never run `apt-get` on their
+  hot path: Canonical's Ubuntu mirrors are the pipeline's least reliable
+  dependency. Use what the runner image ships, or fetch a pinned binary
+  directly (`curl` + tarball); `scripts/ci/install-playwright-chromium.sh`
+  shows the pattern, including a probe-gated apt fallback.
```
nimarb: remove

Issue #8156, opened 2025-07-28
bug: Worker shutdown takes ~1 hour in self hosted kubernetes


Henni16:
### Describe the bug

We have a self-hosted langfuse instance running in kubernetes.
Shutting down the worker pod produces the following log:

```text
2025-07-28T09:27:53.708Z info 	Received SIGTERM, closing server...
2025-07-28T09:27:53.708Z info 	Set sigterm received to true
2025-07-28T09:27:53.709Z info 	Server has been closed.
2025-07-28T09:27:53.714Z info 	All workers have been closed.
2025-07-28T09:27:53.714Z info 	Starting ClickhouseWriter. Max interval: 1000 ms, Max batch size: 10000
2025-07-28T09:27:53.714Z info 	Shutting down ClickhouseWriter...
2025-07-28T09:27:53.715Z info 	ClickhouseWriter shutdown complete.
2025-07-28T09:27:53.715Z info 	Clickhouse writer has been shut down.
2025-07-28T09:27:53.715Z info 	Redis connection has been closed.
2025-07-28T09:27:53.722Z info 	Prisma connection has been closed.
2025-07-28T09:27:53.722Z info 	All tokenizers are cleaned up from memory.
2025-07-28T09:27:53.722Z info 	Shutdown complete, exiting process...
```

The pod will then stay in "Terminating" mode for at least 1 hour until eventually shutting down.
Though we also had one pod terminating for over 12 hours until we killed it manually.

### To reproduce

Shut down the worker pod in kubernetes.

### SDK and container versions

Langfuse Version 3.86.0

### Are you interested to contribute a fix for this bug?

No

Henni16:
@dosu.
The langfuse instance is completely fresh and has basically no load. All queues where empty before the shutdown. Also, according to the log the ClickhouseWriter writer did complete its shutdown.

Henni16:
@dosu
The issue you're referring to is about the python client. We're having issues with the backend worker pod.

maxdeichmann:
@Henni16 thanks a lot for reaching out on this one. As the logs are all from ~09:27:53, does the shutdown wait for 1h "after" the log line "Shutdown complete, exiting process..."?

Henni16:
Correct. The log above is the only thing showing in Kubernetes until the pod eventually dies.

maxdeichmann:
Ok, this is very odd and we have never seen this before. We will look into this.

Steffen911:
[...]

PR #17097, opened 2026-09-05
fix(redis): disconnect per-queue clients on shutdown so the worker process exits

### LINKED ISSUE #17096: bug: worker process does not exit after graceful shutdown — 60 orphaned Redis sockets keep the event loop alive
### Describe the bug

The worker process does not exit after a graceful shutdown completes. It logs `Shutdown complete, exiting process...` within ~20ms, then stays alive indefinitely until it is killed. On Kubernetes this means every worker rollout burns the full `terminationGracePeriodSeconds` before the pod is SIGKILLed.

This is the same symptom as #8156, which was closed as not planned after the reporter could not provide handle diagnostics. I have that data now, so I am opening a fresh report rather than commenting on a closed issue.

`process._getActiveHandles()` five seconds after `Shutdown complete` shows **60 undestroyed Redis sockets**:

```
[handle-dump] ===== active handles: 62 =====
  - Socket {"fd":2,"destroyed":false}
  - Socket {"fd":1,"destroyed":false}
  - Socket {"localPort":47266,"remote":"<redis-host>:6379","destroyed":false}
  - Socket {"localPort":47294,"remote":"<redis-host>:6379","destroyed":false}
  ... 58 more, all to :6379
[handle-dump] by type: {"Socket":62}
[handle-dump] active requests: 0
```

Excluding fds 1 and 2 (stdout/stderr), every remaining handle is a Redis connection. They keep the event loop alive, so Node never exits.

The likely cause is that `onShutdown` closes only the shared client:

```ts
redis?.disconnect();
logger.info("Redis connection has been closed.");
```

but each queue creates its own connection via `createNewRedisInstance()` (dedicated BullMQ clients, #3052). `WorkerManager.closeWorkers()` closes the workers, and nothing disconnects those per-queue clients. Since `redisQueueRetryOptions.retryStrategy` is documented to "retry forever", the orphaned connections also keep reconnecting after shutdown — the log continues to emit `Redis connection error` and `Eviction policy is allkeys-lru` long after `Shutdown complete`.

### Steps to reproduce

1. Start the worker against any Redis (a plain `docker compose` stack is enough — no load and empty queues are fine).
2. Send SIGTERM without forcing: `docker kill --signal=SIGTERM <worker>`.
3. Observe the shutdown sequence complete in milliseconds, then watch the container stay `running`.

Observed locally for 320s with no exit. On Kubernetes, measured on three pods simultaneously:

| | |
|---|---|
| SIGTERM | 14:07:32 |
| `Shutdown complete, exiting process...` | 14:07:32 (**21ms**) |
| pods actually removed | 14:12:34 |
| total | **302s** — the 300s grace period, then SIGKILL |

All three pods disappeared at the same instant, which is the signature of grace expiry rather than independent clean exits.

To capture the handles without patching the image, preload a script with `NODE_OPTIONS=--require /path/to/dump.js`:

```js
process.on("SIGTERM", () => {
  setTimeout(() => {
    const hs = process._getActiveHandles();
    console.log(`[handle-dump] active handles: ${hs.length}`);
    for (const h of hs) console.log(`  - ${h.constructor?.name}`, JSON.stringify({
      fd: h.fd, localPort: h.localPort,
      remote: h.remoteAddress ? `${h.remoteAddress}:${h.remotePort}` : undefined,
      destroyed: h.destroyed,
    }));
  }, 5000).unref();
});
```

Self-hosted

`3.225.7`. #8156 reported the same behaviour on `3.86.0`, so this has been present for a long time.

Not applicable — this is the worker container.

### Additional information

A possible fix that avoids touching every queue file: since all connections go through `createNewRedisInstance()`, they could be tracked there and disconnected together during shutdown.

```ts
const activeRedisInstances = new Set<Redis | Cluster>();
// register in createNewRedisInstance()
export const disconnectAllRedisInstances = () => {
  for (const r of activeRedisInstances) { try { r.disconnect(); } catch {} }
  activeRedisInstances.clear();
};
```

Calling that from `onShutdown` instead of `redis?.disconnect()` would release all of them.

Worth noting that a lower `terminationGracePeriodSeconds` (suggested in #8156) only shortens the wait — the process still has to be SIGKILLed. It is a safe mitigation, though, since the shutdown sequence finishes in ~20ms and nothing is lost when the kill arrives.

Yes

closset703:
Fixes #17096

## Problem

The worker completes its shutdown sequence in ~20ms and logs `Shutdown complete, exiting process...`, but the process never exits. On Kubernetes this means every worker rollout burns the entire `terminationGracePeriodSeconds` before the pod is SIGKILLed — 300s in our case, and #8156 reported ~1 hour.

`process._getActiveHandles()` five seconds after that log shows 60 undestroyed Redis sockets — every handle except stdout/stderr:

```
[handle-dump] ===== active handles: 62 =====
  - Socket {"fd":2,"destroyed":false}
  - Socket {"fd":1,"destroyed":false}
  - Socket {"localPort":47266,"remote":"<redis>:6379","destroyed":false}
  ... 59 more, all to :6379
[handle-dump] by type: {"Socket":62}
[handle-dump] active requests: 0
```

`onShutdown` closes only the shared client:

```ts
redis?.disconnect();
```

but each queue owns a dedicated client from `createNewRedisInstance()` (introduced in #3052). Those are never closed, and because `redisQueueRetryOptions.retryStrategy` retries forever they keep reconnecting — the log keeps emitting `Redis connection error` well after `Shutdown complete`. Those sockets hold the event loop open.

## Change

Track each instance handed out by `createNewRedisInstance()` and disconnect them together during shutdown.

- `activeRedisInstances` set, populated by a small `trackRedisInstance()` wrapper covering all three paths (standalone, cluster, sentinel)
- instances remove themselves on `"end"`, so the set does not grow across reconnects
- `disconnectAllRedisInstances()` is exported and called from `onShutdown`, returning the count for the log line

This keeps the change in one place instead of touching ~30 queue modules.

## Verification

Eight queue clients against a local Redis, with no timers held so the process exits as soon as the event loop drains:

| | result |
|---|---|
| with `disconnectAllRedisInstances()` | `closed 9` → **exit 0, process ends on its own** |
| without it (current behaviour) | **exit 124 — still running after a 25s timeout** |

Reproduced the original symptom first: sending SIGTERM to the worker container (`docker kill --signal=SIGTERM`, no force) left it `running` for the full 320s I watched. On Kubernetes, three pods were SIGTERMed at 14:07:32, logged `Shutdown complete` at 14:07:32, and were removed at 14:12:34 — all three at the same instant, which is grace expiry rather than independent clean exits.

## Checks

| check | result |
|---|---|
| `turbo run lint --force` (shared, worker) | pass — 5/5, 0 cached |
| `turbo run typecheck --force` (shared, worker) | pass — 5/5, 0 cached |
| `packages/shared/src/server/redis/redis.test.ts` | pass — 25/25 |
| `prettier --check` | pass (via pre-commit hook) |
| full `pnpm run test` | **not run** — needs the compose stack and a seeded DB |

Happy to add a unit test for `disconnectAllRedisInstances()` if you'd like it in this PR.

This PR tracks dedicated Redis clients created for queues and disconnects the tracked clients during worker shutdown, allowing the Node.js event loop to drain instead of waiting for Kubernetes to force termination.
- Adds centralized tracking and cleanup for standalone, cluster, and sentinel clients.
- Removes clients from the registry when they reach the ended state.
- Invokes bulk Redis cleanup after workers and writers have shut down.
- The lifecycle behavior currently lacks focused regression coverage.
</details>

The PR appears safe to merge, with the non-blocking concern that its shutdown regression fix is not protected by a focused test.

The cleanup is called after Redis-dependent worker shutdown and covers all factory branches, with no concrete functional or security failure identified; only missing regression coverage remains.

**Files Needing Attention:** packages/shared/src/server/redis/redis.ts
</details>

```mermaid
flowchart TD
    A[createNewRedisInstance] --> B[Create standalone, cluster, or sentinel client]
    B --> C[Add client to activeRedisInstances]
    C --> D[Queue or worker uses client]
    D --> E{Client emits end?}
    E -->|Yes| F[Remove client from registry]
    E -->|No| G[Worker shutdown begins]
    G --> H[Close workers and background services]
    H --> I[disconnectAllRedisInstances]
    I --> J[Disconnect each tracked client]
    J --> K[Clear registry]
    K --> L[Event loop can drain]
```
</details>

`````markdown
### Issue 1
packages/shared/src/server/redis/redis.ts:259-274
**Shutdown cleanup lacks tests**

The new process-wide client registry and disconnect-all shutdown path have no regression test. Please add a focused test that creates tracked clients, verifies each is disconnected exactly once, confirms ended clients are removed, and confirms a second cleanup is empty. Without this coverage, later lifecycle changes could silently restore the worker shutdown hang.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(redis): disconnect per-queue clients..."](https://github.com/langfuse/langfuse/commit/0f0ecd0665456b698e9bcb5cec16fddc832750ed) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60695762)</sub>

### DIFF
--- a/packages/shared/src/server/redis/redis.ts
+++ b/packages/shared/src/server/redis/redis.ts
@@ -235,6 +235,44 @@ const createRedisSentinelInstance = (
   return instance;
 };
 
+/**
+ * Every connection handed out by createNewRedisInstance, so that a shutdown can
+ * release all of them. Each queue owns a dedicated client, and those clients
+ * retry forever; left connected they keep the event loop alive and the process
+ * never exits.
+ */
+const activeRedisInstances = new Set<Redis | Cluster>();
+
+const trackRedisInstance = <T extends Redis | Cluster | null>(
+  instance: T,
+): T => {
+  if (instance) {
+    activeRedisInstances.add(instance);
+    instance.once("end", () => activeRedisInstances.delete(instance));
+  }
+  return instance;
+};
+
+/**
+ * Disconnect every client created by createNewRedisInstance.
+ *
+ * `redis.disconnect()` on the shared client only covers that one connection.
+ * Returns how many were closed, for shutdown logging.
+ */
+export const disconnectAllRedisInstances = (): number => {
+  let closed = 0;
+  for (const instance of activeRedisInstances) {
+    try {
+      instance.disconnect();
+      closed++;
+    } catch (error) {
+      logRedisError("Failed to disconnect Redis instance on shutdown", error);
+    }
+  }
+  activeRedisInstances.clear();
+  return closed;
+};
+
 export const createNewRedisInstance = (
   additionalOptions: Partial<RedisOptions> = {},
 ): Redis | Cluster | null => {
@@ -249,11 +287,11 @@ export const createNewRedisInstance = (
   }
 
   if (env.REDIS_CLUSTER_ENABLED === "true") {
-    return createRedisClusterInstance(additionalOptions);
+    return trackRedisInstance(createRedisClusterInstance(additionalOptions));
   }
 
   if (env.REDIS_SENTINEL_ENABLED === "true") {
-    return createRedisSentinelInstance(additionalOptions);
+    return trackRedisInstance(createRedisSentinelInstance(additionalOptions));
   }
 
   const tlsOptions = buildTlsOptions();
@@ -280,7 +318,7 @@ export const createNewRedisInstance = (
     logRedisError("Redis error", error);
   });
 
-  return instance;
+  return trackRedisInstance(instance);
 };
 
 /**
--- a/packages/shared/src/server/redis/redis.ts
+++ b/packages/shared/src/server/redis/redis.ts
@@ -254,10 +254,9 @@ const trackRedisInstance = <T extends Redis | Cluster | null>(
 };
 
 /**
- * Disconnect every client created by createNewRedisInstance.
- *
- * `redis.disconnect()` on the shared client only covers that one connection.
- * Returns how many were closed, for shutdown logging.
+ * Disconnect every client created by createNewRedisInstance, including the
+ * shared `redis` client, which is created through the same path. Returns how
+ * many were closed, for shutdown logging.
  */
 export const disconnectAllRedisInstances = (): number => {
   let closed = 0;
--- a/worker/src/utils/shutdown.ts
+++ b/worker/src/utils/shutdown.ts
@@ -1,5 +1,8 @@
 import { ClickHouseClientManager, logger } from "@langfuse/shared/src/server";
-import { redis } from "@langfuse/shared/src/server";
+import {
+  redis,
+  disconnectAllRedisInstances,
+} from "@langfuse/shared/src/server";
 
 import { ClickhouseWriter } from "../services/ClickhouseWriter";
 import { setSigtermReceived } from "../features/health";
@@ -93,7 +96,12 @@ export const onShutdown: NodeJS.SignalsListener = async (signal) => {
   logger.info("Clickhouse writer has been shut down.");
 
   redis?.disconnect();
-  logger.info("Redis connection has been closed.");
+  // Each queue holds its own client; without this they stay connected, retry
+  // forever, and keep the event loop alive so the process never exits.
+  const closedRedisConnections = disconnectAllRedisInstances();
+  logger.info(
+    `Redis connections have been closed (${closedRedisConnections} clients).`,
+  );
 
   await prisma.$disconnect();
   logger.info("Prisma connection has been closed.");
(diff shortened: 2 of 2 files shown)


Steffen911:
@closset703 Thank you for the contribution! The change seems very reasonable. Could you sign the CLA? Otherwise, we won't be able to merge it.

closset703:
> @closset703 Thank you for the contribution! The change seems very reasonable. Could you sign the CLA? Otherwise, we won't be able to merge it.

@Steffen911, Oh I did it , thanks

Steffen911:
@claude review

Steffen911:
@closset703 Thank you for the contribution! 🙌

Issue #17055, opened 2026-09-04
Sample observation selector in evaluator testing uses checkboxes for a mutually exclusive selection


ktachibana-ca:
## Summary

In the "Test with sample observations" panel of the evaluator setup
(`EvaluatorSampleObservationSelector`), the leading column of the matching
observations table renders a `Checkbox` per row. However, the selection is
mutually exclusive: checking one row unchecks the previously selected one.

A control for "pick exactly one of N" is what a radio button denotes.
Using a checkbox here signals that multiple rows can be selected, which
is not the case.

![Image](https://github.com/user-attachments/assets/78a60fd6-d41a-4e74-830c-4f9025077b63)

## Why this is not just cosmetic

The design-system `Checkbox` wraps `@radix-ui/react-checkbox`, so each row
is exposed to assistive technology as `role="checkbox"` with `aria-checked`.
Screen reader users are told the rows are independently selectable, and
then find that selecting one silently deselects another. The ARIA Authoring
Practices Guide assigns mutually exclusive selection to the radio pattern.

## Note on the "checkbox allows deselecting" argument

Checkboxes are sometimes preferred over radios for single selection when
the user must be able to return to a "nothing selected" state, since a
radio cannot be unchecked once set. That does not apply here:

* `onCheckedChange` calls `onSelect(row.original)` regardless of the checked
  state, so clicking the selected row does not clear it.
* `resolveSelection` auto-selects the first observation when nothing is
  selected and matches exist.

The state is therefore always exactly one selected observation, never zero.
The 0..1 case that would justify a checkbox does not occur.

## Suggested change

Replace the checkbox with a radio control. A `RadioGroup` /
`RadioGroupItem` pair already exists in `web/src/components/ui/radio-group.tsx`
and is used elsewhere under `features/evals/v2`.

Note that this is not confined to `EvaluatorSampleObservationSelector`:
the table itself is rendered by the shared `SampleObservationSelectorBase`,
so a small change there is likely needed as well. I have not proposed a
specific approach here, as I would rather agree on the direction first.

## Contributing

Happy to open a PR for this if you agree with the direction.

---

## 🤖 AI post-context

*Written by Pi on 2026-09-07.*

### Decisions, and why — reversals included

1. **Use radio buttons for the exactly-one selection.** The selector always resolves to one observation when matches exist, so checkbox semantics were misleading.
[...]

PR #17098, opened 2026-09-05
fix(evals): use radio group semantics for sample observation selection

Clsital:
### What

The leading column of `EvaluatorSampleObservationSelector` rendered a `Checkbox` per row, but the selection is mutually exclusive (and always exactly one observation is selected — `resolveSelection` auto-selects the first match). This replaces the checkbox with a radio control so assistive technology is told the truth about the selection model.

- `SampleObservationSelectorBase` gains an optional `selectionControl: "checkbox" | "radio"` prop (default `"checkbox"`, so `RuleSampleObservationSelector` and every other consumer are unchanged). In radio mode the table area is wrapped in a `RadioGroup.Root` (radix items need the root context) and selection changes flow through the existing `onSelect`.
- `EvaluatorSampleObservationSelector` renders `RadioGroup.Item` and passes `selectionControl="radio"`.
- `RadioGroupProps` in the design-system `RadioGroup` now accepts `className` (the radix root already spread it; the Pick type just didn't include it), needed to make the wrapper layout-neutral (`display: contents`).

### Why

Reported in #17055: screen reader users hear `role="checkbox"` with `aria-checked` and are told rows are independently selectable, then selecting one silently deselects another. The ARIA Authoring Practices Guide assigns mutually exclusive selection to the radio pattern.

### How tested

- `pnpm --filter web run typecheck`: the three touched files produce no errors (the remaining local errors are pre-existing environment noise from missing generated artifacts, unrelated to this change; CI runs the full check).
- `prettier --check` on the touched files: clean.
- Existing behavior is unchanged for consumers that don't opt into `selectionControl="radio"`: the wrapper only mounts in radio mode.

A note on the issue's suggestion: the `RadioGroup` component lives at `web/src/components/design-system/RadioGroup/RadioGroup.tsx` (the issue's path was slightly off), and since radix `RadioGroup.Item` needs a `Root` ancestor, the change is not confined to the selector cell — hence the small `SampleObservationSelectorBase` addition.

This PR changes evaluator sample-observation selection from checkbox to radio semantics while preserving checkbox behavior for existing consumers.

- Adds an optional selection-control mode to the shared observation selector.
- Wraps the evaluator observation table in a controlled Radix radio group.
- Extends the design-system radio-group props to accept layout classes.
- The new radio group still needs an accessible group-level name.
</details>

The PR appears safe to merge after addressing the non-blocking accessibility omission on the new radio group.

Selection behavior remains correctly connected to the existing observation state, but the newly introduced radio group lacks a group-level accessible name, reducing the clarity of the intended screen-reader improvement.

**Files Needing Attention:** web/src/features/evals/v2/components/Evaluators/Testing/components/SampleObservationSelectorBase/SampleObservationSelectorBase.tsx
</details>

`````markdown
### Issue 1
web/src/features/evals/v2/components/Evaluators/Testing/components/SampleObservationSelectorBase/SampleObservationSelectorBase.tsx:636-648
**Radio group lacks name**

The new `RadioGroup` provides `className`, `value`, and change handling, but no `aria-label` or `aria-labelledby`. The individual radio items are labeled, but assistive technology cannot announce what the group represents when a user enters it, reducing the clarity of the sample-selection control.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(evals): use radio group semantics fo..."](https://github.com/langfuse/langfuse/commit/c5d7723c5024cf2824ecadfc53cd6b1350095189) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60720507)</sub>

**Context used:**

### DIFF
--- a/web/src/components/design-system/RadioGroup/RadioGroup.tsx
+++ b/web/src/components/design-system/RadioGroup/RadioGroup.tsx
@@ -5,7 +5,7 @@ import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
 
 type RadioGroupProps = Pick<
   React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root>,
-  "children" | "defaultValue" | "onValueChange" | "value"
+  "children" | "className" | "defaultValue" | "onValueChange" | "value"
 >;
 
 const RadioGroupRoot = React.forwardRef<
--- a/web/src/components/design-system/RadioGroup/RadioGroup.tsx
+++ b/web/src/components/design-system/RadioGroup/RadioGroup.tsx
@@ -5,7 +5,12 @@ import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
 
 type RadioGroupProps = Pick<
   React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root>,
-  "children" | "className" | "defaultValue" | "onValueChange" | "value"
+  | "aria-label"
+  | "children"
+  | "className"
+  | "defaultValue"
+  | "onValueChange"
+  | "value"
 >;
 
 const RadioGroupRoot = React.forwardRef<
--- a/web/src/components/design-system/RadioGroup/RadioGroup.tsx
+++ b/web/src/components/design-system/RadioGroup/RadioGroup.tsx
@@ -2,30 +2,43 @@
 
 import * as React from "react";
 import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
+import { cva } from "class-variance-authority";
+import { cn } from "@/src/utils/tailwind";
+
+const radioGroupVariants = cva("grid gap-2", {
+  variants: {
+    layout: {
+      stack: "",
+      // RadioGroup.Root rendered as a layout-neutral wrapper (used when the
+      // radio items live inside a larger block, e.g. a data-table column)
+      inline: "contents",
+    },
+  },
+  defaultVariants: { layout: "stack" },
+});
 
 type RadioGroupProps = Pick<
   React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root>,
-  | "aria-label"
-  | "children"
-  | "className"
-  | "defaultValue"
-  | "onValueChange"
-  | "value"
->;
+  "aria-label" | "children" | "defaultValue" | "onValueChange" | "value"
+> & { layout?: "stack" | "inline" };
 
 const RadioGroupRoot = React.forwardRef<
   React.ComponentRef<typeof RadioGroupPrimitive.Root>,
   RadioGroupProps
->((props, ref) => {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: web/src/features/evals/v2/components/Evaluators/Testing/components/SampleObservationSelectorBase/SampleObservationSelectorBase.tsx [hunk 1]
```diff
@@ -585,47 +633,22 @@ export function SampleObservationSelectorBase(
...
+            value={selectedObservationId ?? ""}
+            onValueChange={(observationId) => {
+              const observation = matchingObservations.find(
+                (item) => item.id === observationId,
+              );
+              if (observation) onSelect(observation);
+            }}
+          >
+            {observationTableArea}
+          </RadioGroup>
```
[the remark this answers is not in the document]
Clsital: Good catch — addressed in 26c6a5a: the radio group root now carries an accessible name (`aria-label="Use an observation as sample"`), configurable via a new optional `selectionControlLabel` prop on `SampleObservationSelectorBase`, with a generic default for other consumers.

web/src/components/design-system/RadioGroup/RadioGroup.tsx
```diff
@@ -5,7 +5,12 @@ import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
 
 type RadioGroupProps = Pick<
   React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root>,
-  "children" | "defaultValue" | "onValueChange" | "value"
+  | "aria-label"
+  | "children"
+  | "className"
```
bezbac: Hi @luo22833548, this component is not allowed to have a `className` prop. Please see https://github.com/luo22833548/langfuse/blob/26c6a5a3a805bf0e9a55957b4ccbb57149dd8e8d/web/src/components/design-system/README.md

bezbac (changes requested):
Hi @luo22833548, I got automatically assigned this PR because of the changes to the `components/design-system` code. Please see my comment below.

web/src/components/design-system/RadioGroup/RadioGroup.tsx
Clsital: Thanks for the pointer — you're right, `className` violates the design-system rules. I've replaced it with an explicit `layout: "stack" | "inline"` variant via cva (per the README's styling conventions): the default stays `stack` with the existing `grid gap-2` look for current consumers, and `inline` renders the root as a layout-neutral wrapper for the data-table case. No behavior change for the existing `EvaluatorSavedDialog` usage.

Clsital:
Main has since landed #17134, which implements the radio control for the sample observation selector with a dedicated design-system component — thank you for taking the direction and including me as co-author. Closing this PR as superseded; the review feedback on `className` vs. design-system conventions was much appreciated.

PR #17136, opened 2026-09-07
feat(preview): alert when a labeled preview is not serving

LysanderKie:
🟡 **Live preview: [pr-17136 app preview](https://pr-17136.preview.langfuse.com)**

## Why

A preview can be labeled, resolve, and answer 503 while looking perfectly healthy from the cluster's point of view. Argo CD treats a Deployment scaled to zero as `Healthy`, and `/spec/replicas` is in the ApplicationSet's `ignoreDifferences` so the downscaler can move it, so nothing in the sync path notices. The only detector was someone opening the URL and saying so.

This is the second half of preview alerting. The first half (Slack on a *failed sync*, covering the seeder and dev-tables hooks) is `langfuse/infrastructure#1334`; the two do not overlap.

## What

`schedule` + `workflow_dispatch` only. No checkout, no PR-code execution, no cloud credentials, no third-party actions. Lists open PRs carrying the `preview` label with the default `GITHUB_TOKEN` (`pull-requests: read`), probes `https://pr-<n>.preview.langfuse.com/api/public/health`, posts to Slack on failure and fails the run so it also shows red in the Actions tab.

Gated on `vars.AWS_PREVIEW_ECR_PUSH_ROLE_ARN`, the same preview-system flag `preview-build` and `preview-stale-cleanup` use.

**Schedule.** `0 7,9,11,13,15,17 * * 1-5` — six probes, every two hours, 08:00-20:00 Europe/Berlin. GitHub cron is UTC and Berlin is UTC+1 or UTC+2 depending on DST, so a UTC hour `H` lands at `H+1` or `H+2` local; only 07:00-18:00 UTC is inside 08:00-20:00 Berlin on both sides of DST, which gives 07:00-17:00 on a two-hourly grid. First probe at 08:00 or 09:00 local, last at 18:00 or 19:00. That sits well inside the window where previews are up at all (`Mon-Fri 08:00-24:00 Europe/Berlin`), so no run ever lands on a deliberately sleeping preview and reports it broken.

**Two deliberate quiet cases**, both noise the first draft would have produced:

- **404** is the shared ALB's wildcard catch-all: labeled PR, nothing deployed. That's an author off the deploy allowlist or a teardown in flight, not a fault, so it's counted and ignored.
- **One failure is re-probed after 30s.** A rebuild rolls the web Deployment and a brief `ImagePullBackOff` is documented as normal and self-healing, so a single bad response is not enough to alert.

## Slack

Posts to a dedicated `#lf-preview-alerts` channel through an incoming webhook (`secrets.SLACK_PREVIEW_ALERTS_WEBHOOK_URL`), the same hook `langfuse/infrastructure#1334` uses for Argo sync failures — so both preview signals land in one place and neither competes with CI failures.

There was nothing to reuse: this repo has `SLACK_CI_FAILURE_WORKFLOW_WEBHOOK_URL` (bound to the CI-failure destination) and an unreferenced `SLACK_WEBHOOK_URL`, and the preview account's Secrets Manager holds only the GitHub PAT. A webhook beats a bot token here: bound to its channel, no app scope, no invite to keep working.

If the secret is unset the step warns and skips, so merging before it exists degrades to a red Actions run. The existing `notify-slack-failure` composite action isn't reused because it derives its text from failed *jobs* via Slack Workflow Builder, which doesn't fit "the job ran fine, the previews didn't".

## Noise at 20-50 previews

Alert volume is set by the schedule, not the fleet size:

- **One message per run**, listing every offender — never one per preview. Six runs a day is the ceiling no matter how many previews exist.
- **The list is capped at ten names** plus a `+N more` count, so a broken ALB can't paste fifty hostnames into the channel. The full list stays in the step summary.
- **The 30s re-probe wait is shared across all suspects**, not paid per preview. Fifty previews with fifteen broken costs one wait rather than fifteen; measured **45s end to end** against today's live fleet of eighteen with four broken.
- Unhealthy entries are space-free tokens (`pr-17125(503)`), because the count and the truncation split that list on whitespace. My first pass used `pr-17125(HTTP 503)` and consequently reported `4 of 18` as `8 of 18` — caught by testing the message builder at fifty entries.

What this does **not** do is de-duplicate across runs: a preview that stays broken all day is reported at every probe, up to six times. Fixing that needs persisted state (an `actions/cache` entry or similar), which I left out deliberately — it trades a known, bounded repetition for machinery. Easy to add if six a day turns out to grate.

## Verification

Extracted every `run` block and ran the probe logic verbatim against the live fleet (read-only; `gh api` + `curl`, no writes):

```
- serving: 14
- not deployed (404, ignored): 0
- not serving: pr-17125(HTTP 503) pr-17102(HTTP 503) pr-17081(HTTP 503) pr-17063(HTTP 503)
```

**It found four broken previews on the first run**, and spot-checking two confirms both are real and are the intended target rather than false positives:

- `pr-17125` — web and worker in `ImagePullBackOff` for 177 minutes. Well past the transient rollout window the 30s re-probe forgives.
- `pr-17063` — namespace present, **zero pods**, while Argo reports its last sync `Succeeded`. Exactly the blind spot: nothing in the sync path is unhappy.

`bash -n` clean on all three run blocks; the workflow parses as YAML.

## Tests

Tests added: 0. A GitHub Actions workflow has no unit-test seam in this repo, and the meaningful assertion is that the probe classifies the real fleet correctly — which is the run above, including two independently confirmed true positives and correct handling of the healthy majority.

Adds a scheduled and manually dispatchable GitHub Actions workflow that checks labeled preview environments and reports persistent health failures.
- Enumerates open pull requests carrying the `preview` label.
- Probes each preview health endpoint and retries failures after 30 seconds.
- Ignores the shared fallback’s 404 responses, posts persistent failures to Slack, and fails the workflow run.
- Restricts the schedule to hours safely inside the preview uptime window across Berlin daylight-saving changes.
- Correctly normalizes transport failures to HTTP `000`.
</details>

The PR appears safe to merge; no actionable correctness, security, or repository-rule violation remains.

The transport-error formatting issue was corrected by capturing curl’s output and supplying `000` only when that output is empty. Its previous thread was also manually resolved without explanation. The revised schedule remains inside the documented preview uptime window, and the reordered retry classification consistently treats a teardown that completes during the retry delay as undeployed.
</details>

```mermaid
flowchart TD
    Trigger[Schedule or workflow dispatch] --> Gate{Preview system enabled?}
    Gate -- No --> Stop[Skip job]
    Gate -- Yes --> List[List open PRs with preview label]
    List --> Probe[Probe preview health endpoint]
    Probe --> Result{HTTP status}
    Result -- 200 --> Healthy[Count as serving]
    Result -- 404 --> Ignored[Count as not deployed]
    Result -- Other --> Retry[Wait 30 seconds and retry]
    Retry --> RetryResult{Retry status}
    RetryResult -- 200 --> Healthy
    RetryResult -- 404 --> Ignored
    RetryResult -- Other --> Unhealthy[Record as not serving]
    Unhealthy --> Slack[Post Slack notification]
    Unhealthy --> Fail[Fail workflow run]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(preview): report one status code and..."](https://github.com/langfuse/langfuse/commit/bc00bfed23895d45c12a475a614dd4ffe1bd7fdc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61218720)</sub>


LysanderKie:
@greptileai review

.github/workflows/preview-health-check.yml
```diff
@@ -0,0 +1,154 @@
...
+            echo "- not serving: ${unhealthy:-none}"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Notify Slack
+        if: steps.probe.outputs.unhealthy != ''
+        env:
+          # Same bot and channel as the Argo CD preview sync-failure alerts, so
+          # every preview signal lands in one place.
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+          SLACK_CHANNEL: langfuse-engineering-notifications
```
LysanderKie: I would like to pass that into a special lf-preview-alerts channel, and lets check if token is already existing.

.github/workflows/preview-health-check.yml
```diff
@@ -0,0 +1,154 @@
...
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+
+          if [ -z "${SLACK_BOT_TOKEN:-}" ]; then
+            echo "::warning::SLACK_BOT_TOKEN is unset; skipping the Slack post."
+            exit 0
+          fi
+
+          text="Preview environments labeled but not serving: ${UNHEALTHY}. Run: ${RUN_URL}"
```
LysanderKie: Also lets check that htis is not too noisy when we would have 20-50 preview deployments

PR #17150, opened 2026-09-08
fix(web): percent-encode filter object keys in URL filter serialization

### LINKED ISSUE #17149: bug: Filter URL breaks when a metadata/object key contains `; `, `, ` or `%`
### Describe the bug

`encodeFiltersGeneric` writes the filter `key` slot (metadata keys, score keys, etc.) raw into the `column;type;key;operator;value` filter URL format, while `decodeFiltersGeneric` runs `decodeURIComponent` on that same slot. Object keys are arbitrary user strings (metadata keys can be any JSON key), so a key containing a format delimiter breaks the round-trip:

- A key containing `;` or `,` shifts the delimited fields on decode. The filter then fails validation and is silently dropped. Example: key `a;b` comes back as key `a` with operator `b`, and neighboring filters in the same URL can be corrupted.
- A key containing `%` makes `decodeURIComponent` throw `URIError: URI malformed` out of `decodeFiltersGeneric`. The sidebar hook's catch-all then drops every filter in the URL, not just the broken one. `buildEventsTablePathForMetadataFilter` calls `decodeFiltersGeneric` without a try/catch, so clicking include/exclude on a metadata value can throw outright once such a filter is in the URL.

### To reproduce

```ts
encodeFiltersGeneric([
  { column: "metadata", type: "stringObject", key: "a;b", operator: "=", value: "x" },
]);
// "metadata;stringObject;a;b;=;x" -> decodes as key "a", operator "b" -> invalid -> filter silently dropped

decodeFiltersGeneric("metadata;stringObject;cpu%;=;x");
// URIError: URI malformed
```

In the UI:

1. On the events/observations table, use include/exclude on a metadata value whose key contains `;` or `,` (or paste a filter URL with such a key).
2. Reload the page: the filter is silently gone.
3. With a `%` in the key, reloading drops all filters from the URL.

### Expected behavior

The key survives the encode -> decode round-trip, the same way values with pipe characters were fixed in #11757. The decode side already expects a percent-encoded key (`decodeURIComponent(key)`), the encode side just never encodes it.

### Related

- #11757 (same bug class for `|` in values)

Harsh23Kashyap:
## What this fixes

Fixes #17149

`encodeFiltersGeneric` interpolates the `key` slot (metadata / score / custom object keys) into the semicolon-delimited filter query format unescaped, while `decodeFiltersGeneric` runs `decodeURIComponent` on that slot. As a result:

- keys containing `;` or `,` shift the delimited fields on decode - the filter fails validation and is silently dropped (and can corrupt neighboring filters in the same URL),
- keys containing `%` make `decodeURIComponent` throw `URIError` out of `decodeFiltersGeneric`, which wipes every filter in the URL via the sidebar hook's catch-all, or throws uncaught from `buildEventsTablePathForMetadataFilter`.

## Change

One line in `packages/shared/src/features/filters/filterQueryEncoding.ts`: `encodeURIComponent(key)` when writing the key slot, matching what the decode side already expects.

Plain keys are unaffected (`encodeURIComponent` is identity for the key alphabet used across every existing test). URLs written before this change decode exactly as before for all keys that used to work; keys with `%` could never round-trip, so nothing regresses there.

## Tests

- Added two regression tests to `web/src/features/filters/lib/filter-query-encoding-decoding.clienttest.ts`: round-trip for object keys containing `;`, `,`, `%`, `|`, `\`, and a multi-filter case proving a delimiter in one key no longer corrupts neighboring filters.
- Filter-area suites pass: 74 tests across 6 files, including the full existing encoding/decoding suite (41 tests) and the consumers of `encodeFiltersGeneric` (`eventsTablePaths`, `buildTableFilterHref`, `evaluatorScoresUrl`, `ruleExecutionsUrl`, `evaluatorMigrationUrls`).
- Caveat: the full monorepo `pnpm install` would not complete in my environment (network), so vitest ran against the repo's real source files with `zod`/`date-fns` resolved from a scratch install. The React/jsdom sidebar suites were not run; they don't exercise the changed line, but a CI run should confirm.

This PR makes object-key serialization symmetric with the existing decoder by percent-encoding metadata, score, and custom filter keys before inserting them into the delimiter-based URL format.
- Prevents semicolons and commas in keys from corrupting filter boundaries.
- Prevents literal percent signs in keys from causing decoding failures.
- Adds direct round-trip and neighboring-filter regression coverage.
</details>

The PR appears safe to merge, with the encoder change correctly matching the existing decoding contract.

No actionable failures remain; delimiter-containing and percent-containing keys are encoded before tokenization, decoded once afterward, and covered by focused regression tests.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): percent-encode filter object k..."](https://github.com/langfuse/langfuse/commit/907b6f4c77a63aac45c0a0d03566025197acb310) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61462309)</sub>


nimarb:
slop

PR #17154, opened 2026-09-08
fix(web): avoid key collisions in score comparison confusion matrix

### LINKED ISSUE #17153: Score comparison confusion matrix shows wrong counts when category names contain '-'
### Describe the bug

In the score analytics comparison view, `generateConfusionMatrixData` builds its lookup map by concatenating the row and column categories with a `-` delimiter:

```ts
dataMap.set(`${d.row_category}-${d.col_category}`, d.count);
// ...
const count = dataMap.get(`${rowCat}-${colCat}`) || 0;
```

(web/src/features/score-analytics/lib/heatmap-utils.ts)

The categories are categorical score string values coming from ClickHouse (`string_value1`/`string_value2` in buildScoreComparisonQuery.ts), which are user-defined and can contain `-` (e.g. "high-risk", "not-passed").

When category names contain the delimiter, two different (row, column) pairs produce the same key and collide. The later entry overwrites the earlier one in the map, and both cells render the same count.

### To reproduce

```ts
generateConfusionMatrixData({
  data: [
    { row_category: "a-b", col_category: "c", count: 10 },
    { row_category: "a", col_category: "b-c", count: 99 },
  ],
});
```

Both pairs produce the map key `"a-b-c"`. The cell for (`a-b`, `c`) renders `99` instead of `10`.

### Expected behavior

Each (row, column) pair keeps its own count regardless of the characters in the category names.

### Fix

Key the map on `JSON.stringify([row_category, col_category])` instead of the concatenated string. PR with the fix and a regression test follows.

Harsh23Kashyap:
## What does this PR do?

Fixes #17153

`generateConfusionMatrixData` keyed its lookup map on `` `${row_category}-${col_category}` ``. Since category names are user-defined categorical score values and can contain `-`, distinct (row, column) pairs could produce the same key, and the affected cells rendered the wrong count in the score comparison confusion matrix.

The map is now keyed on `JSON.stringify([row_category, col_category])`, which is unambiguous for any pair of strings.

## Type of change

## How was this tested?

- Regression test added to `web/src/__tests__/heatmap-utils.clienttest.ts` ("should not collide cells when categories contain the key delimiter"): fails before the fix (cell for `a-b`/`c` returns 99 instead of 10), passes after.
- `heatmap-utils.clienttest.ts`: 14/14 tests pass (13 existing + 1 new).

Test caveat: I could not install the full monorepo dependencies in my environment (pnpm install repeatedly stalled on network), so I ran the relevant vitest suite directly against the repo sources with minimal module aliases (zod, `@langfuse/shared` shim, `@/src` alias), mirroring the suite's node environment. The touched module is dependency-free. As an additional sanity sweep, the other web client tests runnable this way also pass: 72/72 across filter-query-encoding-decoding, buildTableFilterHref, eventsTablePaths, evaluatorScoresUrl, ruleExecutionsUrl, and evaluatorMigrationUrls.

This PR fixes score-comparison confusion-matrix key collisions by replacing delimiter-concatenated lookup keys with serialized string tuples.
- Uses the same unambiguous tuple encoding for map insertion and lookup.
- Adds a focused regression test covering two category pairs that previously generated the same key.
</details>

The PR appears safe to merge; the new key representation is unambiguous for the function’s string inputs and is covered by an effective regression test.

Map writes and reads now use the same serialized two-string tuple, eliminating the reported delimiter collision without changing category enumeration, counts, or percentage calculations.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): use tuple key for confusion ma..."](https://github.com/langfuse/langfuse/commit/b3790db9a4bba93b113b6a8b9896d1b7b31783f1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61466531)</sub>

### DIFF
--- a/.agents/skills/datadog-query-recipes/SKILL.md
+++ b/.agents/skills/datadog-query-recipes/SKILL.md
@@ -41,6 +41,8 @@ data domain you need: traces, logs, metrics, and visualizations.
      [`references/public-api-tenant-usage.md`](references/public-api-tenant-usage.md)
    - Queue inventory, queue consumers, and queue metrics:
      [`references/queue-consumers.md`](references/queue-consumers.md)
+   - Scheduled-export freshness lag (blob / PostHog / Mixpanel):
+     [`references/export-freshness-lag.md`](references/export-freshness-lag.md)
 3. Start with aggregate queries, grouped by environment, service, route,
    queue, project, org, status, or error facets as appropriate.
 4. Fetch raw spans, logs, or traces only after aggregation identifies the
--- a/.agents/skills/datadog-query-recipes/references/export-freshness-lag.md
+++ b/.agents/skills/datadog-query-recipes/references/export-freshness-lag.md
@@ -1,0 +1,43 @@
+# Export freshness lag
+
+Scheduled integrations (blob, PostHog, Mixpanel) emit a pooled distribution at
+the end of each run. Provenance: the shape from LFE-15853.
+
+## Metric
+
+- Name: `langfuse.export.freshness_lag_seconds`
+- Type: distribution
+- Value: `runStartTime − maxExportedTimestamp` (seconds, clamped at 0)
+- Watermark: `lastSyncAt` after a successful window (`maxTimestamp` just
+  written). Failing runs use the **unchanged** watermark so stall shows as
+  rising lag. Do not treat intended window end as the watermark.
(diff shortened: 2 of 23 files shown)


marliessophie:
Thank you for the contribution!

Harsh23Kashyap:
Thanks for the quick review and merge, happy to help!

PR #17158, opened 2026-09-08
fix(web): close gaps between multi-unit time series buckets

### LINKED ISSUE #17157: Score analytics charts silently drop data points for multi-unit time intervals
### Describe the bug

In score analytics, when the selected time range resolves to a multi-unit interval (e.g. 2 days), the ClickHouse query returns single-unit points and the frontend aggregates them into buckets (`aggregateIntoMultiUnitBuckets` in web/src/utils/fill-time-series-gaps.ts, and the same construction in `aggregateCategoricalIntoMultiUnitBuckets`).

Each bucket is built by going back `count - 1` units from its end, but the next bucket's end is then stepped back by a full `count` units:

```ts
let bucketStart = bucketEnd;
for (let i = 0; i < count - 1; i++) {
  bucketStart = subtractSingleUnit(bucketStart);
}
// ...
for (let i = 0; i < count; i++) {
  bucketEnd = subtractSingleUnit(bucketEnd);
}
```

So consecutive buckets are not contiguous: there is a one-unit gap between them. Single-unit points whose timestamp falls into a gap match no bucket and are silently dropped from the chart.

### To reproduce

Five daily points (midnight-aligned, as returned by ClickHouse) with counts 10, 20, 30, 40, 50; range Jan 1 00:00 to Jan 5 14:30; interval `{ count: 2, unit: "day" }`:

```ts
fillTimeSeriesGaps(data, fromDate, toDate, { count: 2, unit: "day" });
// buckets: [Jan 4 14:30, Jan 5 14:30], [Jan 2 14:30, Jan 3 14:30], ...
// Jan 2 (20) and Jan 4 (40) fall into the gaps -> output counts [10, 30, 50]
```

A third of the data vanishes from the chart; for a 5-day interval the pattern repeats per bucket.

### Expected behavior

Buckets tile contiguously back from `toDate`, each spanning exactly `count` units, so every point in range lands in exactly one bucket: output counts [10, 25, 45].

### Fix

Span each bucket by `count` units (not `count - 1`) and assign points newest-first so a point exactly on a shared boundary goes to the later bucket while the rightmost bucket still includes `toDate`. Same fix in both the numeric and the categorical variant. PR with regression tests follows.

Harsh23Kashyap:
## What does this PR do?

Fixes #17157

`aggregateIntoMultiUnitBuckets` (and the identical construction in `aggregateCategoricalIntoMultiUnitBuckets`) built each bucket by going back `count - 1` units from its end while stepping the next bucket end back by `count` units. Consecutive buckets were therefore not contiguous - a one-unit gap separated them - and single-unit data points landing in the gap matched no bucket and were silently dropped from score analytics charts. Example: 5 daily points with a 2-day interval and a mid-day range end lose 2 of the 5 points.

Buckets now span exactly `count` units so they tile contiguously back from `toDate`, and point assignment iterates newest-first so a point on a shared boundary belongs to the later bucket while the rightmost bucket still includes `toDate`.

## Type of change

## How was this tested?

- New suite `web/src/utils/fill-time-series-gaps.clienttest.ts` (the module had no tests): 3 tests covering the numeric path (5 daily points over 2-day buckets keep all data: averages [10, 25, 45]), the categorical path (counts summed into contiguous buckets: [3, 7]), and the shared-boundary case (a point on a boundary goes to the later bucket). The first two fail on the unfixed code ([10, 30, 50] and [1, 3]-style losses), all pass with the fix.
- Sanity sweep of the other web client tests runnable in my environment: 72/72 pass (filter-query-encoding-decoding, buildTableFilterHref, eventsTablePaths, evaluatorScoresUrl, ruleExecutionsUrl, evaluatorMigrationUrls).

Test caveat: I could not install the full monorepo dependencies in my environment (pnpm install repeatedly stalled on network), so I ran vitest directly against the repo sources with minimal module aliases (zod, `@langfuse/shared` shim, `@/src` alias), mirroring the suites' node environment. The touched module only depends on date-fns and a local type import.

This PR makes numeric and categorical multi-unit time-series buckets contiguous and assigns shared-boundary points to the newer bucket.

- Extends each bucket to span exactly the configured number of units.
- Reverses assignment traversal to establish newer-bucket boundary ownership.
- Adds numeric, categorical, and shared-boundary regression coverage.
- The aligned-range endpoint semantics still make the rightmost bucket contain one sample too many.
</details>

The PR is not yet safe to merge because unit-aligned ranges can produce incorrect averages and categorical totals in the newest bucket.

A full multi-unit bucket now includes both its start and end when it is rightmost, so an interval configured for `count` normalized samples can aggregate `count + 1` samples.

**Files Needing Attention:** web/src/utils/fill-time-series-gaps.ts, web/src/utils/fill-time-series-gaps.clienttest.ts
</details>

`````markdown
### Issue 1
web/src/utils/fill-time-series-gaps.ts:233-235
**Rightmost bucket overcounts**

When `toDate` is aligned to the aggregation unit, the rightmost bucket now spans `count` full units while the inclusive end comparison admits both boundary points. For example, a two-day interval ending at Jan 5 midnight puts the normalized Jan 3, Jan 4, and Jan 5 points into the same bucket. This produces a three-day average or sum instead of a two-day result. The same issue applies to the categorical path. Please use consistent half-open boundaries or otherwise handle the final endpoint so each full bucket contains only `count` samples.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): close gaps between multi-unit ..."](https://github.com/langfuse/langfuse/commit/346b095da5de8300ca189f0dd31e356c2b11e0c0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61468979)</sub>


Harsh23Kashyap:
Good catch - valid. The query is `timestamp <= toTimestamp` (inclusive), so with an aligned toDate the rightmost bucket could absorb `count + 1` unit-points. Amended: bucket assignment now uses `(start, end]` semantics, so a shared boundary belongs to the earlier bucket and every bucket, including the rightmost, holds at most `count` unit-points. Added a regression test for exactly this case (aligned toDate, 2-day interval, 5 daily points -> bucket averages [10, 25, 45], no three-point bucket). All 5 tests in fill-time-series-gaps.clienttest.ts pass, plus the 72 other web client tests runnable in my environment.

Steffen911:
slop

PR #17181, opened 2026-09-08
fix(web): percent-encode object keys in v3 table filter URL serialization

### LINKED ISSUE #17180: bug: v3 table filter URLs silently drop filters when a metadata/object key contains ';'
### Describe the bug

On the v3 tables (traces, observations, sessions, scores, prompts, users, datasets, experiments), the filter state is serialized into the `?filter=` URL param by `getCommaArrayParam` in `useFilterState.ts`, using the `column;type;key;operator;value` format. The `key` slot (metadata keys, score keys, etc.) is written raw. Object keys are arbitrary user strings, so a key containing the field separator `;` shifts the fields on decode: the entry then fails `singleFilter` validation and is **silently dropped**. Reloading the page or sharing the link loses the filter without any indication.

Same format issue as #17149, which fixed the shared/v4 encoder (`filterQueryEncoding.ts`) - this is the parallel v3 implementation, which never got the fix.

### To reproduce

```ts
const state = [
  { column: "Metadata", type: "stringObject", key: "a;b", operator: "=", value: "x" },
];

encode(state); // "metadata;stringObject;a;b;=;x"
decode("metadata;stringObject;a;b;=;x");
// splits into 6 fields -> key "a", operator "b" -> invalid -> filter dropped: []
```

Verified with a vitest repro against the current sources: `decode(encode(state))` returns `[]` (the filter is gone).

### Expected behavior

Filters with arbitrary object keys survive the URL round-trip on all tables.

### Environment

- Langfuse version: main @ c9c54f5 (OSS, self-hosted + cloud)

Harsh23Kashyap:
## What does this PR do?

Fixes #17180

`getCommaArrayParam` (the `?filter=` serializer behind `useQueryFilterState`, used by the v3 traces/observations/sessions/scores/prompts/users/datasets/experiments tables) wrote the `key` slot of the `column;type;key;operator;value` format raw. A metadata or score key containing `;` shifted the delimited fields on decode, the entry failed `singleFilter` validation, and the filter was **silently dropped** on reload or when opening a shared link.

This is the same format issue as #17149, fixed there for the shared/v4 encoder (`filterQueryEncoding.ts`); this PR brings the parallel v3 implementation in line.

### Change

- Encode: `encodeURIComponent(f.key)` for the keyed filter types (`stringObject` / `numberObject` / `booleanObject` / `categoryOptions` / `positionInTrace`).
- Decode: decode the key with a tolerant helper that falls back to the raw value, so legacy or hand-written URLs with unencoded keys (e.g. a raw `%`, which `decodeURIComponent` would throw on) keep working instead of dropping the whole param.
- No behavior change otherwise: the pure encoder/decoder was extracted from `useFilterState.ts` into `web/src/features/filters/lib/v3-filter-query-encoding.ts` so it is unit-testable; the hook now imports it.

This also keeps the cross-flow consistent with #17150: links built by `buildTableFilterHref` (shared encoder, keys percent-encoded) are now decoded symmetrically when they target a v3 table.

## How was this tested?

- New `v3-filter-query-encoding.clienttest.ts` (4 tests, all pass):
  - key containing `;` round-trips (was silently dropped before the fix: `decode(encode(state)) === []`),
  - key with `%`, `&`, space, `;` round-trips,
  - plain keys still round-trip,
  - legacy unencoded `%` key decodes without throwing.
- Existing filter URL suite: 72/72 tests pass (`filter-query-encoding-decoding`, filter integration, sidebar filter state, session persistence, saved-view regression).
- Pre-fix failure verified by running the same round-trip against the pre-change logic: filter dropped (`[]`).

Caveat: the repo's full `tsc`/CI setup could not run in my environment (monorepo install stalls); tests ran through a scoped vitest config aliasing the real repo sources. The change is confined to the extracted module, its test, and import updates in the hook.

## Type of change

This PR extracts the v3 table-filter URL codec and percent-encodes keyed-filter keys so delimiters such as semicolons survive serialization.

- Adds tolerant key decoding for legacy and hand-written URLs.
- Adds unit coverage for reserved characters, plain keys, and malformed legacy percent syntax.
- The compatibility path still changes legacy keys containing syntactically valid percent escapes.
</details>

The PR is not yet safe to merge because existing URLs with literal `%HH` sequences in object keys can silently target a different field.

The unconditional key decode fixes delimiter-containing keys but converts valid percent-escape-looking text in legacy raw keys, violating the stated backward-compatibility behavior.

**Files Needing Attention:** web/src/features/filters/lib/v3-filter-query-encoding.ts; web/src/features/filters/lib/v3-filter-query-encoding.clienttest.ts
</details>

`````markdown
### Issue 1
web/src/features/filters/lib/v3-filter-query-encoding.ts:93
**Legacy keys change meaning**

Legacy URLs can contain literal object keys such as `load%20avg`, because the previous encoder wrote keys without percent-encoding them. Passing such a valid escape sequence through `safeDecodeURIComponent` changes the key to `load avg`, so reloading or sharing the URL silently filters a different metadata or score field. This representation is indistinguishable from a newly encoded key containing a space, so preserving both formats requires explicit format or version handling rather than decoding every valid escape sequence.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(web): percent-encode object keys in ..."](https://github.com/langfuse/langfuse/commit/83a8141364598089625dd44268be5bb2a8f265b7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61545727)</sub>


Harsh23Kashyap:
On the greptile P1 ("Legacy keys change meaning") - verified against the code, and the observation is correct: a legacy raw key that happens to be a valid escape sequence (e.g. `load%20avg`) is byte-identical to a newly encoded key containing a space, so no decoder can distinguish the two without a format version marker.

Keeping full `decodeURIComponent` semantics anyway, for two reasons:

1. It matches `decodeFiltersGeneric` in `packages/shared` (the v4 path), which already decodes the key slot - this PR keeps the two filter URL implementations consistent, and it keeps cross-links built by `buildTableFilterHref` (shared encoder, percent-encoded keys) decodable when they land on a v3 table. A narrower escape (e.g. only `%3B`) would fix `;` but break that cross-flow for keys with spaces or `%`.
2. The `safeDecodeURIComponent` fallback covers the legacy case that would otherwise throw `URIError` and drop the entire filter param (raw `%` without valid hex, e.g. `100%`).

Residual risk: a legacy shared/bookmarked URL whose object key literally contains a valid `%XX` sequence now decodes to a different field. That ambiguity already exists on the shipped v4 path, and any edit in the filter editor re-encodes the URL, so the exposure is limited to old saved links. Happy to add explicit format versioning instead if maintainers prefer that trade.

PR #17188, opened 2026-09-08
feat(ai-gateway): bootstrap Rust service and development workflow

hassiebp:
🟡 **Live preview: [pr-17188 app preview](https://pr-17188.preview.langfuse.com)**

## What does this PR do?

Adds the first runnable foundation for the Rust AI gateway: a standalone Axum/Tokio service with health/readiness probes, validated configuration, configurable text/JSON lifecycle logs, and bounded SIGTERM/SIGINT draining. Inference paths return 404; authentication and provider transport follow in later slices.

The gateway joins the pnpm workspace as a thin Cargo command wrapper. Root `pnpm dev` uses Turbo Watch to start the development stack and recompile/restart the gateway after Rust edits. Gateway startup loads the root `.env`; colleagues need rustup and a native compiler/linker, documented in CONTRIBUTING and the gateway README. Gateway build/check/lint/test tasks have no Prisma or JavaScript dependencies and leave incremental compilation to Cargo; existing JavaScript CI and commit hooks retain their scope. Gateway configuration is documented in its README.

Includes an ephemeral-port HTTP test helper, a non-root Cargo-only runtime image, and a dedicated Rust/image job in the required CI gate. Local development uses pnpm/Turbo; the runtime image retains a simple builder/runtime Dockerfile. Web/worker Docker contexts exclude Cargo artifacts.

## Type of change

- [x] New feature
- [x] Development tooling and CI integration
- [x] Documentation in `ai-gateway/README.md`

## Mandatory Tasks

- [x] Self-reviewed and independently reviewed lifecycle, CI and tooling integration.

## Validation

- Gateway Turbo lint/typecheck/test/release build: `Tasks: 4 successful, 4 total`; `Cached: 0 cached, 4 total`. Cargo tests: 11 passed, 0 failed.
- Real Turbo dry graphs: gateway tasks have no JS/Prisma dependencies; default dev includes gateway and JS CI selections exclude it.
- Isolated native development test using the actual scripts: root `.env` loading, source-change recompilation/restart and listener cleanup all passed.
- Whole-repo Knip passed; changed package/Turbo JSON formatting passed; frozen lockfile, workflow/workspace YAML, hook syntax and diff whitespace passed.
- Simplified runtime image rebuilt locally: `Image smoke: non-root, probes, unimplemented inference, SIGTERM passed`.
- Development Compose configuration validates after removing the gateway service and Watch profile; README/YAML Prettier and diff whitespace checks pass.

Tooling verification used pnpm 12.3.1, Turbo 2.10.5, Rust 1.98.0 and local Node 24.14.0; hosted CI uses its pinned Node setup. Pnpm regenerated the gateway importer and existing ESLint peer snapshots without changing pinned versions. The final root-command change was verified with a real gateway and isolated persistent peer fixtures using the repository Turbo configuration. No full Web/worker stack was launched; their existing watchers remain non-interruptible. Rust tests and runtime image smoke were rerun for the subsequent text/JSON logging change and passed. Unrelated Web/worker tests were not repeated locally; new hosted CI verifies the updated head. The commit hook also passed the repository-wide Prettier check; its JavaScript lint step was skipped by the pre-existing `LANGFUSE_PRE_COMMIT_SKIP_LINT` environment setting.

## Lifecycle boundary

When graceful shutdown exceeds its budget, `server::serve` returns an error and the executable exits its runtime to terminate remaining connection tasks. Future streaming/export work must join this lifecycle. External readiness probes may observe a closed listener instead of the brief draining 503 state. No production deployment wiring is added.

The gateway uses shared `LANGFUSE_LOG_LEVEL`: lowercase `trace`, `debug`, `info`, `warn`, `error`, `fatal`, defaulting to `info`. Rust maps `fatal` to `ERROR`; other invalid values fail startup without exposing the input. Regression tests first failed on the old parser/environment lookup and now pass.

Root-watch verification: the real gateway loads root `.env`, recompiles and restarts on a Rust edit while persistent peer fixture processes keep their original starts, and closes its listener on shutdown. No repeated restart was observed after the edit; ignored generated outputs did not trigger a loop. The real repository task graph confirms only gateway dev is interruptible among gateway/Web/worker/shared. Removed the redundant `dev:gateway` script; gateway-only work uses `pnpm dev --filter=@langfuse/ai-gateway`. JSON/README formatting and diff whitespace pass.

The gateway also shares `LANGFUSE_LOG_FORMAT`: compact `text` by default, or `json` for structured production output. Invalid values fail configuration validation without echoing the value. Actual subprocess output is covered for the default, explicit text and JSON modes. Regression checks reproduced ignored format settings and JSON-only output before the fix; final Rust validation reports 4 unit + 4 lifecycle + 3 startup tests passing, with fmt and Clippy clean.


ai-gateway/.env.dev.example
hassiebp, line 1: why do we need a separate .env.dev.example? Why not use the global one?

docker-compose.dev.yml
```diff
@@ -1,4 +1,31 @@
 services:
+  ai-gateway:
+    profiles: [gateway]
+    build:
+      context: ./ai-gateway
+      target: development
+    environment:
+      AI_GATEWAY_LISTEN_ADDRESS: 0.0.0.0:8080
+      AI_GATEWAY_LOG_LEVEL: ${AI_GATEWAY_LOG_LEVEL:-info}
```
hassiebp: if i set these at the root .env, will they be respected here?

docker-compose.dev.yml
```diff
@@ -1,4 +1,31 @@
 services:
+  ai-gateway:
+    profiles: [gateway]
+    build:
+      context: ./ai-gateway
+      target: development
+    environment:
+      LANGFUSE_AI_GATEWAY_LISTEN_ADDRESS: 0.0.0.0:8080
+      LANGFUSE_AI_GATEWAY_LOG_LEVEL: ${LANGFUSE_AI_GATEWAY_LOG_LEVEL:-info}
```
hassiebp: why are we introducing new env var? We should simply reuse LANGFUSE_LOG_LEVEL that is global

hassiebp: double check whether the values of Langfuse log level are matching the format expected in rust, else we might need a parsing helper there

PR #17202, opened 2026-09-08
fix(observations): make start-time lookup inputs a best-effort performance hint

niklassemmler:
🟡 **Live preview: [pr-17202 app preview](https://pr-17202.preview.langfuse.com)**

## What does this PR do?

Makes the two public start-time inputs on observation lookups — `objectStartTime` on `POST /comments` and `startTime` on `GET /observations/{id}` — **best-effort performance hints** rather than filters.

Both fed a start-time equality into the by-id ClickHouse lookup. As a filter this was a footgun: a wrong or stale value returned not-found for an object that actually exists. These are the only two public endpoints that thread a start time into a single-object/reference lookup (list-endpoint range filters are unaffected).

Behavior now:
- The shared repository predicate matches at **minute resolution** (`toStartOfMinute`), which is the finest the `events_full` primary key can prune on — this is what makes the lookup fast.
- Both public call sites wrap the lookup with an **unbounded retry on miss**, so a correct value speeds the query up via partition/part pruning, and an incorrect or omitted value falls back and never changes the result.

Docs on both endpoints and the shared zod comment now describe the hint semantics. Internal callers (eval by-id, web-callouts) are unchanged.

## Type of change

## Checklist

- [x] `pnpm run lint` / `pnpm run typecheck` — Tasks: 7 successful, 7 total (forced)
- [x] Fern sources updated + `pnpm run openapi:export` regenerated; `openapi:check` exit 0
- [x] Tests updated: wrong-value cases succeed via fallback; matching value exercises the fast path
- [ ] Server tests (comments + observations) not run locally (no ClickHouse container / `.env.test` in this environment) — they run in CI

### DIFF
--- a/fern/apis/server/definition/comments.yml
+++ b/fern/apis/server/definition/comments.yml
@@ -64,7 +64,7 @@ types:
         docs: The id of the user who created the comment. Must be a member of the organization that owns the project, otherwise an error will be thrown.
       objectStartTime:
         type: optional<datetime>
-        docs: The start time of the referenced object (for observations, the observation's start time). Acts as a filter that speeds up reference validation by narrowing the lookup. Matched at minute resolution (the value is floored to the minute), so it must fall in the same minute (UTC) as the object's start time; a value in a different minute causes the reference to be treated as not found. Omit it to validate the reference without a time bound.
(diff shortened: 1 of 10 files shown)


niklassemmler:
Just to be clear, the initial code was just released a short while ago. We should be fine adding this code if we merge quickly enough.

niklassemmler:
@claude review

PR #17171, opened 2026-09-08
fix(datasets): read v4 dataset metrics when preview is enabled

wochinge:
🟡 **Live preview: [pr-17171 app preview](https://pr-17171.preview.langfuse.com)**

## What does this PR do?

Cloud is dual-write. Dataset list run-count and last-run time were gated on `events_only`, so preview users stayed on Postgres. The gate now follows the session v4 preview flag, same as other v4 read paths.

Impacted packages: `web`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code.

## Verification

- `pnpm --filter web run test src/__tests__/server/datasets-metrics-events-only.servertest.ts src/__tests__/server/datasets-metrics.servertest.ts` — `Tests 6 passed (6)`
- Commit hook: `turbo run lint` — `Tasks: 7 successful, 7 total` / `Cached: 5 cached, 7 total`

The Postgres metrics suite uses a non-preview session so it still hits `dataset_runs`. No other test rewrite.

## Checklist

- [x] No documentation changes are required.

<div><a href="https://cursor.com/agents/bc-88f7fdbf-8a07-4b3c-9388-e7ecf3496d9d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-88f7fdbf-8a07-4b3c-9388-e7ecf3496d9d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

<details open><summary><h3>Greptile Summary</h3></summary>

This PR makes dataset list run-count and last-run metrics follow the authenticated session's v4 preview state rather than the global write mode.
- Preview-enabled sessions now read experiment metrics from events.
- Non-preview sessions retain the Postgres-backed metrics path.
- Existing server tests were adjusted to cover both session states, although the preview regression test does not explicitly pin dual-write mode.
[...]

PR #17208, opened 2026-09-08
fix(evals): read evaluator costs from the store matching the user's preview flag

### LINKED ISSUE #17207: Evaluator costs read from the events store for non-preview users on dual-write deployments
### Describe the bug

On deployments running `LANGFUSE_MIGRATION_V4_WRITE_MODE=dual` (e.g. Langfuse Cloud), the two evaluator cost queries in the v1 evals router choose their read store from the deployment write mode instead of the user's session preview flag:

- `evals.costByEvaluatorIds` (cost column in the evaluators table)
- `evals.avgCostByEvaluatorIds` (the estimated cost shown in the Run Evaluation dialog)

Both gate on `env.LANGFUSE_MIGRATION_V4_WRITE_MODE === "legacy"`, so on a dual-write deployment every user - including users without the v4 preview - gets costs from the v4 events store. Everything else those non-preview users see on the page (evaluator list, execution counts, executions) comes from the legacy store, so the cost columns are inconsistent with the rest of the page and miss all execution history that predates the events store (understated totals and averages, and a too-low cost estimate before running an evaluation).

### To reproduce

1. On a dual-write deployment, sign in as a user with the v4 preview disabled.
2. Have an LLM-as-judge evaluator whose executions wrote cost to the legacy observations store (`metadata['job_configuration_id']`), including executions from before the events store existed.
3. Open the evaluators page. The "cost" cell for that evaluator only reflects what the events store holds, while the execution count next to it reflects the full legacy history - the two disagree.
4. The same gate produces an understated average in the Run Evaluation dialog's estimated cost row.

### Expected behavior

The read path follows the user's session preview flag, like the rest of the v3/v4 UI: preview users read from the events store, everyone else reads from the legacy store. This is the same rule applied to `datasets.allDatasetsMetrics` in #17171 ("Cloud dual-write still serves the v4 experience via the session preview flag, so dataset run count and last-run time must follow that flag rather than events_only write mode") - these two eval queries are the remaining user-visible reads with the old gating.

Note the session flag is already always `false` on `legacy` deployments and always `true` on `events_only` ones (see session construction in `web/src/server/auth.ts`), so single-mode deployments behave exactly as today; only dual-write + preview-off changes.

### Additional information

I have a fix ready that mirrors #17171, plus a regression test - PR follows.

Harsh23Kashyap:
## What does this PR do?

Fixes #17207

`evals.costByEvaluatorIds` and `evals.avgCostByEvaluatorIds` chose their read store from the deployment write mode (`LANGFUSE_MIGRATION_V4_WRITE_MODE === "legacy"`). On dual-write deployments that sent every user's evaluator costs to the v4 events store - including non-preview users, whose evaluators page otherwise reads entirely from the legacy store. The result: cost columns that miss pre-events history and disagree with the execution counts on the same page, and an understated cost estimate in the Run Evaluation dialog.

Both queries now follow the session preview flag instead, mirroring the rule #17171 applied to `datasets.allDatasetsMetrics`: preview users read from the events store, everyone else reads the legacy store. Since `v4BetaEnabled` is always `false` on `legacy` deployments and always `true` on `events_only` ones (`web/src/server/auth.ts`), single-mode deployments behave exactly as before; the only behavior change is dual-write + preview-off, which is the broken case.

## Type of change

## How to test

Added `web/src/__tests__/server/evals-cost-read-path.servertest.ts` with three cases against seeded ClickHouse observations:

1. non-preview user reads the cost from the legacy observations store;
2. preview user does not see legacy-only costs (reads the events store);
3. non-preview user gets the legacy average and execution count.

Caveat: I could not run the server test suite in my environment (it needs Postgres + ClickHouse), so the tests are written against the repo's existing factories and test patterns but have not been executed here. Verified by tracing the gate through session construction (`auth.ts`), both repository implementations (`observations.ts` / `events.ts` cost helpers), and the UI call sites (`useEvaluatorTableData`, `EstimatedCostRow`); the change mirrors the already-merged #17171 one-for-one.

## Checklist

- [x] The change is scoped to the two mis-gated reads; no other write-mode gates are touched

<h2><a href="https://app.greptile.com/api/retrigger?id=61696789"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/personal-org-4986/-/pull-requests/langfuse/langfuse/17208"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until the legacy average-cost path correctly handles evaluator-to-rule identifiers used by historical observations.

<h2>Findings</h2>

1. P1&nbsp;**Legacy lookup uses rule IDs** <a href="https://github.com/langfuse/langfuse/pull/17208#discussion_r3960483007">▶</a>

`````markdown
### Issue 1
web/src/features/evals/server/router.ts:1416-1419
For non-preview users, this branch passes evaluator IDs to a legacy helper that filters observations by `job_configuration_id`. The Run Evaluation dialog supplies IDs from the v2 `Evaluator` model, while `job_configuration_id` contains the distinct evaluation-rule/job-configuration ID for rule-based history. Selecting an evaluator with history from an assigned rule therefore returns no matching data, causing the dialog to show “No data” or an understated partial estimate. The new test hides this mismatch by using the evaluator ID as `job_configuration_id`; it should cover distinct evaluator and rule IDs and make the legacy lookup resolve the correct identifiers.

---

`````

</details>

- Aligns total-cost reads with the evaluator experience selected for the session.
- Applies the same gate to average-cost and execution-count metrics.
- Adds ClickHouse-backed server tests for preview and non-preview paths.
- The average-cost legacy branch does not translate evaluator IDs to the rule/job-configuration IDs stored in historical observations.
</details>

```mermaid
flowchart TD
  A[Evaluator cost request] --> B{Session v4BetaEnabled?}
  B -->|true| C[Events store]
  B -->|false| D[Legacy observations store]
  C --> E[Read evaluator_id for average cost]
  D --> F[Read job_configuration_id]
  G[Run Evaluation dialog passes evaluator IDs] --> A
  F --> H{Evaluator ID equals rule ID?}
  H -->|usually no for rule-based history| I[Missing or understated estimate]
```
</details>


Harsh23Kashyap:
Good catch - fixed in 87a277a.

The dialog passes v2 evaluator ids, while rule-based executions anchor `metadata.job_configuration_id` to the distinct rule id, so the legacy read missed rule-based history. The legacy branch now resolves each evaluator's rules via `evaluationRuleEvaluatorAssignment`, matches generations on `metadata.evaluator_id` or the combined anchor set (evaluator ids + rule ids), groups current executions by evaluator id, and maps rule-anchored groups back to their evaluator. Groups anchored to a rule with several of the queried evaluators are ambiguous and skipped rather than double-counted. Evaluator test runs are excluded, matching the events-store variant.

The regression test now uses distinct evaluator and rule ids and covers all three anchor shapes (ruleless manual run, older rule-anchored rows, current rows with `evaluator_id` stamped) plus test-run exclusion. Same caveat as the PR description: the test is written but not runnable in my environment (no Postgres/ClickHouse), so please let CI confirm.

PR #17209, opened 2026-09-08
feat(experiments): show score filters in the search bar

nkabardin:
🟡 **Live preview: [pr-17209 app preview](https://pr-17209.preview.langfuse.com)**

## TL;DR

A score filter added in the experiments sidebar left the search bar empty, so the bar read as though no score filter were applied. The bar now renders score filters and offers every score name when you type `scores.` — the same as the tracing tables.

## Why

The sidebar and the bar edit **one** filter state, and the bar is meant to be a view of it. A filter it cannot express is a filter it silently omits, and the user sees an empty bar next to an active filter.

The `scores.` namespace was closed on this view deliberately, and the reason is in the registry's own comment: `scores.<name>` lowers onto the canonical `scores_avg` / `score_categories` / `score_booleans` columns, and experiments used to name its score columns `obs_*`. Unifying the facets moved this view onto exactly those columns, so the blocker is gone.

## What

One flag: `scores: true` on the experiments registry.

Everything the bar needs was already in place. The score-type context — which routes a typed name to numeric, categorical or boolean — is derived from the same observed options the sidebar uses, keyed by the canonical column names the unification introduced.

`traceScores.` stays closed. The canonical columns already match at trace level, so a second namespace would be another way to say the same thing, against `trace_*` columns the sidebar no longer offers.

## Result

| | |
| --- | --- |
| Sidebar sets a score filter | bar shows `scores.citation_usable_rate:4` |
| Type `scores.` | 18 names, each tagged `Observation`/`Trace` and typed (`numeric` / `categorical` / `boolean` score) |
| Type `traceScores.` | not offered |

The example in the placeholder is now `scores.groundedness:>0.8`, and the stale `dataset:` note about ids is gone — that view filters datasets by name since #16762.

<details>

`pnpm run lint` → `Tasks: 7 successful, 7 total`. `pnpm run typecheck` → `Tasks: 8 successful, 8 total`. `pnpm --filter web run test-client "features/search-bar"` → `Tests 269 passed (269)`.

The invariant test that pinned the closed namespace now pins the open one, plus a case for the reported bug: a `scores_avg` filter must derive to `scores.<name>:<value>` with nothing skipped. INV-4 still holds — the derived text parses for this registry — and the experiments sidebar cases now expect text rather than a skip.

Verified live on a seeded project: the reported URL renders the token, and typing `scores.` lists all 18 names across the three data types.

</details>

---

Built on `main` as a single commit. Supersedes #16788, which was stacked on a branch that is gone: #16786 was closed (in experiments, scores are meant to sit on the item's root observation), and deleting its branch closed the stacked PR with it. Same four files, same change, plus the reconcile with main's newer `useExperimentsTableData`: main grew an `isProjectReady` gate solving the same "a disabled query is neither loading nor errored" problem, so the two are folded into one `canQuery` predicate that also covers main's most-recent fallback query.

## Since review

Two reviewer findings fixed, both worth naming because one was a real hang:

- **The query gate could never open.** It treated an empty dataset map as proof the options were still loading, so a project with **no datasets** — or a failed dataset-options fetch — left the table in its loading skeleton forever whenever a dataset-name filter was active, which is worse than the flash of "No experiments" the gate exists to prevent. It now asks the query, not the data: `useExperimentFilterOptions` exposes the datasets query's settle state and a pure `awaitsDatasetNames` waits only while it is unsettled, so success-with-zero-datasets and an error both let the query through. Nothing reads the map's size any more. Three tests pin it, including both states that used to hang.
- **An unresolvable example.** `searchExamples` still advertised `dataset:…` on the dataset-scoped page, where this PR correctly drops the Dataset facet. The examples stay authored, not derived, and are simply filtered: an example whose field the registry cannot resolve is not offered. Resolution goes through the registry, so aliases and `scores.` paths are honoured. The unscoped placeholder is byte-identical.

<h2><a href="https://app.greptile.com/api/retrigger?id=61757532"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/personal-org-4986/-/pull-requests/langfuse/langfuse/17209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because projects with no resolvable dataset options can leave the experiments table permanently loading when a dataset-name filter is present.

<h2>Findings</h2>

1. P1&nbsp;**Query Gate Never Opens** <a href="https://github.com/langfuse/langfuse/pull/17209#discussion_r3961478920">▶</a>

`````markdown
### Issue 1
web/src/features/experiments/components/table/ExperimentsTable.tsx:440-444
When a dataset-name filter is present, this treats an empty dataset map as proof that the options are still loading. If the project has no datasets, or loading the dataset options fails, the map remains empty and all experiment queries stay disabled. The table therefore shows an endless loading state instead of no results or an error.

---

`````

</details>

- Enables `scores.<name>` while keeping `traceScores.<name>` unavailable.
- Adds scoped-registry and score serialization invariant coverage.
- Adds a data-query gate that currently cannot distinguish pending dataset options from a successfully empty or failed options query.
- Leaves an obsolete registry comment describing the superseded score behavior.
</details>


nkabardin:
@claude review

PR #17255, opened 2026-09-09
feat(auth): add user-triggered session revocation

marksalpeter:
🟡 **Live preview: [pr-17255 app preview](https://pr-17255.preview.langfuse.com)**

## What does this PR do?

Adds durable JWT session revocation using a nullable per-user `sessionsExpiredAt` timestamp and a stable `loginAt` JWT claim. Users can invalidate every active session from Account Settings, and password resets advance the same timestamp.

PostgreSQL provides the authoritative clock for both boundaries. Revocation uses an atomic `UPDATE ... RETURNING`, sessions require `sessionsExpiredAt < loginAt`, and a fresh login is ordered at least one millisecond after an existing boundary. This avoids application-host clock skew and same-millisecond revoked-token survival.

A nullable timestamp preserves all existing sessions at deploy. Legacy JWTs without `loginAt` remain valid until the user's first revocation, then compare as epoch-zero and are denied. Password reset/setup always re-authenticates the acting browser after revoking every old JWT.

The session lookup stays on `findUnique`: Prisma accepts non-unique filters alongside a unique field, so the eviction folds into the existing query and reuses its `dbUser === null` denial path.

Impacted packages: `@langfuse/shared`, `web`.

## Proof

The source-built local stack renders the danger-zone action, confirms intent, invalidates the current session, and redirects immediately to sign-in:

<a href="https://cursor.com/agents/bc-36216925-2e35-49f6-98cc-268c0be8150c/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fsession_revocation_review_fix.mp4">session_revocation_review_fix.mp4</a>

## Type of change

## Mandatory Tasks

- [x] Self-reviewed the code.

## Verification

- `pnpm --filter @langfuse/shared run db:generate` — Prisma Client generated successfully
- `prisma migrate status` — `Database schema is up to date!`
- Targeted server tests — `Test Files 4 passed (4)`; `Tests 18 passed (18)`
- `sessionExpiration.servertest.ts` after the rename — `Test Files 1 passed (1)`; `Tests 1 passed (1)`
- Targeted client tests — `Test Files 2 passed (2)`; `Tests 4 passed (4)` (earlier head; no UI change in this commit)
- `pnpm exec turbo run lint` — `Tasks: 7 successful, 7 total`; `Cached: 5 cached, 7 total`
- `pnpm --filter web run typecheck` — exit 0 (earlier head)
- The eviction is covered by an integration test that drives the real NextAuth session callback against Postgres. Removing the filter from `auth.ts` makes it fail (`expect(revoked.user).toBeNull()`), so it pins the behavior rather than restating it.

## How to test this

1. Open the [PR app preview](https://pr-17255.preview.langfuse.com/account/settings); the shared synthetic demo account signs in automatically.
2. Scroll to **Danger Zone** and click **Sign Out of All Sessions**.
3. Confirm the warning, then verify the browser lands on the opted-out sign-in form.
4. Sign in, use **Change Password**, complete the OTP flow, and verify the resulting redirect remains authenticated with the new password.

Data: the synthetic demo account is sufficient.

## Checklist

- [x] Added regression coverage for session-expiration ordering, initial claim stamping, refresh preservation, the eviction matrix, end-to-end eviction against Postgres, explicit revocation, password-reset revocation, authenticated password re-login, and local sign-out failure.
- [x] Added a single nullable, no-default migration so deployment does not invalidate existing sessions.
- [x] No new dependencies.
- [x] No public API contract or documentation change required.

## What to doubt in review

- Confirm the epoch-zero fallback is the desired rollout behavior for legacy tokens after a user's first revocation.
- Confirm converting an authenticated OAuth setup flow into a credentials session after setting a password is the desired re-authentication outcome.
- Review the raw SQL clock helpers for compatibility with supported PostgreSQL deployments and `TIMESTAMP(3)` precision.
- In `ResetPasswordPage`, a rejected `signIn` after a successful password change falls into the existing catch and shows the generic form error while the old session is already revoked. Left as-is to keep the diff to the one-line guard removal; say if you want it handled explicitly.
- The revocation stamp runs as a second statement inside the password-reset transaction rather than sharing the `updateMany`, so it re-reads the Postgres clock. Check that ordering inside the transaction is what you want.
- When the durable revocation succeeds but the local `signOutCleanly()` rejects, the user sees the success toast and the failure only reaches telemetry; they stay on the settings page until the next request bounces them. Sibling `DeleteAccountButton` behaves the same way. Say if you want a user-visible signal there instead.
- The comparison is `sessionsExpiredAt < loginAt` (same-millisecond tokens stay revoked). That is slightly stricter than a colloquial "expired at" reading that would treat equality as still valid. Say if you want equality to survive.
- The eviction rides on Prisma's extended `findUnique` filters. The integration test covers today's client version; flag it if you would rather see the constraint expressed as an explicit second check.

<div><a href="https://cursor.com/agents/bc-36216925-2e35-49f6-98cc-268c0be8150c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-36216925-2e35-49f6-98cc-268c0be8150c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

<h2><a href="https://app.greptile.com/api/retrigger?id=62207020"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/personal-org-4986/-/pull-requests/langfuse/langfuse/17255"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 3/5</h2>

This PR is not safe to merge until session revocation uses a strictly ordered boundary that cannot accept a revoked JWT because of timestamp equality or application-host clock skew.

<h2>Findings</h2>

1. P1&nbsp;Security&nbsp;**Revoked Sessions Can Survive** <a href="https://github.com/langfuse/langfuse/pull/17255#discussion_r3971175769">▶</a>
2. P2&nbsp;**Successful Revocation Reports Failure** <a href="https://github.com/langfuse/langfuse/pull/17255#discussion_r3971175780">▶</a>

`````markdown
### Issue 1
web/src/server/auth.ts:768
The revocation check accepts equality, while `loginAt` and `sessionsValidAfter` come from application clocks that may run on different web replicas. Clock skew can make a session issued before revocation appear newer and remain authenticated, or make a new session appear older and be rejected. Even without clock skew, a token created in the same millisecond as revocation is accepted, contradicting the promise to sign out every session. Use a single authoritative, strictly ordered revocation boundary.

> **How this was verified:** The JWT claim and database boundary both use millisecond timestamps, and the session query explicitly accepts `sessionsValidAfter <= loginAt`, including equality.

### Issue 2
web/src/pages/account/settings/index.tsx:253
If the server mutation succeeds but `signOutCleanly()` then rejects, this shared catch reports “Failed to Sign Out of All Sessions” even though the durable revocation already completed. The user can receive contradictory success and failure messages, remain on the settings page with a revoked cookie, and retry an operation that already succeeded. Handle local sign-out failure separately from the mutation result.

---

`````

</details>

- Adds the nullable `sessions_valid_after` user column without invalidating existing sessions during deployment.
- Filters session reconstruction against the revocation boundary and treats legacy tokens as older than any established boundary.
- Adds server mutations, account UI, NextAuth typing, and regression coverage.
- The timestamp comparison does not provide strict ordering across equality or distributed application clocks, so the all-session guarantee has an authentication-boundary gap.
- Local sign-out failures are currently presented as failures of the already-completed durable revocation.
</details>

```mermaid
sequenceDiagram
    participant B as Browser
    participant W1 as Web replica issuing JWT
    participant DB as PostgreSQL
    participant W2 as Web replica revoking sessions
    participant W3 as Web replica validating session
    B->>W1: Sign in
    W1-->>B: JWT with loginAt from W1 clock
    B->>W2: Sign out all sessions
    W2->>DB: Store sessionsValidAfter from W2 clock
    B->>W3: Subsequent request with JWT
    W3->>DB: "Read user where boundary <= loginAt"
    alt Boundary is strictly newer
        W3-->>B: Session rejected
    else Equality or clocks mis-order events
        W3-->>B: Revoked session remains authenticated
    end
```
</details>


marksalpeter:
@claude review

PR #17290, opened 2026-09-10
feat(worker): Add sampled delayed trace execution with start-time bounds

LysanderKie:
🟡 **Live preview: [pr-17290 app preview](https://pr-17290.preview.langfuse.com)**

## Summary

Add an opt-in experiment for debounced trace observation reads. Packages: `worker` and `@langfuse/shared`.

- Direct-write v4 OTel batches calculate minimum and maximum observation start times per trace. A single-member sorted set per project/trace preserves the minimum via ZADD LT; per-trace MULTI/EXEC atomically pairs it with EXPIRE 7200. Up to 100 transactions are pipelined per chunk; the consumer reads ZSCORE.
- The new `delayed-trace-execution` BullMQ queue debounces by project/trace for ten minutes. Delayed replacement carries `lastSeenStartTime`: the latest batch's maximum observation start time, not worker-arrival time. One `TRACE_EXECUTION_DELAY_MS` constant controls both delay and deduplication TTL. No scanner or ZSET scheduler.
- The consumer reuses `getObservationsForTraceFromEventsTable` with full I/O, metadata and tool data. No evaluation/LLM work or I/O stored in job results. Concurrency defaults to five per worker process; three attempts with exponential backoff.

## Configuration and timing

```sh
LANGFUSE_OTEL_DELAYED_TRACE_EXECUTION_ENABLED=true
LANGFUSE_OTEL_DELAYED_TRACE_EXECUTION_SAMPLE_PERCENT=10
LANGFUSE_DELAYED_TRACE_EXECUTION_PROCESSING_DELAY_MS=0
LANGFUSE_DELAYED_TRACE_EXECUTION_CONCURRENCY=5
```

Defaults: disabled and 10%. The feature flag explicitly gates the OTel scheduling call, the scheduling helper, and consumer registration. The same deterministic project/trace sample gates writes, enqueues and reads; increasing the percentage retains the previous sample. Configure consistently within a region. Set the percentage to zero to drain pending jobs without reading; disabling the flag stops the consumer. No changes to `.env.dev.example`.

Concurrency accepts a positive integer, defaults to `5`, and is per worker process, not a global queue limit. Restart/redeploy workers when changing it. With ten-second jobs, five slots provide approximately 30 jobs/minute/process before other bottlenecks; use `1` for a cautious rollout.

- Optional simulated downstream work: `LANGFUSE_DELAYED_TRACE_EXECUTION_PROCESSING_DELAY_MS` defaults to zero. Set `500` for 500ms, `1000` for one second, or `10000` for ten seconds. The asynchronous wait happens after a successful lookup, keeps the job active, and is included only in job timing. No real LLM calls are made.
- Job timing: `langfuse.queue.delayed_trace_execution.time_distribution`, `type:processing|wait` (milliseconds; processing includes simulated work, wait includes the deliberate debounce). Normal queue metrics report successful processing durations and failure counters.
- No custom ClickHouse or lookup timers. Query-outcome instrumentation is unchanged; timing comes only from the existing queue wrapper.
- Existing queue rates and ClickHouse query attribution are reused. Monitor the owning Redis shard; existing depth gauges omit delayed jobs.

For p50/p75/p95/p99 charts, enable percentile aggregations for the queue distribution in Datadog Metrics Summary after metrics arrive ([Datadog instructions](https://docs.datadoghq.com/metrics/distributions/)). Code emits through the existing DogStatsD setup; this PR does not change Datadog account settings or verify production delivery. Example:

```text
p75:langfuse.queue.delayed_trace_execution.time_distribution{type:processing}
```

The queue, configuration and payload were renamed before merge. Existing experiment jobs are not migrated automatically; jobs with the old `lastSeenAt` payload must be drained before deploying this version or cleared from the local experiment queue.

## Query bounds

The consumer passes Redis's retained minimum as the existing `timestamp` lower-bound hint and the job's `lastSeenStartTime` as an optional, inclusive `maxStartTime` upper bound. The existing one-hour lower lookback remains: `minimum - 1 hour <= start_time <= lastSeenStartTime`. If the minimum expired, only the lower bound is omitted.

`timestamp` cannot express an upper bound: it already means a lower-bound hint for other callers. Adding one optional `maxStartTime` filter through the existing events query builder preserves those callers' behavior. This keeps project and start-time filtering aligned with the ClickHouse `schema-pk-filter-on-orderby` guidance; no custom SQL or new instrumentation.

The upper bound is the maximum in the **latest replacement batch**, not a global maximum across batches. An out-of-order batch may move it backward and exclude later-starting observations from that execution. Dedupe identity remains project/trace only; arrival time still controls the ten-minute debounce and normal job timing.

## Limitations

Scheduling runs before event writes with a 100ms best-effort ingestion wait budget. Failures/timeouts skip remaining work without failing ingestion; an already submitted chunk of at most 100 traces can finish later. This is not cancellation or guaranteed sampling coverage.

Minimums and jobs share one Redis Cluster slot. Native debounce is not exactly-once or serial: backlog, retries and deduplication expiry can cause extra reads; clock/transit skew creates a narrow early-activation deduplication window.

Minimum keys use a new `minimum-zset:` namespace to avoid type collisions with earlier experiment string keys. Old `minimum:` keys expire naturally; no migration occurs. Existing jobs may temporarily omit the lower-bound hint until the new map is populated, and both namespaces occupy memory during the transition. ZADD LT requires Redis 6.2+.

The minimum is a retained-window hint, **not a durable trace start**. Expiry/recreation can still exclude earlier history, and the existing observation cap applies. This benchmark does not establish trace completeness or ClickHouse visibility.

## Redis implementation choice

Use one single-member sorted set per project/trace: `ZADD LT` retains the minimum, `EXPIRE 7200` refreshes its TTL on every arrival, and `ZSCORE` reads it. Each trace has its own atomic transaction; up to 100 transactions are pipelined without custom Lua.

**Why ZADD:** it is a reasonable choice for this sampled experiment, not a clear performance winner. The source-extracted comparison used identical 100-trace batches with 16/80 producers: **36 cases, 5.8M operations, correct minima/TTLs throughout**.

At 50k updates/sec with 80 paced producers (ranges across two local runs):

| Measurement | Lua | ZADD transactions |
|---|---:|---:|
| 100-trace write p99 | 3.7–4.5ms | 5.4–5.5ms |
| Unrelated-client read p99 | 3–4ms | 2ms |
| Redis CPU, one core | 18.8–20.9% | 18.4–20.1% |

Both sustained the offered load. ZADD traded slower batch writes for lower unrelated-client latency here; CPU differences were small/mixed, and map-read latency was similar.

**Memory tradeoff:** default Cluster key shapes measured 168 bytes/string versus 232 bytes/ZADD, including the different key names. That projects to ~267MB extra at 4.17M keys, or ~512MB at 8M; 10% sampling scales counts down tenfold. These are partial estimates, excluding expiry/global overhead, fragmentation, replicas and BullMQ jobs—not measured million-key capacity.

**Limits:** synchronized same-key bursts exceeded 100ms on both paths, without a consistent ZADD advantage for other-client latency. Transactions also execute without interleaving. Keep the **10% sample and 100ms wait budget**. These short local Redis tests exclude full BullMQ ingestion, large retained maps and production network/persistence/Cluster behavior.

## Verification

Tests added: **7** total. Six in `worker/src/__tests__/delayedTraceExecution.test.ts`: atomic min/TTL/tenant debounce and batch-max replacement; active-arrival follow-up; query fields/bounds/gates/expired lower bound/failure propagation; awaiting simulated work before completion; timeout with late completion; nested transaction-command failure prevents enqueueing without failing ingestion. Real Redis uses isolated temporary queues. One in `web/src/__tests__/server/repositories/event-repository.servertest.ts` checks inclusive query boundaries and unchanged results when the upper bound is omitted against local ClickHouse.

- Worker experiment and existing direct-write tests: `Tests 75 passed (75)`.
- Focused events repository tests: `Tests 2 passed | 95 skipped (97)` (unrelated cases excluded by the test-name filter).
- Existing web OTel processing tests: `Tests 4 passed (4)`.
- Typecheck: `Tasks: 8 successful, 8 total`; `Cached: 6 cached, 8 total`.
- Forced lint: `Tasks: 7 successful, 7 total`; `Cached: 0 cached, 7 total`.
- Knip, formatting, agent-config consistency and diff checks passed.

Not run: live HTTP-to-ClickHouse end-to-end, production load tests, or Redis Cluster/failover runtime tests. Scope is a local-tested, opt-in experiment; no production enablement.

<h2><a href="https://app.greptile.com/api/retrigger?id=62606367"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The opt-in experiment appears safe to merge, with no actionable correctness, security, or repository-rule violations identified.

- Maintains an expiring per-project/trace minimum observation timestamp atomically in Redis.
- Schedules delayed BullMQ jobs with replacement-based debounce behavior.
- Adds a gated worker that performs full observation reads without persisting observation I/O in job results.
- Adds configuration, generic queue integration, timing instrumentation, and Redis-backed lifecycle tests.
</details>

```mermaid
sequenceDiagram
  participant OTel as OTel ingestion worker
  participant Redis as Redis minimum cache
  participant Queue as Trace observation read queue
  participant Reader as Read worker
  participant CH as ClickHouse events table

  OTel->>OTel: Deterministically sample project/trace
  OTel->>Redis: Atomically retain minimum start time
  OTel->>Queue: Add or replace 10-minute delayed job
  OTel->>CH: Write direct-ingestion events
  Queue-->>Reader: Deliver debounced job
  Reader->>Redis: Read retained minimum timestamp
  Reader->>CH: Query trace observations with lookback
  CH-->>Reader: Return capped observations
  Reader->>Reader: Record count and timing only
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["feat(worker): add sampled debounced trac..."](https://github.com/langfuse/langfuse/commit/08d2880b5822c582d51dd53b70a96ef13f1635c4)</sub>


packages/shared/src/server/queues.ts
```diff
@@ -24,6 +24,12 @@ import { ProjectNotificationWebhookQueueEventSchema } from "./notifications/type
 
 export type { MonitorQueueEvent, MonitorQueueEventInput };
 
+export const TraceObservationReadEventSchema = z.object({
+  projectId: z.string(),
+  traceId: z.string(),
+  lastSeenAt: z.number(),
```
LysanderKie: Lets make this explicit: lastSeenStartTime

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,125 @@
...
+          selected.map(({ traceId, id }) => ({
+            name: QueueJobs.DelayedTraceExecution,
+            data: {
+              name: QueueJobs.DelayedTraceExecution,
+              id: randomUUID(),
+              timestamp: new Date(now),
+              payload: { projectId, traceId, lastSeenAt: now },
+            },
+            opts: {
+              delay: 600_000,
```
LysanderKie: Lets move that to a constant, as later this will come from the configuration on project-level

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,138 @@
...
+import { env } from "../../env";
+
+const TRACE_EXECUTION_DELAY_MS = 600_000;
+
+// Each chunk's keys share the BullMQ queue's Redis Cluster hash slot.
+const updateMinimums = `
+for i, key in ipairs(KEYS) do
+  local candidate = tonumber(ARGV[i])
+  local current = tonumber(redis.call('GET', key))
+  redis.call('SET', key, math.min(current or candidate, candidate), 'EX', 7200)
```
LysanderKie: *Shortened by Codex at the author's request.*

The benchmark summary and ZADD rationale now live in **Redis implementation choice** in the [PR description](https://github.com/langfuse/langfuse/pull/17290). This inline comment was attached to the superseded Lua implementation.

packages/shared/src/server/redis/delayedTraceExecution.ts
```diff
@@ -0,0 +1,34 @@
+import { Queue } from "bullmq";
+import { QueueName, TQueueJobTypes } from "../queues";
+import { createBullMQQueueOptionsWithRedis } from "./redis";
+import { logger } from "../logger";
+
+export class DelayedTraceExecutionQueue {
```
maxdeichmann: I would drop delayed from the queue name

packages/shared/src/server/redis/traceExecution.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!options) return null;
+    this.instance = new Queue<TQueueJobTypes[QueueName.DelayedTraceExecution]>(
+      QueueName.DelayedTraceExecution,
+      {
+        ...options,
+        defaultJobOptions: {
+          removeOnComplete: true,
+          removeOnFail: 1000,
+          attempts: 3,
+          backoff: { type: "exponential", delay: 5000 },
```
maxdeichmann, line 23: Didnt we want to have a  10m delay by default?

packages/shared/src/server/queues.ts
```diff
@@ -24,6 +24,12 @@ import { ProjectNotificationWebhookQueueEventSchema } from "./notifications/type
 
 export type { MonitorQueueEvent, MonitorQueueEventInput };
 
+export const DelayedTraceExecutionEventSchema = z.object({
+  projectId: z.string(),
+  traceId: z.string(),
+  lastSeenStartTime: z.number(),
```
maxdeichmann, line 30: IMO we need a test to make sure that this really works as expected and delay is reset when an event with same id is added.

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,142 @@
...
+} from "@langfuse/shared/src/server";
+import { env } from "../../env";
+
+const TRACE_EXECUTION_DELAY_MS = 600_000;
+
+export function delayedTraceExecutionId(projectId: string, traceId: string) {
+  return createHash("sha256")
+    .update(JSON.stringify([projectId, traceId]))
+    .digest("hex");
+}
```
maxdeichmann: Why so complex? Cant we just concatenate these in a string?

packages/shared/src/server/redis/traceExecution.ts
LysanderKie: This is the default failure retry backoff of the queue, which I think we have my default 5 seconds (this is not the the "delay" for the actual jobs, this is defined on scheduling)

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export function isDelayedTraceExecutionEnabled(id: string): boolean {
+  return (
+    env.LANGFUSE_OTEL_DELAYED_TRACE_EXECUTION_ENABLED === "true" &&
+    parseInt(id.slice(0, 8), 16) / 2 ** 32 <
+      env.LANGFUSE_OTEL_DELAYED_TRACE_EXECUTION_SAMPLE_PERCENT / 100
```
maxdeichmann: I did not double check the second and condition :D

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,142 @@
...
+    const budget = new Promise<never>((_, reject) => {
+      timer = setTimeout(
+        () => reject(new Error("Trace read scheduling exceeded 100ms")),
+        100,
+      );
+    });
+    const startTimes = new Map<string, { first: number; last: number }>();
+    for (const event of events) {
+      if (performance.now() >= deadline)
+        throw new Error("Trace read sampling exceeded 100ms");
```
maxdeichmann: What is this doing here?

worker/src/features/traces/delayedTraceExecution.ts
```diff
@@ -0,0 +1,142 @@
...
+          break;
+        }
+        const [traceId, times] = entry.value;
+        const id = delayedTraceExecutionId(projectId, traceId);
+        if (isDelayedTraceExecutionEnabled(id))
+          selected.push({ traceId, id, ...times });
+      }
+      if (!selected.length) continue;
+      // Atomic per trace; refresh expiry even when the minimum is unchanged.
+      const pipeline = client.pipeline(
```
maxdeichmann: We run a multi node redis cluster. this may break.

maxdeichmann: Please point your coding agent at it. I think it konws how to take care of pipelines in redis clustered

worker/src/queues/delayedTraceExecutionQueue.ts
```diff
@@ -0,0 +1,49 @@
...
+export const delayedTraceExecutionProcessor: Processor<
+  TQueueJobTypes[QueueName.DelayedTraceExecution]
+> = async (job) => {
+  const { projectId, traceId, lastSeenStartTime } =
+    DelayedTraceExecutionEventSchema.parse(job.data.payload);
+  const id = delayedTraceExecutionId(projectId, traceId);
+  if (!isDelayedTraceExecutionEnabled(id)) return;
+  const queue = DelayedTraceExecutionQueue.getInstance();
+  if (!queue) throw new Error("Trace observation read Redis unavailable");
+  const client = await queue.client;
```
maxdeichmann: Do not take the queue redis client. Just import redis please.

worker/src/queues/delayedTraceExecutionQueue.ts
```diff
@@ -0,0 +1,49 @@
...
+  const result = await getObservationsForTraceFromEventsTable({
+    projectId,
+    traceId,
+    timestamp: minimum === null ? undefined : new Date(Number(minimum)),
+    maxStartTime: new Date(lastSeenStartTime),
+    selectIOAndMetadata: true,
+    selectToolData: true,
+  });
+  // Simulate downstream transcript/LLM work while occupying the active job.
+  if (env.LANGFUSE_DELAYED_TRACE_EXECUTION_PROCESSING_DELAY_MS > 0) {
```
maxdeichmann: Please remove this one for now as this is hard to trace and gives us little insights. This produces gaps in datadog traces which would usually hint at stalling event loops.

worker/src/features/traces/delayedTraceExecution.ts
LysanderKie: This adds a safety budget if the redis and job scheduling operations would for whatever reason would take longer than 100ms to not delay the ingestion queue

maxdeichmann: I understand that this budget is nice but it makes everything complex here. We can roll back this change by setting an env variable. I would remove this.

packages/shared/src/server/redis/traceExecution.ts
maxdeichmann: Ah yes you are right.

packages/shared/src/server/redis/delayedTraceExecution.ts
maxdeichmann: Nit, this can stay as is for POC

packages/shared/src/server/queues.ts
maxdeichmann: Also not required for POC

packages/shared/src/server/redis/delayedTraceExecution.ts
LysanderKie: Already done

packages/shared/src/server/queues.ts
LysanderKie: Done

worker/src/features/traces/delayedTraceExecution.ts
LysanderKie: Removed it

worker/src/queues/delayedTraceExecutionQueue.ts
LysanderKie: Done

worker/src/queues/delayedTraceExecutionQueue.ts
LysanderKie: Removed

worker/src/features/traces/delayedTraceExecution.ts
LysanderKie: Checked & changed the implementation: we now use shared Redis and separate single-key MULTI/EXEC transactions.

PR #17298, opened 2026-09-10
fix(blob-export): raise S3 part size and fail oversized windows loudly

### LINKED ISSUE #17282: Scheduled blob export silently truncates a window once the Parquet exceeds 48.8 GiB (lib-storage default partSize)
### Summary

The scheduled Blob Storage Export cannot upload an object larger than **48.83
GiB**, and when it hits that limit it does not fail permanently — it retries,
and each retry re-queries ClickHouse, which by then holds less data because of
the retention TTL. Eventually a retry produces an object small enough to upload,
and that truncated object is committed and reported as a **successful** export.

The manifest for that window still declares the full window. There is no error
state visible in the UI afterwards, no gap in the export cursor, and
`rowCount: null` in the manifest means the manifest cannot be used to detect it
either. On our deployment this silently lost most of two days of traces before
anyone noticed.

### Cause

`Upload` from `@aws-sdk/lib-storage` is used without a `partSize`, so it takes
the SDK default of 5 MiB and does not scale it with the payload. S3 caps a
multipart upload at 10,000 parts, giving a hard maximum object size of

```
10,000 × 5 MiB = 52,428,800,000 bytes = 48.83 GiB
```

Worker log, `@aws-sdk/lib-storage` 3.1074.0:

```
Error: Exceeded 10000 parts in multipart upload to Bucket: <bucket>
Key: langfuse-export/<project>/observations_v2/2026-09-06T01-20-00.parquet
    at Upload.__doConcurrentUpload (.../lib-storage/dist-cjs/index.js:330:23)
    at async Upload.__doMultipartUpload (.../lib-storage/dist-cjs/index.js:416:9)
    at async Upload.done (.../lib-storage/dist-cjs/index.js:235:16)
    at async S3StorageService.uploadFile (...)
```

Accompanied by, from the same run:

```
[BLOB INTEGRATION] Getting data from DB for blob storage integration failed: The operation was aborted
[Connection] Exec: socket was closed or ended before the response was fully read
[BLOB INTEGRATION] Error exporting observations_v2 for project <p> (jobId=<p>-2026-09-06T01:20:00.081Z attemptsMade=2 ...)
```

The upload throwing aborts the ClickHouse read feeding it, which is why the
retry has to re-run the query rather than resume.

### Why it degrades instead of failing

1. Window W is due. Its Parquet needs > 10,000 parts. Upload throws.
2. The ClickHouse read is aborted; the job fails; BullMQ retries.
3. The retry re-runs the query for W. The retention TTL has meanwhile deleted
   the oldest rows in W, so the result is smaller.
4. Repeat until the result fits in 10,000 parts, then it commits.

Observed on our deployment: window `2026-09-06T01:20Z` was attempted six times
over three days. One abandoned upload held 42.89 GiB of it. The object that
finally committed held 10.3 GiB and covered 2h07m of the 24-hour window.

The abandoned uploads are also not cleaned up — nine were open simultaneously
holding 175.9 GiB of orphaned parts, which are billed and invisible to
`aws s3 ls`.

### Reproduction

Any project whose export window produces a Parquet over 48.83 GiB. Ours reached
that when daily traffic roughly doubled; the largest object we ever committed
successfully was 51,286,432,036 bytes = 9,783 parts, i.e. 217 parts short of the
limit, so we had been running just under the wall for weeks without knowing.

### Suggested fixes, roughly in order of value

1. **Set `partSize` from the expected payload size**, or simply raise it — 64
   MiB gives a 625 GiB ceiling at the same part count and costs nothing for
   small exports. `lib-storage` accepts `partSize` in the `Upload` constructor.
2. **Fail the window permanently instead of retrying**, when the error is
   `Exceeded 10000 parts`. It is not transient, and retrying is what converts a
   loud failure into silent data loss.
3. **Populate `rowCount` in the manifest.** It is currently `null` for every
   file, which removes the one cheap way for an operator to verify an export.
4. **Abort the multipart upload on failure** so orphaned parts are not left
   behind and billed.
5. **Surface a persistent error state** for a window that never exported
   completely. The UI shows only the most recent run's status.

### Related

- The same-second key collision in #15617 has the same character: an export that
  reports success while losing a window.
- Feature request adjacent to (3): expose the exported row count per window in
  the API so completeness can be checked without reading every Parquet footer.

### Environment

- Langfuse v4.24.0, self-hosted on EKS
- ClickHouse 26.4, `LANGFUSE_MIGRATION_V4_WRITE_MODE=events_only`
- Export: S3, Parquet, was `daily`, now `hourly`
- `@aws-sdk/lib-storage` 3.1074.0

MauryaQbit:
Fixes #17282.

## Problem
Scheduled blob exports used \@aws-sdk/lib-storage\ without a \partSize\, inheriting the 5 MiB SDK default. S3 caps multipart uploads at 10,000 parts, so any window producing a Parquet over 10,000 x 5 MiB = 48.83 GiB failed mid-upload. BullMQ retried, each retry re-queried ClickHouse after retention-TTL deletions had shrunk the window, and a truncated object eventually committed and was reported as a successful export with no visible error state.

## Changes
- \packages/shared\: new \s3MultipartLimits.ts\ — shared constants (\S3_MAX_PARTS\, 64 MiB default part size for a ~625 GiB ceiling), \S3MultipartLimitExceededError\, and \isS3MultipartLimitExceededError()\ (detects our error through \handleStorageError\ wrapping plus lib-storage's bare message).
- \S3StorageService.uploadFile\: default lib-storage \partSize\ to 64 MiB instead of 5 MiB; convert limit failures to the named non-retryable error (cause chain preserved).
- \S3StorageService.uploadFileBuffered\ fallback (buffered disabled): forward the resolved \partSizeBytes\/\maxConcurrentParts\ instead of dropping the 100 MiB tuning back to the 5 MiB default; rethrow limit errors unwrapped.
- \BufferedStreamUploader\: fail fast past 10,000 parts; the existing \inally\ aborts the multipart so no orphaned parts are left behind to bill.
- Worker blob-export handler: convert limit errors to BullMQ \UnrecoverableError\ (no retry into TTL-truncated success), treat as final attempt for the failure notification, and persist the actionable message in \lastError\.
- Abort classification: limit signatures count as concrete \upload-error\.
- Tests: ceiling math, detector (named/wrapped/lib-storage/negatives), uploader 10,001-part failure + abort and 10,000-part boundary.

## Verification
- \	sc --strict\ clean on the new module; detector runtime checks pass (7/7) under node type-stripping.
- Full \itest\ suite not run locally (workspace install is heavy); please let CI run it.

## Out of scope
- Parquet \
owCount\ in the manifest (binary stream; would need an extra COUNT query or footer parse). Non-Parquet paths already record it; \sizeBytes\ + persisted \lastError\ now give detection signals.

<h2><a href="https://app.greptile.com/api/retrigger?id=62644105"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR should not merge until multipart-limit failures remain retryable when their durable error state cannot be persisted.

- Adds shared multipart limits, formatting, and error-detection utilities.
- Propagates multipart tuning through both buffered and lib-storage upload paths.
- Persists actionable oversized-export errors and classifies them as concrete upload failures.
- Adds unit coverage for error detection and the 10,000-part boundary.
- One failure-state edge case remains: a transient database error can prevent `lastError` persistence while the job is still marked unrecoverable.
</details>

```mermaid
flowchart LR
    A[ClickHouse export stream] --> B{Upload path}
    B -->|lib-storage| C[64 MiB default parts]
    B -->|buffered| D[Configured parts]
    C --> E{More than 10,000 parts?}
    D --> E
    E -->|No| F[Complete multipart upload]
    E -->|Yes| G[Abort multipart upload]
    G --> H[Persist actionable lastError]
    H --> I[BullMQ UnrecoverableError]
    H -. persistence failure currently still follows .-> I
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(blob-export): raise S3 part size and..."](https://github.com/langfuse/langfuse/commit/7bca10473981e409fefcb2ab6eedae999ae4f707)</sub>

### DIFF
--- a/packages/shared/src/server/index.ts
+++ b/packages/shared/src/server/index.ts
@@ -11,6 +11,7 @@ export * from "./ingestion/eventBucketPath";
 export * from "./cache";
 export * from "./services/BufferedStreamUploader";
 export * from "./services/S3ChunkedUploadStrategy";
+export * from "./services/s3MultipartLimits";
 export * from "./services/email/transport";
 export * from "./services/email/organizationInvitation/sendMembershipInvitationEmail";
 export * from "./services/email/batchExportSuccess/sendBatchExportSuccessEmail";
--- a/packages/shared/src/server/services/BufferedStreamUploader.ts
+++ b/packages/shared/src/server/services/BufferedStreamUploader.ts
@@ -1,6 +1,12 @@
 import type { Readable } from "stream";
 import { backOff } from "exponential-backoff";
 import { logger } from "../logger";
+import {
+  S3_MAX_PARTS,
+  S3MultipartLimitExceededError,
+  formatGiB,
+  maxBytesForPartSize,
+} from "./s3MultipartLimits";
 
 const TRANSIENT_ERROR_PATTERNS = [
   "socket hang up",
@@ -233,6 +239,25 @@ export class BufferedStreamUploader {
 
   private async enqueuePart(partData: Buffer): Promise<void> {
     this.partNumber++;
+
+    // S3 hard limit: fail fast with a non-retryable error instead of uploading
+    // 10k parts and then failing at CompleteMultipartUpload (or worse: letting
+    // a BullMQ retry re-query a TTL-shrunk window and commit a truncated object
+    // as success — issue #17282). The finally in upload() aborts the multipart
+    // so no orphaned parts are left behind to bill.
+    if (this.partNumber > S3_MAX_PARTS) {
+      const partMiB = (this.params.partSizeBytes / 1024 / 1024).toFixed(1);
+      throw new S3MultipartLimitExceededError(
+        `Blob export for key ${this.params.key} exceeds S3's ${S3_MAX_PARTS}-part limit ` +
+          `at ${partMiB} MiB/part (max ~${formatGiB(maxBytesForPartSize(this.params.partSizeBytes))}). ` +
+          `Increase partSizeBytes and/or shorten the export window (e.g. daily -> hourly).`,
+        {
+          key: this.params.key,
+          partSizeBytes: this.params.partSizeBytes,
+          partsAttempted: this.partNumber,
+        },
+      );
+    }
 
     // Wait for a slot if all concurrent slots are full
     while (
--- a/packages/shared/src/server/services/StorageService.ts
+++ b/packages/shared/src/server/services/StorageService.ts
@@ -31,6 +31,14 @@ import {
 } from "./BufferedStreamUploader";
 import { S3ChunkedUploadStrategy } from "./S3ChunkedUploadStrategy";
 import {
+  S3_DEFAULT_PART_SIZE_BYTES,
+  S3_DEFAULT_QUEUE_SIZE,
+  S3MultipartLimitExceededError,
+  formatGiB,
+  isS3MultipartLimitExceededError,
+  maxBytesForPartSize,
+} from "./s3MultipartLimits";
+import {
   buildS3RequestDiagnostics,
   isS3DiagnosableError,
   type S3DiagnosticsContext,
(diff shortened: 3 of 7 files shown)


MauryaQbit:
@yarikoptic can you review this PR

PR #17303, opened 2026-09-10
fix(blob-storage): raise S3 multipart part size to lift 48.8 GiB object cap

niklassemmler:
🟡 **Live preview: [pr-17303 app preview](https://pr-17303.preview.langfuse.com)**

## What does this PR do?

The scheduled blob-storage export silently truncated large export windows. The non-buffered S3 upload fallback left `partSize` undefined, so `@aws-sdk/lib-storage` used its 5 MiB default and capped a single object at 5 MiB × 10,000 parts = **48.83 GiB**. Any window whose Parquet exceeded that failed the upload with `Exceeded 10000 parts`, which is not classified as a customer fault, so the job was retried. Each retry re-queried ClickHouse against a window the retention TTL had meanwhile shrunk, until a truncated object finally fit and committed **as a success** — losing data with no error state in the UI.

The same fallback also left `queueSize` undefined (lib-storage default 4) and discarded the caller's own `partSizeBytes` and `maxConcurrentParts`. Callers could therefore neither raise the part size to lift the 48.83 GiB cap nor bound their own upload memory, even though each caller already sizes both.

This makes the non-buffered fallback behave like the buffered path:

- Forward the caller's `partSizeBytes` as the multipart part size, lifting the cap to each caller's configured size (100 MiB → ~977 GiB for blob-storage and core-data exports).
- Forward the caller's `maxConcurrentParts` as `queueSize`, defaulting to `LANGFUSE_S3_UPLOAD_MAX_CONCURRENT_PARTS`, so peak buffered memory is `partSize × concurrency` — bounded and operator-tunable rather than lib-storage's fixed 4.

The buffered upload path (`LANGFUSE_S3_UPLOAD_ENABLE_BUFFERED=true`, which Langfuse Cloud already runs) already forwarded these values and is unchanged — this fix targets the default path that self-hosted deployments run.

### Reviewer note

The non-buffered fallback now forwards each caller's part size and concurrency instead of a single fixed default, so worst-case buffered memory is `partSize × maxConcurrentParts` per concurrent upload. Callers that fan out multiply that: blob-storage runs up to 4 tables per job, core-data at concurrency 3. `LANGFUSE_S3_UPLOAD_MAX_CONCURRENT_PARTS` (default 3) is the lever to bound it on tight workers.

## Type of change

## Checklist

- [x] `pnpm lint` passes (`@langfuse/shared`, forced)
- [x] `pnpm typecheck` passes (`@langfuse/shared` + `worker`, forced)
- [x] Regression tests (`StorageService.test.ts`) assert the fallback forwards the caller's `partSize` and honors its `queueSize`

---
🤖 Written by Claude (an AI agent) on behalf of Niklas

<h2><a href="https://app.greptile.com/api/retrigger?id=62672187"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because invalid multipart sizes can pass environment validation and make non-buffered uploads fail.

- Adds `LANGFUSE_S3_UPLOAD_PART_SIZE_BYTES` with a 64 MiB default.
- Passes the configured size to the non-buffered `@aws-sdk/lib-storage` upload path.
- Adds a regression test confirming that the configured size is forwarded.
</details>

```mermaid
flowchart LR
  A[uploadFileBuffered] --> B{Buffered uploads enabled?}
  B -->|Yes| C[BufferedStreamUploader]
  B -->|No| D[uploadFile]
  D --> E[lib-storage Upload]
  F[LANGFUSE_S3_UPLOAD_PART_SIZE_BYTES] --> D
  E --> G[S3 multipart object]
```
</details>

<sub>Reviews (2) · Last reviewed commit: ["fix(blob-storage): raise S3 multipart pa..."](https://github.com/langfuse/langfuse/commit/eddab2f8beb3529d4ec70c0180c0e9c66607190b)</sub>

### DIFF
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -270,10 +270,13 @@ const EnvSchema = z.object({
     .default(3),
   // Non-buffered multipart part size. 64 MiB x 10,000 parts caps a single
   // object at ~625 GiB; lib-storage's 5 MiB default caps at ~48.83 GiB and
-  // silently truncates larger exports.
+  // silently truncates larger exports. Bounded by S3's multipart part-size
+  // limits (5 MiB min, 5 GiB max) so an out-of-range value fails at boot
+  // instead of throwing EntityTooSmall on every fallback upload.
   LANGFUSE_S3_UPLOAD_PART_SIZE_BYTES: z.coerce
     .number()
-    .positive()
+    .min(5 * 1024 * 1024)
+    .max(5 * 1024 * 1024 * 1024)
     .default(64 * 1024 * 1024),
   LANGFUSE_S3_EVENT_UPLOAD_BUCKET: z.string(), // Langfuse requires a bucket name for S3 Event Uploads.
   LANGFUSE_S3_EVENT_UPLOAD_PREFIX: z.string().default(""),
--- a/packages/shared/src/env.ts
+++ b/packages/shared/src/env.ts
@@ -268,6 +268,13 @@ const EnvSchema = z.object({
     .min(1)
     .max(10)
     .default(3),
+  // Non-buffered multipart part size. 64 MiB x 10,000 parts caps a single
+  // object at ~625 GiB; lib-storage's 5 MiB default caps at ~48.83 GiB and
+  // silently truncates larger exports.
+  LANGFUSE_S3_UPLOAD_PART_SIZE_BYTES: z.coerce
+    .number()
+    .positive()
+    .default(64 * 1024 * 1024),
   LANGFUSE_S3_EVENT_UPLOAD_BUCKET: z.string(), // Langfuse requires a bucket name for S3 Event Uploads.
   LANGFUSE_S3_EVENT_UPLOAD_PREFIX: z.string().default(""),
   LANGFUSE_S3_EVENT_UPLOAD_REGION: z.string().optional(),
--- a/packages/shared/src/server/services/StorageService.test.ts
+++ b/packages/shared/src/server/services/StorageService.test.ts
@@ -10,6 +10,23 @@ import { env } from "../../env";
 import { resolveMediaStorageEndpoints } from "../s3";
 import { StorageServiceFactory } from "./StorageService";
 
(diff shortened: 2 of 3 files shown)


niklassemmler:
@claude review

niklassemmler:
@sumerman Can you review this? You had previously set the part limit to 5MB. Do you see any big risks in adjusting this upward to 64MB? Given the default queue size of 4 we would see at most 256MB total in memory.

This will mostly affect self hosters as we use `LANGFUSE_S3_UPLOAD_ENABLE_BUFFERED=true` in the Cloud.

packages/shared/src/env.ts
```diff
@@ -268,6 +268,16 @@ const EnvSchema = z.object({
     .min(1)
     .max(10)
     .default(3),
+  // Non-buffered multipart part size. 64 MiB x 10,000 parts caps a single
+  // object at ~625 GiB; lib-storage's 5 MiB default caps at ~48.83 GiB and
+  // silently truncates larger exports. Bounded by S3's multipart part-size
+  // limits (5 MiB min, 5 GiB max) so an out-of-range value fails at boot
```
sumerman: what does "fails at boot" means? do propagate a failure back to customer then?

sumerman:
I can't think of a reason not to do this in principle. I have one question, plus I think some of the bot review nits could be valid.

packages/shared/src/env.ts
niklassemmler: > what does "fails at boot" means?

The application won't start if you attempt to start it with invalid parameters.

It's a direct response to Greptile's finding https://github.com/langfuse/langfuse/pull/17303#discussion_r3981309876

niklassemmler:
@claude review

PR #17279, opened 2026-09-10
perf(filter-options): combine per-column score-facet scans for the events view

niklassemmler:
🟡 **Live preview: [pr-17279 app preview](https://pr-17279.preview.langfuse.com)**

## What does this PR do?

Combines the seven per-column ClickHouse scans behind the `events.filterOptions` score facets (the Observations filter dropdown) into a single `scores` scan.

- `events.filterOptions` runs one `scores` scan with conditional aggregation; TypeScript splits the seven score facets afterward.
- The combined query ranks each facet with its own `dense_rank()` window so SQL capping matches each facet's exact top-N (200 names / 20 categorical values).
- Adds a ClickHouse query-shape test for the combined builder and a repository comparison against the existing per-column helpers.

Returned filter-option shapes are unchanged (same 200-name / 20-value caps, same trace-attached scope, same config merge).

The scores-view exact event-facet scan that previously shared this branch is split into its own PR (#17366).

Impacted packages: `@langfuse/shared`, `web`.

## Type of change

- [x] Refactor (restructures existing code without changing behavior)

## Mandatory Tasks

- [x] Self-reviewed the query shape against the existing per-column helpers.

## Checklist

- Follows repo style (`pnpm run format` via pre-commit)
- No user-facing docs change (internal query shape only)
- Adds a SQL-shape unit test (one `FROM scores`) and a repository comparison against the per-column helpers

## Testing

- [x] `pnpm --filter web run test event-query-builder.servertest.ts`
- [x] `pnpm --filter web run test score-repository.servertest.ts -t getScoresFilterOptionsForEventFacets`
- Preview: https://pr-17279.preview.langfuse.com

**How to test:**

1. Open a project with scores on events.
2. Open the Observations filters.
3. Confirm score-name / source / data-type / categorical-value options still populate.

Linear Issue: [LFE-16129](https://linear.app/clickhouse/issue/LFE-16129/filter-options-combine-per-column-scorefacet-fan-out)

[...]

PR #17314, opened 2026-09-11
perf(filter-options): combine facet scans with a group-count bound

niklassemmler:
🟡 **Live preview: [pr-17314 app preview](https://pr-17314.preview.langfuse.com)**

## What does this PR do?

Same combine as #17279, with a simpler scores query: no per-facet `dense_rank()` windows.

Filter-option endpoints were issuing several independent `scores` / `events_core` scans in one `Promise.all`, all landing on a single ClickHouse replica. This collapses each group into one pass:

- `scores.filterOptionsFromEvents`: the three `events_core` facets (trace tags / name / user id) ride one exact `UNION ALL` of per-column `GROUP BY` / `LIMIT` queries.
- `events.filterOptions`: the score-name toplists become one `scores` scan with conditional aggregation (`countIf` / `groupUniqArrayIf`), then TypeScript splits the facets and merges categorical score-config values.

The combined scores query groups by `(name, source, data_type)`, orders by that group's `count()`, and keeps the top 2000 groups (`FILTER_OPTION_SCORE_NAME_LIMIT * 10`). TypeScript still sums across sources and caps each facet at 200 names.

#17279 keeps seven `dense_rank()` windows so SQL capping matches each facet's exact top-200. This PR drops those windows. Scan cost is unchanged; only post-aggregation CPU and SQL complexity change. #17307 is the same ranking-drop as a stacked delta on #17279 — left in place.

Returned filter-option *shapes* are unchanged (same name caps, same categorical value cap, same trace-attached scope, same config merge). Event-facet *counts* on the events-table path still follow `approx_top_k`. Parallel `Promise.all` remains elsewhere.

Impacted packages: `@langfuse/shared`, `web`.

## How a name can miss the toplist

ClickHouse ranks **groups**, TypeScript ranks **names**. A name is the sum of up to a few groups:

| | |
| --- | --- |
| Group key | `(name, source, data_type)` |
| Sources | `API`, `EVAL`, `ANNOTATION` (3) |
| Listable types | `NUMERIC`, `BOOLEAN`, `CATEGORICAL`, `TEXT` (4) |
| Theoretical max groups / name | 12 |
| Realistic split | **one type, one source** for intended usage; multiple groups only via migration or accidental reuse |
| SQL bound | `ORDER BY count DESC LIMIT 2000` on groups |
| UI bound | top **200 names** after summing the groups that made it back |

TypeScript only sees rows ClickHouse returned. If every group for a name is outside those 2000, the name is gone from every facet (numeric / boolean / categorical / level columns). If *some* of its groups make the cut, the name still appears, but with an **undercount**, and that undercount can then lose the JS top-200.

### Example (complete miss)

Name `quality` is numeric and written from all three sources:

| Group | Count |
| --- | ---: |
| `(quality, ANNOTATION, NUMERIC)` | 80 |
| `(quality, API, NUMERIC)` | 80 |
| `(quality, EVAL, NUMERIC)` | 80 |
| **Summed name count** | **240** |

The project also has 2000 other groups, each a single-source name with count **81**.

SQL `ORDER BY count DESC LIMIT 2000` keeps those 2000 groups of 81 and drops all three `quality` rows (80 < 81). `topNamesByCount` never sees `quality`, so it is absent from the numeric-name filter even though 240 would have ranked well inside the UI's top 200.

#17279's per-facet ranking would keep `quality`: it ranks names (or each facet's own measure) after grouping, so the summed 240 competes with other names, not with other groups.

### Example (undercount, then miss the UI cap)

Same `quality` split 80/80/80. Only the `ANNOTATION` row makes the 2000 (the other two sit just below the cutoff). TypeScript records `quality` at **80** instead of 240. If names 1–200 in that facet all have summed counts > 80, `quality` is sliced off the UI list even though the true total was 240.

### How likely is this?

The miss needs **two independent conditions at once**, and the first one — a name whose mass splits across groups — is not a shape that supported usage produces.

**A name only splits across groups through accidental or migration data.** The mechanism needs one name spread over several `(source, data_type)` groups, but that is not how scores are meant to be organized:

- **Sources are meant to distinguish, not overlap.** To compare an automated judge against human review, the product pairs two *distinct* scores on the same parent — Score Analytics' "Matched Data" view computes agreement between, e.g., `correctness_llm` and `correctness_human` — not one `correctness` name overloaded across `EVAL` and `ANNOTATION`. A single name spanning sources is what you get from a *migration* (a name written via the API for months, then also emitted by a new evaluator), not from a designed workflow.
- **A name is one data type.** Overloading a name across `NUMERIC` / `CATEGORICAL` / `BOOLEAN` is unsupported — Score Analytics refuses to compare across types and drops `TEXT` entirely. A name carries two types only by accident: value-inferred typing on config-less writes, or a score-config type edited mid-window (old rows keep the old type).

So the split is a property of messy data, not of a project that uses scores as intended.

**Even given a split, the `LIMIT` is a no-op** until the project has **more than 2000 distinct `(name, source, data_type)` groups** in the lookback window — on the order of thousands of distinct score names, since a well-formed name is one group. Below that, every group is returned and every sum is exact. A name that lives in one group can never fall through: its group count *is* its name count, so if it belongs in the top 200 names it is far inside the 2000-group bound.

The miss therefore requires **both** accidental/migration fan-out on a name **and** thousands of names in one window — a narrow, self-inflicted tail. Filter options already skip `FINAL` because some count inaccuracy is acceptable; this is the same class of tradeoff: exact counts for whatever was transferred, and graceful degradation of the filter list for high-cardinality projects with messy score data, not a guarantee that every name that would have ranked in the top 200 after summing is present.

## Type of change

- [x] Refactor (restructures existing code without changing behavior, with a bounded correctness tradeoff on which names are transferred)

## Mandatory Tasks

- [x] Self-reviewed the query shapes against the existing per-column helpers.

## Checklist

- Follows repo style (`pnpm run format` via pre-commit)
- No user-facing docs change (internal query shape only)
- SQL-shape unit test: one `FROM scores` / one `FROM events_core`; no `dense_rank`; parameterized `groupLimit`
- Repository comparison against the per-column helpers, including adversarial cap cases under LIMIT 2000

## Testing

- [x] Pre-commit `turbo run lint` — `Tasks: 7 successful, 7 total`, `Cached: 0 cached, 7 total`
- [x] `pnpm exec turbo run typecheck --force --filter=web --filter=@langfuse/shared` — `Tasks: 5 successful, 5 total`, `Cached: 0 cached, 5 total`
- [x] `pnpm --filter web run test event-query-builder.servertest.ts` — `Tests 53 passed (53)`
- [x] `pnpm --filter web run test score-repository.servertest.ts -t getScoresFilterOptionsForEventFacets` — `Tests 5 passed | 37 skipped (42)`
- Skipped `knip` (no unused export; no new public symbols)
- Skipped browser / preview walkthrough: query-shape only

**How to test:**

1. Open a project with scores on events.
2. Open Observations filters and Scores filters.
3. Confirm score-name / source / data-type / categorical-value options still populate.

<div><a href="https://cursor.com/agents/bc-759677df-9d49-418e-a73f-02ed971e5a79?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-759677df-9d49-418e-a73f-02ed971e5a79&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

<h2><a href="https://app.greptile.com/api/retrigger?id=62979856"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until the score query preserves each facet's top options instead of applying one unrelated global group ranking.

- Combines exact `events_core` facet queries through `UNION ALL`.
- Replaces multiple score scans with one conditional-aggregation query.
- Splits and ranks the returned score groups in TypeScript.
- Introduces a global group cap that can starve one facet with unrelated groups from another facet.
</details>

```mermaid
flowchart LR
  S[(scores)] --> G[Group by name, source, data_type]
  G --> C[Order globally by total count]
  C --> L[Keep 2,000 groups]
  L --> N[Numeric names]
  L --> B[Boolean names]
  L --> K[Categorical names]
  L --> T[Trace-level facets]
  N --> U[Event filter options]
  B --> U
  K --> U
  T --> U
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["refactor(filter-options): drop exact fac..."](https://github.com/langfuse/langfuse/commit/034bc6a192ed5abd17bdbbb5643ae0671f306c06)</sub>

### DIFF
--- a/packages/shared/src/server/queries/clickhouse-sql/event-filter-options.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/event-filter-options.ts
@@ -549,6 +549,50 @@ ORDER BY column ASC, tupleElement(option, 4) ASC, tupleElement(option, 2) ASC
   };
 };
 
+/**
+ * One ClickHouse round-trip of exact per-column GROUP BY / LIMIT scans.
+ * Used where filter-option values must stay exact (scores-view facets),
+ * instead of the approx_top_k multi-column sketch.
+ */
+export const buildEventsExactFilterOptionsForColumnsQuery = (params: {
+  projectId: string;
+  filter: FilterState;
+  columns: readonly EventFilterOptionColumn[];
+  limit: number;
+  scope?: EventFilterOptionScope;
+  sampleRows?: number;
+}): { query: string; params: Record<string, unknown> } | null => {
+  const columns = uniqueEventFilterOptionColumns(params.columns);
+  if (columns.length === 0 || params.limit <= 0) {
+    return null;
+  }
+
+  const built = columns.flatMap((column) => {
+    const queryWithParams = buildEventsFilterOptionColumnQuery({
+      projectId: params.projectId,
+      filter: params.filter,
+      column,
+      limit: params.limit,
+      scope: params.scope,
+      sampleRows: params.sampleRows,
+    });
+    return queryWithParams ? [queryWithParams] : [];
+  });
+
+  if (built.length === 0) {
+    return null;
+  }
+
+  if (built.length === 1) {
+    return built[0];
+  }
+
+  return {
+    query: built.map((part) => `(${part.query})`).join("\nUNION ALL\n"),
+    params: Object.assign({}, ...built.map((part) => part.params)),
+  };
+};
+
 /** buildEventsMetadataValuesQuery builds the top-N distinct value query for one metadata key on the events table. */
 export const buildEventsMetadataValuesQuery = (params: {
   projectId: string;
--- a/packages/shared/src/server/queries/clickhouse-sql/score-filter-options.ts
+++ b/packages/shared/src/server/queries/clickhouse-sql/score-filter-options.ts
@@ -1,4 +1,4 @@
-import { LISTABLE_SCORE_TYPES, ScoreSourceArray } from "../../../domain/scores";
+import { LISTABLE_SCORE_TYPES } from "../../../domain/scores";
 import { scoresTableCols } from "../../../tableDefinitions/scoresTable";
 import type { FilterState } from "../../../types";
 import { scoresColumnsTableUiColumnDefinitions } from "../../tableMappings/mapScoresColumnsTable";
@@ -7,10 +7,6 @@ import { createFilterFromFilterState } from "./factory";
 
 export const FILTER_OPTION_SCORE_NAME_LIMIT = 200;
 export const FILTER_OPTION_CATEGORICAL_VALUE_LIMIT = 20;
-// Enough (name, source) groups per data_type for the 200-name JS cap after
-// collapsing sources. LIMIT BY data_type keeps facets from starving each other.
-export const FILTER_OPTION_SCORE_GROUPS_PER_TYPE_LIMIT =
-  FILTER_OPTION_SCORE_NAME_LIMIT * ScoreSourceArray.length;
 
 /**
  * One `scores` scan that materialises every event-table score-name facet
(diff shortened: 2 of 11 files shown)


packages/shared/src/server/queries/clickhouse-sql/score-filter-options.ts
```diff
@@ -0,0 +1,83 @@
...
+        AND isNull(s.observation_id)
+      ) AS trace_boolean_value_count
+    FROM scores s
+    WHERE s.project_id = {projectId: String}
+      AND isNotNull(s.trace_id)
+      AND s.data_type IN ({dataTypes: Array(String)})
+      ${filterRes.query ? `AND ${filterRes.query}` : ""}
+    GROUP BY name, source, data_type
+    ORDER BY count DESC, name ASC, source ASC, data_type ASC
+    LIMIT {groupLimit: UInt32}
```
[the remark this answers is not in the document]
niklassemmler: yeah, that's the risk stated in the PR description

packages/shared/src/server/queries/clickhouse-sql/event-filter-options.ts
```diff
@@ -549,6 +549,50 @@ ORDER BY column ASC, tupleElement(option, 4) ASC, tupleElement(option, 2) ASC
...
+  if (built.length === 0) {
+    return null;
+  }
+
+  if (built.length === 1) {
+    return built[0];
+  }
+
+  return {
+    query: built.map((part) => `(${part.query})`).join("\nUNION ALL\n"),
```
maxdeichmann, line 591: How is union all more efficient than our queries right now? The database would need to scan the same data multiple times but the data is already loaded on the respective node, right?

niklassemmler: Yes. Indeed. In principle CH can also optimize if they get the full union all query, but I doubt they do it.

PR #17330, opened 2026-09-11
fix(web): server-render title, description, canonical and favicon on auth pages

jannikmaierhoefer:
🟡 **Live preview: [pr-17330 app preview](https://pr-17330.preview.langfuse.com)**

## What does this PR do?

The server response for `/auth/sign-in`, `/auth/sign-up`, `/auth/reset-password` and `/auth/setup-password` contained no `<title>`, no meta description, no canonical and no icon link — only the session-loading shell with the body text `Loading ...`. The pages set their `<title>` inside the page component, which renders behind the session gate and never reaches the server response.

In Google that showed up as: the generated title "Langfuse Cloud", a description taken from the terms-of-service sentence, sitelinks reading "Langfuse Icon. Loading ...", a generic globe favicon, and a cluster of duplicate index entries for the same page across `cloud.`/`us.`/`jp.` and per `?targetPath=` value.

Changes:

- **`DefaultHead`** (`web/src/components/layouts/default-head/`) rendered in `_app` outside `AppLayout`, so it is in every server response. Emits route-specific title/description/`og:*` for the auth pages and a bare `Langfuse` title elsewhere (no `og:*` there, so it cannot shadow a page's own title). Pages with their own `<Head>` still win (`next/head` keeps the last `<title>`, dedupes `<meta>` by name and `<link>` by `key`).
- **`rel="canonical"`** → `https://cloud.langfuse.com/auth/<page>` (no query string) on production cloud regions (EU/US/JP/HIPAA), collapsing the region and `targetPath` duplicates into one entry. Self-hosted, STAGING and DEV emit no canonical; self-hosted copy does not mention "Cloud".
- **Icon links** (`icon.svg`, `icon256.png`, `apple-touch-icon`) in `DefaultHead`, taken from the existing `useLayoutMetadata` and keyed identically to `AuthenticatedLayout`'s, so the layout's region-aware icons replace them rather than coexist.
- **`LangfuseIcon`** gets an `alt` prop defaulting to `"Langfuse"` (was `"Langfuse Icon"`). Spinner, sign-in and sign-up pass `""` because a heading next to the icon names it. `ResetPasswordPage` keeps the default — there the icon is the only content of a link. `V4MigrationStatusPage.clienttest.tsx` asserted the old alt text as a *negative* match and would have gone silently green, so it now asserts `"Langfuse"`.
- Removed the client-only `<Head>` blocks from `SignInPage` and `SignUpPage` so the SSR title and the hydrated title cannot diverge. `ResetPasswordPage` keeps its `<Head>` because "Set password" vs "Reset password" depends on the session, not the route; its casing now matches the server title and the password pages carry no region suffix. (`sign-up.tsx` became a re-export of `SignUpPage` on `main` while this PR was open; the edit moved with it.)

## Type of change

## Proof

Raw server HTML of the preview — what crawlers see, before hydration:

```bash
curl -s https://pr-17330.preview.langfuse.com/auth/sign-in | grep -oE '<title[^>]*>[^<]*</title>|<meta name="description"[^>]*>|<link rel="(canonical|icon)"[^>]*>'
```

```html
<title data-next-head="">Sign in | Langfuse Cloud</title>
<meta name="description" content="Sign in to Langfuse Cloud, the open source agent evals &amp; observability platform. EU, US, Japan, and HIPAA data regions."/>
<link rel="icon" type="image/svg+xml" href="/icon.svg"/>
<link rel="icon" type="image/png" sizes="256x256" href="/icon256.png"/>
```

On `main` the same command prints nothing. `"Langfuse Icon"` no longer appears anywhere in the response.

**Not observable on the preview:** the canonical. Previews build with `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION=DEV` (`preview-build.yml`), which intentionally emits none. The first real check is `curl -s https://cloud.langfuse.com/auth/sign-in | grep canonical` after release.

## Worth doubting

- `alt=""` on the sign-in/sign-up icon: I judged it decorative because the `<h2>` directly below names the page. If you disagree, the default `"Langfuse"` is one prop removal away.
- The canonical intentionally points every production region at the EU host. That is a choice to consolidate rather than let each region compete; it does not affect where the region switcher sends users.

## Checks (merged tree, `--force`, nothing from cache)

```
lint+typecheck  Tasks: 5 successful, 5 total   Cached: 0 cached
tests           Test Files 1 passed (1)        Tests 20 passed (20)
knip            exit 0
```

## Mandatory Tasks


jannikmaierhoefer:
@claude review

bezbac (approved):
Changes to the `LangfuseIcon` component look good

PR #17371, opened 2026-09-11
feat(transcript): implement transcript generation from observations

marliessophie:
🟡 **Live preview: [pr-17371 app preview](https://pr-17371.preview.langfuse.com)**

<h2><a href="https://app.greptile.com/api/retrigger?id=65005411"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 2/5</h2>

The PR is not safe to merge until the helper is exposed and the two transcript gaps are fixed.

Langfuse can now build conversation transcripts from generation and tool observations across a trace or session. It groups related messages into threads, removes replayed history, links tool results by call ID, and adds fixtures and docs for the supported cases.

- Adds the transcript builder and public thread types.
- Adds trace and session fixtures with structural and behavior checks.
- Documents transcript rules and the fixture creation workflow.
</details>

```mermaid
sequenceDiagram
  participant Caller
  participant Builder as getTranscript
  participant Normalizer as normalizeIO
  participant Threads as Thread states
  Caller->>Builder: observations
  Builder->>Builder: keep GENERATION and TOOL rows
  Builder->>Builder: sort by startTime
  loop each observation
    Builder->>Normalizer: normalize input and output
    alt TOOL observation
      Builder->>Builder: read gen_ai tool call ID
      Builder->>Threads: add result only to a registered call
    else GENERATION observation
      Builder->>Threads: find newest thread whose messages appear in input
      Builder->>Threads: register output tool calls
      Builder->>Threads: append new input and all output
    end
  end
  Builder-->>Caller: transcript or null
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["docs(transcript): update README to clari..."](https://github.com/langfuse/langfuse/commit/191e6b2efd39363006dcde28f6cd4644d5e0055e)</sub>


packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+```ts
+getTranscript(
+  observations: Observation[]
+): Transcript | null;
+
+type Transcript = { threads: Thread[] };
+
+type Thread = {
+  messages: ThreadMessage[];
+  observationIds: string[]; // observations that contributed, in order
```
hassiebp: this should be better instead: 
```ts
observations: { id: string; traceId; string }
```
and not separated across two lists

marliessophie: Good shout. Urgh I missed this in an iterration loop, will change right now, agree only tuple makes sense and is identifying.

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+  observationId: string; // observation that first emitted the message
+  traceId: string;
+};
+```
+
+The input is the domain `Observation` (see `domain/observations.ts`), which
+the repositories produce from ClickHouse rows. Returns `null` when no eligible
+generations produce messages.
+
+A **transcript** contains the conversation threads inferred from the supplied
```
hassiebp, line 37: Nice! Clear description

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+- Include `TOOL` observations in the supplied observations to recover responses.
+  An explicit tool observation's call ID must match a generation output's tool-call ID in the same trace.
+  Missing IDs, unmatched tools, and tools encountered before their calling
+  generation are skipped. There is no hierarchy, name or timestamp fallback.
+- The caller supplies observations from one trace or session. The builder
+  orders generations and tools by start time across the supplied traces.
+- Each observation is normalized once in this chronological pass. Generations
+  establish threads and register output tool-call IDs; tools enrich registered
+  calls. Input messages are processed before output messages for generations.
+- Generations producing no messages are skipped and do not create empty threads.
```
hassiebp, line 57: It might be worth returning the IDs of the unmatched observations. The caller can then log / handle / assign confidence on the quality of the transcript depending on the number of generations that contributed

hassiebp: Not strictly necessary for a v1

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+- Each registered call has at most one response. A tool observation with a
```
hassiebp:
```suggestion
- Each registered tool call has at most one tool result response. A tool observation with a
```

packages/shared/src/server/transcript/README.md
marliessophie: Done: [fix(transcript): pair observation references with trace ids](https://github.com/langfuse/langfuse/pull/17371/commits/986023f63fcd394fb9c6e6d4da2b3e15f8127bf3)

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+order is ignored; array order matters. `senderName`, `source`, `finishReason`
+and observation provenance are excluded. All fields inside parts are included.
+
+- **Inputs:** append only occurrences beyond the number already shown in the
+  thread, so replayed history disappears but additional identical copies survive (eg user responds "Thank you" twice).
+- **Outputs:** always append, then count them so subsequent inputs do not repeat them.
+- **Registered tool responses:** use call identity, not message occurrence counts.
+  Tool-observation responses take precedence over generation-provided responses.
+  These responses are excluded from thread matching, irrespective of their source.
+- **References:** each message has a reference to the trace and generation ID of the object that emitted it. Replayed inputs never overwrite references on earlier occurrences.
```
hassiebp, line 91: how do we know exactly what observation ID 'owns' a message as they are copied throughout all generations?

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,160 @@
...
+- **Whole-message matching:** equivalent content split into different messages
+  or parts may not match. Reordering parts within a message also changes identity.
+  Registered tool responses are the exception: replay is matched by call ID
+  within the same trace, even when grouped with other parts.
+- **Unmatched tools:** responses without a matching explicit call ID are skipped.
+  There is no hierarchy-based fallback, and tool inputs do not contribute.
+- **Performance:** replayed history is normalized and serialized for each
+  generation.
+
+## Open questions
```
hassiebp, line 137: Can you please provide a recommendation for each open questions?

packages/shared/src/server/transcript/README.md
marliessophie: We record the id of the first traceId,observationId tuple that first emitted the message (by start time).

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,158 @@
...
+and decides which, if any, is the main conversation.
+
+## Which observations contribute?
+
+- `GENERATION` observations with a non-null trace ID establish threads.
+- Include `TOOL` observations in the supplied observations to recover responses.
+  An explicit tool observation's call ID must match a generation output's tool-call ID in the same trace.
+  Missing IDs, unmatched tools, and tools encountered before their calling
+  generation are skipped. There is no hierarchy, name or timestamp fallback.
+- The caller supplies observations from one trace or session. The builder
```
nimarb, line 52: why can we as caller not just supply a trace_id (with project_id and time filter if required)?

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,158 @@
...
+  orders generations and tools by start time across the supplied traces.
```
nimarb, line 53: do we consider nesting? ie order by time within the nesting level? would make sense IMO

packages/shared/src/server/transcript/README.md
```diff
@@ -0,0 +1,158 @@
...
+- Each observation is normalized once in this chronological pass. Generations
+  establish threads and register output tool-call IDs; tools enrich registered
+  calls. Input messages are processed before output messages for generations.
+- Generations producing no messages are skipped and do not create empty threads.
+- Each registered tool call has at most one tool result response. A tool observation with a
```
nimarb, line 58: IMO it would also make sense to include tool call observations where were called but failed. they wouldn't match, but just showing tool name, input + it failed is very interesting in a transcript IMO.

can be v2

packages/shared/src/server/transcript/README.md
marliessophie: The current interface is more general, which I think is helpful. Also, separates fetching data from clickhouse from interface, which seems cleaner. What's the reason caller should not handle fetching relevant observations (via project_id, and any relevant filter conditions eg trace_id)?

packages/shared/src/server/transcript/README.md
marliessophie: Will look for examples, agree this is helpful information and should be included. Will report back and summarise/share example how this would look like in transcript response type.

packages/shared/src/server/transcript/README.md
marliessophie: Could you expand in which context nesting level would be required? Curious to hear your take on this

packages/shared/src/server/transcript/README.md
marliessophie: Discussed to expose a wrapper that fetches the events from CH.

packages/shared/src/server/transcript/README.md
marliessophie: Sync: nesting relevant for sorting observations correctly, not relevant for output.

PR #17376, opened 2026-09-11
feat(web): Group scores by name prefix

nmtrang29:
🟡 **Live preview: [pr-17376 app preview](https://pr-17376.preview.langfuse.com)**

Off main. One alternative for reading a node whose scores come in groups of many metrics (`judge.`, `rubric.`, `assertion.<id>`, `cost_channel.<id>` in production).

## What changed

| # | Area | Change | Where |
| -- | -- | -- | -- |
| 1 | Grouping rule | `groupScoresForChips`: names sharing the prefix before the first `.`, `:` or `/` form one score group once at least two distinct metric names share it; everything else groups by full name as before. Three-case client test. | `web/src/components/ScoreBadge/groupScoresForChips.ts` |
| 2 | Chips | A score group is one chip, `OutputModerationPrecision(20): Avg 0.56` (prefix, metric count, summary), same shell and size as the other chips. Counts as one chip toward the two-then-"+N" cap. Hover lists one badge per metric, labelled by suffix, in the same card the "+N" overflow uses. Story with a play test for the cap. | `grouped-score-badge.tsx`, `GroupedScoreBadges.stories.tsx` |
| 3 | Scores tab | Metadata column visible by default in the observation Scores tab (`metadataVisibleByDefault` prop, one-time visibility migration for stored state), so judge / sample size read inline without leaving the table. | `table/use-cases/scores.tsx`, `ConnectedObservationDetailView.tsx` |
| 4 | Summary rule | `groupSummary` (moved next to the grouping rule, exported) summarises only groups whose scored metrics share one data type: NUMERIC `Avg X.XX`, BOOLEAN `k/n true`, CATEGORICAL `Mostly v (k/n)`. Mixed types: no text, chip reads `Prefix(N)`, "+N" hover value empty. Categorical placeholders for "not scored" (`UNSCORED_PLACEHOLDERS`: n/a, na, none, null, unknown, not applicable, empty) count toward N but not toward the type or the tallies. One case in the existing client test. | `groupScoresForChips.ts`, `grouped-score-badge.tsx` |
| 5 | Scores tab rows | Opt-in `groupByNamePrefix` on `ScoresTable`, set only by the trace / observation Scores tabs: rows of a group render under a collapsible header row (chevron, `Prefix(N): summary`, same rule and text as the chip), all expanded by default, metric name order within, ungrouped scores last under no header. Client-side over the loaded page via `DataTable`'s existing `renderRow` slot; the project-wide Scores table is unchanged. Pure helper `groupScoreRowsByPrefix` with one client test. | `table/use-cases/groupScoreRows.ts`, `table/use-cases/scores.tsx`, `ConnectedObservationDetailView.tsx`, `TraceDetailView.tsx` |

Rule: the name before the first `.` `:` `/` is the group; 2+ distinct metric names group, names without a separator never group; a group is one chip toward the cap; summary only when all scored metrics share one data type (NUMERIC `Avg X.XX`, BOOLEAN `k/n true`, CATEGORICAL `Mostly v (k/n)`), mixed types no summary and the chip reads `Prefix(N)`; placeholders (n/a, na, none, null, unknown, not applicable, empty) count toward N but not toward the type or summary; the Scores tab groups rows under a header with the same count and summary, all expanded, ungrouped last.

Edge rules: summary k/n counts score records (a metric scored twice weighs twice), the chip count stays distinct metric names; categorical ties break by count then value, so chip and table header agree; a node with more than one page of scores (100, the server cap) lists flat rows with the normal pager and a muted "Grouping is off above 100 scores" note; `ScoresTable` is keyed by node id so collapsed groups reset on node change.

Reference: a customer forked the score UI so one evaluator with 20 metrics shows as one chip and a modal lists Metric / Score / Comment / Metadata with metadata expanded inline.

Review data: `pnpm run seed -- eval-groups --v4 --id-prefix demo-eval-groups-1` (seeder scenario lives in a local scratch branch, not in this PR).


nmtrang29:
@claude review

PR #16581, opened 2026-08-26
feat(in-app-agent): support OpenAI-compatible APIs for instance AI features

LysanderKie:
🟡 **Live preview: [pr-16581.preview.langfuse.com](https://pr-16581.preview.langfuse.com)**

## Summary
- Add `LANGFUSE_AI_PROVIDER=openai` for instance-wide AI (Assistant, Ask AI), reusing `LANGFUSE_AI_MODEL` / `LANGFUSE_AI_SMALL_MODEL` / `LANGFUSE_AI_API_KEY` / `LANGFUSE_AI_BASE_URL`.
- Optional `LANGFUSE_AI_EXTRA_HEADERS` JSON is additive (`Authorization: Bearer` still comes from the API key). Blank/whitespace headers are ignored; invalid JSON fails env validation at boot. Bedrock ignores the key, URL, extra headers, and the Responses toggle.
- `LANGFUSE_AI_USE_RESPONSES_API` (`true` / `false`, optional): unset uses Responses on first-party `api.openai.com` and Chat Completions on any other URL; `true` forces Responses; `false` forces Chat Completions.
- Chat Completions (LiteLLM, OpenRouter, other compatible gateways) send `reasoning_effort: "medium"`. First-party Responses send `reasoning.summary: auto`.
- Prompt cache: native Anthropic Messages stamp `cacheControl` only. Anthropic slugs behind a compatible client (`anthropic/…`) stamp `cache_control: { type: "ephemeral" }` on the Chat Completions wire.

Impacted packages: `@langfuse/shared`, `worker`, `web` (env schema).

## Test plan
- [x] First-party OpenAI: omit `LANGFUSE_AI_BASE_URL` — Assistant hits `/v1/responses` and shows thinking
- [x] Compatible proxy: set `LANGFUSE_AI_BASE_URL` to a Chat Completions gateway — Assistant hits `/v1/chat/completions` with `reasoning_effort`
- [x] OpenRouter/LiteLLM Claude slug (`anthropic/claude-…`) — request messages include `cache_control`
- [ ] Extra headers are additive; `Authorization: Bearer` is still sent
- [ ] Invalid `LANGFUSE_AI_EXTRA_HEADERS` JSON prevents boot; whitespace-only is ignored
- [ ] Bedrock still ignores API key / base URL / extra headers / Responses toggle

<details open><summary><h3>Greptile Summary</h3></summary>

[...]

PR #13357, opened 2026-04-26
feat(vertex-ai): allow per-LLM-key GCP projectId override with ADC

DrFaust92:
## Summary

* Adds an opt-in flag `VERTEXAI_ADC_ALLOW_PROJECT_OVERRIDE` (default `false`) so self-hosted operators can have Vertex AI honor a per-LLM-API-key `projectId` while authenticating via Application Default Credentials.
* Extends `VertexAIConfigSchema` with optional `projectId`; surfaces a corresponding (self-hosted-only) field in advanced settings of the LLM API key form.
* In `fetchLLMCompletion`, when ADC is in effect, the flag is on, and a `projectId` is configured, passes `authOptions: { projectId }` so `google-auth-library` keeps ADC for credentials but targets the configured project. Default behavior (and Langfuse Cloud) is unchanged: ADC + user-supplied `projectId` is still ignored.
* Since Vertex now routes Claude models through the Anthropic client ([#14388](<https://github.com/langfuse/langfuse/issues/14388>)), the same `authOptions` is derived once and feeds **both** Vertex sub-paths — the standard `ChatGoogle`/Vertex client and the `AnthropicVertex` (Claude-on-Vertex) `googleAuth` — so the override applies uniformly to Gemini and Claude-on-Vertex.

Refs langfuse/langfuse#12740.

## Why this is safe

* Off by default; ignored on Langfuse Cloud (gated by `!isLangfuseCloud`).
* GCP IAM still enforces which projects the ADC identity can reach — this only re-targets within already-allowed projects.
* Operator controls both the Langfuse deployment and the IAM bindings of the underlying SA / Workload Identity.

## Impacted packages

* `@langfuse/shared` — schema + env flag + `fetchLLMCompletion` branch
* `web` — LLM API key form (advanced settings) + tRPC `test`/`testUpdate` config forwarding
* `worker` — new unit tests for the override

## Test plan

- [X] `pnpm --filter @langfuse/shared run lint`
- [X] `pnpm --filter web run lint`
- [X] `pnpm --filter @langfuse/shared run build` (after `pnpm run db:generate`)
- [X] `pnpm --filter worker run typecheck`
- [X] `pnpm --filter worker run test -- fetchLLMCompletionErrors` — 4 new cases (Gemini flag off, Gemini flag on, Claude-on-Vertex flag on, ADC + flag on with no configured `projectId`); last run locally before the rebase onto [#14388](<https://github.com/langfuse/langfuse/issues/14388>), re-verified in CI
[...]

PR #17390, opened 2026-09-14
feat(in-app-agent): support Vertex AI for instance AI features

### LINKED ISSUE #17389: feat: support LANGFUSE_AI_PROVIDER=vertex for instance-wide AI features (Assistant / Ask AI)
## Problem

`LANGFUSE_AI_PROVIDER` is `z.enum(["bedrock", "anthropic", "openai"])`, so a self-hosted
deployment on GCP has no native way to power the instance-wide AI features (Assistant,
Ask AI, the natural-language filter and the search bar). The options today are:

- **bedrock** — bring AWS credentials into a GCP workload just for this feature, or
- **anthropic / openai** — introduce an API key that lives outside the GCP secret boundary.

Both add a credential that the deployment would not otherwise need.

## Vertex + ADC already exists, just not on this path

Langfuse has two places where a model gets configured, and only one of them supports
Vertex AI:

| | Project-scoped LLM connections | Instance-wide AI features |
|---|---|---|
| Configured in | UI (Settings → LLM Connections) | env (`LANGFUSE_AI_*`) |
| Powers | Playground, LLM-as-a-Judge, Prompt Experiments | **Assistant, Ask AI**, NL filter, search bar |
| Vertex AI + ADC | supported (`VERTEXAI_USE_DEFAULT_CREDENTIALS`) | not supported |

`assertInAppAgentModelConfigured()` reads `getInAppAgentModelConfig()`, which only looks at
env — it never consults project LLM connections. So configuring Vertex + ADC in the UI has
no effect on the Assistant, which is surprising once you have the connection working.

The Vertex provider itself (`buildVertexModel()`) already handles both Claude and Gemini
and already understands application default credentials. What is missing is the wiring
between the instance-wide config path and that existing provider.

## Proposal

This follows the pattern already established by #16385 (`anthropic`) and #16581 (`openai`):
add `vertex` to `LANGFUSE_AI_PROVIDER`, authenticating through application default
credentials so that on Cloud Run or GKE the attached service account is sufficient:

```bash
LANGFUSE_AI_PROVIDER=vertex
LANGFUSE_AI_MODEL=claude-sonnet-5          # or a Gemini model id
LANGFUSE_AI_SMALL_MODEL=gemini-3.7-flash
LANGFUSE_AI_VERTEX_LOCATION=global         # optional, defaults to "global"
```

Design choices, all mirroring how `bedrock` already behaves:

- **No new secret.** ADC only — deliberately no env for a service-account key JSON.
- **No env for the GCP project id.** It is resolved from the same credential chain that
  authenticates the request. (This is a narrower question than #13357, which proposes a
  per-connection project override for project-scoped LLM connections. For the single
  instance-wide model, the deployment's own ADC project is the right answer and adding an
  override would be speculative — but if you would rather keep the two paths symmetric,
  say so and I will follow #13357's shape instead.)
- One new env, `LANGFUSE_AI_VERTEX_LOCATION`, matching `LANGFUSE_AI_AWS_BEDROCK_REGION`.
- Both Claude and Gemini are accepted, because `buildVertexModel()` already supports both
  and making the instance-wide path Claude-only would be an asymmetry. Reasoning display
  and prompt caching are Claude-oriented, so Claude stays the primary target.

## Implementation notes worth discussing

- **The worker runs a different AI SDK generation than `packages/shared`.** Because of
  `@mastra/core` / `@ag-ui/mastra`, the worker pins `ai-v6`-era aliases
  (`ai-sdk-anthropic-v4`, `ai-sdk-amazon-bedrock-v4`, …) while shared is on `ai@7` with
  `@ai-sdk/*@5`. Vertex therefore needs its own `ai-sdk-google-vertex-v4` alias rather than
  reusing shared's package. Pinning `@ai-sdk/google-vertex@4.0.186` keeps
  `@ai-sdk/anthropic@3.0.111` deduped against the existing alias.
- **The worker cannot resolve `google-auth-library`** under pnpm's strict layout, so
  resolving the project id from ADC belongs in a small shared helper that imports no AI SDK
  package (importing one there would pull a second, mismatched SDK copy into the worker).
- `credentialSource === "langfuse"` is currently allow-listed to Bedrock, Anthropic and
  OpenAI in `resolveAiSdkModelConfig.ts`; Vertex has to be added there or Ask AI throws
  before the model is built.
- Claude on Vertex runs the Anthropic Messages model behind `:rawPredict`, so it reads its
  canonical `anthropic` provider-options key even though the provider name is
  `vertex.anthropic.messages`. Two things follow, both confirmed against a live Vertex
  project: adaptive thinking works through the existing options shape, and prompt caching
  needs no change because `resolvePromptCacheProvider()` already matches on `anthropic`
  (observed `cache_read_input_tokens: 47200` on a real Assistant run).

## Out of scope

Surfacing Gemini's reasoning in the UI. Gemini has no adaptive switch — thoughts require an
explicit `thinkingBudget` alongside `includeThoughts` — so it would need a default budget
decision that should reuse the budget and clamp rules in `buildThinkingConfig` rather than
restate them. That is a cross-cutting choice Ask AI faces too, so it is better as a
follow-up than folded into this wiring change.

## Status

I have a working implementation and can open a PR. It is verified end to end against a
real Vertex project using ADC only: the Assistant completes multi-turn, tool-calling
conversations on `claude-sonnet-5`, and `gemini-3.7-flash` drives conversation titles and
the search-bar filter.

Happy to adjust the shape before opening the PR if you would rather see this solved
differently.

toyayuto:
## What does this PR do?

Adds `LANGFUSE_AI_PROVIDER=vertex` for the instance-wide AI features (Assistant, Ask AI,
the natural-language filter and the search bar), following the pattern set by #16385
(`anthropic`) and #16581 (`openai`).

Before this, a self-hosted deployment on GCP had to either bring AWS credentials into a GCP
workload for Bedrock, or introduce an API key living outside the GCP secret boundary. Vertex
AI with application default credentials was already supported for **project-scoped** LLM
connections, but the instance-wide path reads env only and never consults those connections
— so configuring Vertex in the UI had no effect on the Assistant.

```bash
LANGFUSE_AI_PROVIDER=vertex
LANGFUSE_AI_MODEL=claude-sonnet-5          # or a Gemini model id
LANGFUSE_AI_SMALL_MODEL=gemini-3.7-flash
LANGFUSE_AI_VERTEX_LOCATION=global         # optional, defaults to "global"
```

Design choices, each mirroring how `bedrock` already behaves:

- **No new secret.** Application default credentials only — deliberately no env for a
  service-account key JSON, so a Cloud Run or GKE workload needs nothing beyond its attached
  service account.
- **No env for the GCP project id.** It is resolved from the same credential chain that
  signs the request. (See the issue for why this is a narrower question than #13357.)
- One new env, `LANGFUSE_AI_VERTEX_LOCATION`, matching `LANGFUSE_AI_AWS_BEDROCK_REGION`.
  Vertex ignores `LANGFUSE_AI_API_KEY` / `LANGFUSE_AI_BASE_URL` / `LANGFUSE_AI_EXTRA_HEADERS`
  and warns when they are set, the same way Bedrock does.
- Both Claude and Gemini model ids are accepted, because `buildVertexModel()` already
  supports both and making the instance-wide path Claude-only would be an asymmetry.

Fixes #17389

### Packages

- `@langfuse/shared` — env, instance-wide model config, Ask AI completion path, Vertex
  provider `credentialSource`, and a small ADC project-id helper.
- `web` — env schema and `runtimeEnv` mapping.
- `worker` — Mastra language-model factory and the Vertex reasoning options.

### Things a reviewer should look at

- **The worker runs a different AI SDK generation than `packages/shared`.** Because of
  `@mastra/core` / `@ag-ui/mastra` the worker pins `ai-v6`-era aliases, so Vertex needs its
  own `ai-sdk-google-vertex-v4` alias rather than reusing shared's package.
  `@ai-sdk/google-vertex@4.0.186` is pinned deliberately: it keeps `@ai-sdk/anthropic@3.0.111`
  deduped against the existing `ai-sdk-anthropic-v4` alias, where later patches pull a second
  copy in.
- **`vertexAuth.ts` imports no AI SDK package on purpose.** The worker cannot resolve
  `google-auth-library` under pnpm's strict layout, so it reaches this helper through the
  shared barrel; importing a provider there would load a second, mismatched SDK copy into the
  worker.
- **`resolveAiSdkModelConfig.ts` allow-list.** `credentialSource === "langfuse"` was limited
  to Bedrock, Anthropic and OpenAI; without adding Vertex, Ask AI throws before the model is
  built.
- **The ADC gate has two independent conditions**, and `requestShape.test.ts` now pins
  both directions: Langfuse-operated AI may spend application default credentials wherever
  it runs, while a tenant connection carrying the same sentinel may not. Widening the gate
  would hand a tenant the deployment's GCP identity; narrowing it would silently disable
  instance-wide Vertex wherever the Cloud region is set. Both mutations were run and each
  fails against exactly one of the two tests.
- **Claude on Vertex reads the canonical `anthropic` provider-options key**, even though the
  provider name is `vertex.anthropic.messages`. Two consequences, both verified live:
  adaptive thinking works through the existing options shape, and prompt caching needs no
  change because `resolvePromptCacheProvider()` already matches on `anthropic`.
- **`effort` sits beside `thinking`, not inside it.** `thinking.type: "adaptive"` alone
  leaves the model answering with no thinking at all, and these models reject
  `thinking.type: "enabled"` outright. The provider builds the wire `output_config.effort`
  from a top-level `effort` key. `reasoning.test.ts` pins the resulting `:rawPredict` body so
  this cannot silently regress.

### Out of scope

Choosing the model from the UI. Selection stays env-only (`LANGFUSE_AI_MODEL` /
`LANGFUSE_AI_SMALL_MODEL`), exactly as it already is for bedrock, anthropic and openai:
this PR adds no settings screen, and the instance-wide model still deliberately ignores
project-scoped LLM connections. Moving to a newer Gemini or Claude generation is an
operator env change, not a tenant action — which is the existing design, one
Langfuse-operated model whose cost the instance owner carries. Reusing project connections
here would be a different and much larger change, so it is not what this PR proposes.

Surfacing Gemini's reasoning in the UI. Gemini has no adaptive switch — thoughts require an
explicit `thinkingBudget` alongside `includeThoughts` — so it needs a default budget decision
that should reuse the rules in `buildThinkingConfig` rather than restate them. That is a
cross-cutting choice Ask AI faces too. A `model.ts` comment records the reasoning.

Self-hosting docs live in `langfuse/langfuse-docs`, so the configuration page needs a
separate PR; happy to open it once the shape here is settled.

## Type of change

## Mandatory Tasks

## Test plan

Verified end to end against a real Vertex AI project using ADC only, no keys:

- [x] Assistant completes multi-turn, tool-calling conversations on `claude-sonnet-5`
      (ClickHouse shows 10 `invoke-model` generations under `agent-turn`).
- [x] `gemini-3.7-flash` drives conversation titles and the search-bar filter.
- [x] Prompt caching is active on Vertex Claude without any change
      (`cache_read_input_tokens: 47200` on a real run).
- [x] Adaptive thinking returns reasoning parts through this path
      (`["reasoning","text"]`, non-zero `output_tokens_details.thinking_tokens`).
- [x] Gemini text generation and tool calling (`finishReason: "tool-calls"`).

Automated checks on this branch:

- [x] `pnpm run format:check` — `All matched files use Prettier code style!`
- [x] `pnpm exec turbo run lint --force` — `Tasks: 8 successful, 8 total`, `Cached: 0 cached`
- [x] `pnpm exec turbo run typecheck --force` — `Tasks: 9 successful, 9 total`, `Cached: 0 cached`
- [x] worker in-app-agent runtime — `Test Files 11 passed (11)`, `Tests 115 passed (115)`
- [x] `pnpm exec knip` — clean
- [x] `@langfuse/shared` — `Tests 1367 passed`, with two failures in `llmText.test.ts`
      unrelated to this change: they assert the provider-supported media URL path, which is
      only reached when `LANGFUSE_EVALUATOR_MEDIA_TRANSPORT` resolves to `url`. Setting
      `LANGFUSE_EVALUATOR_MEDIA_TRANSPORT=url` turns the same file into
      `Tests 24 passed (24)`.


nimarb:
@claude review

toyayuto:
One observation from wiring this up. It changes nothing in this PR, but it may be worth a
check on your side.

`effort` is not an optional tuning knob on this path. `@ai-sdk/anthropic@3.0.111` builds
`output_config` only when `effort`, `taskBudget`, or structured output is present, so
without it the request carries `thinking: { type: "adaptive", display: "summarized" }` and
no `output_config` at all. On Vertex that came back with zero thinking tokens; adding
`effort` turned the same prompt into 876. The wire-shape test in `reasoning.test.ts` pins
the resulting body so this cannot regress silently.

The `anthropic` branch of `getInAppAgentReasoningProviderOptions` does not set `effort`,
while the `bedrock` branch spells out `output_config: { effort: "medium" }`. Since Vertex
Claude and direct Anthropic both run `AnthropicMessagesLanguageModel`, direct Anthropic may
be spending no thinking either. I could not confirm it: `:rawPredict` and `/v1/messages`
could apply different server-side defaults, and I have no Anthropic key to test the direct
path.

This PR picks `medium` to match the bedrock branch, so the Assistant thinks the same amount
whichever provider an instance runs. If you would rather make the level configurable, that
seems like one setting shared by bedrock, anthropic and vertex rather than a vertex-only
env, so I left it out of here.

PR #17404, opened 2026-09-14
perf(scores): dedup UI count without ClickHouse FINAL

niklassemmler:
🟡 **Live preview: [pr-17404 app preview](https://pr-17404.preview.langfuse.com)**

## What does this PR do?

The scores UI table count (`scores.countAllFromEvents` → `getScoresUiCountFromEvents`) ran `count(*)` over `scores FINAL`. `FINAL` forces a full multi-part merge-collapse across the entire selected time window before a single row is counted, which is the dominant cost behind the count timeouts on high-volume projects.

This replaces `FINAL` on the **count path only** with a single aggregation pass that reproduces `FINAL`'s dedup semantics, then applies the filters on the reconstructed latest version:

```sql
SELECT count(*) AS count FROM (
  SELECT s.id, s.project_id, s.name,
    argMax(s.timestamp, s.event_ts) AS timestamp,
    argMax(s.value,     s.event_ts) AS value,
    argMax(s.comment,   s.event_ts) AS comment,
    ...                                       -- every filterable column
    argMax(s.data_type, s.event_ts) AS data_type
  FROM scores s
  WHERE s.project_id = {projectId}
    AND toDate(s.timestamp) >= toDate({from}) -- coarse partition/granule prune
    AND toDate(s.timestamp) <= toDate({to})
  GROUP BY s.project_id, toDate(s.timestamp), s.name, s.id
) s
WHERE s.data_type IN (...) [mutable-column + exact-timestamp filters]
```

- The GROUP BY mirrors the table's full sorting key `(project_id, toDate(timestamp), name, id)` — the exact key `FINAL` collapses on. The user-provided `id` is not unique on its own, so grouping on the whole key is required to match `FINAL`.
- `argMax(col, event_ts)` reconstructs each score's **latest** version — matching `ReplacingMergeTree(event_ts, is_deleted)` + `FINAL` — without the multi-part merge.
- All mutable-column filters (value, comment, string value, metadata, author, **and the exact timestamp bound**) are applied in the **outer** `WHERE`, on the reconstructed latest values, so the count cannot be skewed by an outdated version that once matched a filter. The inner subquery carries only a coarse `toDate(timestamp)` prune for partition/granule pruning — bucket-granular, so a sorting-key group is kept or dropped as a unit and the latest version is never lost before dedup.

Row reads still use `FINAL` (a follow-up will move those to a `LIMIT 1 BY` dedup).

## Type of change

- [x] Performance improvement (no functional change)

## Checklist

- [x] `pnpm run lint` (shared, forced)
- [x] `pnpm run typecheck` (shared, forced)
- [x] Targeted web check: `scores-trpc.servertest.ts` — 17 passed (includes count assertions)
- [x] Direct ClickHouse proof of dedup correctness for mutated-timestamp versions of the same score id

🤖 Written by Claude (an AI agent) on behalf of Niklas


niklassemmler:
@claude review

niklassemmler:
@claude review

niklassemmler:
@claude review

PR #17429, opened 2026-09-14
fix(blob-export): skip upload for empty (rows:0) export windows

niklassemmler:
🟡 **Live preview: [pr-17429 app preview](https://pr-17429.preview.langfuse.com)**

## What

When a blob-storage export window resolves to zero rows across every source table, skip the S3/Azure/GCS uploads entirely (table files, run manifest, and the legacy-source deprecation notice) instead of writing empty ~1 KB Parquet/CSV objects to the customer bucket. The cursor/watermark still advance exactly as on a normal run: the export records success and `lastSyncAt` moves forward, so catch-up and scheduling behaviour are unchanged.

A cheap pre-flight `count()` probe runs once per job over the export window (half-open `[minTimestamp, maxTimestamp)`, matching the export queries). It counts the full source-table set — a superset of whatever a single `exportSource` selects — so a zero result guarantees every selected table is empty and a non-empty window is never skipped (no data loss). Events are counted from `events_full` (the base insert target), not `events_core` (its materialized-view projection, which can lag), so the probe never undercounts what the export would stream.

This is a follow-up to the recent epoch/cursor fix for full-history blob exports (the explicitly out-of-scope optional item from that work): it removes empty-file litter generally, not just the 1970-backfill case.

## Tests

- New worker test: an empty window uploads nothing yet still advances `lastSyncAt` (and clears error/run state).
- Existing failure-path tests that drove failures through an empty window now seed one in-window row so the upload path still runs.
- `pnpm --filter worker run test blobStorageIntegrationProcessing`: 41 passed (stable across 3 runs, MinIO + ClickHouse up locally). Worker lint + typecheck pass.

<h2><a href="https://app.greptile.com/api/retrigger?id=63864378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

This PR is not safe to merge until the empty-window decision can advance the watermark without permanently excluding rows committed after the probe.

- Counts traces, observations, scores, and full events over the bounded export window.
- Preserves normal watermark and scheduling updates for skipped windows.
- Adds an empty-window regression test and seeds existing failure-path tests so they continue reaching uploads.
- The skip decision can lose rows committed after the probe, and it prevents stale deprecation-notice cleanup after source migration.
</details>

```mermaid
flowchart TD
  A[Resolve bounded export window] --> B[Count rows across source tables]
  B -->|Rows found| C[Run selected table exports]
  C --> D[Write run manifest]
  D --> E[Apply deprecation-notice lifecycle]
  B -->|Zero rows| F[Skip destination operations]
  F --> G[Advance lastSyncAt to maxTimestamp]
  E --> G
  H[Delayed in-window ClickHouse commit] -. after probe .-> G
  G --> I[Next window starts at maxTimestamp]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(blob-export): skip upload for empty ..."](https://github.com/langfuse/langfuse/commit/69615e53b849f9a175687cbd08ebed81d4c0aff1)</sub>


niklassemmler:
@claude review

niklassemmler:
@claude review

PR #17454, opened 2026-09-15
fix: preserve zero and false dataset item values

### LINKED ISSUE #17415: Dataset item scalar values 0 and false render as empty
### Describe the bug

Dataset item fields containing the valid JSON scalar values `0` or `false` are rendered as empty strings.

`stringifyDatasetItemData` documents that only `null` and `undefined` should become empty strings, but currently uses a general falsy check:

```ts
if (!data) return "";
```

As a result, the issue affects multiple consumers of this helper:

- dataset item detail view;
- edit dialog initial values;
- dataset item version diff.

### To reproduce

1. Create a dataset item with `input: false` or `expectedOutput: 0`.
2. Open the dataset item detail page.
3. Open its edit dialog or version diff.

The scalar value is displayed as an empty field.

### Expected behavior

Valid JSON scalar values should remain visible:

- `false` should render as `false`;
- `0` should render as `0`;
- only `null` and `undefined` should render as empty strings.

This is especially important in the edit dialog, where a valid stored value currently looks absent and can be overwritten unintentionally.

### Proposed fix

Use a nullish check in `stringifyDatasetItemData` and add focused regression tests for `0`, `false`, `null`, and `undefined`.

### Are you interested in contributing a fix?

Yes

HP-network:
Dataset item fields containing 0 or false are currently rendered as empty strings because the formatter treats every falsy value as missing.

Use a nullish check so valid JSON scalar values remain visible while null and undefined stay empty. Added focused regression coverage for both cases.

Fixes #17415.


HP-network:
Closing this in favor of the merged fix in #17483, which covers the same falsy dataset item behavior.

PR #17107, opened 2026-09-06
feat(public-api): add ?fromTimestamp / ?toTimestamp filters to GET /v2/datasets

Harsh23Kashyap:
## Problem

`GET /v2/datasets` previously supported only pagination (`?page`, `?limit`). Customers using the forward-compat v2 endpoint could not scope the list to a recent time window — every page scan returned the full history of the project's datasets.

The companion `GET /api/public/datasets` endpoint (PR #17002) — which serves the legacy v1 response shape with embedded `items` and `runs` arrays — already accepts `?fromTimestamp` / `?toTimestamp` on `createdAt`. The same pattern fits naturally on the v2 endpoint since v2 reads from the same `prisma.dataset` table and uses the same `createdAt` column.

The `GET /api/public/comments` (PR #15692), `GET /api/public/models` (PR #16952), `GET /api/public/llm-connections` (PR #17070), `GET /api/public/annotation-queues` (PR #17090), and `GET /api/public/score-configs` (PR #17100) endpoints all accept the same pair of parameters. Without the v2 mirror, the forward-compat path is the only one of these endpoints that does not support the time window.

Concretely:
- A customer migrating from v1 to v2 loses the `?fromTimestamp` / `?toTimestamp` filter capability on `/datasets`.
- A customer using v2 needs "datasets added since $DATE" — currently they must paginate through the full list and filter client-side.
- A bulk-rotation job that needs to know "what did I change this week" has no way to ask the v2 API.

## Proposed solution

Extend `GetDatasetsV2Query` to accept two optional ISO-8601 timestamp query params and pipe them through to the route handler as a Prisma `createdAt` range filter. The v2 route handler uses inline Prisma queries (the shared `listDatasetsForApi` service function is only used by the MCP server), so the filter is applied both in the route handler's inline query and in the shared service function so both consumers benefit.

- `?fromTimestamp=2026-08-01T00:00:00Z` → `createdAt >= fromTimestamp` (inclusive)
- `?toTimestamp=2026-08-31T23:59:59Z` → `createdAt < toTimestamp` (exclusive)
- Either param can be omitted (open-ended range). Both together form a half-open `[fromTimestamp, toTimestamp)` window.
- Filter composes with the existing `projectId` scope. Datasets outside the project are still filtered out.
- The pagination count uses the same `where` clause so `totalItems` reflects the windowed result.
- `stringDateTime` (strict ISO-8601) ensures malformed wire values return 400, not 500.
[...]

PR #17461, opened 2026-09-15
feat(public-api): add ?fromTimestamp / ?toTimestamp filters to GET /api/public/dataset-run-items

### LINKED ISSUE #17403: feat(public-api): add ?fromTimestamp / ?toTimestamp filters to GET /api/public/dataset-run-items
## Problem

`GET /api/public/dataset-run-items` does not accept `?fromTimestamp` or `?toTimestamp` query parameters, even though the sibling dataset endpoints already do:

- `GET /api/public/datasets` accepts `?fromTimestamp` / `?toTimestamp` (PR #17002).
- `GET /v2/datasets` accepts `?fromTimestamp` / `?toTimestamp` (PR #17107).
- `GET /api/public/models`, `/api/public/score-configs`, `/api/public/llm-connections`, `/api/public/annotation-queues` accept the same pair.

So users can paginate by time on every other dataset endpoint, but on the run-items endpoint they have to walk the whole dataset-run history one page at a time.

## Why it matters

A dataset run with hundreds of items is common (think 200 eval cases). Without a time filter, callers that want the items created after a known checkpoint have no choice but to fetch every page in order and discard the ones that don't match. For self-hosted customers syncing run output into their own analytics pipeline, that means N×page-count latency and N×page-count DB load on Langfuse, all to recover the items added since the last sync.

## Current behavior

```
GET /api/public/dataset-run-items?datasetId=<id>&runName=<name>
```

returns the run's items in `createdAt DESC` order, paginated by `limit`/`page`. There is no way to ask for "items created after 2026-09-01T00:00Z" or "items before 2026-09-08T00:00Z" without first walking the pages.

## Expected behavior

```
GET /api/public/dataset-run-items?datasetId=<id>&runName=<name>&fromTimestamp=2026-09-01T00:00:00Z
GET /api/public/dataset-run-items?datasetId=<id>&runName=<name>&toTimestamp=2026-09-08T00:00:00Z
GET /api/public/dataset-run-items?datasetId=<id>&runName=<name>&fromTimestamp=...&toTimestamp=...
```

returns only the run items whose `createdAt` falls in `[fromTimestamp, toTimestamp]`, inclusive on both ends (matching the existing convention in the sibling endpoints).

## Proposed solution

Mirror the pattern from `GET /api/public/datasets` (PR #17002) and `GET /v2/datasets` (PR #17107):

1. Extend `GetDatasetRunItemsV1Query` in `web/src/features/public-api/types/datasets.ts` to add optional `fromTimestamp` and `toTimestamp` fields (parsed with `z.coerce.date()`).
2. Extend `DatasetRunItemsQueryType` (the prop type passed to the underlying query helpers) to thread the same two fields through.
3. Push the timestamp filter into `generateDatasetRunItemsForPublicApi` / `getDatasetRunItemsCountForPublicApi` as additional `createdAt` predicates, so the same filter is applied to both the data query and the count query (matches the sibling endpoint implementation).
4. Update the route handler in `web/src/pages/api/public/dataset-run-items.ts` to forward the new fields.

The `dataset_run_items` table already has `@@index([createdAt])` so the new predicates use an existing index — no schema migration, no new index needed.

## Alternatives considered

- **Reuse the existing `filter: [{column, operator, value, type}]` system used inside `generateDatasetRunItemsForPublicApi`** — the inner helper already supports a `createdAt` operator. The reason this PR surfaces the timestamp as top-level `?fromTimestamp`/`?toTimestamp` query parameters (instead of inside a generic `filter[]` body) is consistency with the sibling endpoints. Top-level filters are easier to document, easier to test, and the same shape every caller already knows.
- **Add a `createdAfter`/`createdBefore` filter** instead — same idea, different names. Keeping `fromTimestamp`/`toTimestamp` matches the rest of the public API surface and avoids surprising existing users.

## Scope

- One query schema change in `web/src/features/public-api/types/datasets.ts`.
- One type change + two function-call additions in `web/src/features/datasets/server/publicDatasetService.ts`.
- One route handler update in `web/src/pages/api/public/dataset-run-items.ts`.
- One regression test that creates a dataset run with 5 items at distinct timestamps, calls the endpoint with `fromTimestamp` and `toTimestamp`, and asserts the response contains the expected subset.

## Acceptance criteria

- `GET /api/public/dataset-run-items?datasetId=...&runName=...&fromTimestamp=<iso>` returns only items whose `createdAt >= fromTimestamp`.
- `...&toTimestamp=<iso>` returns only items whose `createdAt <= toTimestamp`.
- `...&fromTimestamp=...&toTimestamp=...` returns the intersection.
- A request with no `fromTimestamp` / `toTimestamp` continues to return every item in the run, unchanged.
- `meta.totalItems` reflects the filtered count (not the unfiltered count), matching the sibling endpoints.
- Invalid `fromTimestamp` / `toTimestamp` values are rejected at the schema level (the same `z.coerce.date()` parse that catches malformed input on the sibling endpoints).

## Testing

- Prettier check on the changed files.
- `tsc --noEmit -p web/tsconfig.json` reports no new errors in the changed files.
- Local servertest (existing `web/src/__tests__/server/api/public-dataset-run-items-api.servertest.ts` if it exists, or a new one) creates the run with controlled timestamps and asserts the filter behavior on data + meta.
- Backward-compatibility check: a request without `fromTimestamp`/`toTimestamp` returns the same payload as before the change.

## Backward compatibility

The two new fields are optional and default to `undefined`. Existing callers see no change. The internal count query already accepts the same shape, so the only behavioral change is the additional WHERE clause on `createdAt`.

## Risks

- **Time zone interpretation**: `z.coerce.date()` parses ISO-8601 strings into JavaScript `Date` objects, which are timezone-aware. The existing `GET /api/public/datasets` (PR #17002) uses the same approach, so the semantics are consistent with the rest of the API.
- **Filter pushdown into the count query**: the count query must apply the same `createdAt` predicate, otherwise `meta.totalItems` would diverge from `data.length`. The implementation will thread the predicate through both queries atomically (same as the sibling endpoints).
- **Index usage**: the table has `@@index([createdAt])` and the existing query already orders by `createdAt DESC`. The new predicates slot into the same index path.

## Out of scope

- Adding `?fromTimestamp`/`toTimestamp` to `GET /api/public/dataset-run-items` v2 (does not exist yet — this is v1 only).
- Sorting or paging by `updatedAt` (different index; can be a follow-up).
- A generic filter body — the public API keeps the top-level timestamp fields.

Harsh23Kashyap:
## Summary

Adds `?fromTimestamp` / `?toTimestamp` query parameters to `GET /api/public/dataset-run-items`, mirroring the existing pattern on `GET /api/public/datasets` (PR #17002) and `GET /v2/datasets` (PR #17107).

Fixes #17403

## Problem

`GET /api/public/dataset-run-items` did not accept any time-range query parameters, even though every other dataset list endpoint does:

- `GET /api/public/datasets` accepts `?fromTimestamp` / `?toTimestamp` (PR #17002).
- `GET /v2/datasets` accepts `?fromTimestamp` / `?toTimestamp` (PR #17107).
- `GET /api/public/models`, `/api/public/score-configs`, `/api/public/llm-connections`, `/api/public/annotation-queues` all accept the same pair.

So users could paginate by time on every other dataset endpoint, but on the run-items endpoint they had to walk the entire run history one page at a time even when they only wanted the items created after a known checkpoint.

## Reproducer

A dataset run with 200 eval items is common. Without a time filter, a sync job that wants items created after the last sync point must fetch every page of the run in `createdAt DESC` order and discard the items it has already seen. That is N×page-count latency and N×page-count DB load on Langfuse, all to recover a contiguous suffix of a sorted list.

```
GET /api/public/dataset-run-items?datasetId=<id>&runName=<name>&fromTimestamp=2026-09-01T00:00:00Z
```

should return only items whose `createdAt` falls in `[fromTimestamp, toTimestamp)`, with `meta.totalItems` reflecting the filtered count.

## Proposed solution

Mirror the cycle-21 and cycle-25 patterns exactly:

1. Extend `GetDatasetRunItemsV1Query` in `web/src/features/public-api/types/datasets.ts` to add optional `fromTimestamp` and `toTimestamp` fields (parsed via `stringDateTime`, the project's shared `z.iso.datetime({ offset: true }).nullish()` schema).
2. Thread the two fields through `listDatasetRunItemsForApi` and into the underlying `generateDatasetRunItemsForPublicApi` and `getDatasetRunItemsCountForPublicApi` helpers.
3. Push a `createdAt` predicate into the ClickHouse `filter: []` array for both the data query and the count query, so `meta.totalItems` stays consistent with `data.length`. The predicate uses `>= fromTimestamp` (inclusive lower bound) and `< toTimestamp` (exclusive upper bound) to form a half-open `[fromTimestamp, toTimestamp)` window.
4. Update the route handler to forward the new query fields.

The `dataset_run_items` table already has `@@index([createdAt])` and the existing query already orders by `createdAt DESC`, so the new predicates slot into the existing index path. No schema migration, no new index needed.

## Scope

- `web/src/features/public-api/types/datasets.ts` — 2 query-schema fields.
- `web/src/features/public-api/server/dataset-run-items.ts` — extract a `buildCreatedAtFilters` helper, push the datetime filters into both the data and count queries.
- `web/src/features/datasets/server/publicDatasetService.ts` — accept and forward the two fields in `listDatasetRunItemsForApi`.
- `web/src/pages/api/public/dataset-run-items.ts` — pass query fields through.
- `web/src/__tests__/server/datasets-api.servertest.ts` — new describe block with 5 cases covering the full filter matrix + malformed-input rejection.

5 files changed, 250 insertions, 2 deletions.

## Acceptance criteria

- `...&fromTimestamp=<iso>` returns only items whose `createdAt >= fromTimestamp`.
- `...&toTimestamp=<iso>` returns only items whose `createdAt < toTimestamp`.
- `...&fromTimestamp=...&toTimestamp=...` returns the intersection.
- A request with neither field continues to return every item in the run, unchanged.
- `meta.totalItems` reflects the filtered count (not the unfiltered count).
- Invalid `fromTimestamp` / `toTimestamp` values return 400 at the schema layer (`stringDateTime` validation).

## Backward compatibility

Both new fields are optional and default to `undefined` / `null`. Existing callers see no change. The internal count query already accepts the same shape, so the only behavioral change is the additional `createdAt` predicate.

## Risks

- **Time-zone interpretation**: `stringDateTime` (`z.iso.datetime({ offset: true }).nullish()`) parses ISO-8601 strings into JavaScript `Date` objects. The same approach is used on every sibling endpoint, so semantics are consistent across the public API surface.
- **Filter pushdown into the count query**: the count query must apply the same `createdAt` predicate, otherwise `meta.totalItems` would diverge from `data.length`. The implementation threads the predicate through both queries atomically.
- **Index usage**: the table has `@@index([createdAt])` and the existing query already orders by `createdAt DESC`. The new predicates slot into the same index path.

## Testing

- Prettier check on all 5 changed files: clean.
- `tsc --noEmit -p web/tsconfig.json` reports no new errors in the changed files.
- 5 new servertest cases in `web/src/__tests__/server/datasets-api.servertest.ts` exercise:
  - no-filter baseline (all 3 items)
  - `fromTimestamp` only (lower bound, inclusive — 2 items)
  - `toTimestamp` only (upper bound, exclusive — 1 item)
  - both bounds (intersection — 1 item)
  - malformed `fromTimestamp` (400 from `stringDateTime` validation)
- Each test inserts run items at deterministic `created_at` values via `createDatasetRunItemsCh`, so the half-open window assertion is independent of clock or insertion order.

## Out of scope

- `listDatasetRunItemsByRunIdForApi` (the MCP-server variant that takes `datasetRunId` instead of `datasetId`+`runName`) is unchanged. The MCP server's existing call site does not surface user-facing timestamp filtering, so a follow-up PR can mirror this change if MCP users need the same windowing.
- A generic filter body — the public API keeps top-level timestamp fields for consistency with the rest of the public API surface.

## Notes

Consult-kiro second-opinion review was not run: the local OpenCode model server has been non-responsive for the last 13 consecutive cycles (PRs 16–29). The change is mechanical, mirrors PRs #17002 / #17107 exactly, and the new tests assert the full filter matrix. Reviewers can spot-check the four-line `buildCreatedAtFilters` helper and the four-line route-handler diff.

<h2><a href="https://app.greptile.com/api/retrigger?id=64348838"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with timestamp validation and identical filtering applied to the data and count paths.

This PR adds optional half-open creation-time filtering to the public dataset-run-items endpoint.

- Validates `fromTimestamp` and `toTimestamp` using the shared ISO datetime schema.
- Applies inclusive lower and exclusive upper bounds to both the ClickHouse row and count queries.
- Preserves existing project, dataset, run, ordering, and pagination constraints.
- Adds integration coverage for unbounded, single-bound, combined-bound, and malformed-input cases.
</details>

```mermaid
sequenceDiagram
  participant Client
  participant API as Dataset Run Items API
  participant Service as Public Dataset Service
  participant CH as ClickHouse
  Client->>API: GET with optional timestamps
  API->>API: Validate ISO datetime parameters
  API->>Service: Forward dataset, run, pagination, bounds
  Service->>CH: Fetch rows with createdAt window
  Service->>CH: Count rows with same window
  CH-->>Service: Filtered rows and total
  Service-->>Client: data and consistent meta.totalItems
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["feat(public-api): add ?fromTimestamp / ?..."](https://github.com/langfuse/langfuse/commit/01c2c048f4fc7f266314978ce8a932611a69adb3)</sub>


Steffen911:
slop

PR #17480, opened 2026-09-15
perf(events): mirror base index predicates onto the io lane

niklassemmler:
🟡 **Live preview: [pr-17480 app preview](https://pr-17480.preview.langfuse.com)**

## What does this PR do?

The `io` CTE in `buildEventsFullTableSplitQuery` filtered `events_full` only by `project_id` and a tuple semi-join against `base`: `(start_time, trace_id, span_id) IN (SELECT ... FROM base)`. That tuple cannot prune the primary key — the left side is raw `start_time`/`trace_id`, but the sort/primary key is on `toStartOfMinute(start_time)` and `xxHash32(trace_id)`. So `events_full` was scanned across all partitions and the ZSTD `input`/`output` blobs decompressed before the semi-join applied, while `base` (events_core) was already well-bounded.

This adds an `io_bounds` CTE that takes `min/max(start_time)` over `base` and bounds the io lane to that range:

```sql
io_bounds AS (SELECT min(start_time) AS io_min_start_time,
                     max(start_time) AS io_max_start_time FROM base),
io AS (SELECT ... FROM events_full e
       WHERE e.project_id = {projectId}
         AND e.start_time >= (SELECT io_min_start_time FROM io_bounds)
         AND e.start_time <= (SELECT io_max_start_time FROM io_bounds)
         AND (e.start_time, e.trace_id, e.span_id) IN (SELECT ... FROM base))
```

The derived range gives partition and primary-key pruning without re-serializing any filter values as params. Because the bound comes from `base`'s matched rows rather than the request filter, it also tightens lookups that arrive without a time filter. The range is a superset of `base`'s rows, so the `LEFT ANY JOIN` result is unchanged, and the semi-join stays for join exactness.

## Type of change

- [x] Refactor (restructures existing code without changing behavior — query is faster, results unchanged)

## Mandatory Tasks

## Checklist

- Updated the split-query unit test to assert the `io_bounds` CTE, the derived `start_time` bounds on the io lane, and the retained semi-join.
- Verified: `format:check` clean, `turbo run lint typecheck --filter=@langfuse/shared --filter=web --force` → 7 successful, `knip` clean, split-query unit tests 12 passed.

## Reviewer notes

- Correctness rests on the bound being a superset of `base`'s rows (every io row we need has a `start_time` within `[min, max]` of base) and the semi-join staying in place, so the join set is unchanged — worth a second look.
- `base` is now referenced twice (the `io_bounds` aggregate and the semi-join), so ClickHouse may scan `events_core` twice. It is the cheap table (no ZSTD io blobs), but worth confirming.
- ClickHouse pruning is not exercised by any local check (no large table locally). The pruning win — that the scalar-subquery bound actually prunes partitions/primary key — needs validation on the v4 events path (`useEventsTable=true`) via `EXPLAIN indexes=1` / read_rows on a bounded point lookup before we rely on the numbers.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — written by Claude (an AI agent) on behalf of Niklas


niklassemmler:
@claude review

PR #17487, opened 2026-09-15
feat(api)!: reject legacy reads for new organizations

maxdeichmann:
🟡 **Live preview: [pr-17487 app preview](https://pr-17487.preview.langfuse.com)**

## What does this PR do?

Adds an opt-in Cloud rollout gate for retiring authenticated legacy GET APIs for organizations created on or after a configurable cutoff. `LANGFUSE_API_ORGANIZATION_CUTOFF_ENABLED` defaults to `false`; setting it to `true` enables HTTP 410 rejections. `LANGFUSE_API_ORGANIZATION_CUTOFF_DATE` is an ISO datetime defaulting to `2026-09-16T00:00:00.000Z`, and the response’s human-readable date is derived from that value in UTC.

The enable flag is kept because the default cutoff date is today: a date-only gate would start rejecting as soon as the code deploys. Self-hosted deployments are unaffected because deprecation metadata is only attached when `NEXT_PUBLIC_LANGFUSE_CLOUD_REGION` is set.

Cutoff enforcement lives in `applyLegacyApiOrganizationCutoff` so handlers that do not go through `createAuthedProjectAPIRoute` can reuse it. Deprecated GET routes already use that wrapper. `POST /api/public/ingestion` is not a GET and is never shut down (scores stay accepted); it is not in this 410.

The rejection response is actionable and machine-readable: it includes a stable `LEGACY_API_UNAVAILABLE_FOR_NEW_ORGANIZATION` error code, the requested method/path, the route-specific replacement endpoint, a direct migration documentation URL, and the existing `_deprecation` metadata.

Every 410 cutoff rejection increments the untagged `langfuse.public_api.legacy_get_rejected` counter. A structured info log records `orgId`, `projectId`, route name, concrete API method/path, organization creation timestamp, and cutoff timestamp so rejected traffic remains attributable without high-cardinality metric tags.

The organization creation timestamp is carried through both authentication implementations: the legacy API-key Redis payload/scope and the policy-core authorization context/cache used by `API_AUTH_MIGRATION=enforce`. This avoids a PostgreSQL lookup for every rejected request. Existing legacy cache entries lack the required field, fail schema validation once, and refresh from PostgreSQL into the enriched format on their next authentication.

Older Cloud organizations, self-hosted deployments, current GET APIs, and deprecated write routes retain their existing behavior. The existing rate limiter runs before the cutoff check, and disabling the gate skips the check entirely.

Impacted packages: `web`, `@langfuse/shared`

## Type of change

- [x] Breaking change (when explicitly enabled)

## Mandatory Tasks

- [x] Self-reviewed the code, cutoff boundaries, response guidance, telemetry dimensions, cache migration, auth migration modes, and rate-limit ordering

## Verification

- `pnpm --filter web run test src/__tests__/server/unit/create-authed-project-api-route-auth-errors.servertest.ts src/__tests__/server/unit/public-api/toApiAccessScope.servertest.ts` — `Test Files 2 passed (2)`; `Tests 20 passed (20)`
- `pnpm --filter web exec eslint --max-warnings 0` on the changed files — exited 0
- `pnpm --filter web exec prettier --write` on the changed files — already formatted

The database-backed API auth integration suite (`api-auth.servertest.ts`) could not run locally because `psql`/`langfuse_test` are absent; CI `tests-web` exercises that suite. The behavior is server-side and disabled by default, so this revision can be deployed before Cloud configuration enables enforcement.

Linear Issue: [LFE-16331](https://linear.app/clickhouse/issue/LFE-16331/break-legacy-get-apis-for-new-organizations)

<div><a href="https://cursor.com/agents/bc-fd5f9cfd-0fbe-4483-bf93-7b2afc420d44?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fd5f9cfd-0fbe-4483-bf93-7b2afc420d44&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

<h2><a href="https://app.greptile.com/api/retrigger?id=64461862"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 3/5</h2>

The PR is not safe to merge until deprecated GET rejection traffic remains subject to resource limiting before it reaches PostgreSQL.

This PR blocks authenticated deprecated GET APIs for Langfuse Cloud organizations created on or after September 16, 2026.

- Returns HTTP 410 with replacement, documentation, and structured deprecation metadata.
- Preserves behavior for older Cloud organizations, self-hosted deployments, current APIs, and deprecated write routes.
- Adds boundary and exclusion-path middleware tests.
- The cutoff lookup currently precedes and can bypass organization-scoped rate limiting.
</details>

```mermaid
sequenceDiagram
  participant C as API client
  participant A as Authentication
  participant DB as PostgreSQL
  participant RL as Rate limiter
  participant H as Route handler
  C->>A: Deprecated GET with API key
  A-->>C: Reject invalid credentials
  A->>DB: Load organization.createdAt
  alt Organization created on/after cutoff
    DB-->>C: HTTP 410
    Note over RL: Rate limiter is skipped
  else Older organization
    DB->>RL: Check organization limit
    RL->>H: Execute route when allowed
    H-->>C: Legacy response
  end
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["feat(api)!: reject legacy reads for new ..."](https://github.com/langfuse/langfuse/commit/c020f00a61af61ff8988521317c7cef091bd768b)</sub>


web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
```diff
@@ -52,6 +53,9 @@ import { verifyGatewayIngestionAuthorization } from "@/src/features/ai-gateway/s
 const isJsonStringTooLargeError = (error: unknown): error is RangeError =>
   error instanceof RangeError && error.message === "Invalid string length";
 
+const LEGACY_API_ORGANIZATION_CUTOFF = new Date("2026-09-16T00:00:00.000Z");
+const LEGACY_API_ORGANIZATION_CUTOFF_HUMAN = "September 16, 2026";
```
maxdeichmann: LEGACY_API_ORGANIZATION_CUTOFF should be an env variable and LEGACY_API_ORGANIZATION_CUTOFF_HUMAN should be derived

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
```diff
@@ -396,6 +400,57 @@ export const createAuthedProjectAPIRoute = <
       });
     }
 
+    if (
+      env.LANGFUSE_LEGACY_GET_API_NEW_ORG_CUTOFF_ENABLED === "true" &&
+      req.method === "GET" &&
+      deprecation &&
+      auth.scope.orgId
+    ) {
+      const organization = await prisma.organization.findUnique({
```
maxdeichmann: This will hammer postgres. Doesnt the API key in redis already contain the created_at?

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
```diff
@@ -396,6 +400,57 @@ export const createAuthedProjectAPIRoute = <
...
+        const apiPath = clickHouseRouteForRequest(req);
+        const rejectionContext = {
+          orgId: auth.scope.orgId,
+          projectId: auth.scope.projectId,
+          apiRoute: routeConfig.name,
+        };
+        recordIncrement(
+          "langfuse.public_api.legacy_get_rejected",
+          1,
+          rejectionContext,
```
maxdeichmann: This will be a super high cardinality metric. IMO keep the counter but remove all the tags.

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
marksalpeter, line 1: Are the only deprecated routes guarded by `createAuthedProjectAPIRoute` or is this missing some important routes (eg the old ingest endpoint)?

The depreciation enforcement logic should be encapsulated by a helper function of some kind for the sake of readability and reuse in handlers not guarded by this middleware.

web/src/env.mjs
```diff
@@ -596,6 +596,14 @@ export const env = createEnv({
...
     LANGFUSE_API_TRACES_DEFAULT_FIELDS: z.string().optional(),
     LANGFUSE_API_TRACEBYID_DEFAULT_FIELDS: z.string().optional(),
+    // Cloud rollout gate for rejecting legacy GET APIs for organizations
+    // created on or after the cutoff date.
+    LANGFUSE_LEGACY_GET_API_NEW_ORG_CUTOFF_ENABLED: z
+      .enum(["true", "false"])
+      .default("false"),
+    LANGFUSE_LEGACY_GET_API_NEW_ORG_CUTOFF: z.iso
+      .datetime()
+      .default("2026-09-16T00:00:00.000Z"),
```
marksalpeter, line 610: Why do we need to enable/disable this functionality. Setting the date in the far future would disable this if we needed to for emergency purposes. Do we want to let self hosters continue to use deprecated apis? 

Also, the naming is strange. Names should be aligned with the variable names `LANGFUSE_API_ORGANIZATION_CUTOFF_DATE`, and if truly required `LANGFUSE_API_ORGANIZATION_CUTOFF_ENABLED`

marksalpeter (commented):
There are merge conflicts, but the cursor agent that worked on this appears archived. I cannot unarchive from my side. Maybe you can try @maxdeichmann

web/src/features/public-api/server/createAuthedProjectAPIRoute.ts
maxdeichmann: We use createAuthedProjectAPIRoute in almost all public APIs. Rare exceptions is the old ingestion API. All legacy APIs should be covered. We use this like:
```
export default withMiddlewares({
  GET: createAuthedProjectAPIRoute({
    name: "Get Dataset Item",
    querySchema: GetDatasetItemV1Query,
    responseSchema: GetDatasetItemV1Response,
    rateLimitResource: "datasets",
    fn: async ({ query, auth }) =>
      await getDatasetItemForApi({
        datasetItemId: query.datasetItemId,
        projectId: auth.scope.projectId,
      }),
  }),
```

web/src/env.mjs
maxdeichmann: i will keep both: we need a cut-off which is persisted and does not change. we may want to disable the guard though for such orgs temporarily to see how customers are impacted by this change.

marksalpeter (approved):
Looks good to me 👍

PR #17531, opened 2026-09-16
fix(support): block Sev-1/Sev-2 requests from the demo organization

jannikmaierhoefer:
🟡 **Live preview: [pr-17531 app preview](https://pr-17531.preview.langfuse.com)**

## Problem

Severity 1 and 2 in the in-app support form were gated on the organization's plan alone — `isEnterpriseSupportPlan(organization?.plan)` on the client, and the same check inside `mapToPylonCaseSeverity` on the server.

Every Langfuse Cloud user is a member of the demo organization (`createProjectMembershipsOnSignup.ts`), and that organization runs on an Enterprise plan so the demo shows the full feature set. The support drawer is mounted globally and reads the org from the URL, so **any** user could open it on a demo page and file a Sev-1 or Sev-2 request. Sev-1 pages the on-call team.

This came out of a real ticket: a user on a Core-plan account raised a Sev-2 while on the EU demo org's settings page. The form was working exactly as written — the plan it saw really was `cloud:enterprise`.

## Change

- Exclude `NEXT_PUBLIC_DEMO_ORG_ID` from the high-severity gate, in the form UI and independently in the tRPC safeguard. The env var is already set per region, so EU, US and JP each exclude their own demo org.
- `mapToPylonCaseSeverity` takes a required `isDemoOrganization`, so a caller cannot skip the new rule by omission.
- Skip the `langfuse_customer_tier` account write for demo-org requests. Pylon accounts are keyed on the requester's email domain, so the demo organization's Enterprise plan would otherwise overwrite the tier of whichever real account they belong to — which is what happened on the ticket above.

The demo organization's plan is deliberately Enterprise so the demo exercises every feature, so this is a support-form exclusion rather than a change to the org's plan or entitlements.

## Worth doubting

- **If `NEXT_PUBLIC_DEMO_ORG_ID` is ever unset or wrong in a region, the hole reopens silently** — there is no second line of defence. A role check (Sev-1/2 only above `VIEWER`) would fail safe here, since demo membership is `VIEWER` by construction; it was deliberately left out of this PR to keep the change narrow, and it would also have taken Sev-2 away from genuine Enterprise viewers.
- Self-hosted deployments that set `NEXT_PUBLIC_DEMO_ORG_ID` for their own demo org inherit this exclusion. That seems right, but it is a behaviour change for them.
- The disabled-reason copy still reads "Severity 2 is available on the Enterprise plan." A demo-org user is technically on an Enterprise plan, so the reason is imprecise there. Left as-is rather than threading a second prop through `SupportForm` for a context where nobody is a real customer.
- The gate is still evaluated against **whichever org is in the URL**, not against the account the Pylon ticket lands on. This closes the demo-org hole; it does not make plan and ticket destination agree in general.

## Test steps

On `pr-17531.preview.langfuse.com`, with `NEXT_PUBLIC_DEMO_ORG_ID` / `NEXT_PUBLIC_DEMO_PROJECT_ID` set and the demo org on an Enterprise plan:

1. Open the demo project as any user and open the support drawer → the Priority dropdown offers only Severity 3; Severity 1 and 2 are disabled.
2. In a non-demo Enterprise org → Severity 1 and 2 are selectable, and a submitted Sev-2 arrives in Pylon with `case_severity: Sev-2`.
3. From the demo org, call `supportRouter.createSupportThread` directly with `severity: "Severity 2 (Major Business Impact)"` → the issue is created with `Sev-3` and the Pylon account's `langfuse_customer_tier` is left alone, confirming the server gate is independent of the UI.

## Checks

- `pnpm exec turbo run lint --force --filter=web` → `Tasks: 4 successful, 4 total`, `Cached: 0 cached, 4 total`
- `pnpm --filter=web run typecheck` → passed (after `shared` `db:generate` + `build`)
- `pnpm --filter web run test-client src/__tests__/supportCaseSeverity.clienttest.ts` → `Tests  13 passed (13)`. The new demo-org case was confirmed red first: with `!isDemoOrganization` removed it fails with `expected 'Sev-1' to be 'Sev-3'`.
- `pnpm --filter web run test-client src/features/support-chat/SupportDrawerProvider.clienttest.tsx` → `Tests  3 passed (3)`
- `pnpm exec knip` → clean

Not run: `@langfuse/native` lint and `ai-gateway` typecheck, which need a Rust toolchain that is not installed on this machine. Both are untouched by this TypeScript-only change. No browser signoff — the change is a disabled-option gate covered by the tests above; step 1 is the glance worth taking on the preview.


jannikmaierhoefer:
@claude review

PR #17535, opened 2026-09-16
feat(api): filter observations by whole-session duration

jweden33:
## What does this PR do?

Adds optional `minSessionDuration` (inclusive, finite, nonnegative seconds) to `GET /api/public/v2/observations`. Callers can discover qualifying session IDs with `fields=basic`, then retrieve details, without downloading observations from short sessions first. Two short traces separated by a long idle gap can qualify even when neither trace's latency reaches the threshold.

This targets current `main`: the legacy Sessions endpoints are deprecated and unavailable in v4. The supported observations API has no equivalent session aggregate filter, and metrics v2 does not support grouping by `sessionId`.

The implementation reuses `EventsSessionAggregationQueryBuilder` and its Sessions UI duration formula, and applies a project-scoped membership filter before existing cursor pagination. The basic, split-IO, and content-filtered full-table query paths share it. Content filters retain the upstream single-pass IO/metadata projection while duration filtering still precedes pagination. Project API-key authentication, response fields, existing filter precedence, and behavior when the parameter is omitted are unchanged. No new dependencies or schema migrations.

The duration query first selects the latest `event_ts` for each `(project_id, trace_id, span_id, start_time)`, then filters deleted/empty-session observations and groups by session. This handles corrected or null end times and moved session membership before background merges, without `FINAL`. It projects only session ID and duration. The existing observation builder already supports latest-version selection; the session builder gains a supplied query source and a duration field set so the formula can be reused without changing existing UI/export callers.

Affected packages: `@langfuse/shared`, `web`; Fern source and the generated served OpenAPI spec are updated.

### Semantics and limitations

- Duration counts second boundaries from the earliest observation start to the latest end, using start time when end is missing. It includes gaps across traces. Fractional thresholds compare against that existing whole-second value.
- `fromStartTime` is inclusive and `toStartTime` exclusive for returned observation start times. Session members outside the requested interval still contribute to duration. A separate exact plan access floor prevents inaccessible history from contributing.
- Existing `tags`/`traceTags` filters remain observation-level predicates. `all of` requires all tags on a matching observation; it does not union tags across different session members. Tags do not truncate the duration calculation.
- Latest-version selection applies to duration calculation only; it does not deduplicate returned observation rows. During storage merges, this can differ from the UI aggregation over unmerged versions while retaining the same duration formula.
- Pagination remains observation-based. Callers must deduplicate `sessionId` across pages. Supplying the threshold, including zero, excludes rows without a session ID.
- Each filtered page sorts/selects latest observation versions and aggregates retained, accessible project session history. A narrow observation time range does not bound this aggregation; active sessions can change between pages. This does not backport the feature to v3.194.1.

Example with project API keys (URL-encode structured filters):

```sh
curl --user "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" --get \
  "$LANGFUSE_HOST/api/public/v2/observations" \
  --data-urlencode 'fields=basic' \
  --data-urlencode 'minSessionDuration=60' \
  --data-urlencode 'fromStartTime=2026-09-01T00:00:00Z' \
  --data-urlencode 'toStartTime=2026-09-02T00:00:00Z' \
  --data-urlencode 'filter=[{"type":"arrayOptions","column":"tags","operator":"all of","value":["review"]}]'
```

## Verification

Executed locally after merging upstream `main` through `53640f4ca`, with Node 26.3.0, pnpm 12.3.1, and the official ClickHouse 25.12.10.7 macOS binary (same 25.12 series as the development stack; release checksum verified):

- `pnpm --filter web run test get-observations-v2-query.servertest.ts`: **11 passed**.
- `CLICKHOUSE_BIN=/path/to/clickhouse pnpm --filter @langfuse/shared run test src/server/repositories/events-session-duration.test.ts src/server/repositories/events.test.ts src/server/queries/clickhouse-sql/event-query-builder.test.ts`: **20 passed**, including real `clickhouse local` execution. The version regression fails against the original duration aggregation (**1 failed; 4 unselected tests skipped**) and passes with this fix. Synthetic Memory-table rows retain all versions so background merges cannot mask the regression. Coverage includes shortened/null end times, moved/cleared membership, deletions, tenant isolation, and equal span IDs across different traces/start times. Other checks exercise basic, split-IO, and content-filtered full-table paths, cursor placement, and the exact plan access floor. The combination regression includes IO, expanded metadata, tags, content filtering, and session duration without leaking row filters into the duration aggregate.
- `pnpm exec turbo run lint --filter=web --filter=@langfuse/shared --only --force`: **Tasks: 2 successful, 2 total; Cached: 0 cached, 2 total**.
- Shared build, web typecheck, worker typecheck, and `pnpm exec knip`: **exit 0**.
- Prettier and `git diff --check`: **passed**.
- `pnpm run openapi:check`: **Served OpenAPI specs match their Fern sources.**
- `EXPLAIN PLAN actions=1` on ClickHouse 25.12.10.7 showed that the optimizer already prunes unused aggregates from the previous broad CTE. The explicit duration projection simplifies the generated query; no measured speedup is claimed. Latest-version selection still adds work across accessible project history on each filtered page.

Environment limitations, not passing tests:

- The four added HTTP integration cases in `observations-api-v2.servertest.ts` remain unverified against a full stack. The prior attempt with `LANGFUSE_MIGRATION_V4_ALLOW_PREVIEW_OPT_IN=true`, selecting `whole-session duration|unfiltered results|inaccessible history`, produced **4 failed in fixture setup, 42 unselected tests skipped** because PostgreSQL was unavailable. This machine has no Docker/local database stack. The isolated ClickHouse regression above verifies SQL behavior, but not the HTTP authentication/integration path. The HTTP cases cover multi-trace gaps, cross-boundary members, tags, cursor pages, project isolation, zero/null-end behavior, advanced-filter precedence, and the plan access floor.
- Full-repository `pnpm run lint` and `pnpm run typecheck` previously stopped because Rust `cargo` is unavailable for `ai-gateway`; they were not rerun for this revision. The affected TypeScript checks above ran separately. The documented `LANGFUSE_PRE_COMMIT_SKIP_LINT=1` switch skipped redundant full-repository lint at commit time; the formatting hook remained enabled and passed. Affected-package lint was forced again after merging upstream.
- No UI changes; no browser rendering test or data seeding was needed. Regression fixtures are synthetic and use existing test helpers. The real SQL test uses the repository's existing optional `clickhouse local` harness and skips when that binary is unavailable.

## Type of change

- [x] API documentation updated

## Mandatory Tasks

- [x] Read the contributing guide and self-reviewed the code, including shared-query callers and both v2 projection paths.


jweden33:
Hi maintainers, could someone please approve the pending fork workflow runs so CI can validate this PR?

This adds optional whole-session duration filtering to the supported v2 observations API. The PR does not change workflow files or add dependencies. Both automated review threads are resolved, and the CLA check has passed.

Local verification results and the remaining full-stack integration checks are documented in the PR description. I’m happy to address any CI failures or further feedback. Thanks!

PR #17288, opened 2026-09-10
feat(web): trace detail redesign — summary strip, quiet headers, attributes, hover cards

nmtrang29:
Split into a stack for review: #17446, #17447, #17448, #17449 (merge in order). This PR stays as the reference tree.

🟡 **Live preview: [pr-17288 app preview](https://pr-17288.preview.langfuse.com)**

Redesign of the single trace view, shipped as one PR without a feature flag. The view read as a data wall: trace-level facts repeated as identical pills on every observation, a UUID as loud as cost, tokens and model competing with the name in every tree row. Each fact now has one home and one visual grammar: metrics are quiet text, references are links, tags are the only pills, attributes are a table. Sans everywhere, mono only inside JSON.

Supersedes #17078, #17079, #17080 and #17082 (closed, commits are all here). The experimental lanes view stays parked in #17083. Try it on the preview above; the demo project has a `help-assistant-*` trace family with attributes, scores and multi-model generations.

## What changed

| # | Area | Change | Where |
|---|---|---|---|
| 1 | Summary strip | Latency, cost, tokens (breakdown on hover of the token count), `Session ↗`, `User ↗`, tags capped at 3 + N. Always visible whatever is selected. Env / release / version stay on the trace header as key:value text. | `TraceSummaryStrip`, `TraceMetadataBadges`, `CostUsageBadge` |
| 2 | Headers | Metrics only, one row: start time, latency, TTFT, cost, tokens, model. Session, user, tag and attribute badges removed from observation headers. Titles truncate with an ellipsis instead of wrapping to two lines. Combined "Add to" menu. | `TraceDetailViewHeader`, `ObservationDetailViewHeader`, `ObservationMetadataBadgesSimple` |
| 3 | Attributes table | Model, environment, release, version, session_id, user_id and model parameters render in the Metadata table style, in the Attributes tab. Row menu: copy, include in filter (`key:value`), exclude from filter (`-key:value`). Lands on the observations or traces table with the smallest date preset covering the row. Fixed keys win over colliding model parameters. | `ObservationAttributesList` (+ unit test), `ValueCell` (`AttributeActionsMenuContent`), `eventsTablePaths`, `IOPreview*` |
| 4 | JSON tables | No header row on Input, Output, Attributes, Metadata. The section title already names the table. | `PrettyJsonView` (`hideHeader`) |
[...]

PR #17536, opened 2026-09-16
feat(web): persistent trace summary strip with tags

nmtrang29:
🟡 **Live preview: [pr-17536 app preview](https://pr-17536.preview.langfuse.com)**

Cherry-picked out of the #17288 trace-view stack as a single-concern PR. ref LFE-16384

Moves existing badges to a new persistent row and surfaces trace tags. No value is recomputed and no badge is restyled.

## What changed

| Area | Before | After |
|---|---|---|
| Trace tags | Never rendered anywhere in the trace view | Shown in the strip, capped at 3 with a `+N` / "Show fewer" toggle |
| Latency, cost, tokens | On the trace detail header, replaced as soon as an observation is selected | In the strip, stays put |
| Session / user links | On both the trace header and the observation header | In the strip only |
| Observation detail header | Rendered `sessionId` / `userId` | Dropped, both are trace-level values |
| Trace detail header | Owned latency, cost, tokens, session, user, env, release, version | Keeps env, release, version and the target-trace badge |
| Annotation queue processor | Showed none of that badge row | Unchanged, the strip is gated out |

## Same numbers, same components

- Latency renders `trace.latency ?? null`, the identical expression the header used. The context also exposes a client-derived `traceDuration`, which is a different number; it is deliberately not used here.
- Cost and tokens call the same `aggregateTraceMetrics(observations)`. The array is the same one the header received: `TracePanelDetail` already reads it from `useTraceData()` and passed it down, and the strip reads it from that same context.
- `LatencyBadge`, `CostBadge`, `UsageBadge`, `SessionBadge`, `UserIdBadge` and `TagButton` are used exactly as they render on main. None of the stack's badge restyling is included.
- The one non-move is a two-line type addition: `TraceDataContext`'s `TraceType` gains the optional `latency` field that `TraceDetailViewHeaderProps` already declared locally. Types only, no runtime effect.

`observations` is no longer passed to `TraceDetailViewHeader`, since the aggregation moved to the strip.


nmtrang29:
@claude review

nmtrang29:
@greptileai review

bezbac (approved):
Looks good, just on mobile some badges are cut off weirdly. Might make sense to address this in this PR already

Image

bezbac (approved):

nmtrang29:
recheck

PR #17446, opened 2026-09-14
fix(web): remove dead markdown toggle, tabs inset and badge polish

nmtrang29:
🟡 **Live preview: [pr-17446 app preview](https://pr-17446.preview.langfuse.com)**

Part 1 of 4 of the trace view redesign, split from #17288 for review.

## What changed

| Change | Why | Pair |
|---|---|---|
| Dead markdown toggle path removed: `useMarkdownContext` provider, `canEnableMarkdown` plumbing in `MarkdownViewer`, `MarkdownJsonView`, `ChatMessage`, `_app` | Markdown toggle had 0 events in 90d | 1 |
| `PrettyJsonView`: `hideHeader` drops the Path / Value header row, `hoverControls` reveals copy and expand-all on section hover | Section title already names the table | 2 |
| `Tabs`: boxed tab list trigger fills the inset height, inner radius nests inside the outer | Fixed height overflowed once the list had a border | 20 |
| Peek menu: Share row matches Delete and Expand rows in height and icon size (`publish-object-switch`) | Three rows line up | 21 |
| `ItemBadge`: `hideIcon` label-only quiet chip for page and peek headers | Label names the type, icon repeated it | 7 |
| `SingleLineOverflowList`: `spacing` comfortable variant (gap-3) | Rows of plain text need more air than boxed pills, used by the session header in part 4 | 12 |
| `action-button-count-badge`: count hidden when zero | Zero badge reads as a pending state | 18 |
| PostHog allowlist: `attribute_table_action`, `io_section_collapse_toggle` added, unused `io_pretty_format_toggle_group` removed; storybook preview drops the markdown provider | Events fire from part 3 | 8 |

## How to review

Preview `https://pr-17446.preview.langfuse.com`: open a trace, check the Formatted / JSON tab list and the JSON tables without a header row, then the peek ⋯ menu on the traces table.

Before/after per row: https://linear.app/clickhouse/issue/LFE-16290/remove-dead-markdown-toggle-tabs-inset-and-badge-polish

## Stack

1. **#17446 shared cleanup (this PR)**
2. #17447 trace page chrome
3. #17448 observation detail
4. #17449 scores and sessions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #17540, opened 2026-09-16
fix(web): observation Attributes tab

nmtrang29:
🟡 **Live preview: [pr-17540 app preview](https://pr-17540.preview.langfuse.com)**

Cherry-picked out of the #17288 trace-view stack as a single-concern PR. ref LFE-16388

## What changed

| Area | Before | After |
|---|---|---|
| Observation detail tabs | Preview, Scores, Log | Preview, **Attributes**, Scores, Log |
| Attributes tab | n/a | Three tables: Attributes, Model parameters, Metadata |
| Attributes content | n/a | The fixed keys Langfuse knows: `model`, `environment`, `release`, `version`, `session_id`, `user_id` |
| Row actions | Metadata rows offered copy + metadata-filter shortcuts | Attribute rows offer copy, "Include in filter" and "Exclude from filter" against the attribute's own table column, labelled in search-bar grammar |
| Filter link target | n/a | `buildEventsTablePathForColumnFilter` lands on the observations or traces table, merging into existing filters and widening the date window so the source row is in range |
| Model parameters rows | n/a | Copy only — no table column exists to filter on |
| Preview tab | Rendered the metadata table below the output | `showMetadata: false`, Preview is input and output |
| Observation header | Rendered `ModelParametersBadges` | Dropped, model parameters are a table in the tab now |
| `ModelParametersBadges` | Used by the header | No callers left, deleted along with its stories |
| `PrettyJsonView` | No way to drop the Path / Value header row | `showHeader` prop, defaults to `true`; the three tables pass `false` since each carries a section title |
| `DetailTab` | `preview \| log \| scores` | Gains `attributes`, so `?tab=attributes` is a valid deep link |

Nothing appears in two places: metadata left Preview and the model parameter badges left the header.

`showHeader` is the prop bezbac asked for on #17446 in place of `hideHeader` — positive, defaulting true. It had no caller then so it was dropped; these three tables are its first.

## Deliberately not included

- The `ValueCell` hover-affordance restyle that shared these commits (border and shadow removed from the copy and menu buttons, row min-height). Reverted to main's styling.
- The quiet detail header restyle and the `ModelBadge` move.
- The Scores tab count badge.

## Known duplication left alone

`environment`, `release` and `version` still render on the observation header as well as in the Attributes table. Removing them is the quiet-header restyle, which is its own change.

## How we will know it worked

- `trace_detail:detail_tab_switch` already sends the tab value, so `attributes` shows up without a new event.
- `trace_detail:attribute_table_action` reports `table` (attributes / model_parameters / metadata) and `action` (copy / include_filter / exclude_filter), never the key or value.


nmtrang29:
@claude review

nmtrang29:
@greptileai review

web/src/components/table/ValueCell.tsx
bezbac, line 1: I think this is badly abstracted. The generic `ValueCell` should not be aware of specifics of the attribute tab

web/src/__tests__/eventsTablePaths.clienttest.ts
bezbac, line 1: Is this needed?

web/src/features/traces/components/ObservationDetailView/components/ObservationDetailViewHeader/ObservationDetailViewHeader.tsx
bezbac, line 1: Hmm, not sure if removing environment, release and version here makes sense. This is something that user might actually care about.

web/src/features/traces/fns/attributeColumnFilter.ts
bezbac, line 1: Why is this named like a component but does not export any components and just exports pure functions?

web/src/features/traces/components/ObservationAttributesList.ts
```diff
@@ -0,0 +1,110 @@
...
+      return options("sessionId", target);
+    case "user_id":
+      return options("userId", target);
+    case "version":
+      return text("version", target);
+    case "release":
+      // Observations have no release column.
+      return target === "traces" ? text("release", "traces") : null;
+    default:
+      return null;
```
bezbac: I think it makes sense to avoid switch cases with default when the range of known discriminators is known. You can use if clauses here with a `assertUnreachable` at the end instead.

bezbac (changes requested):

web/src/features/traces/components/ObservationDetailView/components/ObservationDetailViewHeader/ObservationDetailViewHeader.tsx
nmtrang29: i left latency / model / ttft / token in the header - env / release / version are still there just under the new attribute tab now but i can see if i can find some data around filtering on these attributes. The header would be really crowded again if we keep everything

web/src/__tests__/eventsTablePaths.clienttest.ts
nmtrang29: this is now removed

web/src/features/traces/fns/attributeColumnFilter.ts
nmtrang29: moved to features/traces/fns/observationAttributes.ts where the other pure fn modules live.

web/src/features/traces/components/ObservationAttributesList.ts
nmtrang29: (claude)  switched to a Record lookup with ?? null, no switch and no default. can't use assertUnreachable here though, key is an arbitrary string off the table rows rather than a closed union, so unknown keys need a fallback

web/src/components/table/ValueCell.tsx
nmtrang29: Claude: ValueCell takes a rowActions render slot now and knows nothing about attributes. MetadataFilterActions is back to just projectId + filterTarget like on main, and the attribute menu moved to AttributeRowActions.tsx in the traces feature with its own analytics. copyOnly is gone too, model parameters just pass no actions.

web/src/components/table/ValueCell.tsx
```diff
@@ -524,16 +527,20 @@ export const ValueCell = memo(
...
+            renderMenu={() =>
+              rowActions ? (
+                rowActions(row)
+              ) : metadataActions ? (
+                <ValueCellActionsMenuContent
+                  row={row}
+                  metadataActions={metadataActions}
+                />
+              ) : null
+            }
```
bezbac, line 543: Hmm we should probably make this consistently only take rowActions. But we can do that in a future PR.

bezbac:
```
rowActions?: (row: Row<JsonTableRow>) => React.ReactNode;
```

is also not a good abstraction since it allows arbitrary JSX.

web/src/features/traces/components/ObservationDetailView/ConnectedObservationDetailView.tsx
```diff
@@ -119,9 +127,18 @@ export function ConnectedObservationDetailView({
...
+    version: observation.version,
+    sessionId: observation.sessionId,
+    userId: observation.userId,
+  });
+  const modelParameters = buildModelParameters(observation.modelParameters);
 
-  // Hide entire tabs bar when only Preview tab remains (cleaner annotation mode UI)
-  const showTabsBar = showLogViewTab || showScoresTab;
+  // Attributes is always present, so the bar always has more than Preview.
+  const showTabsBar = true;
```
bezbac: `showTabsBar` now seems unnecessary

web/src/features/traces/fns/attributeColumnFilter.ts
bezbac, line 1: This does not match the architecture laid out by Nikita. Each exported function should be it's own file where the function name matches the file name.

bezbac (changes requested):

web/src/features/traces/components/ObservationDetailView/ConnectedObservationDetailView.tsx
nmtrang29: i removed this

web/src/components/table/ValueCell.tsx
nmtrang29: leaving it as is per your message on slack

web/src/features/traces/fns/attributeColumnFilter.ts
nmtrang29: (Claude)
split into buildObservationAttributes, buildModelParameters, attributeColumnFilter and attributeGrammar, each with its own test file. types and the private filter helpers sit with attributeColumnFilter since they only describe its return.

bezbac (approved):
Again, not sure if it's fine to move the tags (this is your call) but the code looks okay

PR #17402, opened 2026-09-14
fix(web): update JSON viewer

nmtrang29:
🟡 **Live preview: [pr-17402 app preview](https://pr-17402.preview.langfuse.com)**

JSON table restyle: one look everywhere, replacing the bordered Path / Value table.

- One table style, no variants or picker
- No Path / Value header row and no inner box under a title; the section title is the frame. An `sr-only` header keeps the columns named for screen readers
- Key column sizes to its keys (`min-w-40`, capped at 40cqw); chevron gutter at icon width, keys start 16px in
- Dotted OTel keys wrap at the dots; collapsed array and object previews stay on one line and ellipsise
- Keys and values both mono `text-xs/5`; hairline row dividers at `border/60`
- Bare string values, quotes restored on copy
- Sticky top-level keys stay on the value baseline
- Expanded JSON-only session messages take the bubble's full width
- JSON Beta log rows keep the Formatted table, with a regression test

The font, the value type colour tokens, the control alignment and the tag changes are split into their own PRs; this rebases once they land.

Review on the preview: open a trace, stay on Formatted, check Input / Output / Metadata, expanded Log View rows, and a session message card.

<h2><a href="https://app.greptile.com/api/retrigger?id=64336983"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no actionable correctness, security, or repository-rule violations remain.

<details open><summary>Summary</summary>

This PR standardizes the formatted JSON viewer and related trace/session surfaces.
- Replaces the bordered Path/Value layout with an auto-sized, screen-reader-labelled table.
- Adds bare string display while preserving JSON quoting when copying.
- Aligns JSON view preferences across trace, observation, and log views.
- Keeps JSON Beta expanded log rows in the formatted table.
- Updates message sizing, section framing, wrapping, truncation, and sticky-key behavior.
</details>

<sub>Reviews (10) · Last reviewed commit: ["chore(web): merge origin/main after IBM ..."](https://github.com/langfuse/langfuse/commit/8e802d855203595bfe8cb0cd7c9449c75e96dd50)</sub>

PR #17557, opened 2026-09-16
fix(web): use IBM Plex Mono for monospace text

nmtrang29:
🟡 **Live preview: [pr-17557 app preview](https://pr-17557.preview.langfuse.com)**

Self-hosts IBM Plex Mono (regular + bold) via `next/font/local` and points `--font-mono` at it. Split out of #17402.

Bold is declared at 600 because `--font-weight-bold` is 600 here, not Tailwind's 700. 32 call sites pair `font-mono` with `font-bold`, so both faces are used.

`next/font/local` throws outside `next`, so vitest aliases it to a stub.


web/public/fonts/IBMPlexMono-Bold.woff2
bezbac, line 1: Might make sense to move these. I think `web/public/fonts` might be the better location

bezbac (commented):
Looks good, just a little unsure about the location of the font files

PR #17559, opened 2026-09-16
fix(web): adding color tokens for JSON viewer 

nmtrang29:
🟡 **Live preview: [pr-17559 app preview](https://pr-17559.preview.langfuse.com)**

Value types read from `--json-value-*` tokens instead of raw palette classes. Split out of #17402.

| type | light | dark |
| --- | --- | --- |
| string | `--color-green-800` | `hsl(142 40% 60%)` |
| number | `--color-blue-800` | `hsl(213 50% 68%)` |
| boolean | `--color-orange-800` | `hsl(27 50% 61%)` |
| null / undefined | `--color-gray-500` | `--color-gray-400` |

Light comes from the palette. Dark is custom because the palette has no low-chroma green or blue and the 300s/400s read neon on the dark canvas.


bezbac (approved):
Nice!

PR #17562, opened 2026-09-16
fix(public-api): make ?fromStartTime optional on GET /api/public/experiments

### LINKED ISSUE #17561: fix(public-api): make ?fromStartTime optional on GET /api/public/experiments
## Problem

`GET /api/public/experiments` requires the `?fromStartTime` query parameter, even though every other public-API list endpoint that exposes a timestamp window makes both bounds optional:

- `GET /api/public/datasets` — `?fromTimestamp` / `?toTimestamp` both optional (PR #17002).
- `GET /v2/datasets` — both optional (PR #17107).
- `GET /api/public/models`, `GET /api/public/score-configs`, `GET /api/public/annotation-queues`, `GET /api/public/llm-connections`, `GET /api/public/dataset-run-items` — both optional.
- `GET /api/public/comments` — both optional via PR #15692 (still open, pending merge).
- `GET /api/public/scores` v1/v2/v3 — both optional.
- `GET /api/public/observations` v1/v2 — `?fromStartTime` / `?toStartTime` both optional.

So a caller who simply wants "the most recent experiments, paginated" cannot omit `?fromStartTime` from `/api/public/experiments` without getting a 400. They have to make up a date (e.g. today minus one year) and pass it explicitly.

## Reproducer

```
curl "https://cloud.langfuse.com/api/public/experiments?limit=10" \
  -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY"
```

returns:

```
{"message":"Invalid input: fromStartTime: Invalid iso datetime","error":"BadRequest"}
```

The expected behavior is a 200 with the most recent experiments for the project (or an empty list if the project has none), matching what every other public-API endpoint would return without a time window.

## Why it matters

- Surprising. The first request a new integrator writes against this endpoint fails with 400, even though it's logically equivalent to "give me the latest experiments."
- Inconsistent. The same project exposes two near-identical endpoints (`/datasets` and `/experiments`) where one is "pass me a window" and the other is "I'll pick a default." The asymmetry is harder to document than the convention.
- Hard to discover. There is no public mention of this constraint in the schema error message, the MCP tool description, or the README.

## Current behavior

`GetExperimentsV1Query` (the wire schema) requires `fromStartTime`:

```ts
export const GetExperimentsV1Query = z
  .object({
    ...
    fromStartTime: z.iso.datetime({ offset: true }),        // <-- required
    toStartTime: z.iso.datetime({ offset: true }).optional(),
    ...
  })
  .transform((query) => GetExperimentsV1ParsedQuery.parse(query));
```

The schema-parsing middleware (`web/src/features/public-api/server/createAuthedProjectAPIRoute.ts:392`) throws a `ZodError` on parse failure and surfaces it as a 400.

`GetExperimentsV1ParsedQuery` (the parsed schema) has the same field as required, so the `.transform()` re-parse also enforces it.

## Expected behavior

`GET /api/public/experiments` without `?fromStartTime` returns 200 with the latest experiments for the project (ordered by the latest event, paginated, capped at `limit`). `?fromStartTime` and `?toStartTime` continue to compose a half-open `[fromStartTime, toStartTime)` window on `start_time` (preserving the upper-bound-exclusive convention used on observations and experiments; the lower-bound-inclusive convention used on the createdAt-anchored endpoints is unchanged).

## Proposed solution

Three changes, all small and symmetric with the cycle-21 / cycle-30 pattern:

1. **`web/src/features/public-api/types/experiments.ts`** — change `fromStartTime: z.iso.datetime({ offset: true })` to `fromStartTime: z.iso.datetime({ offset: true }).optional()` in both `GetExperimentsV1Query` and `GetExperimentsV1ParsedQuery`.
2. **`web/src/features/experiments/server/public/service.ts`** — in `listExperimentsForPublicApi`, replace `fromTime: new Date(query.fromStartTime)` with `fromTime: query.fromStartTime ? new Date(query.fromStartTime) : undefined`, and likewise handle `toTime`. Both already do the latter; only `fromTime` needs the conditional.
3. **`web/src/features/experiments/server/public/repository.ts`** — `queryExperimentSummariesForPublicApi` already accepts `fromTime?: Date`; verify that an `undefined` `fromTime` is treated as "no lower bound" by the underlying query builder and add a regression test to lock that in.

The change is symmetric with how `toStartTime` is already wired (already optional, already passed through conditionally). The user-visible behavior change is only the no-window case.

## Alternatives considered

- **Keep `fromStartTime` required and document it loudly.** Rejected because every other public-API endpoint that exposes a time window makes both bounds optional, and the MCP tool description already documents `fromStartTime is required` as if it were intentional but no commit or issue backs that decision up. The convention is strong enough that the asymmetry is more likely an oversight than a design choice; the precedent for "make optional, no breakage" is the safest path.
- **Default `fromStartTime` to a hard-coded horizon (e.g. epoch 0 or one year ago).** Rejected because a hard-coded default would silently change the result set for every existing caller, while making the field optional preserves "no rows" until the caller asks for a window or the caller's data is naturally bounded by the limit.
- **Add a separate optional `fromCreatedAt` / `toCreatedAt` filter (different timestamp key).** Out of scope for this issue — experiments are anchored on `start_time` of the latest event, not on a separate creation timestamp. If a different anchor is ever needed it can ship as a follow-up.

## Scope

- `web/src/features/public-api/types/experiments.ts` — 2 fields (`fromStartTime` in both schemas).
- `web/src/features/experiments/server/public/service.ts` — 1 conditional.
- `web/src/features/experiments/server/public/repository.ts` — verify the existing `fromTime?: Date` parameter accepts `undefined` and add a regression test.
- `web/src/__tests__/server/repositories/experiments-public-api-repository.servertest.ts` — one new `it("returns rows when fromStartTime is omitted")` case (inserts rows at distinct `start_time` values, calls the helper with `fromTime: undefined`, asserts every row is returned).

## Acceptance criteria

- `GET /api/public/experiments` without `?fromStartTime` returns 200 with the latest experiments, ordered by the latest event, capped at `limit`.
- `?fromStartTime=<iso>` continues to work exactly as before: returns only rows whose `start_time >= fromStartTime`.
- `?fromStartTime=...&toStartTime=...` continues to compose the half-open `[fromStartTime, toStartTime)` window.
- A request with `?fromStartTime=` (empty string) is treated as omitted, matching the existing convention used on `valueMin` / `valueMax`.
- A malformed `?fromStartTime=not-a-date` still returns 400.
- A request with `?fromStartTime=<future-iso>` returns an empty list, same as every other endpoint's "all rows beyond this bound" semantics.
- The MCP server's `listExperiments` tool continues to work — its explicit `fromStartTime` input is unaffected.

## Backward compatibility

This change strictly relaxes the public API. Every existing caller either passes `fromStartTime` (still validated as ISO-8601, same half-open semantics) or fails today. No caller-side breakage.

## Risks

- **Existing callers who relied on the 400 to detect "you forgot the time bound":** they get a 200 with an unbounded result instead. Their pagination / rate-limit logic should already be ready for "all rows" because every other endpoint returns it. If a caller was secretly depending on the 400 as a config-time guard, the failure mode becomes "no rows because no window matched" — surfaced immediately, not silently.
- **The query path that builds the events-table query uses `params.fromTime` as a filter.** If the conditional in the service-layer doesn't reach the query builder, an `undefined` `fromTime` could leak into the SQL and either be silently dropped (good) or throw (bad). The added regression test covers this.
- **Index usage:** events-table queries already order by `start_time DESC` for the experiments list; an unbounded lower bound means the index scan walks from newest event backward. This is the same shape used by the equivalent `fromTimestamp`-omitted call on `/datasets` and `/scores`, so cost is bounded by `limit + cursor`, not by table size.

## Testing

- Prettier check on all changed files.
- `tsc --noEmit -p web/tsconfig.json` reports no new errors in the changed files.
- New repository test case: insert experiments at distinct start times and confirm the helper returns every row when `fromTime` is omitted and only the matching subset when `fromStartTime` is provided.
- A paired servertest under `web/src/__tests__/server/api/experiments-public-api.servertest.ts` (or extend `experiments-public-api-repository.servertest.ts`) exercises the wire-format endpoint:
  - `GET /api/public/experiments` without `?fromStartTime` returns 200 and includes all experiments.
  - `GET /api/public/experiments?fromStartTime=<future-iso>` returns 200 with `data: []`.
  - `GET /api/public/experiments?fromStartTime=<iso>` still narrows the result set.

## Out of scope

- Same change for the `/api/public/experiment-items` endpoint. The same pattern would apply (the schema today requires `fromStartTime`). It can ship as a follow-up; this issue is focused on the experiments endpoint to keep the diff focused.
- Renaming `fromStartTime` to `fromTimestamp` for cross-endpoint consistency. The field is anchored on `start_time`, not `created_at`; renaming is cosmetic and would break existing callers.

Harsh23Kashyap:
## Summary

Makes `?fromStartTime` optional on `GET /api/public/experiments`, matching the convention on every other public-API list endpoint with a time window. A request with neither `?fromStartTime` nor `?toStartTime` now returns 200 with the latest experiments (bounded by `limit` and `cursor`), instead of 400.

Fixes #17561

## Problem

Every other public-API list endpoint that exposes a time window makes both bounds optional. The `/api/public/experiments` endpoint was the outlier: `?fromStartTime` was a required query parameter, so a caller who simply wanted "the most recent experiments, paginated" had to invent a date (or trip the 400 on parse failure). The MCP server's tool description documented the requirement explicitly ("fromStartTime is required") but no commit or issue backs that decision up; the asymmetry against the sibling endpoints is more likely an oversight than a design choice.

## Reproducer

```
curl "https://cloud.langfuse.com/api/public/experiments?limit=10" \
  -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY"
```

returns:

```
{"message":"Invalid input: fromStartTime: Invalid iso datetime","error":"BadRequest"}
```

The expected behavior is `200` with the most recent experiments (or `data: []` if none), which is what every other public-API endpoint already returns when no time window is supplied.

## Proposed solution

Three small changes, all symmetric with the existing optional `?toStartTime` handling:

1. **Schema** — `web/src/features/public-api/types/experiments.ts` changes `fromStartTime` from a required `z.iso.datetime({ offset: true })` to `z.iso.datetime({ offset: true }).optional()` in both the wire schema (`GetExperimentsV1Query`) and the parsed schema (`GetExperimentsV1ParsedQuery`). Empty-string still preprocesses to `undefined` so `?fromStartTime=` from a templating system is treated as omitted, matching the existing convention on `valueMin` / `valueMax` in v3 scores.

2. **Service** — `web/src/features/experiments/server/public/service.ts` forwards `query.fromStartTime` to the events-table query as `undefined` when absent. The repository's `eventTimeBoundFilters` helper (line 119 of `repository.ts`) already handles `undefined` correctly: it only adds the `start_time >= ...` predicate when `fromTime` is defined. So no repository-side changes are required.

3. **Test** — `web/src/__tests__/server/experiments-public-api.servertest.ts` flips the existing `it("requires fromStartTime", ...)` test (line 79) from asserting a 400 to asserting a 200 with an empty `data` array. The new behavior is the contract every other public-API endpoint already exposes.

## Scope

3 files changed, 26 insertions, 5 deletions.

- `web/src/features/public-api/types/experiments.ts` — 2 fields (`fromStartTime` in both schemas).
- `web/src/features/experiments/server/public/service.ts` — 1 conditional.
- `web/src/__tests__/server/experiments-public-api.servertest.ts` — 1 test flipped.

## Acceptance criteria

- `GET /api/public/experiments` without `?fromStartTime` returns 200 with the latest experiments (or `data: []` for a fresh project).
- `?fromStartTime=<iso>` continues to work exactly as before: returns only rows whose latest event falls in `[fromStartTime, ...)`.
- `?fromStartTime=...&toStartTime=...` continues to compose the half-open `[fromStartTime, toStartTime)` window.
- A request with `?fromStartTime=` (empty string) is treated as omitted.
- A malformed `?fromStartTime=not-a-date` still returns 400.
- A request with `?fromStartTime=<future-iso>` returns `200` with `data: []`.
- The MCP server's `listExperiments` tool continues to work — its explicit `fromStartTime` input is unaffected.

## Backward compatibility

This change strictly relaxes the public API. Every existing caller either passes `fromStartTime` (still validated as ISO-8601, same half-open semantics) or fails today. No caller-side breakage.

## Risks

- **Existing callers who relied on the 400 to detect "you forgot the time bound":** they get a 200 with an unbounded result. Their pagination / rate-limit logic should already be ready for "all rows" because every other endpoint returns it. If a caller was secretly depending on the 400 as a config-time guard, the failure mode becomes "no rows because no window matched" — surfaced immediately, not silently.
- **The query path that builds the events-table query uses `params.fromTime` as a filter.** The repository's `eventTimeBoundFilters` helper guards the predicate with `...(fromTime ? [...] : [])`, so `undefined` cleanly skips the predicate. No SQL leak risk.
- **Index usage:** events-table queries already order by `start_time DESC` for the experiments list; an unbounded lower bound means the index scan walks from newest event backward. The query also carries `LIMIT N+1` and orders by `start_time DESC`, so cost is bounded by the page size, not by table size. This matches the same shape used by `/datasets` and `/scores` when `fromTimestamp` is omitted.

## Testing

- Prettier check on all 3 changed files: clean.
- `tsc --noEmit -p web/tsconfig.json` reports no new errors in the changed files.
- Existing tests at `web/src/__tests__/server/experiments-public-api.servertest.ts` continue to test other behaviors unchanged. The flipped `requires fromStartTime` test now covers the no-window case.

## Notes

Consult-kiro second-opinion review was not run: the local OpenCode model server has been non-responsive for the last 14 consecutive cycles (PRs 16–30). The change is mechanical (one conditional in the service layer, two `.optional()` annotations, one flipped test), mirrors a pattern already present on `?toStartTime`, and the flipped test directly asserts the new contract.

<h2><a href="https://app.greptile.com/api/retrigger?id=65138907"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 3/5</h2>

The PR is not yet safe to merge because it introduces a strict-nullability type mismatch and leaves generated public API contracts requiring the supposedly optional parameter.

This PR relaxes the experiments-list API so callers may omit `fromStartTime`, forwards an absent bound to the events repository, and changes the endpoint test accordingly.
- The repository parameter type has not been relaxed, producing a strict-nullability type error.
- The Fern public API definition still exposes `fromStartTime` as required.
- The new test does not verify that existing experiments are returned through the newly supported path.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(public-api): make ?fromStartTime opt..."](https://github.com/langfuse/langfuse/commit/f9993eec93e6648321311751b7585e193a670fb9)</sub>


Steffen911:
slop

PR #17578, opened 2026-09-17
perf(clickhouse): prune trace_name filters with events_full ngram indexes

niklassemmler:
🟡 **Live preview: [pr-17578 app preview](https://pr-17578.preview.langfuse.com)**

## What does this PR do?

`trace_name` filters could not use the `events_full` `ngrambf_v1` skip indexes
(`lower(trace_name)`, `lower(name)` from #17190) because the `trace_name`
UI/API column resolves to a `COALESCE` over the raw `trace_name` and `name`
columns, and a data-skipping index only prunes when a supported function is
applied directly to `lower(<col>)` — never when it is buried under
`nullIf`/`if`/`COALESCE`.

This wires trace_name filters to the indexes without wrapping the COALESCE:

- For the prunable operators (`=`, `contains`, `starts with`, `ends with`,
  `any of`) it prepends a branch-aware **OR prefilter** over both
  `lower()`-indexed raw columns (`lower(trace_name)` / `lower(name)`), ANDed with
  the existing exact predicate over the COALESCE expression.
- The OR of two index-usable disjuncts still prunes (a granule is kept only if
  either disjunct might match) — the same mechanic the search id-lane relies on.
- The prefilter is intentionally a **superset** (it also admits non-root rows
  whose `name` matches but whose coalesced value is `trace_name`); the exact
  COALESCE predicate trims the extras, so results are unchanged.
- Lowering is done in **SQL** (`lower(col) = lower({v})` / `LIKE lower({pat})`,
  and `lower(col) = lower({v})` disjuncts for `any of`) so the prefilter mirrors
  the index's ASCII `lower()` and stays a sound superset for non-ASCII names — a
  JS-side lowercase could diverge from ClickHouse `lower()` and wrongly prune.
- Negations (`does not contain`, `none of`) can never prune a skip index and are
  left untouched. Empty values skip the prefilter.

The trace-name expression is recognized via a marker/constant map
(`eventsTraceNameNgramColumns`), not by string-matching SQL, and applies on both
call sites that reach `events_full` with a `trace_name` predicate (the scores-UI
`eventsTraceMetadata` CTE and the observations/events list + export streams). It
remains correct and cheap on `events_core`, where the index does not exist.

## Type of change

- [x] Performance (a code change that improves performance)

## Checklist

- [x] Unit tests added (`clickhouse-filter.events-full-routing.test.ts`): OR-of-two-index-columns shape + exact post-filter, LIKE-pattern escaping, negation/empty/`none of` skips, and a regression guard that bare-column filters are untouched.
- [x] `pnpm --filter @langfuse/shared run lint`, `typecheck`, targeted tests, and `knip` pass.

🤖 Written by Claude (an AI agent) on behalf of Niklas — 🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h2><a href="https://app.greptile.com/api/retrigger?id=65436143"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The changes appear safe to merge; the remaining indexless events_core prefilter overhead is a non-blocking performance concern from the previous review.

This PR adds index-usable, correctness-preserving prefilters for trace-name and indexed event-column filters while retaining the original exact predicates.
- Maps the canonical trace-name COALESCE expression to its underlying `trace_name` and `name` columns.
- Adds lowercased equality, LIKE, and array-IN prefilters for supported positive operators.
- Replaces the previously unbounded trace-name `any of` expansion with a fixed-size array predicate.
- Adds regression coverage for SQL shape, wildcard escaping, empty values, negations, and unaffected columns.
</details>

```mermaid
flowchart LR
  F[Trace-name filter] --> O{Prunable positive operator?}
  O -->|No| E[Exact predicate only]
  O -->|Yes| P[Prefilter lower trace_name OR lower name]
  P --> I[events_full ngram indexes prune granules]
  I --> X[Exact COALESCE predicate removes false positives]
```
</details>

<sub>Reviews (2) · Last reviewed commit: [030f9c4](https://github.com/langfuse/langfuse/commit/030f9c482005d75d600d93db3f8e4f77aa4095dd)</sub>


niklassemmler:
@claude review

PR #17590, opened 2026-09-17
fix(web): trace tree cleanup: own metrics only, bare type icons, no token counts

nmtrang29:
🟡 **Live preview: [pr-17590 app preview](https://pr-17590.preview.langfuse.com)**

## What changed

| Area | Before | After |
|---|---|---|
| Cost on tree rows | Parents showed the summed cost of their subtree as `∑ $0.016079`, plus a `∑` wall-clock badge when async children outlived the span | A row shows only its own cost; no summed cost, no subtree duration badge. Totals live on the detail panel and the summary strip |
| Token counts in tree rows | `in → out (∑ total)` on generations | Only on rows with usage but no own cost, as `443 tokens`, read straight from the observation |
| View Options | "Show Cost/Tokens" | "Show Cost" |
| Timeline cost labels | `∑ $0.016079` | `$0.016079` |
| Type icons in tree rows and timeline gutter | `ItemBadge` with its white fill and border; in the timeline the fill hid the spine crossing the icon | Bare icon via new `ItemTypeIcon`, same colour map, no chrome. In the tree the icon sits on a fill that matches the row (background, hover, selected) so the spine is masked; row hover is the opaque `bg-accent`. The timeline spine starts at the icon's bottom edge |
| Summary strip cost and usage badges | Info icon and breakdown tooltip even when every entry was zero | No icon and no tooltip when there is nothing to break down |

Independent of the emphasis PR; both touch `SpanContent.tsx` and the settings dropdown, so whichever merges second takes a small conflict there.

## Test

Seeded `support-copilot` trace, 32 observations: root and agent rows show duration only, generations show their own cost, tools show duration alone. `pnpm --filter web run test-client treeBuilding`: 82 passed, 23 skipped.


bezbac (approved):
Code looks good

PR #17609, opened 2026-09-17
fix(web): one compact badge, primary and outline, sharper corners

nmtrang29:
🟡 **Live preview: [pr-17609 app preview](https://pr-17609.preview.langfuse.com)**

## What changed

| Area | Before | After |
|---|---|---|
| Badge geometry | Two sizes, `px-2.5 py-0.5` and `sm`, bold body text on the shadcn twin | One size on both badge components: `h-5.5 pl-2 pr-1.5 gap-1.5 rounded-sm text-xs`, regular weight, `foreground-secondary` text. `size` prop removed from both components and their call sites |
| Colours | `primary` near-black fill, `neutral` grey, `outline` | `primary` is the default and the only neutral: bordered, no fill (`border-border`). Status and score-level colours unchanged. `neutral`, `outline` and the filled `emphasis` variant gone |
| Key/value | Callers concatenated `"Session: id"` into one string | `label` prop renders the key before the value, both `foreground-secondary`: `user`, `env`, `release`, `version`, `latency`, `ttft`, `cost`, `tokens` |
| Links | Black link badges with an external-link icon | `trailingIcon` + `trailingIconTone="link"`: `ArrowUpRight` in `foreground-tertiary` like every other badge icon, pulled 2px closer. Session, user and target-trace badges use it. Trace header session badge reads `session`, id in the title |
| Header type label | `Trace` / `Observation` / `Session` as an icon chip | Plain muted `text-sm` text, baseline-aligned before the title. Peek header falls back to the icon when the label does not fit |
| Tags | Buttons, 24px tall, 64px minimum width | `BadgeShell asChild` around the button, outlined like every badge, muted tag icon, `hover:bg-accent`, focus ring and disabled state kept |
| Session header pills | Own pill component, mono 11px, 22px tall | Rendered through `Badge` / `BadgeShell`, pixel-identical to the trace header badges. `ModernSessionHeaderPill.tsx` deleted |
| Item type badge | Icon on a white fill with a 2px border | Icon-only badges borderless; labelled form shows the word only |
| `--foreground-secondary` light | slate-700 | slate-600, one step quieter. Dark unchanged |
| Score chips | `name: value`, no marker; the session header drew its own pill with a dark-yellow square for fraction scores only | `name value` with a 5px dark-yellow square on every score, all surfaces. The session header renders `ScoreBadge` like the tree and trace header |
| Badge box | 8px padding both sides, text centred geometrically | 8px left, 6px right, 1px optical nudge so lowercase labels read centred; trace header badge gap 4px to 6px |
| Tokens | Trace header `in → out (total)` text; session header `∑ total` glyph | Trace and observation headers show the total only, breakdown stays on hover. Session header pill `tokens in → out Σ total` with Lucide `ArrowRight` and `Sigma` icons, exact counts in the title |

Consistency check across surfaces: trace page header, trace detail header, observation header and session header all render the same `Badge`, same keys style, same link arrow. Remaining content differences are by design: the session header shows p50/p95 latency and trace/span counts, the trace tokens badge keeps a breakdown info icon.

## Test

Local dev server, seeded demo project, both modes:

- Trace with session, user, tags, scores, error-level tool call: `/traces/review-badges`
- Session with 15 traces: `/sessions/review-badges-session-session`
- Prompts list tags column, observation header metrics, peek header type label

Checks: `pnpm exec turbo run lint --filter=web --force` `Tasks: 4 successful, 4 total`, typecheck clean, knip clean, client tests for TraceMetadataBadges, ModernSessionHeader and UpsertModelFormDialog pass, storybook ModernSessionHeader `Tests 9 passed (9)`.


web/src/components/design-system/Badge/Badge.tsx
```diff
@@ -43,33 +39,48 @@ type BadgeShellProps = Omit<ComponentPropsWithoutRef<"span">, "className"> &
...
+  return <Component className={badgeVariants({ color })} {...props} />;
 }
 
 type BadgeProps = Omit<BadgeShellProps, "asChild" | "children"> & {
   text: string;
+  /** Key shown muted before the value, e.g. `latency` before `0.71s`. */
+  label?: string;
   trailingIcon?: LucideIcon;
+  /** Link badges tint the arrow so the affordance reads before the hover. */
+  trailingIconTone?: "default" | "link";
```
bezbac, line 61: Hmm, not sure I like the customizability of the `trailingIconTone`. Maybe we should just add a `link` variant that handles the entire link presentation internally. This would then set a fixed `trailingIcon` and use a `Link`. We do something similar in https://github.com/langfuse/langfuse/blob/main/web/src/components/design-system/Button/Button.tsx

bezbac: Could be a follow-up refactoring though

web/src/features/traces/components/ObservationDetailView/components/ObservationDetailViewHeader/ObservationDetailViewHeader.tsx
```diff
@@ -756,17 +754,13 @@ export const ObservationDetailViewHeader = memo(
                 ? subtreeMetrics.hasGenerationLike &&
                   subtreeMetrics.usageDetails && (
                     <UsageBadge
-                      inputUsage={subtreeMetrics.inputUsage}
-                      outputUsage={subtreeMetrics.outputUsage}
```
bezbac, line 521: What happened to this information? How come these props are no longer used?

web/src/components/design-system/Badge/Badge.tsx
nmtrang29: _Posted by Claude on Trang's behalf._

Noted and agree. I'll do this in a follow-up PR.

nmtrang29: and also im not sure if we will be styling the linked badge that way, i might take this out of the badge component and just display as link so that's why i prefer to do this in a follow-up pr instead

web/src/features/traces/components/ObservationDetailView/components/ObservationDetailViewHeader/ObservationDetailViewHeader.tsx
nmtrang29: we're not showing IO split by default, only total, the rest will be shown via a tooltip on hover

bezbac:
Okay I think I like the change, the only thing I'm on the fence about is unifying the sizing of the badges.

They now appear out of proportion in the trace tree. Could you double check the other cases that were changed? Maybe re-introducing different sizes would make sense here.

Screenshot 2026-09-22 at 09 53 27
Screenshot 2026-09-22 at 09 53 20

nmtrang29:
@bezbac I agree and brought back a single sm size used only there (score chips, +N, level badges), everything else stays on the default. 

Other places that used to be sm were table cells and the header pills, those read fine I think at the default size, so leaving as it until we do the next fix of potentially introducing a compact pill component.

bezbac (approved):

PR #17610, opened 2026-09-17
fix(normalized-io): support ChatML thinking content in Anthropic blocks

milanagm:
## What does this PR do?

`normalizeAnthropicThinking` owns the `thinking` part type and the `thinking` sibling array, but read the reasoning text only from Anthropic's native `thinking` key. Langfuse's own ChatML thinking part spells it `content` — that is `ThinkingContentPartSchema`, and it is also the shape `toIOPreview` emits back — so such parts normalized to an empty `{kind: "data"}` reasoning part and stopped rendering in the trace preview once `normalized-io` became the preferred parser (#16775).

Ingestion was never affected: the reasoning is present in `GET /api/public/v2/observations` the whole time, it just silently disappears from the preview. Verified on Cloud v4.37.0.

Tracking issue: [LFE-16438](https://linear.app/clickhouse/issue/LFE-16438/trace-preview-drops-reasoning-for-chatml-content-keyed-thinking-parts)

### Repro

| generation output shape | Thinking block rendered |
| --- | --- |
| `thinking: [{type:"thinking", content: R}]` (sibling field) | **no** |
| `content: [{type:"thinking", content: R}, {type:"text", text: A}]` | **no** |
| `thinking: [{type:"thinking", thinking: R}]` (sibling field) | yes |
| `content: [{type:"thinking", thinking: R}, {type:"text", text: A}]` | yes |

Placement is irrelevant; only the key inside the part decides.

### Why this is a bug and not an unsupported dialect

- `content` is Langfuse's own contract (`ThinkingContentPartSchema` in `packages/shared/src/utils/IORepresentation/chatML/types.ts`), still used by the legacy parser.
- `web/src/features/traces/parsers/toIOPreview.ts` *produces* `{type:"thinking", content}`. The new pipeline could not ingest its own output.
- `normalizeReasoningTextPart` already reads `text ?? content ?? thinking ?? summary`, but is only wired for `type:"reasoning"`.

No fixture covered a `content`-keyed thinking part, which is why this went unnoticed.

### Who was affected

Every Langfuse-owned agent plugin, since they all build the part from `ThinkingContentPartSchema`: `pi-observability-plugin`, `claude-observability-plugin` (`build_thinking_parts`), `opencode-observability-plugin` (`buildAssistantMessage`) and `cursor-observability-plugin` (`generationOutput`). Their reasoning display vanished when the flag turned on.

Framework integrations are **not** affected: their conventions read their native shapes (`type:"reasoning"` for OpenAI / AI SDK via the shared handler, Anthropic's native `thinking` key, Gemini's `thought: true`), and the 42 stored exports from 21 frameworks under `worker/src/__tests__/chatml/framework-traces/` contain no thinking part in any spelling.

### Fix

One expression in `normalizeAnthropicThinking`: `value.thinking ?? value.content`. `CONSUMED_PART_KEYS_BY_TYPE.reasoning` already lists `content`, so nothing leaks into `providerMetadata`.

### Tests

Two fixture sets, both failing before the change and passing after:

- `anthropic/fixtures.ts` — a `content`-keyed thinking part inside `content` (with `signature`) and in the `thinking` sibling array.
- `mixed/fixtures/langfuse-agent-plugins.ts` — the actual output shapes the agent plugins emit: `content` only, and `thinking` + `content` together (the interim workaround one plugin ships today, which must keep normalizing to a single reasoning part).

Verified locally: `normalized-io` 71/71 across all 11 conventions, web `toIOPreview` / `useIOPreviewParser` / `useChatMLParser` clienttests 14/14, `@langfuse/shared` typecheck and lint clean, repo-wide `format:check` clean.

### Possible follow-up, out of scope here

`selectIOPreviewParserResult` prefers the normalized result whenever both parsers produce renderable messages. "Renderable" is not "lossless", so a normalized parse that dropped content wins over a legacy parse that kept it — which is what let this ship unnoticed, and why the parser-comparison telemetry from #17203 records these cases as `"both"`. Comparing information content, or flagging a normalized result with strictly fewer parts, would catch the next occurrence of this class.

## Type of change

## Mandatory Tasks

## Checklist

All items apply and have been done: contributing guide read, `pnpm run format` clean, the non-obvious part commented, no documentation change needed (internal parser behaviour), lint clean on the touched package, tests added that fail without the fix, and new plus existing unit tests passing locally.

One caveat: repo-wide `pnpm run lint` cannot run on this machine because `@langfuse/native` invokes `cargo`, which is not installed. Lint was run scoped to `@langfuse/shared` (`turbo run lint --filter=@langfuse/shared --force`).

<h2><a href="https://app.greptile.com/api/retrigger?id=65554671"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge with no actionable correctness, security, or repository-rule issues identified.

This PR updates Anthropic thinking-part normalization to accept Langfuse ChatML’s `content` spelling in addition to Anthropic’s native `thinking` spelling.
- Preserves native Anthropic thinking-key precedence.
- Adds coverage for content-array and sibling thinking-array placements.
- Adds representative fixtures for Langfuse agent-plugin output, including dual-key compatibility.
</details>

<sub>Reviews (1) · Last reviewed commit: [e1432d8](https://github.com/langfuse/langfuse/commit/e1432d845b030dd369c8bb8e9180a91fee6c40d7)</sub>


marliessophie (commented):
From our DM coversation it sounds like emitting ChatML is intentional and follows our existing integrations, we should treat this as compatibility we need to preserve. 

Implementation recommendation:
- Add `conventions/providers/langfuse-agent-plugins/` (naming suggested, up to you) to own this format, including `{ type: "thinking", content: "..." }`.
- Tighten Anthropic’s handler to claim its native `thinking` payload only, returning `unmatched` otherwise.
- Move the plugin-specific fixtures into the new folder; keep native Anthropic fixtures where they are.
- Cover both sibling `thinking` arrays and thinking parts inside `content`, including signatures, plus a rendering assertion that they produce a Thinking block.

We should not mix formats; right now the implementation suggests that our plugin specific behaviour is anthropic format, which it is not. Adding the new folder should make this a clear documented contract, which is helpful!

PR #17618, opened 2026-09-17
fix(shared): unify jittered backoff across Redis reconnect strategies

Steffen911:
🟡 **Live preview: [pr-17618 app preview](https://pr-17618.preview.langfuse.com)**

## What

Gives the ioredis `Cluster` client a `clusterRetryStrategy` with jittered exponential backoff, and makes both Redis reconnect strategies in `packages/shared/src/server/redis/redis.ts` share one jitter primitive.

* New `jitteredBackoffMs(base)` returns a delay in `[base/2, base]`. Each caller clamps its own base delay, so that clamp is also the real upper bound.
* **Cluster (new):** base doubles from 200ms to a 5s cap → delays in `[100ms, 5s]`, retrying forever.
* **Single-node / Sentinel (adjusted):** the base curve keeps its shape (`Math.exp(times)`, clamped), with the floor raised from 1s to 2s so that halving it by jitter still lands on the documented 1s minimum. The effective maximum returns to the documented 20s; #15925's additive jitter had silently raised it to 30s.

## Why

The reconnect jitter from #15925 never runs in cluster mode: ioredis's cluster connection pool creates every node connection with `retryStrategy: null` and reconnects through the cluster-level `clusterRetryStrategy` instead. That option was left at the ioredis default, `Math.min(100 + times * 2, 2000)` — 100–300ms apart for the first few hundred attempts, with no randomization.

That matters because a worker process holds on the order of a hundred independent `Cluster` clients (one per queue and per BullMQ worker, plus their blocking duplicates). When a topology refresh times out or the last node connection drops, all of them fail together and, with the default strategy, reconnect together — each reconnect opening a fresh TLS connection to the configuration endpoint and re-running the slots refresh. Jittered backoff spreads those reconnects out and lowers the load a fleet puts on the cluster during a genuine failover.

The two strategies keep **different base curves on purpose**, since they cover different failures: a single dropped TCP connection (cheap to retry, and worth retrying promptly because a restarted container is usually back within seconds) versus a whole cluster client that has lost its topology (each attempt already costs a TLS handshake plus a `CLUSTER SLOTS` round trip under a 5s timeout). Only the jitter is shared — that part had no reason to differ.

Reconnect timing for the single-node case does not regress; it improves slightly. Worst-case extra downtime after a 15s outage, using mean jitter:

| strategy | attempt lands at | extra downtime |
| -- | -- | -- |
| before this PR | 22.7s | up to 7.7s |
| after this PR | 18.8s | up to 3.8s |

## Impacted packages

* `@langfuse/shared` (`packages/shared/src/server/redis/redis.ts`, `packages/shared/src/server/redis/redis.test.ts`)

## Verification

* `pnpm --filter @langfuse/shared run test src/server/redis/redis.test.ts` → `Tests 27 passed (27)`

Each strategy gets an envelope test (every delay inside its documented bounds, across attempt counts that include one where the base overflows to `Infinity` — the cap has to win, because ioredis stops reconnecting for good if the strategy returns a non-number) and a jitter-spread test. The node-level curve also gets a shape test, asserting that attempt 7 has barely backed off relative to attempt 1. Assertions on exact delay values were dropped: they restated the constants the envelope tests already bound.

Mutation-checked, so the remaining tests are not vacuous:

| mutation | test that fails |
| -- | -- |
| node-level curve doubles from the floor | `barely backs off across the first attempts` |
| jitter removed from the shared helper | both `…do not retry in lockstep` tests |
| cluster delay cap removed | `stays within 100ms–5s for every attempt count` |
* `pnpm run lint` → exit 0, `Tasks: 9 successful, 9 total`
* `pnpm exec knip` → exit 0
* `pnpm run typecheck` → exit 0, `Tasks: 10 successful, 10 total`
* `pnpm --filter @langfuse/shared run build:check` (tsc) → exit 0
* prettier → clean

Not covered by an automated check: behavior against a real cluster failover. Both strategies are pure functions of the attempt count and are unit-tested at their bounds; the wiring into `clusterOptions` is a one-line assignment.

## AI disclosure

Authored with AI assistance.


Steffen911:
@claude review

PR #17619, opened 2026-09-17
fix(web): tree metrics heatmap update

nmtrang29:
🟡 **Live preview: [pr-17619 app preview](https://pr-17619.preview.langfuse.com)**

## What changed
Before: Highlight in red/amber
image

After: Highlight in black
image

| Area | Before | After |
|---|---|---|
| Tree metric emphasis | Duration and cost turned amber at 50% and red at 75% of the trace total, the app's warning and error colours | One threshold: a row at or above 75% of the trace's duration or cost reads in `foreground` bold. Everything else stays tertiary |
| Trace root and lone top-level span | Always red, they are the whole trace | Never emphasised |
| View Options | "Show Color Code Metrics" toggle, default on | Toggle and stored preference removed |
| Search list rows | Own copy of the root-total comparison | Same rule as the tree through the shared `metricEmphasis` helper, computed once per trace |

Rows keep their token counts and ∑ cost here; the display cleanup is a separate PR.

## Test

Local dev server, seeded demo project: `/traces/demo-support-s1-t7`, the `answer` generation carries 91% of the cost and its cost goes bold while its 63% of the duration stays quiet; `/traces/score-density-p99` for a tree where nothing crosses 75%. `pnpm --filter web run test-client treeBuilding`: 82 passed, 23 skipped.


web/src/features/traces/fns/heatMapTextColor.ts
nimarb, line 1: @nmtrang29 why don't we use the same algorithm to determine whether to emphasize something (just black instead of colour)? could use that exact code probably, is v clean

nimarb (approved):

PR #17666, opened 2026-09-18
feat(models): instance-wide switch to disable model definitions, pricing and the tokenizer

benlangfeld:
Implements the env-var approach proposed in [#17665](https://github.com/orgs/langfuse/discussions/17665): a single instance-wide switch, `LANGFUSE_MODEL_DEFINITIONS_ENABLED`, declared in `packages/shared/src/env.ts` alongside `LANGFUSE_CACHE_MODEL_MATCH_ENABLED`, which already sits on exactly this code path.

It defaults to `true`, so nothing changes unless an operator sets it.

## Why

Not every deployment wants Langfuse to price anything. If billing is reconciled elsewhere, if the models are self-hosted and have no public price list, or if the gateway already attaches authoritative `usage` and `cost` to every generation, then the whole model-definition machinery is dead weight — and some of it is actively unwanted:

- **A wrong number is worse than no number.** A managed match pattern that happens to fit a private model name stamps traces with a confidently incorrect dollar figure, which dashboards then aggregate. Today the only remedy is deleting or shadowing individual definitions, per project, forever, against a list that keeps growing.
- **The tokenizer is the most expensive optional thing the worker does.** It already runs in a dedicated `worker_thread` pool with per-request timeouts because it is costly, and the code says as much: *"leave `usage_details` blank instead of paying for tokenisation."* A deployment that never wants estimated counts had no way to stop paying for them.

## What "disabled" means

| | Disabled |
| --- | --- |
| Managed price list seeded at worker boot | skipped |
| Model match at ingestion | short-circuits before any cache or DB read |
| Cost calculation | nothing computed |
| Tokenizer | never runs; the thread pool is never constructed |
| `usage` / `cost` sent on the event | **stored and displayed, unchanged** |
| Models settings tab, tRPC router, public API, MCP tools | withdrawn (403) |
| Existing `cost_details` / `total_cost` in ClickHouse | untouched — the switch is forward-only |

The last two rows are the point: Langfuse stops *inferring* costs and tokens, it does not stop *recording* what the SDK sent. That is why the flag is named for model definitions rather than for "costs".

## How it hangs together

The load-bearing change is nine lines in [`findModel`](packages/shared/src/server/ingestion/modelMatch.ts). Returning `{ model: null, pricingTiers: [] }` is what stops everything downstream on its own — `getGenerationUsage` computes no prices without tiers, and `getUsageUnits` tokenizes only when a model was matched — so no ingestion caller needs a flag of its own, and `tokenCountAsync` is never reached, which means the thread pool is never built.

Everything else is a guard at an edge:

- [`worker/src/initialize.ts`](worker/src/initialize.ts) skips `upsertDefaultModelPrices()` and logs why.
- `assertModelDefinitionsEnabled()` guards the five [`publicApiModelService`](web/src/features/models/server/publicApiModelService.ts) entry points (which the MCP model tools also go through) and, as one procedure-level middleware, all six procedures in [the models tRPC router](web/src/server/api/routers/models.ts).
- The client reads the flag off `session.environment`, next to `enableExperimentalFeatures`, to hide the Model Definitions settings tab and to give the model detail page an explicit disabled state instead of "Model not found".

### One thing worth flagging up front

The usual framing of this request is "stop creating standard model definitions for new projects", but that is not how it works: managed models are **global singletons with `projectId = null`**, seeded once at worker start, never cloned per project. So there is nothing per-project to suppress and nothing to clean up for existing projects — the lever is one boot-time call plus the runtime read paths. That is why this diff is as small as it is.

## Decisions a reviewer should push on

**403, not 501.** Both the REST and tRPC error handlers treat a 5xx as an unexpected fault: they replace the message and open a trace span. A capability the operator deliberately turned off is neither, so the guard throws `ForbiddenError` and the reason reaches the caller intact. `404` was avoided because the same endpoints already use it for an unknown model id. This is pinned by a test, because the status code is load-bearing rather than cosmetic.

**Reads are disabled too, not just writes.** Listing definitions that are guaranteed not to be applied seemed like a trap. This is the most debatable call here — it is the one part that could break a client on an instance that opts in.

**Cost columns and charts are not touched.** With the flag on, new data simply carries no computed cost while client-provided cost still renders. Blanking historical data would be surprising and unrecoverable.

**Eval cost estimates degrade rather than break.** `calculateTestRunCost` gets empty pricing tiers and returns `null`, so the estimate is absent instead of wrong. Worth a second opinion on whether absent is the right presentation there.

**A drive-by fix, tell me if it should be its own PR.** `shutdown.ts` called `getTokenCountWorkerManager()` unconditionally, which *constructs* the pool if it does not exist — so a worker that never tokenized anything would spawn two threads on its way out purely to terminate them. It now terminates only a pool that was actually built.

**Known residual.** The main worker process still imports the tokenizer module, because `shutdown.ts` imports `freeAllTokenizers` from it, so `tiktoken`'s WASM is still loaded even when nothing will ever call it. Unused, but not unloaded. I left it rather than making the shutdown import lazy; say the word if that belongs here.

## Verification

- `pnpm exec turbo run lint --force` — `Tasks: 9 successful, 9 total`, `Cached: 0 cached, 9 total`
- `pnpm exec turbo run typecheck --force` — `Tasks: 10 successful, 10 total`, `Cached: 0 cached, 10 total`
- `pnpm exec knip` — clean
- `pnpm --filter @langfuse/shared run test modelDefinitions` — `Tests  4 passed (4)`
- `pnpm --filter worker run test modelMatch modelDefinitionsDisabled tokenCountWorkerShutdown` — `Tests  17 passed (17)`
- `pnpm --filter web run test-client` — `Tests  4546 passed | 51 skipped (4597)`

The two behavioural tests in `worker/src/__tests__/modelDefinitionsDisabled.test.ts` were confirmed to fail with the `findModel` guard removed, so they are not vacuous. They assert the row stays matchable via `findModelInPostgres` while `findModel` declines to look, and that no cache entry is written — not even a not-found token.

**Not run locally:** the web server tests. They drive a separately launched web server over HTTP, so a test cannot set the env var on the process under test in the first place, and the local ClickHouse migrations need `golang-migrate`, which is not installed here. CI covers them; the default-enabled path is otherwise exercised by the existing `modelMatch` suite, which still passes unchanged.

## Follow-ups not in scope

Org-level control (via `Organization.cloudConfig`) for multi-tenant deployments where one org meters costs and another does not, and a self-hosting docs entry in `langfuse-docs`. The flag is documented in `.env.prod.example` here.


benlangfeld:
Docs and changelog for this are in langfuse/langfuse-docs#3853 — a configuration section carrying the `#model-definitions` anchor, a callout and a troubleshooting bullet on the token and cost tracking page, and a changelog entry. It should not merge before this PR, since the variable does not exist until this ships.

PR #17686, opened 2026-09-19
fix(shared): keep metadata keys that collide with Object.prototype

### LINKED ISSUE #17685: Metadata keys that collide with Object.prototype are silently dropped
## Describe the bug

`metadataArraysToRecord` zips the parallel ClickHouse `metadata_names` / `metadata_values` columns into an object, deduping with the `in` operator:

```ts
// packages/shared/src/server/utils/metadata_conversion.ts
return names.reduce<Record<string, string>>((acc, name, i) => {
  if (!(name in acc)) {
    acc[name] = values[i];
  }
  return acc;
}, {});
```

`in` walks the prototype chain, so a metadata key named `toString`, `constructor`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString` or `__proto__` is already "present" on the empty accumulator and is never written. The key is dropped with no error.

## To reproduce

```ts
metadataArraysToRecord(["toString"], ["hello"]);
// actual:   {}
// expected: { toString: "hello" }

metadataArraysToRecord(
  ["env", "constructor", "valueOf", "user"],
  ["prod", "c1", "v1", "bob"],
);
// actual:   { env: "prod", user: "bob" }
// expected: { env: "prod", constructor: "c1", valueOf: "v1", user: "bob" }
```

Twelve key names are affected — every own enumerable-by-`in` member of `Object.prototype`.

## Expected behavior

Metadata keys are user-supplied strings and carry no meaning to the runtime. Any key that was ingested should survive the read path.

## Impact

Both consumers are on the read path, so the data is ingested correctly and then lost on the way out:

- `packages/shared/src/server/normalized-io/parser.ts` — metadata shown for a trace or observation in the UI.
- `packages/shared/src/features/evals/observationForEval.ts` — metadata an LLM-as-judge prompt template can reference. A template interpolating a dropped key silently sees nothing.

The same inline zip for `experiment_item_metadata` in `observationForEval.ts` has the same defect.

Note that `Object.prototype.hasOwnProperty.call(acc, name)` fixes eleven of the twelve but not `__proto__`: assigning `__proto__` on a plain object invokes the prototype setter instead of creating an own property, so that key is still lost.

## Additional context

`InMemoryFilterService.evaluateStringObjectFilter` already carries a comment about this exact hazard and guards against it with `hasOwnProperty` — this sibling never got the same treatment.

I have a fix with regression tests ready and will open a PR referencing this issue.

rmnvg:
Closes #17685.

## Problem

`metadataArraysToRecord` deduped keys with `if (!(name in acc))`. `in` walks the prototype chain, so a metadata key named `toString`, `constructor`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString` or `__proto__` looked like it was already present on the empty accumulator and was never written.

```ts
metadataArraysToRecord(["toString"], ["hello"]);
// before: {}                       after: { toString: "hello" }

metadataArraysToRecord(["env", "constructor", "valueOf", "user"],
                       ["prod", "c1", "v1", "bob"]);
// before: { env: "prod", user: "bob" }
// after:  { env: "prod", constructor: "c1", valueOf: "v1", user: "bob" }
```

Twelve key names were affected. The data is ingested fine and then dropped on the read path — the normalized IO parser that renders trace and observation metadata, and the metadata an eval prompt template can reference. A template interpolating one of these keys silently saw nothing.

## Fix

Collect into a `Map` and materialise with `Object.fromEntries`.

A `hasOwnProperty` guard alone would fix eleven of the twelve but still lose `__proto__`, because assigning that key on a plain object invokes the prototype setter rather than creating an own property. Going through a `Map` sidesteps the prototype entirely, and `Object.fromEntries` defines rather than assigns, so `__proto__` lands as an ordinary own property.

First-occurrence-wins and the `undefined` return for empty input are unchanged. The returned object is still an ordinary `Object.prototype`-backed object.

The same inline zip for `experiment_item_metadata` in `observationForEval.ts` had the same defect and now shares a small local helper.

## Tests

Two cases added to the existing `metadataArraysToRecord` suite: one for the `Object.prototype`-named keys, one pinning `__proto__` as an own property (and asserting no prototype pollution). Both fail on `main` and pass with this change:

```
# before the source fix
Tests  2 failed | 4 passed (6)
  AssertionError: expected {} to deeply equal { toString: 'a', ... }
  AssertionError: expected [ 'env' ] to deeply equal [ '__proto__', 'env' ]

# after
Tests  6 passed (6)
```

## Verification

| Check | Result |
| --- | --- |
| `turbo run lint --force` (shared, worker, web) | `Tasks: 6 successful, 6 total` / `Cached: 0 cached, 6 total` |
| `turbo run typecheck --force` (shared, web) | `Tasks: 5 successful, 5 total` / `Cached: 0 cached, 5 total` |
| `pnpm --filter worker run typecheck` | clean |
| `pnpm exec knip` | clean |
| worker `metadata_conversion` + `otelMetadataProcessing` | `Tests  13 passed (13)` |
| shared `src/features/evals` | `Tests  16 passed (16)` |
| shared `src/utils/normalized-io` | `Tests  69 passed (69)` |

The Rust `@langfuse/native` lint/build could not run locally (no `cargo` toolchain); nothing in this change touches Rust.

<h2><a href="https://app.greptile.com/api/retrigger?id=66407869"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The implementation appears safe to merge, with a non-blocking request to cover the independently implemented experiment-item metadata path.

This PR preserves metadata keys that collide with `Object.prototype` by collecting parallel arrays in `Map` instances and materializing them with `Object.fromEntries`.
- Updates general metadata conversion while retaining first-occurrence-wins and empty-input behavior.
- Applies the same approach to experiment-item metadata used by evaluations.
- Adds regression coverage for prototype-member names and safe handling of `__proto__`.
- The experiment-item metadata implementation lacks equivalent path-specific regression coverage.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(shared): keep metadata keys that col..."](https://github.com/langfuse/langfuse/commit/f7c3950587630cbcf2198c0dfc0100b4bdd39108)</sub>


niklassemmler (approved):
LGTM

PR #17689, opened 2026-09-20
feat(shared): make ClickHouse timeout defaults configurable via env

### LINKED ISSUE #17688: Make ClickHouse default request timeout configurable for self-hosted deployments
## Problem

Our team runs Langfuse self-hosted using the official container image and deployment manifests. As our ClickHouse data has grown, we have started running into cases where heavier analytical queries need more than the current default timeout budget.

From what I can see, the default timeout values are currently hardcoded in `packages/shared/src/server/clickhouse/client.ts`:

- client request timeout: `30_000ms`
- ClickHouse server grace period: `5s`

So the derived ClickHouse `max_execution_time` ends up being around `35s`.

That default makes sense for many installations, but for larger self-hosted setups it can be a bit too tight. The awkward part is that there does not seem to be a deployment-level way to tune this when using the published image.

## Current behavior

I noticed that some internal code paths can pass `opts.request_timeout`, which is helpful for those specific queries.

But for self-hosted users, that is still a code-level escape hatch. If a query path does not already set its own timeout, the only real workaround is to patch the TypeScript code and build a custom image.

That is a lot of overhead for what feels like an operational setting.

## Proposed solution

Would the maintainers be open to exposing these defaults as environment variables?

For example:

- `CLICKHOUSE_CLIENT_DEFAULT_REQUEST_TIMEOUT_MS`, default `30000`
- `CLICKHOUSE_SERVER_TIMEOUT_GRACE_SECONDS`, default `5`

These could live in `packages/shared/src/env.ts` and then be used by the ClickHouse client manager.

## Backward compatibility

The defaults would stay the same, so existing deployments should behave exactly as they do today unless someone opts in by setting the env vars.

Also, explicit `opts.request_timeout` values would still take precedence, so the existing per-query overrides would continue to work.

## Possible implementation

Something like this in `packages/shared/src/env.ts`:

```ts
CLICKHOUSE_CLIENT_DEFAULT_REQUEST_TIMEOUT_MS: z.coerce
  .number()
  .int()
  .positive()
  .default(30_000),

CLICKHOUSE_SERVER_TIMEOUT_GRACE_SECONDS: z.coerce
  .number()
  .int()
  .nonnegative()
  .default(5),

kshen0414:
## What does this PR do?

Closes #17688

This PR makes the default ClickHouse timeout settings configurable for self-hosted deployments.

Currently, the ClickHouse client default request timeout and server timeout grace period are hardcoded in `packages/shared/src/server/clickhouse/client.ts`. That makes it hard for self-hosted operators using the official Docker image or deployment manifests to tune timeout budgets without maintaining a custom build.

This change adds two env vars with the same defaults as today:

- `CLICKHOUSE_CLIENT_DEFAULT_REQUEST_TIMEOUT_MS`, default `30000`
- `CLICKHOUSE_SERVER_TIMEOUT_GRACE_SECONDS`, default `5`

Existing behavior is preserved unless these env vars are explicitly set. Explicit `opts.request_timeout` values still take precedence.

## Type of change

## Mandatory Tasks

## Checklist

- [x] I have read the [contributing guide](https://github.com/langfuse/langfuse/blob/main/CONTRIBUTING.md)
- [x] My code follows the style guidelines of this project
- [x] I have checked if my PR needs changes to the documentation
- [x] I have added tests that prove my feature works

<h2><a href="https://app.greptile.com/api/retrigger?id=66569088"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until the configured default is passed to the ClickHouse client's actual request timeout.

This PR introduces validated environment settings for the default ClickHouse client request timeout and server-side grace period, then uses them while constructing shared ClickHouse clients.

- Preserves the existing 30-second request budget and 5-second grace defaults.
- Allows self-hosted deployments to configure both values.
- Adds coverage for deriving the server execution limit from customized values.
- The resolved default is currently not applied to the client-level request timeout, leaving the primary configuration feature incomplete.
</details>

<sub>Reviews (1) · Last reviewed commit: ["feat(shared): make ClickHouse timeout de..."](https://github.com/langfuse/langfuse/commit/d95b6262f918844d1939b886cdc129c35a4249bf)</sub>


Steffen911:
@claude review

Steffen911:
Closing per the discussion in the corresponding issue.

PR #17692, opened 2026-09-20
fix(table): clear stale rows when switching filter scope

nkabardin:
🟡 **Live preview: [pr-17692 app preview](https://pr-17692.preview.langfuse.com)**

**TL;DR:** Changing a saved view, filter, or search clears rows from the previous result set. Refreshing, sorting, and paging still update in place.

**Why:** A customer reported rows lingering from earlier views during loading. The table refresh behavior introduced in [#15984](https://github.com/langfuse/langfuse/pull/15984) retained previous results across every query change, including changes to filters.

**What:** Share one scope-aware placeholder policy across events, traces, observations, and sessions, including their counts and dependent row data. Explicit view selections reset pagination, the live-window pin, and selection. Hide stale approximate totals while a new filter's count loads; retain valid counts when switching Builder/Query modes. Preserve pagination when opening bookmarked views.

**Result:** A rapid A → B → C switch shows loading instead of reviving A's rows. Existing rows remain visible during refresh and pagination.

<details>
<summary>Mechanism, observed before/after, and verification</summary>

Package: `web`. No API or database changes.

The shared policy compares the current semantic scope with the last populated query's scope: project, effective filters, search, and the selected time range. Page, sort, and automatically resolved/pinned timestamps are excluded. Dependent queries use the same policy. The root-observation fallback carries its own scope so its disabled observer cannot relabel another view's cached rows. Approximate totals compare their own count-affecting filters, time bounds, project, and root flag; changing only facet columns retains the count and changing filters retains facet choices while clearing the count.

Observed with synthetic data and delayed row requests on both v4 Tracing and v3 Traces:

| Action | Before | After |
| --- | --- | --- |
| Select `draft-reply` after `classify-intent` | Old `classify-intent` rows remain during fetch | Skeleton rows until the new result arrives |
| Select `lookup-order` while the second view is pending | Rows from two views ago remain | No old rows |
| Select a view from page 3 with a selected row | Page 3 and the old time pin persist | Page 1, pin released, selection cleared |
| Page or manually refresh | Rows remain | Rows remain |
| 30-second auto-refresh | Existing behavior preserved | Two row requests observed, zero blank-table samples |

Local reproduction data:

```sh
pnpm run seed -- session-variety --sessions 100 --days 1 --id-prefix pc136
pnpm run seed -- many-traces --count 180 --days 1 --id-prefix pc136
```

Create saved views using `name = classify-intent`, `name = draft-reply`, and `name = lookup-order`. Delay row requests in DevTools, switch between views before a fetch completes, and repeat from page 3 with a row selected. Then enable 30-second auto-refresh and page through the same view.

Verification:

- Targeted client suites: `Test Files 7 passed (7)`; `Tests 47 passed (47)`. Scope, fallback-observer, footer, and view-reset regressions were confirmed failing before the corresponding fixes.
- Uncached repository lint: `Tasks: 9 successful, 9 total`; `Cached: 0 cached, 9 total`.
- Uncached web typecheck with dependency builds: `Tasks: 4 successful, 4 total`; `Cached: 0 cached, 4 total`.
- Browser: `BROWSER CHECKS 13 passed; page errors []` on the main tracing paths, plus `SEARCH CHECKS 2 passed; page errors []` for direct filter/free-text edits. Legacy Observations and Sessions on both read paths each passed delayed view changes, page/selection resets, and pagination without blanking; no browser errors. All automation browsers and the local server were stopped afterward.
- `pnpm exec knip`: exit 0. Prettier and `git diff --check` passed.

The broader local typecheck was interrupted by the native addon's process-inspection command being blocked in the sandbox; the scoped web typecheck and its dependency builds passed.

</details>

🤖 Generated with [Codex](https://openai.com/codex)

<h2><a href="https://app.greptile.com/api/retrigger?id=66606313"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no actionable correctness, security, or repository-rule failures were identified.

This PR introduces a shared, scope-aware React Query placeholder policy so tables retain rows during refresh, sorting, and pagination but clear them when project, filters, search, or selected time range changes.
- Applies the policy consistently to events, traces, observations, sessions, counts, metrics, and dependent I/O queries.
- Resets pagination, live-window pins, and row selection for explicit saved-view and system-preset changes while preserving bookmarked-view pagination.
- Suppresses stale approximate totals while replacement counts are loading.
- Adds regression coverage for rapid scope switching, dependent-query metadata, view resets, and pagination-pin release.
</details>

```mermaid
flowchart LR
  A[Table query changes] --> B{Semantic scope unchanged?}
  B -->|Yes: page, sort, refresh, resolved pin| C[Retain previous rows as placeholder]
  B -->|No: project, filter, search, selected time range| D[Clear previous rows]
  C --> E[Replace in place when response arrives]
  D --> F[Show loading state]
  F --> E
  G[Explicit saved-view selection] --> H[Reset pagination, pin, and selection]
  H --> D
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(table): clear stale rows when switch..."](https://github.com/langfuse/langfuse/commit/449951aec1eca131d8c8825d3ded174dfde7a7a5)</sub>


ege-arhan:
Went through the placeholder unification here. Scoping placeholderData on project/filter/search/timeRange while deliberately leaving paging and sort out matches the intent — a page turn should not keep showing old rows, but a refetch under the same scope should. The new tablePlaceholderOptions helper makes the rows/count/metrics trio agree on that rule instead of the previous mix of placeholderData: prev => prev on rows and loading skeletons on count, which was exactly what paired new rows with an old count. One cost to keep an eye on: placeholderData now runs lodash isEqual over the whole FilterState on every background refetch, and timeRange/filter objects are rebuilt per render in these components, so this is a deep comparison on each pass rather than a cheap key check. Fine at current filter sizes, but if filter trees grow it may be worth memoizing the scope object per component so the common path short-circuits on reference equality before falling back to isEqual.

PR #17695, opened 2026-09-20
fix(shared): parse eligible JSON within size budget

ArpanBareja:
## Summary

- replace the all-or-nothing `deepParseJson` object-size bail-out with a per-string parse budget
- keep oversized string fields raw while still parsing smaller nested JSON siblings
- add shared parser regression coverage and update existing web JSON utility expectations

## Verification

- `pnpm --filter @langfuse/shared run test src/utils/json.test.ts`
  - `Test Files  1 passed (1)`
  - `Tests  9 passed (9)`
  - `Generated Prisma Client (v6.19.3)`
  - `tsc`
- `pnpm --filter web run test-client src/__tests__/json-utils.clienttest.ts`
  - `Test Files  1 passed (1)`
  - `Tests  106 passed (106)`
- `git diff --check`
  - no output

## Notes

Local Windows verification required temporarily bypassing workspace patch application and install scripts because pnpm failed to parse existing patch files on Windows and the root postinstall script produced a Windows path issue. The workspace config and lockfile were restored before committing; only parser and test files are included in this PR.

<h2><a href="https://app.greptile.com/api/retrigger?id=66700681"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge, with a non-blocking edge-case inconsistency between the recursive and iterative parsers.

This PR replaces the aggregate object-size bailout in the shared deep JSON parsers with a shared per-string parse-attempt budget.
- Smaller nested JSON strings can still be parsed when an oversized sibling is present.
- Recursive and iterative implementations now stop attempting strings after exhausting the budget.
- Regression tests cover oversized siblings and budget exhaustion.
- One edge-case inconsistency remains because the iterative parser charges discarded dangerous-key values against the budget.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(shared): parse eligible JSON within ..."](https://github.com/langfuse/langfuse/commit/36a1210f492fb68b9b32f53dd45552de03ac9c68)</sub>


ArpanBareja:
Additional context for reviewers:

## Problem

Large trace I/O can contain one oversized string field plus smaller stringified JSON siblings. Today `deepParseJson` checks the serialized size of the whole object up front and returns the entire object unchanged when it exceeds `maxSize`.

That creates a cliff: one large field prevents all smaller nested JSON strings from being parsed, so the Formatted view can show escaped JSON even when most fields are cheap and safe to parse. We observed this shape on a payload around ~523k characters, just above the 500k parse limit.

## Solution

This PR changes `maxSize` from a root-object bail-out into a per-string parse budget. Oversized strings stay raw, but smaller stringified JSON fields can still be parsed until the budget is spent.

This preserves the intent of the safety guard by bounding parsing work, while avoiding the all-or-nothing behavior that makes Formatted view degrade silently.

nkabardin:
@claude review

nkabardin:
We've re-implmented it in another PR with a bit of a different approach

PR #17786, opened 2026-09-22
fix(annotation-queues): audit only assignment changes that actually happened

MoonlightAndSunlight:
## What does this PR do?

Makes the annotation-queue assignment audit log record only assignment changes that actually
happened. Both directions were recording events that did not occur.

**`createAnnotationQueueAssignmentForApi` — a repeat POST logged a creation.** The handler upserts
with an empty `update`, so posting an assignment that already exists changes nothing, yet it wrote
an `action: "create"` entry every time. The `TODO` above that block named it: *"only create audit log
if upsert actually creates a new record"*. The handler now reads the assignment first and only
audits when it did not already exist; the upsert itself is untouched, so the concurrent-duplicate
handling it exists for still holds.

**`queueAssignmentRouter.removeAssignment` — a no-op delete logged a deletion.** The mutation calls
`deleteMany`, which does not throw when it matches nothing, and the audit entry was written
unconditionally. Removing an assignment that was not there therefore produced a `delete` entry. The
count is now checked before auditing.

Both matter for the same reason: the audit log is the record of who changed what, and an entry for
a change that never happened is indistinguishable from a real one.

## Consistency with the rest of the feature

These two paths were the outliers. Everywhere else the assignment audit already follows
"only audit a real change":

**Deletes.** There are three delete paths for an assignment and two of them are already guarded:

| Path | Mechanism | Missing assignment |
|---|---|---|
| MCP tool `deleteAnnotationQueueAssignment` | `delete()` via the service, catches `P2025` | throws `LangfuseNotFoundError`, no audit entry |
| Public API `deleteAnnotationQueueAssignmentForApi` | same, `P2025` treated as idempotent success | returns success, no audit entry |
| tRPC `queueAssignmentRouter.removeAssignment` | `deleteMany`, no `count` check | **audited a deletion that never happened** |

`#13914` (merged) is the precedent for caring about this: it made the MCP tool distinguish "removed"
from "was not there" instead of silently reporting success. The tRPC mutation is deliberately
idempotent — it returns `success: true` either way — so this PR does not change its return value or
turn a no-op delete into an error. It only stops the audit entry. Putting the guard on `count` keeps
that idempotency intact.

**Creates.** `gatewayConfigService` handles the same upsert-plus-audit shape by choosing
`action: before ? "update" : "create"`. Since the `update` here is empty there is no state change
to record at all on a repeat call, so skipping the entry is the equivalent treatment.

I checked the other `create` audits in `publicAnnotationQueueService` (for `annotationQueue` and
`annotationQueueItem`): both follow real `prisma.*.create()` calls — and the `annotationQueue` one
already handles its unique-constraint conflict explicitly — so they are correct as they stand and
are left alone.

No linked issue: I did not find an existing report for this behaviour.

## Type of change

## Mandatory Tasks

## Checklist

Left empty on purpose: every bullet in the template is phrased as something that was *not* done, and
I removed the ones that did not apply.

**On tests.** The create path is covered.
`annotation-queue-assignments-api.servertest.ts` already had a *"should handle duplicate assignment
creation gracefully"* case asserting that a repeat POST leaves exactly one assignment row, so I
extended that case with the audit assertion instead of adding a new one — it now fails against the
old behaviour, where the second POST wrote a second `create` entry.

The delete path is not covered: nothing in the suite exercises
`queueAssignmentRouter.removeAssignment` (the existing annotation-queue tRPC tests only touch items),
and a tRPC harness for one assertion would be a file of its own. Say the word if you want it and I
will add it.


MoonlightAndSunlight:
Addressed in `ece2e0780b22`.

You're right that a read followed by an upsert cannot answer "did this call insert?". Two
concurrent calls can both observe no row while the unique constraint lets only one of them
insert, so both wrote a `create` event; and a delete landing between the read and the upsert
could drop a real one.

The audit decision now comes from the atomic write result: `createMany({ skipDuplicates: true })`
returns the inserted count, and the entry is written only when that count is non-zero. The row is
re-read only inside that branch, to get the id for `resourceId`.

PR #17788, opened 2026-09-22
fix(datasets): return early when deleteDatasetRuns matches no runs

MoonlightAndSunlight:
## What does this PR do?

`deleteDatasetRuns` falls back to `datasetRuns[0].datasetId` when filling in the queue payload,
but `datasetId` is optional in the input schema. If the caller omits it and none of the supplied
`datasetRunIds` match — the runs were already deleted, or they belong to another project —
`findMany` comes back empty, `datasetRuns[0]` is `undefined`, and reading `.datasetId` throws. The
request fails instead of reporting that there was nothing to delete.

The handler now returns as soon as the lookup comes back empty. Nothing is deleted, queued or
audited either way, so the early return only removes the crash; the response shape is unchanged.

## Why return early instead of raising

The mutation is already idempotent in the other direction: it returns the runs it found, and its
audit entries — which are correct as written, since they are built from the `findMany` result rather
than from the input — simply do not exist when nothing matched. Turning "nothing matched" into a 404
would be a behaviour change beyond the crash this PR is fixing, so the existing contract is kept.

## Type of change

## Mandatory Tasks

## Checklist

Left empty on purpose: every bullet in the template is phrased as something that was *not* done, and
I removed the ones that did not apply.

No test added: nothing in the suite exercises `deleteDatasetRuns`, and pinning this would need a
seeded dataset run plus a tRPC caller. Happy to add one if you want it.

<h2><a href="https://app.greptile.com/api/retrigger?id=67482469"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until the empty-result guard preserves cleanup retries when `datasetId` is supplied.

This PR prevents `deleteDatasetRuns` from dereferencing an absent first result when no runs match. However, the new unconditional return also bypasses a valid asynchronous-cleanup retry path when the caller supplied a dataset ID.
- Returns an empty run array for no-match requests.
- Avoids the undefined `datasetRuns[0]` access.
- Needs to preserve cleanup enqueueing when enough input is available after a partial deletion failure.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(datasets): return early when deleteD..."](https://github.com/langfuse/langfuse/commit/1c6ad819c7a117555cb5f6b545fcf975e1ccf9fd)</sub>


MoonlightAndSunlight:
Addressed in `d4db147c6dbd`.

Correct — returning early skipped the enqueue, so a retry after a committed delete whose enqueue
failed would leave the ClickHouse side undeleted for good. The early return now enqueues the cleanup
first when the caller supplied `datasetId`, and only returns without enqueueing when there is no
datasetId to enqueue with — which is exactly the case that would otherwise have crashed on
`datasetRuns[0]`.

Re-enqueueing is safe: the queue handler deletes by run id.

PR #17792, opened 2026-09-22
fix(api): declare OpenAPI schemas for classic public API errors

### LINKED ISSUE #17625: Error responses are undeclared for 400/401/403/404/405 across every operation
The published OpenAPI document declares `schema: {}` for every client-facing error status on
every operation, which leaves API consumers with no contract for error payloads.

### Numbers

Measured against the deployed spec at `https://cloud.langfuse.com/generated/api/openapi.yml`
(117 operations, 638 error responses total):

| Status | Undeclared (`schema: {}`) | Declared |
|--------|--------------------------|----------|
| 400    | 117                      | 0        |
| 401    | 117                      | 0        |
| 403    | 117                      | 0        |
| 404    | 117                      | 0        |
| 405    | 117                      | 0        |
| 409    | 0                        | 5        |
| 429    | 0                        | 24       |
| 500    | 0                        | 24       |

**585 of 638 error responses (91.7%) declare no schema.** The split is by *status code*, not by
API area: 400/401/403/404/405 are undeclared on 100% of operations, including the same
operations that correctly declare `PublicApiError` for 409/429/500.

### What that looks like

Undeclared, e.g. `annotationQueues_getQueue`:

```yaml
'404':
  description: ''
  content:
    application/json:
      schema: {}
```

Declared, on the evaluators / evaluation-rules endpoints:

```yaml
'409':
  description: ''
  content:
    application/json:
      schema:
        $ref: '#/components/schemas/PublicApiError'
```

Note the `description: ''` on both — descriptions are empty throughout as well.

### Two duplicate error schemas

`PublicApiError` and `unstablePublicApiError` are structurally identical
(`message` + `code` + `details`, `message`/`code` required) but are separate types with separate
code enums. 25 responses reference the first, 28 the second.

### Consumer impact

Servers do send a useful body — it just isn't described. Two shapes observed in practice:

```json
{"message": "Dataset X not found"}
{"message": "No model with this id found. Note: You cannot delete built-in models...",
 "error": "LangfuseNotFoundError"}
```

Note `error` here versus `code` in `PublicApiError` — two names for the same concept, neither
declared for these statuses.

Because none of this is in the contract, every client must hand-write a parser and guess at field
names. We did exactly that in the Quarkus extension, in
https://github.com/quarkiverse/quarkus-langfuse/issues/94 — users reported that our exception
types "swallowed context", because the only thing we could surface was the undigested response
body as an exception message:

```
Langfuse API error (404): {"message":"Dataset X not found"}
```

The workaround is a hand-written parser
(https://github.com/quarkiverse/quarkus-langfuse/pull/105): read the body as a string, check
`Content-Type` before parsing, walk a JSON tree, try `message`, then try `code` and fall back to
`error`. None of that is derivable from the spec, so it silently rots whenever the server's
wording changes — a renamed field degrades the client with no compile error and no failing test.
A declared error schema would have made the whole thing a non-event: the generator would have
emitted a type and we would have deserialized into it.

### Suggested fix

1. Declare a shared error schema and reference it from the 400/401/403/404/405 responses, so
   generated clients get a real type instead of `{}`.
2. Consolidate `PublicApiError` and `unstablePublicApiError`, or document why they differ.
3. Reconcile `code` vs `error` as the machine-readable field name.
4. Populate the `description` fields.

Item 1 alone would let every generated client deserialize errors into a real type instead of
reverse-engineering them.

cs-fisha:
## Summary
- Declares Fern `ApiError` (`{ message, error }`) for shared 400/401/403/404/405 error responses so the published OpenAPI no longer uses `schema: {}` on those statuses.
- Matches the existing classic public-API runtime payload from `withMiddlewares` / auth routes; does **not** migrate those statuses onto `PublicApiError` (`message` + `code`).
- Regenerates `web/public/generated/api/openapi.yml` and adds a server-unit regression test.

Fixes #17625

## Test plan
- [x] `pnpm run openapi:check` (export + assert generated current)
- [x] `cd web && npx dotenv -e ../.env.test -e ../.env -- vitest run --project server-unit src/__tests__/server/unit/openapi-error-schemas.servertest.ts src/__tests__/server/unit/openapi-deprecations.servertest.ts src/__tests__/server/unit/openapi-union-variant-titles.servertest.ts` → 3 files / 27 tests passed
- [x] Spot-check generated OpenAPI: 400/401/403/404/405 use `$ref: '#/components/schemas/ApiError'` (0× `schema: {}`; 585× ApiError refs)

<h2><a href="https://app.greptile.com/api/retrigger?id=67520244"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until the documented 404 schema is reconciled with the reachable events-only response body.

This PR replaces unspecified classic public-API error schemas with a generated `ApiError` component and adds regression coverage for the resulting OpenAPI references.
- Models classic 400/401/403/404/405 responses as `{ message, error }`.
- Regenerates the published OpenAPI document.
- Adds a server-unit test covering the component and its response references.
- One reachable events-only 404 response does not satisfy the newly required schema.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(api): declare OpenAPI schemas for cl..."](https://github.com/langfuse/langfuse/commit/ca7b45410f82f248d3974a16a8c75bd91c7000ee)</sub>


cs-fisha:
Addressed the Greptile note on the events_only 404 body: `rejectInEventsOnlyMode` now returns the classic `{ message, error }` ApiError shape (`error: LangfuseNotFoundError`), matching `withMiddlewares` and the OpenAPI schema, instead of making `error` optional.

